PR #8866, opened 2018-06-27
data for upgradeSeriesLock doc

hmlanigan:
part of managed series upgrades: add to-series, from-series, prepareStatus, completeStatus and units to upgradeSeriesLock

qa steps;

juju bootstrap
export JUJU_DEV_FEATURE_FLAGS=upgrade-series
juju deploy ghost --series trusty
juju upgrade-series prepare 0 xenial
juju-db.bash
juju:PRIMARY> db.machineUpgradeSeriesLocks.find()
verify doc has the new data.

### DIFF
--- a/state/machine.go
+++ b/state/machine.go
@@ -217,7 +217,10 @@ type instanceData struct {
 // upgradeSeriesLock holds the attributes relevant to lock a machine during a
 // series update of a machine
 type upgradeSeriesLock struct {
-	Id string
+	Id         string   `bson:"machineid"`
+	ToSeries   string   `bson:"toseries"`
+	FromSeries string   `bson:"fromseries"`
+	Units      []string `bson:"units"`
 }
 
 func hardwareCharacteristics(instData instanceData) *instance.HardwareCharacteristics {
@@ -2061,7 +2064,7 @@ func (m *Machine) VerifyUnitsSeries(unitNames []string, series string, force boo
 // this item exists in the database for a given machine it indicates that a
 // machine's operating system is being upgraded from one series to another - for
 // example from xenial to bionic.
-func (m *Machine) CreateUpgradeSeriesLock() error {
+func (m *Machine) CreateUpgradeSeriesLock(unitNames []string, toSeries string) error {
 	buildTxn := func(attempt int) ([]txn.Op, error) {
 		if attempt > 0 {
 			if err := m.Refresh(); err != nil {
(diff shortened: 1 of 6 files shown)


hmlanigan:
[prepare updates the locking document with the units that are involved, the original series, and requested series](https://trello.com/c/PtkhooQU/300-prepare-updates-the-locking-document-with-the-units-that-are-involved-the-original-series-and-requested-series)

state/machine.go
```diff
@@ -2076,10 +2079,28 @@ func (m *Machine) CreateUpgradeSeriesLock() error {
 			return nil, errors.AlreadyExistsf("upgrade series lock for machine %q", m)
 		}
 		if err = m.isStillAlive(); err != nil {
-			return nil, errors.Wrap(jujutxn.ErrNoOperations, err)
+			return nil, errors.Trace(err)
+		}
+		// Exit early if the Machine series doesn't need to change.
+		fromSeries := m.Series()
+		if fromSeries == toSeries {
+			return nil, errors.Trace(errors.Errorf("machine %s already at series %s", m.Id(), toSeries))
```
ExternalReality: I was wondering if there was a way to wrap these kinds of errors so as to have both the effect of the no transaction errors while still propagating up the relevant error message.

state/machine.go
```diff
@@ -217,7 +217,10 @@ type instanceData struct {
 // upgradeSeriesLock holds the attributes relevant to lock a machine during a
 // series update of a machine
 type upgradeSeriesLock struct {
-	Id string
+	Id         string   `bson:"machineid"`
+	ToSeries   string   `bson:"toseries"`
+	FromSeries string   `bson:"fromseries"`
+	Units      []string `bson:"units"`
```
ExternalReality: Would it be good to have a `status`  field added which is populated with some string enum (of legit statuses) value? The Uniter is interested in the status. Maybe something like below.

```go
//go:generate stringer -type SeriesUpgradeStatus
const (
	Preparing SeriesUpgradeStatus = iota
	Prepared
	Complete
)
```
``` go
 type upgradeSeriesLock struct {
 ...
 status  UpgradeSeriesStatus `bson: "status"`
```
@hmlanigan we should link up on this so that we can determine how the units behave with respect to this.

state/machine.go
hmlanigan: I was wondering the same thing, but haven't found one, there's a comment in txn.Run()
if len(ops) == 0 {
			// Treat this the same as ErrNoOperations but don't suppress other errors.
			return nil
}
However err == ErrNoOperations, returns nil also, not sure how it doesn't suppress other errors.

I think we'll have to some error handling on the call. Adding a todo in machinemanager.

state/machine.go
```diff
@@ -2092,6 +2113,25 @@ func (m *Machine) CreateUpgradeSeriesLock() error {
 	return nil
 }
 
+func (m *Machine) unitsHaveChanged(unitNames []string) (bool, error) {
```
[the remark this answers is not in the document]
hmlanigan: added tests for unitsHaveChanged(), infrastructure for the tests was in machine_test already.

hmlanigan:
@ExternalReality if you approve and will unblock you in the AM, please merge.

state/machine.go
manadart: Do you need to trace an error at its source?

state/machine.go
```diff
@@ -217,9 +217,42 @@ type instanceData struct {
...
 type upgradeSeriesLock struct {
-	Id string
+	Id             string                     `bson:"machineid"`
+	ToSeries       string                     `bson:"toseries"`
+	FromSeries     string                     `bson:"fromseries"`
+	PrepareStatus  MachineSeriesUpgradeStatus `bson:"preparestatus"`
+	PrepareUnits   []unitStatus               `bson:"prepareunits"`
+	CompleteStatus MachineSeriesUpgradeStatus `bson:"completestatus"`
+	CompleteUnits  []unitStatus               `bson:"completeunits"`
 }
```
ExternalReality: Oh, LGTM

hmlanigan:
$$merge$$

PR #8868, opened 2018-06-28
Use stateful set for pods with storage

wallyworld:
## Description of change

For units requiring storage, use a k8s StatefulSet instead of a Deployment Controller to manage the pods.
This PR required a regeneration of the mocks using go generate so ignore those file when reviewing.

The basis is to use a deployment controller for pods without storage and a statefulset for those with. The storage setup logic is essentially the same, but is tweaked a bit because the pv claim creation is not done  manually for stateful sets.

## QA steps

Deploy a test k8s charm with storage. Ensure a pv and pvc is properly set up and the specified mount point used.


caas/kubernetes/provider/k8s.go
```diff
@@ -178,9 +181,31 @@ func (k *kubernetesClient) EnsureOperator(appName, agentPath string, config *caa
...
+				Spec: *pvcSpec,
+			})
+			if err != nil {
+				return errors.Trace(err)
+			}
+			logger.Debugf("created new pvc: %+v", pvc)
+		}
+		storageVol = &core.Volume{Name: volName}
+		storageVol.PersistentVolumeClaim = &core.PersistentVolumeClaimVolumeSource{
+			ClaimName: pvc.Name,
```
ycliuhw: Do we need to handle `NotFound` here?

caas/kubernetes/provider/k8s.go
```diff
@@ -468,17 +481,19 @@ func (k *kubernetesClient) EnsureService(
...
-		return errors.Annotatef(err, "configuring storage for %s", appName)
-	}
-
 	// Add a deployment controller configured to create the specified number of units/pods.
 	numPods := int32(numUnits)
-	if err := k.configureDeployment(appName, unitSpec, params.PodSpec.Containers, &numPods); err != nil {
-		return errors.Annotate(err, "creating or updating deployment controller")
+	if len(params.Filesystems) > 0 {
+		if err := k.configureStatefulSet(appName, unitSpec, params.PodSpec.Containers, &numPods, params.Filesystems); err != nil {
+			return errors.Annotate(err, "creating or updating stateful set")
```
ycliuhw: is it `the StatefulSet` here?

ycliuhw:
It's much better to use `StatefulSet` to ensure our stateful operator app. 
Nice!

caas/kubernetes/provider/k8s.go
wallyworld: yeah, so if the unit requires storage, we use a stateful set
if there's no storage required, we continue to use a deployment controller as it's lighter weight

wallyworld:
$$merge$$

PR #8871, opened 2018-06-28
Make the txn queue patch apply on 2.3.

jameinel:
## Description of change

The max-txn-queue patch was not being applied because it was named '.patch' and not '.diff'. Someone else has already noticed and fixed it in 2.4, but we didn't fix it in 2.3. And thus https://bugs.launchpad.net/juju/+bug/1778907 was opened.

While there, I noticed that applying the diff gives:
```
patching file gopkg.in/mgo.v2/txn/flusher.go
Hunk #3 succeeded at 288 (offset 15 lines).
Hunk #4 succeeded at 350 (offset 15 lines).
Hunk #5 succeeded at 492 (offset 15 lines).
Hunk #6 succeeded at 556 (offset 15 lines).
Hunk #7 succeeded at 950 (offset 15 lines).
```

So I fixed up the diff so that it would apply cleanly instead of needing an offset.

## QA steps

```
$ make add-patches
or
$ make release-build
```
Should both result in a patches source tree that has the MaxTxnQueueLength variable as patch of `gopkg.in/mgo.v2`.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1778907

### DIFF
--- a/patches/max_txn_queue_length_pr463.diff
+++ b/patches/max_txn_queue_length_pr463.diff
@@ -1,6 +1,6 @@
 --- a/gopkg.in/mgo.v2/txn/flusher.go
 +++ b/gopkg.in/mgo.v2/txn/flusher.go
-@@ -244,6 +244,21 @@ NextDoc:
+@@ -270,6 +270,21 @@
  		change.Upsert = false
  		chaos("")
  		if _, err := cquery.Apply(change, &info); err == nil {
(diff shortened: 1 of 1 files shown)


jameinel:
$$merge$$

jameinel:
[1778907 txn-queue patch not in 2.3](https://trello.com/c/y5lQj1xQ/318-1778907-txn-queue-patch-not-in-23)

PR #8872, opened 2018-06-28
2.3 into 2.4

jameinel:
## Description of change

Bring 2.4 up-to-date with the tip of 2.3.

This brings in a small patch to fix the exact lines of one of our dependency patches, and one line of code difference that makes 2.4 buildable with go < 1.9 (introducing type aliases).

## QA steps

Should still build and pass.

## Documentation changes

None.

## Bug reference

See individual patches.


jameinel:
$$merge$$ self approving 2.3 into 2.4

jameinel:
$$merge$$ try again

PR #8873, opened 2018-06-28
WIP: Gosimple linter

SimonRichardson:
## Description of change

A new additional gosimple linter. The aim of this linter is to
reduce redundant code, either by using the defaults implied by
the language (think `copy`, `append`) or using reducing the 
number of `if err != nil {` checks.

## QA steps

Check that it passes all the tests.

## Documentation changes

No changes are involved, but you can run it independently using:

```sh
./scripts/gometalinter
```

## Bug reference

N/A


SimonRichardson:
I'll attempt this another time.

PR #8880, opened 2018-07-01
Suggestion for agree flag test.

ExternalReality:
Suggestion for how to test the agree flag.

### DIFF
--- a/cmd/juju/machine/upgradeseries_test.go
+++ b/cmd/juju/machine/upgradeseries_test.go
@@ -6,6 +6,7 @@ package machine_test
 import (
 	"bytes"
 	"errors"
+	"fmt"
 
 	"github.com/golang/mock/gomock"
 	"github.com/juju/cmd"
@@ -18,6 +19,7 @@ import (
 )
 
 type UpgradeSeriesSuite struct {
+	ctx *cmd.Context
 }
 
 var _ = gc.Suite(&UpgradeSeriesSuite{})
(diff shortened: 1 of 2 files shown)


ExternalReality:
[follow on to PR 8847](https://trello.com/c/w7pdaDLH/321-follow-on-to-pr-8847)

cmd/juju/machine/upgradeseries_test.go
```diff
@@ -144,7 +109,23 @@ func (s *UpgradeSeriesSuite) TestCompleteCommandDoesNotAcceptSeries(c *gc.C) {
...
+}
+
+func (s *UpgradeSeriesSuite) TestPrepareCommandShouldAcceptAgreeAndNotPrompt(c *gc.C) {
+	err := s.runUpgradeSeriesCommandWithConfirmation(c, "n", machine.PrepareCommand, machineArg, seriesArg, "--agree")
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(s.ctx.Stdout.(*bytes.Buffer).String(), gc.Equals, ``)
+}
+
+func stubRemoteAPI(c *gc.C) *mocks.MockUpgradeMachineSeriesAPI {
+	mockController := gomock.NewController(c)
```
manadart: Even though this is for a stub and the only expectations use ```AnyTimes()```, I would still call ```Finish()``` on the controller at some point to ensure you did not get calls had no accompanying expectation (I *think* unexpected methods with no return would not break tests here).

I generally either return the controller from, or pass one into such methods.

ExternalReality:
!!build!!

hmlanigan:
LGTM, curious about the outcome from mandart's comment.

cmd/juju/machine/upgradeseries_test.go
ExternalReality: I don't agree

PR #8886, opened 2018-07-03
Make raft upgrade step idempotent

babbageclunk:
## Description of change
Upgrade steps should be idempotent because they get rerun when upgrading from (for example) pre-release versions or earlier minor versions. In this case the upgrade step to bootstrap the raft cluster needed to check that it hadn't already been done, either by bootstrapping or upgrading to 2.4-rc3. 

## QA steps
* Bootstrap with 2.4-rc3 - this will initialise the raft cluster. Then upgrade to a version with this fix. The upgrade will skip trying to bootstrap the raft cluster again, so the upgrade should succeed and raft should still be running.
* Bootstrap with 2.3.8 and upgrade to a version with this fix - the upgrade should bootstrap the raft cluster.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1779904

### DIFF
--- a/upgrades/raft.go
+++ b/upgrades/raft.go
@@ -5,6 +5,7 @@ package upgrades
 
 import (
 	"net"
+	"os"
 	"path/filepath"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


upgrades/raft.go
```diff
@@ -23,6 +24,13 @@ const jujuMachineKey = "juju-machine-id"
 // being upgraded.
 func BootstrapRaft(context Context) error {
 	agentConfig := context.AgentConfig()
+	storageDir := filepath.Join(agentConfig.DataDir(), "raft")
```
howbazaar: Do you think that just looking for the directory is sufficient?

babbageclunk: Yes - that directory is created in the process of creating the logstore or snapshot store, immediately before the bootstrapping. The only way the directory could exist without the raft cluster having been bootstrapped is if there was catastrophic error while bootstrapping, in which case the bootstrap will have failed (and the controller would have been torn down or the upgrade will have failed).

babbageclunk:
$$merge$$

PR #8893, opened 2018-07-04
LXD Default logger

SimonRichardson:
## Description of change

Provides the current lxdclient logger, which overrides the lxd shared logger to the
new lxd/container package.

This currently uses `init()`, in an ideal scenario, this wouldn't do it this way, but it
makes sure that we do overwrite the shared lxd logger at the best time.

## QA steps

Logs from the lxd logger should be formatted correctly or at least in the exact same
way as they where previously.

## Documentation changes

No changes

## Bug reference

N/A

### DIFF
--- a/container/lxd/logger.go
+++ b/container/lxd/logger.go
@@ -1,0 +1,80 @@
+// Copyright 2016 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package lxd
+
+import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


SimonRichardson:
https://trello.com/c/nULl0hV7/336-lxd-custom-logging

SimonRichardson:
Yeah, that is ugly

SimonRichardson:
$$merge$$

PR #8894, opened 2018-07-04
Check for supported versions

SimonRichardson:
## Description of change

The following checks for the servers supported version, before it
is being used by the provider. It follows the old semantics of not
checking for simple streams.

I'm unsure if this is in the correct place to check, or indeed if
we should be checking the type instead of using the connection info,
but I thought it would be a good PR start

## QA steps

Boot with an incompatible version of LXD and see the warning messages.

## Documentation changes

Should respond with the same logs as previous setup

## Bug reference

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ pre-check:
 	@$(PROJECT_DIR)/scripts/verify.bash
 
 check: godeps pre-check
-	go test $(CHECK_ARGS) -test.timeout=$(TEST_TIMEOUT) $(PROJECT)/...
+	go test $(CHECK_ARGS) -test.timeout=$(TEST_TIMEOUT) $(PROJECT)/... -check.v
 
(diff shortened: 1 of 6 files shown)


SimonRichardson:
https://trello.com/c/pOV38qPO/337-lxd-supported-versions

SimonRichardson:
I've reworked this, so we check on bootstrap, which makes it a lot more consistent.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #8895, opened 2018-07-05
leases: rework lease code in preparation for raft lease store

babbageclunk:
## Description of change

In preparation for a raft-based lease implementation we need to make some changes to the way the current lease code works.

The first change is just a cosmetic one. To make its use clearer, I've renamed core/lease.Client to core/lease.Store - this better reflects how it works in our lease code. (This is the second commit.)

The lease manager won't be a state worker any more, so it needs to manage leases for both namespaces across all models. That means that the core/lease.Store interface methods now take a key (namespace, model UUID and lease name) instead of just a name. (This change is in the first commit.)

The final change is to pull the Checker and Claimer interface methods off the lease manager. Those interfaces don't specify namespace and model, and we don't really want to make Checker and Claimer take keys instead of names (since that won't be convenient from the caller side). Instead worker/lease.Manager now exposes methods to get a Checker or a Claimer for a specific namespace (i.e. singular or leadership) and model.

One consequence of this is that the manager needs to be able to validate lease requests for different namespaces. Instead of accepting a single secretary to do this validation, it now gets a function that returns the secretary to be used for a given namespace.

## QA steps

No functional changes here - smoke test that bootstrapping and deploying multiple units for applications works correctly.

### DIFF
--- a/state/singular.go
+++ b/state/singular.go
@@ -54,7 +54,8 @@ func (s singularSecretary) CheckDuration(duration time.Duration) error {
 // SingularClaimer returns a lease.Claimer representing the exclusive right to
 // manage the model.
 func (st *State) SingularClaimer() corelease.Claimer {
-	return lazyLeaseManager{func() *lease.Manager {
-		return st.workers.singularManager()
+	return lazyLeaseManager{func() (lease.CheckerClaimer, error) {
+		manager := st.workers.singularManager()
+		return manager.Bind(singularControllerNamespace, st.modelUUID())
 	}}
 }
--- a/state/singular.go
+++ b/state/singular.go
@@ -9,8 +9,7 @@ import (
 	"github.com/juju/errors"
 	"gopkg.in/juju/names.v2"
 
-	corelease "github.com/juju/juju/core/lease"
-	"github.com/juju/juju/worker/lease"
+	"github.com/juju/juju/core/lease"
 )
 
 // singularSecretary implements lease.Secretary to restrict claims to either
@@ -53,9 +52,9 @@ func (s singularSecretary) CheckDuration(duration time.Duration) error {
 
 // SingularClaimer returns a lease.Claimer representing the exclusive right to
 // manage the model.
-func (st *State) SingularClaimer() corelease.Claimer {
-	return lazyLeaseManager{func() (lease.CheckerClaimer, error) {
+func (st *State) SingularClaimer() lease.Claimer {
+	return lazyLeaseClaimer{func() (lease.Claimer, error) {
 		manager := st.workers.singularManager()
-		return manager.Bind(singularControllerNamespace, st.modelUUID())
+		return manager.Claimer(singularControllerNamespace, st.modelUUID())
 	}}
 }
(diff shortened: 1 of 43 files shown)


state/state.go
```diff
@@ -454,18 +454,18 @@ func (st *State) getLeaseClient(namespace string) (lease.Client, error) {
 		return nil, errors.Trace(err)
 	}
 
-	client, err := statelease.NewClient(statelease.ClientConfig{
-		Id:          st.leaseClientId,
+	store, err := statelease.NewStore(statelease.StoreConfig{
```
wallyworld: Instead of having to add a new ModelUUID() method to environMongo, I think it's cleaner just to add a model-uuid attribute to StoreConfig. namespace and model-uuid are both used to make the lease key and having these together as attributes on the config struct makes most sense.  The Mongo abstraction should just continue to be the way to get to the database.

state/singular.go
```diff
@@ -54,7 +54,8 @@ func (s singularSecretary) CheckDuration(duration time.Duration) error {
 // SingularClaimer returns a lease.Claimer representing the exclusive right to
 // manage the model.
 func (st *State) SingularClaimer() corelease.Claimer {
-	return lazyLeaseManager{func() *lease.Manager {
-		return st.workers.singularManager()
+	return lazyLeaseManager{func() (lease.CheckerClaimer, error) {
```
wallyworld: I think Bind() needs to be a private method. We just need to expose Claimer() and Checker().
So here we'd just need to access the claim as that's all we're returning:

return lazyLeaseManager{func() (lease.Claimer, error) {
    manager := st.workers.singularManager()
    return manager.Claimer(singularControllerNamespace, st.modelUUID())
}

state/state.go
babbageclunk: Yeah, makes sense. I'll change that.

state/singular.go
babbageclunk: Yeah, you're right - also it's kind of an awkward name. I was thinking about doing that. I'll need to split `lazyLeaseManager` into `lazyLeaseClaimer` and `lazyLeaseChecker`, but that's ok.

worker/lease/manager_cross_test.go
```diff
@@ -0,0 +1,313 @@
...
+			method: "Refresh",
+		}, {
+			method: "ExpireLease",
+			args:   []interface{}{lease1},
+			callback: func(leases map[corelease.Key]corelease.Info) {
+				delete(leases, lease1)
+			},
+		}},
+	}
+	fix.RunTest(c, func(manager *lease.Manager, clock *testing.Clock) {
```
wallyworld: this could be a common helper for wait across namespaces and wait across models

worker/lease/manager_cross_test.go
```diff
@@ -0,0 +1,313 @@
...
+				Holder:   "sgt-howie",
+				Expiry:   offset(time.Second),
+				Trapdoor: corelease.LockedTrapdoor,
+			},
+		},
+		expectCalls: []call{{
+			method: "Refresh",
+		}},
+	}
+	fix.RunTest(c, func(manager *lease.Manager, _ *testing.Clock) {
```
wallyworld: ditto

worker/lease/manager_cross_test.go
babbageclunk: Yeah, good call.

PR #8896, opened 2018-07-05
Introduce charmstore-url as controller config.

Veebers:
## Description of change
Adds the 'charmstore-url' config parameter to the controller enabling users 
to configure which charmstore they consume.

This value is used whenever a charmstore client is created. Some places
occur on the client side (and require a call out to the controller) others
occur on the controller itself.

## Documentation changes
There will need to be some extra command docs added to highlight this 
change: https://bugs.launchpad.net/juju/+bug/1780689

### DIFF
--- a/apiserver/facades/client/application/charmstore.go
+++ b/apiserver/facades/client/application/charmstore.go
@@ -66,8 +66,13 @@ func AddCharmWithAuthorization(st *state.State, args params.AddCharmWithAuthoriz
 		return nil
 	}
 
-	// Open a charm store client.
-	repo, err := openCSRepo(args)
+	// determine which charmstore api url to use.
+	controllerCfg, err := st.ControllerConfig()
+	if err != nil {
+		return err
+	}
+
+	repo, err := openCSRepo(controllerCfg.CharmStoreURL(), args)
 	if err != nil {
 		return err
 	}
@@ -133,8 +138,8 @@ func AddCharmWithAuthorization(st *state.State, args params.AddCharmWithAuthoriz
 	return StoreCharmArchive(st, ca)
 }
 
-func openCSRepo(args params.AddCharmWithAuthorization) (charmrepo.Interface, error) {
-	csClient, err := openCSClient(args)
+func openCSRepo(csURL string, args params.AddCharmWithAuthorization) (charmrepo.Interface, error) {
+	csClient, err := openCSClient(csURL, args)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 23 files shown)


apiserver/facades/client/application/charmstore.go
```diff
@@ -66,8 +66,13 @@ func AddCharmWithAuthorization(st *state.State, args params.AddCharmWithAuthoriz
 		return nil
 	}
 
-	// Open a charm store client.
-	repo, err := openCSRepo(args)
+	// determine which charmstore api url to use.
+	conConfig, err := st.ControllerConfig()
```
wallyworld: Small nit - elsewhere we tend to use controllerCfg
Nice to be consistent

charmstore/client.go
```diff
@@ -85,6 +85,7 @@ func makeWrapper(bakeryClient *httpbakery.Client, server *url.URL) csWrapper {
 	p := csclient.Params{
 		BakeryClient: bakeryClient,
 	}
+	// veebers: I don't think this would be nil anymore as we're being specific.
```
wallyworld: A code grep shows this is called by charmrevisionupdater which passes in nil.
So that needs to be fixed.
If we expect the url to not be nil, we need to check that and return an error if it is.

cmd/juju/application/deploy.go
```diff
@@ -1267,6 +1286,16 @@ func (c *DeployCommand) maybeReadCharmstoreBundleFn(apiRoot DeployAPI) func() (d
 	}
 }
 
+func (c *DeployCommand) getCharmStoreAPIURL(conAPIRoot api.Connection) (string, error) {
+	_, backend := base.NewClientFacade(conAPIRoot, "Controller")
```
wallyworld: This doesn't look right.
To get a controller facade client, we just do:
controller.NewClient(root)

controller/config.go
```diff
@@ -154,6 +157,9 @@ const (
 	// DefaultMaxTxnLogCollectionMB is the maximum size the txn log collection.
 	DefaultMaxTxnLogCollectionMB = 10 // 10 MB
 
+	// DefaultCharmstoreURL empty string will result in the default URL being used.
+	DefaultCharmStoreURL = ""
```
wallyworld: As discussed make this the real value.
Also need tests for this package

Veebers:
[Add charmstore API URL controller config.](https://trello.com/c/VWKhSzza/304-add-charmstore-api-url-controller-config)

Veebers:
$$merge$$

Veebers:
$$merge$$

pmatulis:
Is 'charmstore-url' something that a user can set a value to? Can it be passed with `bootstrap` and `controller-config`?

Veebers:
It must be set at bootstrap time (and cannot, currently, be changed after the fact).
Set like so: (note: the url includes api.)
```
juju bootstrap aws/ap-southeast-1 --config charmstore-url=https://api.staging.jujucharms.com/charmstore
```

PR #8899, opened 2018-07-05
Removing critical error

SimonRichardson:
## Description of change

The following removes the critical error, I'm not sure it should
be there, because we return the error directly under the log
statement.

An example of this error showing in the wrong place is when you've
bootstrapped to lxd on localhost and deploy something. When you
try and grab some metrics about that deploy and you miss type a
model name, it comes with a critical log.

## QA steps

```sh
juju deploy postgresql
juju metrics --all -m=lxc-test
```

The following shouldn't log critical error.

## Documentation changes

Nothing

## Bug reference

N/A


SimonRichardson:
!!build!!

PR #8900, opened 2018-07-05
Migrate storage docs to use hostid not machineid

wallyworld:
## Description of change

The 4 collections used to store volume and filesystem info are migrated to use "hostid" instead of "machineid". The bulk of the change is to write the upgrade step. There was an existing 2.2 upgrade step that needed to be modified to use bson maps instead of juju state structs (arguably all upgrade steps should be written this way anyway).

## QA steps

Deploy Juju 2.1 with postgresql using storage
Upgrade to 2.5
Check db to ensure migrations occurred
Check storage functionality

### DIFF
--- a/state/migration_export.go
+++ b/state/migration_export.go
@@ -1807,7 +1807,7 @@ func (e *exporter) addVolume(vol *volume, volAttachments []volumeAttachmentDoc) 
 		va := volumeAttachment{doc}
 		logger.Debugf("  attachment %#v", doc)
 		args := description.VolumeAttachmentArgs{
-			Machine: va.Machine(),
+			Machine: va.Host().(names.MachineTag),
 		}
 		if info, err := va.Info(); err == nil {
 			logger.Debugf("    info %#v", info)
--- a/state/upgrades.go
+++ b/state/upgrades.go
@@ -518,84 +518,127 @@ func addNonDetachableStorageMachineId(st *State) error {
 	if err != nil {
 		return errors.Trace(err)
 	}
+
+	getPool := func(d bson.M) string {
+		var pool string
+		info, _ := d["info"].(bson.M)
+		params, _ := d["params"].(bson.M)
+		if info != nil {
+			pool = info["pool"].(string)
+		} else if params != nil {
+			pool = params["pool"].(string)
+		}
+		return pool
+	}
+
+	var needsUpgradeTerm = bson.D{
+		{"machineid", bson.D{{"$exists", false}}},
+		{"hostid", bson.D{{"$exists", false}}},
+	}
 	var ops []txn.Op
-	volumes, err := sb.volumes(
-		bson.D{{"machineid", bson.D{{"$exists", false}}}},
-	)
-	if err != nil {
+
+	volumeColl, cleanup := st.db().GetCollection(volumesC)
+	defer cleanup()
+
+	var volData []bson.M
+	err = volumeColl.Find(needsUpgradeTerm).All(&volData)
+	if err != nil && err != mgo.ErrNotFound {
 		return errors.Trace(err)
 	}
-	for _, v := range volumes {
-		var pool string
-		if v.doc.Info != nil {
-			pool = v.doc.Info.Pool
-		} else if v.doc.Params != nil {
-			pool = v.doc.Params.Pool
-		}
-		detachable, err := isDetachableVolumePool(sb, pool)
+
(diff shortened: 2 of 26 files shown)


wallyworld:
!!build!!

state/migration_export.go
```diff
@@ -1807,7 +1807,7 @@ func (e *exporter) addVolume(vol *volume, volAttachments []volumeAttachmentDoc)
 		va := volumeAttachment{doc}
 		logger.Debugf("  attachment %#v", doc)
 		args := description.VolumeAttachmentArgs{
-			Machine: va.Machine(),
+			Machine: va.Host().(names.MachineTag),
```
babbageclunk: Is this just because export isn't supported for caas models yet? Shouldn't description be updated to use Host here too?

state/upgrades.go
```diff
@@ -518,84 +518,109 @@ func addNonDetachableStorageMachineId(st *State) error {
...
+		{"machineid", bson.D{{"$exists", false}}},
+		{"hostid", bson.D{{"$exists", false}}},
+	}
 	var ops []txn.Op
-	volumes, err := sb.volumes(
-		bson.D{{"machineid", bson.D{{"$exists", false}}}},
-	)
-	if err != nil {
+
+	volumeColl, cleanup := st.db().GetCollection(volumesC)
```
babbageclunk: Good call on rewriting this to use mgo datatypes.

state/upgrades.go
```diff
@@ -518,84 +518,109 @@ func addNonDetachableStorageMachineId(st *State) error {
...
 			return errors.Trace(err)
 		}
 		if detachable {
 			continue
 		}
-		attachments, err := sb.VolumeAttachments(v.VolumeTag())
-		if err != nil {
+
+		var attachInfo []bson.M
+		err = volumeAttachColl.Find(bson.D{{"volumeid", v["name"]}}).All(&attachInfo)
```
babbageclunk: Are we worried about the n+1 query structure here? Would it be worth getting all the volume attachments in one query and loading them into a map?

state/upgrades.go
```diff
@@ -518,84 +518,109 @@ func addNonDetachableStorageMachineId(st *State) error {
...
+		if detachable, err := isDetachableFilesystemPool(sb, getPool(f)); err != nil {
 			return errors.Trace(err)
 		} else if detachable {
 			continue
 		}
-		attachments, err := sb.FilesystemAttachments(f.FilesystemTag())
-		if err != nil {
+
+		var attachInfo []bson.M
+		err = filesystemAttachColl.Find(bson.D{{"filesystemid", f["filesystemid"]}}).All(&attachInfo)
```
babbageclunk: Same here.

state/upgrades.go
```diff
@@ -518,84 +518,109 @@ func addNonDetachableStorageMachineId(st *State) error {
...
+		if machineId == "" {
+			machineId = attachInfo[0]["hostid"]
+		}
 		ops = append(ops, txn.Op{
 			C:      volumesC,
-			Id:     v.doc.Name,
+			Id:     v["name"],
 			Assert: txn.DocExists,
 			Update: bson.D{{"$set", bson.D{
 				{"machineid", machineId},
```
babbageclunk: I was expecting you to set it to hostid here. Or is that coming in an upgrade step further down?

state/upgrades.go
```diff
@@ -1602,3 +1627,52 @@ func RemoveContainerImageStreamFromNonModelSettings(st *State) error {
...
+		}
+
+		for _, data := range storageData {
+			machineId := data["machineid"]
+			ops = append(ops, txn.Op{
+				C:      collName,
+				Id:     data["_id"],
+				Assert: txn.DocExists,
+				Update: bson.D{
+					{"$set", bson.D{{"hostid", machineId}}},
```
babbageclunk: Oh, yup.

state/migration_export.go
wallyworld: It will need to be updated at some point, but we need to rev the db description format etc. For  now, it's ok to leave as machine since we don't support migration yet

state/upgrades.go
wallyworld: Right, we need to go through the steps. These 2.2 steps should behave as before to be safe.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #8903, opened 2018-07-08
Stop units after prepare

ExternalReality:
After the `pre-series-upgrade` hook has completed successfully, the `uniter` will stop the processing of any further operations.

### DIFF
--- a/worker/uniter/resolver.go
+++ b/worker/uniter/resolver.go
@@ -52,10 +52,9 @@ func (s *uniterResolver) NextOp(
 		return nil, resolver.ErrTerminate
 	}
 
-	// The if the unit has completed its pre-series-upgrade hook (as noted
-	// by its completed state) then the uniter should idle in the face of
-	// all remote state changes. The unit is waiting to be shutdown by the
-	// machine in preparation for a series upgrade.
+	// If the unit has completed a pre-series-upgrade hook (as noted
+	// by its state) then the uniter should idle in the face of
+	// all remote state changes - the unit is waiting to be shutdown.
 	if localState.UpgradeSeriesStatus == model.UnitCompleted && remoteState.UpgradeSeriesStatus == model.UnitCompleted {
 		return nil, resolver.ErrNoOperation
 	}
--- a/worker/uniter/resolver.go
+++ b/worker/uniter/resolver.go
@@ -52,6 +52,14 @@ func (s *uniterResolver) NextOp(
 		return nil, resolver.ErrTerminate
 	}
 
+	// The if the unit has completed its pre-series-upgrade hook (as noted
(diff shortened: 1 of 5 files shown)


ExternalReality:
[stop unit agents at end of prepare command (once all prepare hooks complete successfully)](https://trello.com/c/QFjQQz9f/343-stop-unit-agents-at-end-of-prepare-command-once-all-prepare-hooks-complete-successfully)

worker/uniter/resolver.go
```diff
@@ -52,6 +52,13 @@ func (s *uniterResolver) NextOp(
 		return nil, resolver.ErrTerminate
 	}
 
+	// If the unit has completed a pre-series-upgrade hook (as noted
+	// by its state) then the uniter should idle in the face of
+	// all remote state changes - the unit is waiting to be shutdown.
+	if localState.UpgradeSeriesStatus == model.UnitCompleted && remoteState.UpgradeSeriesStatus == model.UnitCompleted {
```
ExternalReality: When the resolver loop hits this state, it should only return an `ErrNoOperation`. This should cause the Uniter to idle until the next state change. At this point, the uniter is essentially waiting to be shut down. Which is why the only state change it accepts is that which results in `ErrTerminate`.

worker/uniter/resolver/mock_test.go
```diff
@@ -86,7 +86,15 @@ func (e *mockOpExecutor) Run(op operation.Operation) error {
 
 type mockOp struct {
 	operation.Operation
-	commit func(operation.State) (*operation.State, error)
+	commit  func(operation.State) (*operation.State, error)
```
ExternalReality: This mock is cool, GoMock would certainly help here. We are mocking out the run of the operation. We are only interesting the the side effects of its wrappers (e.g. onCommit, onPrepare) which sync the local and remote state (among other things). Indeed most of these test simply test that the states is being synced correctly.

worker/uniter/resolver/opfactory.go
```diff
@@ -119,9 +119,16 @@ func (s *resolverOpFactory) wrapUpgradeOp(op operation.Operation, charmURL *char
 func (s *resolverOpFactory) wrapHookOp(op operation.Operation, info hook.Info) operation.Operation {
 	switch info.Kind {
 	case hooks.PreSeriesUpgrade:
```
ExternalReality: The on prepare wrapper and on commit wrappers sync the remote and local state. The actual execution of this hook sets the remote state. Which is why the two wrappers are free to set statuses against remote state without waiting for the `RemoteStateWatcher`.

ExternalReality:
$$MERGE$$

PR #8906, opened 2018-07-09
LXD: Validate the host and port of endpoint

SimonRichardson:
## Description of change

The following commit ensures that we have a port for the lxd endpoint.
That way we help people coming to use the lxd-remote flow, will get
a better experience.

## QA steps

Bootstrap a remote host, this should add the default port 8443 if it's not supplied.

## Documentation changes

Users using the CLI should be able to add a URL without the port and it should use the default port to make the user experience of using lxd remoting easier.

## Bug reference

N/A

### DIFF
--- a/container/lxd/connection.go
+++ b/container/lxd/connection.go
@@ -118,7 +118,13 @@ func ConnectLocal() (lxd.ContainerServer, error) {
 
 // ConnectRemote connects to LXD on a remote socket.
 func ConnectRemote(spec ServerSpec) (lxd.ContainerServer, error) {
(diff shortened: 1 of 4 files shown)


SimonRichardson:
https://trello.com/c/g3MFaUmU/345-lxd-ux-ui-ensure-that-there-is-a-default-port-on-the-host

SimonRichardson:
!!build!!

pmatulis:
@SimonRichardson For documentation purposes, I'm going to need more background and specifics. What is this really about in layman's terms?

PR #8907, opened 2018-07-09
WIP: LXD Autodetect clouds

SimonRichardson:
## Description of change

The following autodetects LXD clouds from the users local
clouds.yaml

It always puts the localhost yaml at zero index and there is no
sorting or filtering to prevent duplication.

## QA steps

**To follow, working on manual testing this.**

## Documentation changes

The following changes will read the local clouds.yaml to help detect 
remote lxd clouds. 

## Bug reference

N/A


SimonRichardson:
This isn't required as there is no scenario from the juju CLI you could get too, without changing the yaml files beneath it.

PR #8908, opened 2018-07-10
Add support for caas charms specifying docker command args

wallyworld:
## Description of change

Some k8s charms need to be able to specify the command args for the docker image instead of relying on the default entry point. The pod spec sent by the charm now supports that.
As a driveby, remove old ensure/delete unit code no longer used.

## QA steps

bootstrap a k8s model
deploy a caas charm which specifies command, args, and/or working dir

### DIFF
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -933,7 +933,7 @@ func (k *kubernetesClient) Units(appName string) ([]caas.Unit, error) {
 			unitInfo.FilesystemInfo = append(unitInfo.FilesystemInfo, caas.FilesystemInfo{
 				StorageName:  storageName,
 				Size:         uint64(vol.PersistentVolumeClaim.Size()),
-				FilesystemId: string(pvc.UID),
+				FilesystemId: pvc.Name,
 				MountPoint:   volMount.MountPath,
 				ReadOnly:     volMount.ReadOnly,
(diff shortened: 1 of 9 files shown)


caas/kubernetes/provider/k8s.go
```diff
@@ -933,7 +933,7 @@ func (k *kubernetesClient) Units(appName string) ([]caas.Unit, error) {
 			unitInfo.FilesystemInfo = append(unitInfo.FilesystemInfo, caas.FilesystemInfo{
 				StorageName:  storageName,
 				Size:         uint64(vol.PersistentVolumeClaim.Size()),
-				FilesystemId: string(pvc.UID),
+				FilesystemId: pvc.Name,
```
ycliuhw: not sure if the change from `UID` to `Name` impacts other places that use `FilesystemId` or not?

wallyworld: This was a driveby to improve the id displayed to the user - it's informational only.

ycliuhw: ic. yes, Name makes more sense for human.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #8912, opened 2018-07-10
Code changes for bundle clientfacade/cmd

vinu2003:
This PR addresses CLI facade code changes required for export bundle command which exports the configuration of current model in a YAML file.

## QA steps
Its a Work In Progress PR.
The idea is to execute 'juju export-bundle'

## Documentation changes
Needs Update for the new command.
## Bug reference
N/A


api/bundle/client.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Client allows access to the application API end point.
+type Client struct {
+	base.ClientFacade
+	st     base.APICallCloser
+	facade base.FacadeCaller
+}
+
+// NewClient creates a new client for accessing the bundle api.
+func NewClient(st base.APICallCloser) *Client {
```
howbazaar: I was trying to work out for a bit why we didn't have one of these before. Then it became apparent that the facade method was only ever exposed for the GUI to call in to get the changes for the bundle deploy.

api/bundle/client.go
```diff
@@ -0,0 +1,47 @@
...
+// Package bundle provides access to the bundle api facade.
+// This facade contains api calls that are specific to bundles.
+
+package bundle
+
+import (
+	"github.com/juju/loggo"
+	"gopkg.in/juju/names.v2"
+
+	"github.com/juju/errors"
```
howbazaar: juju/errors should be with the previous import block

api/bundle/client.go
```diff
@@ -0,0 +1,47 @@
...
+func NewClient(st base.APICallCloser) *Client {
+	frontend, backend := base.NewClientFacade(st, "Bundle")
+	return &Client{
+		ClientFacade: frontend,
+		st:           st,
+		facade:       backend}
+}
+
+// ExportBundle exports the current model configuration.
+func (c *Client) ExportBundle(tag names.ModelTag) (params.StringResult, error) {
```
howbazaar: Methods on the api wrappers should unpack the params structs. Ideally there would never be a params struct returned from an API method.

So in this case return `(string, error)`, and get the string value out of the StringResult struct.

howbazaar: Also... the concept to Tags was supposed to be primarily for the wire protocol for calling methods, and for accessing objects in the persistence layer.

There is no need to pass in a model tag because the API is limited to a model itself. There is no argument to pass. The ExportBundle method on the facade takes no args.

howbazaar: You also want an early exit if the best facade version < 2 with a nice error.

api/bundle/client_test.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+var _ = gc.Suite(&bundleMockSuite{})
+
+func newClient(f basetesting.APICallerFunc) *application.Client {
+	return application.NewClient(basetesting.BestVersionCaller{f, 2})
+}
+
+func (s *bundleMockSuite) TestExportBundle(c *gc.C) {
+	// TODO: vinu2003
```
howbazaar: Have one success test for facade version 2, and a failure for facade version 1.

cmd/juju/bundle/exportbundle.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/juju/errors"
+	"github.com/juju/juju/api/bundle"
+	"github.com/juju/juju/apiserver/params"
+	"github.com/juju/juju/cmd/modelcmd"
+	"github.com/juju/juju/cmd/output"
+)
+
+// NewexportbundleCommand returns a fully constructed exportbundle command.
+func NewexportbundleCommand() cmd.Command {
```
howbazaar: Need to camel case all the words `NewExportBundleCommand`.

cmd/juju/bundle/exportbundle.go
```diff
@@ -0,0 +1,105 @@
...
+	out cmd.Output
+	api ExportbundleModelAPI
+}
+
+const exportBundleHelpDoc = `
+Exports the current model configuration into a YAML file.
+
+Examples:
+
+    juju export-bundle
```
howbazaar: We may want to change the actual command name later. We are careful when we introduce any new verbs. But this is fine for now.

cmd/juju/bundle/exportbundle.go
```diff
@@ -0,0 +1,105 @@
...
+`
+
+// Info implements Command.
+func (c *exportBundleCommand) Info() *cmd.Info {
+	return &cmd.Info{
+		Name:    "export-bundle",
+		Purpose: "Exports the current model configuration in a YAML file.",
```
howbazaar: I'd avoid saying "YAML file" and instead say "charm bundle".

cmd/juju/bundle/exportbundle.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// NewexportbundleCommand returns a fully constructed exportbundle command.
+func NewexportbundleCommand() cmd.Command {
+	return modelcmd.Wrap(&exportBundleCommand{}, modelcmd.WrapSkipModelFlags)
+}
+
+type exportBundleCommand struct {
+	modelcmd.ModelCommandBase
+	out cmd.Output
```
howbazaar: We don't want a cmd.Output because bundles are never valid in any other format.

howbazaar: What we do probably want is a -f (--filename) option to write the bundle to a file.
Should default to ctx.Stdout.

cmd/juju/bundle/exportbundle.go
```diff
@@ -0,0 +1,105 @@
...
+
+// SetFlags implements Command.
+func (c *exportBundleCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.ModelCommandBase.SetFlags(f)
+	c.out.AddFlags(f, "yaml", output.DefaultFormatters)
+}
+
+// Init implements Command.
+func (c *exportBundleCommand) Init(args []string) error {
+	modelName := ""
```
howbazaar: There should be no positional args. If a user wants a different model they'd say:

```
juju export-bundle -m other
```
And the api connection machinery in the base type takes care of getting the api connection to the correct model.

cmd/juju/bundle/exportbundle.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Run implements Command.
+func (c *exportBundleCommand) Run(ctx *cmd.Context) error {
+	client, err := c.getAPI()
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	_, modelDetails, err := c.ModelCommandBase.ModelDetails()
```
howbazaar: None of this is necessary, just the `client.ExportBundle()` call

api/bundle/client.go
vinu2003: Done

cmd/juju/bundle/exportbundle.go
vinu2003: done

vinu2003:
Just highlighting the help menu for export-bundle
vinu@vinu-HP-Spectre-x360-Convertible-13-ae0xx:~/work/src/github.com/juju/juju/cmd/juju/bundle$ JUJU_DEV_FEATURE_FLAGS=developer-mode juju export-bundle --help
Usage: juju export-bundle [options]

Summary:
Exports the current model configuration in a charm bundle.

Options:
-B, --no-browser-login  (= false)
    Do not use web browser for authentication
--filename (= "")
    Export Model

Details:
Exports the current model configuration into a YAML file.

Examples:

    juju export-bundle

If --filename is not used, the archive is downloaded to a temp file
which the name of the model with randnumber appended.
The last exported can be identified manually by user.

vinu2003:
!!build!!

vinu2003:
Hi @howbazaar , The CLI work is still evolving. Initially i tried taking things from other command files and seeing how every bits works. Kindly bare with me i will remove it from WIP and keep you posted on the same :).

vinu2003:
Hi @howbazaar , Please take a look at the changes and provide your comments.I have updated the status.
Thanks.

vinu2003:
!!build!!

vinu2003:
!!build!!

vinu2003:
Hi @howbazaar , I have verified the bundle export cmdline functionality by merging tthis PR with PR#8930 and deployed a simple 'media wiki single' bundle from the .yaml file generated using GUI and from command line exported the model config and used the bundle file generated by the juju export-bundle command to deploy back successfully.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+// NewexportbundleCommand returns a fully constructed exportbundle command.
+func NewExportBundleCommand() cmd.Command {
+	return modelcmd.Wrap(&exportBundleCommand{})
+}
+
+type exportBundleCommand struct {
+	modelcmd.ModelCommandBase
+	api ExportBundleModelAPI
```
anastasiamac: I'd rather we had a function to obtain an api, similar to https://github.com/juju/juju/blob/develop/cmd/juju/storage/poolcreate.go#L55

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,129 @@
...
+
+const exportBundleHelpDoc = `
+Exports the current model configuration into a YAML file.
+
+Examples:
+
+    juju export-bundle
+
+If --filename is not used, the archive is downloaded to a temp file
+which the name of the model with time appended.
```
anastasiamac: This does not read very well.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,129 @@
...
+		Name:    "export-bundle",
+		Purpose: "Exports the current model configuration in a charm bundle.",
+		Doc:     exportBundleHelpDoc,
+	}
+}
+
+// SetFlags implements Command.
+func (c *exportBundleCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.ModelCommandBase.SetFlags(f)
+	f.StringVar(&c.Filename, "filename", "", "Export Model")
```
anastasiamac: "Export Model" is not a very good description for the filename option :D

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,129 @@
...
+	Close() error
+	ExportBundle() (string, error)
+}
+
+func (c *exportBundleCommand) getAPI() (ExportBundleModelAPI, error) {
+	if c.api != nil {
+		return c.api, nil
+	}
+
+	api, err := c.NewAPIRoot()
```
anastasiamac: I think you need to cater here for the command line model switch.
For example, 
```
 juju export-bundle -m controller
```

cmd/juju/model/exportbundle.go
anastasiamac: I am not sure that we'd want to limit users to only store the result in the file.

Generally, we provide both options - output into stdout or file. Also, if we do not lock users into a format on the server side, we can be really user-friendly here and eventually allow for other formats. Currently, yaml only is offered since this is what we accept in.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,129 @@
...
+	defer File.Close()
+
+	// Write out the result.
+	_, err = File.WriteString(result)
+	if err != nil {
+		return errors.Annotate(err, "while copying in local file")
+	}
+
+	// Print the local filename.
+	fmt.Fprintln(ctx.Stdout, filename)
```
anastasiamac: Please output something more user friendly - "Bundle successfully exported to %v", filename... or the like.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,129 @@
...
+// ResolveFilename returns the filename used by the command.
+func (c *exportBundleCommand) ResolveFilename() string {
+	filename := c.Filename
+	if filename == "" {
+		_, modelDetails, err := c.ModelCommandBase.ModelDetails()
+		if err != nil {
+			errors.Annotate(err, "retrieving current model configuration details.")
+			return ""
+		}
+		modelTag := names.NewModelTag(modelDetails.ModelUUID)
```
anastasiamac: I am not sure that model UUID is helpful. We do have user friendly model names that user would now the models by, say form 'models' command or 'switch'. I think it'll look better.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	defer client.Close()
+
+	result, err := client.ExportBundle()
+	if err != nil {
+		return err
+	}
+
+	filename := c.ResolveFilename()
+	File, err := os.Create(filename)
```
anastasiamac: No need to capitalise this variable name.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// Print the local filename.
+	fmt.Fprintln(ctx.Stdout, filename)
+	return nil
+}
+
+// ResolveFilename returns the filename used by the command.
+func (c *exportBundleCommand) ResolveFilename() string {
+	filename := c.Filename
```
anastasiamac: Why not just use c.FileName?

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2015 Canonical Ltd.
```
anastasiamac: I wish it was 2015 :D

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (f *fakeExportBundleClient) Close() error {
+	f.MethodCall(f, "Close")
+	return f.NextErr()
+}
+
+func (f *fakeExportBundleClient) ExportBundle() (string, error) {
+	f.MethodCall(f, "ExportBundle", nil)
+	return testing.ModelTag.String(), f.NextErr()
```
anastasiamac: Let's have realistic data here :)

anastasiamac:
closing in favor of https://github.com/juju/juju/pull/8962

PR #8914, opened 2018-07-11
Fixes lp#1781096: model-upgrader worker to uninstall for destroyed model.

anastasiamac:
## Description of change

On a big production systems or systems that have starving mongo, there is a chance that destroyed model may have lingering workers running and re-trying to do work for this ghost model. It is best not to bounce these workers but uninstall them.

This PR deals specifically with model-upgrader. The follow-up PRs will address other workers that might be in the same boat. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1781096


anastasiamac:
$$merge$$

PR #8920, opened 2018-07-11
LXD Default region

SimonRichardson:
## Description of change

The following changes, inserts a default region to the cloud spec, if the cloud spec name is not localhost and if the regions are empty. In this way we can then report back stating that the region is not localhost and can be identified as a different type.

If regions do end up being added to lxd, this insertion of the default region should not happen and user specified regions will be add accordingly. 

## QA steps

1. Add a cloud using lxd
2. Add credentials accordingly
3. Bootstrap a lxd remote cluster

You should see that the region is set to "default".
Consider the `juju status` output:
```sh
$ juju status
Model    Controller   Cloud/Region  Version      SLA          Timestamp
default  lxd-default  lxd/default   2.5-beta1.1  unsupported  14:41:42+01:00
```

## Documentation changes

No changes to the user CLI, but internally we assign a default region to lxd if it's not localhost and there are no regions.

## Bug reference

N/A

### DIFF
--- a/provider/lxd/credentials.go
+++ b/provider/lxd/credentials.go
@@ -12,6 +12,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/juju/juju/provider/lxd/lxdnames"
+
 	"github.com/juju/errors"
 	"github.com/juju/utils"
 	"github.com/lxc/lxd/shared"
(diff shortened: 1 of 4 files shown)


SimonRichardson:
https://trello.com/c/3CBfJkif/323-make-sure-lxd-cluster-reports-as-different-cloud-than-lxd-localhost-for-the-purposes-of-api-calls-to-the-charmstore-check-ins

SimonRichardson:
$$merge$$

SimonRichardson:
failed because another test - $$merge$$

PR #8928, opened 2018-07-12
Change Resource Updates to actually be Updates.

jameinel:
## Description of change

The old code was a delete + reinsert which was never a valid transaction
and actually breaks the TXN logic. This still isn't being properly
covered by running the updates against the database for real, but at
least it should be closer to correct.

## QA steps

I'm still working on manual reproduction steps. It likely requires deploying a charm that has a resource, and then updating that resource. It might also require a resource that is large enough to have download progress trigger.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1776673

### DIFF
--- a/state/resources_mongo.go
+++ b/state/resources_mongo.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/juju/errors"
+	"github.com/kr/pretty"
 	charmresource "gopkg.in/juju/charm.v6/resource"
 	"gopkg.in/mgo.v2/bson"
 	"gopkg.in/mgo.v2/txn"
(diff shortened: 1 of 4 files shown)


jameinel:
[1776673 Invalid Resource Operation](https://trello.com/c/Nj70FfrE/313-1776673-invalid-resource-operation)

state/resources_mongo.go
```diff
@@ -113,16 +113,39 @@ func newInsertResourceOps(stored storedResource) []txn.Op {
...
+		"storage-path":               doc.StoragePath,
+		"download-progress":          doc.DownloadProgress,
+		"timestamp-when-last-polled": doc.LastPolled,
+	}}
+}
+
 func newUpdateResourceOps(stored storedResource) []txn.Op {
 	doc := newResourceDoc(stored)
 
 	// TODO(ericsnow) Using "update" doesn't work right...
```
manadart: Missed deleting this one.

jameinel:
$$merge$$

PR #8929, opened 2018-07-13
Disabled user has no access.

anastasiamac:
## Description of change

In our permission checks we cater for logically-deleted users - they have no access to the system. This PR ensures that the same courtesy is awarded disabled users.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1781442

### DIFF
--- a/state/useraccess.go
+++ b/state/useraccess.go
@@ -186,7 +186,7 @@ func (st *State) userMayHaveAccess(tag names.UserTag) error {
 	if err != nil {
 		return errors.Trace(err)
 	}
-	// Since deleted users will throw an error, we need to check is user has been disabled here.
+	// Since deleted users will throw an error above, we need to check whether the user has been disabled here.
 	if localUser.IsDisabled() {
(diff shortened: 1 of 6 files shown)


anastasiamac:
[disabled user should not have access, lp#1781442](https://trello.com/c/gqCqNhNX/321-disabled-user-should-not-have-access-lp1781442)

state/useraccess.go
```diff
@@ -173,13 +177,26 @@ func newUserAccess(perm *userPermission, userDoc userAccessDoc, object names.Tag
...
+func (st *State) userMayHaveAccess(tag names.UserTag) error {
+	if !tag.IsLocal() {
+		// external users may have access
+		return nil
+	}
+	localUser, err := st.User(tag)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	// Since deleted users will throw an error, we need to check is user has been disabled here.
```
wallyworld: if

anastasiamac:
$$merge$$

PR #8930, opened 2018-07-13
WIP : Code changes for BundleExport

vinu2003:
This PR has server side code changes for export bundle feature.

## QA steps
execute juju export-bundle command in developer-mode
## Documentation changes
N/A
## Bug reference
N/A

yaml file generated by the command export-bundle for mediawiki single:  https://pastebin.ubuntu.com/p/Kg38k6KCc5/

### DIFF
--- a/apiserver/facades/client/bundle/bundle.go
+++ b/apiserver/facades/client/bundle/bundle.go
@@ -22,8 +22,8 @@ import (
 	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/constraints"
 	"github.com/juju/juju/permission"
-	"github.com/juju/juju/state"
 	"github.com/juju/juju/storage"
+	"github.com/juju/os/series"
 )
 
 var logger = loggo.GetLogger("juju.apiserver.bundle")
@@ -72,15 +72,10 @@ func newFacade(ctx facade.Context) (*BundleAPI, error) {
 	authorizer := ctx.Auth()
 	st := ctx.State()
 
-	model, err := st.Model()
-	if err != nil {
-		return nil, errors.Trace(err)
-	}
-
 	return NewBundleAPI(
 		NewStateShim(st),
 		authorizer,
-		model.ModelTag(),
+		names.NewModelTag(st.ModelUUID()),
 	)
 }
 
@@ -170,9 +165,7 @@ func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
 		return fail(err)
 	}
 
-	var exportConfig state.ExportConfig
-
-	b.backend.SetExportconfig(&exportConfig)
+	exportConfig := b.backend.GetExportconfig()
 	logger.Criticalf("XXXXXXXXXXXXXXXXX...........SkipStatusHistory: %v ", exportConfig.SkipStatusHistory)
 	model, err := b.backend.ExportPartial(exportConfig)
 	if err != nil {
@@ -180,7 +173,7 @@ func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
 	}
 
 	// Fill it in charm.BundleData datastructure.
-	bundleData, err := b.FillBundleData(model)
+	bundleData, err := b.fillBundleData(model)
 	if err != nil {
 		return fail(err)
 	}
@@ -200,13 +193,14 @@ func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
 func (u *APIv1) ExportBundle() (_, _ struct{}) { return }
 
 // FillBundle fills the bundledata datastructure required for the exportBundle.
-func (b *BundleAPI) FillBundleData(model description.Model) (*charm.BundleData, error) {
+func (b *BundleAPI) fillBundleData(model description.Model) (*charm.BundleData, error) {
+	// get the default-series from model config.
 	cfg := model.Config()
 	var defaultSeries string
-	for key, _ := range cfg {
-		if key == "default-series" {
-			defaultSeries = fmt.Sprintf("%v", cfg[key])
-		}
+	if value, ok := cfg["default-series"]; !ok {
+		defaultSeries = series.LatestLts()
+	} else {
+		defaultSeries = fmt.Sprint("%v", value)
 	}
 
 	data := &charm.BundleData{
@@ -252,10 +246,14 @@ func (b *BundleAPI) FillBundleData(model description.Model) (*charm.BundleData, 
 	}
 
 	for _, machine := range model.Machines() {
-		constraints := b.constraints(machine.Constraints())
+		var constraints string
+		result := b.constraints(machine.Constraints())
+		if len(result) == 0 {
+			constraints = strings.Join(result, " ")
+		}
 
 		newMachine := &charm.MachineSpec{
-			Constraints: strings.Join(constraints, " "),
+			Constraints: constraints,
 			Annotations: machine.Annotations(),
 			Series:      machine.Series(),
 		}
--- a/apiserver/facades/client/bundle/bundle.go
+++ b/apiserver/facades/client/bundle/bundle.go
@@ -10,6 +10,7 @@ import (
 	"github.com/juju/bundlechanges"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
+	"github.com/juju/os/series"
 	"gopkg.in/juju/charm.v6"
 	"gopkg.in/juju/names.v2"
 
@@ -18,7 +19,9 @@ import (
 	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/constraints"
 	"github.com/juju/juju/permission"
+	"github.com/juju/juju/state"
 	"github.com/juju/juju/storage"
+	"gopkg.in/yaml.v2"
 )
 
 var logger = loggo.GetLogger("juju.apiserver.bundle")
@@ -68,7 +71,7 @@ func newFacade(ctx facade.Context) (*BundleAPI, error) {
 	st := ctx.State()
 
 	return NewBundleAPI(
-		st,
+		stateShim{st},
 		authorizer,
 		names.NewModelTag(st.ModelUUID()),
 	)
@@ -152,6 +155,84 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
 
 // ExportBundle exports the current model configuration as bundle.
 func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
+	if err := b.checkCanRead(); err != nil {
+		return params.StringResult{}, common.ServerError(common.ErrPerm)
+	}
+
+	var exportConfig state.ExportConfig
+
+	// Skip attributes which are not required.
+	exportConfig.SkipActions = true
+	exportConfig.SkipCloudImageMetadata = true
+	exportConfig.SkipCredentials = true
+	exportConfig.SkipIPAddresses = true
+	exportConfig.SkipSSHHostKeys = true
+	exportConfig.SkipStatusHistory = true
+	exportConfig.SkipLinkLayerDevices = true
+
+	model, err := b.backend.ExportPartial(exportConfig)
+	if err != nil {
+		return params.StringResult{}, common.ServerError(err)
+	}
+
+	// Fill it in bundle data struct.
+	var bundledata charm.BundleData
+
+	// Fill in the BundleApplicationSpec of BundleData.
+	app := make(map[string]*charm.ApplicationSpec)
+	for _, application := range model.Applications() {
+		app[application.Name()].Charm = application.CharmURL()
+		app[application.Name()].Series = application.Series()
+		app[application.Name()].NumUnits = len(application.Units())
+		ut := []string{}
+		for _, unit := range application.Units() {
+			ut = append(ut, unit.Machine().Id())
+		}
+		app[application.Name()].To = ut
+
+		app[application.Name()].Expose = application.Exposed()
+		app[application.Name()].Options = application.CharmConfig()
+		app[application.Name()].Annotations = application.Annotations()
+
+		app[application.Name()].EndpointBindings = application.EndpointBindings()
+	}
+
+	// Fill in the BundleMachineSpec of BundleData.
+	mac := make(map[string]*charm.MachineSpec)
+	for _, machine := range model.Machines() {
+		constraints := []string{"arch=" + machine.Constraints().Architecture(),
+			"cpu-cores=" + string(machine.Constraints().CpuCores()),
+			"cpu-power=" + string(machine.Constraints().CpuPower()),
+			"mem=" + string(machine.Constraints().Memory()),
+			"root-disk=" + string(machine.Constraints().RootDisk())}
+		mac[machine.Id()].Constraints = strings.Join(constraints, " ")
+		mac[machine.Id()].Annotations = machine.Annotations()
+		mac[machine.Id()].Series = machine.Series()
+	}
+
+	// Fill in the relation for BundleData.
+	rel := [][]string{}
+	for _, relation := range model.Relations() {
+		endpointRelation := []string{}
+		for _, endpoint := range relation.Endpoints() {
+			endpointRelation = append(endpointRelation, endpoint.ApplicationName()+":"+endpoint.Name())
+		}
+		rel = append(rel, endpointRelation)
+	}
+
+	bundledata.Applications = app
+	bundledata.Machines = mac
+	bundledata.Relations = rel
+	bundledata.Series = series.LatestLts()
+
+	bytes, err := yaml.Marshal(bundledata)
+	if err != nil {
+		return params.StringResult{}, common.ServerError(err)
+	}
+
+	return params.StringResult{
+		Result: string(bytes),
+	}, nil
 	return params.StringResult{}, nil
 }
 
--- a/apiserver/facades/client/bundle/bundle.go
+++ b/apiserver/facades/client/bundle/bundle.go
@@ -5,6 +5,7 @@
 package bundle
 
 import (
+	"strconv"
 	"strings"
 
 	"github.com/juju/bundlechanges"
@@ -13,6 +14,7 @@ import (
 	"github.com/juju/os/series"
 	"gopkg.in/juju/charm.v6"
 	"gopkg.in/juju/names.v2"
+	"gopkg.in/yaml.v2"
 
 	"github.com/juju/description"
 	"github.com/juju/juju/apiserver/common"
@@ -22,7 +24,6 @@ import (
 	"github.com/juju/juju/permission"
 	"github.com/juju/juju/state"
 	"github.com/juju/juju/storage"
-	"gopkg.in/yaml.v2"
 )
 
 var logger = loggo.GetLogger("juju.apiserver.bundle")
@@ -155,9 +156,9 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
 }
 
 // ExportBundle exports the current model configuration as bundle.
-func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
+func (b *BundleAPI) ExportBundle() (params.BytesResult, error) {
 	if err := b.checkCanRead(); err != nil {
-		return params.StringResult{}, common.ServerError(common.ErrPerm)
+		return params.BytesResult{}, common.ServerError(common.ErrPerm)
 	}
 
 	var exportConfig state.ExportConfig
@@ -173,23 +174,27 @@ func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
 
 	model, err := b.backend.ExportPartial(exportConfig)
 	if err != nil {
-		return params.StringResult{}, common.ServerError(err)
+		return params.BytesResult{}, common.ServerError(err)
+	}
+
+	err = model.Validate()
+	if err != nil {
+		return params.BytesResult{}, common.ServerError(err)
 	}
 
 	// Fill it in charm.BundleData datastructure.
-	var bundleData charm.BundleData
-	err = b.FillBundleData(bundleData, model)
+	bundleData, err := b.FillBundleData(model)
 	if err != nil {
-		return params.StringResult{}, common.ServerError(err)
+		return params.BytesResult{}, common.ServerError(err)
 	}
 
 	bytes, err := yaml.Marshal(bundleData)
 	if err != nil {
-		return params.StringResult{}, common.ServerError(err)
+		return params.BytesResult{}, common.ServerError(err)
 	}
 
-	return params.StringResult{
-		Result: string(bytes),
+	return params.BytesResult{
+		Result: bytes,
 	}, nil
 }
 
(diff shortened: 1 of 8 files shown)


vinu2003:
Hi @howbazaar , I have verified the bundle export cmdline functionality by merging tthis PR with PR#8912 and deployed a simple 'media wiki single' bundle from the .yaml file generated using GUI and from command line exported the model config and used the bundle file generated by the juju export-bundle command to deploy back successfully.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+	return params.BytesResult{
+		Result: bytes,
+	}, nil
 }
 
 // Mask the new method from V1 API.
 // ExportBundle is not in V1 API.
 func (u *APIv1) ExportBundle() (_, _ struct{}) { return }
+
+// Fills the bundledata datastructure required for the exportBundle.
```
anastasiamac: Should match method name, i.e.

```
// FillBundleData ...
```

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+func (b *BundleAPI) FillBundleData(model description.Model) (*charm.BundleData, error) {
```
anastasiamac: Any reason why this is an exported method?

It would be helpful if we had an interface declaration for v1 and v2 that would clearly highlight the differences between versions. This would also allow us to track differences in the future when we'd be at 3, 4, 5, etc

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
 }
 
 // ExportBundle exports the current model configuration as bundle.
-func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
-	return params.StringResult{}, nil
+func (b *BundleAPI) ExportBundle() (params.BytesResult, error) {
```
anastasiamac: We cannot just remove 
```
ExportBundle() (params.StringResult, error)
```
as v1 should still have it to comply with the original promise.

Also why would we return a BytesResult? or even a StringResult? We customary return map[string]interface{} for similar situations and let the clients parse the results. For example, see modelmanager.DumpAll()

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
 }
 
 // Mask the new method from V1 API.
 // ExportBundle is not in V1 API.
 func (u *APIv1) ExportBundle() (_, _ struct{}) { return }
+
+// Fills the bundledata datastructure required for the exportBundle.
+func (b *BundleAPI) FillBundleData(model description.Model) (*charm.BundleData, error) {
+	data := &charm.BundleData{
+		Series:       series.LatestLts(),
```
anastasiamac: I am not sure that we want to hard-code it to be latest LTS. Maybe, it should be a juju supported latest lts? However, I would have expected there to be some logic calculation, maybe based on the series of the applications that are in this bundle once we've obtained them?...
Is this what javascript client currently does? We do want the parity, and if it is, this place deserves at least a comment to this effect with a TODO to address.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+			ut = append(ut, unit.Machine().Id())
+		}
+
+		newApplication := &charm.ApplicationSpec{
+			Charm:            application.CharmURL(),
+			Series:           application.Series(),
+			NumUnits:         len(application.Units()),
+			To:               ut,
+			Expose:           application.Exposed(),
+			Options:          application.CharmConfig(),
```
anastasiamac: This does not look right.
CharmConfig() are the settings that charm declares; ApplicationConfig() are the settings that application uses.
Options here need to be application config.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+			Annotations:      application.Annotations(),
+			EndpointBindings: application.EndpointBindings(),
+		}
+
+		data.Applications[application.Name()] = newApplication
+	}
+
+	for _, machine := range model.Machines() {
+		constraints := []string{"arch=" + machine.Constraints().Architecture(),
```
anastasiamac: What happens when machine.Constraints() is empty? It is defined as "omitempty" after all.
Also all individual properties are defined in the same way, so they can be empty too and need to be catered for...

apiserver/facades/client/bundle/bundle_test.go
```diff
@@ -214,3 +215,14 @@ func (s *bundleSuite) TestGetChangesBundleEndpointBindingsSuccess(c *gc.C) {
...
 }
+
+func (s *bundleSuite) TestExportBundleFailNoApplication(c *gc.C) {
+	var cfg state.ExportConfig
+	model, err := s.st.ExportpartialNoApplication(cfg)
+	c.Assert(err, jc.ErrorIsNil)
+
+	c.Assert(model.Validate(), jc.ErrorIsNil)
+
+	_, err = s.facade.FillBundleData(model)
```
anastasiamac: Shouldn't this be calling .ExportBundle() instead?

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+func (b *BundleAPI) FillBundleData(model description.Model) (*charm.BundleData, error) {
+	data := &charm.BundleData{
+		Series:       series.LatestLts(),
+		Applications: make(map[string]*charm.ApplicationSpec),
+		Relations:    [][]string{},
+		Machines:     make(map[string]*charm.MachineSpec),
+	}
+
+	if len(model.Applications()) == 0 {
+		return &charm.BundleData{}, errors.NotFoundf("application")
```
anastasiamac: Is this going to make sense to a user? They have not requested a particular application.

I think we want to say here that there was nothing to export because this model had no applications... or similar message.

apiserver/facades/client/bundle/mock_test.go
```diff
@@ -5,10 +5,48 @@ package bundle_test
 
 import (
 	"github.com/juju/testing"
+	"gopkg.in/juju/names.v2"
+
+	"github.com/juju/description"
```
anastasiamac: move me

apiserver/facades/client/bundle/mock_test.go
```diff
@@ -5,10 +5,48 @@ package bundle_test
...
+	"github.com/juju/juju/state"
 )
 
 type mockState struct {
 	testing.Stub
+	bundle.Backend
+}
+
+func (m *mockState) SetExportconfig(cfg state.ExportConfig) {
+	cfg.SkipActions = true
```
anastasiamac: This would be redundant, especially if this config never changes for bundle export.

apiserver/facades/client/bundle/mock_test.go
```diff
@@ -5,10 +5,48 @@ package bundle_test
...
+	cfg.SkipCloudImageMetadata = true
+	cfg.SkipCredentials = true
+	cfg.SkipIPAddresses = true
+	cfg.SkipSSHHostKeys = true
+	cfg.SkipStatusHistory = true
+	cfg.SkipLinkLayerDevices = true
+}
+
+func (m *mockState) ExportpartialNoApplication(config state.ExportConfig) (description.Model, error) {
```
anastasiamac: This does not look right.

For one mock state can only implement the methods from an actual state.
There are couple of other things here. Let's talk through them f-2-f.

apiserver/facades/client/bundle/state.go
```diff
@@ -3,4 +3,15 @@
...
 package bundle
 
-type Backend interface{}
+import (
+	"github.com/juju/juju/migration"
+	"github.com/juju/juju/state"
+)
+
+type Backend interface {
+	migration.StateExporter
```
anastasiamac: I am not sure that we actually want this interface here (nor this import)... Nor do I think you should change migration interface for the purpose of the bundle export!
 I think it should be just a method declaration  you are after...

```
ExportPartial(cfg state.ExportConfig) (description.Model, error)
```

migration/migration.go
```diff
@@ -29,6 +29,7 @@ var logger = loggo.GetLogger("juju.migration")
 type StateExporter interface {
 	// Export generates an abstract representation of a model.
 	Export() (description.Model, error)
+	ExportPartial(cfg state.ExportConfig) (description.Model, error)
```
anastasiamac: This does not sound valid as we do not want to affect anything to do with migrations :D

apiserver/facades/client/bundle/bundle.go
howbazaar: @anastasiamac I asked @vinu2003 to do it this way as this is the new correct way to implement versioned APIs. 

I agree that it should be a `StringResult` rather than a `BytesResult` as it will be easier for the gui to consume.

@anastasiamac for this call we are wanting to return the bundle, which is a YAML document, not a map[string]interface{}.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
 
 // ExportBundle exports the current model configuration as bundle.
-func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
-	return params.StringResult{}, nil
+func (b *BundleAPI) ExportBundle() (params.BytesResult, error) {
+	if err := b.checkCanRead(); err != nil {
+		return params.BytesResult{}, common.ServerError(common.ErrPerm)
+	}
+
+	var exportConfig state.ExportConfig
```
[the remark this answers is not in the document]
howbazaar: To simplify the extraction of the model information, we are taking advantage of the ExportPartial method on state. Since this needs the ExportConfig, we are including state here for expediency.

Perhaps in the future we may add an ExportForBundle method on state if we really want to not import state here, but until we know more we aren't doing that.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+	exportConfig.SkipSSHHostKeys = true
+	exportConfig.SkipStatusHistory = true
+	exportConfig.SkipLinkLayerDevices = true
+
+	model, err := b.backend.ExportPartial(exportConfig)
+	if err != nil {
+		return params.BytesResult{}, common.ServerError(err)
+	}
+
+	err = model.Validate()
```
howbazaar: Don't call Validate. It won't be valid. The ExportPartial is not defined to produce a valid model export.

apiserver/facades/client/bundle/bundle.go
howbazaar: I agree with @anastasiamac here, the method shouldn't be exported.

apiserver/facades/client/bundle/bundle.go
howbazaar: We definitely don't want to do this.

The Series for the bundle should be the default-series from the model config.

apiserver/facades/client/bundle/bundle.go
howbazaar: @anastasiamac unfortunately not. Bundles don't yest support application config, that is coming later (for CAAS stuff).

The options her are charm config, but we only want to output the values that have been explicitly set, which I'm pretty sure the model export handles.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+		Machines:     make(map[string]*charm.MachineSpec),
+	}
+
+	if len(model.Applications()) == 0 {
+		return &charm.BundleData{}, errors.NotFoundf("application")
+	}
+	for _, application := range model.Applications() {
+		ut := []string{}
+		for _, unit := range application.Units() {
+			ut = append(ut, unit.Machine().Id())
```
howbazaar: We don't add placement for subordinate units.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+		}
+
+		newApplication := &charm.ApplicationSpec{
+			Charm:            application.CharmURL(),
+			Series:           application.Series(),
+			NumUnits:         len(application.Units()),
```
howbazaar: Similarly, we don't have a NumUnits for subordinate applications.

apiserver/facades/client/bundle/bundle.go
howbazaar: Yes, there will need to be function to handle the non-existence of various constraint values.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -151,10 +156,111 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+			Series:      machine.Series(),
+		}
+		data.Machines[machine.Id()] = newMachine
+	}
+
+	for _, relation := range model.Relations() {
+		endpointRelation := []string{}
+		for _, endpoint := range relation.Endpoints() {
+			// skipping the 'peer' role which is not of concern in exporting the current model configuration.
+			if endpoint.Role() != "peer" {
```
howbazaar: You could simplify this loop a little by saying 

```go
if endpoint.Role() == "peer" {
    continue 
}```

File: apiserver/facades/client/bundle/state.go [hunk 19]
```diff
@@ -3,4 +3,15 @@
...
+}
+
+type stateShim struct {
```
howbazaar: We don't need this shim.

migration/migration.go
howbazaar: We don't need two interfaces. The `Backend` interface should just have the `ExportPartial` function.

apiserver/facades/client/bundle/bundle.go
anastasiamac: I see! Ok, it makes sense then to marshal it here and return as params.StringResult.

apiserver/facades/client/bundle/bundle.go
vinu2003: @howbazaar , agree that you have mentioned that in other PR that bundles are never accepted in any other format. the discussion started to have similar interface from server like its done in DumpModel command. Agree with your point and I have made it stringResult itself.

apiserver/facades/client/bundle/bundle.go
vinu2003: addressed this.

vinu2003: description model currently doesnt implement Series(). I will add that.

apiserver/facades/client/bundle/bundle.go
vinu2003: addressed

apiserver/facades/client/bundle/bundle.go
vinu2003: done

apiserver/facades/client/bundle/state.go
vinu2003: Removed.

apiserver/facades/client/bundle/bundle.go
vinu2003: All this versioning confusion started because i changed from StringResult to BytesResult. Its my fault i didn't take into consideration that gui makes use of this. Apologise.

apiserver/facades/client/bundle/bundle.go
vinu2003: missed these from javascript impl.

apiserver/facades/client/bundle/bundle.go
vinu2003: @howbazaar actually mentioned this in other PR.  that no ther format is accpeted for bundle. the confusion started when i changed it to BytesResult. Apologise. I never considered that gui takes this value.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -5,19 +5,24 @@
...
 
 	"github.com/juju/bundlechanges"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	"gopkg.in/juju/charm.v6"
 	"gopkg.in/juju/names.v2"
+	"gopkg.in/yaml.v2"
 
+	"fmt"
+	"github.com/juju/description"
```
anastasiamac: move me and the one above

apiserver/facades/client/bundle/bundle.go
```diff
@@ -150,11 +160,149 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
 	return results, nil
 }
 
+var fail = func(failErr error) (params.StringResult, error) {
```
anastasiamac: Please move this declaration to inside the method where it is relevant. Only ExportBundle uses it.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -150,11 +160,149 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
 func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
-	return params.StringResult{}, nil
+	if err := b.checkCanRead(); err != nil {
+		return fail(err)
+	}
+
+	var exportConfig state.ExportConfig
+
+	b.backend.SetExportconfig(exportConfig)
+	logger.Criticalf("XXXXXXXXXXXXXXXXX...........SkipStatusHistory: %v ", exportConfig.SkipStatusHistory)
```
anastasiamac: This does not belong in production code ;)

apiserver/facades/client/bundle/bundle.go
```diff
@@ -67,10 +72,15 @@ func newFacade(ctx facade.Context) (*BundleAPI, error) {
...
+	model, err := st.Model()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
 	return NewBundleAPI(
-		st,
+		NewStateShim(st),
 		authorizer,
-		names.NewModelTag(st.ModelUUID()),
```
anastasiamac: Why did you change this? Previous implementation was laconic and you did not need to go to .Model().

apiserver/facades/client/bundle/bundle.go
```diff
@@ -150,11 +160,149 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
 // ExportBundle exports the current model configuration as bundle.
 func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
-	return params.StringResult{}, nil
+	if err := b.checkCanRead(); err != nil {
+		return fail(err)
+	}
+
+	var exportConfig state.ExportConfig
+
+	b.backend.SetExportconfig(exportConfig)
```
anastasiamac: Why do u need to it this way? What we discussed this morning would have looked something like https://pastebin.ubuntu.com/p/skHsmffKgb/

This way, you next call becomes

```
b.backend.ExportPartial()
```

apiserver/facades/client/bundle/bundle.go
```diff
@@ -150,11 +160,149 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+		Result: string(bytes),
+	}, nil
 }
 
 // Mask the new method from V1 API.
 // ExportBundle is not in V1 API.
 func (u *APIv1) ExportBundle() (_, _ struct{}) { return }
+
+// FillBundle fills the bundledata datastructure required for the exportBundle.
+func (b *BundleAPI) FillBundleData(model description.Model) (*charm.BundleData, error) {
```
anastasiamac: Please do not export this. This should be fillBundleData...

apiserver/facades/client/bundle/bundle.go
```diff
@@ -150,11 +160,149 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+	cfg := model.Config()
+	var defaultSeries string
```
anastasiamac: This may leave defaultSeries uninitialised...

Something like this 
defaultSeries, ok := cfg["default-series"]
if !ok {
 // should never happen, but just in case it does
// let's set it to latest lts
}

and the for-loop becomes redundant... no need to iterate over the whole of config if we are only after one value.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -150,11 +160,149 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+
+	data := &charm.BundleData{
+		Series:       defaultSeries,
+		Applications: make(map[string]*charm.ApplicationSpec),
+		Relations:    [][]string{},
+		Machines:     make(map[string]*charm.MachineSpec),
+	}
+
+	if len(model.Applications()) == 0 {
+		return &charm.BundleData{}, errors.Errorf("nothing to export as there is no application found.")
```
anastasiamac: Maybe "... there are no applications..."

apiserver/facades/client/bundle/bundle.go
```diff
@@ -150,11 +160,149 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+			}
+		}
+
+		data.Applications[application.Name()] = newApplication
+	}
+
+	for _, machine := range model.Machines() {
+		result := b.constraints(machine.Constraints())
+
+		constraints := []string{"arch=" + *result.Arch,
```
anastasiamac: We do not want to create all this if thher are not specified...

I think you want your b.constraints(...) to return this collection populating only the parameters that were specified...

apiserver/facades/client/bundle/bundle_test.go
```diff
@@ -214,3 +215,12 @@ func (s *bundleSuite) TestGetChangesBundleEndpointBindingsSuccess(c *gc.C) {
 		}
 	}
 }
+
+func (s *bundleSuite) TestExportBundleFailNoApplication(c *gc.C) {
+	var cfg state.ExportConfig
+	model, err := s.st.Exportpartial(cfg)
```
anastasiamac: This is not testing the facade though...

apiserver/facades/client/bundle/bundle.go
```diff
@@ -5,19 +5,24 @@
...
 
 	"github.com/juju/bundlechanges"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
+	"github.com/juju/os/series"
 	"gopkg.in/juju/charm.v6"
 	"gopkg.in/juju/names.v2"
+	"gopkg.in/yaml.v2"
 
+	"github.com/juju/description"
```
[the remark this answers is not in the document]
anastasiamac: This import is not in the right group.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -152,9 +157,142 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
 
 // ExportBundle exports the current model configuration as bundle.
 func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
-	return params.StringResult{}, nil
+	var fail = func(failErr error) (params.StringResult, error) {
```
anastasiamac: This could be just a  local declaration
```
 fail := ...
```

rather than a 
```
var fail = 
```

apiserver/facades/client/bundle/bundle.go
```diff
@@ -152,9 +157,142 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+	return params.StringResult{
+		Result: string(bytes),
+	}, nil
 }
 
 // Mask the new method from V1 API.
 // ExportBundle is not in V1 API.
 func (u *APIv1) ExportBundle() (_, _ struct{}) { return }
+
+// FillBundle fills the bundledata datastructure required for the exportBundle.
```
anastasiamac: Comment needs to match func name and is not necessary for unexported (starting with lower-case) functions unless they are complex. If you decide to keep the comment, however, 'data structure' might be missing a space?..

apiserver/facades/client/bundle/bundle.go
```diff
@@ -152,9 +157,142 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+func (b *BundleAPI) fillBundleData(model description.Model) (*charm.BundleData, error) {
+	// get the default-series from model config.
```
anastasiamac: The comment is not needed as the code is self-explanatory.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -152,9 +157,142 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+				Annotations:      application.Annotations(),
+				EndpointBindings: application.EndpointBindings(),
+			}
+		}
+
+		data.Applications[application.Name()] = newApplication
+	}
+
+	for _, machine := range model.Machines() {
+		var constraints string
```
anastasiamac: This whole paragraph can be simplified to 
```
		newMachine := &charm.MachineSpec{
			Annotations: machine.Annotations(),
			Series:      machine.Series(),
		}
		if result := b.constraints(machine.Constraints()); len(result) != 0 {
			newMachine.Constraints = strings.Join(result, " ")
		}
```

apiserver/facades/client/bundle/bundle.go
```diff
@@ -152,9 +157,142 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+		}
+	}
+	data.Relations = append(data.Relations, endpointRelation)
+
+	return data, nil
+}
+
+func (b *BundleAPI) constraints(cons description.Constraints) []string {
+	if cons == nil {
+		return []string{}
```
anastasiamac: Why not nil?

apiserver/facades/client/bundle/bundle.go
```diff
@@ -152,9 +157,142 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+
+	data := &charm.BundleData{
+		Series:       defaultSeries,
+		Applications: make(map[string]*charm.ApplicationSpec),
+		Relations:    [][]string{},
+		Machines:     make(map[string]*charm.MachineSpec),
+	}
+
+	if len(model.Applications()) == 0 {
+		return &charm.BundleData{}, errors.Errorf("nothing to export as there are no applications.")
```
anastasiamac: Error messages should not contain a full stop as other layers may append more information to the message.

It's kind of counter-intuitive until you see the trace coming from cmd layer :)

apiserver/facades/client/bundle/bundle_test.go
```diff
@@ -214,3 +217,447 @@ func (s *bundleSuite) TestGetChangesBundleEndpointBindingsSuccess(c *gc.C) {
...
+			"uuid": "some-uuid",
+		},
+		CloudRegion: "some-region"})
+	s.st.model.SetStatus(description.StatusArgs{Value: "available"})
+
+	result, err := s.facade.ExportBundle()
+	c.Assert(err, gc.NotNil)
+	c.Assert(result, gc.Equals, params.StringResult{})
+	c.Check(err, gc.ErrorMatches, "nothing to export as there are no applications.")
+	s.st.CheckCallNames(c, "ExportPartial")
```
anastasiamac: This is an unnecessary check here and everywhere else, as the following line checks this and more ;D

vinu2003:
!!build!!

vinu2003:
$$merge$$

PR #8935, opened 2018-07-15
Create a k8s storage class if the provisioner is specified

wallyworld:
## Description of change

If the juju storage pool nominated to be used to create charm storage specifies a k8s provisioner, then we now create that storage class. Previously, the storage class needed to be set up manually by the user.

We also move all of the code to construct the storage class from k8s.go to storage.go

## QA steps

On an AWS k8s setup, create a caas model.

$ juju create-storage-pool k8s-ebs kubernetes storage-class=juju-ebs storage-provisioner=kubernetes.io/aws-ebs parameters.type=gp2
$ juju deploy ./mariadb --storage database=10M,k8s-ebs

### DIFF
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -382,13 +382,15 @@ func (k *kubernetesClient) ensureStorageClass(cfg *storageConfig) error {
 		return nil
 	}
 
+	if !k8serrors.IsNotFound(err) {
+		return errors.Trace(err)
+	}
 	// If it's not found but there's no provisioner specified, we can't
 	// create it so just return not found.
-	if !k8serrors.IsNotFound(err) || cfg.storageProvisioner == "" {
-		if k8serrors.IsNotFound(err) {
-			return errors.NotFoundf("storage class %v", cfg.storageClass)
-		}
-		return errors.Trace(err)
+	if err != nil && cfg.storageProvisioner == "" {
+		return errors.NewNotFound(nil,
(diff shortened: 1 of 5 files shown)


caas/kubernetes/provider/k8s.go
```diff
@@ -380,6 +374,34 @@ func (k *kubernetesClient) maybeGetVolumeClaimSpec(params volumeParams) (*core.P
...
+	}
+
+	// If it's not found but there's no provisioner specified, we can't
+	// create it so just return not found.
+	if !k8serrors.IsNotFound(err) || cfg.storageProvisioner == "" {
+		if k8serrors.IsNotFound(err) {
+			return errors.NotFoundf("storage class %v", cfg.storageClass)
+		}
+		return errors.Trace(err)
+	}
```
ycliuhw: thinking if we could make here flat like: 
```go
if !k8serrors.IsNotFound(err) {
    return errors.Trace(err)
}
// no existing storageclass, so will try to create one
if cfg.storageProvisioner == "" {
    return errors.NotFoundf("storage class %v, but no storage provisioner is available to ensure it.", cfg.storageClass)
}

...
```

wallyworld: I've tweaked the implementation. Note that the suggested code isn't quite correct. It needs to be:
if err != nil && cfg.storageProvisioner == "" {

ycliuhw:
!!build!!

wallyworld:
$$merge$$

PR #8939, opened 2018-07-16
upgradeseries is only applicable for IAAS models

wallyworld:
## Description of change

When setting up the uniter remote watcher, only hook up the upgradeseries watcher on IAAS models.

As a driveby, move a couple of storage watcher vars left in the wrong place.

## QA steps

Ensure CAAS deployments work again.

### DIFF
--- a/worker/uniter/remotestate/watcher.go
+++ b/worker/uniter/remotestate/watcher.go
@@ -251,10 +251,10 @@ func (w *RemoteStateWatcher) loop(unitTag names.UnitTag) (err error) {
 	requiredEvents++
 
 	var (
-		seenStorageChange bool
-		storageChanges    watcher.StringsChannel
+		seenApplicationChange bool
 
-		seenApplicationChange *bool
+		upgradeSeriesChanges    watcher.NotifyChannel
+		seenUpgradeSeriesChange bool
 	)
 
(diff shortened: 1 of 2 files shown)


worker/uniter/remotestate/watcher.go
```diff
@@ -251,10 +251,10 @@ func (w *RemoteStateWatcher) loop(unitTag names.UnitTag) (err error) {
 	requiredEvents++
 
 	var (
-		seenStorageChange bool
-		storageChanges    watcher.StringsChannel
-
 		seenApplicationChange *bool
+
+		upgradeSeriesChanges    watcher.NotifyChannel
+		seenUpgradeSeriesChange *bool
```
jameinel: pointer to bool seems so fragile. Do we need the 'nil' state to know that it was never initialized? It seems safer to just always have the bool even if it doesn't mean anything sometimes.
I realize seenApplicationChange follows that pattern, but it just feels like an easy way to mess up and get a panic.

wallyworld: I've reworked it

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #8940, opened 2018-07-16
Disable 2 way k8s-juju sync

wallyworld:
## Description of change

Disable Juju attempting to add new units when a scale operation is done directly on the k8s cluster (outside of juju). It causes a unit explosion in some cases.

## QA steps

Deploy and scale a k8s model using juju.


wallyworld:
!!build!!

wallyworld:
$$merge$$

PR #8941, opened 2018-07-17
ensure bundle facade versioning

ycliuhw:
## Description of change

added facade versioning for BundleGet because we added `Device` support to this facade.

## QA steps

None

## Bug reference

None

### DIFF
--- a/apiserver/allfacades.go
+++ b/apiserver/allfacades.go
@@ -149,9 +149,7 @@ func AllFacades() *facade.Registry {
 	reg("Backups", 2, backups.NewFacadeV2)
 	reg("Block", 2, block.NewAPI)
 	reg("Bundle", 1, bundle.NewFacadeV1)
-	if featureflag.Enabled(feature.DeveloperMode) {
-		reg("Bundle", 2, bundle.NewFacadeV2)
-	}
+	reg("Bundle", 2, bundle.NewFacadeV2)
 	reg("CharmRevisionUpdater", 2, charmrevisionupdater.NewCharmRevisionUpdaterAPI)
 	reg("Charms", 2, charms.NewFacade)
 	reg("Cleaner", 2, cleaner.NewCleanerAPI)
--- a/apiserver/allfacades.go
+++ b/apiserver/allfacades.go
@@ -149,6 +149,9 @@ func AllFacades() *facade.Registry {
 	reg("Backups", 2, backups.NewFacadeV2)
 	reg("Block", 2, block.NewAPI)
 	reg("Bundle", 1, bundle.NewFacadeV1)
+	if featureflag.Enabled(feature.DeveloperMode) {
+		reg("Bundle", 2, bundle.NewFacadeV2)
+	}
 	reg("CharmRevisionUpdater", 2, charmrevisionupdater.NewCharmRevisionUpdaterAPI)
 	reg("Charms", 2, charms.NewFacade)
 	reg("Cleaner", 2, cleaner.NewCleanerAPI)
--- a/apiserver/facades/client/bundle/bundle.go
+++ b/apiserver/facades/client/bundle/bundle.go
@@ -106,12 +106,35 @@ func (b *BundleAPI) checkCanRead() error {
 // GetChanges returns the list of changes required to deploy the given bundle
 // data. The changes are sorted by requirements, so that they can be applied in
 // order.
+// V1 GetChanges did not support device.
+func (b *APIv1) GetChanges(args params.BundleChangesParams) (params.BundleChangesResults, error) {
+	var results params.BundleChangesResults
+	data, err := charm.ReadBundleData(strings.NewReader(args.BundleDataYAML))
+	if err != nil {
+		return results, errors.Annotate(err, "cannot read bundle YAML")
+	}
+	for key, application := range data.Applications {
+		application.Devices = nil
+		data.Applications[key] = application
+	}
+	return getChanges(data)
+}
+
+// GetChanges returns the list of changes required to deploy the given bundle
+// data. The changes are sorted by requirements, so that they can be applied in
+// order.
 func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleChangesResults, error) {
 	var results params.BundleChangesResults
 	data, err := charm.ReadBundleData(strings.NewReader(args.BundleDataYAML))
 	if err != nil {
 		return results, errors.Annotate(err, "cannot read bundle YAML")
 	}
+	return getChanges(data)
+
+}
+
+func getChanges(data *charm.BundleData) (params.BundleChangesResults, error) {
+	var results params.BundleChangesResults
 	verifyConstraints := func(s string) error {
 		_, err := constraints.Parse(s)
 		return err
--- a/apiserver/facades/client/bundle/bundle.go
+++ b/apiserver/facades/client/bundle/bundle.go
@@ -113,11 +113,50 @@ func (b *APIv1) GetChanges(args params.BundleChangesParams) (params.BundleChange
 	if err != nil {
 		return results, errors.Annotate(err, "cannot read bundle YAML")
 	}
-	for key, application := range data.Applications {
-		application.Devices = nil
-		data.Applications[key] = application
+	verifyConstraints := func(s string) error {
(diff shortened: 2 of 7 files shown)


ycliuhw:
[Add facade upgrade for bundleGet](https://trello.com/c/zWjEIaFR/325-add-facade-upgrade-for-bundleget)

apiserver/allfacades.go
```diff
@@ -149,6 +149,9 @@ func AllFacades() *facade.Registry {
 	reg("Backups", 2, backups.NewFacadeV2)
 	reg("Block", 2, block.NewAPI)
 	reg("Bundle", 1, bundle.NewFacadeV1)
+	if featureflag.Enabled(feature.DeveloperMode) {
```
anastasiamac: You do not need feature flag here. 
Also, you are upping the facade version and in this collection only one instance of it can exist.

apiserver/facades/client/client/bundles.go
```diff
@@ -15,12 +15,14 @@ import (
...
+// Note: any new feature in the future like devices will never be supported here.
 func (c *Client) GetBundleChanges(args params.BundleChangesParams) (params.BundleChangesResults, error) {
 	st := c.api.state()
 
 	bundleAPI, err := bundle.NewBundleAPI(st, c.api.auth, names.NewModelTag(st.ModelUUID()))
 	if err != nil {
 		return params.BundleChangesResults{}, err
 	}
-	return bundleAPI.GetChanges(args)
+	apiV1 := bundle.APIv1{&bundle.APIv2{bundleAPI}}
```
anastasiamac: That's not right. This call should not have changed. Instead your bundle.NewBundleAPI should return v2.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -103,6 +103,62 @@ func (b *BundleAPI) checkCanRead() error {
 	return nil
 }
 
+// GetChanges returns the list of changes required to deploy the given bundle
+// data. The changes are sorted by requirements, so that they can be applied in
+// order.
+// V1 GetChanges did not support device.
+func (b *APIv1) GetChanges(args params.BundleChangesParams) (params.BundleChangesResults, error) {
```
anastasiamac: so is this pointer to a APIv1 or APIv2?
It feels very similar to what we have done in modelmanager...
For example see https://github.com/juju/juju/blob/develop/apiserver/facades/client/modelmanager/modelmanager.go#L835 vs https://github.com/juju/juju/blob/develop/apiserver/facades/client/modelmanager/modelmanager.go#L854.
This is an example of a method change in between different versions. 
You will need to keep an APIv1 call to return data without device support and have another call sitting on APIv2 providing the support. Internally they may call the same func but present data differently.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -160,6 +216,6 @@ func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
 	return params.StringResult{}, nil
 }
 
-// Mask the new method from V1 API.
 // ExportBundle is not in V1 API.
-func (u *APIv1) ExportBundle() (_, _ struct{}) { return }
+// Mask the new method from V1 API.
```
anastasiamac: I am not sure how this got in... We do not generally implement versions this way but rather using interfaces as in https://github.com/juju/juju/blob/develop/apiserver/facades/client/modelmanager/modelmanager.go#L44

apiserver/facades/client/bundle/bundle.go
```diff
@@ -103,6 +103,62 @@ func (b *BundleAPI) checkCanRead() error {
...
+	verifyConstraints := func(s string) error {
+		_, err := constraints.Parse(s)
+		return err
+	}
+	verifyStorage := func(s string) error {
+		_, err := storage.ParseConstraints(s)
+		return err
+	}
+	if err := data.Verify(verifyConstraints, verifyStorage, nil); err != nil {
+		if err, ok := err.(*charm.VerificationError); ok {
```
anastasiamac: I hope that different scopes will take effect here. However, from readability perspective, it seems to be difficult to differentiate between which 'err'...

Also could we please add tests that exercise this code path?

apiserver/facades/client/bundle/bundle.go
ycliuhw: This is existing code from develop. I just changed the doc to make linter happy

apiserver/facades/client/client/bundles.go
ycliuhw: The `NewBundleAPI` returns `BundleAPI`, but facade constructors return `APIv*`

apiserver/facades/client/bundle/bundle.go
ycliuhw: yes, agreed. I did think this. But here the difference between two versions are neither params nor returns. It's in the middle `GUIArgs()`. The idea solution would do whole versioning in bundlechanges repo. I was thinking to do it like that. But I am not very sure if it's worth to put this effect.

apiserver/facades/client/client/bundles.go
anastasiamac: Right, I see. So this functionality exists on both Client and Bundle facade. And the one on the Client facade should behave as previously whereas the on Bundle facade has been revved up...

So this call should probably be bundle.NewFacadeV1().

apiserver/facades/client/bundle/bundle.go
anastasiamac: This looks wrong anyway. The version bump need to be changed to avoid doing this kind of placeholder, I think.

But even if you choose not to undertake this here, the comment should be starting with the method name... Maybe remove this change?

apiserver/facades/client/bundle/bundle.go
anastasiamac: Really? It looks like a brand new method to me?...

apiserver/facades/client/client/bundles.go
anastasiamac: By the way, if here is the only palce that you need the data returned to be different (say with no devices), then why can't we just change the output here once we get the return from bundleAPI.GetChanges()? Would it not also avoid you making a PR against bundelchanges repo?

ycliuhw:
hi @anastasiamac as you suggested, I renamed the GUIArgs to GUIArgsWithDevice in bundlechanges repo and refactored this PR for it as well.
Please take a look when you have time.
Thanks

apiserver/facades/client/bundle/bundle.go
```diff
@@ -143,16 +181,47 @@ func (b *BundleAPI) GetChanges(args params.BundleChangesParams) (params.BundleCh
...
+			_, err := devices.ParseConstraints(s)
+			return err
+		},
 	}
-	return results, nil
+	return getChanges(args, vs, func(changes []bundlechanges.Change, results *params.BundleChangesResults) error {
+		results.Changes = make([]*params.BundleChange, len(changes))
+		for i, c := range changes {
+			var guiArgs []interface{}
+			switch c := c.(type) {
```
anastasiamac: I wonder if it is needed now... I mean we are bringing in the version of bundlechanges that will have the new GUIArgsWith... so we can always call it...

Could you please explain the rationale for the 2 code paths here?

apiserver/facades/client/bundle/bundle.go
```diff
@@ -162,4 +231,4 @@ func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
 
 // Mask the new method from V1 API.
 // ExportBundle is not in V1 API.
-func (u *APIv1) ExportBundle() (_, _ struct{}) { return }
+func (b *APIv1) ExportBundle() (_, _ struct{}) { return }
```
anastasiamac: Is this really needed? :)

apiserver/facades/client/bundle/bundle.go
ycliuhw: Because here in the `[] change`, the change is not always needs to run `GUIArgsWithDevices` but only for `ApplicationChange`.

ycliuhw:
$$merge$$

ycliuhw:
!!build!!

ycliuhw:
$$merge$$

PR #8954, opened 2018-07-20
leases: make manager run claims and expiry asynchronously

babbageclunk:
## Description of change
The raft-based lease store will potentially return timeouts for write operations that happen while the leader is changing. This is because the requests are forwarded over the pubsub hub to the leader - since the hub isn't a guaranteed-delivery system if there's no raft forwarder running the request will be dropped on the floor, and no response will ever be sent. The store will eventually return a timeout in this case.

That means that the lease manager needs to be able to run those requests asynchronously so they don't block handling subsequent requests (after the raft leadership has transferred successfully). It also retries operations that timeout, up to 5 times with exponential backoff. 

## QA steps
This doesn't/shouldn't change behaviour when the lease store operations aren't timing out. Smoke test by bootstrapping a controller, deploying some applications with mutiple units, enabling HA.

### DIFF
--- a/core/lease/store.go
+++ b/core/lease/store.go
@@ -120,9 +120,17 @@ func ValidateString(s string) error {
 	return nil
 }
 
-// ErrInvalid indicates that a Store operation failed because latest state
-// indicates that it's a logical impossibility. It's a short-range signal to
-// calling code only; that code should never pass it on, but should inspect
-// the Store's updated Leases() and either attempt a new operation or return
-// a new error at a suitable level of abstraction.
-var ErrInvalid = errors.New("invalid lease operation")
+var (
+	// ErrInvalid indicates that a Store operation failed because latest state
+	// indicates that it's a logical impossibility. It's a short-range signal to
+	// calling code only; that code should never pass it on, but should inspect
+	// the Store's updated Leases() and either attempt a new operation or return
+	// a new error at a suitable level of abstraction.
+	ErrInvalid = errors.New("invalid lease operation")
+
+	// ErrTimeout indicates that a Store operation failed because it
+	// couldn't update the underlying lease information. This is probably
+	// a transient error due to changes in the cluster, and indicates that
+	// the operation should be retried.
+	ErrTimeout = errors.New("lease operation timed out")
+)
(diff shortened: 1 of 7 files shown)


babbageclunk:
!!build!!

Looks like network issues installing dependencies?

babbageclunk:
!!build!!

Trying again - I can clone juju from the build machine.

babbageclunk:
!!build!!

babbageclunk:
!!build!!

Veebers:
!!build!!

babbageclunk:
If the https problem is on the github side maybe they've fixed it over the weekend?

!!build!!

babbageclunk:
Rerunning the failures it seems like they're intermittent? I can't even see what the failure is for one of them.

!!build!!

worker/lease/manager_async_test.go
```diff
@@ -0,0 +1,506 @@
+// Copyright 2015 Canonical Ltd.
```
anastasiamac: 3 years later...

worker/lease/manager_async_test.go
```diff
@@ -0,0 +1,506 @@
...
+			result <- claimer.Claim("icecream", "rosie", time.Minute)
+		}()
+
+		select {
+		case <-claimCalls:
+		case <-time.After(coretesting.LongWait):
+			c.Fatalf("timed out waiting for claim")
+		}
+
+		// Why three waiters, you ask? I also was confused about that
```
anastasiamac: Awesome! Glad you have figured it out and added a brilliant comment!

core/lease/store.go
```diff
@@ -120,9 +120,17 @@ func ValidateString(s string) error {
...
-// the Store's updated Leases() and either attempt a new operation or return
-// a new error at a suitable level of abstraction.
-var ErrInvalid = errors.New("invalid lease operation")
+var (
+	// ErrInvalid indicates that a Store operation failed because latest state
+	// indicates that it's a logical impossibility. It's a short-range signal to
+	// calling code only; that code should never pass it on, but should inspect
+	// the Store's updated Leases() and either attempt a new operation or return
+	// a new error at a suitable level of abstraction.
+	ErrInvalid = errors.New("invalid lease operation")
```
howbazaar: For future work perhaps, we have Invalid and now Timeout errors in the errors package.

babbageclunk:
$$merge$$

manadart:
@babbageclunk Just a heads up that the unit tests race job has been failing since this one landed.

babbageclunk:
Oh bums - thanks Joe.

PR #8955, opened 2018-07-20
LXD Trust passwords

SimonRichardson:
## Description of change

The following changes allows you to only sign up in interactive mode
with a trust password. Everything else is then created and exchanged
behind the scenes.

Included are some interactive tests that highlight the branching
changes from the schema.

Additional changes to the interactive mode to help with the user UX.

## QA steps

It assumes that you have access to a remote lxd cluster, save the following
file so we can reference it later. 

The LXD server in question should have a trust password configured.

```yaml
clouds:
  lxd:
    type: lxd
    auth-types: [interactive]
    endpoint: https://10.0.0.250:8443
```

```sh
juju add-cloud /path/to/cloud.yaml
juju add-credential lxd
```
When asked, add the trust password, but then skip on all the other optional
values.

## Documentation changes

The following changes allow a one password sign-on to a lxd cluster from juju.
See the demo below for more information.

## Bug reference

N/A

## Demo

[demo](https://asciinema.org/a/192724?autoplay=1&speed=3)

### DIFF
--- a/provider/lxd/credentials.go
+++ b/provider/lxd/credentials.go
@@ -104,23 +104,25 @@ func (environProviderCredentials) CredentialSchemas() map[cloud.AuthType]cloud.C
 		},
 		interactiveAuthType: {
 			{
+				Name: credAttrTrustPassword,
+				CredentialAttr: cloud.CredentialAttr{
+					Description: "The LXD server trust password.",
+					Hidden:      true,
+					Optional:    true,
+				},
+			}, {
 				Name: credAttrClientCert,
 				CredentialAttr: cloud.CredentialAttr{
 					Description:    "The LXD client certificate, PEM-encoded.",
 					ExpandFilePath: true,
+					Optional:       true,
 				},
 			}, {
 				Name: credAttrClientKey,
 				CredentialAttr: cloud.CredentialAttr{
 					Description:    "The LXD client key, PEM-encoded.",
 					ExpandFilePath: true,
-				},
-			}, {
-				Name: credAttrTrustPassword,
-				CredentialAttr: cloud.CredentialAttr{
-					Description: "The LXD server trust password.",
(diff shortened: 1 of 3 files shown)


SimonRichardson:
https://trello.com/c/x1BzAZWC/382-client-cert-and-client-key-for-credentials-should-be-optional-and-generate-if-not-present

provider/lxd/credentials.go
```diff
@@ -218,7 +220,34 @@ func (p environProviderCredentials) FinalizeCredential(
...
+		// passwords can be valid ones...
+		if _, ok := credAttrs[credAttrTrustPassword]; ok {
+			// check to see if the client cert, keys exist, if they do not,
+			// generate them for the user.
+			if _, ok := getClientCertificates(args.Credential); !ok {
+				stderr := ctx.GetStderr()
+				nopLogf := func(s string, args ...interface{}) {
+					fmt.Fprintf(stderr, s+"\n", args...)
+				}
+				clientCert, clientKey, err := p.readOrGenerateCert(nopLogf)
```
jameinel: nopLogf?

SimonRichardson: This was in the old code, I'm just reusing it in this instance, guess the original person didn't want logging in every scenario.

SimonRichardson:
@jameinel not far off at all, in fact I could easily take a look next week.

provider/lxd/credentials.go
SimonRichardson: Personally I don't see why we do that, as it's valuable information. Any opposition in removing it?

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #8959, opened 2018-07-23
ec2 provider reacts to invalid credential.

anastasiamac:
## Description of change

This is the first of the providers to learn to react to invalid credential.

Essentially a successfully bootstrapped system might get the cloud credential it uses invalidated by the provider, say cloud credential gets expired... This event should pause all the interactions with the cloud api until the credential has been updated.

This PR ensures that this happens for aws.

## QA steps

* bootstrap
* add models, deploy applications
* invalidate cloud credential
* check that cannot deploy anything nor create anything that needs to also exist on the cloud
* check logs do not contain infinite number of "invalid credential" errors
* check that all workers that require valid credentials are stopped
* update credential or re-validate it
* check that everything is back to fully functional

### DIFF
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -110,27 +110,30 @@ func (e *environ) Name() string {
 
 // PrepareForBootstrap is part of the Environ interface.
 func (env *environ) PrepareForBootstrap(ctx environs.BootstrapContext) error {
+	callCtx := context.NewCloudCallContext()
+	// Cannot really invalidate a credential here since nothing is bootstrapped yet.
+	callCtx.InvalidateCredentialFunc = func(string) error { return nil }
 	if ctx.ShouldVerifyCredentials() {
-		if err := verifyCredentials(env); err != nil {
+		if err := verifyCredentials(env, callCtx); err != nil {
 			return err
 		}
 	}
 	ecfg := env.ecfg()
 	vpcID, forceVPCID := ecfg.vpcID(), ecfg.forceVPCID()
-	if err := validateBootstrapVPC(env.ec2, env.cloud.Region, vpcID, forceVPCID, ctx); err != nil {
-		return errors.Trace(maybeConvertCredentialError(err))
+	if err := validateBootstrapVPC(env.ec2, callCtx, env.cloud.Region, vpcID, forceVPCID, ctx); err != nil {
+		return errors.Trace(maybeConvertCredentialError(err, callCtx))
 	}
 	return nil
 }
 
 // Create is part of the Environ interface.
 func (env *environ) Create(ctx context.ProviderCallContext, args environs.CreateParams) error {
-	if err := verifyCredentials(env); err != nil {
+	if err := verifyCredentials(env, ctx); err != nil {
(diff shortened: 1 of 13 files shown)


anastasiamac:
[adjust aws  to use callback to invalidate cloud credential when needed](https://trello.com/c/m37mbbl2/324-adjust-aws-to-use-callback-to-invalidate-cloud-credential-when-needed)

provider/ec2/environ.go
```diff
@@ -110,27 +110,30 @@ func (e *environ) Name() string {
 
 // PrepareForBootstrap is part of the Environ interface.
 func (env *environ) PrepareForBootstrap(ctx environs.BootstrapContext) error {
+	callCtx := context.NewCloudCallContext()
+	// Cannot really invalidate a credential here since nothing is bootstrapped yet.
+	callCtx.InvalidateCredentialFunc = func(string) error { return nil }
```
babbageclunk: Maybe doing nothing should be the default behaviour of the context? Would you get a nil-pointer error if this hadn't been set?

provider/ec2/provider.go
```diff
@@ -189,38 +189,48 @@ your details are being verified.`
...
 // user-friendly explanation.
 // All other errors are returned un-wrapped and not annotated.
-var maybeConvertCredentialError = func(err error) error {
+var maybeConvertCredentialError = func(err error, ctx context.ProviderCallContext) error {
 	if err == nil {
 		return nil
 	}
 
+	convert := func(converted error) error {
+		logger.Errorf("XXXXXXXXXXXXXXXXXXX calling invalidate credential")
```
babbageclunk: Debugging output?

anastasiamac: Oh jeez... what gave it away? :D 
Yes, thank you - it is not needed in prod code.

provider/ec2/environ.go
anastasiamac: No, if InvalidateCredentialFunc is not overwritten as above, it will not get a nil pointer but will tell you that this method has not been implemented instead.

I want to keep it this as our default behavior to catch any strange and obscure code paths that may have been missed in my excessive testing. After all, there is always a tiny, remote chance that there is a path that has not been walked by me. I also believe that getting an explicit "not implemented" yet in the log will help diagnose these obscure cases.

So, I would prefer to keep the default as "not implemented" and, in cases such as this one, to have a silent implementation (i.e. returning nil). In this particular case, there is nothing that we can do with an invalid credential -  we have not yet even stored it on the controller and, hence, cannot mark our copy as invalid.

anastasiamac:
infamous - github.com/juju/juju/cmd/juju/machine.TestPackage

anastasiamac:
$$merge$$

anastasiamac:
$$merge$$

provider/ec2/environ.go
babbageclunk: Yeah, that makes sense - I'm happy with needing an explicit do nothing in the rare cases you need that. Having a default that highlights the problem without panicing is good too.

PR #8962, opened 2018-07-23
Client Facade and CLI changes for export bundle

vinu2003:
This PR addresses CLI facade code changes required for export bundle command which exports the configuration of current model in a bundle.
## QA steps
juju  export-bundle --filename <filename>
(or)
juju export-bundle - the output is displayed in stdout.

## Documentation changes
Yes.
## Bug reference

### DIFF
--- a/api/bundle/client.go
+++ b/api/bundle/client.go
@@ -42,5 +42,9 @@ func (c *Client) ExportBundle() (string, error) {
 		return "", errors.Trace(err)
 	}
 
+	if len(result.Result) == 0 {
+		return "", errors.Annotate(result.Error, "export failed")
+	}
+
 	return result.Result, nil
 }
--- a/api/bundle/client.go
+++ b/api/bundle/client.go
@@ -1,0 +1,50 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+// Package bundle provides access to the bundle api facade.
+// This facade contains api calls that are specific to bundles.
+
+package bundle
+
+import (
+	"github.com/juju/errors"
+	"github.com/juju/loggo"
+
+	"github.com/juju/juju/api/base"
+	"github.com/juju/juju/apiserver/params"
+)
+
+var logger = loggo.GetLogger("juju.api.bundle")
+
+// Client allows access to the bundle API end point.
+type Client struct {
+	base.ClientFacade
+	st     base.APICallCloser
+	facade base.FacadeCaller
+}
+
+// NewClient creates a new client for accessing the bundle api.
+func NewClient(st base.APICallCloser) *Client {
+	frontend, backend := base.NewClientFacade(st, "Bundle")
+	return &Client{
+		ClientFacade: frontend,
+		st:           st,
+		facade:       backend}
+}
+
+// ExportBundle exports the current model configuration.
+func (c *Client) ExportBundle() (string, error) {
+	var result params.StringResult
+	if bestVer := c.BestAPIVersion(); bestVer < 2 {
+		return "", errors.Errorf("command not supported on v%d", bestVer)
+	}
+
+	if err := c.facade.FacadeCall("ExportBundle", nil, &result); err != nil {
+		return "", errors.Trace(err)
+	}
+
+	if len(result.Result) == 0 {
+		return "", errors.Errorf("result obtained is incorrect")
+	}
+	return result.Result, nil
+}
--- a/api/bundle/client.go
+++ b/api/bundle/client.go
@@ -34,11 +34,10 @@ func NewClient(st base.APICallCloser) *Client {
 func (c *Client) ExportBundle() (string, error) {
 	var result params.StringResult
 	if bestVer := c.BestAPIVersion(); bestVer < 2 {
-		return "", errors.Errorf("command not supported on v%d", bestVer)
+		return "", errors.Errorf("ExportBundle() on v%d is not supported", bestVer)
 	}
 
 	if err := c.facade.FacadeCall("ExportBundle", nil, &result); err != nil {
-		errors.Annotate(result.Error, "export failed")
 		return "", errors.Trace(err)
 	}
 
--- a/api/bundle/client.go
+++ b/api/bundle/client.go
@@ -19,7 +19,6 @@ var logger = loggo.GetLogger("juju.api.bundle")
 // Client allows access to the bundle API end point.
 type Client struct {
 	base.ClientFacade
-	st     base.APICallCloser
 	facade base.FacadeCaller
 }
 
@@ -28,7 +27,6 @@ func NewClient(st base.APICallCloser) *Client {
 	frontend, backend := base.NewClientFacade(st, "Bundle")
 	return &Client{
 		ClientFacade: frontend,
-		st:           st,
 		facade:       backend}
 }
 
@@ -40,11 +38,8 @@ func (c *Client) ExportBundle() (string, error) {
 	}
 
 	if err := c.facade.FacadeCall("ExportBundle", nil, &result); err != nil {
-		return "", errors.Trace(err)
+		return "", errors.Annotate(result.Error, "export failed")
 	}
 
-	if len(result.Result) == 0 {
-		return "", errors.Errorf("result obtained is incorrect")
-	}
 	return result.Result, nil
 }
--- a/api/bundle/client.go
+++ b/api/bundle/client.go
@@ -42,8 +42,8 @@ func (c *Client) ExportBundle() (string, error) {
 		return "", errors.Trace(err)
 	}
 
-	if len(result.Result) == 0 {
-		return "", errors.Annotate(result.Error, "export failed")
+	if result.Error != nil {
+		return "", errors.Trace(result.Error)
 	}
 
 	return result.Result, nil
--- a/api/bundle/client.go
+++ b/api/bundle/client.go
@@ -38,7 +38,8 @@ func (c *Client) ExportBundle() (string, error) {
 	}
 
 	if err := c.facade.FacadeCall("ExportBundle", nil, &result); err != nil {
-		return "", errors.Annotate(result.Error, "export failed")
+		errors.Annotate(result.Error, "export failed")
+		return "", errors.Trace(err)
 	}
 
 	return result.Result, nil
--- a/api/bundle/client_test.go
+++ b/api/bundle/client_test.go
@@ -59,6 +59,7 @@ func (s *bundleMockSuite) TestExportBundlev2(c *gc.C) {
 			c.Check(id, gc.Equals, "")
 			c.Check(request, gc.Equals, "ExportBundle")
 			c.Assert(args, gc.Equals, nil)
+			c.Assert(response, gc.FitsTypeOf, &params.StringResult{})
 			result := response.(*params.StringResult)
 			result.Result = "applications:\n  " +
 				"ubuntu:\n    " +
@@ -89,4 +90,28 @@ func (s *bundleMockSuite) TestExportBundlev2(c *gc.C) {
 		"series: xenial\n"+
 		"relations:\n"+
 		"- []\n")
+}
+
+func (s *bundleMockSuite) TestExportBundleErrorv2(c *gc.C) {
+	client := newClient(
+		func(objType string, version int,
+			id,
+			request string,
+			args,
+			response interface{},
+		) error {
+			c.Check(objType, gc.Equals, "Bundle")
+			c.Check(id, gc.Equals, "")
+			c.Check(request, gc.Equals, "ExportBundle")
+			c.Assert(args, gc.Equals, nil)
+			c.Assert(response, gc.FitsTypeOf, &params.StringResult{})
+			result := response.(*params.StringResult)
+			result.Result = ""
+			return result.Error
+		}, 2,
+	)
+	result, err := client.ExportBundle()
+	c.Assert(err, gc.NotNil)
+	c.Assert(result, jc.DeepEquals, "")
+	c.Check(err.Error(), jc.Contains, "export failed")
 }
--- a/api/bundle/client_test.go
+++ b/api/bundle/client_test.go
@@ -89,7 +89,7 @@ func (s *bundleMockSuite) TestExportBundlev2(c *gc.C) {
 		"- []\n")
 }
 
-func (s *bundleMockSuite) TestExportBundleErrorv2(c *gc.C) {
+func (s *bundleMockSuite) TestExportBundleNotNilParamsErrorv2(c *gc.C) {
 	client := newClient(
 		func(objType string, version int,
 			id,
@@ -99,12 +99,32 @@ func (s *bundleMockSuite) TestExportBundleErrorv2(c *gc.C) {
 		) error {
 			result := response.(*params.StringResult)
 			result.Result = ""
-			*(response.(*params.StringResult)) = params.StringResult{Error: common.ServerError(errors.New("export failed"))}
+			*(response.(*params.StringResult)) = params.StringResult{Error: common.ServerError(
+				errors.New("export failed nothing to export as there are no applications"))}
 			return result.Error
 		}, 2,
 	)
 	result, err := client.ExportBundle()
 	c.Assert(err, gc.NotNil)
 	c.Assert(result, jc.DeepEquals, "")
-	c.Check(err.Error(), gc.Matches, "export failed")
+	c.Check(err.Error(), gc.Matches, "export failed nothing to export as there are no applications")
+}
+
+func (s *bundleMockSuite) TestExportBundleFailNoParamsErrorv2(c *gc.C) {
+	client := newClient(
+		func(objType string, version int,
+			id,
+			request string,
+			args,
+			response interface{},
+		) error {
+			result := response.(*params.StringResult)
+			result.Result = ""
+			return errors.New("foo")
+		}, 2,
+	)
+	result, err := client.ExportBundle()
+	c.Assert(err, gc.NotNil)
+	c.Assert(result, jc.DeepEquals, "")
+	c.Check(err.Error(), gc.Matches, "foo")
 }
--- a/api/bundle/client_test.go
+++ b/api/bundle/client_test.go
@@ -45,7 +45,8 @@ func (s *bundleMockSuite) TestFailExportBundlev1(c *gc.C) {
 		}, 1,
 	)
 	result, err := client.ExportBundle()
-	c.Assert(err, gc.ErrorMatches, "command not supported on v1")
+	c.Assert(err, gc.NotNil)
+	c.Assert(err.Error(), gc.Equals, "ExportBundle() on v1 is not supported")
 	c.Assert(result, jc.DeepEquals, "")
 }
 
--- a/api/bundle/client_test.go
+++ b/api/bundle/client_test.go
@@ -1,0 +1,92 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package bundle_test
+
+import (
+	jc "github.com/juju/testing/checkers"
+	gc "gopkg.in/check.v1"
+
+	basetesting "github.com/juju/juju/api/base/testing"
+	"github.com/juju/juju/api/bundle"
+	"github.com/juju/juju/apiserver/params"
+	coretesting "github.com/juju/juju/testing"
+)
+
+type bundleMockSuite struct {
+	coretesting.BaseSuite
+	bundleClient *bundle.Client
+}
+
+var _ = gc.Suite(&bundleMockSuite{})
+
+func newClient(f basetesting.APICallerFunc, ver int) *bundle.Client {
+	return bundle.NewClient(basetesting.BestVersionCaller{f, ver})
+}
+
+func (s *bundleMockSuite) TestFailExportBundlev1(c *gc.C) {
+	client := newClient(
+		func(objType string,
+			version int,
+			id,
+			request string,
+			args,
+			response interface{},
+		) error {
+			c.Check(objType, gc.Equals, "Bundle")
+			c.Check(id, gc.Equals, "")
+			c.Check(request, gc.Equals, "ExportBundle")
+			c.Assert(args, gc.Equals, nil)
+			result := response.(*params.StringResult)
+			result.Result = ""
+			return nil
+		}, 1,
+	)
+	result, err := client.ExportBundle()
+	c.Assert(err, gc.ErrorMatches, "command not supported on v1")
+	c.Assert(result, jc.DeepEquals, "")
+}
+
+func (s *bundleMockSuite) TestExportBundlev2(c *gc.C) {
+	client := newClient(
+		func(objType string, version int,
+			id,
+			request string,
+			args,
+			response interface{},
+		) error {
+			c.Check(objType, gc.Equals, "Bundle")
+			c.Check(id, gc.Equals, "")
+			c.Check(request, gc.Equals, "ExportBundle")
+			c.Assert(args, gc.Equals, nil)
+			result := response.(*params.StringResult)
+			result.Result = "applications:\n  " +
+				"ubuntu:\n    " +
+				"charm: cs:trusty/ubuntu\n    " +
+				"series: trusty\n    " +
+				"num_units: 1\n    " +
+				"to:\n    " +
+				"- \"0\"\n    " +
+				"options:\n      " +
+				"key: value\n" +
+				"series: xenial\n" +
+				"relations:\n" +
+				"- []\n"
+			return nil
+		}, 2,
+	)
+	result, err := client.ExportBundle()
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(result, jc.DeepEquals, "applications:\n  "+
+		"ubuntu:\n    "+
+		"charm: cs:trusty/ubuntu\n    "+
+		"series: trusty\n    "+
+		"num_units: 1\n    "+
+		"to:\n    "+
+		"- \"0\"\n    "+
+		"options:\n      "+
+		"key: value\n"+
+		"series: xenial\n"+
+		"relations:\n"+
+		"- []\n")
+}
--- a/api/bundle/client_test.go
+++ b/api/bundle/client_test.go
@@ -4,11 +4,13 @@
 package bundle_test
 
 import (
+	"github.com/juju/errors"
 	jc "github.com/juju/testing/checkers"
 	gc "gopkg.in/check.v1"
 
 	basetesting "github.com/juju/juju/api/base/testing"
 	"github.com/juju/juju/api/bundle"
+	"github.com/juju/juju/apiserver/common"
 	"github.com/juju/juju/apiserver/params"
 	coretesting "github.com/juju/juju/testing"
 )
@@ -55,11 +57,6 @@ func (s *bundleMockSuite) TestExportBundlev2(c *gc.C) {
 			args,
 			response interface{},
 		) error {
-			c.Check(objType, gc.Equals, "Bundle")
-			c.Check(id, gc.Equals, "")
-			c.Check(request, gc.Equals, "ExportBundle")
-			c.Assert(args, gc.Equals, nil)
-			c.Assert(response, gc.FitsTypeOf, &params.StringResult{})
 			result := response.(*params.StringResult)
 			result.Result = "applications:\n  " +
 				"ubuntu:\n    " +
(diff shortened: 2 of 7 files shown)


api/bundle/client.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/juju/juju/api/base"
+	"github.com/juju/juju/apiserver/params"
+)
+
+var logger = loggo.GetLogger("juju.api.bundle")
+
+// Client allows access to the bundle API end point.
+type Client struct {
+	base.ClientFacade
+	st     base.APICallCloser
```
anastasiamac: I wonder if you actually need it? Since you are not using it anywhere except in the constructor to create the client facade instance...

api/bundle/client.go
```diff
@@ -0,0 +1,50 @@
...
+	if bestVer := c.BestAPIVersion(); bestVer < 2 {
+		return "", errors.Errorf("command not supported on v%d", bestVer)
+	}
+
+	if err := c.facade.FacadeCall("ExportBundle", nil, &result); err != nil {
+		return "", errors.Trace(err)
+	}
+
+	if len(result.Result) == 0 {
+		return "", errors.Errorf("result obtained is incorrect")
```
anastasiamac: This is pretty obscure as far as error messages go... Any way we can re-phrase it to be more descriptive/helpful?

api/bundle/client.go
```diff
@@ -0,0 +1,50 @@
...
+	var result params.StringResult
+	if bestVer := c.BestAPIVersion(); bestVer < 2 {
+		return "", errors.Errorf("command not supported on v%d", bestVer)
+	}
+
+	if err := c.facade.FacadeCall("ExportBundle", nil, &result); err != nil {
+		return "", errors.Trace(err)
+	}
+
+	if len(result.Result) == 0 {
```
anastasiamac: What happens if result.Error is not empty?

api/bundle/client_test.go
```diff
@@ -0,0 +1,92 @@
...
+	bundleClient *bundle.Client
+}
+
+var _ = gc.Suite(&bundleMockSuite{})
+
+func newClient(f basetesting.APICallerFunc, ver int) *bundle.Client {
+	return bundle.NewClient(basetesting.BestVersionCaller{f, ver})
+}
+
+func (s *bundleMockSuite) TestFailExportBundlev1(c *gc.C) {
```
anastasiamac: Awesome \o/ Thank you :)

api/bundle/client_test.go
```diff
@@ -0,0 +1,92 @@
...
+				"series: trusty\n    " +
+				"num_units: 1\n    " +
+				"to:\n    " +
+				"- \"0\"\n    " +
+				"options:\n      " +
+				"key: value\n" +
+				"series: xenial\n" +
+				"relations:\n" +
+				"- []\n"
+			return nil
```
anastasiamac: Please also add tests when:
(1) an error is returned;
(2) result.Result has len 0;
(3)  as well as where result has non-nil Error.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,117 @@
...
+
+// NewExportBundleCommand returns a fully constructed export bundle command.
+func NewExportBundleCommand() cmd.Command {
+	return modelcmd.Wrap(&exportBundleCommand{})
+}
+
+type exportBundleCommand struct {
+	modelcmd.ModelCommandBase
+	out cmd.Output
+	api ExportBundleModelAPI
```
anastasiamac: Please have it as func to access api instead of the api directly, something like
```
newAPIFunc func() (ExportBundleAPI, error)
```
There are commands that have similar accessors.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,117 @@
...
+	f.StringVar(&c.Filename, "filename", "", "Export Model")
+}
+
+// Init implements Command.
+func (c *exportBundleCommand) Init(args []string) error {
+	return cmd.CheckEmpty(args)
+}
+
+// ExportBundleAPI specifies the used function calls of the ModelManager.
+type ExportBundleModelAPI interface {
```
anastasiamac: Comment does not match the actual name nor the actual purpose. ModelManager has nothing to do here... cut-and-paste error maybe? Also I think that ExportBundleAPI makes more sense.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,117 @@
...
+	api, err := c.NewAPIRoot()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	return bundle.NewClient(api), nil
+}
+
+// Run implements Command.
+func (c *exportBundleCommand) Run(ctx *cmd.Context) error {
+	client, err := c.getAPI()
```
anastasiamac: This should be calling c.newAPIFunc(). The default implementation of this function would indeed be getAPI() but can be overwritten by tests.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	var result string
+	if client.BestAPIVersion() > 1 {
```
anastasiamac: I am not sure if you need it since you have a similar check in api...

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,117 @@
...
+		result, err = client.ExportBundle()
+		if err != nil {
+			return err
+		}
+	}
+
+	if c.Filename != "" {
```
anastasiamac: This would be cleaner in inverse...
```
if c.Filename == " {
 return c.out.Write(ctx, result)
}
```

and then just proceed with the writing into a file logic. No need for else :D

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,113 @@
...
+type ExportBundleCommandSuite struct {
+	testing.FakeJujuXDGDataHomeSuite
+	fake  *fakeExportBundleClient
+	stub  *jujutesting.Stub
+	store *jujuclient.MemStore
+}
+
+var _ = gc.Suite(&ExportBundleCommandSuite{})
+
+type fakeExportBundleClient struct {
```
anastasiamac: Personal preference is to keep mocks and fakes and other handy pieces at the bottom of the file since the tests are the actual interesting bits here :)

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (s *ExportBundleCommandSuite) SetUpTest(c *gc.C) {
+	s.FakeJujuXDGDataHomeSuite.SetUpTest(c)
+	s.stub = &jujutesting.Stub{}
+	s.fake = &fakeExportBundleClient{
+		Stub:           s.stub,
+		bestAPIVersion: 2,
+	}
+	s.store = jujuclient.NewMemStore()
+	s.store.CurrentControllerName = "testing"
```
anastasiamac: I don't think that you actually need to populate the store to be valid here... You are not using it.

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func (f *fakeExportBundleClient) Close() error { return nil }
+
+func (f *fakeExportBundleClient) ExportBundle() (string, error) {
+	f.MethodCall(f, "ExportBundle")
+	if err := f.NextErr(); err != nil {
+		return "", err
+	}
+	return "applications:\n" +
```
anastasiamac: You want this string to be a variable. You might also want to test what happens if it's empty?

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,113 @@
...
+func (f *fakeExportBundleClient) BestAPIVersion() int {
+	return f.bestAPIVersion
+}
+
+func (s *ExportBundleCommandSuite) SetUpTest(c *gc.C) {
+	s.FakeJujuXDGDataHomeSuite.SetUpTest(c)
+	s.stub = &jujutesting.Stub{}
+	s.fake = &fakeExportBundleClient{
+		Stub:           s.stub,
+		bestAPIVersion: 2,
```
anastasiamac: You definitely want to check what happens when you have v1.

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	err := s.store.UpdateModel("testing", "admin/mymodel", jujuclient.ModelDetails{
+		ModelUUID: testing.ModelTag.Id(),
+		ModelType: coremodel.IAAS,
+	})
+	c.Assert(err, jc.ErrorIsNil)
+	s.store.Models["testing"].CurrentModel = "admin/mymodel"
+}
+
+func (s *ExportBundleCommandSuite) TestExportBundleNoFilename(c *gc.C) {
```
anastasiamac: Please also test what happens if your api call returns an error; as well as file writing logic.

api/bundle/client.go
vinu2003: done

api/bundle/client.go
vinu2003: u r correct. i left this because may be required for signature used in all client facades.

api/bundle/client.go
vinu2003: yes. i have captured this as well.

api/bundle/client_test.go
vinu2003: u r welcome :) it doesnt panic.

cmd/juju/model/exportbundle.go
vinu2003: Done

cmd/juju/model/exportbundle.go
vinu2003: done.

cmd/juju/model/exportbundle.go
vinu2003: done.

cmd/juju/model/exportbundle.go
vinu2003: i think its redundant.

cmd/juju/model/exportbundle.go
vinu2003: done

cmd/juju/model/exportbundle_test.go
vinu2003: done

cmd/juju/model/exportbundle_test.go
vinu2003: I think its required to register the controller. Else we have to do boottsrap or register with other controller.

cmd/juju/model/exportbundle_test.go
vinu2003: i have simplified this attr and also bestVersion as part of fakeExportBundleAPI struct so that each test sets the value.

cmd/juju/model/exportbundle_test.go
vinu2003: same as above. Now each test will set this attr - bestAPIVersion.

cmd/juju/model/exportbundle_test.go
vinu2003: All done.

api/bundle/client_test.go
vinu2003: actually the production code has meaningless chk for len == 0. the server code never returns a non-nil with empty. Only in success case it returns nil. customized based on the calle function and addressed the tests.

api/bundle/client.go
```diff
@@ -0,0 +1,45 @@
...
+
+// ExportBundle exports the current model configuration.
+func (c *Client) ExportBundle() (string, error) {
+	var result params.StringResult
+	if bestVer := c.BestAPIVersion(); bestVer < 2 {
+		return "", errors.Errorf("command not supported on v%d", bestVer)
+	}
+
+	if err := c.facade.FacadeCall("ExportBundle", nil, &result); err != nil {
+		return "", errors.Annotate(result.Error, "export failed")
```
anastasiamac: You should be annotating err here and then there should be a separate handling of result.Error. Both are needed and generally convey different information.

api/bundle/client_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (s *bundleMockSuite) TestExportBundleErrorv2(c *gc.C) {
+	client := newClient(
+		func(objType string, version int,
+			id,
+			request string,
+			args,
+			response interface{},
+		) error {
+			c.Check(objType, gc.Equals, "Bundle")
```
anastasiamac: You do not need check this in every test. Only one test needs these.

cmd/juju/model/export_test.go
```diff
@@ -67,6 +67,15 @@ func NewDumpDBCommandForTest(api DumpDBAPI, store jujuclient.ClientStore) cmd.Co
 	return modelcmd.Wrap(cmd)
 }
 
+// NewDumpCommandForTest returns a DumpCommand with the api provided as specified.
```
anastasiamac: Comment is not correct.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,115 @@
...
+		return cmd.getAPI()
+	}
+	return modelcmd.Wrap(cmd)
+}
+
+type exportBundleCommand struct {
+	modelcmd.ModelCommandBase
+	out        cmd.Output
+	newAPIFunc func() (ExportBundleAPI, error)
+	// name of the charm bundle file.
```
anastasiamac: Either remove or re-phrase. At this stage, with information this comment contains it is not helpful :)

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+const exportBundleHelpDoc = `
+Exports the current model configuration into a YAML file.
+
+Examples:
+
+    juju export-bundle
+
+If --filename is not used, the bundle is displayed in stdout.
```
anastasiamac: Please move to before Examples section. Also, please re-phrase.. For example, I would prefer to see something that says that the filename is optional, the export will be in that file when provided, etc.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,115 @@
...
+`
+
+// Info implements Command.
+func (c *exportBundleCommand) Info() *cmd.Info {
+	return &cmd.Info{
+		Name:    "export-bundle",
+		Purpose: "Exports the current model configuration in a charm bundle.",
```
anastasiamac: "as a bundle"?

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,115 @@
...
+		Doc:     exportBundleHelpDoc,
+	}
+}
+
+// SetFlags implements Command.
+func (c *exportBundleCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.ModelCommandBase.SetFlags(f)
+	c.out.AddFlags(f, "yaml", output.DefaultFormatters)
+	f.StringVar(&c.Filename, "filename", "", "Export Model")
```
anastasiamac: "Export Model" is not actually a what filename is. This description will be displayed as help and in other context, it needs to be related to the variable.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// Init implements Command.
+func (c *exportBundleCommand) Init(args []string) error {
+	return cmd.CheckEmpty(args)
+}
+
+// ExportBundleAPI specifies the used function calls of the ModelManager.
```
anastasiamac: Not ModelManager

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,115 @@
...
+		return c.out.Write(ctx, result)
+	}
+	filename := c.Filename + ".yaml"
+	file, err := os.Create(filename)
+	if err != nil {
+		return errors.Annotate(err, "while creating local file")
+	}
+	defer file.Close()
+
+	// Write out the result.
```
anastasiamac: I ma not sure that these comments help.. The code says it all in this case and below :D

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,165 @@
...
+type ExportBundleCommandSuite struct {
+	testing.FakeJujuXDGDataHomeSuite
+	fake  *fakeExportBundleClient
+	stub  *jujutesting.Stub
+	store *jujuclient.MemStore
+}
+
+var _ = gc.Suite(&ExportBundleCommandSuite{})
+
+func (f *fakeExportBundleClient) Close() error { return nil }
```
anastasiamac: Please move all these fakeExportBundleClient function next to the struct declaration, at the end of the file.

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	err := s.store.UpdateModel("testing", "admin/mymodel", jujuclient.ModelDetails{
+		ModelUUID: testing.ModelTag.Id(),
+		ModelType: coremodel.IAAS,
+	})
+	c.Assert(err, jc.ErrorIsNil)
+	s.store.Models["testing"].CurrentModel = "admin/mymodel"
+}
+
+func (s *ExportBundleCommandSuite) TestExportBundleFailOnv1(c *gc.C) {
```
anastasiamac: I do not think that you need this as there are no version checks on the command now.

I think you can do without bestAPIVersion var on the fake and just keep this test to check that command displays api errors in  an adequate way :D

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,165 @@
...
+	s.fake.CheckCalls(c, []jujutesting.StubCall{
+		{"ExportBundle", nil},
+	})
+
+	out := cmdtesting.Stdout(ctx)
+	c.Assert(out, gc.Equals, "")
+	c.Assert(err, gc.ErrorMatches, "command not supported on v1")
+}
+
+func (s *ExportBundleCommandSuite) TestExportBundleFailEmptyResult(c *gc.C) {
```
anastasiamac: This is not needed as there is no relevant logic for this on the command. This is just checking that you are handling api error and you already have one test for it.

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,165 @@
...
+		{"ExportBundle", nil},
+	})
+
+	out := cmdtesting.Stdout(ctx)
+	c.Assert(out, gc.Equals, "Bundle successfully exported to mymodel.yaml\n")
+}
+
+type fakeExportBundleClient struct {
+	*jujutesting.Stub
+	bestAPIVersion int
```
anastasiamac: You do not need this on your fake.

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,165 @@
...
+		"  machines:\n"+
+		"    \"0\": {}\n"+
+		"    \"1\": {}\n"+
+		"  series: xenial\n"+
+		"  relations:\n"+
+		"  - - wordpress:db\n"+
+		"    - mysql:mysql\n")
+}
+
+func (s *ExportBundleCommandSuite) TestExportBundleSuccessFilename(c *gc.C) {
```
anastasiamac: I would have expected more tests here to ensure that the logic around writing to file is tested.

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,165 @@
...
+	s.fake.bestAPIVersion = 2
+
+	ctx, err := cmdtesting.RunCommand(c, model.NewExportBundleCommandForTest(s.fake, s.store), "--filename", "mymodel")
+	c.Assert(err, jc.ErrorIsNil)
+	s.fake.CheckCalls(c, []jujutesting.StubCall{
+		{"ExportBundle", nil},
+	})
+
+	out := cmdtesting.Stdout(ctx)
+	c.Assert(out, gc.Equals, "Bundle successfully exported to mymodel.yaml\n")
```
anastasiamac: You need to check the contents of the file :)

api/bundle/client.go
vinu2003: get ur point. i missed err here.

vinu2003: i have annotated result.Error instead of err which can be traced.

api/bundle/client_test.go
vinu2003: yup. done

cmd/juju/model/exportbundle.go
vinu2003: done

cmd/juju/model/exportbundle.go
vinu2003: 'bundle file'

cmd/juju/model/exportbundle.go
vinu2003: done

cmd/juju/model/exportbundle.go
vinu2003: cleaned

cmd/juju/model/exportbundle_test.go
vinu2003: done

cmd/juju/model/exportbundle_test.go
vinu2003: i shd have removed this test when i cleaned up the chk in Run(). however, i am pretty uncertain still. Iam cleaning up for now.

cmd/juju/model/exportbundle_test.go
vinu2003: ok

cmd/juju/model/exportbundle_test.go
vinu2003: cleaned up in production code as well.

api/bundle/client.go
```diff
@@ -0,0 +1,46 @@
...
+
+// ExportBundle exports the current model configuration.
+func (c *Client) ExportBundle() (string, error) {
+	var result params.StringResult
+	if bestVer := c.BestAPIVersion(); bestVer < 2 {
+		return "", errors.Errorf("command not supported on v%d", bestVer)
+	}
+
+	if err := c.facade.FacadeCall("ExportBundle", nil, &result); err != nil {
+		errors.Annotate(result.Error, "export failed")
```
anastasiamac: You need to deal with err here and result.Error separately. They are not the same.

cmd/juju/model/exportbundle_test.go
vinu2003: done

api/bundle/client.go
```diff
@@ -0,0 +1,50 @@
...
+	return &Client{
+		ClientFacade: frontend,
+		facade:       backend}
+}
+
+// ExportBundle exports the current model configuration.
+func (c *Client) ExportBundle() (string, error) {
+	var result params.StringResult
+	if bestVer := c.BestAPIVersion(); bestVer < 2 {
+		return "", errors.Errorf("command not supported on v%d", bestVer)
```
anastasiamac: This is not a command. This should be 

return "", errors.NotSupportedf("ExportBundle() on v%d" ....)

api/bundle/client.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	if err := c.facade.FacadeCall("ExportBundle", nil, &result); err != nil {
+		errors.Annotate(result.Error, "export failed")
```
anastasiamac: No we do not want to annotate result.Error here. In fact we do not even want to touch here. 
In fact, there is no guarantee that result is valid since you got an error from the call.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,113 @@
...
+	modelcmd.ModelCommandBase
+	out        cmd.Output
+	newAPIFunc func() (ExportBundleAPI, error)
+	Filename   string
+}
+
+const exportBundleHelpDoc = `
+Exports the current model configuration as bundle.
+
+If --filename is optional. The charm bundle is written in the filename provided as value.
```
anastasiamac: This does not read smoothly to me :)

cmd/juju/model/exportbundle_test.go
anastasiamac: I am pretty sure that we do not want to leave behind the files created by tests.  You would need to have a TearDownTest os.Remove'ing the file.. There are some examples around, especially on the commands that do use files like "create-backup", etc.

api/bundle/client.go
vinu2003: I have removed.

cmd/juju/model/exportbundle.go
vinu2003: i have tried my best :)

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,113 @@
...
+	modelcmd.ModelCommandBase
+	out        cmd.Output
+	newAPIFunc func() (ExportBundleAPI, error)
+	Filename   string
+}
+
+const exportBundleHelpDoc = `
+Exports the current model configuration as bundle.
+
+--filename specifies the file where bundle is written to. Its optional.
```
anastasiamac: Maybe @pmatulis can help phrasing this? :D

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,113 @@
...
+If --filename is not used, the bundle is displayed in stdout.
+
+Examples:
+
+    juju export-bundle
+	juju export-bundle --filename mymodel
```
anastasiamac: The indent is different here.

cmd/juju/model/exportbundle.go
```diff
@@ -0,0 +1,113 @@
...
+		Purpose: "Exports the current model configuration as bundle.",
+		Doc:     exportBundleHelpDoc,
+	}
+}
+
+// SetFlags implements Command.
+func (c *exportBundleCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.ModelCommandBase.SetFlags(f)
+	c.out.AddFlags(f, "yaml", output.DefaultFormatters)
+	f.StringVar(&c.Filename, "filename", "", "bundle file")
```
anastasiamac: Please have a look at other example for flags descriptions in the commands. This is meant to be a phrase, starting with a capital :D

cmd/juju/model/exportbundle_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/juju/cmd/cmdtesting"
+	jujutesting "github.com/juju/testing"
+	jc "github.com/juju/testing/checkers"
+	gc "gopkg.in/check.v1"
+
+	"github.com/juju/juju/cmd/juju/model"
+	coremodel "github.com/juju/juju/core/model"
+	"github.com/juju/juju/jujuclient"
+	"github.com/juju/juju/testing"
+	"os"
```
anastasiamac: move me

cmd/juju/model/exportbundle.go
vinu2003: Thanks anastasiamac. @pmatulis Could you please help us in phrasing this new export-bundle command doc ?

cmd/juju/model/exportbundle.go
vinu2003: I have seen cases with --filename. thgats why i chose this.

cmd/juju/model/exportbundle_test.go
vinu2003: done

cmd/juju/model/exportbundle.go
vinu2003: corrected that

api/bundle/client.go
vinu2003: @anastasiamac We can discuss this tomorrow. I do understand that we have 2 diff errors. 
But in this scenario from server function call ExportBundle() when the function returns error(ie non-nil) the result.Error is always nil. And, in case of  err is nil, there are no scenarios in that function that sets result.Error like for example in credentialValidator.go @@Line77 and @@78.

So what i was trying to point out it for this function ExportBundle(0 from sever side we never come across a state where err=nil  but result.Error="having some value".

cmd/juju/model/exportbundle.go
pmatulis: Conceptually, I see the workflow as a) creating a bundle and then b) exporting it. Therefore:

```
Exports a bundle.
...
If --filename is not used, the configuration is printed to stdout.
...
 --filename specifies an output file.
...
```

That being said, is it actually useful to send output to STDOUT? I'm wondering whether we should not simply enforce an output file.

vinu2003: The user  can also manually redirect the output to a file instead of STDOUT which is same as providing the --fielname option.

cmd/juju/model/exportbundle.go
vinu2003: corrected

cmd/juju/model/exportbundle_test.go
vinu2003: done

cmd/juju/model/exportbundle_test.go
vinu2003: added few more ard the logic of writing files.

cmd/juju/model/exportbundle.go
vinu2003: Thank you @pmatulis

vinu2003:
!!build!!

vinu2003:
!!build!!

vinu2003:
$$merge$$

PR #8968, opened 2018-07-24
Store registry and image path separately.

Veebers:
## Description of change

Splits registry and image data and stores it in the meta data for the docker image resource.
This is to move toward a stable data model for the docker resources with a mind on being able to do a resource-get on behalf of a charm so we can refresh secrets with new passwords.

## QA steps

Covered by unit tests, only data storage at this point.

## Documentation changes

None needed, internal data storage changes.


wallyworld:
$$merge$$

PR #8976, opened 2018-07-25
Add clarifications to snap build instructions

dshcherb:
Clarifications around building locally-modified versions of a Juju snap.

I remember stumbling upon missing changes in the final snap multiple times because I have not committed them. Maybe we can do some kind of auto-check for this but a documented behavior is an improvement.

Likewise applying patches multiple times to the same cloned repo is not idempotent which spends your time when you are actually trying to save it by using `snapcraft clean -s build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,17 @@ Building Juju as a Snap Package
 
 Building
 --------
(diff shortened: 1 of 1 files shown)


howbazaar:
$$merge$$

PR #8981, opened 2018-07-26
Credential override during bootstrap

SimonRichardson:
## Description of change

Background: with the introduction of `autoload-credential` improvements
to the LXD provider, it now makes it possible to be stuck in a situation where
by you can't launch the localhost lxd provider. This is because the logic
around bootstrap identifies (correctly!) that there are multiple credentials 
available to that one provider. The newly auto detected credentials and the
localhost or any other ones the user has successfully launched.

The problem is compounded more, by the fact that "localhost" for the LXD
provider is treated as a special case, as well as the fact that the LXD 
provider deals will different semantics; local vs remote/cluster servers.

So this PR aim to solve that by allowing the operator to specify the credential
to use at the right time.

```sh
juju bootstrap nuc5 --credential=nuc5
```

Note: this only happens when you have multiple credentials, for example
using the `autoload-credentials`

## QA steps

Setup your local lxc to add remote LXD server connections.

```sh
juju bootstrap nuc5 --credential=nuc5
```

Remove that remote LXD server connection from your local LXD setup
and test that it can connect to your localhost

```
```

## Documentation changes

Yes, that fact that `--credential=` argument is available.

## Bug reference

N/A

### DIFF
--- a/cmd/juju/common/cloudcredential.go
+++ b/cmd/juju/common/cloudcredential.go
@@ -50,8 +50,11 @@ func GetOrDetectCredential(
 	// the environment.
 	ctx.Verbosef("no credentials found, checking environment")
 
-	detected, detectErr := modelcmd.DetectCredential(args.Cloud.Name, args.CredentialName, provider)
-	if err != nil && detectErr != nil && args.CredentialName != "" {
+	detected, selected, detectErr := modelcmd.DetectCredential(args.Cloud.Name, args.CredentialName, provider)
+	// Attempt to detect a credential if the credential name is not empty, but
+	// detect credential doesn't return a selected credential based on the
+	// credential name.
+	if !selected && args.CredentialName != "" {
 		return credential, chosenCredentialName, regionName, false, err
 	}
 	if errors.Cause(detectErr) == modelcmd.ErrMultipleCredentials {
--- a/cmd/juju/common/cloudcredential.go
+++ b/cmd/juju/common/cloudcredential.go
@@ -17,7 +17,7 @@ import (
 	"github.com/juju/juju/jujuclient"
 )
 
-// ErrMultipleCredentialsDetected is the error returned by
+// ErrMultipleDetectedCredentials is the error returned by
 // GetOrDetectCredential when multiple credentials are
 // detected, meaning Juju cannot choose one automatically.
 var ErrMultipleDetectedCredentials = errors.New("multiple detected credentials")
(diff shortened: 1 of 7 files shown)


SimonRichardson:
https://trello.com/c/3vpps8BN/402-lxd-select-one-from-the-multi-credential

cmd/juju/common/cloudcredential.go
```diff
@@ -50,11 +50,17 @@ func GetOrDetectCredential(
...
-	detected, err := modelcmd.DetectCredential(args.Cloud.Name, provider)
-	if errors.Cause(err) == modelcmd.ErrMultipleCredentials {
+	detected, selected, detectErr := modelcmd.DetectCredential(args.Cloud.Name, args.CredentialName, provider)
+	// Attempt to detect a credential if the credential name is not empty, but
+	// detect credential doesn't return a selected credential based on the
+	// credential name.
+	if !selected && args.CredentialName != "" {
+		return credential, chosenCredentialName, regionName, false, err
+	}
+	if errors.Cause(detectErr) == modelcmd.ErrMultipleCredentials {
```
anastasiamac: This could be jsut
```
if detectErr != nil {
    if errors.Cause(detectErr) == modelcmd.ErrMultipleCredentials {
        return fail(ErrMultipleDetectedCredentials)
    }
    return fail(errors.Trace(detectErr))
}
```

cmd/juju/controller/destroy.go
```diff
@@ -54,6 +54,7 @@ type destroyCommand struct {
 	destroyModels  bool
 	destroyStorage bool
 	releaseStorage bool
+	credentialName string
```
anastasiamac: I am not too sure that this is a good idea... We have measures to ensure that your user has permissions to destroy a model. I am not sure that we want to a different cloud credential than the one that controller is using... Especially since I am not convinced that we have a check that this new credential will have access to all controller "assets"...

cmd/juju/controller/kill.go
```diff
@@ -63,15 +63,17 @@ func wrapKillCommand(kill *killCommand) modelcmd.Command {
 type killCommand struct {
 	destroyCommandBase
 
-	clock   clock.Clock
-	timeout time.Duration
+	clock          clock.Clock
+	timeout        time.Duration
+	credentialName string
```
anastasiamac: Same concerns as for credential in destroy...

cmd/plugins/juju-metadata/imagemetadata.go
```diff
@@ -109,6 +112,7 @@ func (c *imageMetadataCommand) SetFlags(f *gnuflag.FlagSet) {
 	f.StringVar(&c.Stream, "stream", imagemetadata.ReleasedStream, "the image stream")
 	f.StringVar(&c.VirtType, "virt-type", "", "the image virtualisation type")
 	f.StringVar(&c.Storage, "storage", "", "the type of root storage")
+	f.StringVar(&c.CredentialName, "credential", "", "the credentials type to use")
```
anastasiamac: This seems to be the name not the type of the credentials to use.

PR #8982, opened 2018-07-26
Make sure we set the default credential

SimonRichardson:
## Description of change

Background: with the introduction of autoload-credential improvements
to the LXD provider, it now makes it possible to be stuck in a situation where
by you can't launch the localhost lxd provider. This is because the logic
around bootstrap identifies (correctly!) that there are multiple credentials
available to that one provider. The newly auto detected credentials and the
localhost or any other ones the user has successfully launched.

The problem is compounded more, by the fact that "localhost" for the LXD
provider is treated as a special case, as well as the fact that the LXD
provider deals will different semantics; local vs remote/cluster servers.

So this PR aim to solve that by making sure there is a default one.

```
juju autoload-credentials
juju bootstrap localhost
```

Note: this only happens when you have multiple credentials, for example
using the autoload-credentials

## QA steps

Setup your local lxc to add remote LXD server connections.

```sh
juju autoload-credentials
```

## Documentation changes

N/A

## Bug reference

N/A

### DIFF
--- a/provider/lxd/credentials.go
+++ b/provider/lxd/credentials.go
@@ -163,7 +163,8 @@ func (p environProviderCredentials) DetectCredentials() (*cloud.CloudCredential,
 	}
 
 	return &cloud.CloudCredential{
-		AuthCredentials: authCredentials,
+		DefaultCredential: lxdnames.DefaultCloud,
+		AuthCredentials:   authCredentials,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


SimonRichardson:
https://trello.com/c/3vpps8BN/402-lxd-select-one-from-the-multi-credential

SimonRichardson:
!!build!!

anastasiamac:
!!build!!

anastasiamac:
$$merge$$

PR #8984, opened 2018-07-26
fix lp:1783633 juju destroy-controllers fails when using an Openstack with a self-signed certificate

hmlanigan:
## Description of change

Updated: Write cloud ca certs to the client store file bootstrap-config.yaml.  Add those certs to the bootstrap params used by destroy-controller.

## QA steps

Using an openstack deployment requiring a CACert for access:

1. juju add-cloud, the endpoint url should validate.
2. Add the ca cert to the clouds.yaml for the new cloud:
clouds:
openstack:
type: openstack
.....
ca-certificates:
|
-----BEGIN CERTIFICATE-----
.....
-----END CERTIFICATE-----
3. juju bootstrap; juju deploy ubuntu
4. juju destroy-controller --destroy-all-models --yes --destroy-storage <controller>

Step 4 no longer fails "ERROR destroying controller model: destroying instances: failed to get list of server details"

There should be no change when using other other OpenStack deployments.

## Documentation changes

not yet, there are a few more pieces of the feature to fix.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1783633

### DIFF
--- a/cmd/modelcmd/base.go
+++ b/cmd/modelcmd/base.go
@@ -537,6 +537,7 @@ func (g bootstrapConfigGetter) getBootstrapConfigParams(controllerName string) (
 			IdentityEndpoint: bootstrapConfig.CloudIdentityEndpoint,
 			StorageEndpoint:  bootstrapConfig.CloudStorageEndpoint,
 			Credential:       credential,
+			CACertificates:   bootstrapConfig.CloudCACertificates,
 		},
 		Config: cfg,
 	}, nil
--- a/jujuclient/bootstrapconfigfile_test.go
+++ b/jujuclient/bootstrapconfigfile_test.go
@@ -47,7 +47,48 @@ controllers:
     cloud: maas
     type: maas
     region: 127.0.0.1
+    ca-certificates:
+    - |
+      -----BEGIN CERTIFICATE-----
+      MIIC0DCCAbigAwIBAgIUeIj3r4ocrSubOmb1yPxmoiRfhO8wDQYJKoZIhvcNAQEL
+      BQAwDzENMAsGA1UEAwwET1NDSTAeFw0xODA3MTUxNjE2NTZaFw0xODEwMjQxNjE2
+      NTZaMA8xDTALBgNVBAMMBE9TQ0kwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEK
+      AoIBAQCpCLMLGZpLojudOwrupsbk2ESCQO4/kEOF6L5YHxcUrRxcrxu0DmnwWYcK
+      pjKL9K3U7xSiSL+MtNff7MfBYbV0SOfjHR0/gqwio0JeYONABeeynUZkuXg1CXuG
+      uMHcmPjCAWnLyAnlF4Wwavv6pPdM4l1X4lt1b2ez8G6u4+UPg/zNt473aqOzwMzy
+      B3aToHSHOoDvXQDwtDkR0PimyEtHVz/17AcwSHzMqNGLgLFEx0SPuYJus8WJg1Sn
+      c9kqrvIUBnZzjtbCquCxLRxG2xHdvBxOesbRyJPO0ypqEcTMtrX9rmJce67HG+4h
+      EgLCEpcgfSVyH9PS3wdUAfkr9KE9AgMBAAGjJDAiMA8GA1UdEQQIMAaCBE9TQ0kw
+      DwYDVR0TAQH/BAUwAwEB/zANBgkqhkiG9w0BAQsFAAOCAQEAFIYyqNayVFxZ1jcz
+      fdvEP2yVB9dq8vhSXU4lbkqlPw5q954bLURQzklqMfpXhhIbmrvq6LcLGaSkgmPp
+      CzlxMkjr8oTRVQUqNIfcJQKtwNOAGh7xZ77GPhBlfHJ8VhTFtDXPM/fj8GLr5Oav
+      gy9+QywhESKkwAn4+AubBRRtEDBX9zwc2hT5uqz1x1tcs16tKAZBIekwmMBJKkNs
+      61I+cRHoXtXFh8/upMC6eMAvv6eVHgqpcEWrVLvoBh7ivcsFuUD1IyuIlN4i6roh
+      xcSAzRCXqVe/BBsHqYyd8044vrIG7P7pYGaQm99nFGylTBfSh5g1LrYV7IJP6KkG
+      6JHZXg==
+      -----END CERTIFICATE-----
 `
+
+var fakecert = `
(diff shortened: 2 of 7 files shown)


hmlanigan:
Question: juju controller-config will now display 1 to 2 CACerts.  Should we only all the cert to be displayed with --format tabular?  To make the output easier to read?  This is currently done with juju model-config and cloudinit-userdata.

hmlanigan:
[DestroyController() fails due to x509 auth, find the client used and update](https://trello.com/c/wLXDyjub/400-destroycontroller-fails-due-to-x509-auth-find-the-client-used-and-update)

cmd/juju/commands/bootstrap.go
```diff
@@ -1025,6 +1025,11 @@ func (c *bootstrapCommand) bootstrapConfigs(
 	if err != nil {
 		return bootstrapConfigs{}, errors.Annotate(err, "constructing bootstrap config")
 	}
+
+	if len(cloud.CACertificates) > 0 {
```
ExternalReality: If the cloud has certificates then the controller is going to need access to those certificates, ok, I understand this.

cmd/modelcmd/base.go
```diff
@@ -539,7 +540,12 @@ func (g bootstrapConfigGetter) getBootstrapConfigParams(controllerName string) (
 			Credential:       credential,
 		},
 		Config: cfg,
-	}, nil
+	}
+	if cloudCACerts, ok := bootstrapConfig.ControllerConfig.CloudCACerts(); ok {
```
ExternalReality: If the Controller config was given certs (as per above) the we want to populate the `prepareCfgParams` with them. Looks good.

controller/config.go
```diff
@@ -212,6 +216,9 @@ var (
 	// AllowedUpdateConfigAttributes contains all of the controller
 	// config attributes that are allowed to be updated after the
 	// controller has been created.
+	//
+	// TODO: (hml) 2018-07-25
+	// should CloudCACertsKey be in this list?
```
ExternalReality: Who would know the answer to this question?

hmlanigan:
@wallyworld, CaaS is the only other place that the CloudSpec.CACertificates is used.  Your review would be appreciated to answer a few questions and ensure there isn't a negative impact to CaaS.  Thank you.

controller/config.go
hmlanigan: On reflection, I believe the answer is yes, CloudCACertsKey should be in AllowedUpdateConfigAttributes.  Certificates can expire and need to be refreshed.

mitechie:
I'll bug @wallyworld about how these end up getting updated when the certs expire. That seems like a repeat of coud-creds waiting to happen.

hmlanigan:
destroy-controllers first calls getControllerEnvironFromStore(), if that fails uses getControllerEnvironFromAPI() which does what you're describing.  If getControllerEnvironFromAPI() is used, no changes are necessary.  However the data saved in the store, is not as complete as that saved in state or on the controller and does not include the same cloud info that the api returns.

Is there a reason for bootstrap-config.yaml accessible by the client to contain different data than the bootstrap-params file saved on the controller?

hmlanigan:
@wallyworld, @ExternalReality,  significant rewrite based on review comments, please review again.  TY!  Description is updated

jujuclient/interface.go
```diff
@@ -130,6 +130,10 @@ type BootstrapConfig struct {
 	// when communicating with the cloud's storage service. This will
 	// be empty for clouds that have no storage-specific API endpoint.
 	CloudStorageEndpoint string `yaml:"storage-endpoint,omitempty"`
+
+	// CloudCACertificates contains the CACertificates necessary to
+	// communicate with the cloud infrastructure.
+	CloudCACertificates []string `yaml:"controller-ca-certificates,omitempty"`
```
wallyworld: wrong yaml tag "controller-ca-certificates"

jujuclient/bootstrapconfigfile_test.go
```diff
@@ -47,8 +47,49 @@ controllers:
     cloud: maas
     type: maas
     region: 127.0.0.1
+    controller-ca-certificates:
```
wallyworld: cloud-ca-cert

jujuclient/bootstrapconfigfile_test.go
```diff
@@ -47,8 +47,49 @@ controllers:
...
+      fdvEP2yVB9dq8vhSXU4lbkqlPw5q954bLURQzklqMfpXhhIbmrvq6LcLGaSkgmPp
+      CzlxMkjr8oTRVQUqNIfcJQKtwNOAGh7xZ77GPhBlfHJ8VhTFtDXPM/fj8GLr5Oav
+      gy9+QywhESKkwAn4+AubBRRtEDBX9zwc2hT5uqz1x1tcs16tKAZBIekwmMBJKkNs
+      61I+cRHoXtXFh8/upMC6eMAvv6eVHgqpcEWrVLvoBh7ivcsFuUD1IyuIlN4i6roh
+      xcSAzRCXqVe/BBsHqYyd8044vrIG7P7pYGaQm99nFGylTBfSh5g1LrYV7IJP6KkG
+      6JHZXg==
+      -----END CERTIFICATE-----
 `
 
+var fakecert = `
```
wallyworld: This doesn't match the cert above

jujuclient/bootstrapconfigfile_test.go
```diff
@@ -79,6 +120,7 @@ var testBootstrapConfig = map[string]jujuclient.BootstrapConfig{
 		Cloud:               "maas",
 		CloudType:           "maas",
 		CloudRegion:         "127.0.0.1",
+		CloudCACertificates: []string{fakecert},
```
wallyworld: Where are we checking that this cert is correctly added to the bootstrap config yaml?

hmlanigan: In environs/bootstrap/prepare_test.go, in TestPrepare, a cert is provided to the CloudSpec in the PrepareParams.  Then it checks the controllerStore, for the new controller, including an assert that the store's bootstrap config contains the cert.

jujuclient/bootstrapconfigfile_test.go
hmlanigan: updated

jujuclient/bootstrapconfigfile_test.go
hmlanigan: changed to ca-certificates

jujuclient/interface.go
hmlanigan: changed to ca-certificates

hmlanigan:
!!build!!

hmlanigan:
$$merge$$

PR #8985, opened 2018-07-27
Added logic to consider provisioned H/W for mac

vinu2003:
## Description of change
This PR addresses the changes to consider provisioned hardware characteristics for machines constraints.

execute export-bundle command in developer-mode

## Documentation changes
N/A
## Bug reference
N/A

### DIFF
--- a/apiserver/facades/client/bundle/bundle.go
+++ b/apiserver/facades/client/bundle/bundle.go
@@ -218,7 +218,7 @@ func (b *BundleAPI) fillBundleData(model description.Model) (*charm.BundleData, 
 				Options:     application.CharmConfig(),
 				Annotations: application.Annotations(),
 			}
-			if result := b.constraints(application.Constraints()); len(result) != 0 {
+			if result := b.machineConstraints(application.Constraints()); len(result) != 0 {
 				newApplication.Constraints = strings.Join(result, " ")
 			}
 		} else {
@@ -235,7 +235,7 @@ func (b *BundleAPI) fillBundleData(model description.Model) (*charm.BundleData, 
 				Options:     application.CharmConfig(),
 				Annotations: application.Annotations(),
 			}
-			if result := b.constraints(application.Constraints()); len(result) != 0 {
+			if result := b.machineConstraints(application.Constraints()); len(result) != 0 {
 				newApplication.Constraints = strings.Join(result, " ")
 			}
 		}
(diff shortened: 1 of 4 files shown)


apiserver/facades/client/bundle/bundle.go
```diff
@@ -248,7 +248,44 @@ func (b *BundleAPI) fillBundleData(model description.Model) (*charm.BundleData,
 			Annotations: machine.Annotations(),
 			Series:      machine.Series(),
 		}
-		if result := b.constraints(machine.Constraints()); len(result) != 0 {
+
+		result := b.constraints(machine.Constraints())
+
+		// Get the machine, so we can get the hardware characteristics.
+		// This is done for parity with GUI.
```
wallyworld: No need to mention the GUI. It's just the logic we need to provide.

apiserver/facades/client/bundle/bundle.go
```diff
@@ -248,7 +248,44 @@ func (b *BundleAPI) fillBundleData(model description.Model) (*charm.BundleData,
...
+		if machineState, _ := b.backend.Machine(machine.Id()); machineState != nil {
```
wallyworld: Don't ignore the error. The check should be for err != nil rather than machineState != nil

apiserver/facades/client/bundle/bundle.go
```diff
@@ -248,7 +248,44 @@ func (b *BundleAPI) fillBundleData(model description.Model) (*charm.BundleData,
...
+			hardware, err := machineState.HardwareCharacteristics()
+			if err == nil {
+				result = result[:0]
```
wallyworld: This logic isn't right.
We need to combine the machine constraints and h/w info before splitting the machine constraints into a string.
There's no tests for this new work which would have shown the problem.

apiserver/facades/client/bundle/state.go
```diff
@@ -12,6 +12,7 @@ import (
 type Backend interface {
 	ExportPartial(cfg state.ExportConfig) (description.Model, error)
 	GetExportConfig() state.ExportConfig
+	Machine(id string) (*state.Machine, error)
```
wallyworld: The result needs to be a Machine interface not a state.Machine.
The interface has just the method on it we use in the code and which we then add to the mock

apiserver/facades/client/bundle/bundle.go
```diff
@@ -248,8 +248,13 @@ func (b *BundleAPI) fillBundleData(model description.Model) (*charm.BundleData,
 			Annotations: machine.Annotations(),
 			Series:      machine.Series(),
 		}
-		if result := b.constraints(machine.Constraints()); len(result) != 0 {
+
+		if result := b.hardwareConstraints(machine.Instance()); len(result) != 0 {
 			newMachine.Constraints = strings.Join(result, " ")
+		} else {
+			if result = b.machineConstraints(machine.Constraints()); len(result) != 0 {
```
wallyworld: We don't need this here right?
Since we set machine constraints earlier up above?

wallyworld: Oh wait, I got confused because of the method name. Because it's also used above to get the application constraints, maybe we should revert the name back to "constraints" and have the hardware constraints version called hardwareConstraints is is the case now.

apiserver/facades/client/bundle/state.go
```diff
@@ -32,6 +32,11 @@ func (m *stateShim) GetExportConfig() state.ExportConfig {
 	return cfg
 }
 
+// Machine returns machine with given id.
```
wallyworld: Remove me

apiserver/facades/client/bundle/bundle.go
vinu2003: :) ok.

vinu2003:
$$merge$$

PR #8987, opened 2018-07-27
openstack storage provider now working with CACerts

hmlanigan:
## Description of change

Make use of the cinder client change to allow for tls.Config data to be used when making api calls to an OpenStack requiring CACert as part of authentication.

## QA steps

Using an openstack deployment requiring a CACert for access:
1. Add the ca cert to the clouds.yaml for the new cloud:
clouds:
openstack:
type: openstack
.....
ca-certificates:
|
-----BEGIN CERTIFICATE-----
.....
-----END CERTIFICATE-----
1. juju bootstrap
2. use juju storage functionality.
3. juju destroy-controller should not fail with unable to list volumes.

There should be no change with an openstack deployment not requiring a CACert.

## Documentation changes

Not yet, one more item to be fixed.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1784030


hmlanigan:
Current unit tests of the OpenStack storage provider run against mock clients, so there is not an easy way to verify this change.  The goose libraries do not provide a test double for the cinder OpenStack service.  I can provide credentials and information for this to be tested by hand, against an OpenStack requiring a CACert.  Longer term, a CI test which can be run periodically against such a config would be good.

Testing this PR by hand is facilitated by incorporating https://github.com/juju/juju/pull/8984, but not required.

hmlanigan:
[cinder has own unique client, calling goosehttp.New()](https://trello.com/c/8LtgQ0H1/385-cinder-has-own-unique-client-calling-goosehttpnew)

manadart:
Manually tested with:
- Bootstrap
- `juju deploy postgresql --storage pgdata=cinder,4G,1`
- `juju storage` shows the volume.
- `openstack volume list` shows the volume.

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #8989, opened 2018-07-30
Azure reacts to invalid cloud credentials.

anastasiamac:
## Description of change

This PR ensures that juju azure provider is responsive to azure cloud un-happinness with provided credentials.

Both situations where an invalid credential and invalidated (say, expired) credential is used to communicate with azure cloud is catered for here. 

The most interesting bits are contained in ~/provider/azure/internal/errorutils package. This is the code that examines azure errors and determines whether the error is related to credential and worth reacting to.
All cloud calls that throw errors have now been modified to go through this logic to ensure that if the cloud complained about a credential, juju marks it as invalid.

Unit tests have been added to provide adequate initial coverage.

### DIFF
--- a/provider/azure/environ.go
+++ b/provider/azure/environ.go
@@ -210,7 +210,7 @@ func (env *azureEnviron) initEnviron() error {
 // PrepareForBootstrap is part of the Environ interface.
 func (env *azureEnviron) PrepareForBootstrap(ctx environs.BootstrapContext) error {
 	if ctx.ShouldVerifyCredentials() {
-		if err := verifyCredentials(env); err != nil {
+		if err := verifyCredentials(env, nil); err != nil {
 			return errors.Trace(err)
 		}
 	}
@@ -219,10 +219,10 @@ func (env *azureEnviron) PrepareForBootstrap(ctx environs.BootstrapContext) erro
 
 // Create is part of the Environ interface.
 func (env *azureEnviron) Create(ctx context.ProviderCallContext, args environs.CreateParams) error {
-	if err := verifyCredentials(env); err != nil {
+	if err := verifyCredentials(env, ctx); err != nil {
 		return errors.Trace(err)
 	}
-	return errors.Trace(env.initResourceGroup(args.ControllerUUID, false))
+	return errors.Trace(env.initResourceGroup(ctx, args.ControllerUUID, false))
 }
 
 // Bootstrap is part of the Environ interface.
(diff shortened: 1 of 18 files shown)


anastasiamac:
[adjust azure provider to use callback to invalidate cloud credential when needed](https://trello.com/c/CkG9MeB1/341-adjust-azure-provider-to-use-callback-to-invalidate-cloud-credential-when-needed)

provider/azure/environ.go
```diff
@@ -1232,6 +1241,7 @@ func (env *azureEnviron) deleteVirtualMachine(
 	logger.Debugf("- deleting virtual machine (%s)", vmName)
 	vmResultCh, errCh := vmClient.Delete(env.resourceGroup, vmName, nil)
 	if result, err := <-vmResultCh, <-errCh; err != nil {
+		errorutils.HandleCredentialError(err, ctx)
```
wallyworld: If there's a credential error here, any subsequent call to GetContainerReference below will also fail right? Did we need to do something like:
if maybeStorageClient && validCredential {

of something

provider/azure/environ_test.go
```diff
@@ -1143,6 +1189,33 @@ func (s *environSuite) TestBootstrap(c *gc.C) {
...
 }
 
+func (s *environSuite) TestBootstrapWithInvalidCredential(c *gc.C) {
+	defer envtesting.DisableFinishBootstrap()()
+
+	ctx := envtesting.BootstrapContext(c)
+	env := prepareForBootstrap(c, ctx, s.provider, &s.sender)
+
+	s.createSenderWithUnauthorisedStatusCode(c)
+	// s.sender = append(s.sender,s.initResourceGroupSenders())
```
wallyworld: remove me?

provider/azure/internal/errorutils/errors.go
```diff
@@ -4,11 +4,19 @@
...
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/juju/errors"
+	"github.com/juju/loggo"
+	"github.com/juju/utils/set"
+
+	"github.com/juju/juju/environs/context"
 )
 
+var logger = loggo.GetLogger("juju.provider.azure")
```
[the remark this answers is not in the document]
anastasiamac: I have thought about it but I am on the fence. I'd rather leave it as-is simply because our logging is not generally grouped by package but by functional area and here the functional area is azure provider. SO all the logging for azure provider is either desired or not :)

I have changed the logging of error to Warning although, I do not know what an operator can do with it once the failure occurs.

Greta pickup overall though, as I was not logging the actual failure :) Fixed now.

anastasiamac:
$$merge$$

PR #8996, opened 2018-08-01
Caas revert uneeded splitting and use existing functionality

Veebers:
## Description of change

This branch reverts some validating and splitting of metadata. It instead uses the docker libraries to supply the image path validation.

It also fixes a bug where image names containing numbers wouldn't be considered valid.

## QA steps

Updated the unit tests
Deployed the caas charm that exhibited the validation error.

### DIFF
--- a/caas/kubernetes/provider/dockerconfig.go
+++ b/caas/kubernetes/provider/dockerconfig.go
@@ -4,6 +4,9 @@
 package provider
 
 import (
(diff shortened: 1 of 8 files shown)


wallyworld:
Plus you'll need to update Juju's dependencies.tsv as well.

wallyworld:
$$merge$$

PR #8999, opened 2018-08-01
GCE reacts to invalid cloud credentials.

anastasiamac:
## Description of change

This PR ensures that juju gce provider is responsive to google cloud un-happinness with provided credentials.

The most interesting bits are contained in ~/provider/gce/google/errors.go file. This is the code that examines google errors and determines whether the error is related to credential and worth reacting to.
All cloud calls that throw errors have now been modified to go through this logic to ensure that if the cloud complained about a credential, juju marks it as invalid.

Unit tests have been added to provide adequate initial coverage.

### DIFF
--- a/provider/gce/disks.go
+++ b/provider/gce/disks.go
@@ -88,7 +88,7 @@ func (g *storageProvider) VolumeSource(cfg *storage.Config) (storage.VolumeSourc
 
 type instanceCache map[string]google.Instance
 
-func (c instanceCache) update(gceClient gceConnection, ids ...string) error {
+func (c instanceCache) update(gceClient gceConnection, ctx context.ProviderCallContext, ids ...string) error {
 	if len(ids) == 1 {
 		if _, ok := c[ids[0]]; ok {
 			return nil
@@ -100,7 +100,7 @@ func (c instanceCache) update(gceClient gceConnection, ids ...string) error {
 	}
 	instances, err := gceClient.Instances("", google.StatusRunning)
 	if err != nil {
-		return errors.Annotate(err, "querying instance details")
+		return google.HandleCredentialError(errors.Annotate(err, "querying instance details"), ctx)
 	}
 	for _, instance := range instances {
 		if _, ok := idMap[instance.ID]; !ok {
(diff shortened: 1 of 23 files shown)


anastasiamac:
[adjust gce provider to use callback to invalidate cloud credential when needed](https://trello.com/c/qR6MVUOh/349-adjust-gce-provider-to-use-callback-to-invalidate-cloud-credential-when-needed)

provider/gce/google/errors.go
```diff
@@ -55,3 +60,57 @@ func NewMissingConfigValue(key, field string) error {
...
+				return true
+			}
+		}
+	}
+	return false
+
+}
+
+// AuthorisationFailureStatusCodes contains http status code nad description that signify authorisation difficulties.
+var AuthorisationFailureStatusCodes = map[int]string{
```
wallyworld: I wonder if this is useful or needed in other providers? It seems quite generic.

anastasiamac: Not really generic at all...

For one, for example, azure only has the codes not the description. Also the code for bad request is only applicable to oauth2-compliant clients. So for example, azure does not send it for authentication issues but gce does :)

So, at this stage, fairly provider specific. If I'll encounter duplication in other providers, will pull it into provider/common area or something...

provider/gce/disks.go
```diff
@@ -462,27 +477,32 @@ func (v *volumeSource) attachOneVolume(volumeName string, mode google.DiskMode,
...
 	}
 	return attachment, nil
 }
 
 func (v *volumeSource) DetachVolumes(ctx context.ProviderCallContext, attachParams []storage.VolumeAttachmentParams) ([]error, error) {
 	result := make([]error, len(attachParams))
 	for i, volumeAttachment := range attachParams {
-		result[i] = v.detachOneVolume(volumeAttachment)
+		err := v.detachOneVolume(ctx, volumeAttachment)
+		if err != nil && google.HasDenialStatusCode(err) {
```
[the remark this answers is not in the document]
anastasiamac: True :) and it does \o/ Thank you.

anastasiamac:
$$merge$$

PR #9002, opened 2018-08-02
leases: add raft-FSM lease store

babbageclunk:
## Description of change

Implements a new core/lease.Store as a raft FSM. Instead of running a lease manager for singular and leadership inside each state in the controller agent's state pool, we run one lease manager for the controller agent in the dependency engine. The manager now passes namespace and model uuid along with the lease name - together these are the lease key.

That manager talks to a lease store that reads data directly from the FSM, but all writes (claim, extend, expire) are sent over the pubsub hub to the raft leader to be applied there. We also run another global clock updater to increment the time in the raft lease store (using the same pubsub request mechanism) alongside the clock updater talking to mongo.

The raft lease store still updates the mongo database (the leaseholders collection, to avoid stepping on the toes of the state lease store), but only when leases are claimed or expired, which occur far less frequently than extensions (every 30 seconds).

Lease changes (claims and expiries) are also written to `/var/log/juju/lease.log`. In a multi-controller situation these will need to be combined to get a complete history of lease changes. These logs are rotated and compressed.

The two lease systems (mongo and raft) are both running - which one is in use is controlled by the `raft-leases` feature flag. This means it _can_ be switched on a running controller, although since the lease state isn't synchronised between them there's likely to be some fallout - it's possible for two units to both think they're the leader, for example. I'd advise not switching it on a controller you wouldn't be ok with throwing away for now.

## QA steps

* Bootstrap a controller with `--config="features=[raft-leases]"` and including `juju.worker.lease=TRACE` in `logging-config`.
* Enable HA.
* Deploy percona-cluster and 3 units of wordpress.
* See that one wordpress unit becomes the leader for the application (in status, in logging for `juju.worker.lease.raft` and in the mongo `leaseholders` collection).
* Remove that unit. Another wordpress unit should claim leadership when the original lease expires.
* Use `juju run -m controller <machine> juju-engine-report | less +/raft:` to determine the raft leader. Remove that machine. Once things stabilise, confirm that the lease operations are continuing in a two-controller voter/nonvoter configuration. 
* Add a new controller machine by running `enable-ha` again. The lease manager on that machine should join in with the normal lease operations.

### DIFF
--- a/apiserver/facade/facadetest/context.go
+++ b/apiserver/facade/facadetest/context.go
@@ -5,6 +5,8 @@ package facadetest
 
 import (
 	"github.com/juju/juju/apiserver/facade"
+	"github.com/juju/juju/core/leadership"
+	"github.com/juju/juju/core/lease"
 	"github.com/juju/juju/state"
 )
 
@@ -17,6 +19,10 @@ type Context struct {
 	State_     *state.State
 	StatePool_ *state.StatePool
 	ID_        string
+
+	LeadershipClaimer_ leadership.Claimer
+	LeadershipChecker_ leadership.Checker
+	SingularClaimer_   lease.Claimer
 	// Identity is not part of the facade.Context interface, but is instead
 	// used to make sure that the context objects are the same.
 	Identity string
@@ -66,4 +72,19 @@ func (context Context) Presence() facade.Presence {
 func (context Context) ModelPresence(modelUUID string) facade.ModelPresence {
 	// Potentially may need to add stuff here at some stage.
 	return nil
+}
+
+// LeadershipClaimer implements facade.Context.
+func (context Context) LeadershipClaimer() (leadership.Claimer, error) {
+	return context.LeadershipClaimer_, nil
+}
+
+// LeadershipChecker implements facade.Context.
+func (context Context) LeadershipChecker() (leadership.Checker, error) {
+	return context.LeadershipChecker_, nil
+}
+
+// SingularClaimer implements facade.Context.
+func (context Context) SingularClaimer() (lease.Claimer, error) {
+	return context.SingularClaimer_, nil
 }
--- a/apiserver/facade/interface.go
+++ b/apiserver/facade/interface.go
@@ -6,6 +6,8 @@ package facade
 import (
 	"gopkg.in/juju/names.v2"
 
+	"github.com/juju/juju/core/leadership"
+	"github.com/juju/juju/core/lease"
 	"github.com/juju/juju/core/presence"
 	"github.com/juju/juju/permission"
 	"github.com/juju/juju/state"
@@ -76,6 +78,15 @@ type Context interface {
 	// into Resources to get the watcher in play. This is not really
 	// a good idea; see Resources.
 	ID() string
+
+	// LeadershipClaimer returns a leadership.Claimer tied to the specific model for this context's model.
+	LeadershipClaimer() (leadership.Claimer, error)
+
+	// LeadershipChecker returns a leadership.Checker for this context's model.
+	LeadershipChecker() (leadership.Checker, error)
+
+	// SingularClaimer returns a lease.Claimer for singular leases for this context's model.
+	SingularClaimer() (lease.Claimer, error)
 }
 
 // Authorizer represents the authenticated entity using the API server.
--- a/apiserver/facades/agent/uniter/uniter.go
+++ b/apiserver/facades/agent/uniter/uniter.go
@@ -51,6 +51,7 @@ type UniterAPI struct {
 	st                *state.State
 	auth              facade.Authorizer
 	resources         facade.Resources
+	leadershipChecker leadership.Checker
 	accessUnit        common.GetAuthFunc
 	accessApplication common.GetAuthFunc
 	accessMachine     common.GetAuthFunc
@@ -88,9 +89,16 @@ type UniterAPIV4 struct {
 }
 
 // NewUniterAPI creates a new instance of the core Uniter API.
-func NewUniterAPI(st *state.State, resources facade.Resources, authorizer facade.Authorizer) (*UniterAPI, error) {
+func NewUniterAPI(context facade.Context) (*UniterAPI, error) {
+	authorizer := context.Auth()
 	if !authorizer.AuthUnitAgent() && !authorizer.AuthApplicationAgent() {
 		return nil, common.ErrPerm
+	}
+	st := context.State()
+	resources := context.Resources()
+	leadershipChecker, err := context.LeadershipChecker()
+	if err != nil {
+		return nil, errors.Trace(err)
 	}
 	accessUnit := func() (common.AuthFunc, error) {
 		switch tag := authorizer.GetAuthTag().(type) {
@@ -225,7 +233,7 @@ func NewUniterAPI(st *state.State, resources facade.Resources, authorizer facade
 		ModelWatcher:               common.NewModelWatcher(m, resources, authorizer),
 		RebootRequester:            common.NewRebootRequester(st, accessMachine),
 		UpgradeSeriesAPI:           common.NewExternalUpgradeSeriesAPI(st, resources, authorizer, accessMachine, accessUnit, logger),
-		LeadershipSettingsAccessor: leadershipSettingsAccessorFactory(st, resources, authorizer),
+		LeadershipSettingsAccessor: leadershipSettingsAccessorFactory(st, leadershipChecker, resources, authorizer),
 		MeterStatus:                msAPI,
 		// TODO(fwereade): so *every* unit should be allowed to get/set its
 		// own status *and* its application's? This is not a pleasing arrangement.
@@ -235,6 +243,7 @@ func NewUniterAPI(st *state.State, resources facade.Resources, authorizer facade
 		m:                 m,
 		auth:              authorizer,
 		resources:         resources,
+		leadershipChecker: leadershipChecker,
 		accessUnit:        accessUnit,
 		accessApplication: accessApplication,
 		accessMachine:     accessMachine,
@@ -245,8 +254,8 @@ func NewUniterAPI(st *state.State, resources facade.Resources, authorizer facade
 }
 
 // NewUniterAPIV7 creates an instance of the V7 uniter API.
-func NewUniterAPIV7(st *state.State, resources facade.Resources, authorizer facade.Authorizer) (*UniterAPIV7, error) {
-	uniterAPI, err := NewUniterAPI(st, resources, authorizer)
+func NewUniterAPIV7(context facade.Context) (*UniterAPIV7, error) {
+	uniterAPI, err := NewUniterAPI(context)
 	if err != nil {
 		return nil, err
 	}
@@ -256,8 +265,8 @@ func NewUniterAPIV7(st *state.State, resources facade.Resources, authorizer faca
 }
 
 // NewUniterAPIV6 creates an instance of the V6 uniter API.
-func NewUniterAPIV6(st *state.State, resources facade.Resources, authorizer facade.Authorizer) (*UniterAPIV6, error) {
-	uniterAPI, err := NewUniterAPIV7(st, resources, authorizer)
+func NewUniterAPIV6(context facade.Context) (*UniterAPIV6, error) {
+	uniterAPI, err := NewUniterAPIV7(context)
 	if err != nil {
 		return nil, err
 	}
@@ -267,8 +276,8 @@ func NewUniterAPIV6(st *state.State, resources facade.Resources, authorizer faca
 }
 
 // NewUniterAPIV5 creates an instance of the V5 uniter API.
-func NewUniterAPIV5(st *state.State, resources facade.Resources, authorizer facade.Authorizer) (*UniterAPIV5, error) {
-	uniterAPI, err := NewUniterAPIV6(st, resources, authorizer)
+func NewUniterAPIV5(context facade.Context) (*UniterAPIV5, error) {
+	uniterAPI, err := NewUniterAPIV6(context)
 	if err != nil {
 		return nil, err
 	}
@@ -278,8 +287,8 @@ func NewUniterAPIV5(st *state.State, resources facade.Resources, authorizer faca
 }
 
 // NewUniterAPIV4 creates an instance of the V4 uniter API.
-func NewUniterAPIV4(st *state.State, resources facade.Resources, authorizer facade.Authorizer) (*UniterAPIV4, error) {
-	uniterAPI, err := NewUniterAPIV5(st, resources, authorizer)
+func NewUniterAPIV4(context facade.Context) (*UniterAPIV4, error) {
+	uniterAPI, err := NewUniterAPIV5(context)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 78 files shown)


babbageclunk:
!!build!!

babbageclunk:
!!build!!

core/raftlease/store.go
```diff
@@ -0,0 +1,261 @@
...
+	if err == globalclock.ErrConcurrentUpdate {
+		// Someone else updated before us - get the new time.
+		s.prevTime = s.fsm.GlobalTime()
+	} else if err == nil {
+		s.prevTime = newTime
+	}
+	return errors.Trace(err)
+}
+
+func (s *Store) forwardCommand(command *Command) error {
```
wallyworld: Maybe forwardCommandReturningResponseError?
Or something?
forwardCommand implies that the cmd is sent on and that's it, and any error returned is the result of that on-sending failing. Whereas we are sending on the command and waiting for a response back and if all that happens returning that error. Did we need to separate response error from marshall or send error? I guess not since we check for the error type and can deduce from that if it's a semantic lease related error

core/raftlease/store.go
```diff
@@ -0,0 +1,261 @@
...
+	return &ResponseError{
+		Message: message,
+		Code:    code,
+	}
+}
+
+// RecoverError converts a ResponseError back into the specific error
+// it represents, or into a generic error if it wasn't one of the
+// singleton errors handled.
+func RecoverError(resp *ResponseError) error {
```
wallyworld: does this need to be exported?
(same to a bunch of other things in here)

feature/flags.go
```diff
@@ -51,3 +51,7 @@ const DisableRaft = "disable-raft"
 
 // UpgradeSeries is a development feature flag.
 const UpgradeSeries = "upgrade-series"
+
+// RaftLeases will switch all lease management to be handled by the
+// raft lease FSM, rather than being done in Mongo.
+const RaftLeases = "raft-leases"
```
wallyworld: This is on by default in 2.5 right?

core/raftlease/store_test.go
```diff
@@ -0,0 +1,469 @@
...
+					lease.Key{"warframe", "vauban", "prime"},
+					lease.Request{"vor", time.Second},
+				)
+			}()
+			// Jump time forward further than the 1-second forward
+			// timeout.
+			s.clock.WaitAdvance(2*time.Second, coretesting.LongWait, 1)
+			select {
+			case err := <-errChan:
+				c.Assert(err, jc.Satisfies, lease.IsTimeout)
```
wallyworld: We need a bool to prove we got here
(similar comment for other tests)

state/migration_internal_test.go
```diff
@@ -204,6 +204,11 @@ func (s *MigrationSuite) TestKnownCollections(c *gc.C) {
 		relationNetworksC,
 		firewallRulesC,
 		dockerResourcesC,
+		// Raft leases - TODO
```
wallyworld: // TODO(raft-leases)

(or whatever we use for consistency)

state/state.go
```diff
@@ -419,6 +423,18 @@ func (st *State) start(controllerTag names.ControllerTag, hub *pubsub.SimpleHub)
 // ApplicationLeaders returns a map of the application name to the
 // unit name that is the current leader.
 func (st *State) ApplicationLeaders() (map[string]string, error) {
+	config, err := st.ControllerConfig()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if config.Features().Contains(feature.RaftLeases) {
```
wallyworld: raft by default in 2.5

worker/globalclockupdater/manifold.go
```diff
@@ -11,26 +11,32 @@ import (
...
+	Logger         Logger
 }
 
 func (config ManifoldConfig) Validate() error {
 	if config.ClockName == "" {
 		return errors.NotValidf("empty ClockName")
 	}
-	if config.StateName == "" {
-		return errors.NotValidf("empty StateName")
+	if config.StateName == "" && config.LeaseManagerName == "" {
```
wallyworld: So why not require both and have the one worker update both at the same time?

worker/lease/secretaries.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/juju/juju/core/lease"
+)
+
+// SingularSecretary implements Secretary to restrict claims to either
+// a lease for the controller or a specific model, holdable only by machine-tag
+// strings.
+//
+// It would be nicer to have a single controller-level component managing all
+// singular leases for all models -- and thus be able to validate that
+// proposed holders really are env managers -- but the complexity of threading
```
wallyworld: env managers? that's so juju 1.x

worker/raft/raftforwarder/worker.go
```diff
@@ -0,0 +1,141 @@
...
+func (w *forwarder) handleRequest(_ string, req raftlease.ForwardRequest, err error) {
+	w.config.Logger.Tracef("received %#v, err: %s", req, err)
+	if err != nil {
+		// This should never happen, so treat it as fatal.
+		w.catacomb.Kill(errors.Annotate(err, "requests callback failed"))
+		return
+	}
+	response, err := w.processRequest(req.Command)
+	if err != nil {
+		w.catacomb.Kill(err)
```
wallyworld: annotate here too?

core/raftlease/store.go
```diff
@@ -0,0 +1,261 @@
...
+		Namespace: key.Namespace,
+		ModelUUID: key.ModelUUID,
+		Lease:     key.Lease,
+		Holder:    req.Holder,
+		Duration:  req.Duration,
+	})
+	if err == nil {
+		s.config.Target.Claimed(key, req.Holder)
+	}
+	return errors.Trace(err)
```
wallyworld: I wonder if timeout should be handled differently?
What if system is heavily loaded and 5 seconds isn't enough? Should there be a retry or backoff and retry?

apiserver/facade/facadetest/context.go
```diff
@@ -67,3 +73,18 @@ func (context Context) ModelPresence(modelUUID string) facade.ModelPresence {
 	// Potentially may need to add stuff here at some stage.
 	return nil
 }
+
```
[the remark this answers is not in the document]
babbageclunk: It's accurate though, isn't it? The ModelPresence method is on a different interface, facade.Presence.

apiserver/facade/interface.go
```diff
@@ -76,6 +78,15 @@ type Context interface {
 	// into Resources to get the watcher in play. This is not really
 	// a good idea; see Resources.
 	ID() string
+
+	// LeadershipClaimer returns a leadership.Claimer tied to the specific model for this context.
```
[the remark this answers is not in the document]
babbageclunk: Done.

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -1538,7 +1547,7 @@ func (u *UniterAPI) SetRelationStatus(args params.RelationStatusArgs) (params.Er
 		return unit, nil
 	}
 
-	checker := u.st.LeadershipChecker()
+	checker := u.checker
```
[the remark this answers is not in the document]
babbageclunk: Yeah, that's a good call.

apiserver/facades/controller/singular/singular.go
```diff
@@ -53,22 +58,19 @@ type Backend interface {
...
+func NewFacade(backend Backend, claimer lease.Claimer, auth facade.Authorizer) (*Facade, error) {
 	if !auth.AuthController() {
 		return nil, common.ErrPerm
 	}
 	return &Facade{
 		auth:          auth,
 		modelTag:      backend.ModelTag(),
 		controllerTag: backend.ControllerTag(),
-		claimer:       backend.SingularClaimer(),
+		claimer:       claimer,
```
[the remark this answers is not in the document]
babbageclunk: Agreed on changing it to singular claimer. 
I'm not going to change the facade attribute to LeaseClaimer though - singular and leadership are both applications of leases, so I think it's clearer this way.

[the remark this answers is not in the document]
babbageclunk: Actually, on making this change it seems a bit department of redundancy department. It's the claimer on the singular.Facade, so of course it's a singularClaimer. The uniter one made more sense since the uniter facade does so much other stuff. Ah well, I've done it now.

apiserver/root.go
```diff
@@ -433,6 +435,47 @@ func (ctx *facadeContext) ID() string {
 	return ctx.key.objId
 }
 
+// LeadershipClaimer is part of the facade.Context interface.
+func (ctx *facadeContext) LeadershipClaimer() (leadership.Claimer, error) {
+	if ctx.r.shared.featureEnabled(feature.RaftLeases) {
```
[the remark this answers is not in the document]
babbageclunk: Yup, thanks!

apiserver/root.go
```diff
@@ -542,3 +585,59 @@ func DescribeFacades(registry *facade.Registry) []params.FacadeVersions {
 	}
 	return result
 }
+
+// TODO(raftlease): don't just copy this from state.
```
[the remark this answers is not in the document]
babbageclunk: Done.

cmd/jujud/agent/machine/manifolds.go
```diff
@@ -488,6 +497,17 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 			NewWorker:      globalclockupdater.NewWorker,
 			UpdateInterval: globalClockUpdaterUpdateInterval,
 			BackoffDelay:   globalClockUpdaterBackoffDelay,
+			Logger:         loggo.GetLogger("juju.worker.globalclockupdater.mongo"),
+		}),
+		// We also run another clock updater to feed time updates into
+		// the lease FSM.
+		leaseClockUpdaterName: globalclockupdater.Manifold(globalclockupdater.ManifoldConfig{
```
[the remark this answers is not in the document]
babbageclunk: I started doing that initially but then moved to doing this instead. My thinking was: we want the updates to each clock to be independent (can't have one of them being slow preventing updates to the other one), so that means that they would need to run as separate goroutines inside the worker each talking to their own updater. At that point it seemed cleaner just to have two instances of the worker, each talking to a specific updater.

core/raftlease/fsm.go
```diff
@@ -0,0 +1,411 @@
...
+	mu         sync.Mutex
+	globalTime time.Time
+	entries    map[lease.Key]*entry
+}
+
+func (f *FSM) claim(key lease.Key, holder string, duration time.Duration) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	if _, found := f.entries[key]; found {
+		return lease.ErrInvalid
```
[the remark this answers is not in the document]
babbageclunk: The only (expected/handled) errors that come out of the store are lease.ErrInvalid or lease.ErrTimeout (which is new for raft - the state-based store only returns ErrInvalid). LeaseNotHeld is a higher-level error that the lease manager emits back to client code based on its view of the lease state. 

I could change this, but this behaviour matches what the state store does and the lease manager expects.

core/raftlease/fsm.go
```diff
@@ -0,0 +1,411 @@
...
+	return nil
+}
+
+func (f *FSM) setTime(oldTime, newTime time.Time) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	if f.globalTime != oldTime {
+		return globalclock.ErrConcurrentUpdate
+	}
+	f.globalTime = newTime
```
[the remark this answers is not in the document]
babbageclunk: Leases are stored as start time (when they were claimed/extended) + duration. (They could have been stored as just expiry times, but I think it was done like this to make the translation between global time and local time clearer - I copied this from the state lease implementation.) We could update all of the starts and durations, adding a second to the start and subtracting one from duration, whenever time moves forward, but there's no need to. In a non-distributed system where we were just using the wall clock for time (instead of a replicated time) we wouldn't update every lease each tick.

One related thing I did think about though, is that we could automatically remove leases that have expired here. The state-based lease store doesn't do this because it would require either triggers in mongo, or for the clock updater to know too much about the lease collection. But in the fsm they're right there, so it would be simple and potentially cut out a lot of raft traffic. I didn't want to make that change just yet because it's a bit subtle, but I think the lease manager would just handle it - it already needs to cope with a different manager expiring a lease out from under it.

core/raftlease/fsm.go
```diff
@@ -0,0 +1,411 @@
...
+	defer f.mu.Unlock()
+
+	f.globalTime = snapshot.GlobalTime
+	f.entries = newEntries
+
+	return nil
+}
+
+// Snapshot defines the format of the FSM snapshot.
+type Snapshot struct {
```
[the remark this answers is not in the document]
babbageclunk: It was more convenient for testing, and I have a feeling we're going to need to use the snapshot type for upgrade steps.

core/raftlease/fsm.go
```diff
@@ -0,0 +1,411 @@
...
+	}()
+
+	encoder := yaml.NewEncoder(sink)
+	if err := encoder.Encode(s); err != nil {
+		return errors.Trace(err)
+	}
+	if err := encoder.Close(); err != nil {
+		return errors.Trace(err)
+	}
+	return sink.Close()
```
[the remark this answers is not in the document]
babbageclunk: This method gets called by the raft infrastructure - the docs say that the FSM code should call sink.Close when it's finished with it. We definitely have the same kind of pattern in our code where we need it - otherwise no-one would ever take a WriteCloser, right? :)

core/raftlease/store.go
babbageclunk: You're right - forward is the wrong implication here. How about `runOnLeader`?

I don't think there's any point in splitting out a marshalling/transport error return from lease-specific errors here, because code that directly calls `runOnLeader` doesn't distinguish between those errors - the lease manager does but that's a layer away.

core/raftlease/store.go
babbageclunk: ForwardRequest, ForwardResponse, ResponseError and AsResponseError need to be exported so they can be used from the raftforwarder. RecoverError maybe doesn't? But it seems weird to export a conversion function in one direction and not the reverse. Not sure.

feature/flags.go
babbageclunk: It wasn't, but after we talked about it I'll change it to be on by default. In that case I should really flip the sense of the flag to be "disable-raft-leases"

core/raftlease/store_test.go
babbageclunk: I don't understand this - doesn't the channel write show that we got here? This select can only succeed if we get a timeout error on the channel, otherwise it will fatal, right?

state/migration_internal_test.go
babbageclunk: Good spot!

worker/globalclockupdater/manifold.go
babbageclunk: Covered above, I think.

worker/lease/secretaries.go
babbageclunk: Yeah, this code was mostly just move wholesale from state. I'll update the comment.

worker/raft/raftforwarder/worker.go
babbageclunk: Yeah, nice.

core/raftlease/store.go
babbageclunk: The lease manager does retry with exponential backoff. I'm not totally sure about the numbers around this - I tried to do some experiments to get a feel for what they should be but it was pretty ad-hoc.

feature/flags.go
babbageclunk: Better suggestion (from @wallyworld): `use-legacy-leases`.

babbageclunk:
$$merge$$

babbageclunk:
Intermittent failure in clientMacaroonSuite.TestAddLocalCharmSuccess
https://bugs.launchpad.net/juju/+bug/1788401

$$merge$$

PR #9005, opened 2018-08-02
When finding the unit's index check status type.

ExternalReality:
Post series upgrade hook was checking the wrong status type when trying to update the unit by index.

### DIFF
--- a/worker/uniter/resolver.go
+++ b/worker/uniter/resolver.go
@@ -58,7 +58,8 @@ func (s *uniterResolver) NextOp(
 	// waiting to be shutdown.
 	if localState.UpgradeSeriesPrepareStatus == model.UnitCompleted &&
 		remoteState.UpgradeSeriesPrepareStatus == model.UnitCompleted {
-		return nil, resolver.ErrNoOperation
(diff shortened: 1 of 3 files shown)


ExternalReality:
[Fix issue with status type checks when finding unit index.](https://trello.com/c/PXGaCpfl/428-fix-issue-with-status-type-checks-when-finding-unit-index)

worker/uniter/resolver.go
```diff
@@ -58,7 +58,8 @@ func (s *uniterResolver) NextOp(
 	// waiting to be shutdown.
 	if localState.UpgradeSeriesPrepareStatus == model.UnitCompleted &&
 		remoteState.UpgradeSeriesPrepareStatus == model.UnitCompleted {
-		return nil, resolver.ErrNoOperation
+		logger.Criticalf("We hit here were we are supposed to idle when shutting down.")
```
manadart: Did you mean to leave in this test logging?

ExternalReality: Yes, this logging should remain until the machine agent is ready to shut down the units. This messages should replaced with the code that causes the Uniter to idle, that is, returning a the no-op.

PR #9008, opened 2018-08-03
Bring in recent websocket with handshake timeout.

howbazaar:
There was a potential race condition on the agents when connecting to a controller if interrupted in the middle of the websocket protocol initiation handshake.

This is handled in a more recent version of the websocket library, so bring that in.

45 seconds is the default timeout from the upstream library, but we aren't using their default dialer.

## QA steps

I have not been successful in replicating the error conditions that we only see in very large deployments.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1732587


howbazaar:
$$merge$$

PR #9015, opened 2018-08-03
update openstack cloud schema to include os_cacert

hmlanigan:
## Description of change

Update add-cloud to allow default values of the cloud schema's to get
values from environment variables.  Add environment variables and a new
get cloud ca certificate to the openstack cloud schema.  Add ca-certificates
to cloud schema use for validation.

## QA steps

Run juju add-cloud for openstack.  Should query for an optional cloud ca cert.
1. try without a cert
2. try with a good cert file
3. try with a bad cert file - should create the cloud but Ignore the cert.

Source a novarc file, including a OS_CACERT definition, run juju add-cloud for openstack.
1. notice that the auth url is now a default option.
2. notice that the os cacert is now a default option.

Check that the new cloud definition in ~/.local/share/juju/clouds.yaml contains  ca-certificates: ...

Try juju add-cloud -f with a yaml definition including the ca-certificates, no schema validation error should be printed on the name.

## Documentation changes

Not yet, a discourse update will happen with all of the OS_CACERT changes are made.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1777897
Partially addresses https://bugs.launchpad.net/juju/+bug/1722580

### DIFF
--- a/cmd/juju/cloud/add.go
+++ b/cmd/juju/cloud/add.go
@@ -20,6 +20,8 @@ import (
 	"github.com/juju/juju/cmd/juju/interact"
 	"github.com/juju/juju/environs"
 	"github.com/juju/juju/environs/context"
+	"github.com/juju/utils"
+	"github.com/juju/utils/cert"
 )
 
 type CloudMetadataStore interface {
@@ -203,10 +205,26 @@ func (c *AddCloudCommand) runInteractive(ctxt *cmd.Context) error {
 		ctxt.Infof("Cloud credential is not accepted by cloud provider: %v", reason)
 		return nil
 	}
-	pollster.VerifyURLs = func(s string) (ok bool, msg string, err error) {
-		err = c.Ping(provider, s)
+
+	// VerifyURLs will return true if a schema format type jsonschema.FormatURI is used
+	// and the value will Ping().
+	pollster.VerifyURLs = func(s string) (bool, string, error) {
+		err := c.Ping(provider, s)
 		if err != nil {
 			return false, "Can't validate endpoint: " + err.Error(), nil
+		}
+		return true, "", nil
+	}
+
+	// VerifyCertFile will return true if the schema format type "cert-filename" is used
+	// and the value is readable and a valid cert file.
+	pollster.VerifyCertFile = func(s string) (bool, string, error) {
+		out, err := ioutil.ReadFile(s)
+		if err != nil {
+			return false, "Can't validate CA Certificate file: " + err.Error(), nil
+		}
+		if _, err := cert.ParseCert(string(out)); err != nil {
+			return false, fmt.Sprintf("Can't validate CA Certificate %s: %s", s, err.Error()), nil
 		}
 		return true, "", nil
 	}
@@ -219,6 +237,17 @@ func (c *AddCloudCommand) runInteractive(ctxt *cmd.Context) error {
 	if err != nil {
 		return errors.Trace(err)
 	}
+
+	filename, alt, err := addCertificate(b)
+	switch {
+	case errors.IsNotFound(err):
+	case err != nil:
+		return errors.Annotate(err, "CA Certificate")
+	default:
+		ctxt.Infof("Successfully read CA Certificate from %s", filename)
+		b = alt
+	}
+
 	newCloud, err := c.cloudMetadataStore.ParseOneCloud(b)
 	if err != nil {
 		return errors.Trace(err)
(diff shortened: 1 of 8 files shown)


hmlanigan:
[cloud ca-cert via interactive juju add-cloud](https://trello.com/c/XBcblQaG/414-cloud-ca-cert-via-interactive-juju-add-cloud)

hmlanigan:
!!build!!

mitechie:
Interactive didn't work for me. It output the optional cacert but went stright into the Auth types?

https://pastebin.canonical.com/p/kBwRQv3bDk/

I never hit enter or anything to deal wit hthe filename and I'd expect the authtypes to have a newline after it to line up with other questions.

cmd/juju/cloud/add_test.go
```diff
@@ -757,3 +706,260 @@ func (s *addSuite) TestInvalidCredentialMessage(c *gc.C) {
...
+		"endpoint: http://myopenstack\n" +
+		"regions:\n" +
+		"  regionone:\n" +
+		"    endpoint: http://boston/1.0\n"
+
+	var input = "" +
+		/* Select cloud type: */ "openstack\n" +
+		/* Enter a name for your openstack cloud: */ "os1\n" +
+		/* Enter the API endpoint url for the cloud []: */ "http://myopenstack\n" +
+		/* Enter the filename of the CA certificate to access OpenStack cloud (optional) [] */ "\n" +
```
mitechie: I know these are the tests but wording suggestion

    If your cloud requires a CA certificate please provide the path to the certificate (optional) [none]:

mitechie:
I also note when interactive adding an openstack cloud you get asked for the "API Endpoint url" for the cloud but then you get asked about the region and the region url. Are these different urls? 

In checking our wiki on internal openstack the config is a single url for the region?

mitechie:
Interesting, when I ran juju add-cloud without giving it a name I got the above pastebin. When I ran 

    juju add-cloud testingstack3

it did pause and ask me about the CA cert properly.

mitechie:
ok, I do see the region url defaults to "[use cloud api url]" can this actually be the value vs the generic text?

mitechie:
We should fail if the cert won't load right away as if we typed in a wrong string for the auth type or a ping that failed

https://pastebin.canonical.com/p/rwZQDcM5pX/

mitechie:
I guess I can see the delayed response on the invalid cert as we're not reading it yet...if we can read it right away that'd be great. If not, shoudn't it be an error vs a warning and we should not add the cloud anyway?

https://pastebin.canonical.com/p/rs3P4gnSHx/

mitechie:
Awesome when you have the sourced file how the API url is prefilled with the real value like that. That's great.

cmd/juju/cloud/add_test.go
hmlanigan: There isn't a good way to get rid of a space at the beginning:
"Enter , if required, the path to a CA"... 

Sorry I wasn't more specific, but "Enter " gets tacked to the beginning on the string I provide..

Does "Enter if required, the path to a CA" work for you?

mitechie: no, I don't think that's proper punctuation there. #$@$R#@ this code is here just to irritate me I'm beginning to think.

mitechie: Enter a path to the CA certificate for your cloud if one is required to access it. (optional) [none]

I hate that I can't think of a better way to get the optional-ity of the question towards the front of the line given the stupid constraints of this code.

hmlanigan:
Side effect of the environment variables is that they aren't verified if the FormatURI as the value would be without it.   Risk should be minimal, as with the OpenStack case, the values have been verified by use when interacting with their OpenStack cloud.

cmd/juju/cloud/add_test.go
hmlanigan: updated.

hmlanigan:
Changed the default of Y/n to N/y when asking "Enter another... "
Changed to validate the certificate file when filename provided.

Comment on "[use cloud api url]" outside of scope for this PR.  Please file a bug.
We also need a bug to strip extra newlines and spaces from the input during add-cloud interactive.  Specifying "openstack   " for cloud type fails too.

cmd/juju/cloud/add_test.go
```diff
@@ -65,6 +65,10 @@ func (f *fakeCloudMetadataStore) WritePersonalCloudMetadata(cloudsMap map[string
 
 func (f *fakeCloudMetadataStore) ParseOneCloud(data []byte) (cloudfile.Cloud, error) {
 	results := f.MethodCall(f, "ParseOneCloud", data)
+	if len(results) != 2 {
```
hmlanigan: This was annoying me during testing, if the data doesn't match, the test panics.  Put in for debugging, since the printed []byte is not helpful to reading strings to find what's wrong.

cmd/juju/cloud/add.go
```diff
@@ -235,6 +264,45 @@ func (c *AddCloudCommand) runInteractive(ctxt *cmd.Context) error {
...
 
+// addCertificate reads the cloud certificate file if available and adds the contents
+// to the byte slice with the appropriate key.
+func addCertificate(data []byte) (string, []byte, error) {
+	vals, err := ensureStringMaps(string(data))
+	if err != nil {
+		return "", nil, err
+	}
+	var certificate string
+	filename, ok := vals[cloud.CertFilenameKey].(string)
```
SimonRichardson: This will panic if the `cloud.CertFilenameKey` is not there, because you're using the type assertion as well. If we're 100% sure that this will be fine, ignore my comment.

hmlanigan: Updating:  I can't guarantee the key in the map, but I can guarantee if the key in the map, the value is a string.

hmlanigan:
$$merge$$

PR #9019, opened 2018-08-05
leases: Prevent concurrent updates of the entries map in state/lease.store

babbageclunk:
## Description of change

This fixes a data race caused by the lease manager calling store methods
asynchronously now.

Example race test failure: http://10.125.0.203:8080/job/RunUnittests-race-amd64/584/testReport/github/com_juju_juju_worker_uniter_runner_context/TestPackage/

## QA steps

* Run failing tests under the race checker.

### DIFF
--- a/state/lease/store.go
+++ b/state/lease/store.go
@@ -5,6 +5,7 @@ package lease
 
 import (
 	"fmt"
+	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


babbageclunk:
$$merge$$

PR #9022, opened 2018-08-06
Prevent false negative with go vet

SimonRichardson:
## Description of change

The following error message has been updated to read correctly when
it is outputted to a user at some point (logs/CLI).

NotValidf will insert a " not valid" as the suffix to the error
message and because of that, the error message shown, is non-sensical

## QA steps

N/A

## Documentation changes

N/A

## Bug reference

N/A


SimonRichardson:
$$merge$$

SimonRichardson:
https://trello.com/c/9UWawOSN/442-fix-go-vet-issue-with-caas-k8s-error-message

PR #9024, opened 2018-08-07
Ensure the model owner set in the params for the list offers api call

wallyworld:
## Description of change

When making a ListOffers api call, ensure the model owner attribute is set in the passed params.

## QA steps

Deploy an offer. Grant access to a user.
As that user in another session, juju offers <controller>:<model>
Ensure that any offers are shown.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1785223


wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #9025, opened 2018-08-07
Make juju_engine_report more useful

howbazaar:
The engine report that is available through the introspection endpoint had limited usefulness.

The engine now keeps a count of the number of times a worker has started, and also the time at which the worker was started.

This branch also hooks up the Report method for a number of workers. In particular the model worker, as it has the nested dependency engine.

## QA steps

- bootstrap a controller
- deploy something in the default model
- juju ssh -m controller 0
- juju_engine_report


howbazaar:
Some tests still pending, getting up for the build checks and to follow the diff.

howbazaar:
$$merge$$

PR #9027, opened 2018-08-07
Make tenant name as optional

vinu2003:
## Description of change
This Pr has code changes that makes the tenant Name as optional choice when the openstack credentials are invalid.
## Bug reference
https://bugs.launchpad.net/bugs/1785747


provider/openstack/credentials.go
```diff
@@ -117,7 +117,7 @@ func (c OpenstackCredentials) DetectCredentials() (*cloud.CloudCredential, error
 func (c OpenstackCredentials) detectCredential() (*cloud.Credential, string, string, error) {
 	creds := identity.CredentialsFromEnv()
 	if creds.TenantName == "" {
```
wallyworld: This isn't sufficient - you need to search for all usages of TenantName and fix them - it is used in various places. This PR also doesn't fix the credential schema which as TenantName as a mandatory attribute. You need to QA the change as any QA would have shown these issues. And the PR contains no unit tests.

PR #9030, opened 2018-08-07
Add support for autocert new TLS-ALPN challenge

frankban:
Rather than using the HTTP challenge, requiring port 80 to be open, and therefore an additional handler, we use the new TLS-ALPN challenge, which works using the TLS handshake on 443.

**QA:**
Assuming $DNSNAME is a DNS name that you control:
- bootstrap juju with `--config autocert-dns-name=$DNSNAME`; 
- when the bootstrap process completes, point $DNSNAME to the ip address of the controller;
- wait for the change to propagate;
- run `juju gui`: the output should suggest a URL to visit which includes $DNSNAME;
- visit the GUI at that URL, and it should work;
- deploying charms, running `juju status`, i.e. accessing the API should work.

### DIFF
--- a/apiserver/apiserver.go
+++ b/apiserver/apiserver.go
@@ -290,9 +290,9 @@ func newServer(cfg ServerConfig) (_ *Server, err error) {
 	srv.logSinkWriter = logSinkWriter
 
 	if cfg.PrometheusRegisterer != nil {
-		apiserverCollectior := NewMetricsCollector(&metricAdaptor{srv})
-		cfg.PrometheusRegisterer.Unregister(apiserverCollectior)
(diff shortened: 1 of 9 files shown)


apiserver/apiserver.go
```diff
@@ -290,9 +290,9 @@ func newServer(cfg ServerConfig) (_ *Server, err error) {
 	srv.logSinkWriter = logSinkWriter
 
 	if cfg.PrometheusRegisterer != nil {
-		apiserverCollectior := NewMetricsCollector(&metricAdaptor{srv})
```
frankban: Drive-by fix of a typo.

SimonRichardson: :+1:

frankban:
Thanks for the review!
$$merge$$

PR #9032, opened 2018-08-07
Update the worker.v1 dependency.

howbazaar:
The catacomb, dependency, and workertest packages are now available in "gopkg.in/juju/worker.v1".

This branch removes those packages from the worker package locally and updates all imports and usage to use the packages from worker.v1.

As a drive by, the unnecessary "worker" prefix on the import has been removed.


howbazaar:
$$merge$$

PR #9035, opened 2018-08-08
Fixes lp#1779337: Local charms must have hooks to be valid.

anastasiamac:
## Description of change

As per linked bug, users may easily attempt to deploy local charms that have not been built. Well-built charms have non-empty hooks directory.

This PR checks that local charms are well-built.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1779337

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -4,12 +4,14 @@
 package api
 
 import (
+	"archive/zip"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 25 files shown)


anastasiamac:
[check local charm valid, lp#1779337](https://trello.com/c/KMrEmsOS/379-check-local-charm-valid-lp1779337)

api/client.go
```diff
@@ -329,13 +331,52 @@ func (c *Client) AddLocalCharm(curl *charm.URL, ch charm.Charm) (*charm.URL, err
 		return nil, errors.Errorf("unknown charm type %T", ch)
 	}
 
-	curl, err := c.UploadCharm(curl, archive)
+	anyHooks, err := hasHooks(archive.Name())
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if !anyHooks {
+		return nil, errors.Errorf("invalid charm %q: has no hooks", curl.Name)
```
wallyworld: perhaps suggest to run charm build if it's a reactive charm?

api/client.go
```diff
@@ -329,13 +331,52 @@ func (c *Client) AddLocalCharm(curl *charm.URL, ch charm.Charm) (*charm.URL, err
...
+		return false, err
+	}
+	defer zipr.Close()
+
+	count := 0
+	// Cater for file separator differences between operating systems.
+	hooksPath := filepath.FromSlash("hooks/")
+	for _, f := range zipr.File {
+		if strings.Contains(f.Name, hooksPath) {
+			count++
```
wallyworld: Why not just return true here? I don't get why we need count at all?

featuretests/cmd_juju_deploy_test.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package featuretests
+
+import (
+	"github.com/juju/cmd/cmdtesting"
+	"github.com/juju/errors"
+	"github.com/juju/juju/testcharms"
```
wallyworld: move me

api/client.go
```diff
@@ -329,13 +331,52 @@ func (c *Client) AddLocalCharm(curl *charm.URL, ch charm.Charm) (*charm.URL, err
...
+	curl, err = c.UploadCharm(curl, archive)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
 	return curl, nil
 }
 
+var hasHooks = hasHooksFolder
+
+func hasHooksFolder(name string) (bool, error) {
```
wallyworld: I guess this is covered by feature test.
Would be good to see unit test as well.

api/client.go
anastasiamac: I explained in the comment that finding just 1 occurrence of "hooks/" is not enough :) It could be just an empty folder called "hooks". We need to see more than one reference to "hooks/" in the file names.

api/client.go
anastasiamac: added

featuretests/cmd_juju_deploy_test.go
anastasiamac: done

api/client.go
anastasiamac: We do not really make suggestions in errors coming from apiserver. As discussed elsewhere, current message is sufficient.

anastasiamac:
This was discussed and anticipated... All these tests are currently failing since out testing uses charms that we are no longer consider invalid. I'll address this shortly. 

Test Result (3 failures / +1)

    github.com/juju/juju/worker/provisioner.TestPackage
    github.com/juju/juju/cmd/juju/application.TestPackage
    github.com/juju/juju/cmd/juju/commands.TestPackage

anastasiamac:
$$merge$$

PR #9041, opened 2018-08-09
Fix machine-lock issue with container initialisation.

howbazaar:
With the recent changes in the machine lock, the cancel channel became non-optional. There was the ability to explicitly say "there is no cancel", but we don't want to use that for the container initialisation as it should be cancelled when the provisioner is being shut down.

This branch adds an abort channel to the StartInstanceParams. This is constructed by the provisioner and passed down through the broker calls, which both the KVM and LXD brokers support. This channel is then passed through to the machine lock call.

## QA steps

I tried to replicate on EC2 but it seems the problem occurs more in MAAS and the bridging code.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1786099

### DIFF
--- a/environs/broker.go
+++ b/environs/broker.go
@@ -88,6 +88,10 @@ type StartInstanceParams struct {
 	// changes in status. Its signature is consistent with other
 	// status-related functions to allow them to be used as callbacks.
 	StatusCallback StatusCallbackFunc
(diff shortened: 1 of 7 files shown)


howbazaar:
$$merge$$

PR #9050, opened 2018-08-10
makefile: Reduce the noise from `make install`

babbageclunk:
## Description of change

Now that dep is working, remove the verbose flag to reduce the output from running `make install`. Also, since we now need to explicitly list packages in the `go install` command to avoid the vendored packages, echo a sanitised version of the command, rather than the real (very long) one.

These tweaks reduce the output from a nothing-changed build from 38k to about 3 lines.

## QA steps

Run `make install` - output is mostly restricted to listing packages as they are being built.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ $(GOPATH)/bin/dep:
 
 # populate vendor/ from Gopkg.lock without updating it first (lock file is the single source of truth for machine).
 dep: $(GOPATH)/bin/dep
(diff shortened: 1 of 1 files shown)


babbageclunk:
!!build!!

ycliuhw:
Looks awesome, thanks!

babbageclunk:
$$merge$$

That @veebers!

babbageclunk:
That @Veebers!

$$merge$$

PR #9051, opened 2018-08-10
Fixes lp#1786061: added uuid back for backward compatibility.

anastasiamac:
## Description of change

As per linked bug, some scripts may be broken due to a property rename in yaml output.

This PR adds back old names: uuid' for controller on both 'controllers' and 'show-controller' command as well as 'uuid' for model in 'show-controller' command.

Note that it means that both new and old properties are now part of the output - 'controller-uuid' and 'model-uuid' as well as the old 'uuid' for both controller and model, respectively.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1786061

### DIFF
--- a/cmd/juju/controller/listcontrollers_test.go
+++ b/cmd/juju/controller/listcontrollers_test.go
@@ -139,6 +139,7 @@ controllers:
     current-model: controller
     user: admin
     recent-server: this-is-aws-test-of-many-api-endpoints
+    uuid: this-is-the-aws-test-uuid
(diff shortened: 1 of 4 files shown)


anastasiamac:
[add back 'uuid' to 'controllers' and 'show-controller' output for backward compat, lp#1786061](https://trello.com/c/SMl3z22s/381-add-back-uuid-to-controllers-and-show-controller-output-for-backward-compat-lp1786061)

anastasiamac:
$$merge$$

PR #9052, opened 2018-08-10
Fixes lp#1786253: backup create help and result messaging.

anastasiamac:
## Description of change

From 2.4 we have introduced a slight change in the way that 'create-backup' command was run - backups are now downloaded to the client by default instead of the old way where we stored them remotely by default.

This PR updates help text for the command.
But also ensures that the operator is clear whether the backup is available remotely or not. 

Whilst running unit tests locally, I have also noticed that we were not removing test-generated files. I have adjusted tests' cleanup.

## Documentation changes

@juju/docs I *think* this has been covered when the original behavior change was done to resolve https://bugs.launchpad.net/juju/+bug/1749300 in https://github.com/juju/juju/pull/8663. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1786253

### DIFF
--- a/cmd/juju/backups/create.go
+++ b/cmd/juju/backups/create.go
@@ -20,16 +20,30 @@ import (
 const (
 	notset          = backups.FilenamePrefix + "<date>-<time>.tar.gz"
 	downloadWarning = "downloading backup archives is recommended; " +
-		"backups stored remotely are not guaranteed to be available"
+		"backups stored remotely are not guaranteed to be available."
 )
 
 const createDoc = `
-create-backup requests that Juju creates a backup of its state and prints the
+This command requests that Juju creates a backup of its state and prints the
 backup's unique ID.  You may provide a note to associate with the backup.
 
-The backup archive and associated metadata are stored remotely by Juju, but
-will also be copied locally unless --no-download is supplied. To access the
-remote backups, see 'juju download-backup'.
+By default, the backup archive and associated metadata are downloaded without keeping a copy remotely on the controller.
+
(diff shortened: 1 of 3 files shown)


cmd/juju/backups/create.go
```diff
@@ -20,16 +20,30 @@ import (
...
 backup's unique ID.  You may provide a note to associate with the backup.
 
-The backup archive and associated metadata are stored remotely by Juju, but
-will also be copied locally unless --no-download is supplied. To access the
-remote backups, see 'juju download-backup'.
+By default, the backup archive and associated metadata are downloaded without keeping a copy remotely on the controller.
+
+Use --no-download to avoid getting a local copy of the backup downloaded at the end of the backup process. 
+
+Use --keep-copy option to store a copy of this backup remotely on the controller. Ignored when --no-download is used.
```
wallyworld: I'd prefer it to complain instead of ignoring. The user intent may have been to --keep-copy and they typed --no-download by mistake. We should be strict about accepting valid CLI args.

anastasiamac: Nice :+1: 
added a complain :D

anastasiamac:
$$merge$$

anastasiamac:
$$merge$$

anastasiamac:
unrelated failure in github.com/juju/juju/worker/provisioner.TestPackage
as seen in http://ci.jujucharms.com/job/github-merge-juju/912/

anastasiamac:
$$merge$$

PR #9054, opened 2018-08-10
Version bionic

SimonRichardson:
Note: this is an update to https://github.com/juju/juju/pull/9023 where
it also featured the feature tests. In order to move this PR forward quicker,
I've remove the go code, so we can merge that and then we can have a
PR just for the feature tests.

## Description of change

This makes sure that all LXD images by default are Bionic for 2.5
release. The issue with this is that it can cause issues from this,
because we're now bumping the LTS version for EVERYTHING.

A lot of testing should be done to ensure that we don't cause other
issues for other parts of the system.

## QA steps

Test that launching LXD providers is possible AND that other 
providers/machines aren't broken in changing this.

## Documentation changes

The default LTS for juju is changing from Xenial to Bionic. You can still
over ride this using the series flag if required.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1785124


SimonRichardson:
!!build!!

provider/ec2/local_test.go
```diff
@@ -1566,7 +1566,7 @@ func (t *localServerSuite) TestValidateImageMetadata(c *gc.C) {
 	image_ids, _, err := imagemetadata.ValidateImageMetadata(params)
 	c.Assert(err, jc.ErrorIsNil)
 	sort.Strings(image_ids)
-	c.Assert(image_ids, gc.DeepEquals, []string{"ami-00000133", "ami-00000135", "ami-00000139"})
+	c.Assert(image_ids, gc.DeepEquals, []string{"ami-00001133", "ami-00001135", "ami-00001139"})
```
mitechie: what's this change?

SimonRichardson: This was using mock AMIs and I just wanted to highlight that bionic had a unique AMI code for the test. "ami-00000133" vs "ami-00001133".

SimonRichardson:
$$merge$$

PR #9055, opened 2018-08-10
Ensure uploading of lxd cert once

SimonRichardson:
## Description of change

The following just ensure that we upload the cert once, so that we
can reason about what certificates are in flight at anyone time.

This will help with multiple certificates error

## QA steps

```sh
juju add-cloud lxd ~/path/clouds.yaml
juju add-credential lxd -f ~/path/creds.yaml
juju remove-credential lxd lxd
juju add-credential lxd -f ~/path/creds.yaml
```

## Documentation changes

N/A

## Bug reference

This is step one of many to fix:
https://bugs.launchpad.net/juju/+bug/1786276

### DIFF
--- a/provider/lxd/credentials.go
+++ b/provider/lxd/credentials.go
@@ -393,15 +393,23 @@ func (p environProviderCredentials) finalizeRemoteCredential(
 	}
 
 	// check to see if the cert already exists
-	if err := server.CreateCertificate(api.CertificatesPost{
-		CertificatePut: api.CertificatePut{
-			Name: credentials.Label,
-			Type: "client",
-		},
-		Certificate: base64.StdEncoding.EncodeToString(clientX509Cert.Raw),
-		Password:    trustPassword,
-	}); err != nil {
+	fingerprint, err := clientCert.Fingerprint()
+	if err != nil {
 		return nil, errors.Trace(err)
+	}
+
+	cert, _, err := server.GetCertificate(fingerprint)
+	if err != nil || cert == nil {
+		if err := server.CreateCertificate(api.CertificatesPost{
+			CertificatePut: api.CertificatePut{
+				Name: credentials.Label,
+				Type: "client",
+			},
+			Certificate: base64.StdEncoding.EncodeToString(clientX509Cert.Raw),
+			Password:    trustPassword,
+		}); err != nil {
+			return nil, errors.Trace(err)
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


SimonRichardson:
https://trello.com/c/YJWFwSJ1/447-lxd-multiple-credentials

SimonRichardson:
@hmlanigan turns out I've fixed this bug https://bugs.launchpad.net/juju/+bug/1784721 with this commit, whilst looking at the multi-credential issue. Only realised when looking at the board.

:crying_cat_face:

provider/lxd/credentials.go
```diff
@@ -393,17 +393,25 @@ func (p environProviderCredentials) finalizeRemoteCredential(
...
+				Name: credentials.Label,
+				Type: "client",
+			},
+			Certificate: base64.StdEncoding.EncodeToString(clientX509Cert.Raw),
+			Password:    trustPassword,
+		}); err != nil {
+			return nil, errors.Trace(err)
+		}
+	}
+
```
hmlanigan: nit ln 415: do we want a different msg if the certificate is already there?

provider/lxd/credentials_test.go
```diff
@@ -644,6 +647,61 @@ func (s *credentialsSuite) TestFinalizeCredentialNonLocal(c *gc.C) {
 	c.Assert(got, jc.DeepEquals, &expected)
 }
 
+func (s *credentialsSuite) TestFinalizeCredentialNonLocalWithCertAlreadyExists(c *gc.C) {
+	ctrl := gomock.NewController(c)
+	defer ctrl.Finish()
+
+	deps := s.createProvider(ctrl)
+
+	insecureCred := cloud.NewCredential(cloud.CertificateAuthType, map[string]string{
```
hmlanigan: It appears that the same insecureCred, insecureSpec, secureCred, secureSpec and params are reused in multiple tests, can we set them up to be global or shared etc.?

SimonRichardson:
@hmlanigan unfortunately there is no call for cleaning up credentials from the cloud, they always become an orphan. There are issues where the credential might have expired from the cloud and we can no longer access the cloud, or how do you remove a credential that you're communicating with. 

So it makes the code really complicated to do a really thorough clean up, so the easy path seems to have been chosen.

SimonRichardson:
$$merge$$

PR #9059, opened 2018-08-13
fix lp: 1783407 fix interactive credentials type for lxd

hmlanigan:
## Description of change

Allow for interactive authtype in CredentialSchemas,that isn't in the CloudSchema. If specified will be the default for add-credential.

Allow for a Default to be used in pollster for Schema ArrayTypes.

## QA steps

1. juju add-cloud for lxd with a remote.  The default auth type should be "certificates".
2. juju add-credential for the new lxd cloud should mention auth types of both certificate and interactive.  Interactive being the default.
3. juju add-credential -f <file> for the new lxd cloud with an authtype of interactive should work.
4. juju bootstrap the new lxd cloud should succeed.
3. no change to juju add-credential for azure.

## Documentation changes

Not clear.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1783407

### DIFF
--- a/cmd/juju/interact/pollster_test.go
+++ b/cmd/juju/interact/pollster_test.go
@@ -504,6 +504,76 @@ Select one or more numbers separated by commas:
 	c.Check(s, jc.SameContents, []string{"one", "three"})
 }
 
+func (PollsterSuite) TestQueryArraySchemaDefault(c *gc.C) {
+	schema := &jsonschema.Schema{
+		Singular: "number",
+		Plural:   "numbers",
+		Type:     []jsonschema.Type{jsonschema.ArrayType},
+		Default:  "two",
+		Items: &jsonschema.ItemSpec{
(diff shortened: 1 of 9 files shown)


cmd/juju/interact/pollster_test.go
```diff
@@ -504,6 +504,76 @@ Select one or more numbers separated by commas:
 	c.Check(s, jc.SameContents, []string{"one", "three"})
 }
 
+func (PollsterSuite) TestQueryArraySchemaDefault(c *gc.C) {
```
mitechie: are these missing tests for existing functionality or is this used elsewhere in the PR I missed?

hmlanigan: These are to test that the Default schema piece is used with the array type.  The array type was currently tested without that piece.

hmlanigan:
$$merge$$

PR #9061, opened 2018-08-14
Change default presense implementation to core/presence.

howbazaar:
In 2.4 we added the core/presence package running in tandem with the existing state/presence implementation. The new presence checks could be turned on with a controller freature flag.

This branch changes the default to be the new presence, with the old presence implementation available as a feature flag.

## QA steps

The unit tests passing, followed by CI.

### DIFF
--- a/provider/dummy/environs.go
+++ b/provider/dummy/environs.go
@@ -256,6 +256,7 @@ type environState struct {
 	apiServer      *apiserver.Server
 	apiState       *state.State
 	apiStatePool   *state.StatePool
+	presence       *fakePresence
 	creator        string
 }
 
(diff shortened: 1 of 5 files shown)


provider/dummy/environs.go
```diff
@@ -1797,3 +1799,44 @@ func (*environ) SSHAddresses(ctx context.ProviderCallContext, addresses []networ
...
+func (e *environ) SetAgentStatus(agent string, status presence.Status) {
+	estate, err := e.state()
+	if err != nil {
+		panic(err)
+	}
+	estate.presence.agent[agent] = status
+}
+
+// fakePresence returns alive for all agent alive requests.
+type fakePresence struct {
```
wallyworld: Can we just embed the Recorder interface here to not have to provide empty implementations of everything? Are all those stub methods called?

howbazaar:
$$merge$$

PR #9062, opened 2018-08-14
Version bionic 2.4 backport

SimonRichardson:
## Description of change

This makes sure that all LXD images by default are Bionic for 2.4
release. The issue with this is that it can cause issues from this,
because we're now bumping the LTS version for EVERYTHING.

A lot of testing should be done to ensure that we don't cause other
issues for other parts of the system.

## QA steps

Test that launching LXD providers is possible AND that other 
providers/machines aren't broken in changing this.

## Documentation changes

The default LTS for juju is changing from Xenial to Bionic. You can still
over ride this using the series flag if required.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1785124


SimonRichardson:
https://trello.com/c/tQ5FN36Z/456-backport-lxd-controller-running-xenial-not-bionic-to-24

mitechie:
LGTM

PR #9064, opened 2018-08-15
Avoid starting presence pingers unless needed.

howbazaar:
With the new presence code now being the default in 2.5, we want to avoid starting presence pingers.

If the operator does set the old-presence feature flag, the api server will start presence pingers for the agents. However to ensure that we get all the agents, we restart the apiserver when the old-presence feature flag is either set or unset. This will force disconnect all the agents, and they will reconnect and start the pingers if necessary.

I did start exposing the hub for the JujuConnSuite only to not actually need it. But there is no harm now in leaving it there.

### DIFF
--- a/apiserver/admin.go
+++ b/apiserver/admin.go
@@ -20,6 +20,7 @@ import (
 	"github.com/juju/juju/apiserver/observer"
 	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/core/auditlog"
+	"github.com/juju/juju/feature"
 	"github.com/juju/juju/permission"
 	"github.com/juju/juju/rpc"
 	"github.com/juju/juju/rpc/rpcreflect"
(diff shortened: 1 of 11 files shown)


apiserver/shared.go
```diff
@@ -9,13 +9,22 @@ import (
...
+	"github.com/juju/juju/feature"
+	"github.com/juju/juju/pubsub/apiserver"
 	"github.com/juju/juju/pubsub/controller"
 	"github.com/juju/juju/state"
 )
 
+// SharedHub represents the methods of the pubsub.StructuredHub
+// that are used. The context uses an interface to allow mocking
+// of the hub.
+type SharedHub interface {
```
babbageclunk: This is fun - now any facade can bounce the API server! (Not a serious complaint but, you know...)

pubsub/apiserver/messages.go
```diff
@@ -90,3 +90,11 @@ type PresenceResponse struct {
...
 // topics.
 type OriginTarget common.OriginTarget
+
+// RestartTopic is used by the API server to listen for events that should
+// cause the API server to be bounced.
+const RestartTopic = "apiserver.restart"
+
+// Restart message only contains the local-only indicator as the restart
+// is only ever for the same agent.
+type Restart common.LocalOnly
```
babbageclunk: Cute!

apiserver/admin.go
```diff
@@ -462,26 +463,32 @@ func (shim presenceShim) Start() (presence.Pinger, error) {
...
 func startPingerIfAgent(clock clock.Clock, root *apiHandler, entity state.Entity) error {
-	// worker runs presence.Pingers -- absence of which will cause
-	// embarrassing "agent is lost" messages to show up in status --
-	// until it's stopped. It's stored in resources purely for the
-	// side effects: we don't record its id, and nobody else
-	// retrieves it -- we just expect it to be stopped when the
-	// connection is shut down.
-	agent, ok := entity.(statepresence.Agent)
-	if !ok {
+	tag := entity.Tag()
```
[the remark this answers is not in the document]
howbazaar: I need an early exit, and the check below is only checked if the feature flag is set.  The hunk below should be fully removed when we remove the old presence implementation.

apiserver/shared.go
howbazaar: We can probably remove this when whe presence code is gone :)

howbazaar:
$$merge$$

PR #9066, opened 2018-08-15
2.3 try stopped 1785623

jameinel:
## Description of change

In bug #1785623 when a controller was failing to connect, all we were getting was an unhelpful "try was stopped". I was able to reproduce that with a socket that did not have a valid server running. This just changes the error code path to return the last known error when connecting.

## QA steps

See the associated whitebox tests. Without this, you get "try was stopped" as the error.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1785623

### DIFF
--- a/api/apiclient.go
+++ b/api/apiclient.go
@@ -675,7 +675,6 @@ func dialWebsocketMulti(ctx context.Context, addrs []string, path string, opts d
 			// the addresses are checked with Info.Validate
 			// beforehand.
 			err := errors.Errorf("invalid address %q: %v", addr, err)
-			logger.Criticalf("dialWebsocketMulti got: %v", err)
 			recordTryError(try, err)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


jameinel:
[1785623 "try was stopped"](https://trello.com/c/377krfHY/450-1785623-try-was-stopped)

api/apiclient.go
```diff
@@ -806,6 +806,7 @@ type dialer struct {
 // when appropriate.
 func (d dialer) dial(_ <-chan struct{}) (io.Closer, error) {
 	a := retry.StartWithCancel(d.openAttempt, d.opts.Clock, d.ctx.Done())
+	var lastErr error = nil
```
SimonRichardson: Out of curiosity what's the reason for setting `= nil`, as `var lastErr error` will be nil anyway?

jameinel: Not particularly needed. I think this started out as some sort of "lastErr := nil", but since that isn't a typed nil, I ended up with var ... error, and left the '='.

jameinel:
$$merge$$

PR #9068, opened 2018-08-15
Refactor lock to one state

ExternalReality:
Unify the two lock statuses of the state.


jameinel:
Just checking that this actually ended up merged into the other patch, is that correct?

PR #9069, opened 2018-08-16
More accurate create backup option check.

anastasiamac:
## Description of change

When requesting a backup, it is possible for user to specify not to download created backup file at the same time as instructing Juju not to keep a remote copy, '--no-download --keep-copy=false'. This would mean that we will do a lot of work but there will be no output. Of course, '--no-download --keep-copy' whilst redundant, is actually ok.

This PR tightens this check and ensures that there are unit tests.

### DIFF
--- a/cmd/juju/backups/create.go
+++ b/cmd/juju/backups/create.go
@@ -97,8 +97,8 @@ func (c *createCommand) Init(args []string) error {
 	// all the backup will not be stored anywhere.
 	if c.NoDownload {
 		keepCopySet := false
-		c.fs.Visit(func(blah *gnuflag.Flag) {
-			if blah.Name == "keep-copy" {
+		c.fs.Visit(func(flag *gnuflag.Flag) {
+			if flag.Name == "keep-copy" {
 				keepCopySet = true
 			}
 		})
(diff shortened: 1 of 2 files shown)


anastasiamac:
[better check for --no-download --keep-copy on create backup](https://trello.com/c/5SauFoSM/399-better-check-for-no-download-keep-copy-on-create-backup)

cmd/juju/backups/create.go
```diff
@@ -83,30 +83,28 @@ func (c *createCommand) Info() *cmd.Info {
...
-		}
-		if blah.Name == keepCopyOption {
-			keepCopySet = true
+	// If user specifies that a download is not desired (i.e. no-download == true),
+	// and they have EXPLICITLY not wanted to store a remote backup file copy
+	// (i.e keep-copy == false), then there is no point for us to proceed as
+	// all the backup will not be stored anywhere.
+	if c.NoDownload {
+		keepCopySet := false
+		c.fs.Visit(func(blah *gnuflag.Flag) {
```
babbageclunk: I realise this was also in the old version, but `flag` would be a much better name than `blah`. Petty but I find that kind of thing really throws me off when reading code.

anastasiamac: Oh so true. Changed \o/
(the old version was mine too...)

anastasiamac:
$$merge$$

PR #9070, opened 2018-08-16
Corrected the help for exportbundle command

vinu2003:
This PR has code changes that fix the help for export bundle command. 
There shouldn't be --output and --format options in thge help menu. 

## QA steps
execute juju export-bundle --help command.

### DIFF
--- a/api/bundle/client.go
+++ b/api/bundle/client.go
@@ -34,7 +34,7 @@ func NewClient(st base.APICallCloser) *Client {
 func (c *Client) ExportBundle() (string, error) {
 	var result params.StringResult
 	if bestVer := c.BestAPIVersion(); bestVer < 2 {
-		return "", errors.Errorf("ExportBundle() on v%d is not supported", bestVer)
+		return "", errors.Errorf("this controller version does not support bundle export feature.")
 	}
 
 	if err := c.facade.FacadeCall("ExportBundle", nil, &result); err != nil {
--- a/api/bundle/client_test.go
+++ b/api/bundle/client_test.go
@@ -46,7 +46,7 @@ func (s *bundleMockSuite) TestFailExportBundlev1(c *gc.C) {
 	)
 	result, err := client.ExportBundle()
 	c.Assert(err, gc.NotNil)
-	c.Assert(err.Error(), gc.Equals, "ExportBundle() on v1 is not supported")
+	c.Assert(err.Error(), gc.Equals, "this controller version does not support bundle export feature.")
 	c.Assert(result, jc.DeepEquals, "")
 }
 
--- a/cmd/juju/model/exportbundle.go
+++ b/cmd/juju/model/exportbundle.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/juju/juju/api/bundle"
 	"github.com/juju/juju/cmd/modelcmd"
-	"github.com/juju/juju/cmd/output"
 )
 
 // NewExportBundleCommand returns a fully constructed export bundle command.
@@ -32,7 +31,7 @@ type exportBundleCommand struct {
 }
 
 const exportBundleHelpDoc = `
-Exports the current model configuration as bundle.
+Exports the current model configuration as a reusable bundle.
 
 If --filename is not used, the configuration is printed to stdout.
  --filename specifies an output file.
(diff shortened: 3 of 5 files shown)


cmd/juju/model/exportbundle.go
```diff
@@ -55,7 +54,7 @@ func (c *exportBundleCommand) Info() *cmd.Info {
 // SetFlags implements Command.
 func (c *exportBundleCommand) SetFlags(f *gnuflag.FlagSet) {
 	c.ModelCommandBase.SetFlags(f)
-	c.out.AddFlags(f, "yaml", output.DefaultFormatters)
+	//c.out.AddFlags(f, "yaml", output.DefaultFormatters)
```
anastasiamac: Delete this all together if not needed?...

vinu2003: we dont need to have --format and --output options. AddFLags just injects both cli flags. So i removed.

howbazaar: You'll need to modify the body of the run method too most likely...

anastasiamac: My original comment was to point out that it is not good enough to comment this code out. Remove it! 
We don not want to clatter our production code with commented out code :D

vinu2003:
Hi @howbazaar ,

I havent read your comment. But i just fixed the stdout print as below,
c.out.WriteFormatter(ctx, cmd.FormatYaml, result)

i didnt use fmt.Fprintf()...

question:
Why have you suggested ioutil.writeFile ?

Also,
You have mentioned let user decide the extension and we should not append it to filename.
I have done that and pushed another commit.

But, the GUI appends as '<modelname>.yaml'.
Please let me know your input i can revert back that if we should match with GUI.

mitechie:
For the GUI comparison, the GUI bdoesnt allow you to provide a requested
filename. You just get the same name each time. Since, in the CLI case, we
allow the user to express a filename opinion infeel we should respect it.
If they script some action and provide a filename I believe they'd expect
to be able to use the same filename later in their scripting without issue.

vinu2003:
back and just verified with fmt.Printf() the pipe issie disappeared. So i beleive all cosmetic issues are fixed for bundle export.

vinu2003:
Hi @mitechie Agree with your point.
Thanks for pointing this out. I have also verified it is not mandatory to be with .yaml extension to get deployed. So its all ok for now.

mitechie:
QA ok on the changes made. The help is improved and the format of the terminal output is reusable when piped to a file correctly.

Can you also edit the Summary please? 

    Exports the current model configuration as bundle.

to 

    Exports the current model configuration as a reusable bundle.

Also please add a newline after the example description to line up with the help output template
https://docs.google.com/document/d/1ySjCNqd0x6veLfcBetxLI9NH7qfw3xayLWJNqXMvyW8/edit

In that example can you add the .yaml to the example filename so that it shows that it's respecting what the user types and encourages them to have that even though we don't require it?

Thanks for the updates!

vinu2003:
Hi @mitechie ,
Thanks again for your reviews. I have made the changes requested.

api/bundle/client.go
```diff
@@ -34,7 +34,7 @@ func NewClient(st base.APICallCloser) *Client {
 func (c *Client) ExportBundle() (string, error) {
 	var result params.StringResult
 	if bestVer := c.BestAPIVersion(); bestVer < 2 {
-		return "", errors.Errorf("ExportBundle() on v%d is not supported", bestVer)
+		return "", errors.Errorf("this controller version v%d doen not support bundle feature.", bestVer)
```
mitechie: This output a v1 in the QA work. That's not the controller version so this turns into a bit of misinformation doesn't it?

api/bundle/client_test.go
```diff
@@ -46,7 +46,7 @@ func (s *bundleMockSuite) TestFailExportBundlev1(c *gc.C) {
 	)
 	result, err := client.ExportBundle()
 	c.Assert(err, gc.NotNil)
-	c.Assert(err.Error(), gc.Equals, "ExportBundle() on v1 is not supported")
+	c.Assert(err.Error(), gc.Equals, "this controller version v1 doen not support bundle feature.")
```
mitechie: same here, v1 isn't the controller version

cmd/juju/model/exportbundle.go
```diff
@@ -32,30 +31,30 @@ type exportBundleCommand struct {
 }
 
 const exportBundleHelpDoc = `
-Exports the current model configuration as bundle.
+Exports the current model configuration as a reusable bundle.
```
mitechie: thank you

cmd/juju/model/exportbundle.go
```diff
@@ -32,30 +31,30 @@ type exportBundleCommand struct {
...
 
 If --filename is not used, the configuration is printed to stdout.
  --filename specifies an output file.
 
 Examples:
 
     juju export-bundle
-	juju export-bundle --filename mymodel
+	juju export-bundle --filename mymodel.yaml
```
mitechie: <3

vinu2003:
$$merge$$

PR #9073, opened 2018-08-16
Fixes lp#1780711: Ignore all formatting when only one output is desired.

anastasiamac:
## Description of change

Application configuration string options may contain multi-lined values. When formatted these values get un-desired newlines which are supposedly helpful for human readers and are supplied by json and ymal formatters. However, these newlines are highly undesired when scripting.

As per suggestion in the linked bug, this PR ensures that a single-value output is not formatted. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1780711

### DIFF
--- a/cmd/juju/application/config.go
+++ b/cmd/juju/application/config.go
@@ -355,7 +355,7 @@ func (c *configCommand) getConfig(client applicationAPI, ctx *cmd.Context) error
 		return err
 	}
 	if len(c.keys) == 1 {
-		ctx.Infof("format %v is ignored", c.out.Name())
+		logger.Infof("format %v is ignored", c.out.Name())
 		key := c.keys[0]
 		info, found := results.CharmConfig[key].(map[string]interface{})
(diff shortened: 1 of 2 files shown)


anastasiamac:
[unformat single value config output, lp#1780711](https://trello.com/c/QxHWRk6Q/401-unformat-single-value-config-output-lp1780711)

cmd/juju/application/config.go
```diff
@@ -350,6 +355,7 @@ func (c *configCommand) getConfig(client applicationAPI, ctx *cmd.Context) error
 		return err
 	}
 	if len(c.keys) == 1 {
+		ctx.Infof("format %v is ignored", c.out.Name())
```
howbazaar: I have a strong suspicion that this will confuse people as it will show output any time they ask for something.

Do you think we really need this? Perhaps as debug output?

anastasiamac:
$$merge$$

PR #9076, opened 2018-08-16
Cleanup

vinu2003:
This PR has code changes to make export-bundle command available only in dev mode.
## QA steps
execute the juju export-bundle  command in dev-mode ie with JUJU_FEATURE_FLAG=developer_mode
## Documentation changes
Yes. the doc needs update.


vinu2003:
Yes. The unit tests i missed. thanks Anastasia.

acceptancetests/assess_bundle_export.py
```diff
@@ -1,77 +0,0 @@
-#!/usr/bin/env python
```
howbazaar: We probably don't actually want to remove the test, but just not have it hooked up to a jenkins job right?

vinu2003: Yes. we don't have to. Just to keep it clean in this branch.
Also, I closed the other PR which i created in qa_jenkins. But that is required for 2.5 ci test as we have removed the dev-mode flag there.
If we have this test here it can trigger and the command is not exposed. So better to clean.

vinu2003: @howbazaar - Do i make sense here. correct me if Iam wrong please.

howbazaar: I think it is fine to keep it. In particular the feature flag around bundle-export should be removed again once we release 2.4.2.

It is much easier to revert a very small change, and there is less churn on the codebase. Can we please keep it in the tree?

vinu2003:
$$merge$$

PR #9077, opened 2018-08-16
Feature/custom resource definition

ycliuhw:
## Description of change

Implement `CustomResourceDefinition` feature for caas unit provisioning

## QA steps

```bash
juju deploy ./builds/crd-charm 
kubectl get crd
```

## Documentation changes

N/A


ycliuhw:
[CustomResourceDefinition support](https://trello.com/c/r8SXiWov/389-customresourcedefinition-support)

caas/broker.go
```diff
@@ -109,6 +109,9 @@ type Broker interface {
 	// EnsureService creates or updates a service for pods with the given params.
 	EnsureService(appName string, params *ServiceParams, numUnits int, config application.ConfigAttributes) error
 
+	// EnsureCrd creates or updates a custom resource definition resource.
+	EnsureCrd(appName string, podSpec *PodSpec) error
```
wallyworld: There seems to be a mix of "Crd" and "CustomResourceDefinition" (here and elsewhere).
k8s itself seems to prefer "CustomResourceDefinition" in it's yaml tags etc. I know it's longer but I think it reads better?

caas/containers.go
```diff
@@ -55,8 +58,40 @@ type ContainerSpec struct {
...
+	Group      string                `yaml:"group"`
+	Version    string                `yaml:"version"`
+	Scope      string                `yaml:"scope"`
+	Validation CrdTemplateValidation `yaml:"validation,omitempty"`
+}
+
+// Validate returns an error if the crd is not valid.
+func (crd *CustomResourceDefinition) Validate() error {
+	if crd.Kind == "" {
+		return errors.New("kind is missing")
```
wallyworld: We tend to use errors.NotValidf("missing Kind") etc

caas/kubernetes/provider/base_test.go
```diff
@@ -40,6 +41,10 @@ type BaseSuite struct {
 	mockStorage                *mocks.MockStorageV1Interface
 	mockStorageClass           *mocks.MockStorageClassInterface
 	mockIngressInterface       *mocks.MockIngressInterface
+
+	mockApiextensionsV1     *mocks.MockApiextensionsV1beta1Interface
+	mockApiextensionsClient *mocks.MockApiExtensionsClientInterface
+	mockCrd                 *mocks.MockCustomResourceDefinitionInterface
```
wallyworld: As per other comment, mockCustomResourcDefinition etc

caas/kubernetes/provider/k8s.go
```diff
@@ -66,6 +68,8 @@ type kubernetesClient struct {
 	// namespace is the k8s namespace to use when
 	// creating k8s resources.
 	namespace string
+
+	apiextensionsClient apiextensionsclientset.Interface
```
wallyworld: Can we move this up next to kubernetes.Interface
Not sure if we want to embed it or not?

caas/kubernetes/provider/k8s.go
```diff
@@ -77,19 +81,23 @@ type kubernetesClient struct {
...
-	client, err := newClient(config)
+	k8sClient, apiextensionsClient, err := newClient(config)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	return &kubernetesClient{Interface: client, namespace: namespace}, nil
+	return &kubernetesClient{
+		Interface:           k8sClient,
+		namespace:           namespace,
+		apiextensionsClient: apiextensionsClient,
```
wallyworld: move up

caas/kubernetes/provider/k8s.go
```diff
@@ -503,6 +511,50 @@ func (k *kubernetesClient) DeleteService(appName string) (err error) {
 	return errors.Trace(k.deleteDeployment(appName))
 }
 
+func (k *kubernetesClient) EnsureCrd(appName string, podSpec *caas.PodSpec) error {
```
wallyworld: Comment.
Plus name change I think

caas/kubernetes/provider/k8s.go
```diff
@@ -503,6 +511,50 @@ func (k *kubernetesClient) DeleteService(appName string) (err error) {
...
+			Validation: &apiextensionsv1beta1.CustomResourceValidation{
+				OpenAPIV3Schema: &apiextensionsv1beta1.JSONSchemaProps{
+					Properties: t.Validation.Properties,
+				},
+			},
+		},
+	}
+	apiextensionsV1beta1 := k.apiextensionsClient.ApiextensionsV1beta1()
+	crd, err = apiextensionsV1beta1.CustomResourceDefinitions().Create(crdIn)
+	if k8serrors.IsAlreadyExists(err) {
```
wallyworld: In other places we do this the other way around - attempt to Update() and if we get a k8sNotFound error, we then create.

caas/kubernetes/provider/provider.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/juju/juju/environs"
 	"github.com/juju/juju/environs/config"
 	"github.com/juju/juju/environs/context"
+	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
```
wallyworld: move me

worker/caasunitprovisioner/deployment_worker.go
```diff
@@ -146,7 +146,13 @@ func (w *deploymentWorker) loop() error {
 		if err != nil {
 			return errors.Annotate(err, "cannot parse pod spec")
 		}
-
+		if len(spec.CustomResourceDefinitions) > 0 {
```
wallyworld: This should be internal to EnsureService().
The interface we expose on the caas client reflects generic juju-like semantics (service, units etc) and doesn't expose substrate specific functionality. The idea is that we may at some point support something other than k8s (but we probably never will).

caas/containers.go
```diff
@@ -63,33 +63,33 @@ type PodSpec struct {
...
+	Version    string                             `yaml:"version"`
+	Scope      string                             `yaml:"scope"`
+	Validation CustomResourceDefinitionValidation `yaml:"validation,omitempty"`
 }
 
 // Validate returns an error if the crd is not valid.
 func (crd *CustomResourceDefinition) Validate() error {
 	if crd.Kind == "" {
-		return errors.New("kind is missing")
+		return errors.NotValidf("kind is missing")
```
wallyworld: The Error() string from NotValid errors comes out like this "kind is missing is not valid"
So you just need error.NotValidf("missing Kind")

caas/kubernetes/provider/k8s.go
```diff
@@ -511,9 +510,9 @@ func (k *kubernetesClient) DeleteService(appName string) (err error) {
 	return errors.Trace(k.deleteDeployment(appName))
 }
 
-func (k *kubernetesClient) EnsureCrd(appName string, podSpec *caas.PodSpec) error {
+func (k *kubernetesClient) EnsureCustomResourceDefinition(appName string, podSpec *caas.PodSpec) error {
```
wallyworld: Needs comment

caas/kubernetes/provider/k8s.go
```diff
@@ -547,10 +547,11 @@ func (k *kubernetesClient) ensureCrdTemplate(t *caas.CustomResourceDefinition) (
 		},
 	}
 	apiextensionsV1beta1 := k.apiextensionsClient.ApiextensionsV1beta1()
-	crd, err = apiextensionsV1beta1.CustomResourceDefinitions().Create(crdIn)
-	if k8serrors.IsAlreadyExists(err) {
-		logger.Infof("updating crd %#v", crdIn)
-		crd, err = apiextensionsV1beta1.CustomResourceDefinitions().Update(crdIn)
+	crd, err = apiextensionsV1beta1.CustomResourceDefinitions().Update(crdIn)
+	logger.Debugf("try to update crd %#v", crdIn)
```
wallyworld: Move this log line up above the Update() call

worker/caasunitprovisioner/deployment_worker.go
wallyworld: As discussed on IRC, charms could define custom resources without pods.
We need to allow this but also need to re-think how we model things

caas/kubernetes/provider/provider_test.go
```diff
@@ -160,6 +161,31 @@ containers:
           file1: |
             [config]
             foo: bar
+crd:
```
wallyworld: Also need to fix this

worker/caasunitprovisioner/deployment_worker.go
ycliuhw: yes, agreed.

ycliuhw:
$$merge$$

PR #9078, opened 2018-08-17
provider/oci: Fix data race in terminateInstances

babbageclunk:
## Description of change
`terminateInstances` was appending to the errs and instIds slices from different goroutines in a waitgroup, so `environSuite.TestStopInstancesMultipleFail` would sometimes fail because one of the errors was dropped. I've changed it to send ids and errors down one buffered channel (in a struct so the right id is associated with each error), and then read from the channel after the waitgroup is finished.

## QA steps

Running `environSuite.TestStopInstancesMultipleFail` under the race detector no longer complains, and the test still passes.

### DIFF
--- a/provider/oci/environ.go
+++ b/provider/oci/environ.go
@@ -635,29 +635,40 @@ func (e *Environ) StopInstances(ctx envcontext.ProviderCallContext, ids ...insta
 	return nil
 }
 
+type instError struct {
+	id  instance.Id
+	err error
+}
+
 func (o *Environ) terminateInstances(instances ...*ociInstance) error {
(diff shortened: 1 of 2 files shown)


provider/oci/environ.go
```diff
@@ -635,29 +635,40 @@ func (e *Environ) StopInstances(ctx envcontext.ProviderCallContext, ids ...insta
 	return nil
 }
 
+type instError struct {
```
anastasiamac: I like this a lot :D

babbageclunk:
The OCI implementation in 2.4 is just a stub, so we don't need to backport this.

$$merge$$

PR #9081, opened 2018-08-20
Initial cloud container status storage

Veebers:
## Description of change

Initial plumbing to allow Caas charms to expose status information from within a pod

### DIFF
--- a/state/application.go
+++ b/state/application.go
@@ -2545,6 +2545,17 @@ func (op *AddUnitOperation) Done(err error) error {
 		}
 	}
 	if op.props.CloudContainerStatus != nil {
+		doc := statusDoc{
+			Status:     op.props.CloudContainerStatus.Status,
+			StatusInfo: op.props.CloudContainerStatus.Message,
+			StatusData: mgoutils.EscapeKeys(op.props.CloudContainerStatus.Data),
+			Updated:    timeOrNow(op.props.CloudContainerStatus.Since, u.st.clock()).UnixNano(),
+		}
+		_, err := probablyUpdateStatusHistory(op.application.st.db(), globalCloudContainerKey(op.unitName), doc)
+		if err != nil {
+			return errors.Trace(err)
+		}
+
 		// Ensure unit history is updated correctly
 		unitStatus, err := getStatus(op.application.st.db(), unitGlobalKey(op.unitName), "unit")
 		if err != nil {
--- a/state/application.go
+++ b/state/application.go
@@ -30,6 +30,7 @@ import (
 	"github.com/juju/juju/core/application"
 	"github.com/juju/juju/core/leadership"
 	"github.com/juju/juju/core/status"
+	mgoutils "github.com/juju/juju/mongo/utils"
 	"github.com/juju/juju/network"
 	"github.com/juju/juju/state/presence"
 	"github.com/juju/juju/tools"
@@ -1738,6 +1739,7 @@ func (a *Application) removeUnitOps(u *Unit, asserts bson.D) ([]txn.Op, error) {
 		removeMeterStatusOp(a.st, u.globalMeterStatusKey()),
 		removeStatusOp(a.st, u.globalAgentKey()),
 		removeStatusOp(a.st, u.globalKey()),
+		removeStatusOp(a.st, u.globalCloudContainerKey()),
 		removeConstraintsOp(u.globalAgentKey()),
 		annotationRemoveOp(a.st, u.globalKey()),
 		newCleanupOp(cleanupRemovedUnit, u.doc.Name),
@@ -2391,11 +2393,12 @@ func (a *Application) PasswordValid(password string) bool {
 // UnitUpdateProperties holds information used to update
 // the state model for the unit.
 type UnitUpdateProperties struct {
-	ProviderId  *string
-	Address     *string
-	Ports       *[]string
-	AgentStatus *status.StatusInfo
-	UnitStatus  *status.StatusInfo
+	ProviderId           *string
+	Address              *string
+	Ports                *[]string
+	AgentStatus          *status.StatusInfo
+	UnitStatus           *status.StatusInfo
+	CloudContainerStatus *status.StatusInfo
 }
 
 // UpdateUnits applies the given application unit update operations.
@@ -2492,8 +2495,22 @@ func (op *AddUnitOperation) Build(attempt int) ([]txn.Op, error) {
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
+
 	op.unitName = name
 	ops = append(ops, addOps...)
+
+	if op.props.CloudContainerStatus != nil {
+		now := op.application.st.clock().Now()
+		doc := statusDoc{
+			Status:     op.props.CloudContainerStatus.Status,
+			StatusInfo: op.props.CloudContainerStatus.Message,
+			StatusData: mgoutils.EscapeKeys(op.props.CloudContainerStatus.Data),
+			Updated:    now.UnixNano(),
+		}
+
+		newStatusOps := createStatusOp(op.application.st, globalCloudContainerKey(name), doc)
+		ops = append(ops, newStatusOps)
+	}
 
 	return ops, nil
 }
@@ -2503,7 +2520,7 @@ func (op *AddUnitOperation) Done(err error) error {
 	if err != nil {
 		return errors.Annotatef(err, "adding unit to %q", op.application.Name())
 	}
-	if op.props.AgentStatus == nil && op.props.UnitStatus == nil {
+	if op.props.AgentStatus == nil && op.props.CloudContainerStatus == nil {
 		return nil
 	}
 	// We do a separate status update here because we require all units to be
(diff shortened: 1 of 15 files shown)


apiserver/params/params.go
```diff
@@ -563,14 +564,15 @@ type UpdateApplicationUnits struct {
...
-	Data           map[string]interface{}     `json:"data,omitempty"`
+	ProviderId      string                     `json:"provider-id"`
+	UnitTag         string                     `json:"unit-tag"`
+	Address         string                     `json:"address"`
+	Ports           []string                   `json:"ports"`
+	FilesystemInfo  []KubernetesFilesystemInfo `json:"filesystem-info,omitempty"`
+	Status          string                     `json:"status"`
+	Info            string                     `json:"info"`
+	Data            map[string]interface{}     `json:"data,omitempty"`
+	ContainerStatus status.StatusInfo          `json:"container-status"`
```
wallyworld: params structs don't use non params things like status.StatusInfo
plus, the Status, Info, Data values are what will be container status

state/application.go
```diff
@@ -2379,11 +2379,12 @@ func (a *Application) PasswordValid(password string) bool {
...
-	Address     *string
-	Ports       *[]string
-	AgentStatus *status.StatusInfo
-	UnitStatus  *status.StatusInfo
+	ProviderId           *string
+	Address              *string
+	Ports                *[]string
+	AgentStatus          *status.StatusInfo
+	UnitStatus           *status.StatusInfo
+	CloudContainerStatus *status.StatusInfo
```
wallyworld: UnitStatus -> CloudContainerStatus

state/status.go
```diff
@@ -157,7 +157,25 @@ func (m *ModelStatus) UnitWorkload(unitName string) (status.StatusInfo, error) {
...
 
-	return m.getStatus(unitGlobalKey(unitName), "workload")
+	// Use cloud container status over unit if the cloud container status is
+	// error or active or the unit status hasn't shifted from 'allocating'
+	info, err = m.getStatus(unitGlobalKey(unitName), "workload")
+	if err != nil {
+		return info, err
+	}
+
+	containerInfo, err := m.getStatus(globalCloudContainerKey(unitName), "workload")
```
wallyworld: only want to do this for caas models

state/status.go
```diff
@@ -157,7 +157,39 @@ func (m *ModelStatus) UnitWorkload(unitName string) (status.StatusInfo, error) {
...
+		return info, nil
+	}
+	if err != nil {
+		return info, err
+	}
+
+	if info.Status == status.Allocating || info.Status == status.Active {
+		switch containerInfo.Status {
+		case status.Error:
+			fallthrough
```
wallyworld: Don't need the fallthroughs, can use a comma separated list

state/unit.go
```diff
@@ -1121,15 +1127,13 @@ func (u *Unit) Status() (status.StatusInfo, error) {
...
 	if info.Status != status.Error {
 		info, err = getStatus(u.st.db(), u.globalKey(), "unit")
 		if err != nil {
-			// CAAS units don't have any unit status.
-			if errors.IsNotFound(err) {
-				return status.StatusInfo{}, nil
-			}
 			return status.StatusInfo{}, err
 		}
+		return info, nil
```
wallyworld: don't need this

state/status.go
```diff
@@ -157,7 +157,39 @@ func (m *ModelStatus) UnitWorkload(unitName string) (status.StatusInfo, error) {
 		return info, nil
 	}
 
-	return m.getStatus(unitGlobalKey(unitName), "workload")
+	// (for CAAS models) Use cloud container status over unit if the cloud
```
wallyworld: This needs to be tested; I think the tests for this stuff may sadly be in cmd/juju/status
(would need to check)

state/migration_export.go
```diff
@@ -803,6 +803,12 @@ func (e *exporter) addApplication(ctx addApplicationContext) error {
 		workloadVersionKey := unit.globalWorkloadVersionKey()
 		exUnit.SetWorkloadVersionHistory(e.statusHistoryArgs(workloadVersionKey))
 
+		globalCCKey := unit.globalCloudContainerKey()
```
wallyworld: We were going to delete this right?
It needs to be done when juju/description is updated

Veebers:
!!build!!

state/unit.go
```diff
@@ -380,22 +386,72 @@ func (op *UpdateUnitOperation) Build(attempt int) ([]txn.Op, error) {
...
 		}
+		unitStatus = *op.props.UnitStatus
+	}
+	if op.props.CloudContainerStatus != nil {
+		if err := updateStatus(op.unit.globalCloudContainerKey(), op.props.CloudContainerStatus); err != nil {
+			return nil, errors.Trace(err)
+		}
+		cloudContainerStatus = *op.props.CloudContainerStatus
+	}
+	if unitStatus.Status != "" || cloudContainerStatus.Status != "" {
```
wallyworld: I think with the rework, the only unit status we can get here is Terminated. Right?
Which means we don't need to bother reloading the container status. Terminated wins.

state/unit.go
```diff
@@ -1143,13 +1195,24 @@ func (u *Unit) SetStatus(unitStatus status.StatusInfo) error {
 	if !status.ValidWorkloadStatus(unitStatus.Status) {
 		return errors.Errorf("cannot set invalid status %q", unitStatus.Status)
 	}
+
+	var newHistory *statusDoc
+	if u.modelType == ModelTypeCAAS {
```
wallyworld: Add a comment here as to why we need this

state/status.go
```diff
@@ -302,7 +378,28 @@ func setStatus(db Database, params setStatusParams) (err error) {
 		Updated:    params.updated.UnixNano(),
 	}
 
-	newStatus, historyErr := probablyUpdateStatusHistory(db, params.globalKey, doc)
+	historyDoc := &doc
+	var historyOverwriteStatusExists bool
+	if params.historyOverwrite != nil {
+		historyDoc = params.historyOverwrite
+		// As we redirect history it redirects status. Need to check if
```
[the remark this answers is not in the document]
Veebers: I can and will expand the comment;

probablyUpdateStatusHistory returns if it's a new history record and the status write is skipped if it's not new (i.e. writing the same thing again).
If we have a unit status and an history overwrite status, if the history overwrite is the same as last time, the potentially new actual unit status will never be written, i.e.: charm sets status 'active', we will overwrite history with 'waiting for container' if there is no cloud container status meaning we lose the charm setting active as it'll never be stored anywhere.

state/application.go
```diff
@@ -2503,7 +2527,7 @@ func (op *AddUnitOperation) Done(err error) error {
 	if err != nil {
 		return errors.Annotatef(err, "adding unit to %q", op.application.Name())
 	}
-	if op.props.AgentStatus == nil && op.props.UnitStatus == nil {
+	if op.props.AgentStatus == nil && op.props.UnitStatus == nil && op.props.CloudContainerStatus == nil {
```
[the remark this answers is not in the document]
Veebers: Correct.

state/application.go
```diff
@@ -2527,14 +2551,34 @@ func (op *AddUnitOperation) Done(err error) error {
 			return errors.Trace(err)
 		}
 	}
+	unitStatusInfo := status.StatusInfo{
```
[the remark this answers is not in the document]
Veebers: Ah, right you are. Also, addUnitOps earlier in Build takes care of priming the status with waiting for container etc.

state/application.go
```diff
@@ -2527,17 +2544,29 @@ func (op *AddUnitOperation) Done(err error) error {
...
 	}
-	if op.props.UnitStatus != nil {
-		now := op.application.st.clock().Now()
-		if err := u.SetStatus(status.StatusInfo{
-			Status:  op.props.UnitStatus.Status,
-			Message: op.props.UnitStatus.Message,
-			Data:    op.props.UnitStatus.Data,
-			Since:   &now,
-		}); err != nil {
+	if op.props.CloudContainerStatus != nil {
```
wallyworld: It seems we're failing to update container status history

Veebers: Ah good catch as the createStatusOps doesn't handle the history.

wallyworld: Yep, and the tests don't check for it either. They only load and test unit status history

state/status.go
```diff
@@ -157,7 +157,75 @@ func (m *ModelStatus) UnitWorkload(unitName string) (status.StatusInfo, error) {
...
+
+	containerInfo, err := m.getStatus(globalCloudContainerKey(unitName), "cloud container")
+	if err != nil && !errors.IsNotFound(err) {
+		return info, err
+	}
+	return caasUnitDisplayStatus(info, containerInfo), nil
+}
+
+func caasUnitDisplayStatus(unitStatus status.StatusInfo, containerStatus status.StatusInfo) status.StatusInfo {
+	if unitStatus.Status == status.Terminated {
```
wallyworld: container status can also be terminated right?
So it should be ||

Veebers:
$$merge$$

Veebers:
$$merge$$

Veebers:
$$merge$$

PR #9083, opened 2018-08-20
2.3 txn prune batches 1771906

jameinel:
## Description of change

Brings in the latest txn pruner changes (to only prune in batches), and provides controller config knobs to allow for them to be tuned live, in case 1M batches is not ideal.

## QA steps

```
$ juju bootstrap lxd --debug
# Post bootstrap, you should be able to set the config
$ juju controller-config max-prune-txn-batch-size=10000
# Deploy a lot of applications, so we end up with lots of lease txn churn
$ for k in `seq 0 2` do; for j in `seq 0 9`; do echo $j; for i in `seq 0 9`; do juju deploy cs:~jameinel/ubuntu-lite u$k$j$i --to 0 & done; time wait ; done ; done
# Watch the log for prune events:
$ juju debug-log -m controller --include-module juju.txn --replay
```

It takes 1hr before the first prune event, which shouldn't do much because we don't clean txns that are <1hr old. The second one should end with something like:
```
machine-0: 11:40:19 INFO juju.txn txn pruning complete after 6.766794665s in 5 batches. txns now: 52058, inspected 336 collections, 9420 docs (0 cleaned)
   removed 0 stash docs and 52058 txn docs
```

It should take 5 batches of 10,000 to clean out 50k transactions. You can also trap for max passes with:
```
$ juju controller-config max-prune-txn-batch-size=1000 max-prune-txn-passes=10
```

## Documentation changes

@pmatulis, we probably need to document the new controller config entries well. (`max-prune-txn-batch-size` and `max-prune-txn-passes`)

## Bug reference

https://bugs.launchpad.net/juju/+bug/1771906

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@ TAGS
 parts/
 prime/
 stage/
+vendor/
+_vendor/
 *.snap
 .emacs.desktop
 .emacs.desktop.lock
--- a/cmd/juju/backups/restore_test.go
+++ b/cmd/juju/backups/restore_test.go
@@ -263,7 +263,7 @@ func (s *restoreSuite) TestRestoreReboostrapWritesUpdatedControllerInfo(c *gc.C)
 	})
 }
 
-func (s *restoreSuite) TestRestoreReboostrapControllerConfigDefaults(c *gc.C) {
+func (s *restoreSuite) TestRestoreRebootstrapControllerConfigDefaults(c *gc.C) {
 	metadata := params.BackupsMetadataResult{
 		CACert:       testing.CACert,
 		CAPrivateKey: testing.CAKey,
@@ -292,6 +292,8 @@ func (s *restoreSuite) TestRestoreReboostrapControllerConfigDefaults(c *gc.C) {
 			"max-logs-age":              "72h",
 			"max-logs-size":             "4096M",
 			"max-txn-log-size":          "10M",
+			"max-prune-txn-batch-size":  1000000,
+			"max-prune-txn-passes":      100,
 			"auditing-enabled":          true,
 			"audit-log-capture-args":    false,
 			"audit-log-max-size":        "300M",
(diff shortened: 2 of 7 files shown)


jameinel:
[1771906 TXN Pruner batches in 1M chunks](https://trello.com/c/MSJXMdWS/453-1771906-txn-pruner-batches-in-1m-chunks)

jameinel:
$$merge$$

jameinel:
$$merge$$
PresenceSuite.TestWatchPeriod failed, but that appears to be a transient issue.

pmatulis:
@jameinel Sorry, I'll need more guidance here. First, what is "txn"? And what is the typical use case for each of these new variables?

jameinel:
@pmatulis First, some basic background.
As juju modifies the database, it creates a transaction record for all changes that are made. These records build up over time, so we have to clean them out periodically. It doesn't matter much for small Juju controllers, but for large ones (like prodstack) they can create as much as 200,000 records per hour.
By default we wake up every hour, and check if there are a significant number of new transactions created, and if so, we run a "cleanup and prune" pass across the database. Looking at all the documents to see if any of them mention transactions that have been completed, and then clearing out all transactions that are no longer referenced. Usually this keeps the number of transactions manageable.
We've had a number of issues in the past where due to a bug or corruption the list gets out of hand, and we fail to clean up transactions. (txn stands for transaction.)
In those cases, due to the high rate of new ones being created, we have controllers that have 10M transactions that need to be cleaned up.
What we believe we've observed is that they start failing to clean up, just because there are so many that are pending, that the queries to determine "what should be cleaned up" start failing.
So this changes the default behavior such that it caps the number of transactions we will evaluate at a time to 1M, and then if there are transactions remaining after running the cleanup, we run another cleanup. To avoid having this run forever, we default to only running 100 passes before we stop until the next hour rolls around.
Both of those values are configurable (max-prune-txn-batch-size and max-prune-txn-passes, respectively)
Mostly that was done because we weren't sure on the exact values that would be useful in production, and wanted to make sure we provided a knob in case we guessed them incorrectly. (Most likely, the defaults are sane and nobody needs to touch them, its mostly a just-in-case, but they should be documented.)

PR #9085, opened 2018-08-21
Ensure test storage charms have at least one hook

wallyworld:
## Description of change

Juju now refuses to install charms without at least one hook.
Fix the generated storage test charms to allow acceptance tests to pass.

## QA steps

Run assess_storage.py


wallyworld:
$$merge$$

PR #9086, opened 2018-08-21
Ignore the machine-lock files when looking for the machine log files.

howbazaar:
The action to gather the sizes of the machine logs needs to ignore the machine-lock log files.


howbazaar:
$$merge$$

PR #9094, opened 2018-08-22
Update upstream dependencies to bring in charm.v6 changes

wallyworld:
## Description of change

We need a newer charm.v6 to fix a charm archive path issue.
But there's other deps that need to come in also to satisfy the charm.v6 update. These deps add oci-image and devices support to charms (which are not used in 2.4).


wallyworld:
$$merge$$

PR #9096, opened 2018-08-22
Check cloud credential works for a model.

anastasiamac:
## Description of change

We need to be able to determine if a cloud credential is valid for a model. At this stage, this means "the cloud instances that this credential lists are the same that this model has as machine instances".

This check can be performed using the credential that a model already references. This section of PR is re-factored from migrationtarget facade. We re-factor it to be used both from migrationtarget as well as the update-credential-content functional path.
This check can also be performed using a credential that is not yet referenced by the model. This is the new logical path that this PR introduces. This will be used when we want to replace model credential, i.e. allow models to set a new credential.

### DIFF
--- a/apiserver/common/credentialcommon/backend.go
+++ b/apiserver/common/credentialcommon/backend.go
@@ -29,9 +29,9 @@ type Machine interface {
 	Id() string
 }
 
-// PersistedCloudEntitiesBackend defines what cloud entities where persisted in state
+// CloudEntitiesBackend defines what cloud entities where persisted in state
 // and will be accessed during the check.
-type PersistedCloudEntitiesBackend interface {
+type CloudEntitiesBackend interface {
 	// AllMachines returns all machines in the model.
 	AllMachines() ([]Machine, error)
 }
@@ -48,10 +48,10 @@ type Model interface {
 	Config() (*config.Config, error)
 }
 
-// PersistedModelBackend defines what model specific properties where persisted in state
+// ModelBackend defines what model specific properties where persisted in state
 // and will be accessed during the check.
-type PersistedModelBackend interface {
-	PersistedCloudEntitiesBackend
+type ModelBackend interface {
+	CloudEntitiesBackend
 
 	// Model returns the model entity.
 	Model() (Model, error)
@@ -64,8 +64,8 @@ type stateShim struct {
 	*state.State
 }
 
-// NewPersistedCloudEntitiesBackend creates a backend to use based on state.State.
-func NewPersistedCloudEntitiesBackend(p *state.State) PersistedCloudEntitiesBackend {
+// NewCloudEntitiesBackend creates a backend to use based on state.State.
+func NewCloudEntitiesBackend(p *state.State) CloudEntitiesBackend {
 	return stateShim{p}
 }
 
--- a/apiserver/common/credentialcommon/backend.go
+++ b/apiserver/common/credentialcommon/backend.go
@@ -1,0 +1,95 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package credentialcommon
+
+import (
+	"github.com/juju/errors"
+
+	"github.com/juju/juju/cloud"
+	"github.com/juju/juju/environs/config"
+	"github.com/juju/juju/environs/context"
+	"github.com/juju/juju/instance"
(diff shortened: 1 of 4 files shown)


anastasiamac:
[add credential validity check for a given model](https://trello.com/c/JKzqVvDJ/370-add-credential-validity-check-for-a-given-model)

apiserver/common/credentialcommon/modelcredential.go
```diff
@@ -0,0 +1,124 @@
...
+	return CheckMachineInstances(persisted, provider, callCtx)
+}
+
+// ValidateNewModelCredential checks if a new cloud credential can be valid
+// for a given model.
+// Note that this call does not validate credential against the cloud of the model.
+func ValidateNewModelCredential(backend PersistedModelBackend, newEnv NewEnvironFunc, callCtx context.ProviderCallContext, credential *cloud.Credential) (params.ErrorResults, error) {
+	model, err := backend.Model()
+	if err != nil {
+		return failCredentialValidation(errors.Trace(err))
```
babbageclunk: Can you rename `failCredentialValidation` to `fail` like you used in migrationtarget below? I think the current name is long enough that it obscures what's going on. It needs to be more convenient/readable than the alternative of `return empty, errors.Trace(err)`, otherwise it doesn't pay the cost of having to look the function up to find out what it does.

apiserver/common/credentialcommon/modelcredential_test.go
```diff
@@ -0,0 +1,348 @@
...
+	c.Assert(results, gc.DeepEquals, params.ErrorResults{})
+}
+
+func (s *ModelCredentialSuite) TestCheckMachinesErrorGettingInstances(c *gc.C) {
+	s.provider.allInstancesFunc = func(ctx context.ProviderCallContext) ([]instance.Instance, error) {
+		return nil, errors.New("kaboom")
+	}
+
+	results, err := credentialcommon.CheckMachineInstances(s.backend, s.provider, s.callContext)
+	c.Assert(err, gc.ErrorMatches, "kaboom")
```
babbageclunk: Ha, ok I did write this code.

I just didn't click that you were talking about this, because I didn't think of this as having anything to do with credentials.

apiserver/common/credentialcommon/modelcredential.go
```diff
@@ -0,0 +1,124 @@
...
+			continue
+		}
+		if manual, err := machine.IsManual(); err != nil {
+			return failCredentialValidation(errors.Trace(err))
+		} else if manual {
+			continue
+		}
+		instanceId, err := machine.InstanceId()
+		if err != nil {
+			// TODO (anastasiamac 2018-08-21) do we really want to fail all processing here or just an error result against this machine and keep going?
```
babbageclunk: (Not sure whether this is a note specifically to yourself, but I think the latter - from reading state.Machine.InstanceId it can return an error if the machine hasn't been provisioned yet.)

apiserver/common/credentialcommon/modelcredential.go
anastasiamac: Since this is package level variable, simple "fail" is not descriptive enough. However, I will move this within each func that needs it. This way, it'll be just fail()

apiserver/common/credentialcommon/backend.go
```diff
@@ -0,0 +1,95 @@
...
+	// machine, or a NotProvisionedError, if not set.
+	InstanceId() (instance.Id, error)
+
+	// Id returns the machine id.
+	Id() string
+}
+
+// PersistedCloudEntitiesBackend defines what cloud entities where persisted in state
+// and will be accessed during the check.
+type PersistedCloudEntitiesBackend interface {
```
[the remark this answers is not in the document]
anastasiamac: Good idea. Thanks, will do :D

apiserver/common/credentialcommon/modelcredential.go
anastasiamac: This was a note for reviewer :) I wanted to get a 2nd opinion. And since it looks like we both agree, I'll change this logic to keep going \o/

anastasiamac:
$$merge$$

anastasiamac:
unrelated intermittent failure in github.com/juju/juju/worker/lease.TestPackage

anastasiamac:
$$merge$$

PR #9100, opened 2018-08-22
WIP : Initial changes for openstack Invalidate Creds

vinu2003:
This PR has initial changes for the openstack provider Invalidate Credentials.


provider/openstack/cinder.go
```diff
@@ -217,7 +217,7 @@ var _ storage.VolumeSource = (*cinderVolumeSource)(nil)
 func (s *cinderVolumeSource) CreateVolumes(ctx context.ProviderCallContext, args []storage.VolumeParams) ([]storage.CreateVolumesResult, error) {
 	results := make([]storage.CreateVolumesResult, len(args))
 	for i, arg := range args {
-		volume, err := s.createVolume(arg)
+		volume, err := s.createVolume(arg, ctx)
```
wallyworld: Ideally ctx would be the first argument

provider/openstack/cinder.go
```diff
@@ -262,7 +262,7 @@ func (s *cinderVolumeSource) createVolume(arg storage.VolumeParams) (*storage.Vo
 		if err := s.storageAdapter.DeleteVolume(volumeId); err != nil {
 			logger.Warningf("destroying volume %s: %s", volumeId, err)
 		}
-		return nil, errors.Errorf("waiting for volume to be provisioned: %s", err)
+		return nil, HandleCredentialError(errors.Errorf("waiting for volume to be provisioned: %s", err), ctx)
```
wallyworld: As a drive by, change to error.Annotate(err, "waiting for volume to be provisioned")

here and anywhere else touched by the changes

provider/openstack/cinder.go
```diff
@@ -341,29 +341,29 @@ func (s *cinderVolumeSource) DescribeVolumes(ctx context.ProviderCallContext, vo
 
 // DestroyVolumes implements storage.VolumeSource.
 func (s *cinderVolumeSource) DestroyVolumes(ctx context.ProviderCallContext, volumeIds []string) ([]error, error) {
-	return foreachVolume(s.storageAdapter, volumeIds, destroyVolume), nil
+	return foreachVolume(s.storageAdapter, volumeIds, destroyVolume, ctx), nil
```
wallyworld: ctx first

provider/openstack/errors.go
```diff
@@ -0,0 +1,29 @@
...
+package openstack
+
+import (
+	gooseerrors "gopkg.in/goose.v2/errors"
+
+	"github.com/juju/juju/environs/context"
+	)
+
+func MaybeHandleCredentialError(err error, ctx context.ProviderCallContext) (error, bool) {
+	IsUnauthorized := gooseerrors.IsUnauthorised(err)
```
wallyworld: isUnauthorised

provider/openstack/cinder.go
```diff
@@ -217,8 +217,13 @@ var _ storage.VolumeSource = (*cinderVolumeSource)(nil)
 func (s *cinderVolumeSource) CreateVolumes(ctx context.ProviderCallContext, args []storage.VolumeParams) ([]storage.CreateVolumesResult, error) {
 	results := make([]storage.CreateVolumesResult, len(args))
 	for i, arg := range args {
-		volume, err := s.createVolume(arg)
+		volume, err := s.createVolume(ctx, arg)
 		if err != nil {
+			_, denied := MaybeHandleCredentialError(err, ctx)
+			if denied {
+				errors.Annotate(err, "cannot create volume.")
```
wallyworld: This code doesn't do anything. Take a look inside Annotate() - it returns an error and we are not using it anywhere here.

provider/openstack/errors.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	gooseerrors "gopkg.in/goose.v2/errors"
+
+	"github.com/juju/juju/environs/context"
+)
+
+func MaybeHandleCredentialError(err error, ctx context.ProviderCallContext) (error, bool) {
+	isUnauthorized := gooseerrors.IsUnauthorised(err)
+	//fmt.Printf("XXXXX... IsUnauthorized : %v\n", IsUnauthorized)
+	if ctx != nil && isUnauthorized {
```
wallyworld: If unauthorised, also need to wrap err with common.CredentialNotValid
See how the ec2 provider does it

provider/openstack/cinder.go
```diff
@@ -217,8 +217,13 @@ var _ storage.VolumeSource = (*cinderVolumeSource)(nil)
 func (s *cinderVolumeSource) CreateVolumes(ctx context.ProviderCallContext, args []storage.VolumeParams) ([]storage.CreateVolumesResult, error) {
 	results := make([]storage.CreateVolumesResult, len(args))
 	for i, arg := range args {
-		volume, err := s.createVolume(arg)
+		volume, err := s.createVolume(ctx, arg)
 		if err != nil {
+			_, denied := MaybeHandleCredentialError(err, ctx)
```
wallyworld: None of this code is needed - the createVolume() call itself already does the checking for invalid credential. As written, this code will avoid setting the result error if the credential is invalid which in turn will make the caller think the api has succeeded when it hasn't.

provider/openstack/cinder.go
```diff
@@ -453,8 +458,13 @@ func (s *cinderVolumeSource) ValidateVolumeParams(params storage.VolumeParams) e
 func (s *cinderVolumeSource) AttachVolumes(ctx context.ProviderCallContext, args []storage.VolumeAttachmentParams) ([]storage.AttachVolumesResult, error) {
 	results := make([]storage.AttachVolumesResult, len(args))
 	for i, arg := range args {
-		attachment, err := s.attachVolume(arg)
+		attachment, err := s.attachVolume(ctx, arg)
 		if err != nil {
+			_, denied := MaybeHandleCredentialError(err, ctx)
```
wallyworld: Same comment as above - I'm not sure what the intent here was.

provider/openstack/cinder.go
```diff
@@ -675,7 +686,12 @@ func (ga *openstackStorageAdapter) GetVolume(volumeId string) (*cinder.Volume, e
 
 // SetVolumeMetadata is part of the OpenstackStorage interface.
 func (ga *openstackStorageAdapter) SetVolumeMetadata(volumeId string, metadata map[string]string) (map[string]string, error) {
-	return ga.cinderClient.SetVolumeMetadata(volumeId, metadata)
+	response, err := ga.cinderClient.SetVolumeMetadata(volumeId, metadata)
```
wallyworld: Was the intent here to do an invalid credential check on the error?

provider/openstack/errors.go
```diff
@@ -0,0 +1,29 @@
...
+		invalidateErr := ctx.InvalidateCredential("openstack cloud denied access")
+		if invalidateErr != nil {
+			logger.Warningf("could not invalidate stored openstack cloud credential on the controller: %v", invalidateErr)
+		}
+	}
+	return err, isUnauthorized
```
wallyworld: Do we actually need to use the isAuthorised bool anywhere?
We do in maas but why do we need it here?
Would a better approach be to use what's done with ec2 provider with the maybeConvertCredentialError() method?

PR #9102, opened 2018-08-23
Fix assess container networking functional test.

Veebers:
## Description of change

Update the network container functional test to handle the header text nc puts in the file.
Also change how the reboots are instrumented (previously the reboot would immediately terminate the ssh session raising a CommandProcessError exception).


hmlanigan:
LGTM

PR #9103, opened 2018-08-23
Better recovery from multi-controller cmr errors

wallyworld:
## Description of change

If a cross model relation is created between controllers and the offering controller is offline, the consumer can't create the relation which is expected. But when the offering controller comes back online, the processing doesn't resume. This is because a key watcher is not restarted.
This PR moves the watcher creation to the worker which uses it which fixes the issue.

Also add some logging so that child worker errors are exposed.

## QA steps

Create an offer and firewall off the controller
bootstrap another controller and relate to the offer
restore connectivity and the relation should be created

## Bug reference

https://bugs.launchpad.net/juju/2.3/+bug/1788554

### DIFF
--- a/worker/remoterelations/relationunitsworker.go
+++ b/worker/remoterelations/relationunitsworker.go
@@ -69,7 +69,11 @@ func (w *relationUnitsWorker) Kill() {
 
 // Wait is defined on worker.Worker
 func (w *relationUnitsWorker) Wait() error {
-	return w.catacomb.Wait()
+	err := w.catacomb.Wait()
+	if err != nil {
+		logger.Errorf("error in relation units worker for %v: %v", w.relationTag.Id(), err)
+	}
+	return err
 }
 
 func (w *relationUnitsWorker) loop() error {
--- a/worker/remoterelations/remoteapplicationworker.go
+++ b/worker/remoterelations/remoteapplicationworker.go
@@ -21,8 +21,7 @@ import (
 // changes originating from the offering model and consumes those
 // in the local model.
 type remoteApplicationWorker struct {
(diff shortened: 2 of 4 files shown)


worker/remoterelations/relationunitsworker.go
```diff
@@ -69,7 +69,9 @@ func (w *relationUnitsWorker) Kill() {
 
 // Wait is defined on worker.Worker
 func (w *relationUnitsWorker) Wait() error {
-	return w.catacomb.Wait()
+	err := w.catacomb.Wait()
+	logger.Errorf("error in relation units worker for %v: %v", w.relationTag.Id(), err)
```
manadart: Looks like this will be emitted even when error is nil...

worker/remoterelations/remoteapplicationworker.go
```diff
@@ -69,7 +68,9 @@ func (w *remoteApplicationWorker) Kill() {
 
 // Wait is defined on worker.Worker
 func (w *remoteApplicationWorker) Wait() error {
-	return w.catacomb.Wait()
+	err := w.catacomb.Wait()
+	logger.Errorf("error in remote application worker for %v: %v", w.applicationName, err)
```
manadart: Same here.

worker/remoterelations/relationunitsworker.go
wallyworld: yup, silly me, fixed

wallyworld:
$$merge$$

wallyworld:
[cmr error recovery](https://trello.com/c/vAgjygMn/408-cmr-error-recovery)

PR #9106, opened 2018-08-23
Remove lock and removal related cleanups.

ExternalReality:
Lock should be cleaned up after all is said and done.

### DIFF
--- a/api/upgradeseries/upgradeseries.go
+++ b/api/upgradeseries/upgradeseries.go
@@ -148,3 +148,19 @@ func (s *Client) StartUnitCompletion() error {
 	}
 	return results.Results[0].Error
 }
+
+// FinishUpgradeSeries starts the complete phase for all subordinate units.
+func (s *Client) FinishUpgradeSeries() error {
+	var results params.ErrorResults
+	args := params.Entities{
+		Entities: []params.Entity{{Tag: s.authTag.String()}},
+	}
+	err := s.facade.FacadeCall("FinishUpgradeSeries", args, &results)
+	if err != nil {
+		return err
+	}
+	if len(results.Results) != 1 {
+		return errors.Errorf("expected 1 result, got %d", len(results.Results))
+	}
+	return results.Results[0].Error
+}
--- a/api/upgradeseries/upgradeseries.go
+++ b/api/upgradeseries/upgradeseries.go
@@ -149,7 +149,9 @@ func (s *Client) StartUnitCompletion() error {
 	return results.Results[0].Error
 }
 
-// FinishUpgradeSeries starts the complete phase for all subordinate units.
+// FinishUpgradeSeries notifies the controller that the upgrade process is
+// completely finished. We use the name "Finish" to distinguish this method from
+// the various "Complete" phases.
 func (s *Client) FinishUpgradeSeries() error {
 	var results params.ErrorResults
 	args := params.Entities{
--- a/apiserver/facades/client/machinemanager/machinemanager.go
+++ b/apiserver/facades/client/machinemanager/machinemanager.go
@@ -387,13 +387,6 @@ func (mm *MachineManagerAPI) updateSeriesPrepare(arg params.UpdateSeriesArg) err
 			}
 		}
 	}()
-
-	// TODO (hml) 2018-06-26 managed series upgrade
-	// Next steps in updateSeriesPrepare:
-	// 1. verify success of pre-upgrade-series hook run on each unit
-	// 2. stop each unit's agent on machine
-	// 3. write systemd unit agent files if necessary and copy tools based on series.
-
 	return nil
 }
 
@@ -465,9 +458,6 @@ func (mm *MachineManagerAPI) completeUpgradeSeries(arg params.UpdateSeriesArg) e
 	return machine.CompleteUpgradeSeries()
 }
 
-// [TODO](externalreality) We still need this, eventually the lock is going to cleaned up
-// RemoveUpgradeSeriesLock removes a series upgrade prepare lock for a
-// given machine.
 func (mm *MachineManagerAPI) removeUpgradeSeriesLock(arg params.UpdateSeriesArg) error {
 	machineTag, err := names.ParseMachineTag(arg.Entity.Tag)
 	if err != nil {
--- a/state/machine_upgradeseries.go
+++ b/state/machine_upgradeseries.go
@@ -260,7 +260,7 @@ func (m *Machine) RemoveUpgradeSeriesLock() error {
 			return nil, errors.Trace(err)
 		}
 		if !locked {
-			return nil, errors.NotFoundf("upgrade series lock for machine %q", m)
+			return nil, jujutxn.ErrNoOperations
 		}
 		return removeUpgradeSeriesLockTxnOps(m.doc.Id), nil
 	}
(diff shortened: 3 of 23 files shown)


ExternalReality:
[remove machine lock file once complete runs successfully](https://trello.com/c/3xPYeZDA/505-remove-machine-lock-file-once-complete-runs-successfully)

state/machine_upgradeseries_test.go
```diff
@@ -85,6 +85,15 @@ func (s *MachineSuite) TestRemoveUpgradeSeriesLockUnlocksMachine(c *gc.C) {
 	AssertMachineIsNOTLockedForPrepare(c, mach)
 }
 
+func (s *MachineSuite) TestRemoveUpgradeSeriesLockIsNoOpIfMachineIsNotLocked(c *gc.C) {
```
ExternalReality: This is needed since Removal of the lock may be prompted by an error which triggers cleanup. The error my be an error in the creation of the lock itself which would result in an attempt to remove a non-existent lock. An error which triggers cleanup can result in the attempted remove a non-existent lock, for this an other reasons we do not error when attempting to remove a non-existent lock.

state/machine_upgradeseries.go
```diff
@@ -260,7 +260,7 @@ func (m *Machine) RemoveUpgradeSeriesLock() error {
 			return nil, errors.Trace(err)
 		}
 		if !locked {
-			return nil, errors.NotFoundf("upgrade series lock for machine %q", m)
+			return nil, jujutxn.ErrNoOperations
```
ExternalReality: Remove of lock is idempotent and as such does not throw an error when a lock is not found.

mitechie:
I went to QA this and hit the following: https://pastebin.canonical.com/p/MxTWhbS6tW/

It looks like there's some panic getting tossed that didn't complete prepare and shutting down the unit agent.

manadart:
@mitechie That panic should be resolved in the develop HEAD due to the landing of https://github.com/juju/juju/pull/9105.

api/upgradeseries/upgradeseries.go
```diff
@@ -144,3 +144,19 @@ func (s *Client) StartUnitCompletion() error {
 	}
 	return results.Results[0].Error
 }
+
+// FinishUpgradeSeries starts the complete phase for all subordinate units.
+func (s *Client) FinishUpgradeSeries() error {
```
manadart: Maybe "CompleteUpgradeSeries" so all the verbiage is congruent?

worker/upgradeseries/worker.go
```diff
@@ -268,6 +268,18 @@ func (w *upgradeSeriesWorker) handleCompleteStarted() error {
 	return nil
 }
 
+// handleCompleted givens the worker a chance to perform any final operations
+// before returning control to the server (controller) which will then perform its own
+// post upgrade routine.
+func (w *upgradeSeriesWorker) handleCompleted() error {
```
manadart: Nothing calls this at the moment.

The way the other `handle{Thing}` members are laid out, is that they are invoked based on the machine status **only**. Inside the handlers there are calls the API to check unit statuses.

So I imagine this would be called from another case and look something like:

```go
	units, allConfirmed, err = w.compareUnitAgentServices(w.UnitsCompleted)
	if err != nil {
		return errors.Trace(err)
	}
	if allConfirmed {
		w.logger.Infof("series upgrade complete")
		return errors.Trace(w.FinishUpgradeSeries())
	}
```

mitechie:
Thanks for the heads up on the fix in trunk. I'll update and re-QA.

manadart:
@mitechie There is another issue I am having. Not sure when it was introduced - the worker manifold does not appear to be starting, so the unit goes through its prepare workflow, but nothing else can be progressed.

worker/upgradeseries/worker.go
manadart: Yes, we don't need to check the unit status, because we know machine could not be complete without that.

apiserver/facades/client/machinemanager/machinemanager.go
```diff
@@ -465,9 +458,6 @@ func (mm *MachineManagerAPI) completeUpgradeSeries(arg params.UpdateSeriesArg) e
 	return machine.CompleteUpgradeSeries()
 }
 
-// [TODO](externalreality) We still need this, eventually the lock is going to cleaned up
-// RemoveUpgradeSeriesLock removes a series upgrade prepare lock for a
-// given machine.
 func (mm *MachineManagerAPI) removeUpgradeSeriesLock(arg params.UpdateSeriesArg) error {
```
manadart: Do we need this at all? Doesn't look like it is called by anything.

ExternalReality: We need this because the CLI cleans up upon failure using this method. However, the way the CLI cleans seems as if it will never hit a point where it can fail after creating a lock. That may be changing really soon though.

api/upgradeseries/upgradeseries.go
ExternalReality: I chose finish upgrade series because the word complete is getting somewhat overused and confusing.

worker/uniter/operation/interface.go
```diff
@@ -114,6 +118,9 @@ type Factory interface {
 	// NewResignLeadership creates an operation to ensure the uniter does not
 	// act as application leader.
 	NewResignLeadership() (Operation, error)
+
+	// NewCompleteUpgrade create an operation which resets the local state of the uniter.
```
manadart: "NewCompleteUpgrade creates..."

mitechie:
I pulled the latest develop after confirming that it was back to working as it was end to end. I then re-qa'd this branch on top of that. My expectation was that with this branch cleaning up the lock I'd be able to prepare/complete multiple times. I was not able to do so. I did not get the post-hook execution as well. 

Some notes from my QA:
https://pastebin.canonical.com/p/3X8hGGvQhH/

worker/uniter/resolver/opfactory.go
```diff
@@ -122,20 +135,20 @@ func (s *resolverOpFactory) wrapHookOp(op operation.Operation, info hook.Info) o
 		op = onPrepareWrapper{op, func() {
 			//on prepare the local status should be made to reflect
 			//that the upgrade process for this united has started.
-			s.LocalState.UpgradeSeriesPrepareStatus = s.RemoteState.UpgradeSeriesPrepareStatus
+			s.LocalState.UpgradeSeriesStatus = s.RemoteState.UpgradeSeriesStatus
```
manadart: I would declare this anonymous function as a variable and DRY these op declarations up.

worker/uniter/resolver/opfactory.go
```diff
@@ -48,6 +49,18 @@ func (s *resolverOpFactory) NewSkipHook(info hook.Info) (operation.Operation, er
 	return s.wrapHookOp(op, info), nil
 }
 
+func (s *resolverOpFactory) NewCompleteUpgrade() (operation.Operation, error) {
```
manadart: As discussed, we will need a test for this new piece of behaviour.

mitechie:
QA works like a charm! Thanks for getting the update together here and the ability to rerun is awesome.

worker/uniter/resolver/opfactory.go
ExternalReality: Given that these just changed, I don't know if it would be worth extracting now. They no longer have text in common.

jameinel:
Looks like the test suite hung in:
github.com/juju/juju/featuretests.(*upgradeSuite).TestLoginsDuringUpgrade.func3(0xc42065b860, 0xc4212851d0)


Which doesn't specifically feel like a change here. Will try again

$$merge$$

PR #9108, opened 2018-08-24
Improve k8s storage class lookup

wallyworld:
## Description of change

When the k8s provisioner looks up what storage class to use, we now supply a fallback class which will be used if present, but won't result in a not found error if it doesn't exist. This is so that we can properly allow operator storage to be configured.

## QA steps

deploy a k8s model
deploy a charm and ensure that the operator comes up but with no storage used
set up a storage class called juju-operator-volume
deploy a charm and ensure that the operator storage is created

### DIFF
--- a/caas/kubernetes/provider/export_test.go
+++ b/caas/kubernetes/provider/export_test.go
@@ -36,6 +36,10 @@ func StorageClass(cfg *storageConfig) string {
 	return cfg.storageClass
 }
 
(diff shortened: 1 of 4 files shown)


ycliuhw:
This looks great to me! 
Thanks.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #9113, opened 2018-08-24
fix lp:1787605, lxd provider Ping() should fail with http addr

hmlanigan:
## Description of change

Fix lp:1787605, for the lxd provider, add-cloud should fail when a http address is provided by the user.  This means that the lxd provider Ping() should only succeed with https endpoints.

## QA steps

Run 'juju add-cloud' interactively for a remote lxd cloud.  First try using http, which should fail, then https should succeed.

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1787605

### DIFF
--- a/provider/lxd/provider.go
+++ b/provider/lxd/provider.go
@@ -165,18 +165,22 @@ func (p *environProvider) Ping(ctx context.ProviderCallContext, endpoint string)
 
 	// Ensure the Port on the Host, if we get an error it is reasonable to
 	// assume that the address in the spec is invalid.
-	var err error
-	endpoint, err = lxd.EnsureHostPort(endpoint)
+	lxdEndpoint, err := lxd.EnsureHostPort(endpoint)
 	if err != nil {
 		return errors.Trace(err)
+	}
+
+	// Make sure we have an https url
+	if lxdEndpoint != endpoint {
+		return errors.Errorf("invalid URL %q: only HTTPS is supported", endpoint)
 	}
 
 	// Connect to the remote server anonymously so we can just verify it exists
 	// as we're not sure that the certificates are loaded in time for when the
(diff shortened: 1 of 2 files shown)


hmlanigan:
Suggestions for a better way to fix this are welcome.  Ideally, if you gave an http or https, we could write it to the CloudSpec as https since the lxd endpoint validation converts http to https anyways.  However with the way Ping() and pollster.VerifyURLs() work, that's not straightforward.

hmlanigan:
[bug 1787605 juju should warn on http url during add-cloud for lxd](https://trello.com/c/Seip2Ke4/515-bug-1787605-juju-should-warn-on-http-url-during-add-cloud-for-lxd)

provider/lxd/provider.go
```diff
@@ -165,18 +165,22 @@ func (p *environProvider) Ping(ctx context.ProviderCallContext, endpoint string)
...
-	var err error
-	endpoint, err = lxd.EnsureHostPort(endpoint)
+	lxdEndpoint, err := lxd.EnsureHostPort(endpoint)
 	if err != nil {
 		return errors.Trace(err)
 	}
 
+	// Make sure we have an https url
+	if lxdEndpoint != endpoint {
+		return errors.Errorf("invalid URL %q: only HTTPS is supported", endpoint)
```
SimonRichardson: We don't need this error, we shouldn't care about the URL schema, I think we should just fix it.

SimonRichardson: Having thought about this more, @mitechie what are your thoughts about just throwing an error if the url is wrong and that we then don't swap the url behind the scenes, potentially confusing people?

SimonRichardson:
Approved after discussing in standup

hmlanigan:
$$merge$$

PR #9114, opened 2018-08-26
leases: Don't crash the lease manager on timeouts

babbageclunk:
## Description of change

At startup there's a period where the lease manager is running but raft isn't yet. We get timeouts when workers try to claim singular leases in  that time, but if the lease manager bounces because of them it causes a lot of fallout because that also bounces the API server, breaking API connections and invalidating watchers.

Instead we log the timeout error, as well as responding to the client in the case of claims/extensions timing out. 

## QA steps

* Bootstrap a controller and deploy an application immediately when the bootstrap finishes (ie `juju bootstrap ...  && juju deploy ...`. The deploy should succeed - previously it would complain that the API connection was closed.

### DIFF
--- a/worker/lease/manager.go
+++ b/worker/lease/manager.go
@@ -205,14 +205,25 @@ func (manager *Manager) retryingClaim(claim claim) {
 		}
 	}
 
-	if success {
-		claim.respond(nil)
-	} else if lease.IsTimeout(err) {
+	if lease.IsTimeout(err) {
 		claim.respond(lease.ErrTimeout)
-	} else if err == nil {
-		claim.respond(lease.ErrClaimDenied)
+		manager.config.Logger.Warningf("[%s] retrying timed out while handling claim", manager.logContext)
+		return
 	}
-	// Otherwise we allow the fatal error to send errStopped.
+	if err == nil {
+		if !success {
+			claim.respond(lease.ErrClaimDenied)
+			return
+		}
+		claim.respond(nil)
+		// This is pretty subtle - in the success case we send a nil
+		// back on the errors channel. That indicates to the main loop
+		// that some lease state has changed and gives it a chance to
+		// loop back around and determine a new wakeup time based on
+		// new leases.
+	}
+	// Otherwise we allow the fatal error to send errStopped back to
+	// the client.
 
 	select {
 	case <-manager.catacomb.Dying():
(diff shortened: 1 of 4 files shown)


worker/lease/manager.go
```diff
@@ -207,12 +207,21 @@ func (manager *Manager) retryingClaim(claim claim) {
...
 	if success {
 		claim.respond(nil)
+		// This is pretty subtle - in the success case we send a nil
+		// back on the errors channel. That indicates to the main loop
+		// that some lease state has changed and gives it a chance to
+		// loop back around and determine a new wakeup time based on
+		// new leases.
 	} else if lease.IsTimeout(err) {
 		claim.respond(lease.ErrTimeout)
+		manager.config.Logger.Errorf("[%s] retrying failed in handleClaim", manager.logContext)
```
wallyworld: I don't think this is an error - if this is expected (which it is at startup) we are retrying and expect it to come good etc, can we make it a Warning?
Can the message be tweaked to something like "[%s] retrying lease claim timeout" ?

worker/lease/manager.go
```diff
@@ -330,6 +339,12 @@ func (manager *Manager) retryingTick(now time.Time) {
 		return
 	default:
 	}
+
+	if lease.IsTimeout(err) {
+		// We don't crash on timeouts to avoid bouncing the API server.
+		manager.config.Logger.Errorf("[%s] retrying failed in tick", manager.logContext)
```
wallyworld: Same here etc - warning plus message should say it's a timeout?

worker/lease/manager.go
```diff
@@ -207,12 +207,21 @@ func (manager *Manager) retryingClaim(claim claim) {
 
 	if success {
```
wallyworld: I think this block can be made a little more easy to follow. With the returns, we don't need the else if etc:

```
if lease.IsTimeout(err) {
    claim.respond(lease.ErrTimeout)
    return
}
if err == nil {
    if !success {
        claim.respond(lease.ErrClaimDenied)
        return
    }
    // Do not return here - this is pretty subtle .... etc
    claim.respond(nil)
}
```

worker/lease/manager.go
babbageclunk: Ok, sounds good to me.

worker/lease/manager.go
babbageclunk: That's much nicer, thanks.

worker/lease/manager.go
babbageclunk: Although I don't think the message you're suggesting works - this is logged when retrying has failed at this level (although the clients further up the stack will try to claim the lease again).

babbageclunk:
$$merge$$

babbageclunk:
Aborted the merge because I realised I hadn't changed the tick Errorf to a Warningf.

babbageclunk:
$$merge$$

PR #9115, opened 2018-08-27
Cloud credential validation against model's cloud.

anastasiamac:
## Description of change

When validating a cloud credential for a model, we need to ensure that the credential is valid for the cloud, for example that it is of auth-type that the cloud supports.

Most of this functionality has been written in state. This PR re-factors it for common use, exposes it and uses it in the apiserver call.

### DIFF
--- a/state/cloudcredentials.go
+++ b/state/cloudcredentials.go
@@ -10,7 +10,7 @@ import (
 	"github.com/juju/errors"
 	jujutxn "github.com/juju/txn"
 	"gopkg.in/juju/names.v2"
-	mgo "gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
 	"gopkg.in/mgo.v2/txn"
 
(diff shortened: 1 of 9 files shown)


anastasiamac:
[new credential must be valid for model's cloud](https://trello.com/c/RvH8Qi6k/417-new-credential-must-be-valid-for-models-cloud)

state/cloudcredentials.go
```diff
@@ -299,6 +284,26 @@ func validateCloudCredentials(
 	return ops, nil
 }
 
+func validateCredentialForCloud(nuage cloud.Cloud, tag names.CloudCredentialTag, credential Credential) error {
+	if tag.Cloud().Id() != nuage.Name {
+		return errors.NotValidf("cloud %q", tag.Cloud().Id())
```
wallyworld: perhaps this error message should include both cloud names - the one from the cloud and the one from the credential so the nature of the mismatched is surfaced

anastasiamac: Yes \o/ I agree - am thrilled we think the same..
I wanted to make sure that we have both. This PR includes the error message that we'd get in this case. It'd be something along the lines of:
```
 'validating credential "stratus/bob/foobar" for cloud "dummy": cloud "stratus" not valid`
```

anastasiamac:
And, yes, haha 'nuage' is a cloud in French :) I was trying to figure out how to name the variable without repeating 'cloud' as in this context it was also the name of the package :)

PR #9116, opened 2018-08-27
Use StatePool for initial database connections

howbazaar:
This branch looks to use state.OpenStatePool in the place of state.Open. Ideally we can delete state.Open.

This means that the primary system state instance uses the txn watcher from the statepool.

InitializeState now returns a controller and a state pool. The controller owns the statepool and one of these methods need to have Close called.

## Bug reference

There is a bug related to intermitten test failures in the all_watcher tests, which I think are fixed now.


howbazaar:
!!build!!

howbazaar:
$$merge$$

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #9120, opened 2018-08-28
k8s operators as deployed as statefulsets and use storage correctly

wallyworld:
## Description of change

This PR fixes a number of aspects of k8s operator deployments.

k8s application operators now are deployed using either a deployment controller or statefulset (depending on if storage is used).
The operator provisioning api has been extended to provide the necessary storage information. juju will look for a storage-pool called "operator-storage" and use that if defined.
storage classes are now prefixed with the model name when created since they are global entities. juju will look first for the model class, and then for a global class.

## QA steps

Create a k8s model on aws.

Create a juju operator storage class:
$ juju create-storage-pool operator-storage kubernetes storage-class=juju-operator-storage storage-provisioner=kubernetes.io/aws-ebs parameters.type=gp2

Deploy a charm and see that the operator is deployed as a statefulset and that an ebs volume has been provisioned and used for operator storage.

### DIFF
--- a/api/caasoperatorprovisioner/client.go
+++ b/api/caasoperatorprovisioner/client.go
@@ -14,6 +14,7 @@ import (
 	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/core/life"
 	"github.com/juju/juju/core/watcher"
+	"github.com/juju/juju/storage"
 )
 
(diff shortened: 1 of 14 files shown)


ycliuhw:
Looks awesome!
Thanks!

PR #9122, opened 2018-08-28
Only show meter status for models under SLA.

cmars:
Fixes LP:#1779175

## Description of change

Only display model meter status when an SLA is set.

## QA steps

Deploy a model that collects metrics like cs:canonical-kubernetes or cs:kubernetes-core. Do not set an SLA. `juju status` should no longer display model meter status.

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1779175


cmars:
@manadart PTAL, thanks!

cmars:
$$merge$$

PR #9123, opened 2018-08-28
Allow the hubwatcher to die even if it has pending events to send.

howbazaar:
This fixes an infinite loop in shutdown of the hubwatcher where the hubwatcher wouldn't stop if there were pending messages to send.

## QA steps

Due to the inherent racy nature of this bug, it is very hard to trigger.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1789510


howbazaar:
$$merge$$

PR #9124, opened 2018-08-28
Notify users of units affected by machine series upgrade.

ExternalReality:
## Description of change

When running `upgrade-series prepare` users should be notified of the units that are affected by the upgrade.

## QA steps

* `upgrade-series prepare <machine>`
* You should be notified of the units that would be changed by the issuing of the above command.

## Documentation changes

N/A (Perhaps some Discourse changes at first) 

## Bug reference

N/A

### DIFF
--- a/api/machinemanager/machinemanager.go
+++ b/api/machinemanager/machinemanager.go
@@ -19,10 +19,15 @@ type Client struct {
 	facade base.FacadeCaller
 }
 
+// MakeClient is a constructor function for a machine manager client
+func MakeClient(clientFacade base.ClientFacade, facadeCaller base.FacadeCaller) *Client {
+	return &Client{ClientFacade: clientFacade, facade: facadeCaller}
+}
+
 // NewClient returns a new machinemanager client.
 func NewClient(st base.APICallCloser) *Client {
 	frontend, backend := base.NewClientFacade(st, machineManagerFacade)
-	return &Client{ClientFacade: frontend, facade: backend}
+	return MakeClient(frontend, backend)
 }
 
 // AddMachines adds new machines with the supplied parameters, creating any requested disks.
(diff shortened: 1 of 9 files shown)


ExternalReality:
[Note the units affected during message output](https://trello.com/c/VkM0d6BI/474-note-the-units-affected-during-message-output)

ExternalReality:
!!build!!

ExternalReality:
!!build!!

api/machinemanager/machinemanagernew_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/juju/juju/api/base/mocks"
+	"github.com/juju/juju/api/machinemanager"
+	"github.com/juju/juju/apiserver/params"
+	jujutesting "github.com/juju/juju/testing"
+)
+
+var _ = gc.Suite(&NewMachineManagerSuite{})
+
+type NewMachineManagerSuite struct {
```
manadart: What is the rationale behind the new file/suite here?

api/machinemanager/machinemanager.go
```diff
@@ -19,10 +19,15 @@ type Client struct {
 	facade base.FacadeCaller
 }
 
+// MakeClient is a constructor function for a machine manager client
+func MakeClient(clientFacade base.ClientFacade, facadeCaller base.FacadeCaller) *Client {
```
manadart: Is there a better name that we can use for this?
It is incongruent with the Go idioms new (for references) and make (for object values).

api/machinemanager/machinemanagernew_test.go
ExternalReality: The existing `machinemanager_test.go` uses a home grown mocking mechanism. I wanted to establish the `new` suffixed file to have a place to start systematically moving those tests to use gomock. There are two benefits to this:

1) We can work piecemeal
2) We don't have to mix two mocking styles (in attempt to preserve one file) when transitioning between mocking styles

The plan is to start moving those old style tests and when finished delete the old file and `mv` the new file.

manadart: Understood - strangulation.
Maybe a comment at the top along these lines?

api/machinemanager/machinemanager.go
ExternalReality: That it is will fix. It just `NewClient` was already by the caller.

api/machinemanager/machinemanagernew_test.go
ExternalReality: Going to place my comment above in the code at the top of the file.

ExternalReality:
$$MERGE$$

ExternalReality:
$$MERGE$$

ExternalReality:
$$MERGE$$

PR #9125, opened 2018-08-29
Fixes lp#1789415: Deal with nil value for 'juju config <app> <setting>'.

anastasiamac:
## Description of change

We have recently changed how a value of a single application setting is displayed - it is directly printed on stdout. It is no longer formatted using yaml. This means that we are better handling multi-line string values. However, the corner case with this approach - settings that have nil values. '<nil>' is displayed and breaks existing scripting.
This PR fixes the issue and adds unit test for this scenario.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1789415


anastasiamac:
[empty default app config, lp#1789415](https://trello.com/c/yUu94Njq/422-empty-default-app-config-lp1789415)

anastasiamac:
$$merge$$

PR #9128, opened 2018-08-29
raftleases: Increase the raft-forward timeout threshold to 5s

babbageclunk:
## Description of change
The raft forward timeout needs to be high enough that when we see it we can be very confident that it happened because there was no raft-forwarder listening on the hub. The old value was low enough that we'd sometimes see false reports of timeouts even though the request was actually received and applied to raft. This would mean that the notify target wouldn't be sent the update if a lease was claimed or expired, for example.

There's not much downside to having a high-ish value - the caller will be blocked waiting for a response, but other lease operations aren't queued behind this one.

Also update the pubsub dependency so we don't see debug logging every time a clock update is sent.

## QA steps

* Tests pass, bootstrapping and deploying a set of applications with multiple units, watching the logs for timeout messages (none should happen once the system has started up - some are acceptable while raft is waiting for the peergrouper to publish api details).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -565,11 +565,11 @@
   revision = "5f8741c297b47cba29f747c67f2e31e7c2a36ecb"
 
 [[projects]]
-  digest = "1:f156d17854ec35dd41a88a507103cb0f703f62332fa31315779a303c565667c9"
+  digest = "1:76068d833db78ba2419f72e5efd91d1d4ab5e8ffca1f61c5b5a724795fea6572"
   name = "github.com/juju/pubsub"
   packages = ["."]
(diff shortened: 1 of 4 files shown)


babbageclunk:
Extra detail on the weird addition to Gopkg.lock - it seems like this is a consequence of someone adding an import for a previously-unused package (`gopkg.in/goose.v2/testservices/neutronmodel`) to a test. They didn't run rebuild-dependencies - pretty reasonably, since they weren't adding or updating a dependency, just using a package from an existing one. So this is probably just something that will happen occasionally when we update deps. I think that's ok.

babbageclunk:
$$merge$$

PR #9131, opened 2018-08-29
worker/lease: Ensure the lease manager waits for child goroutines

babbageclunk:
## Description of change

We were seeing errors in test teardown caused by a state-lease store operation trying to use a closed session. The tests will avoid closing the session until the state workers are stopped, but the lease manager would stop before its goroutines were finished. Use a waitgroup to prevent the manager from stopping until in-flight requests have finished or timed out.

## QA steps

* Added a test to verify that the manager waits until the child goroutines are finished. Ran state tests under stress to try to reproduce the error but couldn't.

### DIFF
--- a/worker/lease/manager_async_test.go
+++ b/worker/lease/manager_async_test.go
@@ -504,3 +504,87 @@ func (s *AsyncSuite) TestClaimRepeatedTimeout(c *gc.C) {
 		workertest.CheckAlive(c, manager)
 	})
 }
+
+func (s *AsyncSuite) TestWaitsForGoroutines(c *gc.C) {
+	// The manager should wait for all of its child tick and claim
+	// goroutines to be finished before it stops.
+	tickStarted := make(chan struct{})
+	tickFinish := make(chan struct{})
+	claimStarted := make(chan struct{})
+	claimFinish := make(chan struct{})
+	fix := Fixture{
+		leases: leaseMap{
+			key("legacy"): {
+				Holder: "culprate",
+				Expiry: offset(-time.Second),
+			},
+		},
+		expectCalls: []call{{
+			method: "Refresh",
+		}, {
+			method: "ExpireLease",
+			args:   []interface{}{key("legacy")},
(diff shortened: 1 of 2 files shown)


worker/lease/manager_async_test.go
```diff
@@ -504,3 +504,87 @@ func (s *AsyncSuite) TestClaimRepeatedTimeout(c *gc.C) {
...
+		case <-time.After(coretesting.LongWait):
+			c.Fatalf("timed out waiting for result")
+		}
+
+		workertest.CheckAlive(c, manager)
+
+		// And when we finish the tick the worker stops.
+		close(tickFinish)
+
+		err = workertest.CheckKilled(c, manager)
```
wallyworld: Shouldn't this be in a retry loop - it might take X ms to die

worker/lease/manager_async_test.go
```diff
@@ -504,3 +504,87 @@ func (s *AsyncSuite) TestClaimRepeatedTimeout(c *gc.C) {
...
+		case <-claimStarted:
+		case <-time.After(coretesting.LongWait):
+			c.Fatalf("timed out waiting for claim start")
+		}
+
+		// If we kill the manager now it won't finish until the claim
+		// call finishes (no worries about timeouts because we aren't
+		// advancing the test clock).
+		manager.Kill()
+		workertest.CheckAlive(c, manager)
```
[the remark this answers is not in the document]
babbageclunk: CheckAlive does that - fires off a goroutine waiting for the worker to be dead, then does a select with a ShortWait timeout.

worker/lease/manager_async_test.go
babbageclunk: Same with CheckKilled - in fact it's the same as CheckAlive except the error raise is in the other leg of the select.

babbageclunk: Oh, and the timeout is longer.

babbageclunk:
$$merge$$

PR #9142, opened 2018-08-30
LXD: Interactive Auth Type update

SimonRichardson:
## Description of change

*the following is out dated, but provides context - see addendum*

The following commit attempts to hide certain values from the UI,
whilst still providing values for the schema at a later stage. The
idea for this is that LXD allows the usage of a trust-password, but
that is transient as it's replaced with certificates as soon as
a connection to the server is made.

So the schema really needs to only show a trust password, then the
following values are then filled when a communication is setup. It
seems rather backwards and unique for LXD, the reason being that
the schema that we're using is rather rigid and provides no current
flexibility to enable this sort of flexibility.

It should be noted that the reason we don't keep the trust-password
and just re-establish every time using the trust-password is because
the LXD security documentation[1], categorically states that the
trust-password should be turned off once all server have joined the
cluster.

[1] https://github.com/lxc/lxd/blob/master/doc/security.md#production-setup

## Addendum

So with the following change we no longer need HideFromUI, as we
convert all other auth types into certificate auth type. So we
only need to add a trust-password for the schema and once we require
connectivity to the lxd server we get the certificates and fill them
into the credentials yaml file and we no longer need to talk about
interactive mode.

*note: the QA steps are the same*

## QA steps

Save this file as `clouds.yaml` somewhere, we need to access it later:
*Note:* It assumes you have setup a remote cloud to access.
```yaml
clouds:
    lxd-remote:
        type: lxd
        auth-types: [interactive,certificate]
        endpoint: https://$URL_REMOTE_LXD:8443
```

To replicate this, follow the following steps:

```sh
juju add-cloud lxd-remote /path/to/clouds.yaml
juju add-credential lxd-remote
```

You should see the following output (see demo)

```sh
Enter credential name: lxd-remote

Auth Types
  interactive
  certificate

Select auth type [interactive]: 

Enter trust-password: 

Loaded client cert/key from "/home/$USER/.config/lxc"
Reusing certificate from LXD server.
Credential "lxd-remote" added locally for cloud "lxd-remote".
```

## Documentation changes

Just makes the CLI more understandable.

## Bug reference

N/A

## Demo

[asciicast](https://asciinema.org/a/198924?speed=2)

### DIFF
--- a/cloud/credentials.go
+++ b/cloud/credentials.go
@@ -319,13 +319,6 @@ type CredentialAttr struct {
 	// attributes are provided only to the Juju controllers.
 	Hidden bool
 
-	// HideFromUI controls if the attribute should be displayed at all in the
-	// UI. This differs from Hidden, where we want to display the field, but
-	// hide the input. It's analogous to Optional, but we don't display it in
-	// the UI. The reason for this existence is that we want an attribute
-	// with in the "schema", but we don't want to ask the user for the values.
-	HideFromUI bool
-
 	// FileAttr is the name of an attribute that may be specified instead
 	// of this one, which points to a file that will be read in and its
 	// value used for this attribute.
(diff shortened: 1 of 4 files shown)


SimonRichardson:
!!build!!

Failed because > `ERROR juju.testing mgo.go:311 mongodb has exited without being killed`

SimonRichardson:
!!build!!

Failed because > https://bugs.launchpad.net/juju/+bug/1788401

SimonRichardson:
$$merge$$

PR #9146, opened 2018-08-31
Fixes lp#1788401: intermittent "charm not a valid zip". 

anastasiamac:
## Description of change

I cannot reproduce this locally at all.

However, I believe since CI runs tests in parallel, there could be a discrepancy between client creation at SetUpTest and client closure in TearDownTest.

This PR ensures that client is created within the test. Also everywhere else, unless we are explicitly testing for client closure, we do not close client in tests. So, this PR also removes closing code. 

The failure has been occurring fairly frequently in CI, so I want to build couple of times to see whether this PR is a fix.
Example check-merge failures:
http://ci.jujucharms.com/job/github-check-merge-juju/3267/
http://ci.jujucharms.com/job/github-check-merge-juju/3255/

github-check-merge-juju runs:
1 - Pass
2 - Pass for the test of interest, failed with other intermittent failures http://ci.jujucharms.com/job/github-check-merge-juju/3275/
3 - same as 2 passed for my test but other intertmittents - http://ci.jujucharms.com/job/github-check-merge-juju/3276/
4 - again my test passes, but another intermittent failure - http://ci.jujucharms.com/job/github-check-merge-juju/3277/
5 - Pass
6 - pass for my test, but another intermittent failure elsewhere: http://ci.jujucharms.com/job/github-check-merge-juju/3280/
7 - pass, with unrelated intermittent failure elsewhere: http://ci.jujucharms.com/job/github-check-merge-juju/3281
8- pass for my tests but another unrelated provisioner failure - http://ci.jujucharms.com/job/github-check-merge-juju/3282
9 - pass

## Bug reference

https://bugs.launchpad.net/juju/+bug/1788401

PR #9149, opened 2018-09-02
migrations: Ensure leadership is claimed when importing a model

babbageclunk:
## Description of change

With raft-based leases, the import code that claims leadership for applications in the incoming model needs to move out of state - the global lease manager isn't accessible from there. We now claim leadership in `migration.Import`, passing in a function to get the leadership claimer from the API facade.

This required changing facade context.LeadershipClaimer to accept a model uuid - migrationtarget is a controller facade, so the model being imported isn't the one that we're connected to (since it doesn't exist when the connection is made).

The other wrinkle is that we can't get the legacy leadership claimer from the facade context, because we'd need to get it from the state pool, which in turn would require that we either return a LeadershipClaimer extended with a Close to release the pooled state, or a closer function. It seemed simpler just to check the feature flag and get the claimer from the state we have after importing if LegacyLeases is set. 

## QA steps

* Bootstrap 2 controllers (A and B) and deploy 5 or 6 units of an application in a model on A.
* Migrate the model from A to B.
* After the migration is finished, verify that the leader has been maintained correctly.
* In debug logging for juju.migration check that the leadership was claimed for the right unit during import (so it wasn't just chance that the leadership was carried over).

### DIFF
--- a/apiserver/facades/controller/migrationtarget/migrationtarget_test.go
+++ b/apiserver/facades/controller/migrationtarget/migrationtarget_test.go
@@ -4,10 +4,12 @@
 package migrationtarget_test
 
 import (
+	"io/ioutil"
 	"time"
 
 	"github.com/juju/description"
 	"github.com/juju/errors"
+	"github.com/juju/loggo"
 	"github.com/juju/testing"
 	jc "github.com/juju/testing/checkers"
 	"github.com/juju/utils"
@@ -21,6 +23,8 @@ import (
 	"github.com/juju/juju/apiserver/facades/controller/migrationtarget"
 	"github.com/juju/juju/apiserver/params"
 	apiservertesting "github.com/juju/juju/apiserver/testing"
+	"github.com/juju/juju/core/leadership"
+	"github.com/juju/juju/core/lease"
 	"github.com/juju/juju/environs"
 	"github.com/juju/juju/environs/context"
 	"github.com/juju/juju/instance"
@@ -35,9 +39,10 @@ import (
 type Suite struct {
 	statetesting.StateSuite
 	resources  *common.Resources
-	authorizer apiservertesting.FakeAuthorizer
+	authorizer *apiservertesting.FakeAuthorizer
 
-	callContext context.ProviderCallContext
+	facadeContext facadetest.Context
+	callContext   context.ProviderCallContext
 }
 
 var _ = gc.Suite(&Suite{})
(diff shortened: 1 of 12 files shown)


babbageclunk:
Intermittent failure: https://bugs.launchpad.net/juju/+bug/1788401
Hopefully fixed by #9146 

!!build!!

apiserver/facades/controller/migrationtarget/migrationtarget_test.go
```diff
@@ -35,9 +39,10 @@ import (
 type Suite struct {
 	statetesting.StateSuite
 	resources  *common.Resources
-	authorizer apiservertesting.FakeAuthorizer
+	authorizer *apiservertesting.FakeAuthorizer
```
wallyworld: why this change?
when we store a facade context struct and not a pointer (for example)

migration/migration.go
```diff
@@ -16,14 +17,23 @@ import (
...
+	"github.com/juju/juju/feature"
 	"github.com/juju/juju/resource"
 	"github.com/juju/juju/state"
 	"github.com/juju/juju/tools"
 )
 
 var logger = loggo.GetLogger("juju.migration")
 
+// When we import a new model, we need to give the leaders some time to
+// settle. We don't want to have leader switches just because we migrated an
```
wallyworld: a/an/a

migration/migration.go
```diff
@@ -16,14 +17,23 @@ import (
...
+// model, so this time needs to be long enough to make sure we cover
+// the time taken to migration a reasonable sized model. We don't yet
```
wallyworld: s/migration/migrate

migration/migration.go
```diff
@@ -16,14 +17,23 @@ import (
...
+// know how long this is going to be, but we need something.
+const initialLeaderClaimTime = time.Minute
```
wallyworld: Arbitrary times like this are a bit kudgy. I can see it could take longer than a minute waiting for all the agents to restart again etc. I wonder if we need a way to lock or freeze leadership claims and then unfreeze after migration.
Can we at least make it a todo to come back and take another look

state/upgrades.go
```diff
@@ -1272,6 +1273,8 @@ func MigrateLeasesToGlobalTime(pool *StatePool) error {
 	return runForAllModelStates(pool, migrateModelLeasesToGlobalTime)
 }
 
+const initialLeaderClaimTime = time.Minute
```
wallyworld: This appears duplicated. Can we use a single sourced value?

apiserver/facades/controller/migrationtarget/migrationtarget_test.go
babbageclunk: I needed to be able to change it in a test and have the change be visible in the constructed facade. When it wasn't a pointer then the authorizer was getting copied into the facade context, and changes made in the body of the test had no effect.

migration/migration.go
babbageclunk: Oh, yup!

migration/migration.go
babbageclunk: Yup - I just cut-n-pasted this without reading it closely enough, I'm blaming @mjs for the original typos.

state/upgrades.go
babbageclunk: Actually, it needed to be public in state - the other way produces an import cycle.

migration/migration.go
mjs: I apologise for any migraines my migratory migrations typos may have caused :/

PR #9150, opened 2018-09-03
k8s app/unit status fixes

wallyworld:
## Description of change

Changes for k8s models.
Implement the scale attribute in status using the application scale attribute rather than counting units.
Adapt k8s units to support new presence implementation.

## QA steps

deploy a k8s application
check that unit status is not stuck on lost
scale the application
check status scale value


wallyworld:
$$merge$$

PR #9151, opened 2018-09-03
Fixes part lp#1790424: provisioner keep machine on remove.

anastasiamac:
## Description of change

TestStopInstancesIgnoresMachinesWithKeep started failing more frequently with the recent change that made Juju watcher more responsive.

The problem was with the test that have assumed that machine records were in particular state but has not done anything to ensure that it is so. Specifically we were adding a machine and setting it's haverstmode to keep after removal. The test that would proceed to remove the machine and check that it has been kept. However, with the new watcher, sometimes, the command to remove the machine came before the harvest mode was set causing test to fail. This PR adds a database sync before issuing removal of machine call.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1790424

The other part of this bug is related to another provisioner test failure which I am still chasing locally, and cannot reproduce. However, I wanted to propose this fix as TestStopInstancesIgnoresMachinesWithKeep fails frequently.

### DIFF
--- a/worker/provisioner/provisioner_task.go
+++ b/worker/provisioner/provisioner_task.go
@@ -285,7 +285,7 @@ func (task *provisionerTask) processMachines(ids []string) error {
 	}
 
 	if len(stopping) > 0 {
-		logger.Infof("stopping known instances %v", stopping)
+		logger.Infof("stopping known instances %v", instanceIds(stopping))
 	}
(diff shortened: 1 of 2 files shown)


anastasiamac:
Closing as it's best to re-write this particular test without JujuConnSuite in the mix.

PR #9156, opened 2018-09-03
Fix intermittent test failure.

howbazaar:
The TestClientWatchAllAdminPermission from apiserver/facades/client/client failed intermittently.

When run locally it would fail about 50% of the time. After much digging, I found that this was a race in the underlying multiwatcher code. If the intial GetAll was processed and the underlying changes from the document watchers hadn't been processed, the test would pass, but if the change for the remote application was processed before Next was called in the test, it would fail.

This was due to the entity being assigned into the multiwatcher state was missing the status value. To handle this, if the entity isn't new, we get the status from the old value.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1790548

### DIFF
--- a/state/allwatcher_internal_test.go
+++ b/state/allwatcher_internal_test.go
@@ -372,6 +372,9 @@ func addTestingRemoteApplication(
 		IsConsumerProxy: isProxy,
 	})
 	c.Assert(err, jc.ErrorIsNil)
+	status, err := rs.Status()
+	c.Assert(err, jc.ErrorIsNil)
+
 	return rs, multiwatcher.RemoteApplicationInfo{
 		ModelUUID: st.ModelUUID(),
 		Name:      name,
@@ -379,8 +382,10 @@ func addTestingRemoteApplication(
 		OfferURL:  url,
 		Life:      multiwatcher.Life(rs.Life().String()),
 		Status: multiwatcher.StatusInfo{
-			Current: "unknown",
-			Data:    map[string]interface{}{},
+			Current: status.Status,
+			Message: status.Message,
+			Data:    status.Data,
+			Since:   status.Since,
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


howbazaar:
$$merge$$

howbazaar:
$$merge$$

state/allwatcher_internal_test.go
```diff
@@ -1028,11 +990,12 @@ func (s *allWatcherStateSuite) TestStateWatcher(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 	err = wu.AssignToMachine(m2)
 	c.Assert(err, jc.ErrorIsNil)
+	wpTime := s.state.clock().Now()
 
 	// Look for the state changes from the allwatcher.
 	deltas = tw.All(5)
 
-	zeroOutTimestampsForDeltas(c, deltas)
+	// zeroOutTimestampsForDeltas(c, deltas)
```
wallyworld: delete me

state/allwatcher_internal_test.go
```diff
@@ -1834,7 +1792,7 @@ func (s *allModelWatcherStateSuite) TestStateWatcher(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 
 	deltas = tw.All(1)
-	zeroOutTimestampsForDeltas(c, deltas)
+	//zeroOutTimestampsForDeltas(c, deltas)
```
wallyworld: delete me

state/allwatcher_internal_test.go
```diff
@@ -1860,7 +1818,7 @@ func (s *allModelWatcherStateSuite) TestStateWatcher(c *gc.C) {
...
```
wallyworld: etc etc

howbazaar:
4 other intermittent failures.
$$merge$$

PR #9159, opened 2018-09-04
Handle if collStats returns an int64 for size.

jameinel:
## Description of change

In production, we ran into panic() because Mongo was returning an int64 for a size field instead of an int. We don't have full understanding of how it was happening, but this should at least cover the issue.

Add some testing for how we handle the bson.M result and allow it to be
int or int64 and generate an error rather than panic if it isn't.

## QA steps

Tests added to cover the new code path which was factored out so that it could be directly testable. We don't have exact reproduction for how to get Mongo to behave in this way. One possibility is to just create a logs table that is very very large (>2GB?) and see if it panics without this change.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1790626

### DIFF
--- a/state/logs.go
+++ b/state/logs.go
@@ -864,12 +864,18 @@ func dbCollectionSizeToInt(result bson.M, collectionName string) (int, error) {
 		return 0, nil
 	}
 	if asint, ok := size.(int); ok {
+		if asint < 0 {
+			return 0, errors.Errorf("mongo collStats for %q returned a negative value: %v", collectionName, size)
+		}
 		return asint, nil
 	}
 	if asint64, ok := size.(int64); ok {
 		// 2billion megabytes is 2 petabytes, which is outside our range anyway.
 		if asint64 > math.MaxInt32 {
 			return math.MaxInt32, nil
+		}
+		if asint64 < 0 {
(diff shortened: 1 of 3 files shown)


jameinel:
[1790626 mongo collStats Size](https://trello.com/c/mf3DRF8M/547-1790626-mongo-collstats-size)

state/logs.go
```diff
@@ -854,6 +855,29 @@ func initLogsSession(st ModelSessioner) (*mgo.Session, *mgo.Collection) {
...
+		return 0, nil
+	}
+	if asint, ok := size.(int); ok {
+		return asint, nil
+	}
+	if asint64, ok := size.(int64); ok {
+		// 2billion megabytes is 2 petabytes, which is outside our range anyway.
+		if asint64 > math.MaxInt32 {
+			return math.MaxInt32, nil
+		}
```
SimonRichardson: Is there ever a case that `asint64 < 0` so that we also handle the min case or `math.MinInt32` case?

jameinel: Well, if mongo returns a negative number of megabytes for the size of a collection we certainly have issues. I'm ok handling that case, just in a belt-and-suspenders type of defensive programming. But I'd probably cap the threshold at 0. Cause a negative size would do *really* weird things.

jameinel:
$$merge$$

PR #9160, opened 2018-09-04
use sed -E instead of -r for portability in Makefile

hmlanigan:
## Description of change

Use sed -E instead of -r for portability in Makefile.  Per sed man page they are interchangeable.

## QA steps

make juju on an osx machine, "sed: illegal option -- r" should not appear.
make juju on ubuntu, no change.

## Documentation changes

n/a

## Bug reference

n/a


mitechie:
LGTM and QA ok checking these are replaceable in Ubuntu

    -E, -r, --regexp-extended
                 use extended regular expressions in the script
                 (for portability use POSIX -E).

hmlanigan:
$$merge$$

PR #9161, opened 2018-09-05
raft-leases: move change notification from the store to the forwarder

babbageclunk:
## Description of change

When lease changes (claims or expiries) happen in raft, we need to update the database accordingly. This was being done in the raftlease store, but because commands are sent across the hub, it's hard to be certain whether a command has been successfully applied (and thus that the db needs to be updated) - in particular one can't tell the difference between a message being dropped (because of a raft-leadership change) or one where we just didn't get a response in time.

Instead, move the notification to happen in the forwarder, which is on the other side of the hub applying commands to the raft node directly. FSM.Apply now returns a `FSMResponse` which indicates whether the operation was successful, and can update a notification target with any changes.

I initially thought about putting the notification inside the FSM, but that doesn't work - the FSM can't know whether the commands it sees are current commands or replaying log entries to bring a newly-joined raft node up to the current state. So the fact that the forwarder calls `response.Notify(target)` is key to ensuring that only real commands generate DB changes.

## QA steps

* Bootstrap a controller, deploy an application with multiple units, ensure that leadership changes are reflected in the leaseholders collection in mongo (and in the status output).


core/raftlease/fsm.go
```diff
@@ -153,15 +165,42 @@ type entry struct {
 	duration time.Duration
 }
 
+// response stores what happened as a result of applying a command.
```
wallyworld: Maybe mention in the comment that response implements FSMResponse. Or better add a line just above 
var _ FSMResponse = (*response)(nil)

worker/raft/raftforwarder/manifold.go
```diff
@@ -31,22 +90,82 @@ func (config ManifoldConfig) start(context dependency.Context) (worker.Worker, e
...
+		return nil, errors.Trace(err)
+	}
+
+	st := statePool.SystemState()
+
+	logPath := filepath.Join(agent.CurrentConfig().LogDir(), "lease.log")
+	if err := primeLogFile(logPath); err != nil {
+		// This isn't a fatal error, so log and continue if priming
+		// fails.
+		config.Logger.Errorf(
```
wallyworld: If it's not fatal it should be a warning.

worker/raft/raftforwarder/worker.go
```diff
@@ -127,11 +132,12 @@ func (w *forwarder) processRequest(command []byte) (raftlease.ForwardResponse, e
 		return empty, errors.Trace(err)
 	}
 	respValue := future.Response()
-	responseErr, ok := respValue.(error)
-	if respValue != nil && !ok {
-		return empty, errors.Errorf("FSM response must be an error or nil, got %#v", respValue)
+	response, ok := respValue.(raftlease.FSMResponse)
+	if !ok {
+		return empty, errors.Errorf("expected an FSMResponse, got %#v", respValue)
```
wallyworld: Often helpful to include the type %T in the error message

worker/raft/raftforwarder/manifold.go
```diff
@@ -4,25 +4,84 @@
...
 	CentralHubName string
 
 	RequestTopic string
 	Logger       Logger
 	NewWorker    func(Config) (worker.Worker, error)
+	NewTarget    func(*state.State, io.Writer, Logger) raftlease.NotifyTarget
+}
+
+// Validate checks that the config has all the required values.
+func (config ManifoldConfig) Validate() error {
```
wallyworld: Is this tested anywhere?

core/raftlease/fsm.go
babbageclunk: Will do!

worker/raft/raftforwarder/manifold.go
babbageclunk: Ok.

worker/raft/raftforwarder/worker.go
babbageclunk: Yup

worker/raft/raftforwarder/manifold.go
babbageclunk: Oops, no - I'll add one.

babbageclunk:
$$merge$$

babbageclunk:
Good old provisioner $$merge$$

PR #9162, opened 2018-09-05
remove charmrepo.v2 from dependencies, only charmrepo.v3 is used

hmlanigan:
remove charmrepo.v2 from dependencies, only charmrepo.v3 is used


hmlanigan:
[remove dependency on charmrepo.v2, not used](https://trello.com/c/j94AugKQ/558-remove-dependency-on-charmrepov2-not-used)

hmlanigan:
charmrepo.v2 used by vendor/github.com/juju/bundlechanges/handlers.go

PR #9165, opened 2018-09-06
Add stanza for 18.04:s390x to openstack test image metadata

babbageclunk:
## Description of change

The index entry for 18.04:s390x was added when the default series was bumped to bionic, and image metadata was added for the other architectures, but it seems like s390x was missed. This was preventing the s390x unit test job from passing.

## QA steps

* I removed the corresponding section for 18.04:amd64 and saw that the test failed in the same way when run locally.

### DIFF
--- a/provider/openstack/series_test.go
+++ b/provider/openstack/series_test.go
@@ -170,6 +170,25 @@ var imagesData = `
        }
      }
    },
+   "com.ubuntu.cloud:server:18.04:s390x": {
+     "release": "xenial",
+     "version": "18.04",
+     "arch": "s390x",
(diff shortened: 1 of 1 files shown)


babbageclunk:
$$merge$$

provider/openstack/series_test.go
```diff
@@ -170,6 +170,25 @@ var imagesData = `
        }
      }
    },
+   "com.ubuntu.cloud:server:18.04:s390x": {
+     "release": "xenial",
```
xnox: Don't you mean "bionic" here?

PR #9166, opened 2018-09-06
Add stanza for 18.04:s390x to openstack test image metadata

babbageclunk:
## Description of change

The index entry for 18.04:s390x was added when the default series was bumped to bionic, and image metadata was added for the other architectures, but it seems like s390x was missed. This was preventing the s390x unit test job from passing.

(Backported to 2.4 from #9165 - I'd have just made the change in 2.4 and left it for merging, but the test data has been moved out of export_test in develop.)

## QA steps

* I removed the corresponding section for 18.04:amd64 and saw that the test failed in the same way when run locally.

### DIFF
--- a/provider/openstack/export_test.go
+++ b/provider/openstack/export_test.go
@@ -217,6 +217,25 @@ var imagesData = `
        }
      }
    },
+   "com.ubuntu.cloud:server:18.04:s390x": {
+     "release": "xenial",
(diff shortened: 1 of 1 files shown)


babbageclunk:
This is a trivial backport, so it doesn't need approval.

$$merge$$

PR #9167, opened 2018-09-06
Fix for racy TestRemoveApplicationStopsWatchingApplication

babbageclunk:
## Description of change

This would fail occasionally because the worker loop hadn't had time to get the life for gitlab before the test would set the notfound error - in that case it would never start the watcher for gitlab, and never kill it.

Fix this by setting errors (starting with a nil) before firing any change events.

## QA steps

* Before this change the test would fail about 1 in 5 runs under stress-race, after the change I haven't seen it fail.


babbageclunk:
$$merge$$

babbageclunk:
This looks like an intermittent failure caused by the second ticking over between now() calls. I'll have a go at fixing it tomorrow.

$$merge$$

PR #9172, opened 2018-09-06
Updated ec2-instance-types

SimonRichardson:
## Description of change

The following commit updates the ec2 instance types from the index.json
file and running "go generate". The fallout from that update are these
changes.

## QA steps

The following assumes that you have correctly setup AWS. If not you can follow
the steps [here](https://docs.jujucharms.com/2.4/en/help-aws)

```sh
juju bootstrap aws aws-controller
juju deploy mysql --constraints 'instance-type=m2.xlarge'
```


SimonRichardson:
https://trello.com/c/9jXjeoLl/563-bug-1790647-update-aws-instance-types-23-update

hmlanigan:
I'll QA this one as I review

provider/ec2/internal/ec2instancetypes/instancetypes_test.go
```diff
@@ -26,7 +26,6 @@ func (s *InstanceTypesSuite) TestRegionInstanceTypes(c *gc.C) {
 		"m3.medium", "m3.large", "m3.xlarge", "m3.2xlarge",
 		"c1.medium", "c1.xlarge", "cc2.8xlarge",
 		"c3.large", "c3.xlarge", "c3.2xlarge", "c3.4xlarge", "c3.8xlarge",
-		"cg1.4xlarge",
 		"g2.2xlarge",
```
hmlanigan: Unrelated to this change directly I know, just some questions. TestSupportsClassic() still uses "cg1.4xlarge": assertSupportsClassic("cg1.4xlarge").  Is it correct that the test still passes?  I'm wondering if the test is testing what it thinks it's testing?

SimonRichardson: I agree, I think some of the tests are setup via mocks and therefore some tests still pass even when they're not "live". I don't fully grasp why I've had to change so many tests, so it would be good to get that explained.

provider/ec2/internal/ec2instancetypes/process_cost_data.go
```diff
@@ -179,6 +179,11 @@ func process(in io.Reader) (map[string][]instanceType, metadata, error) {
 		if productInfo.Tenancy != "Shared" {
 			continue
 		}
+		if productInfo.PreInstalledSW != "NA" {
```
jameinel: couldn't it also be "" ?

provider/ec2/local_test.go
```diff
@@ -409,8 +409,8 @@ func (t *localServerSuite) TestSystemdBootstrapInstanceUserDataAndState(c *gc.C)
 	// check that a new instance will be started with a machine agent
 	inst1, hc := testing.AssertStartInstance(c, env, t.ControllerUUID, "1")
 	c.Check(*hc.Arch, gc.Equals, "amd64")
-	c.Check(*hc.Mem, gc.Equals, uint64(3.75*1024))
-	c.Check(*hc.CpuCores, gc.Equals, uint64(1))
+	c.Check(*hc.Mem, gc.Equals, uint64(1024))
```
jameinel: this sounds like our default instance type is going away from a 3.75GB machine with 1 CPU to a 1GB machine with 2 CPUs, which probably isn't the best tradeoff, is it?

SimonRichardson: Yeah, that's why I'm uncomfortable to land this. This seems like a big tradeoff

provider/ec2/internal/ec2instancetypes/process_cost_data.go
SimonRichardson: Unfortunately not according to the `index.json`

```
"preInstalledSw" : "NA",
```
vs
```
 "preInstalledSw" : "SQL Ent",
```

SimonRichardson:
I've verified this manually, that the [bootstrap.go](https://github.com/juju/juju/blob/64063c7f3bcd7c8658436bf3aee67835dd537c4c/environs/bootstrap/bootstrap.go#L168-L178) code does indeed enforce the constraints for ec2

SimonRichardson:
$$merge$$

PR #9178, opened 2018-09-07
2.4: fix build errors found with go 1.11 test

hmlanigan:
## Description of change

Fix build errors found with running go test with go version 1.11. 

## QA steps

Have branch downloaded and dependencies setup.
Install go 1.11
Edit Makefile to not run pre-check when check is run.
make check   <-- intermittent unit test failures are possible, but no build failures should be logged.

### DIFF
--- a/provider/azure/instance.go
+++ b/provider/azure/instance.go
@@ -243,7 +243,7 @@ func (inst *azureInstance) OpenPorts(ctx context.ProviderCallContext, machineId 
 
 		priority, err := nextSecurityRulePriority(nsg, securityRuleInternalMax+1, securityRuleMax)
 		if err != nil {
-			return errors.Annotatef(err, "getting security rule priority for %s", rule)
+			return errors.Annotatef(err, "getting security rule priority for %q", rule)
 		}
 
 		var protocol network.SecurityRuleProtocol
@@ -283,7 +283,7 @@ func (inst *azureInstance) OpenPorts(ctx context.ProviderCallContext, machineId 
 			nil, // abort channel
 		)
 		if err := <-errCh; err != nil {
-			return errors.Annotatef(err, "creating security rule for %s", securityRule)
+			return errors.Annotatef(err, "creating security rule for %q", ruleName)
 		}
 		securityRules = append(securityRules, securityRule)
 	}
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -1320,7 +1320,7 @@ func (e *environ) destroyControllerManagedEnvirons(ctx context.ProviderCallConte
 		// (anastasiamac 2018-03-21) This is strange - we do try
 		// to destroy all volumes but afterwards, if we have encountered any errors,
 		// we will return first one...The same logic happens on detach..?...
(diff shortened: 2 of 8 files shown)


provider/maas/environ.go
```diff
@@ -1723,7 +1723,7 @@ func (environ *maasEnviron) filteredSubnets(ctx context.ProviderCallContext, nod
 		}
 		subnetIdFloat, err := fields["id"].GetFloat64()
 		if err != nil {
-			return nil, errors.Annotatef(err, "cannot get subnet Id: %v")
+			return nil, errors.Annotatef(err, "cannot get subnet Id")
```
anastasiamac: This can probably be just Annotate(

provider/ec2/environ.go
```diff
@@ -1320,7 +1320,7 @@ func (e *environ) destroyControllerManagedEnvirons(ctx context.ProviderCallConte
 		// (anastasiamac 2018-03-21) This is strange - we do try
 		// to destroy all volumes but afterwards, if we have encountered any errors,
 		// we will return first one...The same logic happens on detach..?...
-		return errors.Annotatef(err, "destroying volume %q", volIds[i], err)
+		return errors.Annotatef(err, "destroying volume %q", volIds[i])
```
anastasiamac: Oh wow, yeah.. good call :)

provider/azure/instance.go
```diff
@@ -283,7 +283,7 @@ func (inst *azureInstance) OpenPorts(ctx context.ProviderCallContext, machineId
 			nil, // abort channel
 		)
 		if err := <-errCh; err != nil {
-			return errors.Annotatef(err, "creating security rule for %s", securityRule)
+			return errors.Annotatef(err, "creating security rule for %+v", securityRule)
```
anastasiamac: Why would we want %+v here? What was wrong with %s?

worker/caasoperatorprovisioner/worker.go
```diff
@@ -114,7 +114,7 @@ func (p *provisioner) loop() error {
 			}
 			for app, password := range p.appPasswords {
 				if err := p.ensureOperator(app, password); err != nil {
-					return errors.Annotatef(err, "updating operator for q with new api addresses", app)
+					return errors.Annotatef(err, "updating operator for %q with new api addresses", app)
```
anastasiamac: Ha :) interesting.. thank you for the pick up!

provider/maas/environ.go
hmlanigan: updated

provider/azure/instance.go
hmlanigan: provider/azure/instance.go:286: Annotatef format %s has arg securityRule of wrong type github.com/Azure/azure-sdk-for-go/arm/network.SecurityRule

Looks like network.SecurityRule doesn't implement String()

hmlanigan: i ended up changing this one to %s and ruleName, matched better with the other log and error messages.

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #9179, opened 2018-09-08
fix build errors found with go 1.11 test

hmlanigan:
## Description of change

Fix build errors found with running go test with go version 1.11. 

## QA steps

Have branch downloaded and dependencies setup.
Install go 1.11
Edit Makefile to not run pre-check when check is run.
make check   <-- intermittent unit test failures are possible, but no build failures should be logged.

### DIFF
--- a/provider/maas/environ.go
+++ b/provider/maas/environ.go
@@ -1741,7 +1741,7 @@ func (environ *maasEnviron) filteredSubnets(ctx context.ProviderCallContext, nod
 		}
 		subnetIdFloat, err := fields["id"].GetFloat64()
 		if err != nil {
-			return nil, errors.Annotatef(err, "cannot get subnet Id: %v")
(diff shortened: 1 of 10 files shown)


provider/maas/environ.go
```diff
@@ -1741,7 +1741,7 @@ func (environ *maasEnviron) filteredSubnets(ctx context.ProviderCallContext, nod
 		}
 		subnetIdFloat, err := fields["id"].GetFloat64()
 		if err != nil {
-			return nil, errors.Annotatef(err, "cannot get subnet Id: %v")
+			return nil, errors.Annotatef(err, "cannot get subnet Id")
```
wallyworld: Annotate

hmlanigan: updated

hmlanigan:
$$merge$$

PR #9180, opened 2018-09-09
upgrader: Check available space before downloading binaries

babbageclunk:
## Description of change

If the disk was full when trying to download the agent binary for an upgrade, Juju would log fairly cryptic errors that didn't clearly spell out the problem. Instead of allowing the download or unpacking to just fail, we now check for 250MiB available space before beginning. If there's not enough space the error message logged is:
```
not enough free disk space on "/var/lib/juju" for upgrade: 103MiB available, require 250MiB
```

The free space check is done for the Juju data directory as well as wherever the OS temp storage is (for unpacking).

This exposes and uses `upgrades.CheckFreeDiskSpace` so it can be used from the worker. The function now uses `DiskUsage.Available()`, because in manual testing I couldn't ever get the disk space check to fail using `.Free()` and after some research Available seems like the right choice. I had trouble finding information about the difference between the fields in the underlying `statfs` syscall - the man page just says `f_bfree: free blocks in filesystem` versus `f_bavail: free blocks available to unprivileged user` without any indication of why or in what circumstances you'd use one over the other. [This discussion](https://unix.stackexchange.com/questions/7950/reserved-space-for-root-on-a-filesystem-why) has some detail about superuser-reserved space that suggests it's really/mostly for emergency recovery in nearly-full disk situations. That indicates that Available is the right field for this check - we shouldn't be filling up emergency space with downloaded agent binaries just because the agent is running as root.

## QA steps

* Deploy an application, on a machine in the application use `fallocate -l <size> bigfile` to create a file that fills *almost* all the available space reported by `df -h` - so that there's less than 250MiB free.
* Run `juju upgrade-juju -m controller --build-agent && sleep 20 && juju upgrade-juju ` to upgrade the model.
* Look in the model log, you should see the uniter on the full machine complaining that there's not enough space in /var/lib/juju.
* Delete bigfile; the upgrade should continue and succeed.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1782367

### DIFF
--- a/upgrades/preupgradesteps.go
+++ b/upgrades/preupgradesteps.go
@@ -45,10 +45,10 @@ var MinDiskSpaceMib = uint64(250)
 // containing dir.
 func CheckFreeDiskSpace(dir string, thresholdMib uint64) error {
 	usage := du.NewDiskUsage(dir)
-	free := usage.Free()
-	if free < thresholdMib*humanize.MiByte {
-		return errors.Errorf("not enough free disk space for upgrade at %q: %s available, require %dMiB",
-			dir, humanize.IBytes(free), thresholdMib)
+	available := usage.Available()
+	if available < thresholdMib*humanize.MiByte {
+		return errors.Errorf("not enough free disk space on %q for upgrade: %s available, require %dMiB",
+			dir, humanize.IBytes(available), thresholdMib)
 	}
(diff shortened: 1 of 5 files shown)


babbageclunk:
$$merge$$

PR #9183, opened 2018-09-10
Go vet

SimonRichardson:
## Description of change

The following commit ensures that we handle annotatef correctly in
verify.bash file.


hmlanigan:
$$merge$$

PR #9185, opened 2018-09-11
Wire up caas operator logging to controller

wallyworld:
## Description of change

As well as logging to standard k8s logger, we also send logs to the controller. This conforms with how other agents are expected to operate.

## QA steps

deploy k8s charm
run juju debug-log on the k8s model

### DIFF
--- a/worker/caasoperator/caasoperator.go
+++ b/worker/caasoperator/caasoperator.go
@@ -17,7 +17,6 @@ import (
 	"github.com/juju/utils/arch"
 	"github.com/juju/utils/symlink"
 	"github.com/juju/version"
-	"gopkg.in/juju/charm.v6"
 	"gopkg.in/juju/names.v2"
 	"gopkg.in/juju/worker.v1"
 	"gopkg.in/juju/worker.v1/catacomb"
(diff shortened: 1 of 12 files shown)


worker/caasoperator/caasoperator.go
```diff
@@ -429,7 +429,7 @@ func charmModified(local LocalState, remote remotestate.Snapshot) bool {
 
 func (op *caasOperator) ensureCharm() (*charm.URL, int, error) {
 	charmDir := op.paths.GetCharmDir()
-	if _, err := os.Stat(charmDir); !os.IsNotExist(err) {
+	if _, err := os.Stat(charmDir); err != nil && !os.IsNotExist(err) {
```
babbageclunk: Do you really need the `err != nil` here - doesn't `os.IsNotExist` return false for nil? I tried it in the playground and it worked.

wallyworld: !os.IsNotExist(err) returns true for nil err and thus returns without doing the download. and worse, the returned error is nil and so the caller thinks there's nothing wrong but actually nothing was downloaded

babbageclunk: Ok, I get this now you explained it in IRC.

wallyworld: Actually,the original code was correct after all. The operator was too simplistic in how it determined if a charm already existed. I needed to adopt the code already used in the IAAS uniter.

wallyworld:
$$merge$$

wallyworld:
[fix operator charm download - use code from uniter, add operator state](https://trello.com/c/GffhSfle/448-fix-operator-charm-download-use-code-from-uniter-add-operator-state)

wallyworld:
[wire up k8s operator logging to log to controller](https://trello.com/c/Y8TQ9YRZ/447-wire-up-k8s-operator-logging-to-log-to-controller)

PR #9187, opened 2018-09-12
Add support for deploying k8s charms with constraints

wallyworld:
## Description of change

Deploying k8s charms now supports using constraints. The supported constraints are "mem" and "cpu-power". cpu-power is milli-cpu units.

As a driveby, consolidate the checking logic in state to validate deployment args for iaas and caas models.

## QA steps

juju deploy a k8s charm with constrains, eg --constraints "mem=1G cpu-power=500"
look at the pod and check that the resource limits are correctly applied

### DIFF
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -698,6 +698,16 @@ func (k *kubernetesClient) EnsureService(
 			return errors.Annotatef(err, "configuring devices for %s", appName)
 		}
 	}
+	if mem := params.Constraints.Mem; mem != nil {
+		if err = k.configureConstraint(unitSpec, "memory", fmt.Sprintf("%dMi", *mem)); err != nil {
+			return errors.Annotatef(err, "configuring memory constraint for %s", appName)
+		}
+	}
+	if cpu := params.Constraints.CpuPower; cpu != nil {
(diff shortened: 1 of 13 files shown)


caas/kubernetes/provider/k8s.go
```diff
@@ -1634,6 +1656,22 @@ func mergeDeviceConstraints(device devices.KubernetesDeviceParams, resources *co
 	return nil
 }
 
+func mergeConstraint(constraint string, value string, resources *core.ResourceRequirements) error {
+	if resources.Limits == nil {
+		resources.Limits = core.ResourceList{}
+	}
+	resourceName := core.ResourceName(constraint)
+	if v, ok := resources.Limits[resourceName]; ok {
```
ycliuhw: should we use `Limit` or `Request`?

ycliuhw: as discussed, we only set max for constraints which is Limit.
Awesome, thanks!

wallyworld: It's Limit here because we are setting the max mem/cpu.
We don't yet have a way to specify both using Juju constraint syntax.

wallyworld:
$$merge$$

wallyworld:
[k8s constraints](https://trello.com/c/tQuGWkYS/449-k8s-constraints)

PR #9190, opened 2018-09-13
Support --to placement for k8s charms

wallyworld:
## Description of change

Add placement support for k8s charms. Initially only node selectors are supported since affinity is still in beta.
The k8s tests have also been tweaked to reduce the cut and paste of code.

## QA steps

deploy a k8s charm with placement, eg deploy mariadb-k8s --to kubernetes.io/hostname=host
check the pod is started and the nodeSelector is correctly set

### DIFF
--- a/caas/kubernetes/provider/precheck.go
+++ b/caas/kubernetes/provider/precheck.go
@@ -1,0 +1,55 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package provider
+
+import (
+	"strings"
+
+	"github.com/juju/errors"
+	"github.com/juju/juju/environs"
+	"github.com/juju/juju/environs/context"
+	"github.com/juju/utils/keyvalues"
+)
+
+// PrecheckInstance performs a preflight check on the specified
+// series and constraints, ensuring that they are possibly valid for
+// creating an instance in this model.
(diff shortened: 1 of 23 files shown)


caas/kubernetes/provider/precheck.go
```diff
@@ -0,0 +1,54 @@
...
+	if params.Placement == "" {
+		return nil
+	}
+
+	// Check placement is valid.
+	// Placement is a comma separated list of key-value pairs (node labels).
+	_, err = keyvalues.Parse(strings.Split(params.Placement, ","), false)
+	if err != nil {
+		return errors.NotValidf("placement directive %q", params.Placement)
+	}
```
ycliuhw: Nice. we could fetch the current valid list of node labels to do check further here in the future.

wallyworld: Indeed. Although what's a valid list? I guess we could query for all labels on all nodes?

wallyworld:
$$merge$$

caas/kubernetes/provider/precheck.go
wallyworld: I added a TODO to look at what we might be able to do

ycliuhw: yeah, a list of available nodes. Awesome!

PR #9192, opened 2018-09-13
raftlease: Upgrade step to migrate legacy leases into raft store

babbageclunk:
## Description of change

When upgrading from 2.4.x to 2.5, the existing leases in the mongo store need to be copied into the raft store so that unit leadership will be the same after the upgrade. We read the legacy leases from the mongo store and construct a snapshot with those leases in raftlease.FSM format and write that into the raft snapshot store. When the raft nodes start the snapshot is loaded and the lease state matches what it was before the upgrade. We also update the leaseholders collection in mongo so that the DB is consistent.

## QA steps

* Bootstrap a 2.4 controller. 
* Deploy several application with multiple units. 
* Remove the /0 unit for each application so that the leadership isn't the default (just in case).
* Upgrade to 2.5 with this change.
* Application leadership should stay the same across the upgrade.
* Upgrade again to 2.5 to ensure that the upgrade step is idempotent.

### DIFF
--- a/state/upgrades.go
+++ b/state/upgrades.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/juju/juju/cloud"
 	"github.com/juju/juju/controller"
+	corelease "github.com/juju/juju/core/lease"
 	"github.com/juju/juju/core/status"
 	"github.com/juju/juju/environs/config"
 	"github.com/juju/juju/mongo/utils"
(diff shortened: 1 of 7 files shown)


babbageclunk:
[upgrade step to maintain leases going from 2.4 to 2.5 with raft](https://trello.com/c/DQ3Fjf31/434-upgrade-step-to-maintain-leases-going-from-24-to-25-with-raft)

babbageclunk:
!!build!!

babbageclunk:
Hmm, looks like an intermittent failure, but it's in a test for upgrades. :/
!!build!!

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #9196, opened 2018-09-14
Add LXDProfile to state/charm.

hmlanigan:
## Description of change

Add LXDProfile to state/charm.

## QA steps

Juju deploy the lxd-profile charm (testcharms/charm-repo/quantal/lxd-profile) , check the mongo db on the controller:
db.charms.find({},{"lxd-profile":1})

### DIFF
--- a/state/charm.go
+++ b/state/charm.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/juju/errors"
 	jujutxn "github.com/juju/txn"
+	"github.com/juju/utils/featureflag"
 	"github.com/juju/version"
 	"gopkg.in/juju/charm.v6"
 	"gopkg.in/juju/names.v2"
@@ -18,6 +19,7 @@ import (
 	"gopkg.in/mgo.v2/txn"
 
 	"github.com/juju/juju/core/model"
+	"github.com/juju/juju/feature"
 	"github.com/juju/juju/mongo"
 	mongoutils "github.com/juju/juju/mongo/utils"
 	"github.com/juju/juju/state/storage"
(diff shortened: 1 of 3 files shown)


hmlanigan:
[add charm's lxd profile to state.](https://trello.com/c/qxUbPkqi/585-add-charms-lxd-profile-to-state)

state/charm.go
```diff
@@ -221,6 +231,13 @@ func updateCharmOps(mb modelBackend, info CharmInfo, assert bson.D) ([]txn.Op, e
 		{"pendingupload", false},
 		{"placeholder", false},
 	}
+	if featureflag.Enabled(feature.LXDProfile) {
+		lpc, ok := info.Charm.(charm.LXDProfiler)
+		if !ok {
+			return nil, errors.New("charm doesn't have LXDCharmProfile()")
```
SimonRichardson: I think if the `info.Charm` doesn't have a LXDProfile method, we should just ignore it and assume that it doesn't need to store the profile in the document.

state/charm.go
```diff
@@ -114,6 +117,13 @@ func insertCharmOps(mb modelBackend, info CharmInfo) ([]txn.Op, error) {
 		BundleSha256: info.SHA256,
 		StoragePath:  info.StoragePath,
 	}
+	if featureflag.Enabled(feature.LXDProfile) {
+		lpc, ok := info.Charm.(charm.LXDProfiler)
+		if !ok {
+			return nil, errors.New("charm does no implment LXDProfiler")
```
SimonRichardson: Not sure about this, see below

hmlanigan:
$$merge$$

SimonRichardson:
$$merge$$

state/charm.go
rogpeppe: I *think* I agree with Simon here. Maybe best to accept the old interface too - there can be many Charm implementations, potentially, and they don't all have to implement LXDProfile.

rogpeppe: Also, do you actually want an lxd-profile field in the document when LXDProfile returns nil? (that might be best from a backward-compatibility point of view, because null is different from not-exists in mongo)

state/charm.go
```diff
@@ -314,6 +331,35 @@ func safeConfig(ch charm.Charm) *charm.Config {
 	return escapedConfig
 }
 
+// safeLXDProfile ensures that the LXDProfile that we put into the mongo data
+// store, can in fact store the profile safely by escaping mongo-
+// significant characters in config options.
```
rogpeppe: BTW maps containing mongo-special characters are actually fine to put in documents unless you want to search on them. We've been doing it for a long time and there's no real issue with it AFAIK.

PR #9199, opened 2018-09-17
New test to check public-clouds in streams matches built in.

Veebers:
## Description of change

Add a simple test in tree that checks that public-clouds that is built into Juju matches what is in streams.


Veebers:
[Add test for public-clouds](https://trello.com/c/bP5RrnDk/26-add-test-for-public-clouds)

PR #9201, opened 2018-09-17
Fix tests to not rely on event coalescing.

howbazaar:
Two of the uniter api tests were relying on multiple document changes being coallesced into
a single change. This cannot be guaranteed, so the tests have been updated.


howbazaar:
$$merge$$

babbageclunk:
Should be "coalescing"

PR #9205, opened 2018-09-17
Fix the race in worker lease where the retryingTick doesn't finish.

howbazaar:
There was a race in the retryingTick method where the worker may be asked to shut down between when we checked and when we tried to send the error down the channel. The symptom was the testing timing out after 25 minutes.

This branch removes the naked channel send.


howbazaar:
$$merge$$

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #9206, opened 2018-09-17
Fix a race in state watcher makeControllerIdFilter

babbageclunk:
## Description of change

We've been seeing an intermittent failure in featuretests.upgraderSuite. Running it under the stress-race script revealed a data race. This *might* be the cause of the failure - I'm not clear how, but at least if the race is fixed and the failure still happens we know it wasn't the cause.

The returned filter closure updates a closed-over variable to keep track of the last machines seen (for cases when st.ControllerInfo fails - I'm not sure under what circumstances that happens but it seems important to maintain that behaviour). The function can be called from multiple goroutines so assigning to machines in the enclosing scope triggers a data race.

Rather than adding a lock to the closure, it seemed clearer to rewrite the filter as a struct with a match method.

## QA steps

Running featuretests.upgraderSuite under stress-race doesn't complain anymore.

### DIFF
--- a/state/watcher.go
+++ b/state/watcher.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/juju/collections/set"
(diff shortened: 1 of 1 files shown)


babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #9209, opened 2018-09-18
Update credentials with models, apiserver.

anastasiamac:
## Description of change

'update-credential' command has been updating credential content stored on the controller without checking if the models that use this credential are still accessible.

This PR adds the check for the models at the apiserver layer. Once this lands, api- and cmd-layers need to be adjusted to cater for a new facade version and method signature. The signature of the call had to be changed to cater for the possibility that each updated credential may have more than one error as we can now err out per model. The errors that we can get from a model can also be related to individual machines and instances, so there is a possibility to get more than one error per model. This means that previous approach of an error-per-credential no longer fit. New signature caters for possibility of having more than one error per credential.

This PR also changes how Cloud facade versions are bumped. Previous approach worked well when we only wanted to add new methods to the facade but did not cater for the scenario here where we are changing a signature of the existing method, i.e. previous versions of Cloud facade still have the method but with a different signature.

### DIFF
--- a/apiserver/facades/client/cloud/cloud.go
+++ b/apiserver/facades/client/cloud/cloud.go
@@ -12,6 +12,7 @@ import (
 	"gopkg.in/juju/names.v2"
 
 	"github.com/juju/juju/apiserver/common"
+	"github.com/juju/juju/apiserver/common/credentialcommon"
 	"github.com/juju/juju/apiserver/facade"
 	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/cloud"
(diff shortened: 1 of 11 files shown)


anastasiamac:
[update credential, apiserver: need to check validity for models that use it](https://trello.com/c/8wRmcmkn/409-update-credential-apiserver-need-to-check-validity-for-models-that-use-it)

anastasiamac:
[update-credential, apiserver: check models that use this cred](https://trello.com/c/d3lUBqWf/25-update-credential-apiserver-check-models-that-use-this-cred)

anastasiamac:
unrelated intermittent failure in uniter tests

anastasiamac:
!!build!!

anastasiamac:
another unrelated intermittent

anastasiamac:
!!build!!

apiserver/facades/client/cloud/cloud.go
```diff
@@ -21,56 +22,105 @@ import (
 	"github.com/juju/juju/state"
 )
 
-type CloudV1 interface {
+// CloudV3 defines the methods on the cloud API facade, version 3.
+type CloudV3 interface {
```
babbageclunk: I'm not sure why these interfaces exist - maybe just for documentation purposes? But I think the change you've made to them makes sense, given the signature change.

anastasiamac:
$$merge$$

anastasiamac:
intermittent, unrelated failures in
    github.com/juju/juju/worker/caasoperator.TestPackage
    github.com/juju/juju/worker/peergrouper.TestPackage
    github.com/juju/juju/worker/provisioner.TestPackage
as per http://ci.jujucharms.com/job/github-merge-juju-juju/20/

anastasiamac:
$$merge$$

PR #9211, opened 2018-09-18
Wait for the uniter to start before checking the methods.

howbazaar:
Fix the intermittent test failure. There was a race in the behaviour where if the unit change is acted upon before the methods where checked, the test would fail with an extra call to "Life". Now we wait for the uniter to have been started before we check the calls.


howbazaar:
$$merge$$

PR #9213, opened 2018-09-18
Deal with the entity construction events before testing the watchers.

howbazaar:
There were intermittent test failures due to entities beinng created in setup, or in the test function itself, and the creation events were not consumed before the watchers were created.


wallyworld:
$$merge$$

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #9215, opened 2018-09-18
Add some synchronisation points for the JujuConnSuite.

howbazaar:
Since the JujuConnSuite uses a wall clock to drive the txn poller, the StartSync method now does nothing to the txn poller. There are some tests though that still need to await for txn polling to have been completed before continuing.  This branch adds two different synchronisation methods to the base JujuConnSuite.

`WaitForNextSync` will return when the txn poller has next finished a pass over the txns.logs collection.
`WaitForModelWatchersIdle` waits for all the current watchers on the specified State object for the model to have finished processing any events that had been sent from the txn poller.

### DIFF
--- a/juju/testing/conn.go
+++ b/juju/testing/conn.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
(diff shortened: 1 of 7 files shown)


howbazaar:
$$merge$$

PR #9217, opened 2018-09-18
Make sure we don't ignore the error

SimonRichardson:
## Description of change

This is a drive by fix, whilst creating another PR (to follow)

The following just makes sure that we highlight the error when we
validate the series when deploying. As the err in here would be
nil, so err2 should be outputted instead of a nil err.

## QA steps

N/A


SimonRichardson:
$$merge$$

PR #9219, opened 2018-09-18
Write lxd profile to lxd when found in a charm

hmlanigan:
## Description of change

Add the lxd profile from a charm to lxd on the machine for use. 

## QA steps

1. export JUJU_DEV_FEATURE_FLAGS=lxd-profile
2. juju bootstrap localhost
3. juju deploy ./testcharms/charm-repo/quantal/lxd-profile
4. lxc profile show juju-default-lxd-profile-0
profile should exist and be a superset (without comments) as what's in ./testcharms/charm-repo/quantal/lxd-profile/lxd-profile.yaml, adding name and used_by.
5. unit should deploy per normal (it won't use the new profile yet)

### DIFF
--- a/apiserver/facades/agent/provisioner/provisioninginfo.go
+++ b/apiserver/facades/agent/provisioner/provisioninginfo.go
@@ -11,6 +11,7 @@ import (
 	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 	"github.com/juju/os/series"
+	"github.com/juju/utils/featureflag"
 	"gopkg.in/juju/names.v2"
 
 	"github.com/juju/juju/apiserver/common"
@@ -21,6 +22,8 @@ import (
 	"github.com/juju/juju/environs/imagemetadata"
 	"github.com/juju/juju/environs/simplestreams"
 	"github.com/juju/juju/environs/tags"
+	"github.com/juju/juju/feature"
+	"github.com/juju/juju/provider/lxd"
 	"github.com/juju/juju/state"
 	"github.com/juju/juju/state/cloudimagemetadata"
 	"github.com/juju/juju/state/multiwatcher"
@@ -79,6 +82,16 @@ func (p *ProvisionerAPI) getProvisioningInfo(m *state.Machine, env environs.Envi
 	subnetsToZones, err := p.machineSubnetsAndZones(m)
 	if err != nil {
 		return nil, errors.Annotate(err, "cannot match subnets to zones")
+	}
+
+	// TODO: lxd-profile 2018-09-18
+	// Add withoutControllerSuite.TestProvisioningInfoWithLXDProfile when
+	// lxd profiles added to ProvisioningInfo.
+	if featureflag.Enabled(feature.LXDProfile) {
+		_, err = p.machineLXDProfiles(m, env)
+		if err != nil {
+			return nil, errors.Annotate(err, "cannot write lxd profiles")
+		}
 	}
 
 	endpointBindings, err := p.machineEndpointBindings(m)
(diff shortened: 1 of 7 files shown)


provider/lxd/server.go
```diff
@@ -52,6 +52,7 @@ type Server interface {
 	CreateProfileWithConfig(string, map[string]string) error
 	GetProfile(string) (*lxdapi.Profile, string, error)
 	HasProfile(string) (bool, error)
+	CreateProfile(post lxdapi.ProfilesPost) (err error)
```
hmlanigan: Help needed here pls: I ran mock gen after adding this, however I had to edit by hand, since it created a circular reference and wouldn't work.  Not sure why.  MockServer.CreateProfile() and MockServerMockRecorder.CreateProfile() work find.

hmlanigan:
!!build!!

provider/lxd/server.go
SimonRichardson: :eyes:

SimonRichardson: So I ran `go generate .` from inside `provider/lxd` and although it rearranged the mock file slightly, it still allowed me to run `go test ...` fine.

Two questions: 
 1. what gomock have you installed atm
 1. can you put up the file that it did generate?

apiserver/facades/agent/provisioner/provisioninginfo.go
```diff
@@ -302,6 +315,39 @@ func (p *ProvisionerAPI) machineSubnetsAndZones(m *state.Machine) (map[string][]
...
+func (p *ProvisionerAPI) machineLXDProfiles(m *state.Machine, env environs.Environ) ([]string, error) {
+	profileEnv, ok := env.(lxd.LXDProfiler)
+	if !ok {
+		return nil, nil
+	}
+	units, err := m.Units()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	names := make([]string, 0)
```
SimonRichardson: nit: `var names []string`

apiserver/facades/agent/provisioner/provisioninginfo.go
```diff
@@ -302,6 +315,39 @@ func (p *ProvisionerAPI) machineSubnetsAndZones(m *state.Machine) (map[string][]
...
+		}
+		ch, _, err := app.Charm()
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+		profile := ch.LXDProfile()
+		if profile == nil || (profile != nil && profile.Empty()) {
+			continue
+		}
+		pName := fmt.Sprintf("juju-%s-%s-%d-%d", profileEnv.Name(), app.Name(), ch.Revision(), unit.UnitTag().Number())
```
SimonRichardson: I think we can drop `profileEnv.Name()` here, as the `lxd-profile.yaml` will be the same for name and revision, even if it's on a different provider. They'll be on differing models if you do want to identify between the different providers.

provider/lxd/environ.go
```diff
@@ -104,11 +105,36 @@ func (env *environ) profileName() string {
 	return "juju-" + env.Name()
 }
 
+type LXDProfiler interface {
+	MaybeWriteLXDProfile(pName string, put *charm.LXDProfile) error
+	Name() string
```
SimonRichardson: So I think we can drop the name if the charm name and revision are unique enough.

provider/lxd/environ.go
```diff
@@ -104,11 +105,36 @@ func (env *environ) profileName() string {
...
 	return env.name
 }
 
+// MaybeWriteLXDProfile, write given LXDProfiler to if not already there.
+func (env *environ) MaybeWriteLXDProfile(pName string, put *charm.LXDProfile) error {
+	hasProfile, err := env.server.HasProfile(pName)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if hasProfile {
```
SimonRichardson: Can we guarantee that this is the same Profile that we're putting up, does it harm anything if we push this again, but maybe log out (trace/info level) that the profile already exists?

provider/lxd/environ.go
```diff
@@ -104,11 +105,36 @@ func (env *environ) profileName() string {
 	return "juju-" + env.Name()
 }
 
+type LXDProfiler interface {
```
SimonRichardson: I think this should be in the [environs](https://github.com/juju/juju/blob/develop/environs/interface.go) package?

apiserver/facades/agent/provisioner/provisioninginfo.go
hmlanigan: So Name() gives us the model name. Unless there's a different way in the provisioner code to get it.

provider/lxd/environ.go
hmlanigan: The challenge is the case were the same charm/revision is deployed in to different models.  Howe do we differentiate?

apiserver/facades/agent/provisioner/provisioninginfo.go
mitechie: A machine can only appear in one model. I'm confused why we're concerned with cross model differentiation.

hmlanigan: @mitechie, lxd doesn't know about models.  once we ask lxd to create a profile, it's name must be such that juju can determine which model the profile should be used in.

provider/lxd/server.go
hmlanigan: updating mockgen, as recommended, worked.

apiserver/facades/agent/provisioner/provisioninginfo.go
hmlanigan: done

provider/lxd/environ.go
hmlanigan: dropping Name(), found a different way to get the info in provisioninginfo.go

provider/lxd/environ.go
hmlanigan: LXD errors out if you attempt to create a profile that already exists, logging a message if hasProfile is true.

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #9220, opened 2018-09-18
Use charm defined lxd profile when starting lxd machine

hmlanigan:
## Description of change

Use charm defined lxd profile when starting lxd machine

## QA steps

1. export JUJU_DEV_FEATURE_FLAGS=lxd-profile
2. juju bootstrap localhost
3. juju deploy ./testcharms/charm-repo/quantal/lxd-profile
4. wait for deploy to finish, get container name
4. lxc profile show juju-default-lxd-profile-0-0
profile should exist and be a superset (without comments) as what's in ./testcharms/charm-repo/quantal/lxd-profile/lxd-profile.yaml, adding name and used_by.  Container name from step above should be listed in used_by.
5. unit should deploy per normal, charm profile is used.

### DIFF
--- a/provider/lxd/environ.go
+++ b/provider/lxd/environ.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/juju/errors"
 	"github.com/lxc/lxd/shared/api"
+	"gopkg.in/juju/charm.v6"
 
 	"github.com/juju/juju/environs"
 	"github.com/juju/juju/environs/config"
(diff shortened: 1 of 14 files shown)


hmlanigan:
This PR depends on PR 9219 being merged first.  Tests need to be written.

hmlanigan:
[use the charm lxd profiles when provisioning the machine](https://trello.com/c/3l9IGDRc/594-use-the-charm-lxd-profiles-when-provisioning-the-machine)

provider/lxd/environ.go
```diff
@@ -292,3 +293,24 @@ func (env *environ) DeriveAvailabilityZones(
...
 }
+
+// MaybeWriteLXDProfile implements environs.LXDProfiler.
+func (env *environ) MaybeWriteLXDProfile(pName string, put *charm.LXDProfile) error {
+	hasProfile, err := env.server.HasProfile(pName)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if hasProfile {
+		logger.Debugf("lxd profile %q already exists, not written again", pName)
```
SimonRichardson: This is great feedback for a user, so we can at least track this happened if debugging.

provider/lxd/environ_broker.go
```diff
@@ -179,7 +179,7 @@ func (env *environ) getContainerSpec(
 	}
 	cSpec := lxd.ContainerSpec{
 		Name:     hostname,
-		Profiles: []string{"default", env.profileName()},
+		Profiles: append([]string{"default", env.profileName()}, args.CharmLXDProfiles...),
```
SimonRichardson: I was wondering if order matters?

hmlanigan: "Profiles are applied in the order they are specified so the last profile to specify a specific key wins."  https://lxd.readthedocs.io/en/latest/profiles/

SimonRichardson:
!!build!!

mitechie:
QA notes:

I know the quantal is historical reasons. Can we start a new directory? 

On my lxd setup I had v2.0 and it failed with an error about not able to write a profile. I had to remove the 2.0 deb, lxc-common, lxd-client and install the snap to get it to work. We should error/detect that more clearly if we can. 

After that I got 

    failed to start machine 0 (Failed to detect requested GPU device), retrying in 10s (9 more attempts)

If I just juju add-machine I get a container that gets created, started, running, etc. 

I then updated the charm to remove the gpu from the lxd-profile and deployed a new application as lxd-profile2 and that came up running. 

Getting the profile id out of lxc profile list is kind of :/ so adding a card to make that part of the show-machine details.

hmlanigan:
$$merge$$

PR #9221, opened 2018-09-19
Fix the race in the JujuConnSuite.

howbazaar:
Latest CI run showed some very weird data races in tests that looked very unrelated to recent changes.

The cause was the addition of the sync.Mutex to the JujuConnSuite. It appears that gocheck is taking a copy of the Suite somewhere. To avoid the data race we make the mutex a pointer and set it in the setup test.


howbazaar:
$$merge$$

PR #9225, opened 2018-09-19
Add bundleURL annotation to applications deployed from bundles in the charmstore.

hatched:
The Juju GUI will group applications together in the inspector that have been deployed via bundles from the charmstore. To do this it relies on an application annotation `bundleURL`. This branch adds that value when a bundle is deployed via the CLI from the charmstore.

### DIFF
--- a/cmd/juju/application/bundle.go
+++ b/cmd/juju/application/bundle.go
@@ -338,11 +338,16 @@ func (h *bundleHandler) resolveCharmsAndEndpoints() error {
 }
 
 func (h *bundleHandler) getChanges() error {
+	bundleURL := ""
+	if h.bundleURL != nil {
+		bundleURL = h.bundleURL.String()
+	}
 	changes, err := bundlechanges.FromData(
 		bundlechanges.ChangesConfig{
-			Bundle: h.data,
-			Model:  h.model,
-			Logger: logger,
+			Bundle:    h.data,
+			BundleURL: bundleURL,
+			Model:     h.model,
+			Logger:    logger,
 		})
 	if err != nil {
 		return errors.Trace(err)
@@ -618,18 +623,6 @@ func (h *bundleHandler) addApplication(change *bundlechanges.AddApplicationChang
 	}
 	h.writeAddedResources(resNames2IDs)
 
-	if h.bundleURL == nil {
-		return nil
-	}
-
-	tag := names.NewApplicationTag(p.Application).String()
-	result, err := h.api.SetAnnotation(map[string]map[string]string{tag: {"bundleURL": h.bundleURL.String()}})
-	if err == nil && len(result) > 0 {
-		err = result[0].Error
-	}
-	if err != nil {
-		logger.Debugf("error setting bundleURL annotation for %q: %s", p.Application, err)
-	}
 	return nil
 }
 
--- a/cmd/juju/application/bundle.go
+++ b/cmd/juju/application/bundle.go
@@ -185,7 +185,8 @@ type bundleHandler struct {
 	// data is the original bundle data that we want to deploy.
 	data *charm.BundleData
 
-	// The bundle URL from the charmstore.
+	// bundleURL is the URL of the bundle when deploying a bundle from the
+	// charmstore, nil otherwise.
 	bundleURL *charm.URL
 
 	// unitStatus reflects the environment status and maps unit names to their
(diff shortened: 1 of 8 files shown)


cmd/juju/application/bundle.go
```diff
@@ -183,6 +184,9 @@ type bundleHandler struct {
 	// data is the original bundle data that we want to deploy.
 	data *charm.BundleData
 
+	// The bundle URL from the charmstore.
```
frankban: // bundleURL is the URL of the bundle when deploying a bundle from the char store, nil otherwise.
Or similar?

cmd/juju/application/bundle.go
```diff
@@ -601,6 +607,16 @@ func (h *bundleHandler) addApplication(change *bundlechanges.AddApplicationChang
 		return errors.Annotatef(err, "cannot deploy application %q", p.Application)
 	}
 	h.writeAddedResources(resNames2IDs)
+
+	if h.bundleURL != nil {
+		tag := names.NewApplicationTag(p.Application).String()
+		result, err := h.api.SetAnnotation(map[string]map[string]string{tag: {"bundleURL": h.bundleURL.String()}})
+		if err == nil && len(result) > 0 {
+			err = result[0].Error
+			logger.Debugf("error setting bundleURL annotation for %q: %s", p.Application, err)
```
frankban: You need to check the error before logging it.

cmd/juju/application/bundle_test.go
```diff
@@ -1007,6 +1007,19 @@ func (s *BundleDeployCharmStoreSuite) TestDeployBundleApplicationConstrants(c *g
...
+	testcharms.UploadCharm(c, s.client, "xenial/mysql-42", "mysql")
+	testcharms.UploadCharm(c, s.client, "xenial/wordpress-47", "wordpress")
+	testcharms.UploadBundle(c, s.client, "bundle/wordpress-simple-1", "wordpress-simple")
+	err := runDeploy(c, "bundle/wordpress-simple")
+	c.Assert(err, jc.ErrorIsNil)
+	application, err := s.State.Application("wordpress")
+	c.Assert(err, jc.ErrorIsNil)
+	ann, err := s.Model.Annotations(application)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(ann, jc.DeepEquals, map[string]string{"bundleURL": "cs:bundle/wordpress-simple-1"})
```
frankban: Check annotations for mysql as well?

cmd/juju/application/bundle.go
hatched: done

cmd/juju/application/bundle.go
hatched: done, and removed the additional layer of indentation.

cmd/juju/application/bundle_test.go
hatched: done

mitechie:
!!build!!

mitechie:
!!build!!

mitechie:
!!build!!

mitechie:
!!build!!

cmd/juju/application/bundle.go
```diff
@@ -601,6 +608,19 @@ func (h *bundleHandler) addApplication(change *bundlechanges.AddApplicationChang
...
+		return nil
+	}
+
+	tag := names.NewApplicationTag(p.Application).String()
+	result, err := h.api.SetAnnotation(map[string]map[string]string{tag: {"bundleURL": h.bundleURL.String()}})
+	if err == nil && len(result) > 0 {
+		err = result[0].Error
+	}
+	if err != nil {
+		logger.Debugf("error setting bundleURL annotation for %q: %s", p.Application, err)
```
howbazaar: If we are just going to swallow the error rather than stopping, I think we should use ctx.Infof (on the cmd.Context object) rather than logger.Debug.

This means it will be written out to stderr and more likely to be noticed.

cmd/juju/application/deploy.go
```diff
@@ -1200,6 +1202,7 @@ func (c *DeployCommand) maybeReadLocalBundle() (deployFn, error) {
 			ctx,
 			bundleDir,
 			bundleData,
+			nil,
```
howbazaar: Perhaps a comment here like the filepath one below. just saying `// no bundleURL`

howbazaar:
I think by adding this to the bundlechanges library itself, you wouldn't have to have the custom code in the GUI for adding the bundleURL when deploying a bundle as the annotations to set would come back from the bundle changes call into the API server.

hatched:
It looks like the CI failure is from something else

hatched:
!!build!!

hatched:
!!build!!

hatched:
Thanks @howbazaar @frankban 

$$merge$$

PR #9226, opened 2018-09-19
Fix race in caasoperatorprovisioner test

babbageclunk:
## Description of change

Previously the test would wait until there was at least one call on the caasClient stub but then check for two calls. Most of the time the two calls would happen in quick enough succession that the test would pass, but sometimes (especially under load), it would check the calls before the second one was made. 

Change it to wait for two calls - if they never show up it will still fail in the same way. Also add an assertion on the Calls() length to avoid the test panicking (since the CheckCallNames doesn't stop it).

## QA steps

Reran under stress-race with other load on the machine, didn't see the failure again.


babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #9227, opened 2018-09-19
Add sync methods to StateSuite base suite.

howbazaar:
Enforce the processing of pending txn events before starting the watcher.


howbazaar:
$$merge$$

PR #9229, opened 2018-09-19
Fix intermittent failure in api/watcher test

babbageclunk:
## Description of change

This would fail because it was depending on event coalescing that we can't rely on, so those earlier changes were removed. Also changed uses of state.StartSync to s.WaitForModelWatchersIdle - the former no longer does anything useful in the context of JujuConnSuite because it uses a wallclock.

## QA steps

Reran under stress-race with no failure.


babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #9235, opened 2018-09-20
Add support using the newest gomaasapi to support handling partitions result from a storage constraint.

blakerouse:
## Description of change

MAAS 2.5 added support for a storage constraint to reference a partition. Previous versions of MAAS always only referenced a block device.

## QA steps

First add the `part` tag to a partition in MAAS:
```
maas admin partition add-tag gadex8 6 10 tag=part
```

Bootstrap Juju on MAAS
```
juju add-cloud maas ...
juju add-credential maas ...
```

Add the storage pool with required partition tag and the created `part` tag.
```
juju create-storage-pool maas-part maas tags=partition,part
juju deploy postgresql --storage pgdata=maas-part,1G
```

Wait for the magic!

## Documentation changes

Maybe explain that to get a storage pool to match a partition you need to ensure that `partition` is included as a tag in the storage pool creation.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1713239

### DIFF
--- a/provider/maas/maas2_test.go
+++ b/provider/maas/maas2_test.go
@@ -574,23 +574,123 @@ func (f *fakeFile) ReadAll() ([]byte, error) {
 }
 
 type fakeBlockDevice struct {
-	gomaasapi.BlockDevice
+	id     int
+	uuid   string
 	name   string
+	path   string
 	idPath string
+	model  string
+	serial string
 	size   uint64
+}
+
+func (bd fakeBlockDevice) Type() string {
+	return "blockdevice"
+}
+
+func (bd fakeBlockDevice) ID() int {
+	return bd.id
+}
+
+func (bd fakeBlockDevice) UUID() string {
+	return bd.uuid
 }
 
 func (bd fakeBlockDevice) Name() string {
 	return bd.name
 }
 
+func (bd fakeBlockDevice) Path() string {
+	return bd.path
+}
+
 func (bd fakeBlockDevice) IDPath() string {
 	return bd.idPath
+}
+
+func (bd fakeBlockDevice) Model() string {
+	return bd.model
+}
+
+func (bd fakeBlockDevice) Serial() string {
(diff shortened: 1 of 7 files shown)


anastasiamac:
!!build!!

provider/maas/maas2_test.go
```diff
@@ -574,24 +574,124 @@ func (f *fakeFile) ReadAll() ([]byte, error) {
 }
 
 type fakeBlockDevice struct {
-	gomaasapi.BlockDevice
```
wallyworld: You probably don't need to remove this embedded interface (along with the associated changes to implement all the methods on the fake struct). Embedding the interface means that the struct implements the interface without having to provide all the methods. It is ok to do that if the test doesn't call the methods directly (or else you'll get a nil pointer). If the test does call a method or two, you just need to implement those particular ones.

blakerouse: So I originally did that and I was getting a panic segfault because my fakePartition was not implementing the method I actually called.

Just so I can be sure that the whole interface was implemented and I didn't want to worry about getting the segfault, I changed it to be this way.

It was more of a sanity thing, is that okay?

wallyworld: The preference is not to add unneeded methods. We adopt everywhere else the interface embedding approach so please do that here also. You just need to implement the small number of methods actuallyused in order not to get a seg fault.

blakerouse:
I was able to get this working. It was an issue with how juju filtered out partitions from `lsblk`.

I have fully tested this and postgresql charm now formats the partition and places the data on the partition. So this is good to go.

provider/maas/maas2_test.go
blakerouse: I have also changed this back to only implement what is actually required.

wallyworld:
$merge$$

wallyworld:
$$merge$$

PR #9236, opened 2018-09-20
Fix bionic failures

howbazaar:
There were three tests that were always failing on bionic:
 * worker/apiserver
 * worker/httpserver
 * featuretests debugLogDBSuite

The first two were due to the test suites starting a test mongo themselves, after the isolation suite setup.
The isolation suite removed the JUJU_MONGODB env var, which was set by the jenkins job to tell juju where mongo was. The fix is to add the bionic location into the search paths.

The featuretests failure was due to the replication log setup in the test, and it was actually hitting a mongod bug. Effectively our reset method on the MongoServer in the testing package was causing the mongod to hit an inveriant in the second test and it kills the server. We don't know how to fix that, but we can make the tests each do the restart a mongo dance.


howbazaar:
$$merge$$

PR #9240, opened 2018-09-20
Cloud ACLs for add-model etc

wallyworld:
## Description of change

We now maintain an ACL for managing cloud permissions.

Instead of setting add-model permission on a controller, we now use that permission on a cloud.
For backwards compatibility, setting add-model on a controller adds that permission to the controller cloud.

A new grant-cloud and revoke-cloud command is added to manage cloud permissions. This CLI and facade is based on the existing grant/revoke CLI command - the logic is copied across.
When adding a cloud, the user who is adding it automatically gets admin permission on that cloud.

An upgrade step is included to convert older permissions.

There are 4 commits:
1. state changes
2. facade changes
3. CLI changes
4. upgrade steps

## QA steps

bootstrap an older controller
add a user
grant that user add-model permission on the controller
upgrade to this branch
ensure the user can add-model still
add a k8s cloud
add a new user
grant that new user add-model on the k8s cloud
ensure that new user can add a model to the k8s cloud
add a different user
ensure that user cannot add a model to the k8s cloud


wallyworld:
!!build!!

wallyworld:
$$merge$$

PR #9241, opened 2018-09-20
'update-credential' checks models.

anastasiamac:
## Description of change

Updating credential content on the controller, 'update-credential' command, now also checks the models that use this credential.

We provide more detailed output - what models have been checked? are they visible? if not visible, why?

The PR updates adds new facade call - UpdateCredentialsCheckModels - to version 3 of "Cloud" facade, removing "UpdateCredentials".


anastasiamac:
!!build!!

Veebers:
!!build!!

Veebers:
!!build!!

anastasiamac:
re-base went wrong: will re-propose once cloud facade changes land and the area is stable.

PR #9250, opened 2018-09-21
Fix a race in the Uniter test DSL.

howbazaar:
Fix the uniter race where the database was updated for the new charm before the charm had been uploaded.


howbazaar:
$$merge$$

PR #9251, opened 2018-09-21
Fix the missing application series in bundle-export.

howbazaar:
Fix the exporting of the series of the application when exporting a bundle.

This was evident when the CDK bundle was deployed then exported, the applications were missing the series: xenial part, which caused them to fail to deploy on the machines that were exported as xenial.

### DIFF
--- a/apiserver/facades/client/bundle/bundle.go
+++ b/apiserver/facades/client/bundle/bundle.go
@@ -188,11 +188,22 @@ func (b *BundleAPI) ExportBundle() (params.StringResult, error) {
 	}, nil
 }
 
+// bundleOutput has the same top level keys as the charm.BundleData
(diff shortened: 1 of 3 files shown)


apiserver/facades/client/bundle/bundle.go
```diff
@@ -273,6 +297,26 @@ func (b *BundleAPI) fillBundleData(model description.Model) (*charm.BundleData,
 
 		data.Machines[machine.Id()] = newMachine
 	}
+	// If there is only one series used, make it the default and remove
+	// series from all the apps and machines.
+	size := usedSeries.Size()
```
wallyworld: This can be inlined in the switch below

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #9254, opened 2018-09-26
Fix testsuite for bionic

howbazaar:
This is a backport of the fix used in develop for the mongo issues.

The other test failures from develop aren't in this branch AFAICT. Let's see.


howbazaar:
Looks like standard provisioner intermittent failures.

$$merge$$

PR #9255, opened 2018-09-26
Fix the tests for bionic

howbazaar:
Backport of featuretest failure in bionic.

Unfortunately we can't just use the latest juju/testing update that fixes the mongo paths because it pulls in other clock changes that aren't in 2.4


howbazaar:
$$merge$$

PR #9262, opened 2018-09-27
Fix transaction so worker does not suffer from potential infinite loop.

ExternalReality:
## Description of change

This fixes a condition in which the upgrade series machine worker would loop constantly - on each iteration it would make a call that further fuels the loop. Only by chance would the worker be able to break this loop.


ExternalReality:
Its hard to get in there are write a unit test for this without extracting a public method. Open to suggestions to get a little unit test around this.

ExternalReality:
[chase down race during repeated complete runs](https://trello.com/c/n3DxpOil/623-chase-down-race-during-repeated-complete-runs)

ExternalReality:
$$merge$$

PR #9263, opened 2018-09-28
Add back the state of each unit in a relation to goal state

wallyworld:
## Description of change

Goal state output was changed to adhere to spec but in so doing lost some status that people ended up relying on. So we add that back in. The "relations" block now shows relation status plus the status of each unit in the relation. Also, we were showing peer units in the relation section as well as the units section - this has been fixed as well.

## QA steps

Deploy mediawiki and mariadb and relate.

`$ juju run --unit mariadb/0 goal-state`
``` yaml
units:
  mariadb/0:
    status: active
    since: 2018-09-28 00:59:42Z
relations:
  db:
    mediawiki:
      status: joined
      since: 2018-09-28 01:21:25Z
    mediawiki/0:
      status: active
      since: 2018-09-28 01:33:30Z
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1794739

### DIFF
--- a/apiserver/facades/agent/uniter/uniter.go
+++ b/apiserver/facades/agent/uniter/uniter.go
@@ -2559,7 +2559,7 @@ func (u *UniterAPI) goalStateResult(application *state.Application) (*params.Goa
 		return nil, err
 	}
 	if allRelations != nil {
-		gs.Relations, err = u.goalStateRelations(allRelations)
+		gs.Relations, err = u.goalStateRelations(application.Name(), allRelations)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


apiserver/facades/agent/uniter/uniter.go
```diff
@@ -2598,18 +2602,32 @@ func (u *UniterAPI) goalStateRelations(allRelations []*state.Relation) (map[stri
 			} else {
 				return nil, err
 			}
-			goalState := params.GoalStateStatus{}
-			statusInfo, err := r.Status()
-			if err != nil {
-				return nil, errors.Annotate(err, "getting relation status")
+
+			// Peer units are shown in the Units section, not here.
```
babbageclunk: Why doesn't the `e.Relation.Role == "peer"` check above cover this case? Is it because the relation might be a normal provider/requirer one but to the same application? Maybe worth expanding the comment?

wallyworld: The comment was poorly worded. I've updated to make it clearer

wallyworld:
$$merge$$

PR #9264, opened 2018-09-28
Add support using the newest gomaasapi to support handling partitions result from a storage constraint.

mitechie:
Forward port of https://github.com/juju/juju/pull/9235


mitechie:
$$merge$$

PR #9267, opened 2018-09-30
All lock database updates (unit/machine) set machine level timestamp.

ExternalReality:
The Upgrade Series data contains a lock timestamp at the top level and one on each of its units' data representations. All updates, including those affecting only units, now update the top level timestamp.


ExternalReality:
[Update timestamp on top level upgrade series doc on all changes.](https://trello.com/c/9a6Nyrmg/633-update-timestamp-on-top-level-upgrade-series-doc-on-all-changes)

ExternalReality:
@manadart , it is part of a debugging effort. It seems more correct to update the top level timestamp on all changes rather than just on changes to the machine. Since the same timestamps are kept on the internal log messages that get printed out to the CLI on all changes no data is lost but now the top level lock's timestamp reflects the latest change to any part of the document.

ExternalReality:
$$MERGE$$

PR #9268, opened 2018-10-01
LXD: Ensure target is defined firstly.

SimonRichardson:
## Description of change

The following patch is required when attempting to bootstrap into
a lxd cluster, but the container image isn't found. We believe that
the happens because of an optimisation[1] with in lxd which assumes
the instance is the same as the source. Because of the way the methods
are invoked in juju, can lead to the prior bug[2].

[1] https://github.com/lxc/lxd/blob/c3de74dc45c23c177328a8d7c2a197ebcd521b42/client/lxd_containers.go#L176
[2] https://bugs.launchpad.net/juju/+bug/1793291

## QA steps

This assumes you have a cluster setup of at least 3 nodes and that you can 
communicate between them with a trust password. 

You can follow this [documentation](https://discourse.jujucharms.com/t/setting-up-lxd-cluster-as-a-juju-cloud/73)

```
juju bootstrap lxd-cluster --to zone=node3
```
## Bug reference

https://bugs.launchpad.net/juju/+bug/1793291

### DIFF
--- a/provider/lxd/environ_broker.go
+++ b/provider/lxd/environ_broker.go
@@ -113,18 +113,18 @@ func (env *environ) newContainer(
 	}
 	defer cleanupCallback()
 
-	image, err := env.server.FindImage(args.InstanceConfig.Series, arch, imageSources, true, statusCallback)
+	target, err := env.getTargetServer(ctx, args)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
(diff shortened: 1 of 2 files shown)


SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

SimonRichardson:
https://bugs.launchpad.net/juju/+bug/1793291

PR #9269, opened 2018-10-01
Stop units from getting shut down.

ExternalReality:
## Description of change

Units should no longer be shut down during an `upgrade-series prepare`

### DIFF
--- a/worker/upgradeseries/worker.go
+++ b/worker/upgradeseries/worker.go
@@ -200,31 +200,8 @@ func (w *upgradeSeriesWorker) handlePrepareStarted() error {
 	return errors.Trace(w.transitionPrepareMachine(unitServices))
 }
 
-// transitionPrepareMachine stops all unit agents on this machine and updates
-// the upgrade-series status lock to indicate that upgrade work can proceed.
-// TODO (manadart 2018-08-09): Rename when a better name is contrived for
-// "UpgradeSeriesPrepareMachine".
 func (w *upgradeSeriesWorker) transitionPrepareMachine(unitServices map[string]string) error {
-	w.logger.Infof("stopping units for series upgrade")
-
-	for unit, serviceName := range unitServices {
-		svc, err := w.service.DiscoverService(serviceName)
-		if err != nil {
-			return errors.Trace(err)
-		}
-		running, err := svc.Running()
-		if err != nil {
-			return errors.Trace(err)
-		}
-		if !running {
-			continue
-		}
-
-		if err := svc.Stop(); err != nil {
-			return errors.Annotatef(err, "stopping %q unit agent for series upgrade", unit)
-		}
-	}
-	return errors.Trace(w.SetMachineStatus(model.UpgradeSeriesPrepareMachine, "all units stopped for series upgrade"))
+	return errors.Trace(w.SetMachineStatus(model.UpgradeSeriesPrepareMachine, "all unit prepare hooks completed"))
 }
 
 // handlePrepareMachine handles workflow for the machine with an upgrade-series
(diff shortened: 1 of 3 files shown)


worker/upgradeseries/worker.go
```diff
@@ -200,31 +200,8 @@ func (w *upgradeSeriesWorker) handlePrepareStarted() error {
 	return errors.Trace(w.transitionPrepareMachine(unitServices))
 }
 
-// transitionPrepareMachine stops all unit agents on this machine and updates
-// the upgrade-series status lock to indicate that upgrade work can proceed.
-// TODO (manadart 2018-08-09): Rename when a better name is contrived for
-// "UpgradeSeriesPrepareMachine".
 func (w *upgradeSeriesWorker) transitionPrepareMachine(unitServices map[string]string) error {
```
ExternalReality: Function is left in place for consistency rather than in-lining the body.

ExternalReality:
> In fact, "prepare-machine" may not be needed at all if we don't have to shut down units.

AFAICT each state still does something meaningful even if it just checks to see if its known set of prepared/complete units are the complete set. How it does this, by comparing against the init system's set of services is fine but troublesome. The method should be used as a sanity check but perhaps the controller should have the final say on whether all the units are in the desired state or not.

In any case this PR keeps the units from shutting down.

> Actually, I think you have erroneously removed the workflow transition which will break things.

Where has the transition been removed? I still see the call to update the state and the transition method in place. Let me know if I am not seeing something. :-)

ExternalReality:
Without restarting the Uniter (manually or otherwise) the state of the Uniter will remain in the "prepare complete" state and yet when running the `upgrade-series complete` command we must still progress. Since we no longer automatically stop the Uniter then aborting the process with the `upgrade-seroes complete` command without rebooting the machine will leave the process stuck. The update to this PR fixes this.

ExternalReality:
!!build!!

manadart:
I missed that the status setting was only to report the unit shutdown message and that the change between states still occurs in `handlePrepareStarted`. So the workflow will progress OK.

However regarding my other comment, `handlePrepareStarted` and `handlePrepareMachine` now do the same thing. I think we should transition directly to "prepare-completed" from "prepare-started" and remove the "prepare-machine" status altogether. This is a nice simplification and logging reduction given we didn't like the "prepare-machine" status name anyway.

ExternalReality:
[keep agents running during prepare/complete process](https://trello.com/c/KvdrUSVq/622-keep-agents-running-during-prepare-complete-process)

worker/upgradeseries/worker_test.go
```diff
@@ -147,34 +121,6 @@ func (s *workerSuite) expectMachineUnitServiceDiscovery() {
 	s.expectServiceDiscovery(false)
 }
 
-func (s *workerSuite) TestMachinePrepareMachineUnitFilesWrittenProgressPrepareComplete(c *gc.C) {
```
manadart: This shouldn't be deleted. We need to detect that when the machine is "prepare_started" and all units are "prepare-completed", that we invoke the upgrader and progress the machine to "prepare-completed". 

So the names would become `TestMachinePrepareStartedUnitFilesWrittenProgressPrepareComplete` and `expectMachinePrepareStartedUnitFilesWrittenProgressPrepareComplete` and we just change the first expectation to `exp.MachineStatus().Return(model.UpgradeSeriesPrepareStarted, nil)`

worker/upgradeseries/worker_test.go
```diff
@@ -56,32 +56,6 @@ func (s *workerSuite) SetUpTest(c *gc.C) {
 	s.done = make(chan struct{})
 }
 
-// TestFullWorkflow uses the the expectation scenarios from each of the tests
-// below to compose a test of the whole upgrade-series scenario, from start
-// to finish.
-func (s *workerSuite) TestFullWorkflow(c *gc.C) {
```
manadart: I am uncomfortable with deleting this. It is a statement of expected workflow through the standard life cycle of an upgrade and in addition to confirming, it *demonstrates* the expected overarching operation of the worker.

That said it isn't as important to keep as the deleted test below.

ExternalReality:
$$MERGE$$

PR #9270, opened 2018-10-01
Ensure the order of credentials

SimonRichardson:
## Description of change

The following commit ensure the order of the credentials is always
the same. The code assumed that map orders would always be the same
this isn't the case. So we sort the invalid options before rendering
them out to the user.


mitechie:
$$merge$$

PR #9275, opened 2018-10-02
Fix uniter crash edge case

ExternalReality:
This PR fixes an edge case with the `upgrade-series` command. 

If the `uniter` should stop while running its hooks then when it is restarted by the `complete` sub command the uniter would rerun the interrupted hook which would regresses the units state and halt the machine's agent's `upgrade-series` workflow.

Like any program, we cannot prevent the `uniter` form being stopped by outside forces, the only thing we can do is have it recover gracefully when it is restarted. We take two measures to do this:

1. We ensure a unit's remote `upgrade series` state cannot be regressed by the `upgrade-series` hooks
2. We ensure that the hooks update the state as late in the hooks running process as possible to increase the probability that the `uniter` will not fail or be stopped before completing its hook but after updating its remote state.

## QA

run a script like the following for 45 minutes to an hour:

```bash
#!/bin/bash
while :
do
    juju upgrade-series prepare $1 bionic --agree
    juju upgrade-series complete $1
done
```
Its should not encounter the infinite loop condition and get stuck on a `complete` sub command.

If you trust that the issue is caused by the above you can crash the `uniter` with a panic after its state gets updated but before the hook completes - then restart it and see if the workflow recovers properly.

### DIFF
--- a/core/model/upgradeseries.go
+++ b/core/model/upgradeseries.go
@@ -14,12 +14,11 @@ var UpgradeSeriesStatusOrder map[UpgradeSeriesStatus]int = map[UpgradeSeriesStat
 	UpgradeSeriesNotStarted:       0,
 	UpgradeSeriesPrepareStarted:   1,
 	UpgradeSeriesPrepareRunning:   2,
-	UpgradeSeriesPrepareMachine:   3,
-	UpgradeSeriesPrepareCompleted: 4,
-	UpgradeSeriesCompleteStarted:  5,
-	UpgradeSeriesCompleteRunning:  6,
-	UpgradeSeriesCompleted:        7,
-	UpgradeSeriesError:            8,
+	UpgradeSeriesPrepareCompleted: 3,
+	UpgradeSeriesCompleteStarted:  4,
+	UpgradeSeriesCompleteRunning:  5,
+	UpgradeSeriesCompleted:        6,
+	UpgradeSeriesError:            7,
 }
 
 const (
@@ -34,9 +33,9 @@ const (
 	UpgradeSeriesError            UpgradeSeriesStatus = "error"
 )
 
-// ValidateUnitSeriesUpgradeStatus validates a the input status as valid for a
+// ValidateUpgradeSeriesStatus validates a the input status as valid for a
 // unit, returning the valid status or an error.
-func ValidateUnitSeriesUpgradeStatus(status UpgradeSeriesStatus) (UpgradeSeriesStatus, error) {
+func ValidateUpgradeSeriesStatus(status UpgradeSeriesStatus) (UpgradeSeriesStatus, error) {
 	if _, ok := UpgradeSeriesStatusOrder[status]; !ok {
 		return UpgradeSeriesNotStarted, errors.NotValidf("upgrade series status of %q is", status)
 	}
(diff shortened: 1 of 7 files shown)


worker/uniter/operation/runhook.go
```diff
@@ -251,6 +247,12 @@ func (rh *runHook) Commit(state State) (*State, error) {
 			Step: Queued,
 			Hook: hi,
 		}
+	case hooks.PreSeriesUpgrade:
+		message := createUpgradeSeriesStatusMessage(rh.name, rh.hookFound)
+		err = rh.callbacks.SetUpgradeSeriesStatus(model.UpgradeSeriesPrepareCompleted, message)
+	case hooks.PostSeriesUpgrade:
+		message := createUpgradeSeriesStatusMessage(rh.name, rh.hookFound)
+		err = rh.callbacks.SetUpgradeSeriesStatus(model.UpgradeSeriesCompleted, message)
```
ExternalReality: Here we move the updates down to the commit phase and as close to the end of it as possible.

ExternalReality:
[Fix uniter edge case which would stop the Complete phase.](https://trello.com/c/IWFkndCN/642-fix-uniter-edge-case-which-would-stop-the-complete-phase)

core/model/upgradeseries.go
```diff
@@ -33,3 +45,16 @@ func ValidateUnitSeriesUpgradeStatus(status UpgradeSeriesStatus) (UpgradeSeriesS
 	}
 	return UpgradeSeriesNotStarted, errors.NotValidf("unit upgrade series status of %q", status)
 }
+
+// CompareUpgradeSeriesStatus compares two upgrade series statuses and returns and integer; if the first
+// argument equals the second then 0 is returned; if the second is greater -1 is
+// returned; 1 is returned otherwise.
+func CompareUpgradeSeriesStatus(status1 UpgradeSeriesStatus, status2 UpgradeSeriesStatus) int {
```
manadart: Looks like this needs some validation - if one of the status values is not in the map, then the value will be 0. Also we are placing the burden on the caller to interpret what the -1, 0, 1 correspond to.

I was thinking something like:
```go
func (s UpgradeSeriesStatus) ComesBefore(status UpgradeSeriesStatus) (bool, err) {
    i1, ok := UpgradeSeriesStatusOrder[s]
    if !ok {
        return false, errors.Invalidf("status %q", s)
    }

    i2, ok := UpgradeSeriesStatusOrder[status]
    if !ok {
        return false, errors.Invalidf("status %q", status)
    }

    return s < status, nil
}
```
We only really care about not going backwards.

core/model/upgradeseries.go
```diff
@@ -10,6 +10,18 @@ import (
 // UpgradeSeriesStatus is the current status of a series upgrade for units
 type UpgradeSeriesStatus string
 
+var UpgradeSeriesStatusOrder map[UpgradeSeriesStatus]int = map[UpgradeSeriesStatus]int{
```
manadart: I don't think we need to export this.

core/model/upgradeseries.go
```diff
@@ -10,6 +10,18 @@ import (
...
+	UpgradeSeriesNotStarted:       0,
+	UpgradeSeriesPrepareStarted:   1,
+	UpgradeSeriesPrepareRunning:   2,
+	UpgradeSeriesPrepareMachine:   3,
```
manadart: Can we delete this altogether now that it is unused?

core/model/upgradeseries.go
ExternalReality: We export it for testing. I think it is harmless to export this for testing. Otherwise I have to figure out a way to test the ordering comparitor without knowing the order :-)

core/model/upgradeseries.go
ExternalReality: A little foresight never hurt anyone. Yes `UpgradeSeriesPrepareMachine` is being removed. So we can remove it here with no harm done.

core/model/upgradeseries.go
ExternalReality: I just validated them since a validation function exists already.

ExternalReality: There is also the general convention is C derived languages of return -1,0,1 from comparison functions.

ExternalReality:
$$MERGE$$

ExternalReality:
$$MERGE$$

PR #9278, opened 2018-10-03
Reacting to CAAS Operator changes.

Veebers:
## Description of change

React and report on operator pod changes, funnel status changes from the operator back into theJuju data model.

This will allow Juju to surface error message encountered when spinning up an operator for a CAAS charm deploy.

## QA steps

Manually verify the status data is stored in the DB using the mongo console.
(Modified steps taken from 
  - Boostrap and setup a caas model as per [this post on discourse](https://discourse.jujucharms.com/t/storage-support/154).
  - BEFORE deploying the charm, poison the operator image config so that the operator pod will error (needs to be a valid docker url):
    - For instance I use: ```juju controller-config caas-operator-image-path=veebers/caap-operator@sha256:29498cb4c1f89e28b3f45df6c222e3b404586c56f1fe1c79978398e89b3e0309```
  - Use this script ([found on the wiki here]()) to gain access to the mongo console and check the status data is present:
    - ```db.statuses.find().pretty()``` 
    - We're looking for an entry with an ID like: ```<model uuid>:a#mariadb#operator``` and a message like: ```Error: ImagePullBackOff```

If you repeat the above but dont posion the operator-image you will see the status as 'active'.

## Documentation changes

None.

## Bug reference

N/A

### DIFF
--- a/apiserver/params/params.go
+++ b/apiserver/params/params.go
@@ -590,19 +590,6 @@ type UpdateApplicationServiceArg struct {
 	Addresses      []Address `json:"addresses"`
 }
 
-type UpdateApplicationOperatorArgs struct {
-	Args []UpdateApplicationOperatorArg `json:"args"`
-}
-
-// UpdateApplicationServiceArg holds parameters used to update
-// an application's service definition for the cloud.
-type UpdateApplicationOperatorArg struct {
-	ApplicationTag string                 `json:"application-tag"`
-	Status         string                 `json:"status"`
-	Info           string                 `json:"info"`
-	Data           map[string]interface{} `json:"data,omitempty"`
-}
-
 // DestroyApplicationUnits holds parameters for the deprecated
 // Application.DestroyUnits call.
 type DestroyApplicationUnits struct {
--- a/apiserver/params/params.go
+++ b/apiserver/params/params.go
@@ -590,6 +590,19 @@ type UpdateApplicationServiceArg struct {
 	Addresses      []Address `json:"addresses"`
 }
 
+type UpdateApplicationOperatorArgs struct {
+	Args []UpdateApplicationOperatorArg `json:"args"`
+}
+
+// UpdateApplicationServiceArg holds parameters used to update
+// an application's service definition for the cloud.
+type UpdateApplicationOperatorArg struct {
+	ApplicationTag string                 `json:"application-tag"`
+	Status         string                 `json:"status"`
+	Info           string                 `json:"info"`
+	Data           map[string]interface{} `json:"data,omitempty"`
+}
+
 // DestroyApplicationUnits holds parameters for the deprecated
 // Application.DestroyUnits call.
 type DestroyApplicationUnits struct {
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -1423,7 +1423,7 @@ func (k *kubernetesClient) Operator(appName string) (*caas.Operator, error) {
 		return nil, errors.Trace(err)
 	}
 	if len(podsList.Items) == 0 {
-		return nil, errors.NotFoundf("operator pod")
+		return nil, errors.NotFoundf("operator pod for application %q", appName)
 	}
 
 	opPod := podsList.Items[0]
@@ -1488,23 +1488,6 @@ func (k *kubernetesClient) jujuStatus(podPhase core.PodPhase, terminated bool) s
 		return status.Error
 	case core.PodPending:
 		return status.Allocating
-	default:
-		return status.Unknown
-	}
-}
-
-func (k *kubernetesClient) jujuWorkloadStatus(podPhase core.PodPhase, terminated bool) status.Status {
-	if terminated {
-		return status.Terminated
-	}
-	// No comparable status for Waiting.
-	switch podPhase {
-	case core.PodRunning:
-		return status.Active
-	case core.PodPending:
-		return status.Maintenance
-	case core.PodFailed:
-		return status.Blocked
 	default:
 		return status.Unknown
 	}
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -1258,6 +1258,20 @@ func (k *kubernetesClient) WatchUnits(appName string) (watcher.NotifyWatcher, er
 	return newKubernetesWatcher(w, appName)
 }
 
+// WatchOperator returns a watcher which notifies when there
+// are changes to the operator of the specified application.
+func (k *kubernetesClient) WatchOperator(appName string) (watcher.NotifyWatcher, error) {
+	pods := k.CoreV1().Pods(k.namespace)
+	w, err := pods.Watch(v1.ListOptions{
+		LabelSelector: operatorSelector(appName),
+		Watch:         true,
+	})
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	return newKubernetesWatcher(w, appName)
+}
+
 // jujuPVNameRegexp matches how Juju labels persistent volumes.
 // The pattern is: juju-<storagename>-<digit>
 var jujuPVNameRegexp = regexp.MustCompile(`^juju-(?P<storageName>\D+)-\d+$`)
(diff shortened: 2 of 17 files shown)


caas/kubernetes/provider/k8s.go
```diff
@@ -1441,6 +1485,23 @@ func (k *kubernetesClient) jujuStatus(podPhase core.PodPhase, terminated bool) s
...
+		return status.Terminated
+	}
+	// No comparable status for Waiting.
+	switch podPhase {
+	case core.PodRunning:
+		return status.Active
+	case core.PodPending:
+		return status.Maintenance
+	case core.PodFailed:
+		return status.Blocked
```
ycliuhw: not sure if operator status should map to workload status.
especially, should we show `blocked` if operator provision failed

Veebers:
!!build!!

Veebers:
!!build!!

apiserver/params/params.go
```diff
@@ -590,6 +590,19 @@ type UpdateApplicationServiceArg struct {
 	Addresses      []Address `json:"addresses"`
 }
 
+type UpdateApplicationOperatorArgs struct {
```
wallyworld: These params structs aren't needed

caas/kubernetes/provider/k8s.go
```diff
@@ -1425,6 +1413,70 @@ func (k *kubernetesClient) Units(appName string) ([]caas.Unit, error) {
...
+func (k *kubernetesClient) Operator(appName string) (*caas.Operator, error) {
+	pods := k.CoreV1().Pods(k.namespace)
+	podsList, err := pods.List(v1.ListOptions{
+		LabelSelector: operatorSelector(appName),
+	})
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if len(podsList.Items) == 0 {
+		return nil, errors.NotFoundf("operator pod")
```
wallyworld: include the app name in the not found message

caas/kubernetes/provider/k8s.go
```diff
@@ -1441,6 +1493,23 @@ func (k *kubernetesClient) jujuStatus(podPhase core.PodPhase, terminated bool) s
 	}
 }
 
+func (k *kubernetesClient) jujuWorkloadStatus(podPhase core.PodPhase, terminated bool) status.Status {
```
wallyworld: This method is unused and should be deleted

worker/caasunitprovisioner/application_worker.go
```diff
@@ -109,6 +115,13 @@ func (aw *applicationWorker) loop() error {
 				return errors.Annotatef(err, "failed to start unit watcher for %q", aw.application)
 			}
 		}
+		if appOperatorWatcher == nil {
+			appOperatorWatcher, err = aw.containerBroker.WatchOperator(aw.application)
+			if err != nil {
```
wallyworld: Need to do the
strings.Contains(err.Error(), "unexpected EOF") 
check

worker/caasunitprovisioner/application_worker.go
```diff
@@ -189,6 +202,20 @@ func (aw *applicationWorker) loop() error {
 					return errors.Trace(err)
 				}
 			}
+		case _, ok := <-appOperatorWatcher.Changes():
+			if !ok {
```
wallyworld: Need to do the same processing as for unit changes, ie
debug message
if ! ok stop watcher and continue etc

worker/caasunitprovisioner/application_worker.go
```diff
@@ -189,6 +202,20 @@ func (aw *applicationWorker) loop() error {
...
+				return errors.New("operator watcher channel closed")
+			}
+			logger.Debugf("operator update for %v", aw.application)
+			operator, err := aw.containerBroker.Operator(aw.application)
+			if err != nil {
+				return errors.Trace(err)
```
wallyworld: We should probably check for a not found error here. If not found, the best thing to do I think is set the status as Stopped/Terminated

caas/kubernetes/provider/k8s.go
Veebers: Oh shoot, that got lost in my touch up just now, I've re-introduced it so we can translate the Operator status into a juju application workload status.

caas/kubernetes/provider/k8s.go
Veebers: The intent is to translate the operator pod status into something resembling the application workload status (which is where this info will ultimately end up).
Possibly we might want to do that translation at display time instead.

caas/kubernetes/provider/k8s.go
wallyworld: I think we still want to use jujuStatus() at this layer and do any translation (if needed) in the facade.

caas/kubernetes/provider/k8s.go
```diff
@@ -1425,6 +1413,70 @@ func (k *kubernetesClient) Units(appName string) ([]caas.Unit, error) {
...
+func (k *kubernetesClient) Operator(appName string) (*caas.Operator, error) {
+	pods := k.CoreV1().Pods(k.namespace)
+	podsList, err := pods.List(v1.ListOptions{
+		LabelSelector: operatorSelector(appName),
+	})
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if len(podsList.Items) == 0 {
+		return nil, errors.NotFoundf(fmt.Sprintf("operator pod for application %q", appName))
```
wallyworld: NotFoundf doesn't need fmt.Sprintf

worker/caasunitprovisioner/application_worker.go
```diff
@@ -189,6 +206,31 @@ func (aw *applicationWorker) loop() error {
...
+		case _, ok := <-appOperatorWatcher.Changes():
+			if !ok {
+				logger.Debugf("%v", appOperatorWatcher.Wait())
+				worker.Stop(appOperatorWatcher)
+				appOperatorWatcher = nil
+				continue
+			}
+			logger.Debugf("operator update for %v", aw.application)
+			operator, err := aw.containerBroker.Operator(aw.application)
+			if err != nil {
```
wallyworld: We tend to do this to avoid the nesting

```
if errors.IsNotFound(err) {
    blah
} else if err != nil {
    return errors.Trace(err)
}
```

worker/caasunitprovisioner/application_worker.go
```diff
@@ -189,6 +206,31 @@ func (aw *applicationWorker) loop() error {
...
+				if errors.IsNotFound(err) {
+					logger.Debugf("pod not found for application %q", aw.application)
+					err = aw.provisioningStatusSetter.SetOperatorStatus(aw.application, status.Terminated, "", nil)
```
wallyworld: inline the err check here

```
if err := aw.blah.SetOperatorStatus(); err != nil {
    return errors.Trace(err)
}
```

worker/caasunitprovisioner/application_worker.go
```diff
@@ -189,6 +206,31 @@ func (aw *applicationWorker) loop() error {
...
+					if err != nil {
+						return errors.Trace(err)
+					}
+				} else {
+					return errors.Trace(err)
+				}
+			}
+			err = aw.provisioningStatusSetter.SetOperatorStatus(aw.application, operator.Status.Status, operator.Status.Message, operator.Status.Data)
```
wallyworld: ditto

worker/caasunitprovisioner/application_worker.go
```diff
@@ -189,6 +206,27 @@ func (aw *applicationWorker) loop() error {
...
+				logger.Debugf("%v", appOperatorWatcher.Wait())
+				worker.Stop(appOperatorWatcher)
+				appOperatorWatcher = nil
+				continue
+			}
+			logger.Debugf("operator update for %v", aw.application)
+			operator, err := aw.containerBroker.Operator(aw.application)
+			if errors.IsNotFound(err) {
+				logger.Debugf("pod not found for application %q", aw.application)
+				if err := aw.provisioningStatusSetter.SetOperatorStatus(aw.application, status.Terminated, "", nil); err != nil {
```
wallyworld: This will drop out to the code below and set the status again if err is nil
Probs the easiest fix is to put the last SetOperatorStatus() call below inside an else{}

PR #9280, opened 2018-10-04
Improve help text printed by `list-clouds`

pmatulis:
## Description of change

This improves the help text printed after the table of known clouds (`list-clouds`). It is simpler and easier to read.

I removed the list and explanation of non-baked-in clouds as I found that level of detail out of place in this context. I hope I didn't overstep my bounds there. Anyway, I do believe `add-cloud` can add any cloud type (e.g. 'localhost') so the existing text is misleading.

## QA steps

List clouds (`juju clouds`) and look at the text printed at the very end. It should say:

```
 - Import a cloud credential with `add-credential <cloud>`.
 - List a cloud's regions and endpoints with `show-cloud <cloud>`.
 - Add a cloud to Juju with `add-cloud`.
 - Update cloud metadata with `update-clouds`.
```

Instead of:

```
Try 'list-regions <cloud>' to see available regions.
'show-cloud <cloud>' or 'regions --format yaml <cloud>' can be used to see region endpoints.
Update the known public clouds with 'update-clouds'.
'add-cloud' can add private or custom clouds / infrastructure built for the following provider types:
  - lxd, maas, manual, oci, openstack, oracle, vsphere
```

## Documentation changes

Inspect any `list-clouds` (or `clouds`) example sessions and make the necessary changes.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1795996

### DIFF
--- a/cmd/juju/cloud/list.go
+++ b/cmd/juju/cloud/list.go
@@ -6,7 +6,6 @@ package cloud
 import (
 	"io"
 	"sort"
-	"strings"
 
 	"github.com/juju/ansiterm"
 	"github.com/juju/cmd"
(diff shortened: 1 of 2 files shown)


mitechie:
I thought the goal was to move the text to the --help output vs the list-cloud output?

pmatulis:
In retrospect, I should put back that list of non-baked-in clouds but reword it better.

pmatulis:
@mitechie That was my initial goal but my understanding of the bug discussion was that the text be improved instead.

pmatulis:
I will remove the text and make sure --help is stoked.

PR #9281, opened 2018-10-04
Intermittent failure for cloud info output

SimonRichardson:
## Description of change

The following output is never predictable because the results from
an api call is a map[string]... and the loop over the map never
orders the output. The tests then expects the result of the test
to be the same order, this is never the case as a map is used.

## Bug Link

https://bugs.launchpad.net/juju/+bug/1795742


SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #9284, opened 2018-10-08
Fixes lp#1795495: credential check excludes dead models.

anastasiamac:
## Description of change

There is no need to check whether a cloud credential works for a dead model. By extension, it is also not correct to consider that a dead model is part of the model credential collection. This collection should only contain non-dead models.

This PR fixes cloud credential models query and adds a test to verify that a model that became dead is excluded form the result.  

## Bug reference

https://bugs.launchpad.net/juju/+bug/1795495

### DIFF
--- a/state/cloudcredentials.go
+++ b/state/cloudcredentials.go
@@ -353,8 +353,13 @@ func (st *State) CredentialModels(tag names.CloudCredentialTag) (map[string]stri
 	coll, cleanup := st.db().GetCollection(modelsC)
 	defer cleanup()
 
+	sel := bson.D{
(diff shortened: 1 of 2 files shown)


ycliuhw:
lgtm! thanks!

anastasiamac:
$$merge$$

anastasiamac:
merge failed due to an un-related, intermittent failure in github.com/juju/juju/worker/uniter as per http://ci.jujucharms.com/job/make-check-juju/473/testReport/junit/github/com_juju_juju_worker_uniter/TestPackage/

anastasiamac:
$$merge$$

PR #9285, opened 2018-10-08
Fix multiwatcher waitgroup goroutines

howbazaar:
tomb.Go shouldn't be called multiple times from a non-tomb managed goroutine.

Since the watcher is just interested the goroutine to decrement the wait group, it doesn't need the tomb.

## QA steps

unit tests should still pass

## Documentation changes

No doc changes needed

## Bug reference
partial fix for https://bugs.launchpad.net/juju/+bug/1792299


howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #9290, opened 2018-10-09
Model credential setter and watcher.

anastasiamac:
## Description of change

We need the ability to replace model credential reference to allow users to change model credentials.
This PR introduces model credential setter and its watcher in state layer.

### DIFF
--- a/state/modelcredential_test.go
+++ b/state/modelcredential_test.go
@@ -172,14 +172,12 @@ func (s *ModelCredentialSuite) TestWatchModelCredential(c *gc.C) {
 
 	// Initial event.
 	wc.AssertOneChange()
-	wc.AssertNoChange()
 
 	// Check the watcher reacts to credential reference changes.
 	set, err := m.SetCloudCredential(tag)
 	c.Assert(err, jc.ErrorIsNil)
 	c.Assert(set, jc.IsTrue)
 	wc.AssertOneChange()
-	wc.AssertNoChange()
 
 	// Check the watcher does not react to other changes on this model.
 	err = m.SetDead()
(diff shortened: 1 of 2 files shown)


anastasiamac:
[add model credential setter and watcher](https://trello.com/c/bPX3Shcd/272-add-model-credential-setter-and-watcher)

state/modelcredential_test.go
```diff
@@ -87,6 +88,137 @@ func (s *ModelCredentialSuite) TestValidateCloudCredentialModel(c *gc.C) {
...
+	// Initial event.
+	wc.AssertOneChange()
+	wc.AssertNoChange()
+
+	// Check the watcher reacts to credential reference changes.
+	set, err := m.SetCloudCredential(tag)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(set, jc.IsTrue)
+	wc.AssertOneChange()
+	wc.AssertNoChange()
```
wallyworld: AssertNoChange() is not needed as it is called inside AssertOneChange() above

state/modelcredential_test.go
```diff
@@ -87,6 +88,137 @@ func (s *ModelCredentialSuite) TestValidateCloudCredentialModel(c *gc.C) {
...
+	c.Assert(err, jc.ErrorIsNil)
+	w := m.WatchModelCredential()
+	defer statetesting.AssertStop(c, w)
+	wc := statetesting.NewNotifyWatcherC(c, s.State, w)
+
+	// Initial event.
+	wc.AssertOneChange()
+	wc.AssertNoChange()
+
+	// Check the watcher reacts to credential reference changes.
```
wallyworld: Can we also have a check that setting a different attribute on the watched model does not trigger the credential watcher

anastasiamac: Yes, this is done in the same test - we are changing model's life and seeing no changes.

anastasiamac:
!!build!!

anastasiamac:
$$merge$$

PR #9293, opened 2018-10-09
deploy: Fix bundle panic from inconsistent machine map

babbageclunk:
## Description of change

If `--map-machines=existing` is used when deploying a bundle you can get a state where the bundle and model machines are inconsistent, which caused the old version of the bundlechanges package to emit an add-machine change without putting any units on it (because the application has enough units already). This caused a panic in the deploy code, which expected there'd always be an application for an added machine.

Update the bundlechanges package to a new version which detects this inconsistency and reports an error that hopefully explains the problem and what should be done about it - it looks like this:
```
$ juju deploy bundle.yaml --dry-run --map-machines=existing
ERROR cannot deploy bundle: bundle and machine mapping are inconsistent: need an explicit entry mapping bundle machine "1", perhaps to one of model machines ["36", "37"] - the target should host [grafana]
```

Also add a belt-and-suspenders check so we won't panic in the future if for some reason there's still a machine added with no units.

This is a mostly-trivial forward-port of #9291 - it's simpler than the 2.3 change because the `gopkg.in/juju/charm.v6` dependency didn't need to be upgraded (which had a few knock-on effects).

## QA steps

* Run `juju deploy bundle.yaml --map-machines=existing` with a bundle and model in this state (machines in the bundle missing from the model, but all applications having enough units). See the error.
* Add explicit mapping to `map-machines`, error goes away and deploy continues correctly.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1773357

### DIFF
--- a/cmd/juju/application/bundle.go
+++ b/cmd/juju/application/bundle.go
@@ -601,8 +601,13 @@ func (h *bundleHandler) addMachine(change *bundlechanges.AddMachineChange) error
 
 	deployedApps := func() string {
 		apps := h.applicationsForMachineChange(change.Id())
-		// Note that we always have at least one application that justifies the
-		// creation of this machine.
+		// Note that we *should* always have at least one application
+		// that justifies the creation of this machine. But just in
+		// case, check (see https://pad.lv/1773357).
+		if len(apps) == 0 {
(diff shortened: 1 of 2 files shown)


babbageclunk:
Just forward-porting #9291.
$$merge$$

PR #9294, opened 2018-10-09
Fixes lp#1779677: Add more info to 'settings not found' error & log message. 

anastasiamac:
## Description of change

When things go wrong, the more information we can provide the better.
One of the most common error that we see is 'settings not found'. It is a symptom of a variety of different failures, for example failed model destruction. The message of an error does not convey what settings were not found, for what entity.

This PR adds this information.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1779677


anastasiamac:
Actually, there is a bit of stuttering and ugliness in the messages now...

For example, one of the unit tests's message look like 
``` 
cannot read settings for unit "mysql/0" in relation "wordpress:db mysql:server": settings for key "r#0#provider#mysql/0" in "settings"
```
I'll re-word.

PR #9298, opened 2018-10-10
Merge pull request #9296 from wallyworld/cmr-controller-connect-robust

wallyworld:
## Description of change

Forward port https://github.com/juju/juju/pull/9296

Cross model relations required controller->controller connectivity. If a connection fails, the worker restarts but in the case of a dying relation, the units have already left scope on the consuming side and so the offering side is never informed.

Add logic to the worker which will send a forceCleanup=true along with the published relation dying event; this will be the case if a relation change arrives and the life is dying but there's no child worker running for that relation. The offering side will force any relation units to leave scope and thus the relation on the offering side can properly be removed instead of remaining in the dying state.

## QA steps

bootstrap 2 controllers and create a cmr
stop the agent on the offering side
remove the relation on the consuming side
check logs for connection errors
restart the agent
check that the relation is cleaned up on the offering side

## Bug reference

https://bugs.launchpad.net/juju/+bug/1795499


wallyworld:
$$merge$$

PR #9304, opened 2018-10-10
Fix error on <TAB> autocomplete.

ExternalReality:
## Description of change

`upgrade-series` errors on auto-complete since the first line of its help text contains a keyword that `juju`'s bash auto-compete script interprets in the wrong way. There are several ways to handle this, the approach that is taken here is to tweak the bash auto-completion script. 

## QA steps

Reinstall the script from the juju-repo
```bash
make install-etc
```

Autocomplete an `upgrade-series` command
```bash
upgrade-series <TAB>
juju upgrade-series co<TAB>
```
You should get auto-completion instead of a juju "ERROR ...." style error.

### DIFF
--- a/cmd/juju/machine/upgradeseries.go
+++ b/cmd/juju/machine/upgradeseries.go
@@ -137,7 +137,7 @@ See also:
 func (c *upgradeSeriesCommand) Info() *cmd.Info {
 	return &cmd.Info{
 		Name:    "upgrade-series",
-		Args:    "<action> [args]",
+		Args:    "<command> [args]",
 		Purpose: "Upgrade a machine's series.",
 		Doc:     upgradeSeriesDoc,
 	}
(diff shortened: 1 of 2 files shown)


cmd/juju/machine/upgradeseries.go
```diff
@@ -137,7 +137,7 @@ See also:
 func (c *upgradeSeriesCommand) Info() *cmd.Info {
 	return &cmd.Info{
 		Name:    "upgrade-series",
-		Args:    "<action> [args]",
+		Args:    "<command> [args]",
```
ExternalReality: Technically we don't really need to change this.

ExternalReality:
[juju series-upgrade command auto-complete error](https://trello.com/c/JFQtTlmk/660-juju-series-upgrade-command-auto-complete-error)

ExternalReality:
$$MERGE$$

PR #9312, opened 2018-10-12
Fixes lp#1796618:  Don't upload credential on add-model unless explicitly asked to.

anastasiamac:
## Description of change

Juju will always upload a local credential when 'add-model' command is issued.

This is needed when a user explicitly asks to use a particular local credential via --credential option. In this case, before requesting to add a model, the command will try to upload a credential to the controller. If the credential is new on the controller, it will be uploaded without any further checks. If a credential with this name exists on the controller, this effectively becomes an update of an existing credential which entails check of all the models that use it. This update is not entirely appropriate here. There is a separate 'update-credential' command. Here we only need to cater for 'add' credential functionality (when using local credential is desired) and 'use a named credential that I know exists on controller' (when using controller credential is desired). This aspect will be dealt with in subsequent work when we tackle local vs remote clouds/credentials. At the moment, when specifying a credential explicitly while adding a model and if this credential is already on the controller, this command will try to update credential contents and, hence, run all the check necessary for update.

When a credential is not explicitly specified, Juju should not upload local credential. This PR fixes that.

## QA steps

Expected outcomes.
1: Adding a model without specifying a credential should not try to update credential content stored on controller.
2: Adding a model with a credential that does not exist on the controller will upload that credential.
3. Adding a model with a credential known to the controller will try to update credential content.

* bootstrap 
* add model
* add model by specifying new credential via --credential option
* add model by specifying the same credential used during bootstrap via --credential option [This will fail for k8s models for example]

Sample output for a detailed failure:
```
$ microk8s.config | juju add-k8s k8stest
$ juju add-model test k8stest
Added 'test' model on k8stest with credential 'admin' for user 'admin'
$ juju add-model trial k8stest
Added 'trial' model on k8stest with credential 'admin' for user 'admin'
$ juju add-model trial k8stest --credential=admin
Uploading credential 'k8stest/admin/admin' to controller
Failed to upload credential: some models are no longer visible
Credential invalid for:
  test:
    cloud environ provider provider.kubernetesEnvironProvider not valid
  trial:
    cloud environ provider provider.kubernetesEnvironProvider not valid
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1796618

### DIFF
--- a/cmd/juju/cloud/updatecredential.go
+++ b/cmd/juju/cloud/updatecredential.go
@@ -4,8 +4,6 @@
 package cloud
 
 import (
-	"sort"
-
 	"github.com/juju/cmd"
 	"github.com/juju/errors"
 	"github.com/juju/gnuflag"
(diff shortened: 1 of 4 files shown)


anastasiamac:
[don't upload credential when not asked, lp#1796618](https://trello.com/c/S080PJ7X/475-dont-upload-credential-when-not-asked-lp1796618)

anastasiamac:
$$merge$$

mitechie:
Having an add-model update an existing credential seems like a hidden bomb. Are we sure this is safe and ok to do on the fly? I'd expect an error and forcing a user to be more explicit and update the credential first before using it or just ignoring the local one and preferring the one in the controller in use?

PR #9313, opened 2018-10-12
LXD Profile - Remove feature flag + acceptance tests

SimonRichardson:
## Description of change

The following adds a acceptance test for show machine when using
the lxd profile, along with removing a feature flag from the source code
to allow the acceptance test to pass.

## QA steps

Following the [instructions](https://github.com/juju/juju/blob/develop/acceptancetests/README.md) on installing the dependencies for running the
acceptance tests.

```
./assess_deploy_lxd_profile.py lxd
```


SimonRichardson:
https://trello.com/c/0wH25tfq/670-remove-the-feature-flag-for-lxd-profile

acceptancetests/assess_deploy_lxd_profile.py
```diff
@@ -0,0 +1,82 @@
...
+    client.wait_for(primary)
+
+def assess_juju_lxdprofile_machine(client, args):
+    """ Tests juju status
+
+    Verify with juju status
+    """
+    lxdprofile_machine_verify(client, "juju-{}-lxd-profile-0".format(client.model_name))
+
+def lxdprofile_machine_verify(client, profilename):
```
hmlanigan: Do we want to verify that the profile got written as well?

SimonRichardson: This is verified using the status output. Considering how this works under (gets the information from the lxc API directly) this should be a non-issue.

SimonRichardson:
So the only thing outstanding is the fact that we're using the github reference in the go dependency file. I've got a [PR](https://github.com/juju/charm/pull/265/files) up to fix this in the charm repo. So once that's landed, we're good to go on this one.

SimonRichardson:
$$merge$$

PR #9318, opened 2018-10-15
add-model error when passed nonexistent cloud

Veebers:
## Description of change

Fix the issue when passing the ```add-model``` command a cloud argument that doesn't exist resulting in a nil deference panic.

## QA steps

  - Bootstrap using juju built from develop (or use existing controller)
  - Run: ```juju add-model test this-doesnt-exist```
  - Observe panic

## Documentation changes

N/A

## Bug reference

N/A


Veebers:
$$merge$$

PR #9323, opened 2018-10-16
Fix incorrect machine agent logging path.

ExternalReality:
## Description of change

`Juju` moves the library directory of juju `systemd` init scripts to a the standard location in a `2.4` upgrade step. This upgrade step, however, was writing the services start script with an incorrect logging path. This PR fixes the issue by introducing a `2.4.5` upgrade step which rewrites the files with the correct log path.

## QA steps

This issue only happens on upgrade in `2.4`, due to the error being in an upgrade step, as such we must reproduce the case by upgrading juju from `2.3` to `2.4`:

1. Boostrap with `Juju 2.3`
```bash
snap install juju --channel 2.3 --classic  # Please ensure you are using the right version
juju bootstrap <cloud> test
juju deploy ubuntu # Not strictly necessary since we could check the controller machine
```

2. Upgrade to `Juju 2.4` 
```bash
snap refresh juju --channel 2.4 --classic
juju upgrade-juju -m controller
juju upgrade-juju
```

3. Inspect the `exec-start.sh` to see that the bug has in fact been reproduced
```bash
juju ssh 0 'cat /lib/systemd/system/jujud-machine-0/exec-start.sh'
```
4. Install this PR version of `Juju` (so you get the fix) and upgrade
```bash
snap remove juju
cd <path to juju repo>
go install -v ./...
juju upgrade-juju -m controller --build agent # assuming, of course, you have this PR branch checkout out
juju upgrade-juju
```
6. Wait a few seconds for the files to be rewritten
```bash
juju ssh 0 'cat /lib/systemd/system/jujud-machine-0/exec-start.sh'
```

The script should be corrected and the services should still work.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1793284

### DIFF
--- a/juju/paths/paths.go
+++ b/juju/paths/paths.go
@@ -28,10 +28,18 @@ const (
 	cloudInitCfgDir
 )
 
+const (
+	// NixDataDir is location for agent binaries on *nix operating systems.
+	NixDataDir = "/var/lib/juju"
+
+	// NixDataDir is location for Juju logs on *nix operating systems.
+	NixLogDir = "/var/log"
+)
+
 var nixVals = map[osVarType]string{
 	tmpDir:               "/tmp",
-	logDir:               "/var/log",
-	dataDir:              "/var/lib/juju",
+	logDir:               NixLogDir,
+	dataDir:              NixDataDir,
 	storageDir:           "/var/lib/juju/storage",
 	confDir:              "/etc/juju",
 	jujuRun:              "/usr/bin/juju-run",
--- a/service/agentconf.go
+++ b/service/agentconf.go
@@ -207,7 +207,7 @@ func (s *systemdServiceManager) CreateAgentConf(agentName string, dataDir string
 		kind,
 		name,
 		dataDir,
-		paths.MustSucceed(srvPath, fmt.Errorf("path %s does not exist", srvPath)))
+		srvPath)
 	return AgentConf(info, renderer), nil
 }
 
@@ -315,7 +315,7 @@ func startAgent(name string, kind AgentKind, dataDir string, series string) (err
 		kind,
 		name,
 		dataDir,
-		paths.MustSucceed(srvPath, fmt.Errorf("path %s does not exist", srvPath)),
+		srvPath,
 	)
 	conf := AgentConf(info, renderer)
 	svcName := serviceName(name)
--- a/service/agentconf.go
+++ b/service/agentconf.go
@@ -202,12 +202,12 @@ func (s *systemdServiceManager) CreateAgentConf(agentName string, dataDir string
 		return common.Conf{}, errors.NewNotValid(nil, fmt.Sprintf("agent %q is neither a machine nor a unit", agentName))
 	}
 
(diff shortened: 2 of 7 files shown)


ExternalReality:
[drive by bugfix - log time and date in /var/log/machine-X.log](https://trello.com/c/SrAhJExX/621-drive-by-bugfix-log-time-and-date-in-var-log-machine-xlog)

juju/paths/paths.go
```diff
@@ -28,10 +28,18 @@ const (
 	cloudInitCfgDir
 )
 
+const (
+	// NixDataDir is location for agent binaries on *nix operating systems.
+	NixDataDir = "/var/lib/juju"
+
+	// NixDataDir is location for Juju logs on *nix operating systems.
```
ExternalReality: This is simply a standard `*nix` log dir. You must append "/juju/" to this to get the location for Juju logs on *nix operating systems. Some areas of the code are already doing just that.

service/agentconf.go
```diff
@@ -310,11 +310,12 @@ func startAgent(name string, kind AgentKind, dataDir string, series string) (err
 	if err != nil {
 		return err
 	}
+	srvPath := path.Join(paths.NixLogDir, "juju")
 	info := NewAgentInfo(
 		kind,
 		name,
 		dataDir,
-		paths.MustSucceed(paths.LogDir(series)),
```
ExternalReality: How?

ExternalReality:
!!build!!

ExternalReality:
$$MERGE$$

juju/paths/paths.go
howbazaar: I find it confusing that you have the NixDataDir have the juju suffix but not the NixLogDir (which has a copy / paste error in the comment BTW.

service/agentconf.go
```diff
@@ -202,12 +202,12 @@ func (s *systemdServiceManager) CreateAgentConf(agentName string, dataDir string
 		return common.Conf{}, errors.NewNotValid(nil, fmt.Sprintf("agent %q is neither a machine nor a unit", agentName))
 	}
 
+	srvPath := path.Join(paths.NixLogDir, "juju")
```
howbazaar: This won't work for windows as the NixLogDir isn't right.

ExternalReality:
$$MERGE$$

service/agentconf.go
ExternalReality: Buggers, good catch!

juju/paths/paths.go
ExternalReality: I agree, that is very confusing. I coded this in such a way that it will require minimal change to forward port (for whatever that is worth). The copy Pasta was intentional, when this gets forward ported I will change it will change this and a few other things that are incorrect with respect to the changes in this PR.

service/agentconf.go
manadart: Yes, as I mentioned in https://github.com/juju/juju/pull/9139, this is systemd-specific and should reside under the systemd package. However, that relocation has not yet been undertaken on account of cyclic dependencies that need resolution.

PR #9324, opened 2018-10-16
Prevent duplicate k8s models

wallyworld:
## Description of change

k8s models exist in a namespace named after the unqualified model name.
Extend duplicate checks when creating a new model to account for this k8s behaviour.
At the api facade layer, use the k8s client to check for any existing namespaces.
At the state layer, tweak the dupe checks to use cloud name as qualifier instead of model owner.

## QA steps

deploy a k8s model
attempt to deploy the same model name as a different user


apiserver/facades/client/modelmanager/modelmanager.go
```diff
@@ -89,6 +90,8 @@ type ModelManagerV2 interface {
 	ModelStatus(req params.Entities) (params.ModelStatusResults, error)
 }
 
+type newCaasBrokerFunc func(args environs.OpenParams) (caas.Broker, error)
```
anastasiamac: Oh that's beautiful!! this is what I need to create in credential checks :) Thank you for making me see this!!!

apiserver/facades/client/modelmanager/modelmanager.go
```diff
@@ -499,13 +506,24 @@ func (m *ModelManagerAPI) newCAASModel(cloudSpec environs.CloudSpec,
...
 	}
+
+	// CAAS models exist in a namespace which must be unique.
+	namespaces, err := broker.Namespaces()
+	if err != nil {
+		return nil, errors.Annotate(err, "failed to list namespaces")
+	}
+	nsSet := set.NewStrings(namespaces...)
+	if nsSet.Contains(createArgs.Name) {
+		return nil, errors.NewAlreadyExists(nil, fmt.Sprintf("namespace called %q already exists, would clash with model name", createArgs.Name))
```
anastasiamac: Would not this message be stuttering "..already exists..already exists...."?

wallyworld: No - the New method used is for the case where we want the error type to be AlreadyExists but we need a bespoke message. As opposed to AlreadyExistsf(). All of our error types support this.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #9326, opened 2018-10-16
Remove restriction local charms

kngu9:
## Description of change

This change removes the restriction that collect-metric only works on locally-deployed charms.

## QA steps

Run collect-metrics on any non-locally deployed charm.


kngu9:
!!check!!

kngu9:
!!check!!

kngu9:
!!check!!

PR #9332, opened 2018-10-18
Fix NVMe EBS volumes handling

wallyworld:
## Description of change

Backport https://github.com/juju/juju/pull/9331

More AWS instance types are exposing EBS volumes as NVMe mounts. We need to account for this so that volume registration and Juju storage work correctly together.

## QA steps

Deploy postgres to AWS with Juju 2.4.4. The charm will not come up.
Upgrade to this branch.
Storage is fixed and the charm deploys.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1798001


wallyworld:
$$merge$$

PR #9334, opened 2018-10-18
Ensure all config maps and volumes are deleted when a k8s operator is deleted

wallyworld:
## Description of change

When deleting a k8s application, some operator artifacts (config map and storage volume) were being left behind.

## QA steps

deploy a k8s charm
delete the application
kubectl to check that the operator volume and config maps are gone

### DIFF
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -527,9 +527,16 @@ func (k *kubernetesClient) ensureStorageClass(cfg *storageConfig) (*k8sstorage.S
 func (k *kubernetesClient) DeleteOperator(appName string) (err error) {
 	logger.Debugf("deleting %s operator", appName)
 
-	// First delete the config map.
+	// First delete the config map(s).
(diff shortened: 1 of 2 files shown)


wallyworld:
[operator deletion leaves behind storage pv](https://trello.com/c/tkMvKwvY/485-operator-deletion-leaves-behind-storage-pv)

caas/kubernetes/provider/k8s.go
```diff
@@ -1718,6 +1725,10 @@ func operatorConfigMapName(appName string) string {
 	return operatorName(appName) + "-config"
 }
 
+func operatorConfigurationsConfigMapName(appName string) string {
+	return deploymentName(appName) + "-configurations-config"
```
ycliuhw: wondering what the difference between `operatorConfigurationsConfigMapName` and `operatorConfigMapName`

wallyworld: i think this gets created as part of the stateful set setup maybe? no sure, but it is there and is left behind so we ensure it is deleted

wallyworld:
$$merge$$

PR #9335, opened 2018-10-18
http-server: register handlers before starting the server

babbageclunk:
## Description of change

We occasionally see 404s for valid API requests in tests and controllers at startup time. These happen because the workers that register handlers in the http-server's mux (api-server and raft-transport) depend on it, so they start afterwards which means there's a window when http-server is running with no handlers and responds with 404 to all requests. To prevent this, we extract the http-server's outputs (the mux and authenticator) into their own http-server-args worker, make api-server and raft-transport depend on the new worker, and then make http-server depend on those two so that it won't start until all handlers are registered.

Extra details: 
* If the disable-raft feature flag is set, the raft workers (including raft-transport) won't run, so http-server checks the raft-enabled-flag before requiring raft-transport.
* raft-transport was gated by `ifFullyUpgraded`, but this introduced an implicit dependency loop: raft-transport -> upgrader -> api-caller (implicitly) -> http-server -> raft-transport. The transport doesn't actually need upgrade steps to be run, so I moved that gate to the raft worker proper instead (which definitely does need them - there's one in 2.4->2.5).
* Pulled the handler registration in raft-transport out of the loop so that we know it's happened by the time `NewWorker` returns.

## QA steps

* Bootstrap a controller, deploy some applications and relate them, check status looks right and logs don't report errors. Use juju_engine_report to check that dependencies are as expected.
* Upgrade to 2.5 (with the dependency changes).
* Bootstrap with `features=[disable-raft]` and make the same checks.

## Documentation changes

None.

### DIFF
--- a/worker/caasoperator/caasoperator_test.go
+++ b/worker/caasoperator/caasoperator_test.go
@@ -281,14 +281,14 @@ func (s *WorkerSuite) TestWorkerDownloadsCharm(c *gc.C) {
 }
 
 func (s *WorkerSuite) assertUniterStarted(c *gc.C) (worker.Worker, watcher.NotifyChannel) {
-	var (
-		uniterStarted      int32
-		applicationChannel atomic.Value
-	)
+	applicationChannel := make(chan watcher.NotifyChannel)
 	s.config.StartUniterFunc = func(runner *worker.Runner, params *uniter.UniterParams) error {
 		c.Assert(params.UnitTag.Id(), gc.Equals, "gitlab/0")
-		atomic.AddInt32(&uniterStarted, 1)
-		applicationChannel.Store(params.ApplicationChannel)
+		select {
+		case applicationChannel <- params.ApplicationChannel:
+		case <-time.After(coretesting.LongWait):
+			c.Fatalf("timeout sending application channel")
+		}
 		return nil
 	}
 
@@ -306,12 +306,12 @@ func (s *WorkerSuite) assertUniterStarted(c *gc.C) (worker.Worker, watcher.Notif
 		c.Fatal("timed out sending unit change")
 	}
 
-	for attempt := coretesting.LongAttempt.Start(); attempt.Next(); {
-		if atomic.LoadInt32(&uniterStarted) > 0 {
-			return w, applicationChannel.Load().(watcher.NotifyChannel)
-		}
+	select {
+	case channel := <-applicationChannel:
+		return w, channel
+	case <-time.After(coretesting.LongWait):
+		c.Fatalf("timeout while waiting for uniter to start")
 	}
-	c.Fatalf("timeout while waiting for uniter to start")
 	panic("not reachable")
 }
 


babbageclunk:
There are caasoperator failures here that seem intermittent, I can't reproduce them even under stress-race. :(

!!build!!

babbageclunk:
Although I did find a race that causes a panic in the tests, so I've fixed that.

worker/httpserver/manifold.go
```diff
@@ -99,11 +111,27 @@ func (config ManifoldConfig) start(context dependency.Context) (_ worker.Worker,
...
-	var clock clock.Clock
-	if err := context.Get(config.ClockName, &clock); err != nil {
+	var mux *apiserverhttp.Mux
+	if err := context.Get(config.MuxName, &mux); err != nil {
 		return nil, errors.Trace(err)
 	}
 
+	// We don't actually need anything from these workers, but we
+	// shouldn't start until they're available.
+	if err := context.Get(config.APIServerName, nil); err != nil {
```
howbazaar: I'm kinda surprised that this works, but I'm pleased it does.

worker/httpserver/manifold.go
```diff
@@ -99,11 +111,27 @@ func (config ManifoldConfig) start(context dependency.Context) (_ worker.Worker,
...
+	if err := context.Get(config.RaftEnabledName, &raftEnabled); err != nil {
+		return nil, errors.Trace(err)
+	}
+	if raftEnabled.Check() {
+		if err := context.Get(config.RaftTransportName, nil); err != nil {
+			return nil, errors.Trace(err)
+		}
+	}
+
 	var stTracker workerstate.StateTracker
```
howbazaar: I don't think we need the state tracker any more. I think it was juts there to create the authenticator.

worker/httpserver/manifold.go
```diff
@@ -159,55 +180,7 @@ func (config ManifoldConfig) start(context dependency.Context) (_ worker.Worker,
...
-	authenticator httpcontext.LocalMacaroonAuthenticator
-	cleanupOnce   sync.Once
-	cleanup       func()
-}
-
-func (w *wrapperWorker) Wait() error {
-	err := w.Worker.Wait()
-	w.cleanupOnce.Do(w.cleanup)
-	return err
+	return common.NewCleanupWorker(w, func() { stTracker.Done() }), nil
```
howbazaar: If we don't need the tracker, this gets simpler.

worker/httpserverargs/manifold.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/juju/juju/apiserver/apiserverhttp"
+	"github.com/juju/juju/apiserver/httpcontext"
+	workerstate "github.com/juju/juju/worker/state"
+)
+
+// ManifoldConfig holds the resources needed to run an httpserverargs
+// worker.
+type ManifoldConfig struct {
+	ClockName string
```
howbazaar: I don't expect you to change this, but I think a manifold that just holds the clock is a waste given that we always need it and the original manifold config has the clock. I prefer to just pass the clock itself in rather than get it from the clock worker. I'd loke to change all workers to do this.

worker/httpserverargs/manifold.go
```diff
@@ -0,0 +1,145 @@
...
+			return nil
+		case authenticatorType.AssignableTo(elemType):
+			rv.Elem().Set(reflect.ValueOf(w.authenticator))
+			return nil
+		}
+	}
+	return errors.Errorf("unexpected output type %T", out)
+}
+
+func newWorker(
```
howbazaar: Very nice and simple. I like it.

worker/httpserver/manifold.go
babbageclunk: I was hoping to get rid of it too, but unfortunately we still need it for `newTLSConfig`.

worker/httpserver/manifold.go
babbageclunk: :(

worker/caasoperator/caasoperator_test.go
```diff
@@ -281,14 +281,14 @@ func (s *WorkerSuite) TestWorkerDownloadsCharm(c *gc.C) {
 }
 
 func (s *WorkerSuite) assertUniterStarted(c *gc.C) (worker.Worker, watcher.NotifyChannel) {
-	var (
-		uniterStarted      int32
-		applicationChannel atomic.Value
-	)
+	applicationChannel := make(chan watcher.NotifyChannel, 1)
```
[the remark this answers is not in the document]
babbageclunk: I wasn't sure that the start func would only be called once, but I am now - I'll change it to unbuffered.

worker/httpserver/manifold.go
babbageclunk: Yeah, same - it's documented in the interface, so the behaviour's deliberate. Thanks Will!

worker/httpserverargs/manifold.go
babbageclunk: Thanks!

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

babbageclunk:
This is a fix for (the 404s in) https://bugs.launchpad.net/juju/+bug/1792299

PR #9340, opened 2018-10-18
2.3 assertion txn patch 1798485

jameinel:
## Description of change

This is related to: https://github.com/juju/mgo/pull/3.

We recently added support to limit the length of `txn-queue` to limit fallout when we have a broken transaction. We also added pruning of `txn-queue` for committed transactions. However, it seems in [lp:1798485](https://bugs.launchpad.net/juju/+bug/1798485) that we can have more that 1000 transactions get committed within an hour that only 'assert' against a given document. (In this case it was a machine document.)

Rather than prune more frequently, it seems better to fix this at the source. Rather than *never* cleaning up the `txn-queue` for documents that are only an assert, we now check if there is a significant number of stale transactions, we go ahead and remove them from the `txn-queue`. 

## QA steps

You can see the test cases that were added in the patch.
I haven't worked out a way to manually expose this issue yet.

## Documentation changes

None.

## Bug reference

[lp:1798485](https://bugs.launchpad.net/juju/+bug/1798485)


jameinel:
!!build!!

jameinel:
$$merge$$

PR #9346, opened 2018-10-19
http-server: register handlers before starting the server

babbageclunk:
## Description of change

This is a forward-port of #9335 to the develop branch.

We occasionally see 404s for valid API requests in tests and controllers at startup time. These happen because the workers that register handlers in the http-server's mux (api-server and raft-transport) depend on it, so they start afterwards which means there's a window when http-server is running with no handlers and responds with 404 to all requests. To prevent this, we extract the http-server's outputs (the mux and authenticator) into their own http-server-args worker, make api-server and raft-transport depend on the new worker, and then make http-server depend on those two so that it won't start until all handlers are registered.

Extra details: 
* If the disable-raft feature flag is set, the raft workers (including raft-transport) won't run, so http-server checks the raft-enabled-flag before requiring raft-transport.
* raft-transport was gated by `ifFullyUpgraded`, but this introduced an implicit dependency loop: raft-transport -> upgrader -> api-caller (implicitly) -> http-server -> raft-transport. The transport doesn't actually need upgrade steps to be run, so I moved that gate to the raft worker proper instead (which definitely does need them - there's one in 2.4->2.5).
* Pulled the handler registration in raft-transport out of the loop so that we know it's happened by the time `NewWorker` returns.

## QA steps

* Bootstrap a controller, deploy some applications and relate them, check status looks right and logs don't report errors. Use juju_engine_report to check that dependencies are as expected.
* Bootstrap a 2.4 controller (with #9335 included), upgrade to 2.5, check that everything works as above.

## Documentation changes

None.

# Bug reference
https://bugs.launchpad.net/juju/+bug/1792299


babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #9341, opened 2018-10-18
Fix repeate upgrade series message txns

ExternalReality:
## Description of change

`upgrade-series` commands user messages where being marked as having been seen multiple times. This would cause unnecessary database activity during series upgrades. 

## QA steps

<TBA>

## Bug reference

<TBA>

PR #9348, opened 2018-10-19
Transaction fix follow on

ExternalReality:
## Description of change

These commits were supposed to go onto #9341 but were mistakenly not added. There commits must be added or the feature related to #9341 will be in a broken state.

### DIFF
--- a/state/machine_upgradeseries.go
+++ b/state/machine_upgradeseries.go
@@ -507,12 +507,22 @@ func (m *Machine) GetUpgradeSeriesMessages() ([]string, bool, error) {
 	if err != nil {
 		return nil, false, errors.Trace(err)
 	}
+	// finished means that a subsequent call to this method, while the
+	// Machine Lock is of a similar Machine Status, would return no
+	// additional messages (notifications). Since the value of this variable
+	// is returned, callers may choose to close streams or stop watchers
+	// based on this information.
+	finished := lock.MachineStatus == model.UpgradeSeriesCompleted ||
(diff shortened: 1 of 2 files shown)


ExternalReality:
$$MERGE$$

ExternalReality:
$$MERGE$$

state/machine_upgradeseries.go
```diff
@@ -507,32 +507,34 @@ func (m *Machine) GetUpgradeSeriesMessages() ([]string, bool, error) {
...
+		return []string{}, finished, nil
+	}
 	sort.Slice(unseenMessages, func(i, j int) bool {
 		return unseenMessages[i].Timestamp.Before(unseenMessages[j].Timestamp)
 	})
 	messages := make([]string, 0)
 	for _, unseenMessage := range unseenMessages {
 		messages = append(messages, unseenMessage.Message)
 	}
-	err = m.SetUpgradeSeriesMessagesAsSeen(unseenMessages)
```
jameinel: Thanks for the original fix, but doesn't this cause us to set things that are already set?
Or is the issue that we were only looking at the in-memory representation and didn't notice that we hadn't recorded that the message was seen?

ExternalReality: This does set the status of all the messages every time. The reason for this that we need the index of lock message in the array to construct the database transaction. 

Once a user asks to fetch the current messages that are in the db we pull those messages into memory and send them across the wire. We consider those messages to be seen that have been pulled into memory. Therefore we can consider all of the in-memory messages to have been 'seen'.

PR #9360, opened 2018-10-24
Rework the api-caller worker retry strategies.

jameinel:
## Description of change

This changes how the workers try to connect to the controller, and how
they retry their connections.

Rather than trying all controllers every 200ms for 3s on, 3s off, this
relies on the exponential backoff of the dependency engine. And instead
we try 1 controller at a time, with a 200ms delay between controllers.

We also shuffle the order that we try controllers with each retry. This
should help balance the load on controllers, and avoid having them
hammered when starting up.

We max out the delay between retries to 2 minutes, and backoff by 20%
each time. This means it takes around 20 retries in 10 minutes before we
cap out at 2min per retry. (which is way below the original 1500 retries
in 10 minutes.)

## QA steps

```
$ juju bootstrap lxd
$ juju enable-ha
$ juju deploy simple-charm
```
Kill/stop various controllers and monitor the logs to see how the agents are trying to reconnect.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1799365
https://bugs.launchpad.net/juju/+bug/1793245

### DIFF
--- a/worker/apicaller/connect.go
+++ b/worker/apicaller/connect.go
@@ -4,7 +4,6 @@
 package apicaller
 
 import (
-	"math/rand"
 	"time"
 
 	"github.com/juju/errors"
@@ -83,9 +82,6 @@ func connectFallback(
 ) (
 	conn api.Connection, didFallback bool, err error,
 ) {
-
-	// Encourage load balancing by defaulting to a different controller each time.
-	rand.Shuffle(len(info.Addrs), func(i, j int) { info.Addrs[i], info.Addrs[j] = info.Addrs[j], info.Addrs[i] })
 	// We expect to assign to `conn`, `err`, *and* `info` in
 	// the course of this operation: wrapping this repeated
 	// atom in a func currently seems to be less treacherous
--- a/worker/apicaller/connect.go
+++ b/worker/apicaller/connect.go
@@ -4,6 +4,7 @@
 package apicaller
 
 import (
+	"math/rand"
 	"time"
 
 	"github.com/juju/errors"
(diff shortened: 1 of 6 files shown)


worker/apicaller/connect.go
```diff
@@ -83,6 +84,8 @@ func connectFallback(
 	conn api.Connection, didFallback bool, err error,
 ) {
 
+	// Encourage load balancing by defaulting to a different controller each time.
+	rand.Shuffle(len(info.Addrs), func(i, j int) { info.Addrs[i], info.Addrs[j] = info.Addrs[j], info.Addrs[i] })
```
4a6f656c: Unless I'm missing something, it seems like it would be better to do this in the code that actually handles the parallel connections (`dialWebsocketMulti` in `api/apiclient.go`), so that all clients that connect to the API server randomise the controller order (for example, with this diff `juju status` will continue to hit the first controller in the list).

jameinel: I wasn't sure about doing all clients, but some discussions convinced me that it should be reasonable. I do think there were reasons why we were intentionally not doing so, but I think that was more when we expected average latency to each controller being different, which is not actually likely to be true.

jameinel:
$$merge$$ I believe I've addressed all the concerns of Tim and Joe.

jameinel:
$$merge$$ test timed out on the bot, but I cannot reproduce locally

mitechie:
Looks like gofmt is sad and fails to go through the check job atm

jameinel:
$$merge$$ I think I found out why the 2 tests were failing intermittently, because we no longer dial in the order they expected.
But both pass reliably for me now.

PR #9366, opened 2018-10-24
diff-bundle: Compare bundle with model and report differences

babbageclunk:
## Description of change

This adds a new command, `diff-bundle` that compares a local or charmstore bundle and reports differences between them. It supports includes and overlays in the same way as the deploy command, and uses the same code to extract model information to compare it to the bundle.

More information about the command including options and output is in the discourse post: https://discourse.jujucharms.com/t/new-diff-bundle-command/247

## QA steps

* Deploy a bundle to a model, and then compare the bundle to the model. Edit the bundle to change details like number of units or exposure and add machines, applications and relations, and make different changes to the model. See that they're reflected in the diff output.
* Remove units and machines deployed to the model from the bundle, and deploy it again (so those units get recreated with different IDs from the ones in the bundle). Run a diff and see that the different machines are reported. Use the --map-machines options to indicate which bundle machines correspond to the new machines and the differences should be removed.

## Documentation changes

Yes, this will need documentation. I'll do some more narrative/task-oriented documentation in discourse to help with this.

### DIFF
--- a/cmd/juju/application/bundle.go
+++ b/cmd/juju/application/bundle.go
@@ -57,6 +57,54 @@ type deploymentLogger interface {
 	Infof(string, ...interface{})
 }
 
+// composeBundle adds the overlays and bundle includes into the passed bundle
+// data struct.
+func composeBundle(data *charm.BundleData, ctx *cmd.Context, bundleDir string, overlayFileNames []string) error {
+	if err := processBundleOverlay(data, overlayFileNames...); err != nil {
+		return errors.Trace(err)
+	}
+	if bundleDir == "" {
+		bundleDir = ctx.Dir
+	}
+	if err := processBundleIncludes(bundleDir, data); err != nil {
+		return errors.Annotate(err, "unable to process includes")
+	}
+	return nil
+}
+
+func verifyBundle(data *charm.BundleData, bundleDir string) error {
+	verifyConstraints := func(s string) error {
+		_, err := constraints.Parse(s)
+		return err
+	}
+	verifyStorage := func(s string) error {
+		_, err := storage.ParseConstraints(s)
+		return err
+	}
+	verifyDevices := func(s string) error {
+		_, err := devices.ParseConstraints(s)
+		return err
+	}
+	var verifyError error
+	if bundleDir == "" {
+		verifyError = data.Verify(verifyConstraints, verifyStorage, verifyDevices)
+	} else {
+		verifyError = data.VerifyLocal(bundleDir, verifyConstraints, verifyStorage, verifyDevices)
+	}
+
+	if verr, ok := errors.Cause(verifyError).(*charm.VerificationError); ok {
+		errs := make([]string, len(verr.Errors))
+		for i, err := range verr.Errors {
+			errs[i] = err.Error()
+		}
+		return errors.New("the provided bundle has the following errors:\n" + strings.Join(errs, "\n"))
+	}
+	if verifyError != nil {
+		return errors.Trace(verifyError)
+	}
+	return nil
+}
+
 // deployBundle deploys the given bundle data using the given API client and
 // charm store client. The deployment is not transactional, and its progress is
 // notified using the given deployment logger.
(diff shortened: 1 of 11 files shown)


cmd/juju/application/bundle.go
```diff
@@ -57,27 +57,22 @@ type deploymentLogger interface {
...
-	bundleMachines map[string]string,
-) (map[*charm.URL]*macaroon.Macaroon, error) {
-
-	if err := processBundleOverlay(data, bundleOverlayFile...); err != nil {
-		return nil, err
+// composeBundle adds the overlays and bundle includes into the passed bundle
+// data struct.
+func composeBundle(data *charm.BundleData, ctx *cmd.Context, bundleDir string, overlayFileNames []string) error {
+	if err := processBundleOverlay(data, overlayFileNames...); err != nil {
+		return errors.Trace(err)
```
wallyworld: maybe an annotate here too, like below?

cmd/juju/application/bundle.go
```diff
@@ -92,27 +87,47 @@ func deployBundle(
...
 	}
-	if verifyError != nil {
-		if verr, ok := verifyError.(*charm.VerificationError); ok {
-			errs := make([]string, len(verr.Errors))
-			for i, err := range verr.Errors {
-				errs[i] = err.Error()
-			}
-			return nil, errors.New("the provided bundle has the following errors:\n" + strings.Join(errs, "\n"))
+
+	if verr, ok := errors.Cause(verifyError).(*charm.VerificationError); ok {
```
wallyworld: nice pickup

cmd/juju/application/bundle.go
```diff
@@ -92,27 +87,47 @@ func deployBundle(
...
+		errs := make([]string, len(verr.Errors))
+		for i, err := range verr.Errors {
+			errs[i] = err.Error()
 		}
-		return nil, errors.Trace(verifyError)
+		return errors.New("the provided bundle has the following errors:\n" + strings.Join(errs, "\n"))
+	}
+	if verifyError != nil {
```
wallyworld: don't need this
errors.Trace(err) returns nil if err is nil

cmd/juju/application/bundlediff.go
```diff
@@ -0,0 +1,275 @@
...
+const bundleDiffDoc = `
+Bundle can be a local bundle file or the name of a bundle in
+the charm store. The bundle can also be combined with overlays (in the
+same way as the deploy command) before comparing with the model.
+
+The map-machines option works similarly as for the deploy command, but
+existing is always assumed, so it doesn't need to be specified.
+
+Examples:
+    juju diff-bundle localbundle.yaml
```
wallyworld: can you include a -m example?

cmd/juju/application/bundlediff.go
```diff
@@ -0,0 +1,275 @@
...
+
+// AllowInterspersedFlags is part of cmd.Command.
+func (c *bundleDiffCommand) AllowInterspersedFlags() bool { return true }
+
+// Info is part of cmd.Command.
+func (c *bundleDiffCommand) Info() *cmd.Info {
+	return &cmd.Info{
+		Name:    "diff-bundle",
+		Args:    "<bundle file or name>",
+		Purpose: "Compare a bundle against a model and report any differences.",
```
wallyworld: remove the trailing "."

cmd/juju/application/deploy.go
```diff
@@ -1164,66 +1164,72 @@ func (c *DeployCommand) maybePredeployedLocalCharm() (deployFn, error) {
...
-func (c *DeployCommand) maybeReadLocalBundle() (deployFn, error) {
-	bundleFile := c.CharmOrBundle
-	var bundleDir string
-	isDir := false
-	resolveDir := false
-
+// readLocalBundle returns the bundle data and bundle dir (for
+// resolving includes) for the bundleFile passed in. If the bundle
+// file doesn't exist we return nil.
+func readLocalBundle(ctx *cmd.Context, bundleFile string) (*charm.BundleData, string, error) {
```
wallyworld: We should return NotFoundError and have the callers deal with that if the bundle doesn't exist.
That's a lot more obvious and in line with expected semantics

cmd/juju/application/deploy.go
```diff
@@ -1164,66 +1164,72 @@ func (c *DeployCommand) maybePredeployedLocalCharm() (deployFn, error) {
...
+	bundleData, bundleDir, err := readLocalBundle(ctx, bundleFile)
+	if charmrepo.IsInvalidPathError(err) {
+		return nil, errors.Errorf(""+
+			"The charm or bundle %q is ambiguous.\n"+
+			"To deploy a local charm or bundle, run `juju deploy ./%[1]s`.\n"+
+			"To deploy a charm or bundle from the store, run `juju deploy cs:%[1]s`.",
+			bundleFile,
+		)
+	}
+	if err != nil {
```
wallyworld: if err != nil && !errors.IsNotFound(err)

etc

cmd/juju/application/bundlediff_test.go
```diff
@@ -0,0 +1,460 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"reflect"
+	"strings"
+
+	"github.com/juju/loggo"
+
```
wallyworld: delete line

cmd/juju/application/bundlediff_test.go
```diff
@@ -0,0 +1,460 @@
...
+	"github.com/juju/juju/apiserver/params"
+	"github.com/juju/juju/cmd/juju/application"
+	"github.com/juju/juju/constraints"
+	"github.com/juju/juju/core/model"
+	"github.com/juju/juju/jujuclient"
+	"github.com/juju/juju/jujuclient/jujuclienttesting"
+	"github.com/juju/juju/testing"
+)
+
+var logger = loggo.GetLogger("xtian")
```
wallyworld: I prefer GetLogger("wallyworld") :-P

cmd/juju/application/bundle.go
babbageclunk: yeah, good call

cmd/juju/application/bundle.go
babbageclunk: Fixed!

cmd/juju/application/bundlediff.go
babbageclunk: Done.

cmd/juju/application/bundlediff.go
babbageclunk: Removed.

cmd/juju/application/bundlediff_test.go
babbageclunk: oops!

cmd/juju/application/bundlediff_test.go
babbageclunk: ha

cmd/juju/application/deploy.go
babbageclunk: Yeah, I agree - the top-level functions in the deploy fallback pipeline can still return nil to indicate "try the next one", but that shouldn't happen further down the stack. Fixing.

babbageclunk: I used NotValid rather than NotFound - partly because that seems more correct and partly because it's less likely to be a real error that comes back.

cmd/juju/application/deploy.go
babbageclunk: Done

babbageclunk:
$$merge$$

cmd/juju/application/bundlediff.go
babbageclunk: Gah, this is wrong - we have a test that enforces all Purposes must have a terminal "."!

babbageclunk:
$$merge$$

PR #9367, opened 2018-10-24
Rework --help for `set-credential` command

pmatulis:
## Description of change

 - Reworked text based on well-reviewed documentation PR https://github.com/juju/docs/pull/3119
 - Changed summary line
 - Added example 
 - Added 'See also' commands
 - Removed text on line 'modelCredentialDoc' (?)

@anastasiamac I really think the concept of "linking" a credential to a model is the best here.

## QA steps

View the help (`juju help set-credential`) and see the revised output.

## Documentation changes

N/A, but there is some precision here that would be good to eventually get into the docs.

## Bug reference

N/A


pmatulis:
[Improve help for `set-credential` command](https://trello.com/c/24fwf2N3/728-improve-help-for-set-credential-command)

cmd/juju/model/setcredential.go
```diff
@@ -61,7 +61,7 @@ func (c *modelCredentialCommand) Info() *cmd.Info {
 	return &cmd.Info{
 		Name:    "set-credential",
 		Args:    "<cloud name> <credential name>",
-		Purpose: "Change cloud credential for the current or specified model.",
+		Purpose: "Links a remote credential to a model.",
```
anastasiamac: I do not like this wording:
1. It does not 'link". At the very best, it "re-links". I did not just pick a word out of the sky. Either "change" or "replace" or even "change reference"...?
2. It is not just "remote" credential. If needed, we do upload local credential.

cmd/juju/model/setcredential.go
```diff
@@ -171,12 +171,28 @@ func (c *modelCredentialCommand) findCredentialLocally(ctx *cmd.Context) (*cloud
...
+The credential specified may exist locally (on the client), remotely (on the
+controller), or both. The command will error out if the credential is neither
+remotely nor locally stored.
 
-If the specified credential does not exist remotely on the controller, but does exist locally on the client,
-Juju will upload it to the controller first and then replace the model's credential.
+When remote, the credential will be linked to the specified model.
 
+When local and not remote, the credential will first be uploaded to the
+controller and then linked.
```
anastasiamac: We do have IT-savy users. I am not sure that "link" is the technical jargon I'd want to use in this context - reference is better for our audience?..

cmd/juju/model/setcredential.go
pmatulis: 1.a Re-link: not always. You can get a remote credential that is not associated with any model (by first linking another credential to its model). You would then link that orphaned credential to a different model. Anyway, re-linking is still linking. :)
1.b The problem I see with the words "change a credential..." or "replace a credential..." is that it sounds like the content of the credential is changing or the content is being replaced, which isn't the case here. We're really just "pointing" a credential to a model.
2. It still always links a remote credential with a model. If we don't see the upload as an extra it gets unnecessarily complicated to summarise.

pmatulis: Ahhh.

pmatulis:
$$merge$$

PR #9371, opened 2018-10-25
Delay relation-joined until unit has address

wallyworld:
## Description of change

Wait until a unit has an address before running relation joined. This is only really applicable to k8s models.

NB We introduce a proper unit test suite in api/uniter - the older full stack test file was renamed to legacy_unit_test.go so the diff is larger because of this. But the bulk of the changes are just due to the rename. When reviewing, legacy_unit_test.go can be skipped as it's just the renamed unit_test.go. And the unit_test.go content appears as a +/- diff but is really just the addition of a new test suite.

## QA steps

deploy k8s model
deploy gitlab
wait until done
deploy and relate mariadb immediately
check logs for empty address warning


mitechie:
How does this compare to kelvins branch?

wallyworld:
This PR is superseded by https://github.com/juju/juju/pull/9369

PR #9374, opened 2018-10-26
Model credential changes adjustment for validity watcher.

anastasiamac:
## Description of change

Since model's cloud credential can now change, this PR adjusts credential validity worker that gates cloud communications between a cloud provider and a model.
Previously the worker only watched the changes to cloud credential content for this model.
With this PR, the worker also gains a watcher that notifies it of changes to what cloud credential a model uses. 

## QA steps

1. bootstrap
2. change model credential for a default model
3. observe credential validity worker restart

### DIFF
--- a/apiserver/facades/agent/credentialvalidator/credentialvalidator.go
+++ b/apiserver/facades/agent/credentialvalidator/credentialvalidator.go
@@ -16,8 +16,18 @@ import (
 
 var logger = loggo.GetLogger("juju.api.credentialvalidator")
 
-// CredentialValidator defines the methods on credentialvalidator API endpoint.
-type CredentialValidator interface {
+// CredentialValidatorV2 defines the methods on version 2 facade for the
+// credentialvalidator API endpoint.
+type CredentialValidatorV2 interface {
+	InvalidateModelCredential(reason string) (params.ErrorResult, error)
+	ModelCredential() (params.ModelCredential, error)
+	WatchCredential(params.Entity) (params.NotifyWatchResult, error)
+	WatchModelCredential() (params.NotifyWatchResult, error)
+}
+
+// CredentialValidatorV1 defines the methods on version 1 facade
+// for the credentialvalidator API endpoint.
+type CredentialValidatorV1 interface {
 	InvalidateModelCredential(reason string) (params.ErrorResult, error)
 	ModelCredential() (params.ModelCredential, error)
 	WatchCredential(params.Entity) (params.NotifyWatchResult, error)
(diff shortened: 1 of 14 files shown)


anastasiamac:
[adjust existing credential validator worker to react to model cred changes](https://trello.com/c/OVGOZeGg/469-adjust-existing-credential-validator-worker-to-react-to-model-cred-changes)

apiserver/facades/agent/credentialvalidator/credentialvalidator.go
```diff
@@ -97,3 +123,30 @@ func (api *CredentialValidatorAPI) ModelCredential() (params.ModelCredential, er
...
 }
+
+// Mask out new methods from the old API versions. The API reflection
+// code in rpc/rpcreflect/type.go:newMethod skips 2-argument methods,
+// so this removes the method as far as the RPC machinery is concerned.
+//
+// WatchModelCredential did not exist prior to v2.
+func (*CredentialValidatorAPIV1) WatchModelCredential(_, _ struct{}) {}
+
+// WatchModelCredential returns a NotifyWatcher that watches what cloud credential a model uses.
```
babbageclunk: Maybe this would be clearer as "returns a NotifyWatcher that signals when the model's cloud credential changes."?

worker/credentialvalidator/worker.go
```diff
@@ -64,42 +70,42 @@ func NewWorker(config Config) (worker.Worker, error) {
...
+		modelCredentialWatcher: mcw,
 	}
 
+	// The watcher needs to be added to the worker's catacomb plan
+	// here in order to be controlled by this worker's lifecycle events:
+	// for example, to be destroyed when this worker is destroyed, etc.
+	// We also add the watcher to the Plan.Init collection to ensure that
+	// the worker's Plan.Work method is executed after the watcher
+	// is initialised and watcher's changes collection obtains the changes.
+	// Watchers that are added using catacomb.Add method
```
babbageclunk: I don't think this comment's right - the watcher's changes collection is initialised when it's created. Using `catacomb.Add` should be the same as putting the worker in `plan.Init` (they both use `catacomb.add` under the hood) - the only difference is that if the catacomb's been killed for some reason before the watcher is `.Add`ed the watcher will be stopped immediately, which isn't needed if the watcher is in `plan.Init` (since the catacomb couldn't have been killed).

It's fine to put the watcher in `plan.Init`, but if there's some subtle bug with adding watchers via `catacomb.Add` then we have a big problem because lots of workers do that.

babbageclunk:
It looks like there's a watcher that combines events from multiple notify watchers: `apiserver/common/MultiNotifyWatcher`.

worker/credentialvalidator/worker.go
anastasiamac: So, the comment was made with relation to the actual implementation of catacomb.Invoke().
I have done a lot of investigations and testing when I first wrote this worker (and this comment... note it was just moved not really added). See
https://github.com/juju/worker/blob/v1/catacomb/catacomb.go#L91

I needed the watchers in catacomb by the time I invoked to react correctly to initial change. I think maybe majority of the situations you describe do not care for initial state of events and only care about subsequent changes?...I don't know the answer for the situations you are thinking of, but I do know that here I needed watchers in Plan.Init.

apiserver/facades/agent/credentialvalidator/credentialvalidator.go
anastasiamac: Thank you for the tip :) I looked at common.MultiNotifyWatcher as per your suggestion... I am not too sure that it will actually be of much benefit to me here. We can be in situation where we do not have a credential to watch but we do want to watch a model-credential reference changing from nil/empty to something concrete. It feels like an easier job done with 2 watchers.

I'll keep it as-is for now. We can circle back if needed.

anastasiamac:
$$merge$$

PR #9380, opened 2018-10-28
Fix for windows filepath handling and error message string.

Veebers:
## Description of change

Fixes file path handling with windows as well as a drive-by error message fix.


Veebers:
!!build!!

cmd/juju/resource/deploy.go
```diff
@@ -284,23 +284,21 @@ func (d deployUploader) checkExpectedResources(filenames map[string]string, revi
...
+	f, err := os.Open(path)
+	if err == nil {
 		defer f.Close()
-		details, err = unMarshalDockerDetails(f)
+		details, err := unMarshalDockerDetails(f)
 		if err != nil {
 			return details, errors.Trace(err)
 		}
+		return details, nil
 	} else if err := resources.ValidateDockerRegistryPath(path); err == nil {
```
anastasiamac: I wonder if we want to at least log the original err?

Veebers:
$$merge$$

PR #9381, opened 2018-10-29
Refactor common error parsing for auth failures.

anastasiamac:
## Description of change

Some Juju providers already react to an invalid cloud credential.
This code is refactored for ease of use by other providers. For example, the next-in-line provider to benefit from this code is openstack. It will only need to define a function that determines if a given error is related to auth and call a common method.

The most interesting logic has been moved to provider/common/errors.go and maas provider logic changed to call these refactored methods when processing errors from the cloud.

### DIFF
--- a/provider/common/errors.go
+++ b/provider/common/errors.go
@@ -86,7 +86,7 @@ var AuthorisationFailureStatusCodes = set.NewInts(
 // MaybeHandleCredentialError determines if a given error relates to an invalid credential.
 //  If it is, the credential is invalidated and the return bool is true.
 // Original error is returned untouched.
-func MaybeHandleCredentialError(isAuthError func(error) bool, err error, ctx context.ProviderCallContext) (error, bool) {
+func MaybeHandleCredentialError(isAuthError func(error) bool, err error, ctx context.ProviderCallContext) bool {
 	denied := isAuthError(errors.Cause(err))
 	if ctx != nil && denied {
 		invalidateErr := ctx.InvalidateCredential("cloud denied access")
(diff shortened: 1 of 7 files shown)


anastasiamac:
pre-check is fine except for win tests which are fixed in https://github.com/juju/juju/pull/9380

provider/common/errors.go
```diff
@@ -80,3 +82,24 @@ var AuthorisationFailureStatusCodes = set.NewInts(
 	http.StatusForbidden,
 	http.StatusProxyAuthRequired,
 )
+
+// MaybeHandleCredentialError determines if a given error relates to an invalid credential.
+//  If it is, the credential is invalidated and the return bool is true.
+// Original error is returned untouched.
+func MaybeHandleCredentialError(isAuthError func(error) bool, err error, ctx context.ProviderCallContext) (error, bool) {
```
babbageclunk: If the original error is always returned untouched, it shouldn't return it.

provider/common/errors.go
```diff
@@ -80,3 +82,24 @@ var AuthorisationFailureStatusCodes = set.NewInts(
...
+		if invalidateErr != nil {
+			logger.Warningf("could not invalidate stored cloud credential on the controller: %v", invalidateErr)
+		}
+	}
+	return err, denied
+}
+
+// HandleCredentialError determines if a given error relates to an invalid credential.
+// If it is, the credential is invalidated. Original error is returned untouched.
+func HandleCredentialError(isAuthError func(error) bool, err error, ctx context.ProviderCallContext) error {
```
babbageclunk: Same here.

provider/common/errors.go
anastasiamac: Ahga \o/ yes, historical reasons as some providers indeed do but this code would not...
The reason why the errors are returned untouched is so that they can be used used in one line of code in similar way to errors.Trace(), etc...

anastasiamac:
$$merge$$

anastasiamac:
unrelated intermittent FAIL: upgrade_test.go:97: upgradeSuite.TestLoginsDuringUpgrade

anastasiamac:
and unrelated intermittent fail in github.com/juju/juju/worker/remoterelations.TestPackage

anastasiamac:
$$merge$$

PR #9383, opened 2018-10-29
Rework --help for `update-credential` command

pmatulis:
## Description of change

This improves the help output for the `update-credential` command.

Question: The current end of the help section has ``` `[1:] ``` at the very end. I didn't change that but all other commands I've seen lately drop to a new line and then end with a ``` ` ```.

## QA steps

View the help (`juju help update-credential`) and see the revised output.

## Documentation changes

N/A, but there is some precision here that would be good to eventually get into the docs.

## Bug reference

N/A

### DIFF
--- a/cmd/juju/cloud/updatecredential.go
+++ b/cmd/juju/cloud/updatecredential.go
@@ -17,31 +17,31 @@ import (
 )
 
 var usageUpdateCredentialSummary = `
-Updates a controller credential for a cloud.`[1:]
+Updates a remote credential.`[1:]
 
 var usageUpdateCredentialDetails = `
-Cloud credentials for controller are used for model operations and manipulations.
-Since it is common to have long-running models, it is also common to 
-have these cloud credentials become invalid during models' lifetime.
-When this happens, a user must update the cloud credential that 
-a model was created with to the new and valid details on controller.
+This command updates a credential cached on a controller. It does this by
+uploading the identically named local credential and validating it against the
+cloud. Use the ` + "`credentials`" + ` and ` + "`show-credential`" + ` commands to view local and
+remote credentials respectively.
 
-This command allows to update an existing, already-stored, named,
-cloud-specific controller credential.
+This command does not affect what models the updated credential may be related
+to. See command ` + "`set-credential`" + ` for that.
 
-NOTE: 
-This is the only command that will allow you to manipulate cloud
-credential for a controller. 
-All other credential related commands, such as 
-` + "`add-credential`" + `, ` + "`remove-credential`" + ` and  ` + "`credentials`" + ` 
-deal with credentials stored locally on the client not on the controller.
+This is the only command that can change the contents of an active credential.
+It is typically preceded by the ` + "`add-credential --replace`" + `  command.
 
 Examples:
-    juju update-credential aws mysecrets
+
+For cloud 'google', update remote credential 'jen':
+
+    juju update-credential google jen
 
 See also: 
     add-credential
-    credentials`[1:]
+    credentials
+    set-credential
+    show-credential`[1:]
 
 type updateCredentialCommand struct {
 	modelcmd.ControllerCommandBase


pmatulis:
[improve help for `update-credential` command](https://trello.com/c/RrktCaJd/733-improve-help-for-update-credential-command)

anastasiamac:
The [1:] drops the first char in a string. It all depends on how the string is constructed. Please do not remove it here.

I have not reviewed this PR yet. Will do soon.

cmd/juju/cloud/updatecredential.go
```diff
@@ -17,31 +17,31 @@ import (
...
-This command allows to update an existing, already-stored, named,
-cloud-specific controller credential.
-
-NOTE: 
-This is the only command that will allow you to manipulate cloud
-credential for a controller. 
-All other credential related commands, such as 
-` + "`add-credential`" + `, ` + "`remove-credential`" + ` and  ` + "`credentials`" + ` 
-deal with credentials stored locally on the client not on the controller.
+This command updates a credential cached on a controller. It does this by
```
anastasiamac: 'cached' is not suitable here. Maybe 'stored'?

cmd/juju/cloud/updatecredential.go
```diff
@@ -17,31 +17,31 @@ import (
...
+uploading the identically named local credential and validating it against the
```
anastasiamac: What? No it does not and even it did, it makes no sense to say so. How is it relevant to the usage?

cmd/juju/cloud/updatecredential.go
```diff
@@ -17,31 +17,31 @@ import (
...
+cloud. Use the ` + "`credentials`" + ` and ` + "`show-credential`" + ` commands to view local and
+remote credentials respectively.
+
+This command does not affect what models the updated credential may be related
```
anastasiamac: I do not see why this needs to be mentioned here?

cmd/juju/cloud/updatecredential.go
```diff
@@ -17,31 +17,31 @@ import (
...
+to. See command ` + "`set-credential`" + ` for that.
+
+This is the only command that can change the contents of an active credential.
+It is typically preceded by the ` + "`add-credential --replace`" + `  command.
```
anastasiamac: This is not needed here and in fact when the UX and flow will be addressed, this will become yet-another-place to revisit. Whoever will be doing the work, will need to be aware that this will become outdated.

cmd/juju/cloud/updatecredential.go
pmatulis: I wanted to ensure the user that when they "update a credential" that any other settings related to that credential are not affected. Referring to `set-credential` is just an extra.

cmd/juju/cloud/updatecredential.go
pmatulis: It *looks* like there is validation that the new credential is the same as the credential stored in its models. So, if this is the case, my bad, not cloud validation, but model validation. I said cloud because the error text is unclear:

```
Credential invalid for:
  pete-model:
    
The provided credentials could not be validated and 
may not be authorized to carry out the request.
Ensure that your account is authorized to use the Amazon EC2 service and 
that you are using the correct access keys. 
These keys are obtained via the "Security Credentials"
page in the AWS console.
: AWS was not able to validate the provided access credentials (AuthFailure)
Controller credential "pete" for user "admin" on cloud "aws" not updated: some models are no longer visible.
```

In passing, what is one supposed to do when the command errors out like this, which would be almost all the time?

cmd/juju/cloud/updatecredential.go
pmatulis: OK, I recall the term "cached" being used in the past.

cmd/juju/cloud/updatecredential.go
anastasiamac: Before we update the cloud credential content, we validate that it works for all the models that refer to it. If the credential succeeds/is valid for all models that use it, then its contents are updated on the controller. The message will list what models it was validated against.

If the model validation fails for even one of the models that use this credential, the update of the credential content will not happen. We will list the models that failed and the error, they would have failed with. In you r case, the credential content was still invalid and was not passing basic auth. There could be scenarios where the model references some machine instances but the new credential does not see them or vice versa - the instances are visible but Juju is not aware of them.

When  model validation for a new credential fails, depending on what error the model has, the user can consider:
(i) choosing different credential content and using update-credential again; 
(ii) choosing different credential for the failing model and using set-crdential;
(iii) if the issue is with machines/instances, addressing that, i.e. removing machine, etc (note that force remove machine is yet to come)...

cmd/juju/cloud/updatecredential.go
anastasiamac: Sure, but this feels like a general flow and is the concern outside of this page.. It's kind of like saying that add-model or deploy is preceded by 'bootstrap'...

It would suffice to say that the credential should be known to juju locally via either 'add-credential' or 'add-credential --replace'... 

Again, it feels like a lot of effort for the area that will have to be re-worded yet again shortly once the new changes to this CRUD commands are in.

cmd/juju/cloud/updatecredential.go
```diff
@@ -17,31 +17,31 @@ import (
 )
 
 var usageUpdateCredentialSummary = `
-Updates a controller credential for a cloud.`[1:]
+Updates a remote credential.`[1:]
```
mitechie: I don't know that remote is the best term here. Is remote jaas? Is it another machine?

mitechie:
    Beep..Boop...

PR #9391, opened 2018-11-01
Fixes lp#1800724: switching to non-existing model succeeds.

anastasiamac:
## Description of change

As per linked bug, it was possible to successfully switch to an already-destroyed model if:
* that model was considered as 'current' on this client;
* it was destroyed;
* there were no other model switches performed since this model's destruction.

In other words, if a previous current model was the same as the model the operator decided to switch to, and this model was destroyed, Juju will consider it as a no-op and will not let the operator know that the model no longer exists.

This was happening because we have short circuited the logic that sets current model - if the model name is the same, we would not perform certain checks.

This PR removes the short cut ensuring that the model name is always checks against known models.   

## QA steps

1. bootstrap;
2. add-model [this will also make it *current* on this client];
3. destroy-model form 2;
4. switch t model form 2. 

We should get an error. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1800724

### DIFF
--- a/cmd/juju/commands/switch_test.go
+++ b/cmd/juju/commands/switch_test.go
@@ -323,3 +323,32 @@ func (s *SwitchSimpleSuite) addController(c *gc.C, name string) {
 		User: "admin",
 	}
 }
+
+func (s *SwitchSimpleSuite) TestSwitchCurrentModelInStore(c *gc.C) {
+	s.store.CurrentControllerName = "same"
+	s.addController(c, "same")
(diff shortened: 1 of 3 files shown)


anastasiamac:
[can switch to destroyed model, lp#1800724](https://trello.com/c/dTbm5zsT/498-can-switch-to-destroyed-model-lp1800724)

anastasiamac:
$$merge$$

PR #9400, opened 2018-11-02
Openstack provider learns to react to an invalid cloud credential.

anastasiamac:
## Description of change

Juju uses cloud credentials stored on the controller against a model to communicate with the underlying cloud.
However, this credential may become invalid, for eg. expire. When this happens, Juju needs to stop trying to communicate with the cloud and wait for a user to provide either a new credential for the model or update the old one.

This PR ensures that openstack provider behaves as described.
The most interesting bit is the definition of what constitutes auth failure by the cloud, see errors.go. The rest is mechanical - ensure that errors from cloud calls are examined and when needed mark a cloud credential as invalid.

### DIFF
--- a/provider/openstack/firewaller.go
+++ b/provider/openstack/firewaller.go
@@ -20,6 +20,7 @@ import (
 	"github.com/juju/juju/environs/context"
 	"github.com/juju/juju/instance"
 	"github.com/juju/juju/network"
+	"github.com/juju/juju/provider/common"
 )
 
 const (
@@ -99,7 +100,7 @@ type switchingFirewaller struct {
 	fw Firewaller
 }
 
-func (f *switchingFirewaller) initFirewaller() error {
+func (f *switchingFirewaller) initFirewaller(ctx context.ProviderCallContext) error {
 	f.mu.Lock()
 	defer f.mu.Unlock()
 	if f.fw != nil {
@@ -109,6 +110,7 @@ func (f *switchingFirewaller) initFirewaller() error {
 	client := f.env.client()
 	if !client.IsAuthenticated() {
 		if err := authenticateClient(client); err != nil {
+			common.HandleCredentialError(IsAuthorisationFailure, err, ctx)
 			return errors.Trace(err)
 		}
 	}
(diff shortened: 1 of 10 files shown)


anastasiamac:
[adjust openstack provider to use callback to invalidate cloud credential when needed](https://trello.com/c/lUPvCKik/360-adjust-openstack-provider-to-use-callback-to-invalidate-cloud-credential-when-needed)

provider/openstack/firewaller.go
```diff
@@ -748,17 +774,32 @@ func (c *neutronFirewaller) updateGroupControllerUUID(group *neutron.SecurityGro
...
+	return nil
 }
 
 // IngressRules implements Firewaller interface.
 func (c *neutronFirewaller) IngressRules(ctx context.ProviderCallContext) ([]network.IngressRule, error) {
-	return c.ingressRules(c.ingressRulesInGroup)
+	rules, err := c.ingressRules(ctx, c.ingressRulesInGroup)
+	if err != nil {
+		common.HandleCredentialError(IsAuthorisationFailure, err, ctx)
+		return rules, errors.Trace(err)
```
wallyworld: This line can just be moved below replacing
return rules, nil
(here and elsewhere)
since errors.Trace(nil) returns nil

provider/openstack/provider.go
```diff
@@ -1810,12 +1920,22 @@ func validateAuthURL(authURL string) error {
...
+	return subnets, nil
 }
 
 // NetworkInterfaces is specified on environs.Networking.
 func (e *Environ) NetworkInterfaces(ctx context.ProviderCallContext, instId instance.Id) ([]network.InterfaceInfo, error) {
-	return e.networking.NetworkInterfaces(instId)
+	infos, err := e.networking.NetworkInterfaces(instId)
+	if err != nil {
+		common.HandleCredentialError(IsAuthorisationFailure, err, ctx)
+		return infos, errors.Trace(err)
```
wallyworld: ditto about err handling

provider/openstack/firewaller.go
anastasiamac: True. However, isn't it clearer to return nil when you know that there is no errors?

anastasiamac:
$$merge$$

PR #9403, opened 2018-11-05
Fixes lp#1793303: deploy fails if model missing from local cache.

anastasiamac:
## Description of change

Juju commands are able to query the controller if a user specifies a model that their client does not know about.

However, deploy command has been recently changed to support CAAS models. The change meant that Juju needed to know model type to validate fully at command.Init() stage. 

There is a scenario, see linked bug, where Juju will not be able to determine model type from local cache (the model is not there) and has to go to the controller. However, at Init() stage, the API connection has not been established yet. So, in cases where specified model is not known locally, we need to postpone some model-type based checks until command.Run() stage which can get API connection and refresh local models knowledge.

This PR does just that. Note that for all other scenarios, we still perform the validation as soon as possible, in command.Init().

## QA steps

1. bootstrap
2. remove default model details from client's local cache
3. 'juju deploy ubuntu -m default'

[SUCCESS] 'ubuntu' application gets deployed without errors and your local models' cache gains 'default' model details back.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1793303

### DIFF
--- a/cmd/juju/application/deploy.go
+++ b/cmd/juju/application/deploy.go
@@ -386,7 +386,7 @@ type DeployCommand struct {
 	machineMap string
 	flagSet    *gnuflag.FlagSet
 
-	notLocalModel bool
+	unknownModel bool
 }
 
 const deployDoc = `
@@ -723,7 +723,7 @@ func (c *DeployCommand) Init(args []string) error {
 		// at Init() we do not yet have an API connection.
 		// So we do not want to fail here if we encountered NotFoundErr, we want to
 		// do a late validation at Run().
-		c.notLocalModel = true
+		c.unknownModel = true
 	}
 	switch len(args) {
 	case 2:
@@ -764,7 +764,7 @@ func (c *DeployCommand) Init(args []string) error {
 		// at Init() we do not yet have an API connection.
 		// So we do not want to fail here if we encountered NotFoundErr, we want to
 		// do a late validation at Run().
-		c.notLocalModel = true
+		c.unknownModel = true
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


anastasiamac:
[Juju 2.5 deploy fails if model missing from local cache https://bugs.launchpad.net/bugs/1793303](https://trello.com/c/ChVPBBDE/510-juju-25-deploy-fails-if-model-missing-from-local-cache-https-bugslaunchpadnet-bugs-1793303)

jameinel:
From the description I assume we're not trying to handle the case where the model is known than the client doesn't connect, the model is deleted and replaced by a CAAS model.
It does seem like relying on the local cache in order to do validation is always a bit risky.

Is it just to give a better user experience most of the time by having faster cli parsing feedback?

anastasiamac:
@jameinel,
I am not sure I understand your first statement.

All we are catering for here is the case where a client does not know about the model that the operator wants to deploy to. However, the model exists on the controller. This works with all previous versions of Juju.

We have very recently added validation to 'deploy' command Init() method that needs model type for its logic. However, in the situation described above, we would not know model type until we are able to make a connection to controller. This connection is possible in commands' Run(). So for this particular case, the validation is postponed until Run().

If model is known locally, the validation will still happen early in Init() and not in Run(). 

If your local model information is outdated, this command will behave as all others and is not a concern of this PR.

cmd/juju/application/deploy.go
```diff
@@ -385,6 +385,8 @@ type DeployCommand struct {
 
 	machineMap string
 	flagSet    *gnuflag.FlagSet
+
+	notLocalModel bool
```
Veebers: I'm not sure notLocalModel is the best name here, perhaps something like notKnownModel or notCachedModel or perhaps even modelKnownAtInitTime (no that's too wordy :-) ) to indicate we're dealing with a model that we don't currently know about.

anastasiamac: I was originally thinking notKnownLocallyModel but is it really adding more clarity?

anastasiamac: I'll just rename to unknownModel.

anastasiamac:
$$merge$$

PR #9405, opened 2018-11-06
undertaker watches namespace for caas

ycliuhw:
## Description of change

1. undertaker watches namespace for caas model;
2. fixed a few related bugs;

## QA steps

1. prepare a caas model;
2. deploy some workload;
3. juju destroy-model t1 --debug -y --destroy-storage

Now, we will see the destroy cmd waits until namespace's terminated.

## Documentation changes

No document change is required because it's internal logic change.

### DIFF
--- a/caas/kubernetes/provider/k8s_test.go
+++ b/caas/kubernetes/provider/k8s_test.go
@@ -331,8 +331,10 @@ func (s *K8sBrokerSuite) TestDestroy(c *gc.C) {
 			v1.ListOptions{LabelSelector: "juju-model==test"},
 		).Times(1).
 			Return(s.k8sNotFoundError()),
+		// still terminating.
 		s.mockNamespaces.EXPECT().Get("test", v1.GetOptions{IncludeUninitialized: true}).Times(1).
-			Return(nil, errors.New("any non k8s not found error")),
+			Return(ns, nil),
+		// terminated, not found returned.
 		s.mockNamespaces.EXPECT().Get("test", v1.GetOptions{IncludeUninitialized: true}).Times(1).
 			Return(nil, s.k8sNotFoundError()),
 	)
@@ -348,7 +350,7 @@ func (s *K8sBrokerSuite) TestDestroy(c *gc.C) {
 
 	err := s.broker.Destroy(context.NewCloudCallContext())
 	c.Assert(err, jc.ErrorIsNil)
-	c.Assert(workertest.CheckKilled(c, s.watcher), gc.ErrorMatches, "bad catacomb Kill: tomb.ErrDying")
+	c.Assert(workertest.CheckKilled(c, s.watcher), jc.ErrorIsNil)
 }
 
 func (s *K8sBrokerSuite) TestDeleteOperator(c *gc.C) {
--- a/caas/kubernetes/provider/k8s_test.go
+++ b/caas/kubernetes/provider/k8s_test.go
@@ -323,18 +323,18 @@ func (s *K8sBrokerSuite) TestDestroy(c *gc.C) {
 		// first delete returns nil
 		s.mockNamespaces.EXPECT().Delete("test", s.deleteOptions(v1.DeletePropagationForeground)).Times(1).
 			Return(nil),
-		// Get now returns non NotFoundError.
-		s.mockNamespaces.EXPECT().Get("test", v1.GetOptions{IncludeUninitialized: true}).Times(1).
-			Return(nil, errors.New("any non k8s not found error")),
-		s.mockNamespaces.EXPECT().Delete("test", s.deleteOptions(v1.DeletePropagationForeground)).Times(1).
-			Return(s.k8sNotFoundError()),
-		s.mockNamespaces.EXPECT().Get("test", v1.GetOptions{IncludeUninitialized: true}).Times(1).
-			Return(nil, s.k8sNotFoundError()),
 		s.mockStorageClass.EXPECT().DeleteCollection(
 			s.deleteOptions(v1.DeletePropagationForeground),
 			v1.ListOptions{LabelSelector: "juju-model==test"},
 		).Times(1).
 			Return(s.k8sNotFoundError()),
+		// Get now returns non NotFoundError.
+		s.mockNamespaces.EXPECT().Get("test", v1.GetOptions{IncludeUninitialized: true}).Times(1).
+			Return(nil, errors.New("any non k8s not found error")),
+		// s.mockNamespaces.EXPECT().Delete("test", s.deleteOptions(v1.DeletePropagationForeground)).Times(1).
+		// 	Return(s.k8sNotFoundError()),
+		s.mockNamespaces.EXPECT().Get("test", v1.GetOptions{IncludeUninitialized: true}).Times(1).
+			Return(nil, s.k8sNotFoundError()),
 	)
 
 	go func(w *watch.FakeWatcher) {
--- a/caas/kubernetes/provider/k8s_test.go
+++ b/caas/kubernetes/provider/k8s_test.go
@@ -351,6 +351,7 @@ func (s *K8sBrokerSuite) TestDestroy(c *gc.C) {
 	err := s.broker.Destroy(context.NewCloudCallContext())
 	c.Assert(err, jc.ErrorIsNil)
 	c.Assert(workertest.CheckKilled(c, s.watcher), jc.ErrorIsNil)
+	c.Assert(namespaceWatcher.IsStopped(), jc.IsTrue)
 }
 
 func (s *K8sBrokerSuite) TestDeleteOperator(c *gc.C) {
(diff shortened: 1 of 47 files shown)


ycliuhw:
[k8s model destruction - new worker to wait for namespace deletion to complete, similar to IAAS](https://trello.com/c/hav4VDX4/514-k8s-model-destruction-new-worker-to-wait-for-namespace-deletion-to-complete-similar-to-iaas)

caas/kubernetes/provider/k8s_test.go
```diff
@@ -276,19 +313,42 @@ func (s *K8sBrokerSuite) TestDestroy(c *gc.C) {
...
+			Return(namespaceWatcher, nil),
 		s.mockNamespaces.EXPECT().Delete("test", s.deleteOptions(v1.DeletePropagationForeground)).Times(1).
-			Return(s.k8sNotFoundError()),
+			Return(nil),
 		s.mockStorageClass.EXPECT().DeleteCollection(
 			s.deleteOptions(v1.DeletePropagationForeground),
 			v1.ListOptions{LabelSelector: "juju-model==test"},
 		).Times(1).
 			Return(s.k8sNotFoundError()),
+		s.mockNamespaces.EXPECT().Get("test", v1.GetOptions{IncludeUninitialized: true}).Times(1).
```
wallyworld: This doesn't look right? We expect either nil if namespace still exists, or not found when it goes away. Any other error should cause Destroy() to fail

caas/kubernetes/provider/k8s_test.go
```diff
@@ -276,19 +313,42 @@ func (s *K8sBrokerSuite) TestDestroy(c *gc.C) {
...
+			if !w.IsStopped() {
+				clk.WaitAdvance(time.Second, testing.LongWait, 1)
+				f(ns)
+			}
+		}
+	}(namespaceWatcher, s.clock)
+
 	err := s.broker.Destroy(context.NewCloudCallContext())
 	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(workertest.CheckKilled(c, s.watcher), gc.ErrorMatches, "bad catacomb Kill: tomb.ErrDying")
```
wallyworld: Normally we expect CheckKilled to test for a nil error.
"bad catacomb kill" doesn't look right

caas/kubernetes/provider/k8swatcher.go
```diff
@@ -16,24 +17,26 @@ import (
...
 )
 
-// kubernetesWatcher reports changes to kubernetes
+// KubernetesWatcher reports changes to kubernetes
 // resources. A native kubernetes watcher is passed
 // in to generate change events from the kubernetes
 // model. These events are consolidated into a Juju
 // notification watcher event.
-type kubernetesWatcher struct {
+type KubernetesWatcher struct {
```
wallyworld: instead of exporting this can we add a line to export_test.go

caas/kubernetes/provider/k8swatcher.go
```diff
@@ -62,13 +65,16 @@ func (w *kubernetesWatcher) loop() error {
 			}
 			logger.Tracef("received k8s event: %+v", evt.Type)
 			if pod, ok := evt.Object.(*core.Pod); ok {
-				logger.Tracef("%v(%v) = %v, status=%+v", pod.Name, pod.UID, pod.Labels, pod.Status)
+				logger.Debugf("%v(%v) = %v, status=%+v", pod.Name, pod.UID, pod.Labels, pod.Status)
```
wallyworld: let's revert to trace here and below

caas/kubernetes/provider/provider.go
```diff
@@ -50,7 +51,7 @@ func (kubernetesEnvironProvider) Open(args environs.OpenParams) (caas.Broker, er
 	if err := validateCloudSpec(args.Cloud); err != nil {
 		return nil, errors.Annotate(err, "validating cloud spec")
 	}
-	broker, err := NewK8sBroker(args.Cloud, args.Config, newK8sClient)
+	broker, err := NewK8sBroker(args.Cloud, args.Config, newK8sClient, nil, jujuclock.WallClock)
```
wallyworld: I guess we could pass in newKubernetesWatcher here instead of nil, and remove the nil check in NewK8sBroker. I think that would be cleaner?

state/model.go
```diff
@@ -1160,7 +1160,7 @@ func (m *Model) destroyOps(
 			prereqOps = storageOps
 		}
 	} else {
-		if !m.isControllerModel() {
+		if !m.isControllerModel() && m.Type() == ModelTypeIAAS {
```
wallyworld: Let's add a good comment here to explain why k8s models are different

state/state.go
```diff
@@ -211,12 +211,38 @@ func ControllerAccess(st *State, tag names.Tag) (permission.UserAccess, error) {
 	return st.UserAccess(tag.(names.UserTag), st.controllerTag)
 }
 
-// RemoveAllModelDocs removes all documents from multi-model
+// RemoveModel removes all documents from multi-model
```
wallyworld: The comment needs to be changed since RemoveModel() does more than just remove the documents

ycliuhw:
!!build!!

caas/kubernetes/provider/k8s_test.go
ycliuhw: yes, you are right. just fixed a bug in `kubernetesWatcher.loop`. now it's nil

caas/kubernetes/provider/provider.go
ycliuhw: agreed. changed.

caas/kubernetes/provider/k8swatcher.go
ycliuhw: now it's exposed in export_test

caas/kubernetes/provider/k8s_test.go
ycliuhw: yes, fixed!

ycliuhw:
!!build!!

ycliuhw:
!!build!!

ycliuhw:
!!build!!

state/state.go
```diff
@@ -211,12 +211,59 @@ func ControllerAccess(st *State, tag names.Tag) (permission.UserAccess, error) {
...
+	}
+	if err = st.db().Run(buildTxn); err != nil {
+		return errors.Trace(err)
+	}
+	return nil
+}
+
+// RemoveModel sets current model to dead then removes all documents from
+// multi-model collections.
+func (st *State) RemoveModel() error {
```
wallyworld: I think this should be RemoveDyingModel(). I think that we should never short circuit model.Destroy() to set the next life to Dead. This also improves IAAS model destruction because even for IAAS models Destroy() can take time to run. So in places where we set next model life to Dead because there are no entity refs, perhaps the next life should always be Dying; the undertaker runs environ.Destroy(); then calls RemoveDyingModel(). I think that allows the SetDyingModelToDead()  call to not be needed.

ycliuhw:
!!build!!

ycliuhw:
!!build!!

cmd/juju/controller/destroy.go
```diff
@@ -250,7 +250,8 @@ upgrade the controller to version 2.3 or greater.
 		}
 
 		updateStatus := newTimedStatusUpdater(ctx, api, controllerEnviron.Config().UUID(), clock.WallClock)
-		envStatus := updateStatus(0)
+		// wait for 2 seconds to let empty hosted models changed from alive to dying.
+		envStatus := updateStatus(2 * time.Second)
```
wallyworld: We can't do this - we can't rely on an arbitrary wait time. It is fragile and will break on systems like prodstack that are heavily loaded. Without the wait time, what happened? An error? It may be that we need to introduce a new for loop to poll status to wait for models to transition to dead

state/model.go
```diff
@@ -1225,6 +1219,7 @@ func (m *Model) destroyOps(
 				if !IsModelNotEmptyError(err) {
 					return nil, errors.Trace(err)
 				}
+				logger.Errorf("destroy controller: got alive empty hosted model....")
```
wallyworld: this should be debug right?

state/state.go
```diff
@@ -211,12 +211,55 @@ func ControllerAccess(st *State, tag names.Tag) (permission.UserAccess, error) {
...
+	}
+	if err := st.db().Run(buildTxn); err != nil {
+		return errors.Trace(err)
+	}
+	return nil
+}
+
+// RemoveDyingModel sets current model to dead then removes all documents from
+// multi-model collections.
+func (st *State) RemoveDyingModel() error {
```
wallyworld: We need a test for models in all life states - alive, dying, dead. Do we have all those tests or de we need to add one?

ycliuhw: yes, we have some tests for this, and I just added more tests for both RemoveDyingModel and setDyingModelToDead.

ycliuhw:
!!build!!

ycliuhw:
!!build!!

ycliuhw:
$$merge$$

PR #9416, opened 2018-11-07
apiserver: Skip unprovisioned machines when checking cloud creds

babbageclunk:
## Description of change

If I do an add-unit with a bad cloud credential, a new machine gets created in the model but when the provisioner tries to start a corresponding instance the credential will get invalidated. If I try to use update-credential to mark the credential as valid again, it fails because of the NotProvisioned error from getting the machine's instance id. Change the check to skip unprovisioned machines - we don't expect the cloud to know about these anyway.

## QA steps

* Bootstrap a controller with a credential that you can disable in the cloud provider (and deploy an application).
* Disable the credential.
* Add a unit to the application. This will create a machine for the unit, but when we try to provision it the cred will be invalidated.
* Reenable the cred in the cloud.
* Use update-credential to reactivate the credential - it succeeds.
* The machine and unit get added successfully.

### DIFF
--- a/apiserver/common/credentialcommon/modelcredential.go
+++ b/apiserver/common/credentialcommon/modelcredential.go
@@ -119,7 +119,11 @@ func checkMachineInstances(backend PersistentBackend, provider CloudProvider, ca
 			continue
 		}
 		instanceId, err := machine.InstanceId()
-		if err != nil {
+		if errors.IsNotProvisioned(err) {
+			// Skip over this machine; we wouldn't expect the cloud
+			// to know about it.
+			continue
+		} else if err != nil {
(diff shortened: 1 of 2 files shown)


anastasiamac:
unit tests passed, windows failed

babbageclunk:
$$merge$$

anastasiamac:
[when adding a machineon the model with invalid credential, machine will be stored in state](https://trello.com/c/G4tR8udz/522-when-adding-a-machineon-the-model-with-invalid-credential-machine-will-be-stored-in-state)

babbageclunk:
$$merge$$

PR #9422, opened 2018-11-08
lease-clock-updater: don't crash during startup

babbageclunk:
## Description of change

We used to get a noisy error from the lease clock updater during controller startup, because the raft forwarder wasn't running when it tried to update the clock so the lease store would return a timeout error. To avoid this we make the lease clock updater start depend on the raft forwarder, so that it doesn't try to update the clock until it's available. Also add a timeout error to the globalclock interface and have the clock updater worker handle it, rather than crashing.

## QA steps

* Bootstrap a controller.
* The timeout error from the lease clock updater doesn't appear in the logs during controller startup any more.


babbageclunk:
$$merge$$

wallyworld:
[prevent lease-clock-updater timeout crash during startup](https://trello.com/c/A5l8Xorq/458-prevent-lease-clock-updater-timeout-crash-during-startup)

PR #9424, opened 2018-11-08
Enable invalid credential handling for LXD

Veebers:
## Description of change

Invalidate the credential whenever we get an authorisation error back from any provider action.

### DIFF
--- a/provider/lxd/environ_broker_test.go
+++ b/provider/lxd/environ_broker_test.go
@@ -23,15 +23,21 @@ import (
 type environBrokerSuite struct {
 	lxd.EnvironSuite
 
-	callCtx        context.ProviderCallContext
-	defaultProfile *api.Profile
+	callCtx           *context.CloudCallContext
+	defaultProfile    *api.Profile
+	invalidCredential bool
 }
 
 var _ = gc.Suite(&environBrokerSuite{})
 
 func (s *environBrokerSuite) SetUpTest(c *gc.C) {
 	s.BaseSuite.SetUpTest(c)
-	s.callCtx = context.NewCloudCallContext()
+	s.callCtx = &context.CloudCallContext{
+		InvalidateCredentialFunc: func(string) error {
+			s.invalidCredential = true
+			return nil
+		},
+	}
 	s.defaultProfile = &api.Profile{
 		ProfilePut: api.ProfilePut{
 			Devices: map[string]map[string]string{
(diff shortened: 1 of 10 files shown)


provider/lxd/environ_broker_test.go
```diff
@@ -350,7 +361,28 @@ func (s *environBrokerSuite) TestStartInstanceNoTools(c *gc.C) {
...
+	)
+
+	env := s.NewEnviron(c, svr, nil)
+	_, err := env.StartInstance(s.callCtx, s.GetStartInstanceArgs(c, "bionic"))
+	c.Assert(err, gc.ErrorMatches, "not authorized")
+	c.Assert(s.invalidCredential, jc.IsTrue)
+}
+
 func (s *environBrokerSuite) TestStopInstances(c *gc.C) {
+	c.Assert(s.invalidCredential, jc.IsFalse)
```
anastasiamac: Why is this here?
If we are checking that the credential is valid in the beginning of the test, I would have expected that the test makes some calls to invalidate the credential and then check that it is indeed invalid at the end. But it does not seem to be the case here :)

I think you wanted to make this call int he beginning of TestStopInstancesInvalidCredentials... It is missing from there :)

provider/lxd/environ_test.go
```diff
@@ -200,6 +250,121 @@ func (s *environSuite) TestDestroyController(c *gc.C) {
...
+
+	s.Client.Containers = append(s.Client.Containers, *machine0)
+
+	// RemoveContainers will error not-auth.
+	s.Client.Stub.SetErrors(nil, nil, nil, nil, nil, errTestUnAuth)
+
+	err := s.Env.DestroyController(s.callCtx, s.Config.UUID())
+	c.Assert(err, gc.ErrorMatches, "not authorized")
+	c.Assert(s.invalidCredential, jc.IsTrue)
+	s.Stub.CheckCalls(c, []gitjujutesting.StubCall{
```
anastasiamac: I would have expected to see a check that controller was not destroyed here :)

anastasiamac:
[adjust lxd provider to use callback to invalidate cloud credential when needed](https://trello.com/c/ymHuqezY/362-adjust-lxd-provider-to-use-callback-to-invalidate-cloud-credential-when-needed)

Veebers:
$$merge$$

PR #9425, opened 2018-11-08
Only show local subordinates in goal state output

wallyworld:
## Description of change

Ensure that only local subordinate units are in the "relations" section of goal state for a given unit.
The tests required a bit of surgery to swap over the unit for which the api call was being made, since mysql is the unit that could best handle having a subordinate added to it.

## QA steps

deploy a charm with multiple units
add a subordinate relation
check goal state to see that only each local subordinate unit is in goal state for any given principal

## Bug reference

https://bugs.launchpad.net/juju/+bug/1801765


testcharms/charm-repo/quantal/mysql/metadata.yaml
```diff
@@ -6,3 +6,6 @@ provides:
     interface: mysql
   server-admin:
     interface: mysql-root
+requires:
+  monitoring-client:
+    interface: monitoring
```
jameinel: I think we want a trailing newline :)

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -2634,16 +2632,21 @@ func (u *UniterAPI) goalStateRelations(appName string, allRelations []*state.Rel
...
 	if err != nil {
 		return nil, err
 	}
 	unitsGoalState := params.UnitsGoalState{}
 	for _, unit := range allUnits {
+		// Ignore subordinates belonging to other units.
+		pn, ok := unit.PrincipalName()
+		if ok && pn != principalName {
+			continue
+		}
```
jameinel: Presumably this is the core of the change, and the rest is just being able to pass in principalName into this code, and a small amount of refactoring.

wallyworld: yup

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
[goal state calculation for subordinates](https://trello.com/c/Nb36Prh7/523-goal-state-calculation-for-subordinates)

PR #9430, opened 2018-11-09
Add add-unit support for kubernetes models.

Veebers:
## Description of change

Put a veneer over the ```scale-application``` command and allow ```add-unit``` to work with K8s models albeit supporting only the ```num-units``` argument.

## Documentation changes

Added a 'note' in the commands help text although this might need to be expanded.

### DIFF
--- a/api/application/client.go
+++ b/api/application/client.go
@@ -596,10 +596,10 @@ type ScaleApplicationParams struct {
 	// ApplicationName is the application to scale.
 	ApplicationName string
 
-	// Scale is the target number of units which should exist
+	// Scale is the target number of units which should should be running.
 	Scale int
 
-	// ScaleChange is the amount of change to the target amount of existing units
+	// ScaleChange is the amount of change to the target number of existing units.
 	ScaleChange int
 }
 
@@ -608,6 +608,11 @@ func (c *Client) ScaleApplication(in ScaleApplicationParams) (params.ScaleApplic
 	if !names.IsValidApplication(in.ApplicationName) {
 		return params.ScaleApplicationResult{}, errors.NotValidf("application %q", in.ApplicationName)
 	}
+
+	if err := validateApplicationScale(in.Scale, in.ScaleChange); err != nil {
+		return params.ScaleApplicationResult{}, errors.Trace(err)
+	}
+
 	args := params.ScaleApplicationsParams{
 		Applications: []params.ScaleApplicationParams{{
 			ApplicationTag: names.NewApplicationTag(in.ApplicationName).String(),
@@ -935,4 +940,15 @@ func (c *Client) GetLXDProfileUpgradeMessages(applicationName string, watcherId 
 		messages[k].Message = v.Message
 	}
 	return messages, nil
+}
+
+func validateApplicationScale(scale, scaleChange int) error {
+	if scale == 0 && scaleChange == 0 {
+		return errors.NotValidf("scale of 0")
+	} else if scale < 0 && scaleChange == 0 {
+		return errors.NotValidf("scale < 0")
+	} else if scale != 0 && scaleChange != 0 {
+		return errors.NotValidf("requesting both scale and scale-change")
+	}
+	return nil
 }
--- a/api/application/client.go
+++ b/api/application/client.go
@@ -596,8 +596,11 @@ type ScaleApplicationParams struct {
 	// ApplicationName is the application to scale.
 	ApplicationName string
 
-	// Scale is the number of units which should be running.
+	// Scale is the target number of units which should exist
 	Scale int
+
+	// ScaleChange is the amount of change to the target amount of existing units
+	ScaleChange int
 }
 
 // ScaleApplication sets the desired unit count for one or more applications.
@@ -609,6 +612,7 @@ func (c *Client) ScaleApplication(in ScaleApplicationParams) (params.ScaleApplic
 		Applications: []params.ScaleApplicationParams{{
 			ApplicationTag: names.NewApplicationTag(in.ApplicationName).String(),
 			Scale:          in.Scale,
+			ScaleChange:    in.ScaleChange,
 		}},
 	}
 	var results params.ScaleApplicationResults
--- a/api/application/client_test.go
+++ b/api/application/client_test.go
@@ -1182,6 +1182,41 @@ func (s *applicationSuite) TestScaleApplicationArity(c *gc.C) {
 	c.Assert(err, gc.ErrorMatches, "expected 1 result, got 2")
 }
 
+func (s *applicationSuite) TestScaleApplicationValidation(c *gc.C) {
+	apiCaller := basetesting.APICallerFunc(
+		func(objType string, version int, id, request string, a, response interface{}) error {
+			return nil
+		},
+	)
+	client := application.NewClient(apiCaller)
+
+	for i, test := range []struct {
+		scale       int
+		scaleChange int
+		errorStr    string
+	}{{
+		scale:       5,
+		scaleChange: 5,
+		errorStr:    "requesting both scale and scale-change not valid",
+	}, {
+		scale:       0,
+		scaleChange: 0,
+		errorStr:    "scale of 0 not valid",
+	}, {
+		scale:       -1,
+		scaleChange: 0,
+		errorStr:    "scale < 0 not valid",
+	}} {
+		c.Logf("test #%d", i)
+		_, err := client.ScaleApplication(application.ScaleApplicationParams{
+			ApplicationName: "foo",
+			Scale:           test.scale,
+			ScaleChange:     test.scaleChange,
+		})
+		c.Assert(err, gc.ErrorMatches, test.errorStr)
+	}
+}
+
 func (s *applicationSuite) TestScaleApplicationError(c *gc.C) {
 	apiCaller := basetesting.APICallerFunc(
 		func(objType string, version int, id, request string, a, response interface{}) error {
--- a/api/application/client_test.go
+++ b/api/application/client_test.go
@@ -1154,6 +1154,36 @@ func (s *applicationSuite) TestScaleApplication(c *gc.C) {
 	})
 }
 
+func (s *applicationSuite) TestChangeScaleApplication(c *gc.C) {
+	apiCaller := basetesting.APICallerFunc(
+		func(objType string, version int, id, request string, a, response interface{}) error {
+			c.Assert(request, gc.Equals, "ScaleApplications")
+			args, ok := a.(params.ScaleApplicationsParams)
+			c.Assert(ok, jc.IsTrue)
+			c.Assert(args, jc.DeepEquals, params.ScaleApplicationsParams{
+				Applications: []params.ScaleApplicationParams{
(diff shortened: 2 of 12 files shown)


api/application/client.go
```diff
@@ -596,8 +596,11 @@ type ScaleApplicationParams struct {
 	// ApplicationName is the application to scale.
 	ApplicationName string
 
-	// Scale is the number of units which should be running.
+	// Scale is the target number of units which should exist
```
wallyworld: I think "be running." is ok?
If it is changed to exist, end with a "."

api/application/client.go
```diff
@@ -609,6 +612,7 @@ func (c *Client) ScaleApplication(in ScaleApplicationParams) (params.ScaleApplic
 		Applications: []params.ScaleApplicationParams{{
 			ApplicationTag: names.NewApplicationTag(in.ApplicationName).String(),
 			Scale:          in.Scale,
+			ScaleChange:    in.ScaleChange,
```
wallyworld: We should validate scale args here, just like in.ApplicatioName above

apiserver/facades/client/application/application.go
```diff
@@ -1582,9 +1582,14 @@ func (api *APIBase) ScaleApplications(args params.ScaleApplicationsParams) (para
 		return params.ScaleApplicationResults{}, errors.Trace(err)
 	}
 	scaleApplication := func(arg params.ScaleApplicationParams) (*params.ScaleApplicationInfo, error) {
-		if arg.Scale < 0 {
+		if arg.Scale == 0 && arg.ScaleChange == 0 {
+			return nil, errors.NotValidf("scale of 0")
+		} else if arg.Scale < 0 && arg.ScaleChange == 0 {
 			return nil, errors.NotValidf("scale < 0")
+		} else if arg.Scale != 0 && arg.ScaleChange != 0 {
+			return nil, errors.NotValidf("requesting scale and scale-change")
```
wallyworld: requesting both

apiserver/facades/client/application/application.go
```diff
@@ -1596,6 +1601,14 @@ func (api *APIBase) ScaleApplications(args params.ScaleApplicationsParams) (para
 		} else if err != nil {
 			return nil, errors.Trace(err)
 		}
+		if arg.ScaleChange != 0 {
+			currentScale := app.GetScale()
+			newScale := currentScale + arg.ScaleChange
```
wallyworld: It's not quite this simple as we should add an assert that the current scale hasn't changed. This functionality needs to be moved to state where an assert can be used. ie introduce a ChangeScale() method.

cmd/juju/application/addunit.go
```diff
@@ -35,6 +36,10 @@ with any application or model constraints. This command also supports the
 placement directive ("--to") for targeting specific machines or containers,
 which will bypass application and model constraints.
 
+Note:
+For Kubernetes models the only valid argument is -n, --num-units anything
```
wallyworld: full stop after --num-units and Anything on next line

cmd/juju/application/addunit.go
```diff
@@ -184,6 +188,15 @@ func (c *addUnitCommand) Init(args []string) error {
 	if err := cmd.CheckEmpty(args[1:]); err != nil {
 		return err
 	}
+	modelType, err := c.ModelType()
```
wallyworld: We won't always know the model type in Init()
(see destroy)

cmd/juju/application/addunit.go
```diff
@@ -184,6 +188,15 @@ func (c *addUnitCommand) Init(args []string) error {
...
+	if err != nil {
+		return err
+	}
+	if modelType == model.CAAS {
+		if c.PlacementSpec != "" || len(c.AttachStorage) != 0 {
+			return errors.New("Kubernetes models only supports num-units flag")
```
wallyworld: ... support --num-units"

api/application/client.go
Veebers: I'm not sure what validation makes sense here as remove-unit will pass this a negative number for the ScaleChange. Thoughts?

wallyworld: No, I meant that we should use the same validation as on the server side, eg both can't be set

apiserver/facades/client/application/application_unit_test.go
```diff
@@ -639,6 +639,58 @@ func (s *ApplicationSuite) TestScaleApplicationsCAASModel(c *gc.C) {
...
+		}},
+	})
+	app := s.backend.applications["postgresql"]
+	app.CheckCall(c, 0, "ChangeScale", 5)
+}
+
+func (s *ApplicationSuite) TestScaleApplicationsCAASModelScaleArgCheck(c *gc.C) {
+	application.SetModelType(s.api, state.ModelTypeCAAS)
+	s.backend.applications["postgresql"].scale = 2
+	results, err := s.api.ScaleApplications(params.ScaleApplicationsParams{
```
wallyworld: These would typically be done as a loop
eg see github.com/juju/juju/cmd/juju/subnet/remove_test.go
func (s *RemoveSuite) TestInit(c *gc.C)

cmd/juju/application/addunit.go
```diff
@@ -217,6 +243,28 @@ func (c *addUnitCommand) Run(ctx *cmd.Context) error {
...
+		return err
+	}
+
+	if modelType == model.CAAS {
+		_, err = apiclient.ScaleApplication(application.ScaleApplicationParams{
+			ApplicationName: c.ApplicationName,
+			ScaleChange:     c.NumUnits,
+		})
+		if params.IsCodeUnauthorized(err) {
+			common.PermissionsMessage(ctx.Stderr, "add a unit")
```
wallyworld: "scale an application" perhaps

state/application.go
```diff
@@ -1337,6 +1337,39 @@ func (a *Application) GetScale() int {
 	return a.doc.DesiredScale
 }
 
+// ChangeScale alters the existing scale by the provided change amount, returning the new amount.
+// This is used on CAAS models.
+func (a *Application) ChangeScale(scaleChange int) (int, error) {
+	newScale := a.doc.DesiredScale + scaleChange
+	if newScale < 1 {
```
wallyworld: newScale < 0

state/application.go
```diff
@@ -1337,6 +1337,39 @@ func (a *Application) GetScale() int {
...
+		return a.doc.DesiredScale, errors.NotValidf("cannot remove more units than currently exist")
+	}
+	buildTxn := func(attempt int) ([]txn.Op, error) {
+		if attempt > 0 {
```
wallyworld: Need to redo the check that resulting scale is >=0 since if attempt > 0, the scale may have changed underneath us

state/application.go
```diff
@@ -1337,6 +1337,39 @@ func (a *Application) GetScale() int {
...
+				return nil, errors.Trace(err)
+			} else if !alive {
+				return nil, applicationNotAliveErr
+			}
+		}
+		return []txn.Op{{
+			C:  applicationsC,
+			Id: a.doc.DocID,
+			Assert: bson.D{{"life", Alive},
+				{"charmurl", a.doc.CharmURL},
```
wallyworld: So this is a bug in Scale() also.
If any of these asserts fail, up above in if attempt > 0, we need to refresh the app to get the latest data, or else we'll just keep looping
			if err := a.Refresh(); err != nil {
				return nil, errors.Trace(err)
			}

api/application/client_test.go
```diff
@@ -1182,6 +1182,41 @@ func (s *applicationSuite) TestScaleApplicationArity(c *gc.C) {
 	c.Assert(err, gc.ErrorMatches, "expected 1 result, got 2")
 }
 
+func (s *applicationSuite) TestScaleApplicationValidation(c *gc.C) {
```
wallyworld: Not a validation test, but we're missing a test for success where scale change is used but not scale. ie a test similar to TestScaleApplication() but for scale change

cmd/juju/application/addunit.go
```diff
@@ -35,6 +36,10 @@ with any application or model constraints. This command also supports the
 placement directive ("--to") for targeting specific machines or containers,
 which will bypass application and model constraints.
 
+Note:
```
wallyworld: As with comment in remove-unit PR, I think we need to explain up front that k8s and cloud models are different and preface a block of text for each type.

cmd/juju/application/addunit_test.go
```diff
@@ -66,6 +66,17 @@ func (f *fakeApplicationAddUnitAPI) AddUnits(args apiapplication.AddUnitsParams)
 	return nil, nil
 }
 
+func (f *fakeApplicationAddUnitAPI) ScaleApplication(args apiapplication.ScaleApplicationParams) (params.ScaleApplicationResult, error) {
```
wallyworld: As with destroy, we need test(s) similar to TestDestroyUnknownModelCallsRefresh etc

state/application.go
```diff
@@ -1337,6 +1337,46 @@ func (a *Application) GetScale() int {
...
+			alive, err := isAlive(a.st, applicationsC, a.doc.DocID)
+			if err != nil {
+				return nil, errors.Trace(err)
+			} else if !alive {
+				return nil, applicationNotAliveErr
+			}
+			if err := a.Refresh(); err != nil {
+				return nil, errors.Trace(err)
+			}
+			newScale = a.doc.DesiredScale + scaleChange
```
wallyworld: Something to ponder - I wonder if the scale changes from underneath us, we should just error and not try again. ie if I saw an app had 5 units and I wanted to add 2 more expecting 7, and the scale changed to 4 before my txn ran, the end result would be 6 and not what i expected. Maybe it's better just to error in that case? With a suitable message. Although add-unit has the same problem so maybe not an issue to address right now

state/application.go
```diff
@@ -1337,6 +1337,46 @@ func (a *Application) GetScale() int {
...
+	}
+	buildTxn := func(attempt int) ([]txn.Op, error) {
+		if attempt > 0 {
+			alive, err := isAlive(a.st, applicationsC, a.doc.DocID)
+			if err != nil {
+				return nil, errors.Trace(err)
+			} else if !alive {
+				return nil, applicationNotAliveErr
+			}
+			if err := a.Refresh(); err != nil {
```
wallyworld: Maybe too subtle - I was hoping for a driveby fix in Scale() also :-)

Veebers:
$$merge$$

cmd/juju/application/addunit.go
```diff
@@ -217,6 +252,28 @@ func (c *addUnitCommand) Run(ctx *cmd.Context) error {
 	}
 	defer apiclient.Close()
 
+	if c.unknownModel {
+		if err := c.validateArgs(); err != nil {
+			return errors.Trace(err)
+		}
+	}
+
+	modelType, err := c.ModelType()
```
wallyworld: Same comment as for remove-unit - I think we pass model type to validate() method.
Also, can we be consistent with naming of the method between add and remove

Veebers:
$$merge$$

PR #9436, opened 2018-11-12
Add remove-unit support for kubernetes models.

Veebers:
## Description of change

Put a veneer over the ```scale-application``` command and allow ```remove-unit``` to work with K8s models albeit supporting only the ```num-units``` argument.

## Documentation changes

Added a 'note' in the commands help text although this might need to be expanded.

### DIFF
--- a/cmd/juju/application/removeunit.go
+++ b/cmd/juju/application/removeunit.go
@@ -14,6 +14,7 @@ import (
 	"github.com/juju/juju/api/storage"
 	"github.com/juju/juju/cmd/juju/block"
 	"github.com/juju/juju/cmd/modelcmd"
+	"github.com/juju/juju/core/model"
 )
 
 // NewRemoveUnitCommand returns a command which removes an application's units.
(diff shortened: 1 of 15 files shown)


cmd/juju/application/removeunit.go
```diff
@@ -41,56 +52,94 @@ model.
...
 Examples:
 
     juju remove-unit wordpress/2 wordpress/3 wordpress/4
 
+    juju remove-unit wordpress/2 --destroy-storage
+
+For Kubernetes models only a single application can be supplied and only the
+--num-units argument supported.
+Specific units cannot be targeted for removal as that is handled by Kubernetes,
+instead the total number of units to be removed are specified.
```
wallyworld: is specified

cmd/juju/application/removeunit.go
```diff
@@ -144,38 +193,57 @@ func (c *removeUnitCommand) unitsHaveStorage(unitNames []string) (bool, error) {
 // Run connects to the environment specified on the command line and destroys
 // units therein.
 func (c *removeUnitCommand) Run(ctx *cmd.Context) error {
+	logger.Criticalf("here we are cc")
```
wallyworld: so we are

cmd/juju/application/removeunit.go
```diff
@@ -144,38 +193,57 @@ func (c *removeUnitCommand) unitsHaveStorage(unitNames []string) (bool, error) {
...
 	client, apiVersion, err := c.getAPI()
 	if err != nil {
 		return err
 	}
 	defer client.Close()
 
+	logger.Criticalf("here we are 1")
```
wallyworld: but i want to be there, not here

cmd/juju/application/removeunit.go
```diff
@@ -144,38 +193,57 @@ func (c *removeUnitCommand) unitsHaveStorage(unitNames []string) (bool, error) {
...
 	if apiVersion < 4 {
 		return c.removeUnitsDeprecated(ctx, client)
 	}
+
+	logger.Criticalf("here we are 2")
```
wallyworld: ditto etc and below

cmd/juju/application/removeunit.go
```diff
@@ -144,38 +193,57 @@ func (c *removeUnitCommand) unitsHaveStorage(unitNames []string) (bool, error) {
...
+	if err := c.validateRemovalByModelType(); err != nil {
+		return errors.Trace(err)
+	}
+
+	logger.Criticalf("here we are 3")
+	modelType, err := c.ModelType()
```
wallyworld: I think validateRemovalByType() could take model type as an arg since we get it inside the method and also again here.

Veebers:
$$merge$$

PR #9439, opened 2018-11-12
Fixes lp#1802872: model workers changes for 'not found credential', consider invalid.

anastasiamac:
## Description of change

It is possible, under unusual circumstances described in the linked bug, to end up having a model that refers to a cloud credential that no longer exists on the controller - we do not have contents for this credential like pwd, access key, etc [note that the exact type of the content depends on the underlying cloud and is irrelevant to this change].

This has implications for model manifolds and workers - we want to consider that circumstance as invalid.

This PR adjusts existing credential validator worker to treat this situation the same as if the credential is invalid, i.e. stop all cloud communications that depends on a valid credential. The same applies to a model that does not refer to any credentials but has been deployed to a cloud that requires a credential.

As a consequent of this, 'environ-tracker' worker actually, logically, depends on the credential validity flag. So this PR adjust model manifolds and its tests accordingly. Once the 'environ-tracker' became conditional, both instance-poller and a firewaller started panicking. These panics were fixed as well.

Finally, as a drive-by, addressed coding style and warnings that came up as pre-commit checks.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1802872


anastasiamac:
[model credential not found to be treated as a condition to stop cloud calls from the model](https://trello.com/c/BODEyYE0/531-model-credential-not-found-to-be-treated-as-a-condition-to-stop-cloud-calls-from-the-model)

PR #9441, opened 2018-11-12
Seed random number generator during agents init()

timClicks:
This commit calls `rand.Seed()` with the current UTC time as a UNIX timestamp. Code changes affect the  machine, unit, caasoperaor agents as wells as the CLI.

I'm currently unsure what the best strategy for testing and documenting this change should be, so tagging the PR as WIP. 

----

## Description of change

From the [original bug report](https://bugs.launchpad.net/juju/+bug/1802824):

> Juju recently gained the ability to randomise which controller address it connected to. Except that it's not so random because Go requires that the random number generator be seeded first and this is not currently done.

## QA steps

Using the CLI multiple times should result in the addresses that controller selects being different each time.

## Documentation changes

Currently undocumented

## Bug reference

* https://bugs.launchpad.net/juju/+bug/1802824

### DIFF
--- a/cmd/juju/main.go
+++ b/cmd/juju/main.go
@@ -4,13 +4,16 @@
 package main
 
 import (
+	"math/rand"
 	"os"
+	"time"
 
 	"github.com/juju/cmd"
 	"github.com/juju/loggo"
 
 	"github.com/juju/juju/cmd/juju/commands"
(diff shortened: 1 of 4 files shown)


timClicks:
closing (pushing to the wrong branch, also we'll use the init() functions, rather than the Init() methods.)

PR #9442, opened 2018-11-12
Merge 2.4 changes into develop

babbageclunk:
## Description of change

There were two areas with non-trivial conflicts.  In apiserver/facades/client/bundle, the changes for k8s bundles in 2.5 overlapped a bit with 2.4 changes to use machine constraints rather than instance details.

The other was in the engine reports added to state/watcher/hubwatcher.go and txnwatcher.go - these caused some changes in the timer usage so WaitAdvance calls needed tweaking. The new reports required some changes in the worker/state tests as well (since the report now contains a request count which changes with each report request).

## QA steps

* Ran unit tests, bootstrapped a new controller for a basic smoke test.

### DIFF
--- a/apiserver/facades/client/bundle/bundle_test.go
+++ b/apiserver/facades/client/bundle/bundle_test.go
@@ -796,6 +796,13 @@ func (s *bundleSuite) addMinimalMachineWithConstraints(model description.Model, 
 		Architecture: "amd64",
 		Memory:       8 * 1024,
 		RootDisk:     40 * 1024,
+		CpuCores:     2,
+		CpuPower:     100,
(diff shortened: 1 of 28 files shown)


apiserver/facades/client/bundle/bundle_test.go
```diff
@@ -889,9 +898,9 @@ applications:
     - "0"
 machines:
   "0":
-    constraints: arch=amd64 mem=4096 root-disk=16384
+    constraints: arch=amd64 mem=8192 root-disk=40960
```
wallyworld: This whole test should be deleted, ie delete TestExportBundleModelWithHardwareConstraints

PR #9450, opened 2018-11-13
Gosigma - move to a fork instead of patch file

SimonRichardson:
## Description of change

The following commit removes the patch file for gosigma race
issues and instead depends on a fork with the fixes applied. This
way we can make sure that the source of juju is correct, because
the source code of gosigma is applied at source.

## QA steps

 1. Ensure that all tests pass
 1. Run race detector to see that races have disappeared for gosigma.

## Documentation changes

Developers of juju should no longer provide patch files for upstream
changes and instead use `dep` source re-writing to do the same job.

## Bug

https://bugs.launchpad.net/juju/+bug/1795007


Gopkg.toml
```diff
@@ -37,8 +37,9 @@
   name = "github.com/EvilSuperstars/go-cidrman"
 
 [[constraint]]
-  revision = "31228935eec685587914528585da4eb9b073c76d"
+  revision = "a27b59fe2be9d16e2b66f0d7acf5c852f844cc2b"
   name = "github.com/altoros/gosigma"
+  source = "github.com/wallyworld/gosigma"
```
jameinel: Should this be "wallyworld" ? I would have expected us to use github.com/juju/gosigma so that it doesn't require any single person to fix it.
Also, I didn't audit the actual branch to make sure it has the right patch applied, but I'm happy with the concept.

SimonRichardson: I applied Cunningham's Law to this :smile:

SimonRichardson:
https://trello.com/c/At0CwN6k/788-remove-gosigma-patch-files-and-favour-dep-re-writing

SimonRichardson:
!!build!!

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

PR #9454, opened 2018-11-13
Switch from a series of patches to a different branch.

jameinel:
## Description of change

Switch to github.com/juju/mgo which is gopkg.in/mgo.v2 with our patches
applied.

## QA steps

All of the test suite should continue to apply. One manual test that you could do is checkout the existing 2.5 develop and run "make add-patches", and then diff the final content with the content in make dep.

## Documentation changes

None.

## Bug reference

None.


jameinel:
$$merge$$

mitechie:
go not found? 

$$merge$$

PR #9455, opened 2018-11-13
Re-instantiate previously skipped test

SimonRichardson:
## Description of change

The following test was previously skipped whilst I worked out what
the correct flow should have been. The test now falls inline with
the previous intentions.

## QA steps

Running the following should ensure that the test passes correctly.

```
go test -v ./state/... -check.v -check.f=ApplicationSuite.TestSetCharmProfile
```


SimonRichardson:
$$merge$$

PR #9460, opened 2018-11-14
Fix test: httpserver now blocks accept calls when shutting down.

howbazaar:
The test that confirmed that the httpserver stayed running while there were connections, also tried to get a value from the httpserver. However this behaviour changed recently such that when the httpserver is asked to shutdown, it blocks socket accept calls.

This made the test fail intermittently based on which goroutine got woken up first to either deal with the "shutdown now" call or the "accept this connection" call.

The fix is to just not make the http call. It is unnecessary for the test.


howbazaar:
$$merge$$

PR #9463, opened 2018-11-14
Fixes lp#1802872: model workers changes for 'not found credential', consider invalid.

anastasiamac:
## Description of change

This is an adapted forward-port of https://github.com/juju/juju/pull/9439. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1802872


anastasiamac:
self-approving because the code is the same. It's where it is that differs since 2.5 has had a lot of movement in this area.

anastasiamac:
$$merge$$

anastasiamac:
[model credential not found to be treated as a condition to stop cloud calls from the model](https://trello.com/c/BODEyYE0/531-model-credential-not-found-to-be-treated-as-a-condition-to-stop-cloud-calls-from-the-model)

PR #9467, opened 2018-11-15
Fixes lp#1802875: Check models when deleting credential.

anastasiamac:
## Description of change

As per linked bug, Juju was previously allowing to delete a credential from the system even if there were models using it. 

This PR will not allow deletion of a used credential unless it is forced.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1802875

### DIFF
--- a/apiserver/facades/client/cloud/cloud.go
+++ b/apiserver/facades/client/cloud/cloud.go
@@ -8,6 +8,7 @@ package cloud
 import (
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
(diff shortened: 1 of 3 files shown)


anastasiamac:
[change remove credential behavior to verify if it is in use by the models](https://trello.com/c/M8qICYjG/539-change-remove-credential-behavior-to-verify-if-it-is-in-use-by-the-models)

apiserver/facades/client/cloud/cloud.go
```diff
@@ -720,6 +760,33 @@ func (api *CloudAPI) RevokeCredentialsCheckModels(args params.RevokeCredentialAr
...
+			logger.Warningf("could not get models that use credential %v: %v", tag, err)
+		}
+		if len(models) != 0 {
+			// map keys are notoriously randomly ordered
+			uuids := []string{}
+			for uuid := range models {
+				uuids = append(uuids, uuid)
+			}
+			sort.Strings(uuids)
+			for _, uuid := range uuids {
```
wallyworld: This could result in a lot of individual warning messages. It would be better to compose a single multi-line message with the model uuids listed and log a single message, eg
```
WARNING: credential foo cannot be deleted as it is used by these models:
 - uuid
 - uuid
```

anastasiamac:
$$merge$$

PR #9468, opened 2018-11-15
Register the alternative name for lxd

SimonRichardson:
## Description of change

The following registers the alternative name for the localhost
lxd provider. This way we can bootstrap juju just using the
alternative name "lxd" rather than the perceived more popular one
"localhost".

The question of if this is correct or not is a bit of a concern.
Why we don't reconcile both localhost and lxd names and normalise
them to mean the same one when bootstrapping seems a mystery.

Anyway, I'll supply a good deal of Cunningham's Law to this and
see what the fallout to that is.

## QA steps

```sh
juju bootstrap lxd
```

## Documentation changes

This is a regression from 2.4, where you could bootstrap using
"lxd" correctly.

## Bug reference

N/A

### DIFF
--- a/provider/lxd/credentials.go
+++ b/provider/lxd/credentials.go
@@ -111,16 +111,14 @@ func (environProviderCredentials) CredentialSchemas() map[cloud.AuthType]cloud.C
 }
 
 // RegisterCredentials is part of the environs.ProviderCredentialsRegister interface.
-func (p environProviderCredentials) RegisterCredentials(credentialName string) (map[string]*cloud.CloudCredential, error) {
+func (p environProviderCredentials) RegisterCredentials(cld cloud.Cloud) (map[string]*cloud.CloudCredential, error) {
 	// only register credentials if the operator is attempting to access "lxd"
 	// or "localhost"
-	if credentialName != lxdnames.DefaultCloud && credentialName != lxdnames.DefaultCloudAltName {
+	cloudName := cld.Name
+	if cloudName != lxdnames.DefaultCloud && cloudName != lxdnames.DefaultCloudAltName {
 		return make(map[string]*cloud.CloudCredential), nil
 	}
 
-	// register the local lxd credentials, not we register both "localhost" and
-	// "lxd" credentials at the same type, as they can be used interchangeably
-	// through out the use of the commands.
 	nopLogf := func(msg string, args ...interface{}) {}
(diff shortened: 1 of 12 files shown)


SimonRichardson:
https://trello.com/c/fsvnAZBj/798-unable-to-bootstrap-local-lxd-provider-if-using-lxd-as-the-provider-name-instead-of-localhost

SimonRichardson:
Also check if there is no lxd present locally, to allow bootstrapping.

SimonRichardson:
> Also check if there is no lxd present locally, to allow bootstrapping.

Disabling lxd (via a snap) and attempting to bootstrap to a local provider works the same as 2.4, which is to say that it outputs the following https://pastebin.canonical.com/p/7WnvnT92v8/

provider/lxd/credentials.go
```diff
@@ -130,6 +139,12 @@ func (p environProviderCredentials) RegisterCredentials() (map[string]*cloud.Clo
 				lxdnames.DefaultCloud: *localCertCredential,
 			},
 		},
+		lxdnames.DefaultCloudAltName: {
```
jameinel: why register both, rather than just registering the one that matches?

CloudCredentials{
 credentialName:
}

provider/lxd/credentials.go
```diff
@@ -111,7 +111,16 @@ func (environProviderCredentials) CredentialSchemas() map[cloud.AuthType]cloud.C
 }
 
 // RegisterCredentials is part of the environs.ProviderCredentialsRegister interface.
-func (p environProviderCredentials) RegisterCredentials() (map[string]*cloud.CloudCredential, error) {
+func (p environProviderCredentials) RegisterCredentials(credentialName string) (map[string]*cloud.CloudCredential, error) {
```
jameinel: Isn't this "cloudName" not "credentialName" ?
"Register the Credentials for the cloud named X"

But also, why are we restricting it to just the cloud Name, couldn't we put in the full Cloud definition, and then in the future we might do something special based on, say, the Auth URL or something like that?

SimonRichardson: @jameinel that's not a bad idea, I will change that!

provider/lxd/credentials.go
SimonRichardson: Sure, makes sense.

SimonRichardson:
$$merge$$

PR #9472, opened 2018-11-16
More robust controller-api port handling.

howbazaar:
This work does two things:
 * it changes the trigger to open the api-port if the controller-api-port is set
 * adds a report for better observability in the juju_engine_report

Instead of waiting from an event from the peer grouper, the trigger is now an api connection from itself. All api connections are already published, and we know that whenever the httpserver is restarted, the api-caller will have to reconnect as its connection would have been dropped. When we get a published api connection event from ourselves, we initiate the opening of the api-port, after the configured delay of course.

To facilitate this, the agent name is now part of the manifold config. There is no need to get this immutable piece of information out of the agent worker itself.

As a drive by, the current version is added to the agent worker report, as I found myself wanting this information regularly.

## QA steps

* bootstrap lxd
* enable-ha
* `juju controller-config controller-api-port=17071`
* run the juju_engine_report on the controller machines

## Bug reference

https://bugs.launchpad.net/juju/+bug/1803484

### DIFF
--- a/cmd/jujud/agent/machine.go
+++ b/cmd/jujud/agent/machine.go
@@ -459,8 +459,9 @@ func (a *MachineAgent) Run(*cmd.Context) (err error) {
 	}
 
 	agentConfig := a.CurrentConfig()
+	agentName := a.Tag().String()
 	machineLock, err := machinelock.New(machinelock.Config{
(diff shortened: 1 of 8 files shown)


howbazaar:
$$merge$$

PR #9473, opened 2018-11-16
Fix metadata plugins

wallyworld:
## Description of change

The juju metadata plugins actually don't need a model but were written such that a model was mandatory. This was a limitation of the plugin architecture.

Change the plugin architecture to allow plugins to be either controller or model commands, and allow them to run without a mandatory controller/model if appropriate. This allows image metadata to be generated without having to bootstrap which is the common use case.

As part of the work, introduce a JUJU_CONTROLLER env var which operates similarly to JUJU_MODEL.

And as a drive by, delete an obsolete upgrade-mongo plugin (2nd commit).

## QA steps

run juju metadata generate-image without a running controller and check it works.

## Documentation changes

We can document the new JUJU_CONTROLLER env var.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1799250


wallyworld:
[Juju-Openstack provider "juju metadata generate-image" needs a controller to exist](https://trello.com/c/OUWHlBPu/543-juju-openstack-provider-juju-metadata-generate-image-needs-a-controller-to-exist)

wallyworld:
$$merge$$

PR #9478, opened 2018-11-19
vsphere: add enable-disk-uuid model config setting

babbageclunk:
## Description of change
By default VMs created in vsphere don't expose consistent disk UUIDs to the guest OS - this means that k8s (for example) deployed into a vsphere model can't use the vsphere storage features. This can be changed by setting disk.EnableUUID=true on the VM, but we don't want to force that in all situations since in some cases operators might need it to be off: https://kb.vmware.com/s/article/2146204

This PR adds model config option enable-disk-uuid to the vsphere provider - it defaults to true, but can still be set to false if the operator needs to use vsphere features that require it to be off.

Since machines in existing models will have the underlying VM config option set to false (by default), an upgrade step sets enable-disk-uuid=false when upgrading to 2.5 to be consistent.

## QA steps

* Bootstrap a vsphere controller and add an application. The machines created in the controller model and default model should have `disk.EnableUUID=true` in the VM config options in the vsphere console. `enable-disk-uuid` should default to false in the model-config output.
* Bootstrap another controller with `--config="enable-disk-uuid=false" - the machines in that controller should have `disk.EnableUUID=false`.
* Bootstrap a 2.4.x controller in vsphere and upgrade it to a version with this change. The model-config option should be set to false, and new machines created in the model should get `disk.EnableUUID` set to false.

## Documentation changes

* Yes - this new option will need to be documented. A key point to make is that the setting will be applied to new machines in the model, but changing the value won't go and change the VM setting on existing machines - doing this would require stopping and starting the VM, so it needs to be done by the operator.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1751858

### DIFF
--- a/provider/vsphere/config.go
+++ b/provider/vsphere/config.go
@@ -15,6 +15,7 @@ const (
 	cfgPrimaryNetwork  = "primary-network"
 	cfgExternalNetwork = "external-network"
 	cfgDatastore       = "datastore"
+	cfgEnableDiskUUID  = "enable-disk-uuid"
 )
 
 // configFields is the spec for each vmware config value's type.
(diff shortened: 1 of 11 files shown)


babbageclunk:
It looks like the lease manager isn't handling ErrDying properly when retrying. I'll fix that in a separate PR.

$$merge$$

PR #9479, opened 2018-11-19
More permissive test based on counts.

howbazaar:
Backport 2.5 fix. Explicit counts are different on other architectures.


howbazaar:
$$merge$$

PR #9482, opened 2018-11-19
Replace '.' with '-' when creating bridge name to avoid confict with VLANs

wupeka:
## Description of change
Creating bridge br-device.123 for VLANed device.123 can confuse network scripts, avoid it by replacing '.' with '-' in generated bridge name.

## QA steps
Described in bug report

## Documentation changes
None

## Bug reference
https://bugs.launchpad.net/juju/+bug/1804018

### DIFF
--- a/network/containerizer/bridgepolicy.go
+++ b/network/containerizer/bridgepolicy.go
@@ -180,14 +180,16 @@ var skippedDeviceNames = set.NewStrings(
 // 2. Add b- to device name, if it doesn't fit in 15 characters then:
 // 3a. For devices starting in 'en' remove 'en' and add 'b-'
 // 3b. For all other devices 'b-' + 6-char hash of name + '-' + last 6 chars of name
+// 4. If using the device name directly always replace '.' with '-', to make sure that bridges from VLANs won't break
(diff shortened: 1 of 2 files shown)


mitechie:
!!build!!

wupeka:
!!build!!

mitechie:
!!build!!

jameinel:
so
a) If it wasn't working before, we don't have to worry about backward compatibility :)
b) The general networking logic *should* be "do we have a bridge in space X, if so, use it, else create it". (to support the case where you define the bridge in MAAS before Juju even gets to see the machine. So for sites if they *did* have it working, we should continue to use the existing one.

I think that part is fine.
Obviously we need to fix the test suite for this.

jameinel:
@wupeka its good to see you around, btw. Thanks for the patch.

jameinel:
It seems odd to me that the bridge name would declare whether it was a VLAN or not, but I guess that is already true for the ETH device. So LGTM, we'll need to sort out any extra test issues.

jameinel:
This has landed in 2.4 and develop from the above PR.

PR #9484, opened 2018-11-20
Remove "oracle" and "oci" clouds from juju add-cloud's interactive prompt

timClicks:
This commit removes the "oracle" cloud from `juju add-cloud`'s list of clouds supported by the interactive setup.

----

## QA steps

Run `juju add-cloud`

Expect to see:

```
Cloud Types
 lxd
 maas
 manual
 openstack
 vsphere

Select cloud type:
````

## Documentation changes

Change to Juju CLI: Running `juju add-cloud` no longer presents `oracle` as an option.

Rationale: Oracle Cloud Infrastructure (OCI) is now the favored cloud for Juju developers. Oracle's public cloud offering has been renamed "Oracle Classic" in their documentation. Accordingly, within Juju, the "oracle" cloud has been deprecated in favor of the "oci" cloud provider. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1802540

### DIFF
--- a/cmd/juju/cloud/add.go
+++ b/cmd/juju/cloud/add.go
@@ -33,26 +33,63 @@ type CloudMetadataStore interface {
 }
 
 var usageAddCloudSummary = `
-Adds a user-defined cloud to Juju from among known cloud types.`[1:]
+Adds a cloud definition to Juju.
+`[1:]
 
 var usageAddCloudDetails = `
+Juju needs to know how to connect to clouds. A cloud definition 
+describes a cloud's endpoints and authentication requirements. Each
+definition is stored and accessed later as <cloud name>.
+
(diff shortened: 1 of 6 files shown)


timClicks:
[Traces of Oracle should be removed from interactive mode](https://trello.com/c/Lo8axZuQ/529-traces-of-oracle-should-be-removed-from-interactive-mode)

mitechie:
The goal is that there is no oracle-classic support and that there's a
single "oracle" that is the new OCI work.

This was ok'd by looking at model stats and seeing that the only models out
there on the old oracle classic were our own testing ones and to have the
cleanest path forward for Juju users going forward.

timClicks:
$$merge$$

PR #9486, opened 2018-11-20
Include an mgo patch to filter txn-queue during Insert and Remove

jameinel:
## Description of change

This is the patch from
  https://github.com/juju/mgo/pull/4
  https://github.com/globalsign/mgo/pull/301

It filters the txn-queue when copying a document from or to the stash,
just like we would do during a normal Update operation. This prevents a
sequence of Insert + Remove pairs from filling up the txn-queue.

## QA steps

The associated test shows the basic change. To trigger this manually you can:

```
$ juju bootstrap lxd
$ while true; do juju deploy cs:~jameinel/ubuntu-lite; sleep 90; juju remove-application ubuntu-lite; sleep 30 done
```

While doing that, you should be able to go into the juju controller and either
```
> db.constraints.find()
> db.txns.stash.find({"_id.c": "constraints"})
```
And see that there is a document whose txn-queue is growing quite large. With this patch, the txn-queue should stay around 1 or 2 entries.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1804197


jameinel:
$$merge$$

PR #9494, opened 2018-11-22
Use extended wwn if available for block devices

wallyworld:
## Description of change

For logical block devices, it's possible they are identified by an extended wwn; the "normal" wwn itself identifies the controller. Add support for recording the extended wwn if present as that's how the volumes will appear under /dev/disk/by-id and is expected to be what is reported by the provider, eg MAAS.

Also add trace level debugging for troubleshooting.

## QA steps

Will be tested on site.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1778033

### DIFF
--- a/worker/diskmanager/lsblk.go
+++ b/worker/diskmanager/lsblk.go
@@ -188,7 +188,7 @@ func addHardwareInfo(dev *storage.BlockDevice) error {
 		return errors.Annotate(err, msg)
 	}
 
-	var devpath, idBus, idSerial string
+	var devpath, idBus, idSerial, wwnWithExtension string
 
 	s := bufio.NewScanner(bytes.NewReader(output))
 	for s.Scan() {
@@ -210,6 +210,8 @@ func addHardwareInfo(dev *storage.BlockDevice) error {
 			idSerial = value
 		case "ID_WWN":
 			dev.WWN = value
+		case "ID_WWN_WITH_EXTENSION":
+			wwnWithExtension = value
 		default:
 			logger.Tracef("ignoring line: %q", line)
 		}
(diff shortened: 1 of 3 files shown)


worker/diskmanager/lsblk.go
```diff
@@ -218,6 +220,14 @@ func addHardwareInfo(dev *storage.BlockDevice) error {
 		return errors.Annotate(err, "cannot parse udevadm output")
 	}
 
+	// For cases where there are logical disks attached to a
+	// controller (eg RAID), the controller itself has a WWN
+	// which is the ID_WWN value and each disk has a WWN with
+	// a vendor extension ID_WWN_WITH_EXTENSION added to
+	// identify the disk via the /dev/disk/by-id path.
+	if wwnWithExtension != "" {
```
babbageclunk: Just checking I understand - the reason for doing this here rather than in the ID_WWN_WITH_EXTENSION case is because that and ID_WWN might be in the output in either order, and we always want the extended version to take precedence if it's present. Is that right?

worker/diskmanager/lsblk_test.go
```diff
@@ -83,6 +83,15 @@ ID_WWN=foo
 `, storage.BlockDevice{WWN: "foo"})
 }
 
+func (s *ListBlockDevicesSuite) TestListBlockDevicesExtendedWWN(c *gc.C) {
+	// If ID_WWN_WITH_EXTENSION is found, then we should use that
+	// in preference to the ID_WWN value.
+	s.testListBlockDevicesExtended(c, `
+ID_WWN=foo
+ID_WWN_WITH_EXTENSION=foobar
```
babbageclunk: Might be worth flipping the order of these or adding another test with them in reverse to ensure the with_extension takes priority.

worker/diskmanager/lsblk.go
wallyworld: edzakery

worker/diskmanager/lsblk_test.go
wallyworld: indeed

wallyworld:
$$merge$$

PR #9497, opened 2018-11-23
Allow the disabling of state prometheus metrics

howbazaar:
Adds a controller feature flag that if "disable-state-metrics" is set, the prometheus gatherer isn't registered.

This is primarily a short term fix to allow more robust gathering of data from large controllers.


wallyworld:
$$merge$$

PR #9498, opened 2018-11-23
uniter: Be smarter about when we run config-changed (part 1)

babbageclunk:
## Description of change
At the moment, when a controller agent restarts all of the unit agents that were connected to it reconnect and then immediately fire config-changed, even if no configuration has actually changed. On controllers with thousands of units, this can overload the controller agent while all the units get all their config and run any other hook commands in response.

The uniter always runs config-changed when it first starts because it doesn't track any local state that would enable it to see that the config is the same and avoid running the hook.

Change the API to expose Uniter.WatchConfigSettingsHash - this works like .WatchConfigSettings, except it sends a hash of the current config when it signals. The uniter stores that value in the local uniter state file, so that it persists across agent restarts, and checks it before firing config-changed.

This bumps the uniter API to version 9.

This PR plumbs the config hash through the API and uniter, but the uniter also watches trust config and addresses, so the behaviour won't change until those other parts are done in followups.

## QA steps

* Bootstrap a controller and model, deploy some applications and relate them. 
* See that the uniter state files for various units includes the config-hash.
* Change a config value for one of the applications.
* See that the config-hash changes when the config-changed hook fires.

## Documentation changes

None

### DIFF
--- a/apiserver/facades/agent/uniter/uniter.go
+++ b/apiserver/facades/agent/uniter/uniter.go
@@ -35,7 +35,9 @@ import (
 
 var logger = loggo.GetLogger("juju.apiserver.uniter")
 
-// UniterAPI implements the latest version (v8) of the Uniter API.
+// UniterAPI implements the latest version (v9) of the Uniter API,
+// which changes WatchConfigSettings to a StringsWatcher that returns
+// a hash of the current config.
 type UniterAPI struct {
 	*common.LifeGetter
 	*StatusAPI
@@ -64,9 +66,16 @@ type UniterAPI struct {
 	cloudSpec       cloudspec.CloudSpecAPI
 }
 
+// UniterAPIV8 adds SetContainerSpec, GoalStates, CloudSpec,
+// WatchTrustConfigSettings, WatchActionNotifications,
+// UpgradeSeriesStatus, SetUpgradeSeriesStatus.
+type UniterAPIV8 struct {
+	UniterAPI
+}
+
 // UniterAPIV7 adds CMR support to NetworkInfo.
 type UniterAPIV7 struct {
-	UniterAPI
+	UniterAPIV8
 }
 
 // UniterAPIV6 adds NetworkInfo as a preferred method to calling NetworkConfig.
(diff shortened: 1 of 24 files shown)


babbageclunk:
Note to self - work out whether/how the uniter restarts the WatchConfigSettings(Hash) watcher when the application's charm is updated. If it doesn't it'll need to!

worker/uniter/remotestate/watcher_test.go
```diff
@@ -188,9 +188,9 @@ func (s *WatcherSuiteIAAS) TestSnapshot(c *gc.C) {
 	s.signalAll()
 	assertNotifyEvent(c, s.watcher.RemoteStateChanged(), "waiting for remote state change")
 
-	// Note that the configuration change is updated on both application and
-	// charm config which increments it twice.
-	expectedVersion := 3
+	// Note that the configuration change is updated on both application config and
```
wallyworld: I think this should be " on both trust changes and address changes "
The uniter watches for trust changes, and on the server side it's done in terms of app config changes.

worker/uniter/remotestate/watcher_test.go
```diff
@@ -211,9 +212,9 @@ func (s *WatcherSuiteCAAS) TestSnapshot(c *gc.C) {
...
```
wallyworld: ditto

worker/uniter/resolver_test.go
```diff
@@ -368,3 +368,57 @@ func (s *resolverSuite) TestRunHookStopRetryTimer(c *gc.C) {
 	c.Assert(err, gc.Equals, resolver.ErrNoOperation)
 	s.stub.CheckCallNames(c, "StartRetryHookTimer", "StopRetryHookTimer")
 }
+
+func (s *resolverSuite) TestRunsConfigChangedIfConfigHashChanges(c *gc.C) {
```
wallyworld: A test for no op if hashes haven't changed?

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -2673,3 +2693,59 @@ func (u *UniterAPI) goalStateUnits(app *state.Application, principalName string)
...
+	return result, nil
+}
+
+// Mask WatchConfigSettingsHash from the v8 API. The API reflection
+// code in rpc/rpcreflect/type.go:newMethod skips 2-argument methods,
+// so this removes the method as far as the RPC machinery is
+// concerned.
+
+// WatchConfigSettingsHash isn't on the v8 API.
+func (u *UniterAPIV8) WatchConfigSettingsHash(_, _ struct{}) {}
```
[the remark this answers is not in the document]
babbageclunk: I'd like to do that in a separate PR - once I go through and make sure nothing else uses it.

state/unit_test.go
```diff
@@ -165,6 +165,118 @@ func (s *UnitSuite) TestWatchConfigSettings(c *gc.C) {
...
+	// Add an attribute that comes before.
+	err = s.application.UpdateCharmConfig(charm.Settings{
+		"alphabetic": 1,
+	})
+	c.Assert(err, jc.ErrorIsNil)
+	// Sanity check.
+	config, err := s.unit.ConfigSettings()
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(config, gc.DeepEquals, charm.Settings{
+		// Ints that get round-tripped through mongo come back as int64.
```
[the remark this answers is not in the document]
babbageclunk: Good call, I'll add a test for that.

worker/uniter/remotestate/watcher_test.go
babbageclunk: Right, that makes sense.

worker/uniter/resolver_test.go
babbageclunk: Ok, I'll add that.

babbageclunk:
I can confirm that the hash keeps getting updated if I change a config value after a charm upgrade, so the uniter must be recreating the config settings hash watcher. I'll make sure I check the same thing happens when the other two hashes are done.

babbageclunk:
$$merge$$

PR #9502, opened 2018-11-27
Merge 2.4 develop incl azure

Veebers:
## Description of change

Merges latest 2.4 changes into develop including:

  - Azure SDK updates
  - New engine report

### DIFF
--- a/provider/azure/environ.go
+++ b/provider/azure/environ.go
@@ -4,6 +4,7 @@
 package azure
 
 import (
+	stdcontext "context"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -12,11 +13,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/Azure/azure-sdk-for-go/arm/compute"
-	"github.com/Azure/azure-sdk-for-go/arm/disk"
-	"github.com/Azure/azure-sdk-for-go/arm/network"
-	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
-	"github.com/Azure/azure-sdk-for-go/arm/storage"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-08-01/network"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2018-07-01/storage"
 	azurestorage "github.com/Azure/azure-sdk-for-go/storage"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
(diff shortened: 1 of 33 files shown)


provider/azure/environ.go
```diff
@@ -1243,17 +1260,26 @@ func (env *azureEnviron) deleteVirtualMachine(
...
 		if errorutils.MaybeInvalidateCredential(err, ctx) {
 			return invalidatedCredential(err, msg)
 		}
-		if !isNotFoundResponse(result.Response) {
+		if !isNotFoundResponse(autorest.Response{vmFuture.Response()}) {
 			return errors.Annotate(err, msg)
 		}
 	}
-
+	if err == nil {
```
wallyworld: this can just be else {

provider/azure/environ.go
```diff
@@ -1243,17 +1260,26 @@ func (env *azureEnviron) deleteVirtualMachine(
...
+		err = vmFuture.WaitForCompletionRef(sdkCtx, vmClient.Client)
```
wallyworld: can't all these future api calls also error due to an invalid cred?

provider/azure/environ.go
```diff
@@ -1292,46 +1327,78 @@ func (env *azureEnviron) deleteVirtualMachine(
...
-		resultCh, errCh := nicClient.Delete(env.resourceGroup, nicName, nil)
-		if result, err := <-resultCh, <-errCh; err != nil {
+		nicFuture, err := nicClient.Delete(sdkCtx, env.resourceGroup, nicName)
+		if err != nil {
 			msg := "deleting NIC"
 			if errorutils.MaybeInvalidateCredential(err, ctx) {
 				return invalidatedCredential(err, msg)
 			}
-			if !isNotFoundResponse(result) {
+			if !isNotFoundResponse(autorest.Response{nicFuture.Response()}) {
```
wallyworld: I wonder we need both a isNotFoundResponse to take a http.Response and a isNotFoundResult to take the autorest.Response from the Result() call and have the isNotFoundResponse wrap and call isNotFoundResult.
Avoids all this struct boilerplate

provider/azure/environ.go
```diff
@@ -1243,17 +1260,33 @@ func (env *azureEnviron) deleteVirtualMachine(
...
 	logger.Debugf("- deleting virtual machine (%s)", vmName)
-	vmResultCh, errCh := vmClient.Delete(env.resourceGroup, vmName, nil)
-	if result, err := <-vmResultCh, <-errCh; err != nil {
-		msg := "deleting virtual machine"
+	vmErrMsg := "deleting virtual machine"
+	vmFuture, err := vmClient.Delete(sdkCtx, env.resourceGroup, vmName)
+	if err != nil {
 		if errorutils.MaybeInvalidateCredential(err, ctx) {
-			return invalidatedCredential(err, msg)
+			return invalidatedCredential(err, vmErrMsg)
```
wallyworld: This invalidatedCredential() wrapper func makes no sense - it just annotates the error with a message? Why do we need it instead of just calling errors.Annotate() directly.
Existing code, but....

provider/azure/environ.go
```diff
@@ -1243,17 +1260,33 @@ func (env *azureEnviron) deleteVirtualMachine(
...
 		}
-		if !isNotFoundResponse(result.Response) {
-			return errors.Annotate(err, msg)
+		if !isNotFoundResponse(vmFuture.Response()) {
+			return errors.Annotate(err, vmErrMsg)
+		}
+	} else {
+		err = vmFuture.WaitForCompletionRef(sdkCtx, vmClient.Client)
+		if err != nil {
+			if errorutils.MaybeInvalidateCredential(err, ctx) {
```
wallyworld: So here is where it really doesn't make sense.
You just need
return errorutils.HandleCredentialError(errors.Annotate(err, "deleting virtual machine"), ctx)

Veebers:
$$merge$$

PR #9504, opened 2018-11-27
oci: add authorisation checking mechanism to API responses

timClicks:
We make use of the juju [`HandleCredentialError`][1] callback mechanism to mark credentials as invalid within Juju when that's detected client-side. 

All API calls that require a [`CallProviderContext`][2] have had this callback added.

[1]: https://github.com/juju/juju/provider/common
[2]: https://github.com/juju/juju/environs/context

## Description of change

This change enables Juju to short circuit subsequent requests to the API with credentials that are no longer valid. This reduces the load on the server by preventing workers (for example) to continually poll the API with requests. The short circuit  behaviour makes OCI consistent with the other providers.  

## QA steps

 * bootstrap into oci
 * deploy a workload
 * invalidate the user's credentials (via the oracle console?)
 * attempt to perform another action within juju

This should fail fast and further.

## Documentation changes

None. This change shouldn't be user-facing, except that authorisation errors are now handled more gracefully. 

## Bug reference

n/a

### DIFF
--- a/provider/oci/common/errors.go
+++ b/provider/oci/common/errors.go
@@ -1,0 +1,61 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details
+
+package common
+
+import (
+	"github.com/juju/collections/set"
+	ociCommon "github.com/oracle/oci-go-sdk/common"
+
+	"github.com/juju/juju/environs/context"
+	"github.com/juju/juju/provider/common"
+)
+
+// Oracle bundles authorisation errors into HTTP 401, HTTP 404 and
+// HTTP 409, without specifically indicating that it's an authorisation
+// failure. Each response includes a Code field that we can match against,
+// but they remain intentionally ambiguous.
+//
+// The codes we match against are:
+//   - HTTP 401 ("NotAuthenticated")
+//   - HTTP 404 ("NotAuthorizedOrNotFound")
+//   - HTTP 409 ("NotAuthorizedOrResourceAlreadyExists")
+//
+// As we're not generating any API calls manually, it's unlikely
+// that we'll be striking URIs that don't exist at all, therefore we assume
+// auth issues are causing the errors.
+//
+// For more details, see https://docs.cloud.oracle.com/iaas/Content/API/References/apierrors.htm
+var authErrorCodes = set.NewStrings(
+	"NotAuthenticated",
+	"NotAuthorizedOrResourceAlreadyExists",
+	"NotAuthorizedOrNotFound",
+)
+
+// IsAuthorisationFailure reports whether the error is related to
+// attempting to access the provider with invalid or expired credentials.
+func IsAuthorisationFailure(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	serviceError, ok := err.(ociCommon.ServiceError)
+	if !ok {
+		// Just to double check, also try the SDK's
+		// implementation. This isn't checked first, because
+		// it is hard to test.
+		serviceError, ok = ociCommon.IsServiceError(err)
+	}
+
+	if ok && authErrorCodes.Contains(serviceError.GetCode()) {
+		return true
+	}
+
+	return false
+}
+
+// HandleCredentialError marks the current credentials as invalid internally
+//  if Oracle believes that they are expired
+func HandleCredentialError(err error, ctx context.ProviderCallContext) {
+	common.HandleCredentialError(IsAuthorisationFailure, err, ctx)
+}
--- a/provider/oci/common/errors_test.go
+++ b/provider/oci/common/errors_test.go
@@ -1,0 +1,60 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package common_test
+
+import (
+	"fmt"
+
+	ocicommon "github.com/oracle/oci-go-sdk/common"
+	gc "gopkg.in/check.v1"
+
+	"github.com/juju/errors"
+	"github.com/juju/juju/provider/oci/common"
+	"github.com/juju/juju/testing"
+	jc "github.com/juju/testing/checkers"
+)
+
+type errorsSuite struct {
+	testing.BaseSuite
+}
+
+var _ = gc.Suite(&errorsSuite{})
+
+type MockServiceError struct {
+	ocicommon.ServiceError
+
+	code string
+}
+
+func (a MockServiceError) Error() string {
+	return fmt.Sprintf("Mocked error %s", a.GetCode())
+}
+
+func (a MockServiceError) GetCode() string { return a.code }
+
+func (s *errorsSuite) TestServiceErrorsCanTriggerIsAuthorisationFailure(c *gc.C) {
+	err := MockServiceError{code: "NotAuthenticated"}
+	result := common.IsAuthorisationFailure(err)
+	c.Assert(result, jc.IsTrue)
+
+	err = MockServiceError{code: "InternalServerError"}
+	result = common.IsAuthorisationFailure(err)
+	c.Assert(result, jc.IsFalse)
+}
+
+func (s *errorsSuite) TestUnknownErrorsDoNotTriggerIsAuthorisationFailure(c *gc.C) {
+	err1 := errors.New("unknown")
+	for _, err := range []error{
+		err1,
+		errors.Trace(err1),
+		errors.Annotate(err1, "really unknown"),
+	} {
+		c.Assert(common.IsAuthorisationFailure(err), jc.IsFalse)
+	}
+}
+
+func (s *errorsSuite) TestNilDoesNotTriggerIsAuthorisationFailure(c *gc.C) {
+	result := common.IsAuthorisationFailure(nil)
+	c.Assert(result, jc.IsFalse)
+}
--- a/provider/oci/environ.go
+++ b/provider/oci/environ.go
@@ -72,13 +72,14 @@ func (e *Environ) ecfg() *environConfig {
 	return e.ecfgObj
 }
 
-func (e *Environ) allInstances(tags map[string]string) ([]*ociInstance, error) {
+func (e *Environ) allInstances(ctx envcontext.ProviderCallContext, tags map[string]string) ([]*ociInstance, error) {
 	compartment := e.ecfg().compartmentID()
 	request := ociCore.ListInstancesRequest{
 		CompartmentId: compartment,
 	}
 	response, err := e.Compute.ListInstances(context.Background(), request)
 	if err != nil {
+		providerCommon.HandleCredentialError(err, ctx)
 		return nil, errors.Trace(err)
 	}
 
@@ -101,6 +102,7 @@ func (e *Environ) allInstances(tags map[string]string) ([]*ociInstance, error) {
 		}
 		inst, err := newInstance(val, e)
 		if err != nil {
+			providerCommon.HandleCredentialError(err, ctx)
 			return nil, errors.Trace(err)
 		}
 		ret = append(ret, inst)
@@ -108,7 +110,7 @@ func (e *Environ) allInstances(tags map[string]string) ([]*ociInstance, error) {
 	return ret, nil
 }
 
-func (e *Environ) getOCIInstance(id instance.Id) (*ociInstance, error) {
+func (e *Environ) getOCIInstance(ctx envcontext.ProviderCallContext, id instance.Id) (*ociInstance, error) {
 	instanceId := string(id)
 	request := ociCore.GetInstanceRequest{
 		InstanceId: &instanceId,
@@ -116,6 +118,7 @@ func (e *Environ) getOCIInstance(id instance.Id) (*ociInstance, error) {
 
 	response, err := e.Compute.GetInstance(context.Background(), request)
 	if err != nil {
+		providerCommon.HandleCredentialError(err, ctx)
 		return nil, errors.Trace(err)
 	}
 
@@ -183,6 +186,7 @@ func (e *Environ) AvailabilityZones(ctx envcontext.ProviderCallContext) ([]commo
 	domains, err := e.Identity.ListAvailabilityDomains(ociCtx, request)
 
 	if err != nil {
+		providerCommon.HandleCredentialError(err, ctx)
 		return nil, errors.Trace(err)
 	}
 
@@ -194,10 +198,11 @@ func (e *Environ) AvailabilityZones(ctx envcontext.ProviderCallContext) ([]commo
 	return zones, nil
 }
 
-// InstanceAvailabilityzoneNames implements common.ZonedEnviron.
+// InstanceAvailabilityZoneNames implements common.ZonedEnviron.
 func (e *Environ) InstanceAvailabilityZoneNames(ctx envcontext.ProviderCallContext, ids []instance.Id) ([]string, error) {
 	instances, err := e.Instances(ctx, ids)
 	if err != nil && err != environs.ErrPartialInstances {
+		providerCommon.HandleCredentialError(err, ctx)
 		return nil, err
 	}
 	zones := []string{}
@@ -216,7 +221,7 @@ func (e *Environ) DeriveAvailabilityZones(ctx envcontext.ProviderCallContext, ar
 	return nil, nil
 }
 
-func (e *Environ) getOciInstances(ids ...instance.Id) ([]*ociInstance, error) {
+func (e *Environ) getOciInstances(ctx envcontext.ProviderCallContext, ids ...instance.Id) ([]*ociInstance, error) {
 	ret := []*ociInstance{}
 
 	compartmentID := e.ecfg().compartmentID()
@@ -226,6 +231,7 @@ func (e *Environ) getOciInstances(ids ...instance.Id) ([]*ociInstance, error) {
 
 	instances, err := e.Compute.ListInstances(context.Background(), request)
 	if err != nil {
+		providerCommon.HandleCredentialError(err, ctx)
 		return nil, errors.Trace(err)
 	}
 
@@ -236,6 +242,7 @@ func (e *Environ) getOciInstances(ids ...instance.Id) ([]*ociInstance, error) {
 	for _, val := range instances.Items {
 		oInstance, err := newInstance(val, e)
 		if err != nil {
+			providerCommon.HandleCredentialError(err, ctx)
 			return nil, errors.Trace(err)
 		}
 		for _, id := range ids {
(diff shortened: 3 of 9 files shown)


timClicks:
[adjust oci provider to use callback to invalidate cloud credential when needed](https://trello.com/c/xmpnE5cd/368-adjust-oci-provider-to-use-callback-to-invalidate-cloud-credential-when-needed)

provider/oci/common/errors.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details
+
+package common
+
+import (
+	"github.com/juju/collections/set"
+	ociCommon "github.com/oracle/oci-go-sdk/common"
+
+	envContext "github.com/juju/juju/environs/context"
```
wallyworld: Only need to alias the import if there's a name clash

provider/oci/common/errors.go
```diff
@@ -0,0 +1,64 @@
...
+	return false
+}
+
+// HandleCredentialError marks the current credentials as invalid internally
+//  if Oracle believes that they are expired
+func HandleCredentialError(err error, ctx envContext.ProviderCallContext) {
+	common.HandleCredentialError(IsAuthorisationFailure, err, ctx)
+}
+
+func init() {
```
wallyworld: No need for init, just assign to the var directly.

```
var authErrorCodes = set.NewStrings(
    "NotAuthenticated", 
    "NotAuthorizedOrResourceAlreadyExists",
    "NotAuthorizedOrNotFound",
)
```

provider/oci/common/errors_test.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package common_test
+
+import (
+	"fmt"
+
+	"github.com/juju/errors"
+	. "github.com/juju/testing/checkers"
```
wallyworld: By convention we alias this to "jc"

provider/oci/common/errors_test.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (s *errorsSuite) TestServiceErrorsCanTriggerIsAuthorisationFailure(c *gc.C) {
+	err := MockAuthError{}
+	result := oci.IsAuthorisationFailure(err)
+	c.Assert(result, IsTrue)
+}
+
+func (s *errorsSuite) TestUnknownErrorsDoNotTriggerIsAuthorisationFailure(c *gc.C) {
+	err1 := errors.New("unknown")
```
wallyworld: Don't really need to Cause() case

```
	err1 := errors.New("unknown")
 	for _, err := range []error{
            err1,
            errors.Trace(err1),
            errors.Annotate(err1, "really unknown"),
        } {
            c.Assert(oci.IsAuthorisationFailure(err), IsFalse)
	}
```

provider/oci/common/errors_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (a MockAuthError) GetHTTPStatusCode() int { return 401 }
+
+func (a MockAuthError) GetMessage() string { return "" }
+
+func (a MockAuthError) GetOpcRequestID() string { return "123abc" }
+
+func (s *errorsSuite) TestOCIServiceErrorInterfaceCorrectlyMocked(c *gc.C) {
+	// compile-time check
+	var _ ociCommon.ServiceError = (*MockAuthError)(nil)
```
wallyworld: Don't need this is a test, just declare this as a package var

provider/oci/common/errors_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/juju/errors"
+	. "github.com/juju/testing/checkers"
+
+	ociCommon "github.com/oracle/oci-go-sdk/common"
+	gc "gopkg.in/check.v1"
+
+	"github.com/juju/juju/provider/oci"
+)
+
+type errorsSuite struct {
```
wallyworld: I can't see where the suite is registered. What was the output when you ran the tests? I bet it said 0 tests run :-)

provider/oci/common/errors_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (a MockAuthError) GetOpcRequestID() string { return "123abc" }
+
+func (s *errorsSuite) TestOCIServiceErrorInterfaceCorrectlyMocked(c *gc.C) {
+	// compile-time check
+	var _ ociCommon.ServiceError = (*MockAuthError)(nil)
+	c.Assert(true, IsTrue)
+}
+
+func (s *errorsSuite) TestServiceErrorsCanTriggerIsAuthorisationFailure(c *gc.C) {
```
wallyworld: This test will never pass as MockAuthError is not a serviceError

provider/oci/common/errors_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+package common_test
+
+import (
+	"fmt"
+
+	"github.com/juju/errors"
+	. "github.com/juju/testing/checkers"
+
+	ociCommon "github.com/oracle/oci-go-sdk/common"
```
wallyworld: ocicommon

provider/oci/common/errors_test.go
```diff
@@ -0,0 +1,62 @@
...
+	gc "gopkg.in/check.v1"
+
+	"github.com/juju/juju/provider/oci"
+)
+
+type errorsSuite struct {
+	commonSuite
+}
+
+type MockAuthError struct{}
```
wallyworld: You want this

```
type mockAuthError struct {
    ocicommon.ServiceError
}

func (m mockAuthError) GetCode() string { return "NotAuthenticated" }
```

The other functions are unnecesary

provider/oci/common/errors.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// IsAuthorisationFailure reports whether the error is related to
+// attempting to access the provider with invalid or expired credentials.
+func IsAuthorisationFailure(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	serviceError, ok := ociCommon.IsServiceError(err)
```
wallyworld: serviceError, ok := err.(ocicommon.ServiceError)

provider/oci/common/errors.go
```diff
@@ -0,0 +1,64 @@
...
+	if authErrorCodes.Contains(serviceError.GetCode()) {
+		return true
+	}
+
+	return false
+}
+
+// HandleCredentialError marks the current credentials as invalid internally
+//  if Oracle believes that they are expired
+func HandleCredentialError(err error, ctx envContext.ProviderCallContext) {
```
wallyworld: this needs to return an error

provider/oci/common/errors.go
```diff
@@ -0,0 +1,64 @@
...
+	common.HandleCredentialError(IsAuthorisationFailure, err, ctx)
```
wallyworld: follow up with 
return err

provider/oci/environ.go
```diff
@@ -183,6 +183,7 @@ func (e *Environ) AvailabilityZones(ctx envcontext.ProviderCallContext) ([]commo
 	domains, err := e.Identity.ListAvailabilityDomains(ociCtx, request)
 
 	if err != nil {
+		providerCommon.HandleCredentialError(err, ctx)
```
wallyworld: return nil, providerCommon.HandleCredentialError(errors.Trace(err), ctx)

etc

provider/oci/environ.go
```diff
@@ -226,6 +228,8 @@ func (e *Environ) getOciInstances(ids ...instance.Id) ([]*ociInstance, error) {
 
 	instances, err := e.Compute.ListInstances(context.Background(), request)
 	if err != nil {
+		// todo(tsm) add ctx as arg
```
wallyworld: Why not add the missing ctx arg :-)
It's needed.

provider/oci/environ.go
```diff
@@ -393,10 +406,16 @@ func (e *Environ) DestroyController(ctx envcontext.ProviderCallContext, controll
 
 	err = e.StopInstances(ctx, ids...)
 	if err != nil {
+		providerCommon.HandleCredentialError(err, ctx)
 		return err
 	}
 	logger.Debugf("Cleaning up network resources")
-	return e.cleanupNetworksAndSubnets(controllerUUID, "")
+	err = e.cleanupNetworksAndSubnets(controllerUUID, "")
+	if err != nil {
```
wallyworld: No need for nil check, handle that in the helper func if necessary

return providerCommon.HandleCredentialError(err, ctx)

provider/oci/instance.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/juju/juju/environs/instances"
 	"github.com/juju/juju/instance"
 	"github.com/juju/juju/network"
+	providerCommon "github.com/juju/juju/provider/oci/common"
```
wallyworld: No need for alias

provider/oci/common/errors.go
timClicks: Looks like I made the wrong call here. I went for consistency through the package.

provider/oci/common/errors.go
timClicks: This is the SDK code's approach, but I like the type conversion idea.

provider/oci/environ.go
timClicks: Was slightly scared of changing method signatures - wanted to confirm that's what you wanted to do :)

provider/oci/instance.go
timClicks: Yeah this was just for consistency, will fix.

timClicks:
@wallyworld PR ready to be re-reviewed

provider/oci/environ.go
```diff
@@ -393,10 +403,12 @@ func (e *Environ) DestroyController(ctx envcontext.ProviderCallContext, controll
...
 	err = e.StopInstances(ctx, ids...)
 	if err != nil {
-		return err
+		err = providerCommon.HandleCredentialError(err, ctx)
+		return errors.Trace(err)
 	}
 	logger.Debugf("Cleaning up network resources")
-	return e.cleanupNetworksAndSubnets(controllerUUID, "")
+	err = e.cleanupNetworksAndSubnets(controllerUUID, "")
+	return providerCommon.HandleCredentialError(err, ctx)
```
wallyworld: This spot does it "correctly", in accordance with that the providerCommon.HandleCredentialError() method signature offers.

provider/oci/environ.go
```diff
@@ -649,12 +677,16 @@ func (o *Environ) terminateInstances(instances ...*ociInstance) error {
 			defer wg.Done()
 			if err := inst.deleteInstance(); err != nil {
 				errCh <- instError{id: inst.Id(), err: err}
+				// todo(tsm) add ctx as argument
```
wallyworld: remove these comments
you need to do the error processing down below as you iterate the errCh

provider/oci/common/errors.go
```diff
@@ -0,0 +1,62 @@
...
+	if err == nil {
+		return false
+	}
+
+	serviceError, ok := err.(ociCommon.ServiceError)
+	if !ok {
+		// Just to double check, also try the SDK's
+		// implementation. This isn't checked first, because
+		// it is hard to test.
+		serviceError, ok = ociCommon.IsServiceError(err)
```
babbageclunk: I can't see how there would be circumstances where this would be true but the cast above would fail - doesn't the internal `serviceError` type implement `ociCommon.ServiceError`?

provider/oci/storage_volumes.go
```diff
@@ -534,6 +543,7 @@ func (v *volumeSource) DetachVolumes(ctx envcontext.ProviderCallContext, params
 		if !ok {
 			currentAttachments, err := v.volumeAttachments(param.InstanceId)
 			if err != nil {
```
wallyworld: If there's a credential error here, there's no point continuing with the next loop below.
A call to MaybeHandleCredentialError will tell you if that's the case
(same with other loops in the storage area potentially, look at other providers for inspiration)

provider/oci/common/errors.go
babbageclunk: I can't see how there would be circumstances where this would be true but the cast above would fail - doesn't the internal `serviceError` type implement `ociCommon.ServiceError`?

timClicks: This was mostly paranoia.

provider/oci/environ.go
```diff
@@ -183,6 +183,7 @@ func (e *Environ) AvailabilityZones(ctx envcontext.ProviderCallContext) ([]commo
 	domains, err := e.Identity.ListAvailabilityDomains(ociCtx, request)
 
 	if err != nil {
+		err = providerCommon.HandleCredentialError(err, ctx)
```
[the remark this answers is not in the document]
timClicks: I don't have a preference either way, but it looks like returning nothing is the path of least resistance at this point

provider/oci/storage_volumes.go
timClicks: I've added some logic to avoid unnecessary calls within that method, but haven't added anything similar to the rest of the storage methods yet.

provider/oci/environ.go
```diff
@@ -393,10 +406,13 @@ func (e *Environ) DestroyController(ctx envcontext.ProviderCallContext, controll
...
 	if err != nil {
-		return err
+		providerCommon.HandleCredentialError(err, ctx)
+		return errors.Trace(err)
 	}
 	logger.Debugf("Cleaning up network resources")
-	return e.cleanupNetworksAndSubnets(controllerUUID, "")
+	err = e.cleanupNetworksAndSubnets(controllerUUID, "")
+	providerCommon.HandleCredentialError(err, ctx)
+	return errors.Trace(err)
```
timClicks: This needs to have the check for nil re-introduced

provider/oci/environ.go
```diff
@@ -194,10 +198,11 @@ func (e *Environ) AvailabilityZones(ctx envcontext.ProviderCallContext) ([]commo
 	return zones, nil
 }
 
-// InstanceAvailabilityzoneNames implements common.ZonedEnviron.
+// InstanceAvailabilityZoneNames implements common.ZonedEnviron.
```
timClicks: Thought that this was too minor to justify adding to another PR

provider/oci/common/errors.go
```diff
@@ -0,0 +1,62 @@
...
+		return true
+	}
+
+	return false
+}
+
+// HandleCredentialError marks the current credentials as invalid internally
+//  if Oracle believes that they are expired
+func HandleCredentialError(err error, ctx context.ProviderCallContext) error {
+	common.HandleCredentialError(IsAuthorisationFailure, err, ctx)
```
wallyworld: This method no longer needs to return an error if we are calling it and not using the error result

provider/oci/environ.go
```diff
@@ -393,10 +406,17 @@ func (e *Environ) DestroyController(ctx envcontext.ProviderCallContext, controll
...
 	err = e.StopInstances(ctx, ids...)
 	if err != nil {
-		return err
+		providerCommon.HandleCredentialError(err, ctx)
+		return errors.Trace(err)
 	}
 	logger.Debugf("Cleaning up network resources")
-	return e.cleanupNetworksAndSubnets(controllerUUID, "")
+	err = e.cleanupNetworksAndSubnets(controllerUUID, "")
+	if err != nil {
```
wallyworld: no need for nil check here, just do
```
providerCommon.HandleCredentialError(err, ctx)
return err
```

as the auth err check func should return false for nil, and the cleanupNetworksAndSubnets shoulf already have applied the trace wrapper

provider/oci/storage_volumes.go
```diff
@@ -527,20 +541,41 @@ func (v *volumeSource) AttachVolumes(ctx envcontext.ProviderCallContext, params
...
 	instanceAttachmentMap := map[instance.Id][]ociCore.IScsiVolumeAttachment{}
+
+	var isAuthFailure = func(err error) bool {
+		return allProvidersCommon.MaybeHandleCredentialError(common.IsAuthorisationFailure, err, ctx)
+	}
+
 	for idx, param := range params {
+		// If we have already encountered a credential error, skip subsequent requests
+		if authProblems {
+			ret[idx] = errors.Trace(err)
```
wallyworld: There are so many "err" instances that it's a bug waiting to happen - let's declare a "credErr" and assign any credential error to that and use it to avoid confusion.

provider/oci/environ.go
timClicks: I actually put it back in because I wanted to be able to call `errors.Trace(err)`. If you're happy to go without the trace. then this is an easy fix.

provider/oci/storage_volumes.go
```diff
@@ -15,7 +15,8 @@ import (
 	envcontext "github.com/juju/juju/environs/context"
 	"github.com/juju/juju/environs/tags"
 	"github.com/juju/juju/instance"
-	providerCommon "github.com/juju/juju/provider/oci/common"
+	allProvidersCommon "github.com/juju/juju/provider/common"
```
timClicks: Using this long alias to avoid confusion with the other aliases in the package

provider/oci/storage_volumes.go
```diff
@@ -154,11 +163,21 @@ func (v *volumeSource) CreateVolumes(ctx envcontext.ProviderCallContext, params
 	if params == nil {
 		return []storage.CreateVolumesResult{}, nil
 	}
+	var credErr error
```
timClicks: The pattern that I'm using is to define `credError` is updated from `nil` when a problem with the credentials is detected. We can then test that independently of other errors that might be generated upstream.

timClicks:
Reattempting build (appears to have failed due some transient issue) 

!!build!!

provider/oci/storage_volumes.go
```diff
@@ -154,11 +163,21 @@ func (v *volumeSource) CreateVolumes(ctx envcontext.ProviderCallContext, params
...
 	instanceMap := map[instance.Id]*ociInstance{}
 	for i, volume := range params {
-		vol, err := v.createVolume(volume, instanceMap)
+		if credErr != nil {
+			results[i].Error = errors.Trace(credErr)
+			continue
+		}
+		vol, err := v.createVolume(ctx, volume, instanceMap)
 		if err != nil {
+			if isAuthFailure(err, ctx) {
```
wallyworld: Don't need this wrapper AND the call to common.HandleCredentialError(err, ctx) below
MaybeHandleCredentialError does the same thing that HandleCredentialError does

timClicks:
Merging as `make check` passed in CI and Windows/macOS failures are caused by other issues.

$$merge$$

PR #9506, opened 2018-11-27
Make upgrade-series auto-complete work like other commands.

ExternalReality:
## Description of change

Remove auto-completion tweak that is deprecated by the commands restructuring.

## QA steps

Compare the auto-completion of the `upgrade-series` command with that of the `juju remove-application` command. You will see that in both cases the options are completed indefinitely. Thus `upgrade-series` auto-completion is now consistent with other commands in that if it hits options it will indefinitely complete the options tag not making it to subsequent things like \<machine\> or \<application\> in the command usage string.

### DIFF
--- a/etc/bash_completion.d/juju
+++ b/etc/bash_completion.d/juju
@@ -60,11 +60,6 @@ print ("\n".join(j.get("applications", {}).keys()))
 '   < ${cache_fname}
 }
 
-# Print (return) upgradeseries commands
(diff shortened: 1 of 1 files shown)


ExternalReality:
[Fix auto-complete for new upgrade-series command order.](https://trello.com/c/JQ6q68bl/761-fix-auto-complete-for-new-upgrade-series-command-order)

ExternalReality:
$$MERGE$$

PR #9508, opened 2018-11-28
feature/juju status surface storage deletion better

ycliuhw:
## Description of change

`juju destroy-model` cmd now exits and prints debug cmd rather than hanging there forever if there is an error in destroying machine/filesystem/volume.

## QA steps
1. prepare an IAAS model on aws(could be any cloud, here make assumption for step 3);
2. deploy an application with storage;
3. set the ebs volume to DeleteOnTermination=False;
4. run `juju destroy-model <model-name> --destroy-storage -t 10s` (prints error report after timeout);
4. run `juju destroy-model <model-name> --destroy-storage`, then ctrl +C (prints error report when it exits);
sample output: 
```
⇒  juju destroy-model t1 --destroy-storage -y -t 15s
Destroying model
Waiting on model to be removed, 1 machine(s), 1 application(s), 1 filesystems(s)...
Waiting on model to be removed, 1 machine(s), 1 application(s), 1 filesystems(s)...
Waiting on model to be removed, 1 machine(s), 1 application(s), 1 filesystems(s)...
Waiting on model to be removed, 1 machine(s), 1 application(s), 1 filesystems(s)...
Waiting on model to be removed, 1 machine(s), 1 filesystems(s)...
Waiting on model to be removed, 1 error(s), 1 machine(s), 1 filesystems(s)...
Waiting on model to be removed, 1 error(s), 1 machine(s), 1 filesystems(s)...
^Cctrl+c detected, aborting...

The following errors were encountered during destroying the model.
You can fix the problem causing the errors and run destroy-model again.

Resource    Id  Message
Filesystem  0   removing filesystem: testing invalid errors!!!!!
```

## Documentation changes

slightly changed error handling, probably no doc change required? 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1734287

### DIFF
--- a/cmd/juju/model/destroy.go
+++ b/cmd/juju/model/destroy.go
@@ -6,6 +6,7 @@ package model
 import (
 	"bytes"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/juju/cmd"
@@ -359,6 +360,31 @@ func newTimedModelStatus(ctx *cmd.Context, api DestroyModelAPI, tag names.ModelT
 			}
 			return nil
 		}
+		isErrorMessage := func(s string) bool {
+			return strings.Contains(s, "error")
+		}
+		logger.Criticalf("status -> \n%+v", status)
+		for _, s := range status {
+			for _, v := range s.Machines {
+				if isErrorMessage(v.Status) {
+					ctx.Infof("unable to destroy machine: %+v, \nerror: %q\ntry `juju status` for more details.", v, v.Status)
+					return nil
+				}
+			}
+			for _, v := range s.Filesystems {
+				if isErrorMessage(v.Status) {
+					ctx.Infof("unable to destroy filesystem: %+v, \nerror: %q\ntry `juju status --storage` for more details.", v, v.Status)
+					return nil
+				}
+			}
+			for _, v := range s.Volumes {
+				if isErrorMessage(v.Status) {
+					ctx.Infof("unable to destroy volume: %+v, \nerror: %q\ntry `juju status --storage` for more details.", v, v.Status)
+					return nil
+				}
+			}
+		}
+
 		if l := len(status); l != 1 {
 			ctx.Infof("error finding model status: expected one result, got %d", l)
 			return nil
--- a/cmd/juju/model/destroy.go
+++ b/cmd/juju/model/destroy.go
@@ -6,7 +6,6 @@ package model
 import (
 	"bytes"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/juju/cmd"
@@ -27,6 +26,7 @@ import (
 	rcmd "github.com/juju/juju/cmd/juju/romulus"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/core/model"
+	corestatus "github.com/juju/juju/core/status"
 )
 
 const (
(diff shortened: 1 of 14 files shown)


cmd/juju/model/destroy.go
```diff
@@ -359,6 +360,30 @@ func newTimedModelStatus(ctx *cmd.Context, api DestroyModelAPI, tag names.ModelT
 			}
 			return nil
 		}
+		isError := func(s string) bool {
+			return corestatus.Error.Matches(corestatus.Status(s))
+		}
+		for _, s := range status {
+			for _, v := range s.Machines {
+				if isError(v.Status) {
+					ctx.Infof("unable to destroy machine: %q, \ngot error: %q\ntry `juju status` for more details.", v.Id, v.Message)
```
wallyworld: we need to allow for v.Message being empty

cmd/juju/model/destroy.go
```diff
@@ -359,6 +360,30 @@ func newTimedModelStatus(ctx *cmd.Context, api DestroyModelAPI, tag names.ModelT
...
+					return nil
```
wallyworld: We don't want to return nil in case a script is checking the error code.
Return ErrSilent

worker/machiner/machiner.go
```diff
@@ -191,7 +191,11 @@ func (mr *Machiner) Handle(_ <-chan struct{}) error {
 			logger.Tracef("machine still has storage attached")
 			return nil
 		}
-		return errors.Annotatef(err, "%s failed to set machine to dead", mr.config.Tag)
+		err = errors.Annotatef(err, "%s failed to set machine to dead", mr.config.Tag)
```
wallyworld: Be careful here.
1. err = errors.Annotatef(err, "failed to set status")
is masking the error we really care about. It just needs to be logged as it's not the main issue.
2. We need to check the behaviour - this CodeMachineHasAttachedStorage etc error I think may be part of the normal cleanup - separate workers will be doing the job of removing units. We need to check if it's really appropriate to set an error here and hence stop the destroy CLI, or if it is expected this issue will be transient and go away when the other workers run, in which case we don't want to set status to error.

worker/storageprovisioner/filesystem_ops.go
```diff
@@ -234,7 +234,7 @@ func removeFilesystems(ctx *context, ops map[names.FilesystemTag]*removeFilesyst
 			reschedule = append(reschedule, ops[tag])
 			statuses = append(statuses, params.EntityStatusArgs{
```
wallyworld: Ideally we would see if the error is fatal - the detach operation is rescheduled and may succeed again second time.
One option is to wait till we get X errors and only then mark the status as error and instead of rescheduling, just leave the status as error.

ycliuhw: ah, sorry, ignore me. we are still scheduling the retry even setting the status to Error.

ycliuhw:
!!build!!

ycliuhw:
!!build!!

ycliuhw:
!!build!!

ycliuhw:
!!build!!

ycliuhw:
!!build!!

worker/machiner/machiner.go
```diff
@@ -191,7 +191,11 @@ func (mr *Machiner) Handle(_ <-chan struct{}) error {
...
+		if e := mr.machine.SetStatus(status.Error, errors.Annotate(err, "destroying machine").Error(), nil); e != nil {
```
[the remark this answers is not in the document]
ycliuhw: added

ycliuhw:
$$merge$$

PR #9509, opened 2018-11-28
More robust cmr when controllers disconnect

wallyworld:
## Description of change

When controllers in a cmr scenario lost contact, it's possible that relations get in a stuck state. We're now more aggressive about removing remote units from scope, and also on the consuming side set the remote app to terminated if the offer has been removed. Removing a terminated remote application also triggers extra cleanup of the units to allow any relations to also be deleted.

## QA steps

Run a multi-controller cmr scenario.
firewall the offering side so outbound traffic to the consuming controller is blocked
forcibly remove the offer
open the firewall
consuming side shows saas application in status as terminated
remove-saas will remove the application plus the relations and run the departed/broken hooks

## Bug reference

https://bugs.launchpad.net/juju/+bug/1804362

### DIFF
--- a/state/relation.go
+++ b/state/relation.go
@@ -263,6 +263,33 @@ func (r *Relation) Destroy() (err error) {
 		}
 	}()
 	rel := &Relation{r.st, r.doc}
+
+	remoteApp, isCrossModel, err := r.RemoteApplication()
+	if err != nil {
+		return errors.Trace(err)
+	}
+	// If the status of the consumed app is terminated, we will never
+	// get an orderly exit of units from scope so force the issue.
+	if isCrossModel {
+		statusInfo, err := remoteApp.Status()
+		if err != nil && !errors.IsNotFound(err) {
+			return errors.Trace(err)
+		}
+		if err == nil && statusInfo.Status == status.Terminated {
+			logger.Debugf("forcing cleanup of units for %v", remoteApp.Name())
+			remoteUnits, err := rel.AllRemoteUnits(remoteApp.Name())
+			if err != nil {
+				return errors.Trace(err)
+			}
+			logger.Debugf("got %v relation units to clean", len(remoteUnits))
+			for _, ru := range remoteUnits {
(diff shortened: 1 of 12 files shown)


wallyworld:
[removing a cmr from a defunct model](https://trello.com/c/IRlUdDjF/553-removing-a-cmr-from-a-defunct-model)

state/relation.go
```diff
@@ -556,17 +583,17 @@ func (r *Relation) AllRemoteUnits(appName string) ([]*RelationUnit, error) {
 }
 
 // IsCrossModel returns whether this relation is a cross-model
-// relation.
-func (r *Relation) IsCrossModel() (bool, error) {
+// relation, and if so, the name of the remote application.
+func (r *Relation) IsCrossModel() (bool, *RemoteApplication, error) {
```
babbageclunk: Maybe it would be clearer to rename this to `RemoteApplication() (*RemoteApplication, error)` and have nil indicate that it's not cross-model? It doesn't seem to be used in many places. Or have both and implement `IsCrossModel()` in terms of `RemoteApplication()`?

state/remoteapplication_test.go
```diff
@@ -852,6 +851,67 @@ func (s *remoteApplicationSuite) assertDestroyWithReferencedRelation(c *gc.C, re
...
+	s.assertInScope(c, mysqlru, appStatus == nil || *appStatus != status.Terminated)
+	s.assertInScope(c, wpru, true)
+}
+
+func (s *remoteApplicationSuite) TestDestroyNoStatus(c *gc.C) {
+	s.assertDestroyAppWithStatus(c, nil)
+}
+
+func (s *remoteApplicationSuite) TestDestroyNotTerminated(c *gc.C) {
+	st := status.Active
```
babbageclunk: st is pretty overloaded. Maybe remoteStatus?

state/relation.go
wallyworld: We need to retain the bool. I can rename to RemoteApplication()

PR #9510, opened 2018-11-28
Fix anon auth in notify watcher

wallyworld:
One line fix due to an omission from an earlier forward port.
Properly allow anon invocations of Stop() on notify watcher.


wallyworld:
$$merge$$

PR #9501, opened 2018-11-26
uniter: Smarter config-changed part 2: trust config

babbageclunk:
## Description of change

This follows on from #9498 - it adds a hash watcher for application trust config settings and stores the current hash in the uniter state file. The final watcher for addresses will be in a subsequent PR.

## QA steps

* Bootstrap a controller and model, deploy some applications and relate them. 
* See that the uniter state files for various units includes the trust-hash.
* Change the trust config for one of the applications using `juju trust <application>`.
* See that the trust-hash changes when the config-changed hook fires.

## Documentation changes

None

PR #9518, opened 2018-11-28
uniter: Smarter config-changed pt3 - address config

babbageclunk:
## Description of change

This is a follow-up to #9498 and #9501 to prevent config-changed from being run unless the config has actually, like, changed.

Add Machine.WatchAddressesHash and Unit.WatchContainerAddressesHash that emit the hash of the current addresses, and feed that down to the uniter (in the RemoteStateWatcher).

This is the last item that can trigger config-changed so I've removed the ConfigVersion field from local state and the remote state snapshot. That means that we only fire the hook when one or more of the config, trust and address hashes is different from the state on disk. There were a lot of uniter tests that assumed that config-changed would always run after the unit agent restarted, those have been fixed - in `UniterSuite.TestUniterRelations` there was one that was using the hook to extract relation data so I change configuration explicitly so that still works.

`UniterSuite.TestRunCommand` has a workaround fix until I can come back to sort out what the underlying problem is.

## QA steps

* Deploy an application to a new model.
* See that the config-changed hook for the new unit runs after install.
* Look in the uniter state file to see the hashes for config, trust and addresses.
* Change the config, see that the config-changed hook runs and the hash in the state file is updated.
* Do the same for trust.
* Not sure how to change addresses for the unit, but if you can, do that too.
* Stop and start the unit agent.
* Config-changed shouldn't fire.
* Stop and start the controller.
* When the unit agent reconnects it shouldn't run config-changed.

## Documentation changes

We will need to clearly highlight this behaviour change in charming documentation - it's likely there are some charms that (incorrectly) rely on the config-changed hook running after agent start to start their workload, rather than setting up service files.


apiserver/facades/agent/uniter/uniter_test.go
```diff
@@ -2589,6 +2589,90 @@ func (s *uniterSuite) TestWatchCAASUnitAddresses(c *gc.C) {
...
+
+	result, err := uniterAPI.WatchUnitAddressesHash(args)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(result, gc.DeepEquals, params.StringsWatchResults{
+		Results: []params.StringsWatchResult{
+			{Error: apiservertesting.ErrUnauthorized},
+			{
+				StringsWatcherId: "1",
+				// The container doesn't have an address.
+				Changes: []string{""},
```
wallyworld: Why not an empty sha256 hash like for IAAS units?

state/unit_test.go
```diff
@@ -2453,3 +2453,73 @@ func (s *CAASUnitSuite) TestWatchContainerAddresses(c *gc.C) {
...
+	addr := "10.0.0.1"
+	updateUnits.Updates = []*state.UpdateUnitOperation{unit.UpdateOperation(state.UnitUpdateProperties{
+		Address: &addr,
+	})}
+	err = s.application.UpdateUnits(&updateUnits)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(err, jc.ErrorIsNil)
+	wc.AssertChange("0b94e9ad737e2ff4d50dc9d8cd443d336e6a6d638b038e31b973959be89d5dec")
+
+	// Set different machine addresses: reported.
```
wallyworld: container

state/watcher.go
```diff
@@ -3709,40 +3709,143 @@ func (w *containerAddressesWatcher) loop() error {
...
+
+// WatchAddressesHash returns a StringsWatcher that emits the hash of
+// the machine's (sorted) addresses whenever they change.
+func (m *Machine) WatchAddressesHash() StringsWatcher {
+	w := &hashWatcher{
+		commonWatcher: newCommonWatcher(m.st),
+		out:           make(chan []string),
+		collection:    machinesC,
+		id:            m.doc.DocID,
+		hash: func() (string, error) {
```
wallyworld: This is split out as a separate func for hashSettingsWatcher
Do same here?

state/watcher.go
```diff
@@ -3709,40 +3709,143 @@ func (w *containerAddressesWatcher) loop() error {
...
+				// Addresses guarantees that each value will only be
+				// returned once - addresses from provider take
+				// precedence over those from the machine.
+				return addresses[i].Value < addresses[j].Value
+			})
+			hash := sha256.New()
+			for _, address := range addresses {
+				hash.Write([]byte(address.Value))
+				hash.Write([]byte(address.Type))
+				hash.Write([]byte(address.Scope))
```
wallyworld: scope is included in the hash yet in some tests where scope is change but address is the same the watcher doesn't fire?

worker/uniter/util_test.go
```diff
@@ -585,7 +585,8 @@ func (s verifyRunning) step(c *gc.C, ctx *context) {
 	if s.minion {
 		hooks = append(hooks, "leader-settings-changed")
 	}
-	hooks = append(hooks, "config-changed")
+	// We don't expect config-changed to always run on agent restart
```
wallyworld: yay

state/watcher.go
```diff
@@ -3709,40 +3709,143 @@ func (w *containerAddressesWatcher) loop() error {
...
+// WatchContainerAddressesHash returns a StringsWatcher that emits a
+// hash of the unit's container address whenever it changes.
+func (u *Unit) WatchContainerAddressesHash() StringsWatcher {
+	firstCall := true
+	w := &hashWatcher{
+		commonWatcher: newCommonWatcher(u.st),
+		out:           make(chan []string),
+		collection:    cloudContainersC,
+		id:            u.st.docID(u.globalKey()),
+		hash: func() (string, error) {
```
wallyworld: also break this out to keep things consistent?

apiserver/facades/agent/uniter/uniter_test.go
babbageclunk: This is an implementation detail - because a cloud container can only have 0 or 1 address, we check for a nil pointer and return "", rather than creating the hash. A machine can have many addresses, so we loop over a slice and feed them into the hash - if there are zero we end up with a specific hash value (for no items). I don't think it matters specifically, unless you have strong feelings about it I'm not going to change it.

state/unit_test.go
babbageclunk: Fixed!

worker/uniter/util_test.go
babbageclunk: yay!

babbageclunk:
$$merge$$

PR #9524, opened 2018-11-29
apiserver/facades/client/cloud: avoid naturalsort

rogpeppe:
The updateCredentials methods use naturalsort to sort model UUIDs which is a bit weird and means it's harder for external packages to test against. Avoid that, and also make the tests avoid the ordering dependency by putting the UUID into the mock model (in general an API endpoint implementation should be free to do its work in an arbitrary order or even concurrently if desired).


hmlanigan:
!!build!!

rogpeppe:
$$merge$$

PR #9525, opened 2018-11-30
Fixes lp#1804630: panic when logging failures.

anastasiamac:
## Description of change

There are some details that are not available when failures occur. Re-arrane log messages to only access available information.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1804630

### DIFF
--- a/provider/openstack/provider.go
+++ b/provider/openstack/provider.go
@@ -1148,7 +1148,8 @@ func (e *Environ) StartInstance(ctx context.ProviderCallContext, args environs.S
 				break
 			}
 			if server == nil {
-				continue
+				logger.Warningf("may have lost contact with nova api while creating instances, some stray instances may be around and need to be deleted")
+				break
 			}
 			var serverDetail *nova.ServerDetail
 			serverDetail, err = waitForActiveServerDetails(client, server.Id, 5*time.Minute)
@@ -1164,7 +1165,7 @@ func (e *Environ) StartInstance(ctx context.ProviderCallContext, args environs.S
 				if serverDetail.Fault != nil {
 					faultMsg = fmt.Sprintf(" with fault %q", serverDetail.Fault.Message)
 				} else {
-					logger.Debugf("getting activer server details from nova failed without fault details")
+					logger.Debugf("getting active server details from nova failed without fault details")
 				}
 				logger.Infof("Deleting instance %q in ERROR state%v", server.Id, faultMsg)
 				if err = e.terminateInstances([]instance.Id{instance.Id(server.Id)}); err != nil {
--- a/provider/openstack/provider.go
+++ b/provider/openstack/provider.go
@@ -1155,15 +1155,25 @@ func (e *Environ) StartInstance(ctx context.ProviderCallContext, args environs.S
 			} else if serverDetail.Status == nova.StatusActive {
 				break
 			} else if serverDetail.Status == nova.StatusError {
+				if server == nil {
+					break
+				}
 				// Perhaps there is an error case where a retry in the same AZ
 				// is a good idea.
-				logger.Infof("Instance %q in ERROR state with fault %q", server.Id, serverDetail.Fault.Message)
-				logger.Infof("Deleting instance %q in ERROR state", server.Id)
+				faultDetails := ""
+				faultMsg := ""
+				if serverDetail != nil && serverDetail.Fault != nil {
+					faultDetails = serverDetail.Fault.Message
+					faultMsg = fmt.Sprintf(" with fault %q", faultMsg)
+				}
+				logger.Infof("Deleting instance %q in ERROR state%v", server.Id, faultMsg)
 				if err = e.terminateInstances([]instance.Id{instance.Id(server.Id)}); err != nil {
 					logger.Debugf("Failed to delete instance in ERROR state, %q", err)
 				}
 				server = nil
-				err = errors.New(serverDetail.Fault.Message)
+				if faultDetails != "" {
+					err = errors.New(faultDetails)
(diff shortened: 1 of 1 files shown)


anastasiamac:
[lp#1804630, panic when logging openstack provider failures](https://trello.com/c/aeOPU6nz/561-lp1804630-panic-when-logging-openstack-provider-failures)

provider/openstack/provider.go
```diff
@@ -1155,15 +1155,25 @@ func (e *Environ) StartInstance(ctx context.ProviderCallContext, args environs.S
...
+				if server == nil {
+					break
+				}
 				// Perhaps there is an error case where a retry in the same AZ
 				// is a good idea.
-				logger.Infof("Instance %q in ERROR state with fault %q", server.Id, serverDetail.Fault.Message)
-				logger.Infof("Deleting instance %q in ERROR state", server.Id)
+				faultDetails := ""
+				faultMsg := ""
+				if serverDetail != nil && serverDetail.Fault != nil {
```
hmlanigan: Suggest we move the check of serverDetail != nil to before line 1155.

provider/openstack/provider.go
```diff
@@ -1155,15 +1155,25 @@ func (e *Environ) StartInstance(ctx context.ProviderCallContext, args environs.S
 			} else if serverDetail.Status == nova.StatusActive {
 				break
 			} else if serverDetail.Status == nova.StatusError {
+				if server == nil {
+					break
```
hmlanigan: We're in deeper trouble if this is true.  Suggest moving to before 1150 and set err in this case, otherwise we'll try to use 'server' at ln 1208.

provider/openstack/provider.go
```diff
@@ -1155,15 +1155,25 @@ func (e *Environ) StartInstance(ctx context.ProviderCallContext, args environs.S
...
+					faultDetails = serverDetail.Fault.Message
+					faultMsg = fmt.Sprintf(" with fault %q", faultMsg)
+				}
+				logger.Infof("Deleting instance %q in ERROR state%v", server.Id, faultMsg)
 				if err = e.terminateInstances([]instance.Id{instance.Id(server.Id)}); err != nil {
 					logger.Debugf("Failed to delete instance in ERROR state, %q", err)
 				}
 				server = nil
-				err = errors.New(serverDetail.Fault.Message)
+				if faultDetails != "" {
```
hmlanigan: we need to set err even if we didn't get the faultDetails, otherwise we'll try to continue on instead of failing the instance and returning to the provisioner.

Instance %q is in an Error State, unable to determine cause??

provider/openstack/provider.go
```diff
@@ -1155,15 +1155,25 @@ func (e *Environ) StartInstance(ctx context.ProviderCallContext, args environs.S
...
 				// Perhaps there is an error case where a retry in the same AZ
 				// is a good idea.
-				logger.Infof("Instance %q in ERROR state with fault %q", server.Id, serverDetail.Fault.Message)
-				logger.Infof("Deleting instance %q in ERROR state", server.Id)
+				faultDetails := ""
+				faultMsg := ""
+				if serverDetail != nil && serverDetail.Fault != nil {
+					faultDetails = serverDetail.Fault.Message
+					faultMsg = fmt.Sprintf(" with fault %q", faultMsg)
+				}
```
hmlanigan: Suggest a logger.Debug if the check at line 1165 fails.

anastasiamac:
!!build!!

provider/openstack/provider.go
```diff
@@ -1147,23 +1147,31 @@ func (e *Environ) StartInstance(ctx context.ProviderCallContext, args environs.S
...
 			} else if serverDetail.Status == nova.StatusError {
 				// Perhaps there is an error case where a retry in the same AZ
 				// is a good idea.
-				logger.Infof("Instance %q in ERROR state with fault %q", server.Id, serverDetail.Fault.Message)
-				logger.Infof("Deleting instance %q in ERROR state", server.Id)
+				faultMsg := " unable to determine fault details"
+				if serverDetail.Fault != nil {
+					faultMsg = fmt.Sprintf(" with fault %q", serverDetail.Fault.Message)
+				} else {
+					logger.Debugf("getting activer server details from nova failed without fault details")
```
hmlanigan: nit s/activer/active

provider/openstack/provider.go
```diff
@@ -1147,23 +1147,31 @@ func (e *Environ) StartInstance(ctx context.ProviderCallContext, args environs.S
 			if err != nil {
 				break
 			}
+			if server == nil {
+				continue
```
hmlanigan: when testing with instrumented code, this causes juju to not delete the instance that was created.  suggest a message to the user, to look for stray instances.

i believe the break should be used, it will allow the provisioner to retry, though an instance will still be left behind. Given the past changes to parallel the provisioner, perhaps the for loop here should have been removed then, as retries should be handled by the it. maybe a change for another day.

provider/openstack/provider.go
anastasiamac: Oh good pick! I have just moved the code around without noticing it has a typo :) will fix.

provider/openstack/provider.go
anastasiamac: I agree - I'll add a WARN log message to look for stray instances and will change 'continue' to 'break'. 

As for the rest of the comment, is it possible for you to file a tech debt bug describing the improvement and its benefit?

anastasiamac:
$$merge$$

anastasiamac:
$$merge$$

PR #9470, opened 2018-11-15
Ensure that we handle the case of non-LXD

SimonRichardson:
## Description of change

The following ensures that we handle what happens if we're not
running on LXD containers.

## QA steps

```
juju bootstrap localhost lxd-profile-test
juju deploy ./testcharms/charm-repo/quantal/lxd-profile
juju upgrade-charm lxd-profile --path ./testcharms/charm-repo/quantal/lxd-profile
watch juju status
```

## Documentation changes

The following change ensures that we don't have any issues with running
on KVM containers et al.

## Bug reference

N/A

PR #9535, opened 2018-12-03
Ensure that we handle the case of non-LXD

SimonRichardson:
Supersedes https://github.com/juju/juju/pull/9470

## Description of change

We actually don't care about this TODO, as we handle it in other
ways, which include:

 - Only provisioning the correct machine/container with the correct
   provisioner task. If the provisioner task doesn't support the
   lxd profile, then this becomes a no-op.
 - In the #9513 PR we always clean up after a lxd profile, so that
   any documents created in this process are never left in an
   orphan state.

## QA steps

N/A

## Documentation changes

All providers are treated equally, as long as we always clean
up after ourselves.

### DIFF
--- a/worker/provisioner/provisioner.go
+++ b/worker/provisioner/provisioner.go
@@ -405,6 +405,10 @@ func (p *containerProvisioner) getRetryWatcher() (watcher.NotifyWatcher, error) 
 }
 
 func (p *containerProvisioner) getProfileWatcher() (watcher.StringsWatcher, error) {
+	// Note: we don't care what type the container is when watching. The
+	// provisioner task will make this become a no-op.
+	// Also we'll always clean up any documents once the uniter has finished
+	// deploying/upgrading a charm.
 	machine, err := p.getMachine()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


SimonRichardson:
https://trello.com/c/s6WVPg30/871-handle-the-case-of-non-lxd-providers

SimonRichardson:
https://trello.com/c/s6WVPg30/871-handle-the-case-of-non-lxd-providers

worker/provisioner/provisioner.go
```diff
@@ -405,10 +405,6 @@ func (p *containerProvisioner) getRetryWatcher() (watcher.NotifyWatcher, error)
 }
 
 func (p *containerProvisioner) getProfileWatcher() (watcher.StringsWatcher, error) {
-	if p.containerType != instance.LXD {
```
hmlanigan: Should we add a note on why we're not checking this?  For our future selves and head scratching later?

SimonRichardson:
$$merge$$

PR #9536, opened 2018-12-03
ensure we can set series on the subordinate.

ExternalReality:
## Description of change

The `assess_upgrade_series` acceptance tests adds a step to ensure that subordinate applications can be set to the correct series after a success issuance of the `upgrade-series` command.

## QA steps

Run the acceptance test


mitechie:
$$merge$$

PR #9538, opened 2018-12-03
(WIP) juju show-model indicates whether the model is the controller model 

timClicks:
## Description of change

`apiserver/params/model.ModelInfo` is extended  with an extra field, `IsController`. Adding a field, rather than a method simplifies reflection for when the API is sent over the wire.

## QA steps

* `juju list-models --format=yaml` includes an `"is-controller"` field for each model
* `juju show-model <model-name> --format=yaml` includes an `"is-controller"` field 
* `juju status --format=yaml` includes an `"is-controller"` field

## Documentation changes

A note should be made that the output has been updated for `list-models`, `show-model` and `status` commands.

## Bug reference

lp:1803247 https://bugs.launchpad.net/juju/+bug/1803247

### DIFF
--- a/cmd/juju/controller/listmodels.go
+++ b/cmd/juju/controller/listmodels.go
@@ -16,11 +16,13 @@ import (
 	"gopkg.in/juju/names.v2"
 
 	"github.com/juju/juju/api/base"
+	"github.com/juju/juju/api/controller"
 	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/cmd/output"
 	"github.com/juju/juju/core/model"
+	"github.com/juju/juju/environs/config"
 	"github.com/juju/juju/jujuclient"
 )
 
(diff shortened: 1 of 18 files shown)


timClicks:
Note, there is a comment in juju/juju/apiserver/facades/client/client/client_test.go TestModelInfo that says we don't want clients to know the controller UUID. Perhaps that's why this is more complicated than it seems like it should be.

> The controller UUID is not returned by the ModelInfo endpoint on the Client facade.

timClicks:
[unable to determine if a model is a controller model](https://trello.com/c/1ZOKh0bb/560-unable-to-determine-if-a-model-is-a-controller-model)

cmd/juju/controller/listmodels.go
```diff
@@ -81,6 +83,18 @@ func (c *modelsCommand) SetFlags(f *gnuflag.FlagSet) {
 
 // Run implements Command.Run
 func (c *modelsCommand) Run(ctx *cmd.Context) error {
+	ctrlRoot, err := c.NewAPIRoot()
```
howbazaar: I think we should be returning the is-controller flag back in the API call when asking for the model list.

Otherwise other API users will need to duplicate this logic.

cmd/juju/model/show.go
```diff
@@ -84,6 +86,16 @@ func (c *showModelCommand) Init(args []string) error {
 
 // Run implements Command.Run.
 func (c *showModelCommand) Run(ctx *cmd.Context) error {
+	ctrlRoot, err := c.NewControllerAPIRoot()
```
howbazaar: Similarly here.

timClicks:
Closing this PR - will redo

PR #9542, opened 2018-12-05
bug 1805567: don't assume if charm has no lxd profile, that any applied lxd profile is from provious versions 

hmlanigan:
## Description of change

if the charm doesn't have a profile and no profile has been previously
applied, set the new instanceCharmProfileData to NotRequired

## QA steps

1. charm pull ubuntu
2.  juju deploy path-to-charm/ubuntu
2. juju deploy ./testcharms/charm-repo/quantal/lxd-profile-alt --to lxd 
2. juju deploy ./testcharms/charm-repo/quantal/lxd-profile
2. juju deploy neutron-openvswitch;juju add-relation neutron-openvswitch ubuntu
2.  juju upgrade-charm ubuntu --path   path-to-charm/ubuntu
2.  juju upgrade-charm lxd-profile-alt --path ./testcharms/charm-repo/quantal/lxd-profile-alt
2.  juju upgrade-charm lxd-profile --path ./testcharms/charm-repo/quantal/lxd-profile
2.  juju deploy ./testcharms/charm-repo/quantal/lxd-profile-subordinate;
2.  juju add-relation lxd-profile-subordinate lxd-profile
2. juju add-relation lxd-profile-subordinate ubuntu
2.  juju upgrade-charm lxd-profile-subordinate --path ./testcharms/charm-repo/quantal/lxd-profile-subordinate
3. lxc profile list

## Documentation changes

n/a

## Bug reference

https://bugs.launchpad.net/juju/+bug/1805567

### DIFF
--- a/state/machine.go
+++ b/state/machine.go
@@ -2215,7 +2215,8 @@ func (m *Machine) VerifyUnitsSeries(unitNames []string, series string, force boo
 }
 
 // SetUpgradeCharmProfile sets a application name and a charm url for
-// machine's needing a charm profile change.  For a container only.
+// machine's needing a charm profile change.  For an LXD container or
+// machine only.
 func (m *Machine) SetUpgradeCharmProfile(appName, chURL string) error {
 	buildTxn := func(attempt int) ([]txn.Op, error) {
 		if attempt > 0 {
@@ -2227,6 +2228,7 @@ func (m *Machine) SetUpgradeCharmProfile(appName, chURL string) error {
 		if life == Dead || life == Dying {
 			return nil, ErrDead
 		}
+
 		charmURL, err := charm.ParseURL(chURL)
 		if err != nil {
 			return nil, errors.Trace(err)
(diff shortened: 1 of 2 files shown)


state/machine.go
```diff
@@ -2237,40 +2239,56 @@ func (m *Machine) SetUpgradeCharmProfile(appName, chURL string) error {
...
 		}
 
-		provisioned := true
+		ops := []txn.Op{
+			{
+				C:      machinesC,
+				Id:     m.doc.DocID,
+				Assert: isAliveDoc,
+			},
+			// TODO: lxd profile - enhance checks for this set of transactions
```
hmlanigan: this is my next card related to excessive content

hmlanigan:
!!build!!

state/machine.go
```diff
@@ -2237,40 +2239,56 @@ func (m *Machine) SetUpgradeCharmProfile(appName, chURL string) error {
...
-		}
-		// we need to ensure that the machine is provisioned before we attempt
-		// to set that the upgrade charm profile complete to not required.
-		// If the profiles are empty and there is nothing to apply, we can
-		// assume that nothing is required.
-		if len(profiles) == 0 && emptyProfile {
-			var ops []txn.Op
-			if provisioned {
-				ops = append(ops, m.checkCharmProfilesIsEmptyOp())
+		} else {
```
jameinel: can we get a comment here? I think using an explicit flag might be clearer, eg:

if err != nil {
  if errors.IsNotProvisioned(err) {
    // we don't need to fail here
    provisioned = false
  } else {
    return nil, errors.Trace(err)
  }

if provisioned {
  ops = append(ops, txn.Op{...}
}

Might more clearly convey why the code is structured that way.

state/machine.go
```diff
@@ -2237,40 +2239,56 @@ func (m *Machine) SetUpgradeCharmProfile(appName, chURL string) error {
...
+			ops = append(ops, txn.Op{
+				C:      instanceDataC,
+				Id:     m.doc.DocID,
+				Assert: bson.D{{"charm-profiles", profiles}},
```
jameinel: I should clarify that I think this mechanism of clearly asserting the list of profiles hasn't changed since we looked at it is correct. It doesn't matter if the old value is an empty list or one with things in it.
The important bit is that we're making assumptions of what to do based on the list of profiles that we read. If that list changes concurrently, then we need to reevaluate what to do.

state/machine.go
```diff
@@ -2215,7 +2215,8 @@ func (m *Machine) VerifyUnitsSeries(unitNames []string, series string, force boo
 }
 
 // SetUpgradeCharmProfile sets a application name and a charm url for
-// machine's needing a charm profile change.  For a container only.
+// machine's needing a charm profile change.  For an LXD container or
```
jameinel: sets "an" application name

state/machine.go
hmlanigan: babbageclunk helped me find at least one excessive contention yesterday.  it's not new to this pr, but has been on going.  that work is being done in a different pr.  but perhaps there is more than one.

We're reading fresh, but also the charm itself was found from the same charmURL argument. It's a type issue of string vs charm.URL for the assert.

state/machine.go
hmlanigan: There is a case where this isn't correct.  "charm-profiles" is setup with omit empty, so if there are no charm-profiles, this check fails.  Babbageclunk helped me find it. Will be updating this code shortly.

state/machine.go
hmlanigan: restructured the code section and added comments.

state/machine.go
hmlanigan: done

hmlanigan:
The tests failing, provisionerSuite.TestCharmProfileChangeInfo and provisionerSuite.TestCharmProfileChangeInfoSubordinate, haven't failed when I run it locally.  Trying to sort that out.

api/provisioner/provisioner_test.go
```diff
@@ -1000,8 +1000,8 @@ func (s *provisionerSuite) TestWatchModelMachinesCharmProfiles(c *gc.C) {
 	wc := watchertest.NewStringsWatcherC(c, w, s.BackingState.StartSync)
 	defer wc.AssertStops()
 
-	// Update the upgrade-charm charm profile to trigger watcher.
-	s.machine.SetUpgradeCharmProfile("app-name", "local:quantal/lxd-profile-0")
+	// Trigger the watcher.
+	err = s.machine.SetUpgradeCharmProfile("lxd-profile", "local:quantal/lxd-profile-0")
```
hmlanigan: drive by fix

hmlanigan:
This past test failure had nothing to do  provisionerSuite.TestCharmProfileChangeInfo* tests, it was a timeout panic with the raft worker

hmlanigan:
$$merge$$

hmlanigan:
QA looks good!!

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #9543, opened 2018-12-05
WIP: Old-style, non-multiseries charms should not support all series.

ExternalReality:
## Description of change

Charms that don't list any series in their metadata (i.e legacy non-multiseries charms), in some cases validated against all valid series. This change corrects that behavior. For example `juju` allows a deployed Wordpress charm to be set to any series even though the documentation for `set-series` indicates that only series supported by the charm should be allowed.

## QA steps

`juju deploy wordpress`
`juju set-series wordpress cosmic`

The above now errors whereas before it would work fine.

### DIFF
--- a/state/application.go
+++ b/state/application.go
@@ -1313,10 +1313,14 @@ func (a *Application) VerifySupportedSeries(series string, force bool) error {
 	if err != nil {
 		return err
 	}
-	_, seriesSupportedErr := charm.SeriesForCharm(series, ch.Meta().Series)
+	supportedSeries := ch.Meta().Series
+	if len(supportedSeries) == 0 {
+		supportedSeries = append(supportedSeries, ch.URL().Series)
+	}
+	_, seriesSupportedErr := charm.SeriesForCharm(series, supportedSeries)
 	if seriesSupportedErr != nil && !force {
 		return &ErrIncompatibleSeries{
-			SeriesList: ch.Meta().Series,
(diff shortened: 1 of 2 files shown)


ExternalReality:
[Fix bug in application commands that allowed old-style, non-multi-series charms to validate against any series.](https://trello.com/c/7JHL6tfO/879-fix-bug-in-application-commands-that-allowed-old-style-non-multi-series-charms-to-validate-against-any-series)

state/machine_test.go
```diff
@@ -2562,12 +2562,12 @@ func (s *MachineSuite) setupTestUpdateMachineSeries(c *gc.C) *state.Machine {
 	err = ru.EnterScope(nil)
 	c.Assert(err, jc.ErrorIsNil)
 
-	ch2 := state.AddTestingCharmMultiSeries(c, s.State, "wordpress")
-	app2 := state.AddTestingApplicationForSeries(c, s.State, "precise", "wordpress", ch2)
-	unit2, err := app2.AddUnit(state.AddUnitParams{})
-	c.Assert(err, jc.ErrorIsNil)
-	err = unit2.AssignToMachine(mach)
-	c.Assert(err, jc.ErrorIsNil)
+	// ch2 := state.AddTestingCharmMultiSeries(c, s.State, "mysql")
```
manadart: Delete me.

PR #9544, opened 2018-12-06
raft leases: Forward commands concurrently

babbageclunk:
## Description of change

Using raft leases on a controller with much slower IO revealed that the lease commands were being applied one-at-a-time - I didn't realise it, but pubsub callbacks are processed sequentially, not as the messages come in. This doesn't cause an apparent problem on a machine where the disk is very fast (even when in HA and with lots of applications) but with a slow disk and many applications it can mean that requests come in faster than they're applied, and all the lease operations start timing out. 

Change the forwarder to apply commands in parallel - the raft library expects this and puts more commands into the heartbeat messages to dramatically improve throughput.

Adds lots of trace logging and Prometheus metrics to provide more insight into what's happening in the raft lease store and inside the raft node.

## QA steps

* Bootstrap a controller on a machine with slow disk and enable HA.
* Deploy more than 100 applications.
* Monitor the raft lease request times with Prometheus or juju_metrics.
* Ensure that the values are well under the 5 second threshold.
* Watch for timeouts - we shouldn't see these increasing unless the controller is still starting or there has just been a raft leader election.
* Create a new unit and remove the current leader - ensure that leadership is transferred.

## Documentation changes
None

### DIFF
--- a/cmd/jujud/agent/machine/manifolds.go
+++ b/cmd/jujud/agent/machine/manifolds.go
@@ -857,14 +857,15 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 		// The global lease manager tracks lease information in the raft
 		// cluster rather than in mongo.
 		leaseManagerName: ifController(leasemanager.Manifold(leasemanager.ManifoldConfig{
-			AgentName:      agentName,
-			ClockName:      clockName,
-			CentralHubName: centralHubName,
-			FSM:            leaseFSM,
-			RequestTopic:   leaseRequestTopic,
-			Logger:         loggo.GetLogger("juju.worker.lease.raft"),
-			NewWorker:      leasemanager.NewWorker,
-			NewStore:       leasemanager.NewStore,
+			AgentName:         agentName,
+			ClockName:         clockName,
+			CentralHubName:    centralHubName,
+			FSM:               leaseFSM,
+			RequestTopic:      leaseRequestTopic,
+			Logger:            loggo.GetLogger("juju.worker.lease.raft"),
+			MetricsRegisterer: config.PrometheusRegisterer,
+			NewWorker:         leasemanager.NewWorker,
+			NewStore:          leasemanager.NewStore,
 		})),
 
 		validCredentialFlagName: credentialvalidator.Manifold(credentialvalidator.ManifoldConfig{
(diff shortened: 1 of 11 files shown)


babbageclunk:
[fix raft lease timeout problem found at Jalasoft](https://trello.com/c/hF1LYMOp/569-fix-raft-lease-timeout-problem-found-at-jalasoft)

cmd/jujud/agent/machine/manifolds.go
```diff
@@ -290,6 +290,9 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 	controllerTag := agentConfig.Controller()
 
 	leaseFSM := raftlease.NewFSM()
+	if config.PrometheusRegisterer != nil {
+		raft.RegisterMetrics(config.PrometheusRegisterer)
```
howbazaar: We need to unregister too.

core/raftlease/store.go
```diff
@@ -53,18 +57,29 @@ type StoreConfig struct {
...
+	result := &Store{
 		fsm:      config.FSM,
 		hub:      config.Hub,
 		config:   config,
 		prevTime: config.FSM.GlobalTime(),
+		metrics:  newMetricsCollector(),
+	}
+
+	if config.MetricsRegisterer != nil {
+		config.MetricsRegisterer.Unregister(result.metrics)
```
howbazaar: The unregister should happen on tear down, closing... etc.

worker/raft/metrics.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+// RegisterMetrics connects the metrics gathered in the
+// hashicorp/raft library code with our prometheus collector.
+func RegisterMetrics(registerer prometheus.Registerer) error {
+	sink, err := pmetrics.NewPrometheusSink()
+	if err != nil {
+		return errors.Trace(err)
+	}
+	registerer.Unregister(sink)
```
howbazaar: Why are you calling unregister before register?

worker/raft/raftforwarder/worker.go
```diff
@@ -137,7 +144,9 @@ func (w *forwarder) processRequest(command []byte) (raftlease.ForwardResponse, e
 	if !ok {
 		return empty, errors.Errorf("expected an FSMResponse, got %T: %#v", respValue, respValue)
 	}
+	w.config.Logger.Tracef("%d: after fsm response", reqID)
```
howbazaar: I think these two trace lines could probably be removed now...

babbageclunk:
Fixes https://bugs.launchpad.net/juju/+bug/1805764

PR #9547, opened 2018-12-06
Prevent accidental cross channel charm upgrades.

ExternalReality:
## Description of change

When running upgrade charm without specifying a channel there was chance that a charm would be upgraded across channels. For example, a charm that was originally obtained from the Beta channel could be accidentally upgraded to a charm from the Stable channel.

## QA steps

Release a charm to the beta channel:

`juju release <charm> --channel beta`

Deploy the charm:

`juju deploy <charm>`

Release a revision to stable channel

`juju release <charm> --channel stable`

Release a revision to beta channel

`juju release <charm> --channel beta`

Upgrade the charm of the deployed application without specifying a channel

`juju upgrade-charm <application>`

Notice that after this patch the beta channel's revision should be deployed and not the stable channel's revision. This because the previous deployment's channel should be respected by the `upgrade-charm` command.

## Documentation changes

It may be useful to specifically note that the previous deployment's/upgrade's channel will be respected by this command and if another channel is desired it must be specified.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1804669

### DIFF
--- a/apiserver/facades/client/application/get.go
+++ b/apiserver/facades/client/application/get.go
@@ -87,6 +87,7 @@ func (api *APIBase) getConfig(
 		ApplicationConfig: appConfigInfo,
 		Constraints:       constraints,
 		Series:            app.Series(),
+		Channel:           string(app.Channel()),
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


ExternalReality:
Note that another implementation option would be to remove the channel info from client object since, as @ericsnowcurrently, points out in comments that information is likely to become stale in long running threads that hold a client. The client should always reach out to the DB for channel info since it is likely to change between client instantiations. This solution is adequate since for cli commands the client is short lived.

ExternalReality:
$$MERGE$$

PR #9548, opened 2018-12-07
Parallelise the raft apply calls and hook up metrics

howbazaar:
This work is is a combination of work from myself and @babbageclunk.

If using spinning rust for disks rather than SSDs, the i/o time significantly impacts how raft performs. This branch parallelises the requests from the raftforwarder to the underlying raft FSM.

Also extra debugging was added, and prometheus metrics added for the lease manager, and hooked up to the metrics available from raft.

### DIFF
--- a/worker/lease/manager.go
+++ b/worker/lease/manager.go
@@ -141,7 +141,7 @@ func (manager *Manager) Wait() error {
 
 // loop runs until the manager is stopped.
 func (manager *Manager) loop() error {
-	if collector, ok := manager.config.Store.(prometheus.Collector); ok {
+	if collector, ok := manager.config.Store.(prometheus.Collector); ok && manager.config.PrometheusRegisterer != nil {
 		// The store implements the collector interface, but the lease.Store
 		// doen't expose those.
 		manager.config.PrometheusRegisterer.Register(collector)
--- a/worker/lease/manager.go
+++ b/worker/lease/manager.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/juju/clock"
 	"github.com/juju/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"gopkg.in/juju/worker.v1/catacomb"
 	"gopkg.in/retry.v1"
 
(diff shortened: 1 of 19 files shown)


howbazaar:
Supercedes #9544.

worker/raft/metrics.go
```diff
@@ -17,7 +17,7 @@ func registerMetrics(registerer prometheus.Registerer) (prometheus.Collector, er
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	_, err = metrics.NewGlobal(metrics.DefaultConfig("jujumetrics"), sink)
+	_, err = metrics.NewGlobal(metrics.DefaultConfig("juju"), sink)
```
babbageclunk: I think this needs to be something other than "juju" (which is used for metrics in our code) - otherwise there's a possibility of the register failing because of name collisions between juju/juju metrics and hashicorp/raft ones.

worker/lease/manager.go
```diff
@@ -140,6 +141,13 @@ func (manager *Manager) Wait() error {
 
 // loop runs until the manager is stopped.
 func (manager *Manager) loop() error {
+	if collector, ok := manager.config.Store.(prometheus.Collector); ok {
+		// The store implements the collector interface, but the lease.Store
+		// doen't expose those.
+		manager.config.PrometheusRegisterer.Register(collector)
```
babbageclunk: Yeah, it's definitely nicer to do this here - for some reason it didn't occur to me that the type collecting the metrics didn't need to be the one to register itself.

worker/raft/metrics.go
howbazaar: I checked, and they all say "juju_raft_..."

howbazaar: If we find it a problem later, we can look at it then, but this makes all our metrics register nicely with the same prefix.

howbazaar:
$$merge$$

howbazaar:
$$merge$$

howbazaar:
$$merge$$

worker/raft/metrics.go
babbageclunk: Yeah, fair enough.

PR #9549, opened 2018-12-07
Report whether a model is a controller model

timClicks:
## Description of change

This PR adds the ability to identify whether a model is a controller model from the CLI and the GUI. 

From the [bug report](https://bugs.launchpad.net/juju/+bug/1803247)

> Because the controller model is "special" the GUI blocks certain actions in it, like the deletion action.
>
> The issue is that we can no longer determine if a model is a controller model as the `model-uuid` and the `controller-uuid` no longer match and the name "controller" can be changed.
>
> There doesn't appear to be any information in the `ModelInfo` call that can be used to determine if a model is a controller model. An 'is-controller' key would be ideal.

## QA steps

Step 0: Bootstrap and deploy a bundle:

- `juju bootstrap`
- `juju deploy wiki-simple`

Step 1: Verify that `show-model controller` includes `is-controller: true`. 

```plain
$ juju show-model controller
controller:
  name: admin/controller
  short-name: controller
  ...
  is-controller: true
  ...
``` 

Step 2: Verify that `show-model default` includes `is-controller: false`. 

```plain
$ juju show-model default
controller:
  name: admin/default
  short-name: default
  ...
  is-controller: false
  ...
``` 

## Documentation changes

todo

## Bug reference

lp: https://bugs.launchpad.net/juju/+bug/1803247


timClicks:
[unable to determine if a model is a controller model](https://trello.com/c/1ZOKh0bb/560-unable-to-determine-if-a-model-is-a-controller-model)

timClicks:
Looks like the [build failed](http://ci.jujucharms.com/job/make-check-juju/1954/) for something transient

>         2018-12-12 07:06:22 WARNING juju.testing mgo.go:212 failed to start mongo: mongod failed to listen on port 41040
>      mgo.go:469: mongod failed to listen on port 41040

!!build!!

apiserver/facades/client/client/client.go
```diff
@@ -526,16 +526,19 @@ func (c *Client) ModelInfo() (params.ModelInfo, error) {
...
+		DefaultSeries:  config.PreferredSeries(conf),
+		CloudTag:       names.NewCloudTag(model.Cloud()).String(),
+		CloudRegion:    model.CloudRegion(),
+		ProviderType:   conf.Type(),
+		Name:           conf.Name(),
+		Type:           string(model.Type()),
+		UUID:           model.UUID(),
+		OwnerTag:       model.Owner().String(),
+		Life:           params.Life(model.Life().String()),
+		ControllerUUID: state.ControllerTag().String(),
```
howbazaar: I do question whether or not the ControllerUUID is something that the user should really be caring about, but it probably can't hurt that much.

apiserver/facades/client/modelmanager/modelinfo_test.go
```diff
@@ -217,10 +220,12 @@ func (s *modelInfoSuite) TestModelInfo(c *gc.C) {
 		{"ModelUUID", nil},
 		{"GetBackend", []interface{}{s.st.model.cfg.UUID()}},
 		{"Model", nil},
+		{"ControllerModelUUID", nil},
 		{"AllMachines", nil},
 		{"LatestMigration", nil},
 	})
 	s.st.model.CheckCalls(c, []gitjujutesting.StubCall{
+		{"UUID", nil},
```
howbazaar: why do we have another call to UUID?

apiserver/facades/client/modelmanager/modelinfo_test.go
```diff
@@ -708,6 +713,11 @@ func (st *mockState) ControllerUUID() string {
 	return st.controllerUUID
 }
 
+func (st *mockState) IsController() bool {
+	st.MethodCall(st, "IsController")
+	return st.controllerUUID == st.model.UUID()
```
howbazaar: Ah... is this the extra call to UUID?

apiserver/facades/client/modelmanager/modelmanager.go
```diff
@@ -1001,11 +1002,14 @@ func (m *ModelManagerAPI) getModelInfo(tag names.ModelTag) (params.ModelInfo, er
 		return params.ModelInfo{}, errors.Trace(err)
 	}
 
+	isControllerModel := st.ControllerModelUUID() == model.UUID()
```
howbazaar: You should just use `st.IsController()`

cmd/juju/common/model.go
```diff
@@ -27,6 +27,7 @@ type ModelInfo struct {
 	Type           model.ModelType             `json:"model-type" yaml:"model-type"`
 	ControllerUUID string                      `json:"controller-uuid" yaml:"controller-uuid"`
 	ControllerName string                      `json:"controller-name" yaml:"controller-name"`
+	IsController   bool                        `json:"is-controller" yaml:"is-controller"`
```
howbazaar: I think both json and yaml should also be `omitempty`.

cmd/juju/controller/listmodels.go
```diff
@@ -232,6 +232,7 @@ type ModelSummary struct {
 
 	ControllerUUID     string                  `json:"controller-uuid" yaml:"controller-uuid"`
 	ControllerName     string                  `json:"controller-name" yaml:"controller-name"`
+	IsController       bool                    `json:"is-controller" yaml:"is-controller"`
```
howbazaar: Also omitempty here too.

state/modelsummaries.go
```diff
@@ -113,6 +116,7 @@ func newProcessorFromModelDocs(st *State, modelDocs []modelDoc, user names.UserT
 			Life:               doc.Life,
 			Owner:              doc.Owner,
 			ControllerUUID:     doc.ControllerUUID,
+			IsController:       doc.UUID == ctrlModelUUID,
```
howbazaar: Just use `st.IsController()`

apiserver/facades/client/client/client.go
timClicks: That's a fair point, but one that I didn't really question as it was already part of the struct. Happy to pull it out - but perhaps we should file another ticket and see if there is anyone relying on it?

apiserver/facades/client/modelmanager/modelmanager.go
timClicks: I've needed to add an `IsController` method to model as that `st` object isn't actually an instance of `state.State`

timClicks:
$$build$$

state/modelsummaries.go
```diff
@@ -43,7 +43,9 @@ type ModelSummary struct {
 	Type           ModelType
 	Owner          string
 	ControllerUUID string
-	Life           Life
+	IsController   bool
+
```
wallyworld: do we need the extra new line?

cmd/juju/controller/listmodels.go
timClicks: This has the slightly awkward consequence of hiding the field from the user-visible output [when it's `false`](https://github.com/golang/go/issues/13284).

timClicks: After chatting with @wallyworld, we're going to remove `omitempty` from the structs within `cmd/...` so that the field is present when viewed in the cli

timClicks:
([unit tests passed](http://ci.jujucharms.com/job/make-check-juju/2050/) on a commit that has been removed via rebase)

!!merge!!

timClicks:
(retrying because the [build errors](http://ci.jujucharms.com/job/make-check-juju/2052/) appear to be related to mongodb/raft/leases )

$$merge$$

PR #9551, opened 2018-12-07
scripts/copyright.bash: use grep and find options available on both ubuntu and macOS

hmlanigan:
## Description of change

scripts/copyright.bash: use grep and find options available on both ubuntu and macOS
scripts/verify.bash: move copyright earlier in the checks

## QA steps

1. make pre-check <-- succeeds
2. remove first 3 lines of state/machine.go
3. make pre-check   <-- should fail because file doesn't have Copyright


mitechie:
$$merge$$

PR #9552, opened 2018-12-08
Fix deadlocks in shutdown.

howbazaar:
While doing excessive testing of another branch, I found that there were a few different deadlocks that were possible during shutdown of machine agents. These deadlocks only occurred sometimes, and only with a bunch of preconditions, so they are quite hard to reproduce, especially to write test cases for.

The multiwatcher has a naked channel send, and the reciever of this channel may not be listening as the listener has a select that includes tome.Dying.

The state change is to allow the workers for the state objects in the pool to be stopped before we actually close the state objects. This is because if there has been a multiwatcher created on the state object, an all watcher worker is started. If this all watcher is processing an event while the pool is being closed, it may want to get a state object out of the pool for processing, but the mutex blocks this. So the pool shutdown is now much more careful about how long it holds the locks for, and also checks against a closing state pool in more places.

### DIFF
--- a/state/multiwatcher.go
+++ b/state/multiwatcher.go
@@ -304,7 +304,10 @@ func (sm *storeManager) respond() {
 
 		req.changes = changes
 		w.revno = sm.all.latestRevno
-		req.reply <- true
+		select {
+		case req.reply <- true:
+		case <-sm.tomb.Dying():
+		}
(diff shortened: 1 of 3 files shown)


howbazaar:
$$merge$$

PR #9553, opened 2018-12-10
Update prometheus

howbazaar:
The updated prometheus brings in some useful testutil functions for testing metrics.


howbazaar:
$$merge$$

PR #9557, opened 2018-12-10
LXD missing methods - 2.5beta

SimonRichardson:
## Description of change

Implement missing methods for 2.5-beta2

The following are required to allow operators to upgrade from
2.5-beta2 to 2.5rc. The reason why we need the following no-op
operations is because we no longer utilise the client to monitor
the upgrading of a charm. Instead we use the uniter to oversee
the correct upgrade path (logical in he insight), but in early betas
the client expects to be able to call and watch upgrading. In the 
following code we just still allow those methods to be called, but
we just perform a no-op.

## QA steps

deploy with 2.5-rc or equivalent
```
juju bootstrap localhost
```
using the 2.5-beta2 client attempt to upload
``` 
juju upgrade lxd-profile-alt --path ./testcharms/charm-repo/quantal/lxd-profile-alt
```

## Documentation changes

This should allow clients on beta2.5 to move to rc1

## Bug reference

N/A

### DIFF
--- a/apiserver/facades/client/application/application.go
+++ b/apiserver/facades/client/application/application.go
@@ -780,6 +780,41 @@ func (api *APIBase) SetCharmProfile(args params.ApplicationSetCharmProfile) erro
 	return application.SetCharmProfile(args.CharmURL)
 }
 
+// WatchLXDProfileUpgradeNotifications returns a watcher that fires on LXD
+// profile events.
+func (api *APIBase) WatchLXDProfileUpgradeNotifications(entity params.Entity) (params.NotifyWatchResult, error) {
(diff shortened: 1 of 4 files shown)


SimonRichardson:
https://trello.com/c/RBrBh1Wq/888-lxd-missing-methods-25

howbazaar:
If the method was introduced and removed before the RC, then I think we're OK.

Does the client make any other calls to server methods that may not be there in older controllers?
See https://bugs.launchpad.net/juju/+bug/1807666
If these calls are essential to the charm being deployed, we should error nicely. If not, we should be able to deploy on older controllers.

PR #9559, opened 2018-12-10
cmd: Fix help text does not include --debug

MalloZup:
Add info about ``--debug`` flag
## Description of change

The help text output for list **now** the --debug option along with the other options available for a command.

## QA steps
`juju help` shoud show improved help message.

``juju --debug`` should show some debug infos as also `juju add-cloud --debug` etc. (but this commands should already working).

## Documentation changes
`Cmd` juju shows now the debug option.

## Bug reference
Fix https://bugs.launchpad.net/juju/+bug/1628151

## add info:

I'm open if you have a better layout/description for the cmd helper output

### DIFF
--- a/cmd/juju/commands/main.go
+++ b/cmd/juju/commands/main.go
@@ -111,7 +111,13 @@ Example help commands:
     `[1:] + "`juju help`" + `          This help page
     ` + "`juju help commands`" + ` Lists all commands
     ` + "`juju help deploy`" + `   Shows help for command 'deploy'
-`
+
+Debug mode:
+  --debug flag enable debug mode for each command.
+
+  example:
+    ` + "`juju add-cloud --debug`" + ` execute add-cloud in debug mode
+ `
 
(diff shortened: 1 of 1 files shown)


howbazaar:
I feel that what we really want is a section in the help that lists all the inherited flags from the supercommand.
  --debug
  --show-log
  --logging-config
  --verbose
  --quiet

There is a reason we don't have these on every single command, but if we could have them grouped, it would be more palatable. Although this probably needs to be fixed at the github.com/juju/cmd package level.

MalloZup:
@anastasiamac @howbazaar thx for feedback.  yop i was thinking the same but i wanted to show up in a PR and waiting for feedback  because i didn't know all the other options, for a generic approach.  :)
Yop the possible fix should be on the pkg level `cmd` i will adress once in have some free time. 

thx :sunflower:

anastasiamac:
@howbazaar,
Agreed, I feel like we should have all these options available through 'help' on every command. I am happy to group them after all the command options, although I do not see why we cannot just mix supercommand options among command-specific options and order all alphabetically. I'll add this to the bug.

anastasiamac:
@MalloZup,
Thank you for bringing it to our attention. I'll close this PR for now.

PR #9561, opened 2018-12-11
juju client no longer calls SetCharmProfile on upgrade-charm.

howbazaar:
The setting of the charm profile is now done on the apiserver when SetCharm is called.

This means older clients will just do the right thing, and newer clients will work with older controllers.

## QA steps

- bootstrap both 2.4.7 controller and 2.5-rc1 controller
- deploy test charm to both: ./testcharms/charm-repo/quantal/lxd-profile-alt
- upgrade-charm --path to both

## Bug reference

https://bugs.launchpad.net/juju/+bug/1807666

### DIFF
--- a/apiserver/facades/client/application/application.go
+++ b/apiserver/facades/client/application/application.go
@@ -739,6 +739,10 @@ func (api *APIBase) SetCharm(args params.ApplicationSetCharm) error {
 	if err != nil {
 		return errors.Trace(err)
 	}
+	if err := application.SetCharmProfile(args.CharmURL); err != nil {
+		return errors.Annotatef(err, "unable to set charm profile")
+	}
+
 	channel := csparams.Channel(args.Channel)
 	return api.applicationSetCharm(
 		args.ApplicationName,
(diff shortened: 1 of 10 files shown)


apiserver/facades/client/application/application.go
```diff
@@ -755,8 +759,11 @@ func (api *APIBase) SetCharm(args params.ApplicationSetCharm) error {
 	)
 }
 
-// SetCharmProfile a new charm's url on deployed machines for changing the profile used
-// on those machine.
+// SetCharmProfile a new charm's url on deployed machines for changing the
```
wallyworld: typo

apiserver/facades/client/application/application.go
```diff
@@ -755,8 +759,11 @@ func (api *APIBase) SetCharm(args params.ApplicationSetCharm) error {
...
+// profile used on those machine.
+//
+// TODO: consider removing this public method as it is now done as part of
```
wallyworld: Can we remove now? Harder to remove later after release

state/application.go
```diff
@@ -901,6 +901,10 @@ func (a *Application) DeployedMachines() ([]*Machine, error) {
 		// whether principal or subordinate.
 		id, err := u.AssignedMachineId()
 		if err != nil {
+			if errors.IsNotAssigned(err) {
```
wallyworld: strictly speaking there should be a test for this

apiserver/facades/client/application/application.go
howbazaar: removed.

apiserver/facades/client/application/application.go
howbazaar: yes, removing now

state/application.go
howbazaar: tests added

wallyworld:
$$merge$$

PR #9570, opened 2018-12-12
Fix charm revision updater test.

howbazaar:
The headers sent to identify new charm revisions includes the architectures that the charms are deployed with. The test was asserting against the host architecture running the tests for this, which is wrong.


babbageclunk:
$$merge$$

PR #9571, opened 2018-12-12
Fix migration failure case.

howbazaar:
The permission that was being set in the python script hasn't been valid since pre 2.0. It is possible that recent changes removed the backwards compatibility. The CI test should use the right name.


babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #9576, opened 2018-12-13
Prevent error when using Legacy Leases

SimonRichardson:
## Description of change

The following code ensures that we don't fail when attempting to
use the old legacy leases.

## QA steps

```
juju bootstrap localhost test --debug --build-agent=true --config features="[legacy-leases]" --bootstrap-series="xenial"
juju upgrade-series 0 prepare bionic -y
```

## Documentation changes

This is a bug fix, not functionality has changed.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1808324

### DIFF
--- a/worker/upgradeseries/worker.go
+++ b/worker/upgradeseries/worker.go
@@ -12,6 +12,7 @@ import (
 	"gopkg.in/juju/worker.v1"
 	"gopkg.in/juju/worker.v1/catacomb"
 
+	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/core/model"
 	"github.com/juju/juju/service"
 	"github.com/juju/os/series"
@@ -356,9 +357,9 @@ func (w *upgradeSeriesWorker) pinLeaders() (err error) {
 		// utilising the legacy leases store, then we should display the warning
 		// in the log and return out. Unpinning leaders should be safe as that
 		// should be considered a no-op
-		if errors.IsNotImplemented(errors.Cause(err)) {
-			w.logger.Warningf("failed to pin machine applications %+v", err)
-			return
+		if params.IsCodeNotImplemented(err) {
+			w.logger.Infof("failed to pin machine applications %+v", err)
+			return nil
 		}
 		return errors.Trace(err)
 	}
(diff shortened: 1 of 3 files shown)


worker/upgradeseries/worker.go
```diff
@@ -352,6 +352,14 @@ func (w *upgradeSeriesWorker) pinLeaders() (err error) {
 
 	results, err := w.PinMachineApplications()
 	if err != nil {
+		// If pin machine applications method return not implemented because it's
+		// utilising the legacy leases store, then we should display the warning
+		// in the log and return out. Unpinning leaders should be safe as that
+		// should be considered a no-op
+		if errors.IsNotImplemented(errors.Cause(err)) {
```
jameinel: IsNotImplemented already handles the Cause portion, so you don't have to do that here.

worker/upgradeseries/worker.go
```diff
@@ -352,6 +352,14 @@ func (w *upgradeSeriesWorker) pinLeaders() (err error) {
...
+			w.logger.Warningf("failed to pin machine applications %+v", err)
```
jameinel: Thinking about this, maybe this should be Infof, as it isn't something we want someone to file a bug about/investigate more thoroughly.

Did you see the actual error under this circumstance? do we trust that it will be something a human will understand?

worker/upgradeseries/worker.go
SimonRichardson: So during testing, we don't actually get a NotImplemented error, so this doesn't currently work. Is there anyway to check if the error is not implemented after it's crossed the API boundary?

SimonRichardson: It's ok I found out how to do it :smile:

worker/upgradeseries/worker.go
SimonRichardson: I've moved it to Infof.

> do we trust that it will be something a human will understand?

You can see the error message here:

> machine-0: 13:01:20 INFO juju.worker.upgradeseries failed to pin machine applications  unable to get leadership pinner; pinning is not available with the legacy lease manager not implemented

jameinel: I'd rather we play with it a bit to make a bit more sense:

failed to pin machine applications, with legacy lease manager leadership pinning is not implemented

Thoughts?

SimonRichardson:
!!build!!

Lease failure on a mongo box, not entirely sure why.

SimonRichardson:
!!build!!

SimonRichardson:
!!build!!

PR #9577, opened 2018-12-13
Fix other addmodel sites.

howbazaar:
Python has single quoted strings as well as double. Test was failing because code was checking internal frozenset prior to making call.


mitechie:
TY $$merge$$

PR #9583, opened 2018-12-14
upgrade caas testing charms and pull from charmstore because charmsto…

ycliuhw:
## Description of change

upgrade caas testing charms and pull from charmstore because charmstore support caas charms now; delete local charms;

## QA steps

nil

## Documentation changes

nil

## Bug reference

nil


wallyworld:
$$merge$$

PR #9584, opened 2018-12-14
Charm profile resolver tests

SimonRichardson:
## Description of change

Add tests for the resolver around charm profile

The following commit just adds missing tests to the resolver, so
that we can be a bit more confident around changing this code in
the future.

## QA steps

These are additive to what already exists.

Relates to: https://github.com/juju/juju/pull/9578

PR #9585, opened 2018-12-14
Un-place missing upgrade charm profile data

SimonRichardson:
## Description of change

This is a revert of 38b5255759be8578010c04a132cb23224f8c2789 with
added notes to prevent people from making the same mistake that I
did.

Although the code is generalised for other users to use, if nobody
else is using it, then we should delete it for the sake of our
sanity.

This is analogous to https://github.com/juju/juju/pull/9584 so that we 
have more tests around this feature.

## QA steps

N/A


hmlanigan:
$$merge$$

PR #9586, opened 2018-12-17
Fix model destroy tests

wallyworld:
## Description of change

After destroying a model, the current model reference is now retained so we need to run a status check to see that it is gone.

## QA steps

 ./assess_destroy_model.py ian $GOPATH/bin/juju ~/jujutest/artifacts

## Bug reference

https://bugs.launchpad.net/juju/+bug/1808250


mitechie:
TY +1

PR #9589, opened 2018-12-18
Add display name to instance data

timClicks:
This commit makes it possible for provider code to pass through a display name when instances are created.

Implementations of `InstanceBroker` can add `DisplayName` `string` to their `StartInstanceResult`
instance that's returned with `StartInstance`. That string is propagated through to stat as part of `instanceData` and surfaced up when end-users call `juju status`.

The PR touches many files as it has required some reworking of a few of the provisioning mechanics through each stage of the API. Here are the major public API changes:

* The [`Machine.SetProvisioned`](https://github.com/timClicks/juju/blob/f4b75139fcfef12af7a0b8a6e13d96b216de92fe/state/machine.go#L1436) method takes a new parameter, `displayName string`
* the `MachineProvisioner` interface has been modified. [`MachineProvisioner.SetInstanceInfo`](https://github.com/timClicks/juju/blob/f4b75139fcfef12af7a0b8a6e13d96b216de92fe/api/provisioner/machine.go#L86) now includes a `displayName string` argument. 

**Backwards compatibility**

_Provider code_

No changes are required to code that does not wish to make use of instance display names. The default behaviour is to revert to the instance ID. Current provider code remains unchanged.

_API internals_

Adding a new field may require bumping the façade number. This hasn't yet been done within this PR. 

**Future work**

* Add the ability to create a custom display name via the CLI `juju set-display-name`
* Extend the functionality into `state/addmachine.go`. `addmachine.go` does not alter `instanceData`. Provisioners that make use of this code to provision machines—generally the manual provisioners—won't be able to set an instance's display name without more work.  
 
## QA steps

Bootstrap into Oracle provider, add a machine and see that the `Inst id` column is only 6 significant "figures":

```plain
$ juju bootstrap --config compartment-id=ocid1... oracle tsm-phx
...

$ juju add-machine -n 2
created machine 0
created machine 1

$ juju status
Model    Controller  Cloud/Region         Version      SLA          Timestamp
default  tsm-phx     oracle/us-phoenix-1  2.6-beta1.1  unsupported  21:07:57+13:00

Machine  State    DNS              Inst id    Series  AZ  Message
0        pending  129.146.175.166  ...pbcehq  bionic      running
1        pending  129.146.173.212  ...6qhlsq  bionic      running

$ juju machines --format=yaml
model: default
machines:
  "0":
    juju-status:
      current: started
      since: 10 Jan 2019 21:10:07+13:00
      version: 2.6-beta1.1
    dns-name: 129.146.175.166
    ip-addresses:
    - 129.146.175.166
    - 10.0.1.2
    instance-id: ocid1.instance.oc1.phx.abyhqljtnhgn33m35rurkociav3aoyrgdqrwgfngoungqqjwzgoy6mpbcehq
    display-name: '...pbcehq'
    machine-status:
      current: running
      message: running
      since: 10 Jan 2019 21:07:52+13:00
    series: bionic
    network-interfaces:
      ens3:
        ip-addresses:
        - 10.0.1.2
        mac-address: 00:00:17:02:cb:3a
        gateway: 10.0.1.1
        is-up: true
    hardware: arch=amd64
...
```

## Documentation changes

The output for the MAAS and OCI providers will look different _when interacting with newly provisioned machines_. Machines that have already been provisioned will still have the same display name.

This might create an awkward situation within `juju show-status` where some instances will have long, ugly instance ids, whereas others have a shorter one.

## Bug reference

* https://bugs.launchpad.net/juju/+bug/1792210

### DIFF
--- a/api/base/types.go
+++ b/api/base/types.go
@@ -41,13 +41,14 @@ type ModelStatus struct {
 
 // Machine holds information about a machine in a juju model.
 type Machine struct {
-	Id         string
-	InstanceId string
-	HasVote    bool
-	WantsVote  bool
-	Status     string
-	Message    string
-	Hardware   *instance.HardwareCharacteristics
+	Id          string
+	InstanceId  string
+	DisplayName string
+	HasVote     bool
+	WantsVote   bool
+	Status      string
+	Message     string
+	Hardware    *instance.HardwareCharacteristics
 }
 
 // ModelInfo holds information about a model.
--- a/api/common/modelstatus.go
+++ b/api/common/modelstatus.go
@@ -102,12 +102,13 @@ func constructModelStatus(model names.ModelTag, owner names.UserTag, r params.Mo
 			result.CoreCount += int(*mm.Hardware.Cores)
 		}
 		result.Machines[j] = base.Machine{
-			Id:         mm.Id,
-			InstanceId: mm.InstanceId,
-			HasVote:    mm.HasVote,
-			WantsVote:  mm.WantsVote,
-			Status:     mm.Status,
-			Message:    mm.Message,
+			Id:          mm.Id,
+			InstanceId:  mm.InstanceId,
+			DisplayName: mm.DisplayName,
+			HasVote:     mm.HasVote,
+			WantsVote:   mm.WantsVote,
+			Status:      mm.Status,
+			Message:     mm.Message,
 		}
 	}
 	return result
--- a/api/modelmanager/modelmanager.go
+++ b/api/modelmanager/modelmanager.go
@@ -134,11 +134,12 @@ func convertParamsModelInfo(modelInfo params.ModelInfo) (base.ModelInfo, error) 
 	result.Machines = make([]base.Machine, len(modelInfo.Machines))
 	for i, m := range modelInfo.Machines {
 		machine := base.Machine{
-			Id:         m.Id,
-			InstanceId: m.InstanceId,
-			HasVote:    m.HasVote,
-			WantsVote:  m.WantsVote,
-			Status:     m.Status,
+			Id:          m.Id,
+			InstanceId:  m.InstanceId,
+			DisplayName: m.DisplayName,
+			HasVote:     m.HasVote,
+			WantsVote:   m.WantsVote,
+			Status:      m.Status,
 		}
 		if m.Hardware != nil {
 			machine.Hardware = &instance.HardwareCharacteristics{
--- a/api/provisioner/machine.go
+++ b/api/provisioner/machine.go
@@ -84,7 +84,7 @@ type MachineProvisioner interface {
 	// SetInstanceInfo sets the provider specific instance id, nonce, metadata,
 	// network config for this machine. Once set, the instance id cannot be changed.
 	SetInstanceInfo(
-		id instance.Id, nonce string, characteristics *instance.HardwareCharacteristics,
+		id instance.Id, displayName string, nonce string, characteristics *instance.HardwareCharacteristics,
 		networkConfig []params.NetworkConfig, volumes []params.Volume,
 		volumeAttachments map[string]params.VolumeAttachmentInfo, charmProfiles []string,
 	) error
@@ -360,7 +360,7 @@ func (m *Machine) DistributionGroup() ([]instance.Id, error) {
 
 // SetInstanceInfo implements MachineProvisioner.SetInstanceInfo.
 func (m *Machine) SetInstanceInfo(
-	id instance.Id, nonce string, characteristics *instance.HardwareCharacteristics,
+	id instance.Id, displayName string, nonce string, characteristics *instance.HardwareCharacteristics,
 	networkConfig []params.NetworkConfig, volumes []params.Volume,
 	volumeAttachments map[string]params.VolumeAttachmentInfo, charmProfiles []string,
 ) error {
@@ -369,6 +369,7 @@ func (m *Machine) SetInstanceInfo(
 		Machines: []params.InstanceInfo{{
 			Tag:               m.tag.String(),
 			InstanceId:        id,
+			DisplayName:       displayName,
 			Nonce:             nonce,
 			Characteristics:   characteristics,
 			Volumes:           volumes,
--- a/apiserver/common/machine.go
+++ b/apiserver/common/machine.go
@@ -57,6 +57,7 @@ func (st *stateShim) Machine(id string) (Machine, error) {
 type Machine interface {
 	Id() string
 	InstanceId() (instance.Id, error)
+	InstanceNames() (instance.Id, string, error)
 	WantsVote() bool
 	HasVote() bool
 	Status() (status.StatusInfo, error)
@@ -122,10 +123,11 @@ func ModelMachineInfo(st ModelManagerBackend) (machineInfo []params.ModelMachine
 			Status:    status,
 			Message:   statusInfo.Message,
 		}
-		instId, err := m.InstanceId()
+		instId, displayName, err := m.InstanceNames()
 		switch {
 		case err == nil:
 			mInfo.InstanceId = string(instId)
+			mInfo.DisplayName = displayName
 		case errors.IsNotProvisioned(err):
 			// ok, but no instance ID to get.
 		default:
--- a/apiserver/common/machine_test.go
+++ b/apiserver/common/machine_test.go
@@ -105,7 +105,8 @@ func (s *machineSuite) TestMachineHardwareInfo(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 	c.Assert(info, jc.DeepEquals, []params.ModelMachineInfo{
 		{
-			Id: "1",
+			Id:          "1",
+			DisplayName: "",
 			Hardware: &params.MachineHardware{
 				Arch:     &amd64,
 				Mem:      &gig,
@@ -113,7 +114,8 @@ func (s *machineSuite) TestMachineHardwareInfo(c *gc.C) {
 				CpuPower: &one,
 			},
 		}, {
-			Id: "2",
+			Id:          "2",
+			DisplayName: "",
 		},
 	})
 }
@@ -121,7 +123,21 @@ func (s *machineSuite) TestMachineHardwareInfo(c *gc.C) {
 func (s *machineSuite) TestMachineInstanceInfo(c *gc.C) {
 	st := mockState{
 		machines: map[string]*mockMachine{
-			"1": {id: "1", instId: instance.Id("123"), status: status.Down, hasVote: true, wantsVote: true},
+			"1": {
+				id:        "1",
+				instId:    "123",
+				status:    status.Down,
+				hasVote:   true,
+				wantsVote: true,
+			},
+			"2": {
+				id:          "2",
+				instId:      "456",
+				displayName: "two",
+				status:      status.Allocating,
+				hasVote:     false,
+				wantsVote:   true,
+			},
 		},
 	}
 	info, err := common.ModelMachineInfo(&st)
@@ -133,6 +149,68 @@ func (s *machineSuite) TestMachineInstanceInfo(c *gc.C) {
 			Status:     "down",
 			HasVote:    true,
 			WantsVote:  true,
+		},
+		{
+			Id:          "2",
+			InstanceId:  "456",
+			DisplayName: "two",
+			Status:      "allocating",
+			HasVote:     false,
+			WantsVote:   true,
+		},
+	})
+}
+
+func (s *machineSuite) TestMachineInstanceInfoWithEmptyDisplayName(c *gc.C) {
+	st := mockState{
+		machines: map[string]*mockMachine{
+			"1": {
+				id:          "1",
+				instId:      "123",
+				displayName: "",
+				status:      status.Down,
+				hasVote:     true,
+				wantsVote:   true,
+			},
+		},
+	}
+	info, err := common.ModelMachineInfo(&st)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(info, jc.DeepEquals, []params.ModelMachineInfo{
+		{
+			Id:          "1",
+			InstanceId:  "123",
+			DisplayName: "",
+			Status:      "down",
+			HasVote:     true,
+			WantsVote:   true,
+		},
+	})
+}
+
+func (s *machineSuite) TestMachineInstanceInfoWithSetDisplayName(c *gc.C) {
+	st := mockState{
+		machines: map[string]*mockMachine{
+			"1": {
+				id:          "1",
+				instId:      "123",
+				displayName: "snowflake",
+				status:      status.Down,
+				hasVote:     true,
+				wantsVote:   true,
+			},
+		},
+	}
+	info, err := common.ModelMachineInfo(&st)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(info, jc.DeepEquals, []params.ModelMachineInfo{
+		{
+			Id:          "1",
+			InstanceId:  "123",
+			DisplayName: "snowflake",
+			Status:      "down",
+			HasVote:     true,
+			WantsVote:   true,
 		},
 	})
 }
@@ -169,6 +247,7 @@ type mockMachine struct {
 	containerType      instance.ContainerType
 	hw                 *instance.HardwareCharacteristics
 	instId             instance.Id
+	displayName        string
 	hasVote, wantsVote bool
 	status             status.Status
 	statusErr          error
@@ -190,6 +269,11 @@ func (m *mockMachine) Life() state.Life {
 
 func (m *mockMachine) InstanceId() (instance.Id, error) {
 	return m.instId, nil
+}
+
+func (m *mockMachine) InstanceNames() (instance.Id, string, error) {
+	instId, err := m.InstanceId()
+	return instId, m.displayName, err
 }
 
 func (m *mockMachine) WantsVote() bool {
--- a/apiserver/common/modelstatus_test.go
+++ b/apiserver/common/modelstatus_test.go
@@ -134,6 +134,7 @@ func (s *modelStatusSuite) TestModelStatus(c *gc.C) {
 		Jobs:            []state.MachineJob{state.JobManageModel},
 		Characteristics: &instance.HardwareCharacteristics{CpuCores: &eight},
 		InstanceId:      "id-4",
+		DisplayName:     "snowflake",
 		Volumes: []state.HostVolumeParams{{
 			Volume: state.VolumeParams{
 				Pool: "modelscoped",
@@ -193,7 +194,7 @@ func (s *modelStatusSuite) TestModelStatus(c *gc.C) {
 			OwnerTag:           s.Owner.String(),
 			Life:               params.Alive,
 			Machines: []params.ModelMachineInfo{
-				{Id: "0", Hardware: &params.MachineHardware{Cores: &eight}, InstanceId: "id-4", Status: "pending", WantsVote: true},
+				{Id: "0", Hardware: &params.MachineHardware{Cores: &eight}, InstanceId: "id-4", DisplayName: "snowflake", Status: "pending", WantsVote: true},
 				{Id: "1", Hardware: stdHw, InstanceId: "id-5", Status: "pending"},
 			},
 			Volumes: []params.ModelVolumeInfo{{
--- a/apiserver/facades/controller/charmrevisionupdater/testing/suite.go
+++ b/apiserver/facades/controller/charmrevisionupdater/testing/suite.go
@@ -91,7 +91,7 @@ func (s *CharmSuite) AddMachine(c *gc.C, machineId string, job state.MachineJob)
 	controllerCfg, err := s.jcSuite.State.ControllerConfig()
 	c.Assert(err, jc.ErrorIsNil)
 	inst, hc := jujutesting.AssertStartInstanceWithConstraints(c, s.jcSuite.Environ, s.jcSuite.ProviderCallContext, controllerCfg.ControllerUUID(), m.Id(), cons)
-	err = m.SetProvisioned(inst.Id(), "fake_nonce", hc)
+	err = m.SetProvisioned(inst.Id(), "", "fake_nonce", hc)
 	c.Assert(err, jc.ErrorIsNil)
 }
 
--- a/apiserver/params/internal.go
+++ b/apiserver/params/internal.go
@@ -384,11 +384,21 @@ type LifeResults struct {
 	Results []LifeResult `json:"results"`
 }
 
-// InstanceInfo holds a machine tag, provider-specific instance id, a nonce, and
-// network config.
+// InstanceInfo holds information about an instance. Instances are
+// typically virtual machines hosted by a cloud provider but may also
+// be a container.
+//
+// The InstanceInfo struct contains three categories of information:
+//  - interal data, as the machine's tag and the tags of any attached
+//    storage volumes
+//  - naming and other provider-specific information, including the
+//    instance id and display name
+//  - configuration information, including its attached storage volumes,
+//    charm profiles and networking
 type InstanceInfo struct {
 	Tag             string                            `json:"tag"`
 	InstanceId      instance.Id                       `json:"instance-id"`
+	DisplayName     string                            `json:"display-name"`
 	Nonce           string                            `json:"nonce"`
 	Characteristics *instance.HardwareCharacteristics `json:"characteristics"`
 	Volumes         []Volume                          `json:"volumes"`
--- a/apiserver/params/model.go
+++ b/apiserver/params/model.go
@@ -283,13 +283,14 @@ type ModelInfoListResults struct {
 
 // ModelMachineInfo holds information about a machine in a model.
 type ModelMachineInfo struct {
-	Id         string           `json:"id"`
-	Hardware   *MachineHardware `json:"hardware,omitempty"`
-	InstanceId string           `json:"instance-id,omitempty"`
-	Status     string           `json:"status,omitempty"`
-	Message    string           `json:"message,omitempty"`
-	HasVote    bool             `json:"has-vote,omitempty"`
-	WantsVote  bool             `json:"wants-vote,omitempty"`
+	Id          string           `json:"id"`
+	Hardware    *MachineHardware `json:"hardware,omitempty"`
+	InstanceId  string           `json:"instance-id,omitempty"`
+	DisplayName string           `json:"display-name,omitempty"`
+	Status      string           `json:"status,omitempty"`
+	Message     string           `json:"message,omitempty"`
+	HasVote     bool             `json:"has-vote,omitempty"`
+	WantsVote   bool             `json:"wants-vote,omitempty"`
 }
 
 // MachineHardware holds information about a machine's hardware characteristics.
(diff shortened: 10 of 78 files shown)


environs/instances/instance.go
```diff
@@ -14,6 +14,10 @@ type Instance interface {
 	// Id returns a provider-generated identifier for the Instance.
 	Id() instance.Id
 
+	// DisplayName returns a provider-generated, user-friendly
+	// instance name for display purposes.
+	DisplayName() instance.DisplayName
```
timClicks: I've added `DisplayName` to the instance interface. Although that means that every provider needs to implement it, we get the benefit of type safety. Let me know if this is overkill.

state/machine.go
```diff
@@ -1241,6 +1242,25 @@ func (m *Machine) InstanceId() (instance.Id, error) {
 	return instData.InstanceId, err
 }
 
+// DisplayName returns the provider-specific display name for this
+// machine, falling back to the instance id if it does not exist.
+func (m *Machine) DisplayName() (instance.DisplayName, error) {
```
timClicks: Perhaps a better way to do this would be to simply return `""`, rather than jumping through hoops to provide a fallback option?

state/machine.go
```diff
@@ -1419,12 +1439,13 @@ func (m *Machine) DesiredSpaces() (set.Strings, error) {
...
 func (m *Machine) SetProvisioned(
 	id instance.Id,
+	displayName instance.DisplayName,
 	nonce string,
 	characteristics *instance.HardwareCharacteristics,
 ) (err error) {
 	defer errors.DeferredAnnotatef(&err, "cannot set instance data for machine %q", m)
 
-	if id == "" || nonce == "" {
+	if id == instance.UnknownId || nonce == "" {
```
timClicks: Kind of unnecessary, but I added it in as the constant has been defined within `core/instance` and using it means IDEs can make use of it.

state/machine.go
```diff
@@ -1442,17 +1463,18 @@ func (m *Machine) SetProvisioned(
...
-		Mem:        characteristics.Mem,
-		RootDisk:   characteristics.RootDisk,
-		CpuCores:   characteristics.CpuCores,
-		CpuPower:   characteristics.CpuPower,
-		Tags:       characteristics.Tags,
-		AvailZone:  characteristics.AvailabilityZone,
+		DocID:       m.doc.DocID,
+		MachineId:   m.doc.Id,
+		InstanceId:  id,
+		DisplayName: &displayName,
```
timClicks: Adding `DisplayName` is the substantive change here. The rest is gofmt noise.

api/base/types.go
```diff
@@ -41,13 +41,14 @@ type ModelStatus struct {
...
-	Id         string
-	InstanceId string
-	HasVote    bool
-	WantsVote  bool
-	Status     string
-	Message    string
-	Hardware   *instance.HardwareCharacteristics
+	Id          string
+	InstanceId  instance.Id
+	DisplayName instance.DisplayName
```
wallyworld: no need to make this type - display name can be a string

apiserver/common/machine.go
```diff
@@ -125,12 +126,21 @@ func ModelMachineInfo(st ModelManagerBackend) (machineInfo []params.ModelMachine
...
 		switch {
 		case err == nil:
-			mInfo.InstanceId = string(instId)
+			mInfo.InstanceId = instId
 		case errors.IsNotProvisioned(err):
 			// ok, but no instance ID to get.
 		default:
 			return nil, errors.Trace(err)
 		}
+		displayName, err := m.DisplayName()
```
wallyworld: No need for separate block of code here - just get the instance id and display name together above

apiserver/params/model.go
```diff
@@ -6,6 +6,7 @@ package params
 import (
 	"time"
 
+	"github.com/juju/juju/core/instance"
```
wallyworld: grouping

cmd/juju/controller/showcontroller.go
```diff
@@ -273,7 +274,7 @@ type MachineDetails struct {
 	ID string `yaml:"id,omitempty" json:"id,omitempty"`
 
 	// InstanceID holds the cloud instance id of the machine.
-	InstanceID string `yaml:"instance-id,omitempty" json:"instance-id,omitempty"`
+	InstanceID instance.Id `yaml:"instance-id,omitempty" json:"instance-id,omitempty"`
```
wallyworld: This is purely for output purposes so no need for anything other than a string

core/instance/instance.go
```diff
@@ -11,11 +11,17 @@ import (
 // instance (physical or virtual machine allocated in the provider).
 type Id string
 
+// DisplayName is used for user-friendly reporting of the instance's
+// status. The DisplayName is not intended to be used internally within
```
wallyworld: s/status/name

environs/instances/instance.go
wallyworld: Yeah, not needed. DisplayName is just a string

state/machine.go
wallyworld: You could consider a new method which returns both instance.Id and display name in a struct. I think that could simplify things a fair bit.

environs/instances/instance.go
timClicks: Just to clarify. Keep the method, but use string as the type?

cmd/juju/controller/showcontroller.go
timClicks: Sorted.

api/common/modelstatus.go
```diff
@@ -103,7 +103,7 @@ func constructModelStatus(model names.ModelTag, owner names.UserTag, r params.Mo
 		}
 		result.Machines[j] = base.Machine{
 			Id:         mm.Id,
-			InstanceId: mm.InstanceId,
+			InstanceId: string(mm.InstanceId),
```
wallyworld: There should be no change to InstanceId handling.
But we want to copy DisplayName here right?

api/modelmanager/modelmanager.go
```diff
@@ -134,7 +134,7 @@ func convertParamsModelInfo(modelInfo params.ModelInfo) (base.ModelInfo, error)
 	for i, m := range modelInfo.Machines {
 		machine := base.Machine{
 			Id:         m.Id,
-			InstanceId: m.InstanceId,
+			InstanceId: string(m.InstanceId),
```
wallyworld: ditto

api/provisioner/machine.go
```diff
@@ -84,7 +84,8 @@ type MachineProvisioner interface {
 	// SetInstanceInfo sets the provider specific instance id, nonce, metadata,
 	// network config for this machine. Once set, the instance id cannot be changed.
 	SetInstanceInfo(
-		id instance.Id, nonce string, characteristics *instance.HardwareCharacteristics,
+		id instance.Id, displayName instance.DisplayName, nonce string,
```
wallyworld: displayName string

api/provisioner/machine.go
```diff
@@ -360,7 +361,7 @@ func (m *Machine) DistributionGroup() ([]instance.Id, error) {
 
 // SetInstanceInfo implements MachineProvisioner.SetInstanceInfo.
 func (m *Machine) SetInstanceInfo(
-	id instance.Id, nonce string, characteristics *instance.HardwareCharacteristics,
+	id instance.Id, displayName instance.DisplayName, nonce string, characteristics *instance.HardwareCharacteristics,
```
wallyworld: ditto

apiserver/facades/controller/charmrevisionupdater/testing/suite.go
```diff
@@ -84,14 +84,17 @@ func (s *CharmSuite) AddMachine(c *gc.C, machineId string, job state.MachineJob)
...
+
 	c.Assert(err, jc.ErrorIsNil)
 	c.Assert(m.Id(), gc.Equals, machineId)
 	cons, err := m.Constraints()
 	c.Assert(err, jc.ErrorIsNil)
 	controllerCfg, err := s.jcSuite.State.ControllerConfig()
 	c.Assert(err, jc.ErrorIsNil)
 	inst, hc := jujutesting.AssertStartInstanceWithConstraints(c, s.jcSuite.Environ, s.jcSuite.ProviderCallContext, controllerCfg.ControllerUUID(), m.Id(), cons)
-	err = m.SetProvisioned(inst.Id(), "fake_nonce", hc)
+	displayName, err := m.DisplayName()
```
wallyworld: This isn't right - you are setting the displayname of a machine to the displayname of a machine.
The display name comes from the act of starting an instance - so the AssertStartInstanceWithConstraints() helper above needs to return  an inst implementation that has the display name added

provider/maas/instance.go
```diff
@@ -40,17 +40,14 @@ func (mi *maas1Instance) String() string {
...
 
 func (mi *maas1Instance) Id() instance.Id {
-	return maasObjectId(mi.maasObject)
-}
-
-func maasObjectId(maasObject *gomaasapi.MAASObject) instance.Id {
-	// Use the node's 'resource_uri' value.
+	// Note: URI excludes the hostname
 	return instance.Id(maasObject.URI().String())
 }
```
timClicks: Decided to remove this function, as it's only used in a single place.

timClicks:
[user friendly instance display names](https://trello.com/c/upHDMfXI/574-user-friendly-instance-display-names)

apiserver/facades/controller/charmrevisionupdater/testing/suite.go
```diff
@@ -92,6 +92,9 @@ func (s *CharmSuite) AddMachine(c *gc.C, machineId string, job state.MachineJob)
 	c.Assert(err, jc.ErrorIsNil)
 	inst, hc := jujutesting.AssertStartInstanceWithConstraints(c, s.jcSuite.Environ, s.jcSuite.ProviderCallContext, controllerCfg.ControllerUUID(), m.Id(), cons)
 	err = m.SetProvisioned(inst.Id(), "fake_nonce", hc)
+	instId, err := m.InstanceId()
```
wallyworld: We don't need these lines I don't think. We just need to ensure the machine is provisioned hence the call to SetProvisioned()

apiserver/params/internal.go
```diff
@@ -384,11 +384,22 @@ type LifeResults struct {
...
 }
 
-// InstanceInfo holds a machine tag, provider-specific instance id, a nonce, and
-// network config.
+// InstanceInfo holds information about an instance. Instances are
+// typically virtual machines hosted by a cloud provider but may also
+// be a container.
+//
+// The InstanceInfo struct contains three categories of information:
+//
```
wallyworld: no need for blank line here

environs/broker.go
```diff
@@ -105,6 +105,11 @@ type StartInstanceParams struct {
 // StartInstanceResult holds the result of an
 // InstanceBroker.StartInstance method call.
 type StartInstanceResult struct {
+	// DisplayName is an optional human-readable string that's used
+	// for display purposes only. When empty, the output defaults to
```
wallyworld: No need for the 2nd sentence - it's up to individual renderers how this is used and not really the concern of this layer

juju/testing/conn.go
```diff
@@ -309,13 +310,17 @@ func (s *JujuConnSuite) OpenAPIAsNewMachine(c *gc.C, jobs ...state.MachineJob) (
 	if len(jobs) == 0 {
 		jobs = []state.MachineJob{state.JobHostUnits}
 	}
+
+	displayName := "foo"
+	instId := instance.Id(displayName)
```
wallyworld: We should have display name and inst id different to each other we we are going to set display name. Otherwise leave display name empty. Is there need to set display name here at all? Most providers won't set it. We can set it when needed in those tests that do care.

provider/maas/environ.go
```diff
@@ -1095,6 +1095,7 @@ func (env *maasEnviron) StartInstance(
 	}
 
 	return &environs.StartInstanceResult{
+		DisplayName:       hostname,
```
wallyworld: need an updated test

provider/oci/environ.go
```diff
@@ -637,6 +637,17 @@ func (e *Environ) StartInstance(ctx envcontext.ProviderCallContext, args environ
 	}
 	logger.Infof("started instance %q", *machineId)
 
+	var displayName string
+	if machineId == nil {
```
wallyworld: we know that when we get here, machineId is not nil so no need for a check

provider/oci/environ.go
```diff
@@ -645,8 +656,9 @@ func (e *Environ) StartInstance(ctx envcontext.ProviderCallContext, args environ
 	}
 
 	result := &environs.StartInstanceResult{
-		Instance: instance,
-		Hardware: instance.hardwareCharacteristics(),
+		DisplayName: displayName,
```
wallyworld: also needs a test

state/allwatcher_internal_test.go
```diff
@@ -973,7 +972,7 @@ func (s *allWatcherStateSuite) TestStateWatcher(c *gc.C) {
 		Arch: &arch,
 		Mem:  &mem,
 	}
-	err = m0.SetProvisioned("i-0", "bootstrap_nonce", hc)
+	err = m0.SetProvisioned(instance.Id("i-0"), "bootstrap_nonce", hc)
```
wallyworld: these casts in the SetProvisioned() calls should not be needed as this code is not being changed

state/machine.go
```diff
@@ -1241,6 +1242,27 @@ func (m *Machine) InstanceId() (instance.Id, error) {
 	return instData.InstanceId, err
 }
 
+// Names returns both the provider's instance id and a user-friendly
```
wallyworld: InstanceNames

state/machine.go
```diff
@@ -1241,6 +1242,27 @@ func (m *Machine) InstanceId() (instance.Id, error) {
...
+// display name. The display name is intended used for human input and
+// is ignored internally.
+func (m *Machine) InstanceNames() (instance.Id, string, error) {
```
wallyworld: Missing a test
You should implement InstanceId() in terns of InstanceNames()
```
func (m *Machine) InstanceId() (instance.Id, error) {
    instId, _, err := m.InstanceNames()
    return instId, err
}
```

state/machine.go
```diff
@@ -1241,6 +1242,27 @@ func (m *Machine) InstanceId() (instance.Id, error) {
...
+	instData, err := getInstanceData(m.st, m.Id())
+	if errors.IsNotFound(err) {
+		err = errors.NotProvisionedf("machine %v", m.Id())
+	}
+	if err != nil {
+		return "", "", err
+	}
+	var displayName string
+	if instData.DisplayName == nil {
+		displayName = ""
```
wallyworld: No need for this - displayName is initialised as "". Just need the != nil case

testing/factory/factory.go
```diff
@@ -90,6 +90,7 @@ type MachineParams struct {
 	Nonce           string
 	Constraints     constraints.Value
 	InstanceId      instance.Id
+	DisplayName     string
```
wallyworld: Where is this used? Looks like still need to add some code?

state/multiwatcher/multiwatcher.go
```diff
@@ -130,6 +130,7 @@ type Address struct {
 type MachineInfo struct {
 	ModelUUID                string                            `json:"model-uuid"`
 	Id                       string                            `json:"id"`
+	DisplayName              string                            `json:"display-name,omitempty"`
```
wallyworld: Is this being set anywhere? Looks like need to add code?

apiserver/params/params.go
```diff
@@ -213,6 +213,7 @@ type AddMachineParams struct {
 	// is set.
 	InstanceId              instance.Id                      `json:"instance-id"`
 	Nonce                   string                           `json:"nonce"`
+	DisplayName             string                           `json:"display-name"`
```
wallyworld: This doesn't appear to be used

environs/broker.go
```diff
@@ -105,6 +105,11 @@ type StartInstanceParams struct {
...
+	// StartInstanceResult.Instance.Id()
+	DisplayName string
```
wallyworld: This doesn't appear to be used - it needs to be fed into the SetInstanceInfo() API call.
(with the appropriate api versioning)

api/base/types.go
```diff
@@ -41,13 +41,14 @@ type ModelStatus struct {
...
-	Id         string
-	InstanceId string
-	HasVote    bool
-	WantsVote  bool
-	Status     string
-	Message    string
-	Hardware   *instance.HardwareCharacteristics
+	Id          string
+	InstanceId  string
+	DisplayName string
```
wallyworld: I can't see the CLI code which uses this - I guess it's still in progress?

apiserver/common/modelstatus_test.go
```diff
@@ -193,8 +193,8 @@ func (s *modelStatusSuite) TestModelStatus(c *gc.C) {
 			OwnerTag:           s.Owner.String(),
 			Life:               params.Alive,
 			Machines: []params.ModelMachineInfo{
-				{Id: "0", Hardware: &params.MachineHardware{Cores: &eight}, InstanceId: "id-4", Status: "pending", WantsVote: true},
-				{Id: "1", Hardware: stdHw, InstanceId: "id-5", Status: "pending"},
+				{Id: "0", Hardware: &params.MachineHardware{Cores: &eight}, InstanceId: "id-4", DisplayName: "id-4", Status: "pending", WantsVote: true},
```
wallyworld: displayname should either be "" or different to instance id

apiserver/common/machine_test.go
```diff
@@ -88,51 +88,64 @@ func (s *machineSuite) TestMachineHardwareInfo(c *gc.C) {
 	one := uint64(1)
 	amd64 := "amd64"
 	gig := uint64(1024)
+	displayName := "_"
 	st := mockState{
 		machines: map[string]*mockMachine{
-			"1": {id: "1", life: state.Alive, containerType: instance.NONE,
+			"1": {id: "1", displayName: &displayName, life: state.Alive, containerType: instance.NONE,
```
wallyworld: why a pointer here?

state/machine.go
timClicks: Resolved as we're no longer passing back a `*string`

environs/broker.go
timClicks: Now available within the provider code

api/base/types.go
timClicks: This struct is used by `api/common/modelstatus.go`, `api/common/modelmanager.go`, et al

core/instance/instance.go
timClicks: New type has been removed.

api/common/modelstatus.go
timClicks: Fixed.

api/provisioner/machine.go
timClicks: Fixed.

api/provisioner/machine.go
timClicks: Fixed.

apiserver/facades/controller/charmrevisionupdater/testing/suite.go
timClicks: Fixed

provider/oci/environ.go
timClicks: I've pulled this functionality out into its own function. IIRC I still include a defensive check.

provider/oci/environ.go
timClicks: Turns out that this is much harder to test than I thought it would be. The OCI provider doesn't seem to actually test `StartInstance`. All of the current tests appear to focus on interactions with the upstream OCI library.

state/allwatcher_internal_test.go
timClicks: Oh it was for increasing consistency. There were a few pre-existing casts and thought it would be cleaner if it were the same. Will revert

apiserver/params/params.go
timClicks: Have removed the field. Only manual providers such as ssh and winrm use the `InstanceId` and `Nonce` fields. And even then, using `DisplayName` here wouldn't be much use as the params are used to fill in `MachineTemplate`s. `MachineTemplate` does not have a `DisplayName` field.

apiserver/facades/client/modelmanager/modelmanager.go
```diff
@@ -1101,7 +1101,8 @@ func (m *ModelManagerAPI) getModelInfo(tag names.ModelTag) (params.ModelInfo, er
 		}
 	}
 	if canSeeMachines {
-		if info.Machines, err = common.ModelMachineInfo(st); shouldErr(err) {
+		info.Machines, err = common.ModelMachineInfo(st)
```
wallyworld: This change isn't needed

provider/maas/environ_whitebox_test.go
```diff
@@ -217,6 +217,40 @@ func (suite *environSuite) TestStartInstanceStartsInstance(c *gc.C) {
 	c.Check(err, jc.Satisfies, errors.IsNotFound)
 }
 
+func (suite *environSuite) TestStartInstanceReturnsHostnameAsDisplayName(c *gc.C) {
```
wallyworld: You don't need a whole new test here - just add an extra assert to TestStartInstanceStartsInstance or similar

provider/maas/maas2_test.go
```diff
@@ -309,6 +309,7 @@ func (m *fakeMachine) InterfaceSet() []gomaasapi.Interface {
 }
 
 func (m *fakeMachine) Start(args gomaasapi.StartArgs) error {
+	logger.Criticalf("%s", args)
```
wallyworld: remove me

state/machine.go
```diff
@@ -1231,14 +1232,25 @@ func (m *Machine) SetAgentPresence() (*presence.Pinger, error) {
 // InstanceId returns the provider specific instance id for this
 // machine, or a NotProvisionedError, if not set.
 func (m *Machine) InstanceId() (instance.Id, error) {
+	instId, _, err := m.InstanceNames()
+	if err != nil {
```
wallyworld: don't need this err check
just do
return instId, err

testing/factory/factory.go
timClicks: Have added the DisplayName to default settings within the factory, but haven't modified any tests to make use of the field directly yet.

apiserver/facades/client/modelmanager/modelmanager.go
timClicks: Sorry - leftover from adding logging

provider/maas/environ_whitebox_test.go
timClicks: I originally thought that too. The reason that I've needed the extra test is that `TestStartInstanceStartsInstance` doesn't use a method that exposes `StartInstanceResult`. That's alluded to within the comments of the new test:

>     // Create controller. We're not doing testing StartInstance on node0 because
>     // env.ControllerInstances and many of juju/testing/instance helper
>     // functions don't provide us with access to the environs.StartInstanceResult

wallyworld: It's easy to fix though. Instead of the existing test calling AssertStartInstance() helper, just break out the 3 lines inside that method and use in the test. Better to do that than add the extra LOC in a whole new test

wallyworld: If you drill down a couple of API calls, the start instance result is available to use.
AssertStartInstance -> StartInstance -> StartInstanceWithConstraints -> StartInstanceWithParams

The latter returns a result struct. It's just a bit of tweaking to make it work.

timClicks: Sorted

cmd/juju/status/status_internal_test.go
```diff
@@ -4833,6 +4855,24 @@ wordpress:logging-dir  logging:logging-directory  logging    subordinate
 	c.Assert(string(output), gc.Equals, expected)
 }
 
+func (s *StatusSuite) TestStatusWithFormatYaml(c *gc.C) {
```
timClicks: @wallyworld are these two what you were after? 1/2

cmd/juju/status/status_internal_test.go
```diff
@@ -4833,6 +4855,24 @@ wordpress:logging-dir  logging:logging-directory  logging    subordinate
...
+	ctx := s.prepareTabularData(c)
+	defer s.resetContext(c, ctx)
+	code, stdout, stderr := runStatus(c, "--format", "yaml")
+	c.Check(code, gc.Equals, 0)
+	c.Check(string(stderr), gc.Equals, "")
+	c.Assert(string(stdout), jc.Contains, "display-name: snowflake")
+}
+
+func (s *StatusSuite) TestStatusWithFormatJson(c *gc.C) {
```
timClicks: @wallyworld 2/2

cmd/juju/status/status_internal_test.go
```diff
@@ -3717,11 +3717,33 @@ func (sm startMachineWithHardware) step(c *gc.C, ctx *context) {
 	cfg, err := ctx.st.ControllerConfig()
 	c.Assert(err, jc.ErrorIsNil)
 	inst, _ := testing.AssertStartInstanceWithConstraints(c, ctx.env, environscontext.NewCloudCallContext(), cfg.ControllerUUID(), m.Id(), cons)
-	err = m.SetProvisioned(inst.Id(), "fake_nonce", &sm.hc)
+	err = m.SetProvisioned(inst.Id(), "", "fake_nonce", &sm.hc)
 	c.Assert(err, jc.ErrorIsNil)
 	ctx.pingers[m.Id()] = pinger
 }
 
+type startAliveMachineWithDisplayName struct {
```
wallyworld: You don't need all the code duplication - you can just put the displayname attribute on startAliveMachine and embed that struct in this one.

cmd/juju/status/status_internal_test.go
```diff
@@ -3717,11 +3717,33 @@ func (sm startMachineWithHardware) step(c *gc.C, ctx *context) {
...
+	pinger := ctx.setAgentPresence(c, m)
+	cons, err := m.Constraints()
+	c.Assert(err, jc.ErrorIsNil)
+	cfg, err := ctx.st.ControllerConfig()
+	c.Assert(err, jc.ErrorIsNil)
+	inst, hc := testing.AssertStartInstanceWithConstraints(c, ctx.env, environscontext.NewCloudCallContext(), cfg.ControllerUUID(), m.Id(), cons)
+	err = m.SetProvisioned(inst.Id(), sm.displayName, "fake_nonce", hc)
+	c.Assert(err, jc.ErrorIsNil)
+	ctx.pingers[m.Id()] = pinger
+	_, displayName, err := m.InstanceNames()
```
wallyworld: Don't do the displayname asserts here - that's what the tests do. This step() func is just to get things set up.

cmd/juju/status/status_internal_test.go
wallyworld: These work, thank you. Also need though a tabular version of the test because that's where the decision to use the display name really matters.

timClicks: I've done that on line 4840  by checking machine 1's line looks like:

```plain
1        started  10.0.1.1  snowflake     quantal
```

timClicks:
Encountered (hopefully intermittent) issue with [firewaller_test.go](http://ci.jujucharms.com/job/make-check-juju/2424/consoleFull).

```plain
00:23:24.040 	machine 1 not found (not found)
```

!!build!!

timClicks:
$$merge$$

manadart:
Firewaller tests failures should be addressed by https://github.com/juju/juju/pull/9641

PR #9598, opened 2018-12-28
Move the constraints package to core/constraints.

howbazaar:
A very mechanical move of the constaints package to core/constraints.


howbazaar:
$$merge$$

PR #9603, opened 2019-01-08
Clear the reboot flag in the worker before exiting.

howbazaar:
If the juju-reboot hook command was executed by a charm that was running on the controller, the machine agent would attempt to connect to the apiserver after it had been shut down.

This change moves the "clear the reboot flag" work to the worker itself before it returns the error. Should there have been a problem stopping another of the services or listing machines for the container managers, the machine agent would have still been in a restart loop. This way clearing the flag out prior to returning the error, we also remove that restart loop problem.

Drive by fix for a uniter test race failure.

## QA steps

Grab the ubuntu-lite charm locally, and add an install hook with the following body:

```bash
#!/bin/bash
status-set maintenance rebooting
juju-reboot
```

Then execute the following:

```
juju bootstrap lxd test
juju deploy -m controller ../path/to/charm u --to 0
```

* check the logs of the controller machine to see that it restarted.
* juju show-status-log -m contgroller u/0 shows the charm started fine

## Bug reference

https://bugs.launchpad.net/juju/+bug/1810313


howbazaar:
It is possible now that a reboot will be requested, but due to an error in the handling of that request, we'll lose the restart request. However, the situations that cause a shutdown request to fail are:
 * listing containers on the machine failing
 * listing the services
 * stopping a given service
These are unlikely to fail on any given machine. Now at scale, everything fails. so perhaps @jameinel's suggestion of documenting the behaviour a bit more so charms that actually case could be slightly more defensive in their programming. Such that a start hook that was expecting the install hook to cause a reboot would check to ensure the software is available, and if not fail and ask for a reboot.

howbazaar:
Changing to land against develop. Can backport once we release 2.5.0.

howbazaar:
$$merge$$

PR #9606, opened 2019-01-09
upgrade-model: Upgrades are always available if --build-agent is set

babbageclunk:
## Description of change

Since we'll always increment the build number when uploading newly-built binaries, we shouldn't bail early if the versions match. This means you can still upgrade to a dev version if you bootstrapped with an edge snap.

## QA steps

* Bootstrap with the edge snap.
* Checkout and build the tip of the same branch.
* Upgrade the controller with `juju upgrade-model -m controller --build-agent --debug`.
* The local tree should be built and uploaded, then the upgrade to the .1 build version should succeed.

## Documentation changes
None


mitechie:
    Beep..Boop...

PR #9609, opened 2019-01-09
Running 'juju status' shouldn't end up with INFO messages in the controller log

jameinel:
## Description of change

When investigating controller logs at INFO level, I noticed a lot of messages about "is a subordinate match". This triggers when you run "juju status", but it doesn't make sense to default to printing it all the time.

## QA steps

```
$ juju bootstrap --model-defaults logging-config="<root>=INFO"
$ juju deploy ubuntu
$ juju deploy telegraf:juju-info ubuntu:juju-info
$ juju relate subordinate ubuntu
$ juju status
```
Prior to this, you should see the message in "juju debug-log -m controller", and after it should default to DEBUG level.

## Documentation changes

None.

## Bug reference

None.


jameinel:
$$merge$$

PR #9612, opened 2019-01-09
Bug 1808947

hmlanigan:
## Description of change

Some instanceCharmProfileData docs were not removed as they should have been causing failures in destroying machines.  Using a subordinate charm with a profile on a machine in a cloud that doesn't support lxd profiles caused one doc left behind.  Added code in the uniter's install code flow to remove any instanceCharmProfileData for the machine.  Also enhanced checking in destroyHostOps to remove stray instanceCharmProfileData.

## QA steps

First do these steps with an lxd cloud, then a maas or was cloud:
```sh
juju bootstrap 
juju deploy cs:nova-compute
juju deploy cs:neutron-openvswitch-255
juju add-relation neutron-openvswitch nova-compute
```

once it's settled, `juju destroy-model`

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1808947

### DIFF
--- a/state/unit.go
+++ b/state/unit.go
@@ -724,8 +724,6 @@ func (u *Unit) destroyHostOps(a *Application) (ops []txn.Op, err error) {
 			return nil, err
 		}
 		if profile != "" {
-			// TODO (hml) 09-jan-2019
-			// how do we know when to clean this doc up?
 			logger.Tracef("Setup to remove charm profile %q, removing unit from machine %s", profile, m.Id())
 			ops = append(ops, m.SetUpgradeCharmProfileOp(a.Name(), "", lxdprofile.EmptyStatus))
 		}
--- a/state/unit.go
+++ b/state/unit.go
@@ -703,8 +703,8 @@ func (u *Unit) destroyHostOps(a *Application) (ops []txn.Op, err error) {
 				Assert: txn.DocExists,
 				Remove: true,
 			})
-		} else if errors.IsNotFound(err) {
-			logger.Tracef("Error getting instance charm profile data for machine %s, %s", m.Id(), err.Error())
+		} else if !errors.IsNotFound(err) {
+			logger.Errorf("Error getting instance charm profile data for machine %s, %s", m.Id(), err.Error())
 		}
 	} else {
 		machineAssert = bson.D{{"$or", []bson.D{
--- a/state/unit.go
+++ b/state/unit.go
@@ -692,14 +692,43 @@ func (u *Unit) destroyHostOps(a *Application) (ops []txn.Op, err error) {
 			{{"jobs", bson.D{{"$nin", []MachineJob{JobManageModel}}}}},
 			{{"hasvote", bson.D{{"$ne", true}}}},
 		}}}
+		// Clean up any instanceCharmProfileData docs for this machine before
+		// it is destroyed.
+		_, err := getInstanceCharmProfileData(m.st, m.doc.DocID)
+		if err == nil {
+			logger.Tracef("Remove instance charm profile data for machine %s", m.Id())
+			ops = append(ops, txn.Op{
+				C:      instanceCharmProfileDataC,
+				Id:     m.doc.DocID,
+				Assert: txn.DocExists,
+				Remove: true,
+			})
+		} else if errors.IsNotFound(err) {
+			logger.Tracef("Error getting instance charm profile data for machine %s, %s", m.Id(), err.Error())
+		}
 	} else {
 		machineAssert = bson.D{{"$or", []bson.D{
 			{{"principals", bson.D{{"$ne", []string{u.doc.Name}}}}},
 			{{"jobs", bson.D{{"$in", []MachineJob{JobManageModel}}}}},
 			{{"hasvote", true}},
 		}}}
-		logger.Tracef("Remove charm profile from machine %s for %s", m.Id(), a.Name())
-		ops = append(ops, m.SetUpgradeCharmProfileOp(a.Name(), "", lxdprofile.EmptyStatus))
+		// Remove any charm profile applied to the machine for this unit,
+		// if this is a unit removal from a machine which will not be
+		// destroyed.
+		machineProfiles, err := m.CharmProfiles()
+		if err != nil {
+			return nil, err
+		}
+		profile, err := lxdprofile.MatchProfileNameByAppName(machineProfiles, u.ApplicationName())
+		if err != nil {
+			return nil, err
+		}
+		if profile != "" {
+			// TODO (hml) 09-jan-2019
+			// how do we know when to clean this doc up?
+			logger.Tracef("Setup to remove charm profile %q, removing unit from machine %s", profile, m.Id())
+			ops = append(ops, m.SetUpgradeCharmProfileOp(a.Name(), "", lxdprofile.EmptyStatus))
+		}
 	}
 
 	// If removal conditions satisfied by machine & container docs, we can
--- a/state/unit_test.go
+++ b/state/unit_test.go
@@ -17,6 +17,7 @@ import (
 	"gopkg.in/juju/worker.v1"
 
 	"github.com/juju/juju/core/application"
+	"github.com/juju/juju/core/lxdprofile"
 	"github.com/juju/juju/core/status"
 	"github.com/juju/juju/instance"
 	"github.com/juju/juju/network"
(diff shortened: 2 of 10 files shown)


worker/provisioner/provisioner_task.go
```diff
@@ -368,7 +368,9 @@ func (task *provisionerTask) processProfileChanges(ids []string) error {
 			}
 		} else if err != nil {
 			logger.Errorf("cannot upgrade machine's lxd profile: %s", err.Error())
-			m.SetUpgradeCharmProfileComplete(lxdprofile.AnnotateErrorStatus(err))
+			if err2 := m.SetUpgradeCharmProfileComplete(lxdprofile.AnnotateErrorStatus(err)); err2 != nil {
```
hmlanigan: drive by

worker/provisioner/provisioner_task.go
```diff
@@ -383,16 +385,18 @@ func (task *provisionerTask) processProfileChanges(ids []string) error {
 			if err2 := m.SetStatus(status.Started, "", nil); err2 != nil {
 				return errors.Annotatef(err2, "cannot set error status for machine %q agent", m)
 			}
-			m.SetUpgradeCharmProfileComplete(lxdprofile.SuccessStatus)
+			if err2 := m.SetUpgradeCharmProfileComplete(lxdprofile.SuccessStatus); err2 != nil {
```
hmlanigan: drive by

worker/uniter/operation/deploy.go
```diff
@@ -51,6 +51,12 @@ func (d *deploy) Prepare(state State) (*State, error) {
 	if err := d.checkAlreadyDone(state); err != nil {
 		return nil, errors.Trace(err)
 	}
+	if d.kind == Install {
+		// TODO: check for terminal status?
```
SimonRichardson: We should check for terminal status here before removing the upgrade charm profile data. If the status is not terminal, we should return out an error, but we should remove the profile data so that subsequent deploys should proceed.

hmlanigan: There isn't a good way to determine if a doc in terminal status should exist.  The only time it should exist at this point, is for 2nd unit on a machine, such as a subordinate.  

Maybe checking to see if there is a doc, because the doc should already exist if it's going to.  If the doc exists, check the status for terminal.

hmlanigan: What happens if we error out here?  Is the uniter restarted?  and Perhaps the change would be complete by that point?

At this point we have the charmURL... could we see if the charm has a profile, and thus a doc to be removed?

state/unit_test.go
```diff
@@ -614,6 +618,9 @@ func (s *UnitSuite) TestRemoveUnitMachineNoDestroyCharmProfile(c *gc.C) {
 
 	host, err := s.State.AddMachine("quantal", state.JobHostUnits)
 	c.Assert(err, jc.ErrorIsNil)
+	err = host.SetProvisioned("inst-id", "fake_nonce", nil)
+	c.Assert(err, jc.ErrorIsNil)
+
```
hmlanigan: @SimonRichardson, this test is failing at line 636.  I'm not sure if this part is still valid or not.  Thoughts?

SimonRichardson: I'll have a look, from your PR.

worker/uniter/operation/deploy.go
SimonRichardson: > What happens if we error out here? Is the uniter restarted?

So, it depends... on [resume install](https://github.com/juju/juju/blob/d865bd728bc8219d410f97b9be4bc49da0167b9a/worker/uniter/uniter.go#L209-L215) it will cause the uniter to return out and not attempt a restart. BUT code later on will cause the uniter to [process the same operation](https://github.com/juju/juju/blob/d865bd728bc8219d410f97b9be4bc49da0167b9a/worker/uniter/uniter.go#L352-L392) and only certain error results will cause termination.

> At this point we have the charmURL... could we see if the charm has a profile, and thus a doc to be removed?

Yes, but I'd probably have a bespoke function in this case.

state/unit_test.go
SimonRichardson: I'm wondering if we change the `unit.go` to the following (see below), that way we don't need to check for provisioning issues. Although this is a bit more defensive in terms of programming style, it does mean that we don't have to care as much about the `m.CharmProfiles()` machineProfile errors.

```
diff --git a/state/unit.go b/state/unit.go
index 1f0072e..620367e 100644
--- a/state/unit.go
+++ b/state/unit.go
@@ -716,18 +716,21 @@ func (u *Unit) destroyHostOps(a *Application) (ops []txn.Op, err error) {
                // if this is a unit removal from a machine which will not be
                // destroyed.
                machineProfiles, err := m.CharmProfiles()
-               if err != nil {
-                       return nil, err
-               }
-               profile, err := lxdprofile.MatchProfileNameByAppName(machineProfiles, u.ApplicationName())
-               if err != nil {
-                       return nil, err
-               }
-               if profile != "" {
-                       // TODO (hml) 09-jan-2019
-                       // how do we know when to clean this doc up?
-                       logger.Tracef("Setup to remove charm profile %q, removing unit from machine %s", profile, m.Id())
-                       ops = append(ops, m.SetUpgradeCharmProfileOp(a.Name(), "", lxdprofile.EmptyStatus))
+               if err == nil {
+                       profile, err := lxdprofile.MatchProfileNameByAppName(machineProfiles, u.ApplicationName())
+                       if err != nil {
+                               return nil, err
+                       }
+                       if profile != "" {
+                               // TODO (hml) 09-jan-2019
+                               // how do we know when to clean this doc up?
+                               logger.Tracef("Setup to remove charm profile %q, removing unit from machine %s", profile, m.Id())
+                               ops = append(ops, m.SetUpgradeCharmProfileOp(a.Name(), "", lxdprofile.EmptyStatus))
+                       }
+               } else {
+                       if !errors.IsNotProvisioned(err) {
+                               return nil, err
+                       }
                }
        }
```

SimonRichardson: So the test now fails because we never call `SetUpgradeCharmProfileOp`. So the `chAppName` will always equal `""`. We should seriously think if this is what we intend.

hmlanigan: chAppName is always equal to "" here, because I need to set some charm profiles in the instance charm profile data.  :-)  overnight allowed for a light bulb moment to happen.

hmlanigan: To your code example, I had been thinking of a similar thing.  However that felt like coding for the unit tests to work, not because it was truly needed.  I'm up in the air on that.

state/unit.go
```diff
@@ -692,14 +692,43 @@ func (u *Unit) destroyHostOps(a *Application) (ops []txn.Op, err error) {
...
+		if err != nil {
+			return nil, err
+		}
+		profile, err := lxdprofile.MatchProfileNameByAppName(machineProfiles, u.ApplicationName())
+		if err != nil {
+			return nil, err
+		}
+		if profile != "" {
+			// TODO (hml) 09-jan-2019
+			// how do we know when to clean this doc up?
```
SimonRichardson: Just wondering if this TODO is required anymore?

state/unit_test.go
SimonRichardson: We'll keep this the way it currently is, as the tests shouldn't be short circuiting actual real world setup. We can always use this as a plan B if required.

state/watcher.go
```diff
@@ -2032,7 +2032,7 @@ func (w *instanceCharmProfileDataWatcher) merge(change watcher.Change) (bool, er
 			logger.Debugf("instanceCharmProfileData NOT mgo err not found")
 			return false, err
 		}
-		logger.Debugf("instanceCharmProfileData mgo err not found")
+		logger.Tracef("instanceCharmProfileData for %s on machine %s: mgo err not found", w.applicationName, machineId)
```
SimonRichardson: Good catch, I was obviously being optimistic about the error level here.

worker/uniter/operation/deploy.go
```diff
@@ -51,6 +51,17 @@ func (d *deploy) Prepare(state State) (*State, error) {
 	if err := d.checkAlreadyDone(state); err != nil {
 		return nil, errors.Trace(err)
 	}
+	if d.kind == Install {
+		// When a 2nd principal or subordinate unit having an lxd profile,
+		// is added to a machine, it is installed on the machine via the same
+		// mechanism as a charm upgrade with a profile.  However there is not
+		// a watcher waiting for the instanceCharmProfileData doc status to
+		// go into a terminal state so it can be deleted.
```
SimonRichardson: > However there is not a watcher waiting for the instanceCharmProfileData doc status to go into a terminal state so it can be deleted.

Maybe we should have that in the end, but I'm unsure what happens around timing issues of the uniter, but this is probably the most reliable realistic way at bullet proofing this feature.

We should remember this for the future.

hmlanigan:
$$merge$$

PR #9614, opened 2019-01-10
raft: Set a timeout when starting raft-transport

babbageclunk:
## Description of change

We were seeing a problem where the raft worker would occasionally hang during shutdown, and looking at a goroutine dump of the hung agent and experimenting with a similar situation extracted from a test that sometimes hangs in the same way lead to the discovery that no timeout was being set on the transport. The worker passes the timeout value from its config into the transport, but the manifold doesn't initialise it.

## QA steps

I haven't managed to directly reproduce the hang in an agent, so I haven't verified this other than in the test scenario. But it passes a smoke test:
* Bootstrap and enable HA.
* Upgrade the controller so it shuts down and starts again.
* The upgrade completes successfully.

## Documentation changes

None.

### DIFF
--- a/provider/azure/environ.go
+++ b/provider/azure/environ.go
@@ -1639,7 +1639,7 @@ func (env *azureEnviron) deleteControllerManagedResourceGroups(controllerUUID st
 	errs := make([]error, len(groupNames))
 	for i, name := range groupNames {
 		groupName := to.String(name)
-		logger.Debugf("  - deleting resource group %q", name)
+		logger.Debugf("  - deleting resource group %q", groupName)
 		wg.Add(1)
 		go func(i int) {
 			defer wg.Done()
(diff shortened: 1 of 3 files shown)


babbageclunk:
Yeah, I think you're right - bumped it up to 30s.

babbageclunk:
$$merge$$

babbageclunk:
To try to provoke the original problem and prove this fix I ran an upgrade loop overnight, but neither exhibited it - they both died from running out of disk space. I'll try it again over the weekend with this script (to keep the disk usage constant).

```bash
set -e
while :; do
    echo controller...
    juju upgrade-juju -m controller --build-agent
    sleep 60
    echo model...
    juju upgrade-juju
    sleep 30
    echo removing oldest binaries...
    juju run -m controller --all -- bash -c 'ls -1vd /var/lib/juju/tools/2.4.8.*-bionic-amd64 | head -1 | xargs sudo rm -r'
done
```

babbageclunk:
Remember to run tests after tiny review tweaks, doh.
$$merge$$

babbageclunk:
ugh
$$merge$$

PR #9615, opened 2019-01-10
fix nginx ingress patch typo and largerd db size a bit

ycliuhw:
## Description of change

fix nginx ingress patch typo and largerd db size a bit.

## QA steps

N/A

## Documentation changes

N/A

## Bug reference

N/A


ycliuhw:
$$merge$$

PR #9618, opened 2019-01-10
Adding unit tests to worker/provisioner for processOneMachineProfileChange and processProfileChange.

hmlanigan:
Adding unit tests to worker/provisioner for processOneMachineProfileChange and processProfileChange.


worker/provisioner/provisioner_task_test.go
```diff
@@ -21,23 +21,27 @@ import (
...
 	"github.com/juju/juju/controller/authentication"
+	"github.com/juju/juju/core/lxdprofile"
 	"github.com/juju/juju/core/status"
 	"github.com/juju/juju/core/watcher"
 	"github.com/juju/juju/core/watcher/watchertest"
 	"github.com/juju/juju/environs"
 	"github.com/juju/juju/environs/config"
 	"github.com/juju/juju/environs/context"
 	"github.com/juju/juju/environs/imagemetadata"
+	environsmocks "github.com/juju/juju/environs/mocks"
```
SimonRichardson: Not sure where this is coming from, but it currently doesn't exist - missing file?

hmlanigan: I created it before making the LXDProfileInstanceBroker interface.  No longer need for both and only half cleaned up.  Thanks for the catch.

hmlanigan:
$$merge$$

PR #9619, opened 2019-01-10
raft: Set a timeout when starting raft-transport

babbageclunk:
## Description of change

We were seeing a problem where the raft worker would occasionally hang during shutdown, and looking at a goroutine dump of the hung agent and experimenting with a similar situation extracted from a test that sometimes hangs in the same way lead to the discovery that no timeout was being set on the transport. The worker passes the timeout value from its config into the transport, but the manifold doesn't initialise it.

This is a direct forward-port of #9614 from 2.4.

## QA steps

I haven't managed to directly reproduce the hang in an agent, so I haven't verified this other than in the test scenario. But it passes a smoke test:
* Bootstrap and enable HA.
* Upgrade the controller so it shuts down and starts again.
* The upgrade completes successfully.

## Documentation changes

None.

### DIFF
--- a/provider/azure/environ.go
+++ b/provider/azure/environ.go
@@ -1674,7 +1674,7 @@ func (env *azureEnviron) deleteControllerManagedResourceGroups(ctx context.Provi
 	errs := make([]error, len(groupNames))
 	for i, name := range groupNames {
 		groupName := to.String(name)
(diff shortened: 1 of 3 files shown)


babbageclunk:
$$merge$$

PR #9620, opened 2019-01-11
Handle aws xen pv block devices

wallyworld:
## Description of change

AWS can provision xen pv volumes such that udev info is missing any hardware identifying characteristics. In this case, we need to force a fallback to match on device name, which is not perfect but it's all we can do.

## QA steps

bootstrap aws
juju deploy postgresql --storage pgdata=1G,ebs

Also test google and maas clouds

## Bug reference

https://bugs.launchpad.net/juju/+bug/1809478

### DIFF
--- a/apiserver/common/storagecommon/storage.go
+++ b/apiserver/common/storagecommon/storage.go
@@ -239,9 +239,11 @@ func volumeAttachmentDevicePath(
 		volumeAttachmentInfo.DeviceName != "" ||
 		volumeAttachmentInfo.DeviceLink != "" {
 		// Prefer the volume attachment's information over what is
-		// in the published block device information.
+		// in the published block device information, but only if the
+		// block device information actually has any device links. In
+		// some cases, the block device has very little hw info published.
 		var deviceLinks []string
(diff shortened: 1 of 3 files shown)


apiserver/common/storagecommon/storage.go
```diff
@@ -241,7 +241,7 @@ func volumeAttachmentDevicePath(
 		// Prefer the volume attachment's information over what is
 		// in the published block device information.
 		var deviceLinks []string
-		if volumeAttachmentInfo.DeviceLink != "" {
+		if volumeAttachmentInfo.DeviceLink != "" && len(blockDevice.DeviceLinks) > 0 {
```
babbageclunk: So, this is only using volumeAttachmentInfo.DeviceLink if the block device has device links?

babbageclunk: Ah, from reading below it sounds like blockDevice comes from the udevadm output, so empty device links is used to detect the weird case where the output udevadm output is nearly empty, thus forcing the matching to fall back to name, is that right? Probably worth a comment if so.

wallyworld:
$$merge$$

PR #9607, opened 2019-01-09
Handle deployments with storage and placement on maas

wallyworld:
## Description of change

Deploying a charm with storage to maas fails if a placement directive is used, eg --to zone=<blah>.
The PR fixes the underlying  deployment logic to support that case.

As a driveby, move a test for placement with lxd profiles next to the other similar tests.

## QA steps

bootstrap on maas
juju deploy postgresql --storage pgdata=1G,maas --to zone=<blah>

## Bug reference
https://bugs.launchpad.net/juju/+bug/1765959

PR #9599, opened 2019-01-07
Create an /etc/fstab entry when storage is attached

wallyworld:
## Description of change

When a block device is attached for storage, create an /etc/fstab entry.
Delete the /etc/fstab entry when storage is detached.

## QA steps

bootstrap aws
deploy postgresql with storage
ssh into the postgresql machine and check that there's an fstab entry matching the mtab entry
detach the storage
ensure that the fstab entry is removed

## Bug reference

https://bugs.launchpad.net/juju/+bug/1783419

PR #9621, opened 2019-01-11
Backport storage fixes

wallyworld:
## Description of change

Backport some storage fixes from develop:

https://github.com/juju/juju/pull/9599 
When a block device is attached for storage, create an /etc/fstab entry.
Delete the /etc/fstab entry when storage is detached.

https://github.com/juju/juju/pull/9607
Deploying a charm with storage to maas fails if a placement directive is used, eg --to zone=<blah>.
The PR fixes the underlying deployment logic to support that case.

Also forward port AWS storage fix from 2.4:

https://github.com/juju/juju/pull/9620
AWS can provision xen pv volumes such that udev info is missing any hardware identifying characteristics. In this case, we need to force a fallback to match on device name, which is not perfect but it's all we can do.

## QA steps

bootstrap aws and maas and test postgresql deployments with storage

## Bug reference

https://bugs.launchpad.net/juju/+bug/1783419
https://bugs.launchpad.net/juju/+bug/1765959
https://bugs.launchpad.net/juju/+bug/1809478

### DIFF
--- a/apiserver/common/storagecommon/blockdevices.go
+++ b/apiserver/common/storagecommon/blockdevices.go
@@ -42,23 +42,27 @@ func MatchingBlockDevice(
 	for _, dev := range blockDevices {
 		if planBlockInfo.HardwareId != "" {
 			if planBlockInfo.HardwareId == dev.HardwareId {
+				logger.Tracef("plan hwid match on %v", volumeInfo.HardwareId)
 				return &dev, true
 			}
 		}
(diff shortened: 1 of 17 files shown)


wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

jamesbeedy:
💯 awesome!! Thank you!!!

PR #9624, opened 2019-01-11
Webscale reporting

SimonRichardson:
## Description of change

Allow sending metrics to a influx db for the webscale tests

## QA steps

Install influxdb locally (I've used docker):
```
docker run -p 8086:8086 \
      -v $PWD:/var/lib/influxdb \
      influxdb
```

Then run the following:
```
./assess_deploy_webscale.py
```


acceptancetests/assess_deploy_webscale.py
```diff
@@ -139,13 +154,23 @@ def main(argv=None):
...
+            calculate_total_time(timings),
+            len(timings),
+            calculate_max_time(timings),
+            (time.time() - begin),
+        )
 
-        log.info("The timings for deployment: total txn time: {}, total nums txn: {}, max txn time: {}, max time (seconds): {}".format(total_time, total_txns, max_time, since))
+        log.info("Metrics for deployment: {}".format(metrics))
+
+        rclient = reportingClient(args.reporting_uri)
```
mitechie: This feels like we should have at least a try/catch with some logging in case of failure to connect?

acceptancetests/assess_deploy_webscale.py
```diff
@@ -139,13 +154,23 @@ def main(argv=None):
...
+        rclient.report(metrics, tags={
+            "git-sha": args.git_sha,
+            "charm-bundle": args.charm_bundle,
+        })
+        log.info("Metrics successfully sent to report storage")
```
mitechie: Another were maybe this should only be output if we know report worked where here it could be run w/o successful report?

acceptancetests/reporting.py
```diff
@@ -0,0 +1,90 @@
...
+
+__metaclass__ = type
+
+DBNAME = 'txn_metrics'
+POLICYNAME = 'txn_metric'
+
+class _Reporting:
+    """_Reporting represents a class to report metrics upon"""
+
+    __metaclass__ = abc.ABCMeta
```
mitechie: do we need the __metaclass__ calls? I'd think they'd be dropable.

acceptancetests/reporting.py
```diff
@@ -0,0 +1,90 @@
...
+            "max_time": "txn_metric.max_time",
+            "test_duration": "txn_metric.test_duration",
+        }
+
+    @abc.abstractmethod
+    def report(self, metrics, tags):
+        """Report the metrics to the underlying reporting client
+        """
+
+class InfluxDB(_Reporting):
```
mitechie: If this is the reporting client let's call it something like that so it doesn't appear to be something from the influxdb package. 

Maybe just InfluxClient or even ReportingClient?

acceptancetests/reporting.py
```diff
@@ -0,0 +1,90 @@
...
+                    "tags": tags,
+                    "time": int(now.strftime('%s')),
+                    "fields": {
+                        "value": metrics[key],
+                    },
+                }
+                series.append(pointValue)
+        self.client.write_points(series, retention_policy=POLICYNAME)
+
+def makeMetrics(total_time, total_num_txns, max_time, test_duration):
```
mitechie: this should just be make_metrics if it's just a function in the package.

acceptancetests/reporting.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    return {
+        "total_time": total_time,
+        "total_num_txns": total_num_txns,
+        "max_time": max_time,
+        "test_duration": test_duration,
+    }
+
+def reportingClient(uri):
```
mitechie: same with this reporting_client, but that seems like a noun vs a function that's doing something. maybe get_reporting_client or influx_connection or something?

acceptancetests/requirements.txt
```diff
@@ -37,3 +37,4 @@ python-novaclient>=6.0.2
 pyOpenSSL>=17.5.0
 boto3>=1.5.18
 pyaml>=17.12.1
+influxdb>=5.2.1
```
mitechie: alphabetical please

acceptancetests/reporting.py
SimonRichardson: So this is called by `./assess_deploy_webscale.py` - so that we don't have to keep track of the metric labels outside of the `reporting` module.

mitechie: right, I just mean python standards would be function names would be written with underscores.

SimonRichardson: Ah yes, so it should, I changed it! You probably want to code review this whole folder - :see_no_evil:

SimonRichardson:
@mitechie Updated to be more consistent.

SimonRichardson:
$$merge$$

PR #9625, opened 2019-01-14
2.5 txn update

jameinel:
## Description of change

This brings in the txn pruner changes from 2.4 into the 2.5 branch. This changes the code that keeps the txns collection small. The old algorithm would snapshot the list of completed transaction ids and then remove ones that are referenced, and then proceed to remove txns that are known to not be referenced.

The new algorithm reads a small batch of transactions (1000), then looks at the docs that they mention, and cleanup any docs that reference them, and then removes the transactions, since they are now known to be unreferenced. (This is safe to do, because once a transaction is in Complete or Aborted status, any other transaction can come along and remove it from the txn-queue of the docs.)

The main advantage of this is that we get to the point where transactions start being removed much quicker. The old one needed a lot of work to cache the list of transaction ids and then iterate all docs in the database to see what was referenced. It also had a separate pass to cleanup txn-queue for transactions that were already completed.

## QA steps

Bootstrapping and deploying a couple of charms, and then waiting for 1hr for the transaction pruner to be triggered should show the transactions being pruned. The code is also present in the mgopurge-1.8 release, which would allow you to trigger the pruning on demand.

## Documentation changes

Already handled when this code landed in the 2.4 series.

## Bug reference

See the earlier 2.4 changes.


wallyworld:
$$merge$$

jameinel:
$$merge$$

PR #9626, opened 2019-01-14
Fix list-offers for k8s models

wallyworld:
## Description of change

list-offers was failing for k8s models because it was trying to load space info and k8s models don't support spaces.
The cross model API calls also needed facade definitions to be updated for k8s models.

## QA steps

deploy mariadb to a k8s model
offer mariadb
check list-offers
deploy mediawiki to an iaas model
add a relation and check that everything works
do for a single and multiple controller scenario

## Bug reference

https://bugs.launchpad.net/juju/+bug/1811058


wallyworld:
$$merge$$

PR #9627, opened 2019-01-14
Revert "2.5 - Update LXD dependency"

Veebers:
Reverts juju/juju#9617

As per bug[1] reverting these changes to unblock an rc2 release and we'll land these changes at a later date (with bug fixes).

[1] https://bugs.launchpad.net/lxd/+bug/1811703


Veebers:
$$merge$$

PR #9637, opened 2019-01-17
WIP: IAAS deploy stack

SimonRichardson:
## Description of change

The following code abstracts away deployment of a stack, so that
we can utilise the same code for deploying both IAAS and CAAS stacks.

This should make it a lot easier to interchange between the two
when writing tests.


SimonRichardson:
https://trello.com/c/IyLsYD9H/956-create-iaas-client-instead-of-caas-for-deploying-a-bundle

acceptancetests/deploy_stack.py
```diff
@@ -95,6 +96,15 @@
     source: /dev/null
     type: disk
 """
+IAAS_LXD_PROFILE = """
```
mitechie: do we need this for the iaas model? I don't see what's in the bundle here?

mitechie: Hmm, so in looking at this since the bundle is provided by the calling into this maybe the lxd profile bits should be passed in vs hard coded as well. Ok for this PR but a general question for the future.

SimonRichardson: Yes, good shout! I'm definitely not sold on this bit.

SimonRichardson:
$$merge$$

PR #9639, opened 2019-01-18
Up timeout for checks.

Veebers:
## Description of change

Update the primary/sub relations functional test to wait a bit longer.
During ci-run the lxd machines can be under a bit of load so things take a little longer.


Veebers:
$$merge$$

PR #9644, opened 2019-01-21
Install fake ensure mongo is per suite not test.

anastasiamac:
## Description of change

For efficiency, setting up and confirming mongo is installed is better done per suite rather than per test.

Prior to this change, I could not stress test [1]  upgradeSuite at all. It still fails after 3 runs but at least it has the 3 runs now. I'll keep digging into failures but this change is aimed to make flaky tests less flaky.

[1] https://github.com/juju/juju/wiki/Stress-Test

### DIFF
--- a/cmd/jujud/agent/util_test.go
+++ b/cmd/jujud/agent/util_test.go
@@ -67,7 +67,6 @@ func (s *commonMachineSuite) SetUpSuite(c *gc.C) {
 	s.AgentSuite.SetUpSuite(c)
 	s.PatchValue(&jujuversion.Current, coretesting.FakeVersionNumber)
 	s.PatchValue(&stateWorkerDialOpts, mongotest.DialOpts())
-	s.fakeEnsureMongo = agenttest.InstallFakeEnsureMongo(s)
 }
 
 func (s *commonMachineSuite) SetUpTest(c *gc.C) {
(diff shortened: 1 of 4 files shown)


cmd/jujud/agent/util_test.go
```diff
@@ -67,10 +67,7 @@ func (s *commonMachineSuite) SetUpSuite(c *gc.C) {
 	s.AgentSuite.SetUpSuite(c)
 	s.PatchValue(&jujuversion.Current, coretesting.FakeVersionNumber)
 	s.PatchValue(&stateWorkerDialOpts, mongotest.DialOpts())
-}
-
-func (s *commonMachineSuite) TearDownSuite(c *gc.C) {
-	s.AgentSuite.TearDownSuite(c)
```
ycliuhw: just wondering why we remove teardown suite.

anastasiamac:
So, the test tear down should be responsible for cleaning mongo/returning it to pristine condition. The suite is responsible for installing it.

cmd/jujud/agent/util_test.go
anastasiamac: It's a redundant call. If there is nothing in the TearDown but the call to super.TearDown, the infrastructure will call it anyway.

wallyworld:
@Veebers InstallFakeEnsureMongo does not actually set up mongo - it just fakes out the calls to do so so.
There's no harm in moving it to be setup per suite. Given how little it does - just records what calls are made to set up mongo and returns a fake version, it's interesting that it makes a difference.

PR #9645, opened 2019-01-21
Don't swallow register error.

anastasiamac:
## Description of change

Register user command clears out macaroon if an error occurred during user registration and currently our CI fails to do that.
However, during the investigation it was discovered that the actual error that is causing the clean up was never visible.


anastasiamac:
self-approving to see what CI logs.

anastasiamac:
$$merge$$

PR #9657, opened 2019-01-23
PruneTxnSleepTime was meant to be optional.

jameinel:
## Description of change

The other field used 'intOrDefault' which properly handles the default
vaule when the parameter isn't available. But sleep time was a string,
and doesn't have a strOrDefault. So this does an inline "if we don't
have it, use the default".

## QA steps

Upgrade from a version that doesn't have 'prune-txn-sleep-time' set, and see that when the pruner wakes up, it doesn't panic.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1812954

### DIFF
--- a/controller/config.go
+++ b/controller/config.go
@@ -580,7 +580,15 @@ func (c Config) PruneTxnQueryCount() int {
 
 // PruneTxnSleepTime is the amount of time to sleep between batches.
 func (c Config) PruneTxnSleepTime() time.Duration {
-	val, _ := time.ParseDuration(c.mustString(PruneTxnSleepTime))
+	asInterface, ok := c[PruneTxnSleepTime]
+	if !ok {
+		asInterface = DefaultPruneTxnSleepTime
+	}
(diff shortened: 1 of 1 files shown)


controller/config.go
```diff
@@ -580,7 +580,15 @@ func (c Config) PruneTxnQueryCount() int {
...
-	val, _ := time.ParseDuration(c.mustString(PruneTxnSleepTime))
+	asInterface, ok := c[PruneTxnSleepTime]
+	if !ok {
+		asInterface = DefaultPruneTxnSleepTime
+	}
+	asStr, ok := asInterface.(string)
+	if !ok {
+		asStr = DefaultPruneTxnSleepTime
+	}
+	val, _ := time.ParseDuration(asStr)
```
SimonRichardson: What's the default for `time.ParseDuration` if it's an error, I'm guessing 0? Do we cater for 0 sleep time?

jameinel:
$$merge$$

PR #9660, opened 2019-01-23
If SetSupportedContainers doesn't change the value, don't write to the db

jameinel:
## Description of change

On startup, machine agents call `SetSupportedContainers`. If that doesn't actually change the supported containers, no need to write to the database.

Note the test infrastructure requires https://github.com/juju/txn/pull/49

## QA steps

You can use:
```
$ juju bootstrap lxd lxd
$ juju add-machine
$ juju model-config -m controller logging-config='<root>=INFO;juju.state.txn=TRACE'
# restart machine-0 in either the model or the controller, see what txns are run
```
## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1812981

### DIFF
--- a/state/machine.go
+++ b/state/machine.go
@@ -1847,6 +1847,20 @@ func isSupportedContainer(container instance.ContainerType, supportedContainers 
 
 // updateSupportedContainers sets the supported containers on this host machine.
 func (m *Machine) updateSupportedContainers(supportedContainers []instance.ContainerType) (err error) {
+	if m.doc.SupportedContainersKnown {
+		if len(m.doc.SupportedContainers) == len(supportedContainers) {
+			equal := true
+			for i := range m.doc.SupportedContainers {
+				if m.doc.SupportedContainers[i] != supportedContainers[i] {
+					equal = false
+					break
+				}
+			}
+			if equal {
+				return nil
+			}
+		}
+	}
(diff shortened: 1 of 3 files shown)


SimonRichardson:
@jameinel to get the CI to pass before the https://github.com/juju/txn/pull/49 lands, we could update the Gopkg.toml to point to the branch.

state/machine.go
```diff
@@ -1847,6 +1847,20 @@ func isSupportedContainer(container instance.ContainerType, supportedContainers
 
 // updateSupportedContainers sets the supported containers on this host machine.
 func (m *Machine) updateSupportedContainers(supportedContainers []instance.ContainerType) (err error) {
+	if m.doc.SupportedContainersKnown {
+		if len(m.doc.SupportedContainers) == len(supportedContainers) {
+			equal := true
+			for i := range m.doc.SupportedContainers {
+				if m.doc.SupportedContainers[i] != supportedContainers[i] {
```
SimonRichardson: What happens if the order of supported containers is different, yet they are the same (can that even happen)?

Example:
```A, B, C -> C, A, B```

SimonRichardson:
@jameinel closing this in favour of this merged version https://github.com/juju/juju/pull/9672

PR #9669, opened 2019-01-25
Fix goroutine and websocket leak in logsink handler

babbageclunk:
## Description of change

If the ServeHTTP handler finishes before the receiveLogs goroutine, it can be stuck in the select trying to send on logCh. Add a stop channel to the goroutine that is always closed when the handler finishes and include that in the select.

This should fix the leak we see in prodstack, although we don't understand why it's happening now - none of the logsink code has changed between 2.4 and 2.5

## QA steps

* Basic smoketest - bootstrapped a controller, deployed some applications and ensured that logging was still getting to the database using debug-log.
* Haven't verified the goroutine leak because we haven't been able to reproduce that.

## Documentation changes
None

## Bug reference
Fixes https://bugs.launchpad.net/juju/+bug/1813104

### DIFF
--- a/apiserver/logsink/logsink.go
+++ b/apiserver/logsink/logsink.go
@@ -67,6 +67,10 @@ func NewHTTPHandler(
 		newLogWriteCloser: newLogWriteCloser,
 		abort:             abort,
 		ratelimit:         ratelimit,
+		newStopChannel: func() (chan struct{}, func()) {
+			ch := make(chan struct{})
+			return ch, func() { close(ch) }
+		},
 	}
 }
 
@@ -75,6 +79,11 @@ type logSinkHandler struct {
 	abort             <-chan struct{}
 	ratelimit         *RateLimitConfig
 	mu                sync.Mutex
+
+	// newStopChannel is overridden in tests so that we can check the
+	// goroutine exits when prompted.
+	newStopChannel  func() (chan struct{}, func())
+	receiverStopped bool
 }
 
 // Since the logsink only receives messages, it is possible for the other end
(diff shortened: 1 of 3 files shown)


wallyworld:
Needs a test before landing

babbageclunk:
$$merge$$

apiserver/logsink/logsink.go
```diff
@@ -156,6 +167,9 @@ func (h *logSinkHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 				}
 			case m, ok := <-logCh:
 				if !ok {
+					h.mu.Lock()
+					defer h.mu.Unlock()
```
jameinel: I think this is clearer as
mu.Lock()
var = true
mu.Unlock()

no need for defer when you are done right away.

apiserver/logsink/logsink.go
```diff
@@ -233,6 +247,10 @@ func (h *logSinkHandler) receiveLogs(socket *websocket.Conn, endpointVersion int
 			// Send the log message.
 			select {
 			case <-h.abort:
+				// The API server is stopping.
+				return
+			case <-stop:
+				// The ServeHTTP handler has stopped.
 				return
 			case logCh <- m:
```
jameinel: so the failure is that the api server isnt stopping but the log channel is?

jameinel:
We ping once every 60s, so it is possible that is why I wasn't able to reproduce this. The failure seems to only be if the Ping to ensure the connection is alive fails.

PR #9672, opened 2019-01-25
If SetSupportedContainers doesn't change the value, don't write to the db

SimonRichardson:
## Description of change

On startup, machine agents call SetSupportedContainers. If that
doesn't actually change the supported containers, no need to write
to the database.

This supersedes https://github.com/juju/juju/pull/9660 as it contains
the full PR required - includes the landed https://github.com/juju/txn/pull/49
changes as well.
## QA steps

You can use:
```
$ juju bootstrap lxd lxd
$ juju add-machine
$ juju model-config -m controller logging-config='<root>=INFO;juju.state.txn=TRACE'
# restart machine-0 in either the model or the controller, see what txns are run
```
## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1812981

### DIFF
--- a/state/machine.go
+++ b/state/machine.go
@@ -2010,6 +2010,20 @@ func isSupportedContainer(container instance.ContainerType, supportedContainers 
 
 // updateSupportedContainers sets the supported containers on this host machine.
 func (m *Machine) updateSupportedContainers(supportedContainers []instance.ContainerType) (err error) {
+	if m.doc.SupportedContainersKnown {
+		if len(m.doc.SupportedContainers) == len(supportedContainers) {
+			equal := true
+			for i := range m.doc.SupportedContainers {
+				if m.doc.SupportedContainers[i] != supportedContainers[i] {
+					equal = false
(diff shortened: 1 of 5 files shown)


state/machine.go
```diff
@@ -2010,6 +2010,20 @@ func isSupportedContainer(container instance.ContainerType, supportedContainers
 
 // updateSupportedContainers sets the supported containers on this host machine.
 func (m *Machine) updateSupportedContainers(supportedContainers []instance.ContainerType) (err error) {
+	if m.doc.SupportedContainersKnown {
+		if len(m.doc.SupportedContainers) == len(supportedContainers) {
+			equal := true
+			for i := range m.doc.SupportedContainers {
+				if m.doc.SupportedContainers[i] != supportedContainers[i] {
```
hmlanigan: wondering if we can be sure of the order these two slices are in.

hmlanigan:
$$merge$$

PR #9681, opened 2019-01-29
Enable storage-pool update/delete API support.

Veebers:
## Description of change

Expose the ability to update and delete storage pools on the api.

## QA steps

This branch doesn't expose the functionality to the cli, just enables it via api calls. There are unit tests.

## Documentation changes

The followup PR will address the documentation additions as it will actually enable the CLI access to the functionality.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1812350

### DIFF
--- a/apiserver/facades/client/storage/storage.go
+++ b/apiserver/facades/client/storage/storage.go
@@ -14,17 +14,20 @@ import (
 	"github.com/juju/juju/apiserver/common/storagecommon"
 	"github.com/juju/juju/apiserver/facade"
 	"github.com/juju/juju/apiserver/params"
+	"github.com/juju/juju/caas"
 	"github.com/juju/juju/core/status"
+	"github.com/juju/juju/environs"
 	"github.com/juju/juju/environs/context"
 	"github.com/juju/juju/environs/tags"
 	"github.com/juju/juju/permission"
 	"github.com/juju/juju/state"
+	"github.com/juju/juju/state/stateenvirons"
 	"github.com/juju/juju/storage"
 	"github.com/juju/juju/storage/poolmanager"
 )
 
-// APIv3 implements the storage v3 API.
-type APIv3 struct {
+// StorageAPI implements the latest version (v5) of the Storage API which adds Update and Delete.
+type StorageAPI struct {
 	backend       backend
 	storageAccess storageAccess
 	registry      storage.ProviderRegistry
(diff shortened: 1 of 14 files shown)


Veebers:
[delete and update storage pools](https://trello.com/c/rlJvRuiS/596-delete-and-update-storage-pools)

apiserver/facades/client/storage/storage.go
```diff
@@ -1111,3 +1134,21 @@ func (a *APIv4) importFilesystem(
 
 // Destroy was dropped in V4, replaced with Remove.
 func (*APIv4) Destroy(_, _ struct{}) {}
+
+// DeletePool deletes the named pool
+func (a *APIv5) DeletePool(poolName string) error {
```
wallyworld: You'd need to mask these on V4 if you change the embedding to the current approach

storage/poolmanager/interface.go
```diff
@@ -28,6 +31,7 @@ type PoolManager interface {
 
 type SettingsManager interface {
 	CreateSettings(key string, settings map[string]interface{}) error
+	UpdateSettings(key string, settings map[string]interface{}) error
```
wallyworld: You want to Replace the current settings

storage/poolmanager/interface.go
```diff
@@ -19,6 +19,9 @@ type PoolManager interface {
 	// Delete removes the pool with name from state.
 	Delete(name string) error
 
+	// Update updates existing pool configuration with new or changed values
+	Update(name string, attrs map[string]interface{}) error
```
wallyworld: This needs to be Replace - provider type needs to be accounted for as well.

storage/poolmanager/interface.go
```diff
@@ -48,6 +52,15 @@ func (m MemSettings) CreateSettings(key string, settings map[string]interface{})
 	return nil
 }
 
+// UpdateSettings is part of the SettingsManager interface.
+func (m MemSettings) UpdateSettings(key string, settings map[string]interface{}) error {
+	if _, ok := m.Settings[key]; !ok {
+		return errors.NotFoundf("settings with key %q", key)
+	}
+	m.Settings[key] = settings
```
wallyworld: This is a replace operation (which is correct). But it's not an update.

storage/poolmanager/poolmanager.go
```diff
@@ -78,6 +78,14 @@ func (pm *poolManager) Delete(name string) error {
 	return errors.Annotatef(err, "deleting pool %q", name)
 }
 
+// Replace is defined on PoolManager interface.
+func (pm *poolManager) Replace(name string, attrs map[string]interface{}) error {
+	if name == "" {
+		return MissingNameError
+	}
+	return pm.settings.ReplaceSettings(globalKey(name), attrs)
```
wallyworld: This will break stuff - see Create()
We need to ensure the settings have the name and type added.
Also, where's the validation?
The only difference between create and replace is the existence of a previous pool so we need to refactor the methods to use common code.
On that basis, Replace should probably also support updating the type.

state/settings.go
```diff
@@ -373,6 +373,22 @@ func removeSettingsOp(collection, key string) txn.Op {
 	}
 }
 
+// replaceSettings replaces the settings value for key.
```
wallyworld: values

storage/poolmanager/poolmanager.go
wallyworld: These issue would have been picked p if there were tests added to poolmanager_test.go :-)

PR #9684, opened 2019-01-29
Exit http server worker after a time even if client mux still in use

wallyworld:
## Description of change

The http-server worker waits for (tracked) users of its http muxer to all exit before the worker loop terminates. Even though all the tracked workers are supposed to terminate when the server's dying channel is closed, some appear to stay stuck, maybe in a blocking read (not sure).

This PR will exit the http-server work after 1 minute even if not all mux clients have exited. The expectation is that any which are blocked on reads will then exit because the http socket will close.
If such an exit does occur, we dump the goroutines to /var/log/juju/apiserver-debug.log for further analysis.

## QA steps

QA is to ensure nothing is broken - bootstrap a controller, upgrade juju on it etc.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1813261

### DIFF
--- a/cmd/jujud/agent/machine/manifolds.go
+++ b/cmd/jujud/agent/machine/manifolds.go
@@ -243,6 +243,11 @@ type ManifoldsConfig struct {
 	// This is used by a number of workers to ensure serialisation of actions
 	// across the machine.
 	MachineLock machinelock.Lock
+
+	// MuxShutdownWait is the maximum time the http-server worker will wait
+	// for all mux clients to gracefully terminate before the http-worker
+	// exits regardless.
+	MuxShutdownWait time.Duration
 }
 
 // Manifolds returns a set of co-configured manifolds covering the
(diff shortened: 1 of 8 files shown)


wallyworld:
discussed on irc, 1 minute seems ok for now

wallyworld:
$$merge$$

wallyworld:
$$merge$$

cmd/jujud/agent/machine/manifolds.go
```diff
@@ -743,6 +748,8 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 			PrometheusRegisterer: config.PrometheusRegisterer,
 			AgentName:            config.AgentName,
 			Clock:                config.Clock,
+			MuxShutdownWait:      config.MuxShutdownWait,
+			LogDir:               "/var/log/juju", //agentConfig.LogDir(),
```
manadart: WIll this cause an issue for non-*nix builds?

jameinel: It probably isn't correct, but it is a valid directory on windows. It just happens to be "C:" or "D:" based on what drive the PWD is.

So its not great, and something like LogDir() or something based on DataDir() probably is better.

wallyworld:
[http worker termination backstop](https://trello.com/c/em6xBlMH/605-http-worker-termination-backstop)

PR #9689, opened 2019-01-30
Install mongodb via snap install juju-db

timClicks:
## Description of change

This PR enables the underlying `monogod` server that holds Juju's state to be installed via `snap`.

However, this PR should be understood as a **proof of concept** implementation only. Installing `mongod` is only available via enabling a feature flag: `mongodb-snap`.

## QA steps

`juju bootstrap`, `juju deploy` and `juju destroy-controller` complete successfully in the following environments:

`xenial` & `bionic`
- `export JUJU_DEV_FEATURE_FLAGS=mongodb-snap`
- `juju bootstrap --config default-series={series} localhost "c-{series}-jujudb-snap"
- `juju deploy wiki-simple`
- `juju destroy-controller --destroy-all-models -y`

`xenial`, `bionic` & `trusty`
- `unset JUJU_DEV_FEATURE_FLAGS`
- `juju bootstrap --config default-series={series} localhost "c-{series}-jujudb-classic"
- `juju deploy wiki-simple`
- `juju destroy-controller --destroy-all-models -y`

## Documentation changes

Juju's internal state database, an instance of MongoDB, can now be deployed via the `juju-db` snap package. This is an experimental feature, enabled by the `mongodb-snap` feature flag.  

## Bug reference

* https://bugs.launchpad.net/juju/+bug/1815688

### DIFF
--- a/mongo/mongo.go
+++ b/mongo/mongo.go
@@ -22,11 +22,15 @@ import (
 	"github.com/juju/packaging/manager"
 	"github.com/juju/replicaset"
 	"github.com/juju/utils"
+	"github.com/juju/utils/featureflag"
 	"gopkg.in/mgo.v2"
 
 	"github.com/juju/juju/controller"
+	"github.com/juju/juju/feature"
 	"github.com/juju/juju/network"
 	"github.com/juju/juju/service"
+	"github.com/juju/juju/service/common"
+	"github.com/juju/juju/service/snap"
 )
 
 var (
@@ -57,6 +61,13 @@ const (
 	// JujuMongoPackage is the mongo package Juju uses when
 	// installing mongo.
 	JujuMongoPackage = "juju-mongodb3.2"
+
+	// JujuDbSnap is the snap of MongoDB that Juju uses.
+	JujuDbSnap = "juju-db"
+
+	// JujuDbSnapMongodPath is the path that the juju-db snap
+	// makes mongod available at
+	JujuDbSnapMongodPath = "/snap/bin/juju-db.mongod"
 
 	// JujuMongoToolsPackage is the mongo package Juju uses when
 	// installing mongo tools to get mongodump etc.
@@ -433,26 +444,46 @@ func EnsureServer(args EnsureServerParams) (Version, error) {
 	return ensureServer(args, mongoKernelTweaks)
 }
 
+func setupDataDirectory(args EnsureServerParams) error {
+	dbDir := DbDir(args.DataDir)
+	if err := os.MkdirAll(dbDir, 0700); err != nil {
+		return errors.Annotate(err, "cannot create mongo database directory")
+	}
+
+	if err := UpdateSSLKey(args.DataDir, args.Cert, args.PrivateKey); err != nil {
+		return errors.Trace(err)
+	}
+
+	err := utils.AtomicWriteFile(sharedSecretPath(args.DataDir), []byte(args.SharedSecret), 0600)
+	if err != nil {
+		return errors.Annotatef(err, "cannot write mongod shared secret to %v", sharedSecretPath(args.DataDir))
+	}
+
+	if err := os.MkdirAll(logPath(dbDir), 0644); err != nil {
+		return errors.Annotate(err, "cannot create mongodb logging directory")
+	}
+
+	return nil
+}
+
 func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string) (Version, error) {
 	var zeroVersion Version
 	tweakSysctlForMongo(mongoKernelTweaks)
+
+	if featureflag.Enabled(feature.MongoDbSnap) {
+		// TODO(tsm): push this to earlier in the bootstrapping process
+		if args.DataDir != dataPathForJujuDbSnap {
+			logger.Warningf("overwriting args.dataDir (set to %v) to %v", args.DataDir, dataPathForJujuDbSnap)
+			args.DataDir = dataPathForJujuDbSnap
+		}
+	}
+
 	logger.Infof(
 		"Ensuring mongo server is running; data directory %s; port %d",
 		args.DataDir, args.StatePort,
 	)
 
-	dbDir := filepath.Join(args.DataDir, "db")
-	if err := os.MkdirAll(dbDir, 0700); err != nil {
-		return zeroVersion, errors.Errorf("cannot create mongo database directory: %v", err)
-	}
-
-	oplogSizeMB := args.OplogSize
-	if oplogSizeMB == 0 {
-		var err error
-		if oplogSizeMB, err = defaultOplogSize(dbDir); err != nil {
-			return zeroVersion, errors.Trace(err)
-		}
-	}
+	setupDataDirectory(args)
 
 	if err := installMongod(series.MustHostSeries(), args.SetNUMAControlPolicy); err != nil {
 		// This isn't treated as fatal because the Juju MongoDB
@@ -469,13 +500,12 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string) 
 	}
 	logVersion(mongoPath)
 
-	if err := UpdateSSLKey(args.DataDir, args.Cert, args.PrivateKey); err != nil {
-		return zeroVersion, errors.Trace(err)
-	}
-
-	err = utils.AtomicWriteFile(sharedSecretPath(args.DataDir), []byte(args.SharedSecret), 0600)
-	if err != nil {
-		return zeroVersion, errors.Errorf("cannot write mongod shared secret: %v", err)
+	oplogSizeMB := args.OplogSize
+	if oplogSizeMB == 0 {
+		oplogSizeMB, err = defaultOplogSize(DbDir(args.DataDir))
+		if err != nil {
+			return zeroVersion, errors.Trace(err)
+		}
 	}
 
 	// Disable the default mongodb installed by the mongodb-server package.
@@ -492,53 +522,50 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string) 
 		}
 	}
 
-	svcConf := newConf(ConfigArgs{
-		DataDir:       args.DataDir,
-		DBDir:         dbDir,
-		MongoPath:     mongoPath,
-		Port:          args.StatePort,
-		OplogSizeMB:   oplogSizeMB,
-		WantNUMACtl:   args.SetNUMAControlPolicy,
-		Version:       mongodVersion,
-		Auth:          true,
-		IPv6:          network.SupportsIPv6(),
-		MemoryProfile: args.MemoryProfile,
-	})
-	svc, err := newService(ServiceName, svcConf)
+	mongoArgs := generateConfig(mongoPath, args.DataDir, args.StatePort, oplogSizeMB, args.SetNUMAControlPolicy, mongodVersion, args.MemoryProfile)
+	logger.Debugf("creating mongo service configuration for mongo version: %d.%d.%s at %q",
+		mongoArgs.Version.Major, mongoArgs.Version.Minor, mongoArgs.Version.Patch, mongoArgs.MongoPath)
+
+	svc, err := mongoArgs.asService()
 	if err != nil {
 		return zeroVersion, errors.Trace(err)
 	}
-	installed, err := svc.Installed()
-	if err != nil {
-		return zeroVersion, errors.Trace(err)
-	}
-	if installed {
-		exists, err := svc.Exists()
+
+	// TODO(tsm): refactor out to service.Configure
+	if featureflag.Enabled(feature.MongoDbSnap) {
+		err = mongoArgs.writeConfig(configPath(args.DataDir))
 		if err != nil {
 			return zeroVersion, errors.Trace(err)
 		}
-		if exists {
-			logger.Debugf("mongo exists as expected")
-			running, err := svc.Running()
-			if err != nil {
-				return zeroVersion, errors.Trace(err)
-			}
 
-			if !running {
-				return mongodVersion, errors.Trace(svc.Start())
-			}
-			return mongodVersion, nil
+		err := snap.SetSnapConfig(ServiceName, "configpath", configPath(args.DataDir))
+		if err != nil {
+			return zeroVersion, errors.Trace(err)
 		}
+
+		err = service.ManuallyRestart(svc)
+		if err != nil {
+			logger.Criticalf("unable to (re)start mongod service: %v", err)
+			return zeroVersion, errors.Trace(err)
+		}
+
+		return mongodVersion, nil
 	}
 
-	if err := svc.Stop(); err != nil {
-		return zeroVersion, errors.Annotatef(err, "failed to stop mongo")
+	running, err := svc.Running()
+	if err != nil {
+		return zeroVersion, errors.Trace(err)
 	}
+	if running {
+		return mongodVersion, nil
+	}
+
+	dbDir := DbDir(args.DataDir)
 	if err := makeJournalDirs(dbDir); err != nil {
-		return zeroVersion, fmt.Errorf("error creating journal directories: %v", err)
+		return zeroVersion, errors.Errorf("error creating journal directories: %v", err)
 	}
 	if err := preallocOplog(dbDir, oplogSizeMB); err != nil {
-		return zeroVersion, fmt.Errorf("error creating oplog files: %v", err)
+		return zeroVersion, errors.Errorf("error creating oplog files: %v", err)
 	}
 
 	if err := service.InstallAndStart(svc); err != nil {
@@ -610,6 +637,17 @@ func installPackage(pkg string, pacconfer config.PackagingConfigurer, pacman man
 }
 
 func installMongod(operatingsystem string, numaCtl bool) error {
+	if featureflag.Enabled(feature.MongoDbSnap) {
+		prerequisites := []snap.App{snap.NewApp("core")}
+		backgroundServices := []snap.BackgroundService{{"daemon", true}}
+		conf := common.Conf{Desc: ServiceName + " snap"}
+		service, err := snap.NewService(ServiceName, conf, snap.Command, "edge", "jailmode", backgroundServices, prerequisites)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		return service.Install()
+	}
+
 	// fetch the packaging configuration manager for the current operating system.
 	pacconfer, err := config.NewPackagingConfigurer(operatingsystem)
 	if err != nil {
@@ -678,7 +716,6 @@ func installMongod(operatingsystem string, numaCtl bool) error {
 			}
 		}
 	}
-
 	return nil
 }
 
--- a/mongo/mongo_test.go
+++ b/mongo/mongo_test.go
@@ -117,7 +117,6 @@ func (s *MongoSuite) makeConfigArgs(dataDir string) mongo.ConfigArgs {
 		OplogSizeMB: 1024,
 		WantNUMACtl: false,
 		Version:     s.mongodVersion,
-		Auth:        true,
 		IPv6:        true,
 	}
 }
@@ -256,7 +255,7 @@ func (s *MongoSuite) TestEnsureServerServerExistsAndRunning(c *gc.C) {
 	s.assertMongoConfigFile(c)
 
 	c.Check(s.data.Installed(), gc.HasLen, 0)
-	s.data.CheckCallNames(c, "Installed", "Exists", "Running")
+	s.data.CheckCallNames(c, "Running")
 }
 
 func (s *MongoSuite) TestEnsureServerSetsSysctlValues(c *gc.C) {
@@ -305,7 +304,7 @@ func (s *MongoSuite) TestEnsureServerServerExistsNotRunningIsStarted(c *gc.C) {
 	s.assertMongoConfigFile(c)
 
 	c.Check(s.data.Installed(), gc.HasLen, 0)
-	s.data.CheckCallNames(c, "Installed", "Exists", "Running", "Start")
+	s.data.CheckCallNames(c, "Running", "Install", "Stop", "Start")
 }
 
 func (s *MongoSuite) TestEnsureServerServerExistsNotRunningStartError(c *gc.C) {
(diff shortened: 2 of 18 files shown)


agent/agent.go
```diff
@@ -18,13 +18,15 @@ import (
 	"github.com/juju/loggo"
 	"github.com/juju/os/series"
 	"github.com/juju/utils"
+	//	"github.com/juju/utils/featureflag"
```
wallyworld: delete me, and below

mongo/mongo.go
```diff
@@ -7,6 +7,7 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"fmt"
+	"github.com/juju/juju/service/snap"
```
wallyworld: move me

mongo/mongo.go
```diff
@@ -435,26 +441,38 @@ func EnsureServer(args EnsureServerParams) (Version, error) {
 
 func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string) (Version, error) {
 	var zeroVersion Version
+	var err error
```
wallyworld: Can't see where this is used. In any case, if needed, it needs to be declared next to where it's used.

mongo/mongo.go
```diff
@@ -435,26 +441,38 @@ func EnsureServer(args EnsureServerParams) (Version, error) {
...
 	tweakSysctlForMongo(mongoKernelTweaks)
 	logger.Infof(
 		"Ensuring mongo server is running; data directory %s; port %d",
 		args.DataDir, args.StatePort,
```
wallyworld: So with the check below, data dir here is wrong, so this log Info really needs to be part of the "if using snap" conditional below

mongo/mongo.go
```diff
@@ -435,26 +441,38 @@ func EnsureServer(args EnsureServerParams) (Version, error) {
...
 	)
 
+	if featureflag.Enabled(feature.MongoDbSnap) {
+		if args.DataDir != snap.DefaultDataPath {
+			logger.Warningf("overwriting args.dataDir (set to %v) to %v", args.DataDir, snap.DefaultDataPath)
```
wallyworld: Don't really care about if it's overwritten or not.
The important bit is - we're using the snap. So we just need a log message stating that the snap is used.
Also, it's not sufficient just to override data dir here. It needs to be set properly up front depending on if the snap is used, since other things like backups need to be able to access it.
In the code, we have a lot of this:

```
dbPath := filepath.Join(dataDir, "db")
```

You need to create a helper function and return the right thing depending on the feature flag

mongo/mongo.go
```diff
@@ -435,26 +441,38 @@ func EnsureServer(args EnsureServerParams) (Version, error) {
...
-		var err error
 		if oplogSizeMB, err = defaultOplogSize(dbDir); err != nil {
 			return zeroVersion, errors.Trace(err)
 		}
 	}
 
-	if err := installMongod(series.MustHostSeries(), args.SetNUMAControlPolicy); err != nil {
+	series := series.MustHostSeries()
+	if featureflag.Enabled(feature.MongoDbSnap) {
+		series = "snap"
```
wallyworld: let's not abuse series in this way
You need to introduce a decision point here for snap vs package.
installMongod() below is very package specific in all it does.
Consider a new installMongodSnap() func to call for a snap to do snappy type things

mongo/mongo.go
```diff
@@ -523,7 +543,21 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string)
 			if err != nil {
 				return zeroVersion, errors.Trace(err)
 			}
-
+			if featureflag.Enabled(feature.MongoDbSnap) {
```
wallyworld: Probably time to introduce a new function for the "if exists" branch given all the indenting etc.
But this is the wrong place for all this.
Setting up mongo should be part of the ServiceActions.Install(), ie implemented in service/snap.go

mongo/mongo.go
```diff
@@ -604,6 +638,10 @@ func installPackage(pkg string, pacconfer config.PackagingConfigurer, pacman man
 	// apply release targeting if needed.
 	if pacconfer.IsCloudArchivePackage(pkg) {
 		pkg = strings.Join(pacconfer.ApplyCloudArchiveTarget(pkg), " ")
+	} else if featureflag.Enabled(feature.MongoDbSnap) {
```
wallyworld: This doesn't belong here - installPackage is really for apt/yum etc only

mongo/mongo.go
```diff
@@ -694,6 +731,8 @@ func packagesForSeries(series string) ([]string, []string) {
 		return []string{"juju-mongodb"}, []string{}
 	case "xenial", "artful":
 		return []string{JujuMongoPackage, JujuMongoToolsPackage}, []string{}
+	case "snap":
```
wallyworld: this will disappear also when the abstraction is fixed

service/snap/snap_test.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2019 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package snap_test
+
+// import (
```
wallyworld: fix me

service/snap/snap.go
```diff
@@ -0,0 +1,422 @@
...
+
+	"github.com/juju/juju/feature"
+	"github.com/juju/juju/service/common"
+)
+
+const (
+	Command           = "snap"
+	CustomConfigPath  = "/var/snap/juju-db/common/custom.config"
+	DefaultConfigPath = "/var/snap/juju-db/common/default.config"
+	DefaultDataPath   = "/var/snap/juju-db/common"
```
wallyworld: this is what needs to be the dbDir (via that helper function mentioned earlier) if snap is used

service/snap/snap.go
```diff
@@ -0,0 +1,422 @@
...
+	"github.com/juju/utils/set"
+	"github.com/juju/utils/shell"
+
+	"github.com/juju/juju/feature"
+	"github.com/juju/juju/service/common"
+)
+
+const (
+	Command           = "snap"
+	CustomConfigPath  = "/var/snap/juju-db/common/custom.config"
```
wallyworld: And these are then relative to the dbDir using filepath.Join

service/snap/snap.go
```diff
@@ -0,0 +1,422 @@
...
+func IsRunning() (bool, error) {
+	if runtime.GOOS == "windows" {
+		return false, nil
+	}
+
+	cmd := exec.Command(Command, "version")
+	out, err := cmd.CombinedOutput()
+	logger.Debugf("snap version output: %#v", string(out[:]))
+	if err == nil {
+		if !featureflag.Enabled(feature.MongoDbSnap) {
```
wallyworld: No need to use feature flag at this level.
The primitives in this package should be purely snap related

service/service.go
```diff
@@ -36,6 +38,12 @@ var linuxInitSystems = []string{
 	InitSystemUpstart,
 }
 
+// func init() {
```
wallyworld: Delete me?

mongo/service.go
```diff
@@ -143,133 +168,377 @@ func sharedSecretPath(dataDir string) string {
...
+	return nil
+}
+
 // EnsureServiceInstalled is a convenience method to [re]create
-// the mongo service.
-func EnsureServiceInstalled(dataDir string, statePort, oplogSizeMB int, setNUMAControlPolicy bool, version Version, auth bool, memProfile MemoryProfile) error {
-	mongoPath, err := Path(version)
+// the mongo service. It assumes that the necessary packages are
+// already installed and that the file system includes secrets at dataDir.
+func EnsureServiceInstalled(dataDir string, statePort int, oplogSizeMB int, setNUMAControlPolicy bool, version Version, auth bool, memProfile MemoryProfile) error {
```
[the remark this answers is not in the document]
timClicks: Sounds good to me

service/snap/snap.go
```diff
@@ -0,0 +1,512 @@
...
+// the name parameter to fetch and install a snap with a matching name, then uses
+// default policies for the installation. To install a snap with --classic confinement,
+// or via --edge, --candidate or --beta, then create the Service via another method.
+func NewServiceFromName(name string, conf common.Conf) (Service, error) {
+	Prerequisites := []App{}
+	BackgroundServices := []BackgroundService{}
+	Channel := defaultChannel
+	ConfinementPolicy := defaultConfinementPolicy
+
+	if name == "juju-db" {
```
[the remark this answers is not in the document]
timClicks: yeah totally yuck - will fix

mongo/mongodfinder.go
```diff
@@ -37,6 +40,11 @@ func NewMongodFinder() *MongodFinder {
 
 // FindBest tries to find the mongo version that best fits what we want to use.
 func (m *MongodFinder) FindBest() (string, Version, error) {
+	if featureflag.Enabled(feature.MongoDbSnap) {
+		v, err := m.findVersion(JujuDbSnapMongodPath)
+		return JujuDbSnapMongodPath, v, err
```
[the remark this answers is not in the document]
timClicks: Good point, will fix

mongo/service.go
```diff
@@ -143,133 +168,377 @@ func sharedSecretPath(dataDir string) string {
...
+
+func (mongoArgs *ConfigArgs) startCommand() string {
+	if featureflag.Enabled(feature.MongoDbSnap) {
+		// TODO(tsm): work out how to bridge mongoService and service.Service
+		// to access the StartCommands method, rather than duplicating code
+		return snap.Command + " start --enable " + ServiceName
+	}
+
+	cmd := ""
+	if mongoArgs.WantNUMACtl {
```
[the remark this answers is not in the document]
timClicks: Okay, will do

mongo/service.go
```diff
@@ -126,12 +146,17 @@ func ReStartService() error {
 	if err != nil {
 		return errors.Trace(err)
 	}
+	return restartService(svc)
+}
+
+func restartService(svc mongoService) error {
```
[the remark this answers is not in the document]
timClicks: IIRC The mongoService interface doesn't define Restart

mongo/service.go
```diff
@@ -143,133 +168,377 @@ func sharedSecretPath(dataDir string) string {
...
+	if usingMongo2 {
+		mongoArgs.SSLOnNormalPorts = true
+	} else {
+		mongoArgs.SSLMode = "requireSSL"
+	}
+
+	if featureflag.Enabled(feature.MongoDbSnap) {
+		mongoArgs.Syslog = false
+		mongoArgs.LogAppend = true
+		mongoArgs.LogPath = logPath(dataDir)
```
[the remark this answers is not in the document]
timClicks: I'm uncertain, to be honest. Happy to pull this out of the conditional. Can check

mongo/mongo.go
```diff
@@ -298,33 +309,23 @@ func mongoPath(
...
-	if _, err = stat(path); err == nil {
+
+	if foundVersion.NewerThan(version) >= 0 {
 		return path, nil
 	}
-	logger.Infof("could not find a suitable binary for %q", version)
-	errMsg := fmt.Sprintf("no suitable binary for %q", version)
-	return "", errors.New(errMsg)
+
+	path, err = lookPath("mongod")
```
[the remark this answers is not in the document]
timClicks: This was an attempt at eliminating code duplication by using mongofinder.FindBest(). Happy to revert.

mongo/mongo.go
```diff
@@ -298,33 +309,23 @@ func mongoPath(
...
-		} else {
-			return "", err
-		}
+	finder := NewMongodFinder()
+	path, foundVersion, err := finder.FindBest()
+
+	if err != nil && !errors.IsNotFound(err) {
+		return "", errors.Trace(err)
 	}
-	path := JujuMongodPath(version)
```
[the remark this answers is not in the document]
timClicks: I believe that the path returned by JujuMongodPath() is checked by mongofinder.FindBest() and isn't needed here. Will check.

mongo/mongo.go
```diff
@@ -433,28 +434,56 @@ func EnsureServer(args EnsureServerParams) (Version, error) {
...
+	err := utils.AtomicWriteFile(sharedSecretPath(args.DataDir), []byte(args.SharedSecret), 0600)
+	if err != nil {
+		return errors.Annotatef(err, "cannot write mongod shared secret to %v", sharedSecretPath(args.DataDir))
+	}
+
+	if err := os.MkdirAll(logPath(dbDir), 0644); err != nil {
+		return errors.Annotate(err, "cannot create mongodb logging directory")
+	}
+
+	if err := makeJournalDirs(dbDir); err != nil {
```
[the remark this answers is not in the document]
timClicks: My intention was to simplify all of the nested conditionals in the original code. Moving those two pieces into that function felt appropriate they should be no-ops when mongo is already in-place and now all setup filesystem operations are in one place.

mongo/mongo.go
```diff
@@ -433,28 +434,56 @@ func EnsureServer(args EnsureServerParams) (Version, error) {
...
-		var err error
-		if oplogSizeMB, err = defaultOplogSize(dbDir); err != nil {
-			return zeroVersion, errors.Trace(err)
-		}
-	}
+	setupDataDirectory(args)
 
-	if err := installMongod(series.MustHostSeries(), args.SetNUMAControlPolicy); err != nil {
+	series := series.MustHostSeries()
+	logger.Infof("installing mongo with series %s", series)
```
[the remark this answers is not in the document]
timClicks: Sorry, was hang over from my debugging. Fixed locally.

mongo/mongo.go
```diff
@@ -464,18 +493,16 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string)
 	}
 	finder := NewMongodFinder()
 	mongoPath, mongodVersion, err := finder.FindBest()
-	if err != nil {
+	if err != nil && !errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
timClicks: Good find. This was a carryover from a previous iteration where I would install the snap later.

mongo/mongo.go
```diff
@@ -492,53 +519,41 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string)
...
+		mongoArgs.Version.Major, mongoArgs.Version.Minor, mongoArgs.Version.Patch, mongoArgs.MongoPath)
+
+	svc, err := mongoArgs.asService()
 	if err != nil {
 		return zeroVersion, errors.Trace(err)
 	}
-	if installed {
-		exists, err := svc.Exists()
+
+	if featureflag.Enabled(feature.MongoDbSnap) {
```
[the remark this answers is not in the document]
timClicks: Totally agree.

mongo/prealloc.go
```diff
@@ -72,6 +74,10 @@ func defaultOplogSize(dir string) (int, error) {
 		return smallOplogSizeMB, nil
 	}
 
+	if featureflag.Enabled(feature.MongoDbSnap) {
```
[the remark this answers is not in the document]
timClicks: Misunderstanding of snap's fs permissions. Have reverted.

service/common/conf.go
```diff
@@ -81,25 +81,28 @@ func (c Conf) Validate(renderer shell.Renderer) error {
 	if c.ExecStart == "" {
 		return errors.New("missing ExecStart")
 	}
-	for field, cmd := range map[string]string{
+
+	fieldsToCheck := map[string]string{
```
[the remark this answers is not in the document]
timClicks: Personal preference.. I find the inline back-to-front. I'll revert

service/discovery.go
```diff
@@ -71,9 +73,11 @@ func versionInitSystem(ser string) (string, error) {
 		return "", errors.Wrap(err, notFound)
 	}
 
-	switch seriesos {
-	case os.Windows:
+	if seriesos == os.Windows {
```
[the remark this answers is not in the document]
timClicks: Oh - it was to support snaps as a special case when I was passing `"snap"` though as a series but I forgot to repair all of the code

service/service.go
```diff
@@ -245,42 +258,37 @@ func Restart(name string) error {
...
-
-}
-
-func restart(svc Service) error {
-	// Use the Restart method, if there is one.
-	if svc, ok := svc.(RestartableService); ok {
-		if err := svc.Restart(); err != nil {
+// ManuallyRestart restarts the service by applying
+// its Restart method or by falling back to calling Stop and Start
+func ManuallyRestart(svc ServiceActions) error {
```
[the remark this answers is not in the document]
timClicks: It's used by the mongo package.. should be possible to refactor/remove when Configure is a service method

mongo/mongo_test.go
```diff
@@ -596,7 +617,8 @@ func (s *MongoSuite) assertTestMongoGetFails(c *gc.C, series string, packageMana
 
 	// Verify that EnsureServer continued and started the mongodb service.
 	c.Check(s.data.Installed(), gc.HasLen, 0)
-	s.data.CheckCallNames(c, "Installed", "Exists", "Running", "Start")
+	s.data.CheckCallNames(c, "Running", "Install", "Stop", "Start")
+	//	s.data.CheckCallNames(c, "Installed", "Exists", "Running", "Start")
```
wallyworld: remove me

mongo/mongo_test.go
```diff
@@ -617,38 +639,40 @@ func (s *MongoSuite) TestInstallMongodServiceExists(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 
 	c.Check(s.data.Installed(), gc.HasLen, 0)
-	s.data.CheckCallNames(c, "Installed", "Exists", "Running")
+	s.data.CheckCallNames(c, "Running")
```
wallyworld: Why has existing behaviour changed? We need to ensure that non-snap service setup is exactly how it has always been

mongo/mongo_test.go
```diff
@@ -617,38 +639,40 @@ func (s *MongoSuite) TestInstallMongodServiceExists(c *gc.C) {
...
 }
 
 func (s *MongoSuite) TestNewServiceWithReplSet(c *gc.C) {
-	conf := mongo.NewConf(s.makeConfigArgs(c.MkDir()))
+	confArgs := s.makeConfigArgs(c.MkDir())
+	conf := mongo.NewConf(&confArgs)
```
wallyworld: Is there a reason why this was changed?
It's idiomatic to pass in smallish structs  to methods that don't modify the data

mongo/mongo.go
```diff
@@ -469,13 +500,12 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string)
...
 	logVersion(mongoPath)
 
-	if err := UpdateSSLKey(args.DataDir, args.Cert, args.PrivateKey); err != nil {
-		return zeroVersion, errors.Trace(err)
-	}
-
-	err = utils.AtomicWriteFile(sharedSecretPath(args.DataDir), []byte(args.SharedSecret), 0600)
-	if err != nil {
-		return zeroVersion, errors.Errorf("cannot write mongod shared secret: %v", err)
+	oplogSize := args.OplogSize
```
wallyworld: The original var was called oplogSizeMB to convey to the user the units.
There's no need to change that var name

mongo/mongo.go
```diff
@@ -492,53 +522,50 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string)
...
 	}
+
+	dbDir := DbDir(args.DataDir)
 	if err := makeJournalDirs(dbDir); err != nil {
-		return zeroVersion, fmt.Errorf("error creating journal directories: %v", err)
+		return Version{}, errors.Errorf("error creating journal directories: %v", err)
 	}
-	if err := preallocOplog(dbDir, oplogSizeMB); err != nil {
-		return zeroVersion, fmt.Errorf("error creating oplog files: %v", err)
+	if err := preallocOplog(dbDir, args.OplogSize); err != nil {
```
wallyworld: Ooops, we have set oplogSizeMB above and are not using it. The change here is a mistake it seems.
Also, why the sudden aversion to zeroVersion?

mongo/service_test.go
```diff
@@ -23,144 +23,143 @@ var _ = gc.Suite(&serviceSuite{})
...
+		dataDir,
+		port,
+		oplogSizeMB,
+		false,
+		mongodVersion,
+		mongo.MemoryProfileDefault,
+	)
+	conf := mongo.NewConf(confArgs)
+
+	// expected := common.Conf{
```
wallyworld: ??

mongo/service_test.go
```diff
@@ -23,144 +23,143 @@ var _ = gc.Suite(&serviceSuite{})
...
+		} else if field == "--oplogSize" {
+			hasNoPrealloc = true
+		} else if field == "--sslOnNormalPorts" {
+			hasSslOnNormalPorts = true
+		}
 	}
-
-	c.Check(conf, jc.DeepEquals, expected)
-	c.Check(strings.Fields(conf.ExecStart), jc.DeepEquals, strings.Fields(expected.ExecStart))
+	c.Check(hasSmallFiles, gc.Equals, true)
```
wallyworld: We can't not check the expected conf

mongo/mongo.go
timClicks: Ah, I was shadowing the [field name](https://github.com/juju/juju/blob/484c173746b4560d724f099aa60e6c516ccf79ac/mongo/mongo.go#L449) to make it obvious that they're connected.

Are you happy for the local variable to be renamed back to `oplogSizeMB` or would you like that change to be made further upstream where the user is more likely to encounter it?

mongo/mongo.go
timClicks: Good catch

I'll get `zeroVersion` in there.

mongo/service_test.go
timClicks: This was removed because the ordering of Go's maps are not stable. The intention behind skipping the check of the whole `conf` struct is to make it clear that this test is intended to test arguments that relate to Mongo 2.4 (`--smallfiles`,  `--noprealloc` and `--sslOnNormalPorts`). Will fix.

mongo/service_test.go
```diff
@@ -23,144 +27,209 @@ var _ = gc.Suite(&serviceSuite{})
...
-			" --journal" +
-			" --replSet juju" +
-			" --quiet" +
-			" --oplogSize 10" +
-			" --auth" +
-			" --keyFile '/var/lib/juju/shared-secret'" +
-			" --sslMode requireSSL" +
-			" --storageEngine wiredTiger" +
-			" --wiredTigerCacheSizeGB 1",
+	// confArgs := mongo.GenerateConf(
```
wallyworld: delete me

timClicks:
[Build failed](http://ci.jujucharms.com/job/make-check-juju/2912/consoleFull) in (hopefully unrelated) area:

```plain
00:23:59.774 === RUN   TestPackage
00:23:59.774 2019-02-12 20:49:20 WARNING juju.testing mgo.go:212 failed to start mongo: mongod failed to listen on port 44457
00:23:59.774 --- FAIL: TestPackage (0.09s)
00:23:59.774     mgo.go:469: mongod failed to listen on port 44457
00:23:59.774 FAIL
00:23:59.774 FAIL	github.com/juju/juju/worker/authenticationworker	1.859s
```

!!build!!

timClicks:
$$merge$$

PR #9697, opened 2019-01-31
Omit model list when attempting to show controller for user with login access

achilleasa:
## Description of change

Prior to this PR, when a user with login access attempted to `juju show-controller` they would get back `{}` as the result. The current implementation attempts to fetch the controller models; the call fails with a permission denied error which gets suppressed and the code moves to the next controller. If the command is invoked without an argument then the code only iterates the active controller which results in the empty set being returned to the client.

This PR introduces a permission check to the show controller implementation. When the check fails, the code defaults to an empty model list which allows users to access non-model controller details which may be already cached locally (see the commit log of f39ccd1 for additional info why this fix will not work for users with the `add-model` permission).

## QA steps
```
juju add-user foo
juju logout
juju login -u foo -c lxd-test
juju show-controller

lxd-test:
  details:
    uuid: 321f8195-3e34-4676-8ca9-05a5697acafe
    controller-uuid: 321f8195-3e34-4676-8ca9-05a5697acafe
    api-endpoints: ['10.153.103.93:17070']
    ca-cert: |
      -----BEGIN CERTIFICATE-----
      ...
      -----END CERTIFICATE-----
    cloud: localhost
    region: localhost
    agent-version: (unauthorized access)
  current-model: admin/default
  account:
    user: foo
    access: login
``` 

## Bug reference
https://bugs.launchpad.net/juju/+bug/1808202

### DIFF
--- a/cmd/juju/controller/showcontroller.go
+++ b/cmd/juju/controller/showcontroller.go
@@ -139,6 +139,15 @@ func (c *showControllerCommand) Run(ctx *cmd.Context) error {
 		}
 
 		var details ShowControllerDetails
+
+		// If user lacks sufficient access level to display controller details,
+		// append the empty controller details and attach a permission error.
+		if !permission.Access(access).EqualOrGreaterControllerAccessThan(permission.SuperuserAccess) {
+			details.Errors = append(details.Errors, errors.Unauthorizedf("permission denied").Error())
+			controllers[controllerName] = details
+			continue
+		}
+
 		allModels, err := client.AllModels()
 		if err != nil {
 			details.Errors = append(details.Errors, err.Error())
--- a/cmd/juju/controller/showcontroller.go
+++ b/cmd/juju/controller/showcontroller.go
@@ -138,25 +138,33 @@ func (c *showControllerCommand) Run(ctx *cmd.Context) error {
 			one.AgentVersion = c.agentVersion(client, ctx)
 		}
 
-		var details ShowControllerDetails
+		var (
+			details   ShowControllerDetails
+			allModels []base.UserModel
+		)
 
-		// If user lacks sufficient access level to display controller details,
-		// append the empty controller details and attach a permission error.
-		if !permission.Access(access).EqualOrGreaterControllerAccessThan(permission.SuperuserAccess) {
-			details.Errors = append(details.Errors, errors.Unauthorizedf("permission denied").Error())
-			controllers[controllerName] = details
-			continue
-		}
-
-		allModels, err := client.AllModels()
-		if err != nil {
-			details.Errors = append(details.Errors, err.Error())
-			continue
-		}
-		// Update client store.
-		if err := c.SetControllerModels(c.store, controllerName, allModels); err != nil {
-			details.Errors = append(details.Errors, err.Error())
-			continue
+		// NOTE: this user may have been granted AddModelAccess which
+		// should allow them to list only the models they have access to.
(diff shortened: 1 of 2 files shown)


cmd/juju/controller/showcontroller.go
```diff
@@ -139,6 +139,15 @@ func (c *showControllerCommand) Run(ctx *cmd.Context) error {
 		}
 
 		var details ShowControllerDetails
+
+		// If user lacks sufficient access level to display controller details,
+		// append the empty controller details and attach a permission error.
+		if !permission.Access(access).EqualOrGreaterControllerAccessThan(permission.SuperuserAccess) {
```
SimonRichardson: I have no idea myself, but should it always be compared to a `superuser`, or should this be a case of does the user have access to controller through access controls.

i.e. `juju grant $USER read controller`

mitechie:
So I tested this out and have some feedback. 

User with only login you get the {}, ok. However I think in this case you don't have any models, fine, but you should still get controller details that are know (especially ones in your local cache of data cat .local/share/juju/controllers.yaml) anyway. 

When I show-controller on JAAS I get that data. 

    jaas:
      details:
        uuid: a030379a
        controller-uuid: a030379a
        api-endpoints: ['jimm.jujucharms.com:443']
        ca-cert: ""
        cloud: ""
        agent-version: (not implemented)
      models:
      account:
        user: rharding@external
        access: (not implemented)

I don't see any reason we'd not include this data out to the user as a result of show-controller. 

I then granted my external user 'add-model' access on the controller. I still get the {}. So now I have permissions to create new models on the controller. I should get to see an empty model list and the other data as well imo. 

So next I used that user to add a model:

 juju add-model ricktest --credential company
Uploading credential 'google/rick/company' to controller
Added 'ricktest' model on google/us-east1 with credential 'company' for user 'rick'

Now show-controller is still an empty {} but I do have a model on that controller. 

In the end, I think that if the data is already in your local cache and you have valid login or higher permissions we should show that data along with an empty models list. If we have models on the controller it's great to show the models I have access to on there.

achilleasa:
@mitechie  While going through the permission bits, I noticed the [AddModelAccess](https://github.com/juju/juju/blob/develop/permission/access.go#L36) one but the comparison code for [controller](https://github.com/juju/juju/blob/develop/permission/access.go#L92) access seems to ignore it so I went with `SuperuserAccess` instead.  

On a side-note, `juju grant -h` does not list `add-model` in the controller permission list. It *does* let me grant it but it does not show up in `juju list-users` (I get `login` instead). 

The permission comparison code for CloudAccess takes `AddModelAccess` into account but I am not sure how to grab its value in that particular controller without having to modify `ControllerAccessAPI` to expose it.

The original code silently failed when trying to list the models (the suppressed error was: authorization failed) so I guess that if I figure out the right set of permission constraints (maybe something like `controller access > Login || cloud(?) access >= AddModelAccess`) that allow users to list models without errors we could make your last example work as expected.

pmatulis:
@achilleasa Take a look here: https://bugs.launchpad.net/juju/+bug/1811143

achilleasa:
@SimonRichardson I have pushed an extra commit (and updated the PR description) with a workaround to address @mitechie concerns. Let me know what you think and whether it makes sense to squash the 2 commits as the previous solution yields incorrect results for users with `add-model` permission.

achilleasa:
$$merge$$

achilleasa:
$$merge$$

PR #9699, opened 2019-01-31
Bug 1813044 Changing instanceCharmProfileData docID to <machine>#<unit>

hmlanigan:
## Description of change

Change instanceCharmProfileData to be per machine and per unit, rather than simply per machine.  This avoids collisions when upgrading multiple charms at once, and while deploying bundles.

`SetUpgradeCharmProfileComplete()` and `RemoveUpgradeCharmProfileData()` now require an unit name.  `WatchModelMachinesCharmProfiles()` and `WatchContainersCharmProfiles()` now pass a new string of `"<machine-id>#<unit-name>"` when triggered to the provisioner.

## QA steps

1. `juju deploy -n 6 ./testcharms/charm-repo/quantal/lxd-profile`
1. wait for the machines to start provisioning
3. `juju deploy ~/charms/ubuntu --to 0 ; juju add-unit ubuntu -n 5 --to 1,2,3,4,5`
2. `juju deploy ~/charms/ntp`
3. `juju deploy ./testcharms/charm-repo/quantal/lxd-profile-subordinate`
4. `juju add-relation lxd-profile lxd-profile-subordinate`
5. `juju add-relation ntp ubuntu`
4. once settled:
```juju upgrade-charm lxd-profile-subordinate --path ./testcharms/charm-repo/quantal/lxd-profile-subordinate; juju upgrade-charm lxd-profile --path ./testcharms/charm-repo/quantal/lxd-profile ; juju upgrade-charm ntp --path ~/charms/ntp ; juju upgrade-charm ubuntu --path ~/charms/ubuntu```

Also test with deploy --to lxd.  With PR version server and client; as well as 2.5.0 server and PR version client.

## Documentation changes

n/a

## Bug reference

https://bugs.launchpad.net/juju/+bug/1813044

### DIFF
--- a/api/provisioner/machine.go
+++ b/api/provisioner/machine.go
@@ -120,7 +120,7 @@ type MachineProvisioner interface {
 	WatchContainersCharmProfiles(ctype instance.ContainerType) (watcher.StringsWatcher, error)
 
 	// CharmProfileChangeInfo retrieves the info necessary to change a charm
-	// profile used by a machine, for the give unit.
+	// profile used by a machine, for the given unit.
 	CharmProfileChangeInfo(string) (CharmProfileChangeInfo, error)
 
 	// SetCharmProfiles records the given slice of charm profile names.
@@ -128,7 +128,7 @@ type MachineProvisioner interface {
 
 	// SetUpgradeCharmProfileComplete records the result of updating
 	// the machine's charm profile(s), for the given unit.
-	SetUpgradeCharmProfileComplete(string, string) error
+	SetUpgradeCharmProfileComplete(unitName string, message string) error
 
 	// RemoveUpgradeCharmProfileData completely removes the instance charm profile
 	// data for a machine and the given unit, even if the machine is dead.
--- a/api/provisioner/machine.go
+++ b/api/provisioner/machine.go
@@ -120,19 +120,19 @@ type MachineProvisioner interface {
 	WatchContainersCharmProfiles(ctype instance.ContainerType) (watcher.StringsWatcher, error)
 
 	// CharmProfileChangeInfo retrieves the info necessary to change a charm
-	// profile used by a machine.
-	CharmProfileChangeInfo() (CharmProfileChangeInfo, error)
+	// profile used by a machine, for the give application.
+	CharmProfileChangeInfo(string) (CharmProfileChangeInfo, error)
 
 	// SetCharmProfiles records the given slice of charm profile names.
 	SetCharmProfiles([]string) error
 
-	// SetUpgradeCharmProfileComplete recorded that the result of updating
-	// the machine's charm profile(s)
-	SetUpgradeCharmProfileComplete(string) error
+	// SetUpgradeCharmProfileComplete records the result of updating
+	// the machine's charm profile(s), for the given application.
+	SetUpgradeCharmProfileComplete(appName string, msg string) error
 
 	// RemoveUpgradeCharmProfileData completely removes the instance charm profile
-	// data for a machine, even if the machine is dead.
-	RemoveUpgradeCharmProfileData() error
+	// data for a machine and the provided application, even if the machine is dead.
+	RemoveUpgradeCharmProfileData(string) error
 }
 
 // Machine represents a juju machine as seen by the provisioner worker.
@@ -572,11 +572,16 @@ type CharmProfileChangeInfo struct {
 }
 
 // CharmProfileChangeInfo implements MachineProvisioner.CharmProfileChangeInfo.
-func (m *Machine) CharmProfileChangeInfo() (CharmProfileChangeInfo, error) {
+func (m *Machine) CharmProfileChangeInfo(appName string) (CharmProfileChangeInfo, error) {
 	var results params.ProfileChangeResults
-	args := params.Entities{Entities: []params.Entity{
-		{Tag: m.tag.String()},
-	}}
+	args := params.ProfileArgs{
+		Args: []params.ProfileArg{
+			{
+				Entity:  params.Entity{Tag: m.tag.String()},
+				AppName: appName,
+			},
+		},
+	}
 	err := m.st.facade.FacadeCall("CharmProfileChangeInfo", args, &results)
 	if err != nil {
 		return CharmProfileChangeInfo{}, err
@@ -627,12 +632,13 @@ func (m *Machine) SetCharmProfiles(profiles []string) error {
 }
 
 // SetUpgradeCharmProfileComplete implements MachineProvisioner.SetUpgradeCharmProfileComplete.
-func (m *Machine) SetUpgradeCharmProfileComplete(message string) error {
+func (m *Machine) SetUpgradeCharmProfileComplete(appName, message string) error {
 	var results params.ErrorResults
 	args := params.SetProfileUpgradeCompleteArgs{
 		Args: []params.SetProfileUpgradeCompleteArg{
 			{
 				Entity:  params.Entity{Tag: m.tag.String()},
+				AppName: appName,
 				Message: message,
 			},
 		},
@@ -652,12 +658,13 @@ func (m *Machine) SetUpgradeCharmProfileComplete(message string) error {
 }
 
 // RemoveUpgradeCharmProfileData implements MachineProvisioner.RemoveUpgradeCharmProfileData.
-func (m *Machine) RemoveUpgradeCharmProfileData() error {
+func (m *Machine) RemoveUpgradeCharmProfileData(appName string) error {
 	var results params.ErrorResults
-	args := params.Entities{
-		Entities: []params.Entity{
+	args := params.ProfileArgs{
+		Args: []params.ProfileArg{
 			{
-				Tag: m.tag.String(),
+				Entity:  params.Entity{Tag: m.tag.String()},
+				AppName: appName,
 			},
 		},
 	}
--- a/api/provisioner/machine.go
+++ b/api/provisioner/machine.go
@@ -120,7 +120,7 @@ type MachineProvisioner interface {
 	WatchContainersCharmProfiles(ctype instance.ContainerType) (watcher.StringsWatcher, error)
 
 	// CharmProfileChangeInfo retrieves the info necessary to change a charm
-	// profile used by a machine, for the give application.
+	// profile used by a machine, for the give unit.
 	CharmProfileChangeInfo(string) (CharmProfileChangeInfo, error)
 
 	// SetCharmProfiles records the given slice of charm profile names.
@@ -572,13 +572,13 @@ type CharmProfileChangeInfo struct {
 }
 
 // CharmProfileChangeInfo implements MachineProvisioner.CharmProfileChangeInfo.
-func (m *Machine) CharmProfileChangeInfo(appName string) (CharmProfileChangeInfo, error) {
+func (m *Machine) CharmProfileChangeInfo(unitName string) (CharmProfileChangeInfo, error) {
 	var results params.ProfileChangeResults
 	args := params.ProfileArgs{
 		Args: []params.ProfileArg{
 			{
-				Entity:  params.Entity{Tag: m.tag.String()},
-				AppName: appName,
+				Entity:   params.Entity{Tag: m.tag.String()},
+				UnitName: unitName,
 			},
 		},
 	}
@@ -637,9 +637,9 @@ func (m *Machine) SetUpgradeCharmProfileComplete(appName, message string) error 
 	args := params.SetProfileUpgradeCompleteArgs{
 		Args: []params.SetProfileUpgradeCompleteArg{
 			{
-				Entity:  params.Entity{Tag: m.tag.String()},
-				AppName: appName,
-				Message: message,
+				Entity:   params.Entity{Tag: m.tag.String()},
+				UnitName: appName,
+				Message:  message,
 			},
 		},
 	}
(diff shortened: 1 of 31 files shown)

### CODE REVIEW

File: apiserver/facades/agent/provisioner/provisioner.go [hunk 1]
```diff
@@ -1414,15 +1425,15 @@ func (a *ProvisionerAPI) CACert() (params.BytesResult, error) {
 
 // CharmProfileChangeInfo retrieves the info necessary to change a charm
 // profile used by a machine.
-func (p *ProvisionerAPI) CharmProfileChangeInfo(machines params.Entities) (params.ProfileChangeResults, error) {
-	results := make([]params.ProfileChangeResult, len(machines.Entities))
+func (p *ProvisionerAPI) CharmProfileChangeInfo(args params.ProfileArgs) (params.ProfileChangeResults, error) {
```
hmlanigan: I'm wondering if we should keep the old version with the ProvisionerAPIV7

apiserver/facades/agent/uniter/lxdprofile_test.go
```diff
@@ -24,15 +24,13 @@ type lxdProfileSuite struct {
 
 	machineTag1 names.MachineTag
 	unitTag1    names.UnitTag
-	unitTag2    names.UnitTag
```
hmlanigan: drive by, these were un-used.

apiserver/facades/agent/provisioner/provisioner.go
SimonRichardson: Is there a chance we get old agents trying to connect to V7 and how does that work not that we require the application name to upgrade a profile. I guess for V7 we could attempt to work it out potentially?

hmlanigan:
@jameinel can you please review?

state/machine_test.go
```diff
@@ -170,96 +172,120 @@ func (s *MachineSuite) testSetUpgradeCharmProfileWithoutLXDProfile(c *gc.C, prof
...
+func (s *MachineSuite) TestNextUpgradeCharmProfileApplicationNameZero(c *gc.C) {
+	s.assertSetUpUnitForLXDProfile(c)
+	m := s.machine
+
+	name, err := m.NextUpgradeCharmProfileUnitName()
+	c.Assert(err, gc.ErrorMatches, "cannot get instance charm profile data for machine 1: instance charm profile data for machine 1 not found")
+	c.Assert(name, gc.Equals, "")
+}
+
+//func (s *MachineSuite) TestSingleUpgradeCharmProfileApplicationNameMoreThanOne(c *gc.C) {
```
SimonRichardson: Any reason why this is commented out?

state/machine.go
```diff
@@ -2487,6 +2577,28 @@ func (m *Machine) RemoveUpgradeCharmProfileData() error {
 	return nil
 }
 
+func (m *Machine) NextRemoveUpgradeCharmProfileData() error {
```
SimonRichardson: Probably need to test this.

state/machine.go
```diff
@@ -2487,6 +2577,28 @@ func (m *Machine) RemoveUpgradeCharmProfileData() error {
...
+	logger.Debugf("trying deleted instance charm profile data for machine %s here", m.Id())
+	instData, err := getNextInstanceCharmProfileData(m.st, m.Id(), true)
+	if err != nil {
+		return errors.Annotatef(err, "cannot get instance charm profile data for machine %v", m.Id())
+	}
+	return m.RemoveUpgradeCharmProfileData(instData.UpgradeCharmProfileUnit)
+}
+
+func (m *Machine) NextSetUpgradeCharmProfileComplete(msg string) error {
```
SimonRichardson: Probably need to test this.

state/watcher.go
```diff
@@ -803,19 +803,37 @@ func (w *minUnitsWatcher) Changes() <-chan []string {
...
-	return m.st.watchCharmProfiles(isChildRegexp)
+	isChildRegexp := fmt.Sprintf("^%s/%s/%s#%s$", m.doc.DocID, ctype, names.NumberSnippet, names.UnitSnippet)
+	return m.st.watchCharmProfiles(isChildRegexp, true)
 }
 
-func (st *State) watchCharmProfiles(regExp string) (StringsWatcher, error) {
+// WatchModelMachinesCharmProfiles returns a StringsWatcher that notifies of
+// changes to the upgrade charm profile charm url for a machine.
+//
+// TODO - is this needed?  2019-02-01
```
SimonRichardson: Is this `TODO` relevant...

state/watcher.go
```diff
@@ -893,38 +921,40 @@ func (w *modelFieldChangeWatcher) initial() (set.Strings, error) {
...
-		if _, ok := w.known[machineId]; ok {
-			logger.Tracef("stopped field change watching for machine %q", machineId)
+		if _, ok := w.known[docId]; ok {
+			logger.Tracef("stopped field change watching for %q", docId)
 		}
-		delete(w.known, machineId)
-		machineIds.Remove(machineId)
+		delete(w.known, docId)
+		// TODO - check this correct.
+		watchSet.Remove(docId)
```
SimonRichardson: So the idea behind this remove, is that we don't look for documents that are no longer needed. If those documents [do come back and they're classified as](https://github.com/juju/juju/pull/9699/files#diff-5a822a48b96dd79a90eda5ba852495c4R977) `!isKnown` (they're not known), then we re-add them in the merge. So in theory this shouldn't be an issue. 

Does that sound reasonable assumption?

SimonRichardson:
!!build!!

apiserver/facades/agent/provisioner/provisioner.go
hmlanigan: there is a ProvisionerAPIV7. CharmProfileChangeInfo above at 1501.

state/machine_test.go
hmlanigan: this was related to only returning a name from  NextUpgradeCharmProfileUnitName if 1 doc existed.  that functionality has been changed to allow more than one.  Added a test case to cover the change

api/provisioner/machine.go
```diff
@@ -652,12 +658,13 @@ func (m *Machine) SetUpgradeCharmProfileComplete(message string) error {
...
+func (m *Machine) RemoveUpgradeCharmProfileData(appName string) error {
 	var results params.ErrorResults
-	args := params.Entities{
-		Entities: []params.Entity{
+	args := params.ProfileArgs{
+		Args: []params.ProfileArg{
 			{
-				Tag: m.tag.String(),
+				Entity:   params.Entity{Tag: m.tag.String()},
+				UnitName: appName,
```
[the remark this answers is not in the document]
hmlanigan: updated

api/provisioner/machine.go
```diff
@@ -652,12 +658,13 @@ func (m *Machine) SetUpgradeCharmProfileComplete(message string) error {
 }
 
 // RemoveUpgradeCharmProfileData implements MachineProvisioner.RemoveUpgradeCharmProfileData.
-func (m *Machine) RemoveUpgradeCharmProfileData() error {
+func (m *Machine) RemoveUpgradeCharmProfileData(appName string) error {
```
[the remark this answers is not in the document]
hmlanigan: updated

api/provisioner/machine.go
```diff
@@ -627,13 +632,14 @@ func (m *Machine) SetCharmProfiles(profiles []string) error {
...
-func (m *Machine) SetUpgradeCharmProfileComplete(message string) error {
+func (m *Machine) SetUpgradeCharmProfileComplete(appName, message string) error {
 	var results params.ErrorResults
 	args := params.SetProfileUpgradeCompleteArgs{
 		Args: []params.SetProfileUpgradeCompleteArg{
 			{
-				Entity:  params.Entity{Tag: m.tag.String()},
-				Message: message,
+				Entity:   params.Entity{Tag: m.tag.String()},
+				UnitName: appName,
```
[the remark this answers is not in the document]
hmlanigan: updated

api/provisioner/machine.go
```diff
@@ -627,13 +632,14 @@ func (m *Machine) SetCharmProfiles(profiles []string) error {
 }
 
 // SetUpgradeCharmProfileComplete implements MachineProvisioner.SetUpgradeCharmProfileComplete.
-func (m *Machine) SetUpgradeCharmProfileComplete(message string) error {
+func (m *Machine) SetUpgradeCharmProfileComplete(appName, message string) error {
```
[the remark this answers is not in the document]
hmlanigan: updated

api/provisioner/machine.go
```diff
@@ -120,19 +120,19 @@ type MachineProvisioner interface {
...
 
 	// SetCharmProfiles records the given slice of charm profile names.
 	SetCharmProfiles([]string) error
 
-	// SetUpgradeCharmProfileComplete recorded that the result of updating
-	// the machine's charm profile(s)
-	SetUpgradeCharmProfileComplete(string) error
+	// SetUpgradeCharmProfileComplete records the result of updating
+	// the machine's charm profile(s), for the given unit.
+	SetUpgradeCharmProfileComplete(string, string) error
```
jameinel: I think putting the variable names into the interface will make it clearer which is the machine I'd and which is the unit.
I realize the other ones don't do this, but especially when there is more than one parameter of the same type.

apiserver/facades/agent/provisioner/provisioner.go
```diff
@@ -1442,13 +1531,43 @@ func (p *ProvisionerAPI) CharmProfileChangeInfo(machines params.Entities) (param
 	return params.ProfileChangeResults{Results: results}, nil
 }
 
-func machineChangeProfileChangeInfo(machine ProfileMachine, st ProfileBackend) (params.ProfileChangeResult, error) {
-	nothing := params.ProfileChangeResult{}
```
jameinel: calling this "ChangeProfileChangeInfo" makes me a bit concerned. "changed relative to what"?

Most operations we do, we want to get an idea of "this is the current state that we want" and then we compare it to the state we are currently at

the state we seem to want to compare to is lxc profile show (the api equivalent). Otherwise we risk half applying a profile and then we end up re applying it or never applying the second part.

I think that could help with issues about subordinates showing up unexpected, etc.

state/watcher.go
hmlanigan: I've resolved the TODO, removing WatchModelMachinesCharmProfiles as it's no longer called.  Only WatchContainersCharmProfiles is needed for compatibility.

state/machine.go
hmlanigan: added

state/machine.go
hmlanigan: added

state/machine.go
```diff
@@ -2232,10 +2155,164 @@ func (m *Machine) VerifyUnitsSeries(unitNames []string, series string, force boo
...
+	}
+	return instData, nil
+}
+
+func getNextInstanceCharmProfileData(st *State, id string, used bool) (instanceCharmProfileData, error) {
+	logger.Debugf("trying to get next instance charm profile data for machine %s here", id)
+	collection, closer := st.db().GetCollection(instanceCharmProfileDataC)
+	defer closer()
+
+	machineRegExp := fmt.Sprintf("^%s:%s#%s$", st.ModelUUID(), id, names.UnitSnippet)
```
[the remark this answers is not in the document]
hmlanigan: done.  added more comments as well.

api/provisioner/machine.go
hmlanigan: done

api/provisioner/machine.go
```diff
@@ -120,19 +120,19 @@ type MachineProvisioner interface {
 	WatchContainersCharmProfiles(ctype instance.ContainerType) (watcher.StringsWatcher, error)
 
 	// CharmProfileChangeInfo retrieves the info necessary to change a charm
-	// profile used by a machine.
-	CharmProfileChangeInfo() (CharmProfileChangeInfo, error)
+	// profile used by a machine, for the give unit.
```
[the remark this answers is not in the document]
hmlanigan: done

apiserver/facades/agent/provisioner/provisioner.go
hmlanigan: Talked about this offline and it's on the list of things to address in the future.

state/machine.go
```diff
@@ -2487,6 +2499,128 @@ func (m *Machine) RemoveUpgradeCharmProfileData() error {
...
+				return nil, errors.Trace(err)
+			}
+		}
+		life := m.Life()
+		if life == Dead || life == Dying {
+			return nil, ErrDead
+		}
+		docId := m.instanceCharmProfileDataId(unitName)
+		data, err := getInstanceCharmProfileData(m.st, docId)
+		if err != nil {
```
SimonRichardson: If the error is not found, should we not just return out here with `ErrNoOperations`, in case something has removed it behind our backs?

state/machine.go
```diff
@@ -2487,6 +2499,128 @@ func (m *Machine) RemoveUpgradeCharmProfileData() error {
...
+		return []txn.Op{
+			{
+				C:      machinesC,
+				Id:     m.doc.DocID,
+				Assert: isAliveDoc,
+			},
+			{
+				C:      instanceCharmProfileDataC,
+				Id:     docId,
+				Assert: bson.D{{"being-used", false}},
```
SimonRichardson: This is ok, if we ensure above that we can't get here easily, otherwise this maybe a cause of txn looping exhaustion.

hmlanigan: Which piece is the cause of concern?  All pieces of this are checked inside buildTxn() afaik, or am I missing something?

state/machine.go
hmlanigan: This is only called from NextUpgradeCharmProfileUnitName().  The Next*() calls won't remove behind our backs since being-used still false.  If something else deletes it, we're in trouble and should return an error.

state/machine.go
hmlanigan: Nothing missing.

hmlanigan:
$$merge$$

PR #9700, opened 2019-02-01
Support long tokens in dummy-sink and dummy-source

jameinel:
## Description of change

We are calling status-set, but if the token is 12kB it will do bad
things. So instead we print with a cut to shorter length. Also handle
when the token has spaces in it.

This is only a tweak to our dummy charms for testing purposes.

## QA steps

```
$ juju bootstrap lxd lxd --debug
$ juju deploy ./dummy-sink
$ juju deploy ./dummy-source
$ juju relate dummy-source dummy-sink
$ juju config dummy-source token="this is a very long token that would otherwise overflow status"
$ watch juju status
```

You should see the status message for dummy-source ends up being trimmed rather than causing any problems with the display.

## Documentation changes

None.

## Bug reference

None.


jameinel:
$$merge$$

PR #9702, opened 2019-02-01
Remove juju- prefix from k8s pod names

wallyworld:
## Description of change

When creating k8s pods, do not use "juju-" as a prefix.

## QA steps

deploy and manage a k8s model with apps


wallyworld:
$$merge$$

PR #9706, opened 2019-02-03
2.5 leases backoff

jameinel:
## Description of change

When getting a Claim from a client, lease.Manager would dutifully try to get the claim from the backend store. And if there was a problem getting the claim, the manager would retry the request. It was actually doing 2 layers of retries (AFAICT). One in a tight loop, and one in a slow-backoff loop if it got Timeout.

However, the issue is that it used the in-memory information about Leases to determine whether to be doing:
a) Claim() the lease, because there is no current holder
b) Extend() the lease, because *I'm* the current holder
c) Fail the claim because someone else is the current holder.

But the in-memory information is kept in sync via the raft log, asynchronously from doing any leadership claims. Which implies that if the lease synchronization was at all slow, then (a) or (b) would be happening, getting a "sorry you have made an invalid claim" response, and then immediately trying again based on the same out-of-date information. Now, a sync *could* occur during this time, but we didn't really wait for it in any meaningful fashion.

With the proposed change, we use the same "try to make a claim, but if it fails, wait with exponential backoff" before trying again, whether it was "that claim was invalid" or it was "that claim timed out". This should give better opportunities for the system to reach convergence. A slightly better solution for ErrInvalid would be to actually wait for the next sync Raft AppendEntries to occur. But we don't currently expose that.

## QA steps

I have not done manual QA of this yet.  My best guess is that you could introduce intentional lag in handling of AppendEntries and create lots of applications and units vying for leadership. You'd have to be in an HA controller. Then one of the secondary raft engines could be requested to claim a lease, and it would try to do so with outdated information. Note that you'd also need a fair bit of churn on the lease Expiry, because you need who-the-holder to actually be changing. (So you need something that is wanting leases, but failing to hold on to them properly, and something to introduce lag between the primary raft and secondary engines.)

Note that we *did* see this in Prodstack 4.5.

## Documentation changes

No user facing changes.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1814424


worker/lease/manager_claim_test.go
```diff
@@ -73,9 +74,20 @@ func (s *ClaimSuite) TestClaimLease_Success_SameHolder(c *gc.C) {
...
-		err := getClaimer(c, manager).Claim("redis", "redis/0", time.Minute)
-		c.Check(err, jc.ErrorIsNil)
+	fix.RunTest(c, func(manager *lease.Manager, clock *testclock.Clock) {
+		// On the first attempt, we don't see ourselves in the leases, so we try
+		// to Claim the lease. But Primary thinks we already have the lease, so it
+		// refuses. After claiming, we wait 50ms to let the refresh happen, then
+		// we notice that we are the holder, so we Extend instead of Claim.
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
```
babbageclunk: These are very cunning.

jameinel:
$$merge$$

PR #9707, opened 2019-02-03
raft-leases: don't try to expire leases when the store will expire them itself

babbageclunk:
## Description of change

The raft lease store will automatically expire leases as time advances - it doesn't need the lease manager to do it. Avoiding this reduces the number of store.Leases() calls the manager needs to make, which should make it faster under heavy load.

Add an Autoexpire() method to the lease store interface - this returns false for the mongo store but true for the raft lease store. If autoexpire is true for a store we still wake up for ticks (so we can signal blocks that are waiting for leases to expire) but we don't scan all the leases looking for ones that need to be expired.

## QA steps

* Bootstrap and deploy multiple units of an app.
* Remove the leader unit.
* The lease should expire as usual and one of the other units should claim leadership.
* There shouldn't be any errors in the lease manager log about it trying to expire leases.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/worker/lease/manager.go
+++ b/worker/lease/manager.go
@@ -175,8 +175,13 @@ func (manager *Manager) choose(blocks blocks) error {
 	case check := <-manager.checks:
 		return manager.handleCheck(check)
 	case manager.now = <-manager.nextTick(manager.now):
-		manager.wg.Add(1)
-		go manager.retryingTick(manager.now)
+		// If we need to expire leases we should do it, otherwise this
+		// is just an opportunity to check for blocks that need to be
+		// notified.
+		if !manager.config.Store.Autoexpire() {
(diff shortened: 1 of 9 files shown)


worker/lease/manager.go
```diff
@@ -175,8 +175,13 @@ func (manager *Manager) choose(blocks blocks) error {
 	case check := <-manager.checks:
 		return manager.handleCheck(check)
 	case manager.now = <-manager.nextTick(manager.now):
```
howbazaar: Doesn't nextTick still copy Leases?
Is there a way to not do this?

babbageclunk: Yeah, that's coming in the next PR.

babbageclunk:
$$merge$$

PR #9709, opened 2019-02-04
Use the newer testing clock.

jameinel:
## Description of change

The testing clock now iterates faster when given a long timeout, and
gives stack traces when the waiters don't match expectations.

This is just cleanup of the testing clock that helped when I was debugging something else, but we may as well get it into our core code. Only changes the testing clock, so it may change the test suite, but doesn't change production code.

https://github.com/juju/clock/pull/3
https://github.com/juju/clock/pull/4

We need (4) https://github.com/juju/juju/pull/9712

## QA steps

None.

## Documentation changes

None.

## Bug reference

None.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -434,14 +434,14 @@
   revision = "b8d279cc30ffcc57dab7a51c4bc39be08fb2b569"
 
 [[projects]]
-  digest = "1:acc1f840c3b249910754a9d5f5cb825674ca2d2011f185c6130b78f7c5ea0df8"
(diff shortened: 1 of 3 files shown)


jameinel:
It appears that we have quite a few places that expected being able to pass a timeout of '0s' to work correctly, which I apparently broke. The old code would always check 10 times, and call time.Sleep(0) each time. The new one notices the timeout immediately.

jameinel:
InstancePoller was the only place using that syntax, I just change it to use ShortWait.
$$merge$$

jameinel:
21:08:31 make: go: Command not found (doesn't sound like my code)

$$merge$$

PR #9710, opened 2019-02-04
Remove juju- from k8s artifact names

wallyworld:
## Description of change

When creating k8s artifacts like pods, volume claims, services etc, do not use the "juju-" prefix.
So that 2.5.0 deployed models continue to work, when an app is acted on, first see if it has been created using the legacy naming, and if so, continue to use the "juju-" prefixes for that app.

This started as just a change to k8s.go
When doing QA, I noticed that Juju doesn't handle well the case where the cluster is scaled down to 0 pods outside Juju. So add extra handling for that case.

## QA steps

bootstrap 2.5.0
deploy a k8s model with apps with storage
upgrade juju to 2.5.1
ensure that kubectl and juju status reflect a running model using the legacy names
scale out some apps and check
deploy some new apps with 2.5.1
ensure juju status shows a working system
scale out some apps and check

### DIFF
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -398,7 +398,7 @@ func (k *kubernetesClient) deleteSecret(imageSecretName string) error {
 // application exists.
 func (k *kubernetesClient) OperatorExists(appName string) (bool, error) {
 	statefulsets := k.AppsV1().StatefulSets(k.namespace)
-	_, err := statefulsets.Get(operatorName(appName), v1.GetOptions{IncludeUninitialized: true})
+	_, err := statefulsets.Get(k.operatorName(appName), v1.GetOptions{IncludeUninitialized: true})
 	if k8serrors.IsNotFound(err) {
 		return false, nil
 	}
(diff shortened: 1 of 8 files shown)


caas/kubernetes/provider/k8s.go
```diff
@@ -719,11 +729,15 @@ func (k *kubernetesClient) DeleteOperator(appName string) (err error) {
 		return errors.Trace(err)
 	}
 
+	deploymentName := appName
+	if legacy {
+		deploymentName = "juju-" + appName
```
SimonRichardson: Considering this is repeated a few times, do you think we should have a function for this `func legacyName(name string) string`, so we don't end up fat fingering this in the future?

caas/kubernetes/provider/k8s.go
```diff
@@ -1283,13 +1303,14 @@ func (k *kubernetesClient) deleteVolumeClaims(appName string, p *core.Pod) ([]st
...
-		valid := volMount.Name == operatorVolumeClaim(appName) || jujuPVNameRegexp.MatchString(volMount.Name)
-		if !valid {
-			logger.Debugf("ignoring non-Juju attachment %q", volMount.Name)
-			continue
-		}
+		//valid := volMount.Name == operatorVolumeClaim(appName) || pvNameRegexp.MatchString(volMount.Name)
+		//if !valid {
+		//	logger.Debugf("ignoring non-Juju attachment %q", volMount.Name)
+		//	continue
+		//}
```
SimonRichardson: Can we remove this comment?

caas/kubernetes/provider/k8s.go
```diff
@@ -1554,12 +1579,12 @@ func (k *kubernetesClient) Units(appName string) ([]caas.Unit, error) {
...
-				logger.Debugf("ignoring non-Juju attachment %q", volMount.Name)
-				continue
-			}
-			storageName := jujuPVNameRegexp.ReplaceAllString(volMount.Name, "$storageName")
+			//valid := pvNameRegexp.MatchString(volMount.Name)
+			//if !valid {
+			//	logger.Debugf("ignoring non-Juju attachment %q", volMount.Name)
+			//	continue
+			//}
+			//storageName := pvNameRegexp.ReplaceAllString(volMount.Name, "$storageName")
```
SimonRichardson: Again can we remove this comment?

wallyworld: doh, done

caas/kubernetes/provider/k8s.go
wallyworld: done

caas/kubernetes/provider/k8s.go
```diff
@@ -1283,13 +1302,14 @@ func (k *kubernetesClient) deleteVolumeClaims(appName string, p *core.Pod) ([]st
 		volumesByName[pv.Name] = pv
 	}
 
+	//	pvNameRegexp := jujuPVNameRegexp(appName)
```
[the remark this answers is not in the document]
wallyworld: fixed

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
[fix PVC, operator names](https://trello.com/c/2Tuo6z6J/608-fix-pvc-operator-names)

PR #9711, opened 2019-02-04
LXDProfile - Machine dead

SimonRichardson:
## Description of change

The following commit attempts to ensure that we don't process
dead machines when applying profiles. If the machine is dead,
we will continue onwards as if that machine is no longer around.

Note this means that we could end up leaving profiles in an orphaned
state. Cleaning up these orphaned profiles shouldn't do any harm,
as each one is unique to a application on a host/container. If
there is a chance that a conflict arises and all machines have been
blown away, then you can safely remove them directly from LXC.

## QA steps

TBA


SimonRichardson:
This has been updated to https://github.com/juju/juju/pull/9718

PR #9719, opened 2019-02-05
LXDProfile - Provision profile error

SimonRichardson:
## Description of change

The following commit changes how the instance poller works, because
of the fact that LXC reports a container running even if the LXD
profile applied failed.

So when applying a profile change to a container, the profile is
first added to LXD. The profile name is then added to the container
spec. Once the container is started it can sometimes error out
with an error. The problem occurs when the instance poller comes
around and asks for the instance status. LXD returns that response
as running, as it doesn't know that applying a profile that fails
should be a terminal state. We could attempt to apply the profile
again, but the LXD client doesn't tell us what the error is and
what was the cause. So retrying a profile might not be worth it,
missing device vs just unable to apply the profile at that time.

Instead we find out when the profile didn't apply and we put the
machine in to a profile error status, which can then be picked up
by the poller, which in turn can then work out if we should keep
the current error message and prevent it from become running or
we should move along.

The underlying fix for this is to get LXD to go into an ERROR state
if when applying a profile it fails, instead of a RUNNING state.

## QA steps

```sh
juju bootstrap lxd 
juju deploy ~/bundle-no-sub.yaml 
watch --color -n 1 juju status --color
```

Check that the error is kept when the machine `0` can't apply the
profile.

```
Model    Controller  Cloud/Region  Version  SLA          Timestamp
default  xxx         lxd/default   2.5.1.1  unsupported  17:42:05Z

App          Version  Status   Scale  Charm        Store  Rev  OS      Notes
lxd-profile  18.04    waiting    2/5  lxd-profile  local    0  ubuntu  
ubuntu       18.04    waiting    2/5  ubuntu-lite  local    0  ubuntu  

Unit            Workload  Agent       Machine  Public address  Ports  Message
lxd-profile/0*  active    idle        0        10.178.104.124         ready
lxd-profile/1   active    idle        1        10.178.104.226         ready
lxd-profile/2   waiting   allocating  2                               waiting for machine
lxd-profile/3   waiting   allocating  3                               waiting for machine
lxd-profile/4   waiting   allocating  4                               waiting for machine
ubuntu/0*       active    idle        0        10.178.104.124         ready
ubuntu/1        active    idle        1        10.178.104.226         ready
ubuntu/2        waiting   allocating  2                               waiting for machine
ubuntu/3        waiting   allocating  3                               waiting for machine
ubuntu/4        waiting   allocating  4                               waiting for machine

0        started  10.178.104.124  juju-be6237-0  bionic      cannot upgrade machine's lxd profile: Failed to add mount for device: Failed to run: /snap/lxd/current/bin/lxd forkmount lxc-mount juju-be6237-0 /var/snap/lxd/common/lxd/containers /var/snap/lxd/common/lxd/logs/juju-be6237-0/lxc.conf /var/snap/lxd/common/lxd/devices/juju-be6237-0/unix.tun.dev-net-tun /dev/net/tun none 4096: 
1        started  10.178.104.226  juju-be6237-1  bionic      Running
2        down                     pending        bionic      cannot upgrade machine's lxd profile: machine 2 not provisioned
3        down                     pending        bionic      cannot upgrade machine's lxd profile: machine 3 not provisioned
4        down                     pending        bionic      cannot upgrade machine's lxd profile: machine 4 not provisioned
```

bundle-no-sub.yaml
```yaml
series: bionic
applications:
  lxd-profile:
    charm: ~/go/src/github.com/juju/juju/testcharms/charm-repo/quantal/lxd-profile
    num_units: 9
    to:
    - "0"
    - "1"
    - "2"
    - "3"
    - "4"
    - "5"
    - "6"
    - "7"
    - "8"
  ubuntu:
    charm: cs:~jameinel/ubuntu-lite
    num_units: 9
    to:
    - "0"
    - "1"
    - "2"
    - "3"
    - "4"
    - "5"
    - "6"
    - "7"
    - "8"
machines:
  "0": {}
  "1": {}
  "2": {}
  "3": {}
  "4": {}
  "5": {}
  "6": {}
  "7": {}
  "8": {}
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1813044

### DIFF
--- a/apiserver/facades/agent/provisioner/provisioner.go
+++ b/apiserver/facades/agent/provisioner/provisioner.go
@@ -1341,7 +1341,10 @@ func (p *ProvisionerAPI) setOneInstanceStatus(canAccess common.AuthFunc, arg par
 		logger.Debugf("failed to SetInstanceStatus for %q: %v", mTag, err)
 		return err
 	}
+	// When the status passed is an error condition, also set the status to
+	// that error condition.
 	if status.Status(arg.Status) == status.ProvisioningError ||
+		status.Status(arg.Status) == status.ProvisioningProfileError ||
 		status.Status(arg.Status) == status.Error {
 		s.Status = status.Error
 		logger.Debugf("SetInstanceStatus triggering SetStatus for %#v", s)
--- a/core/status/status.go
+++ b/core/status/status.go
@@ -219,6 +219,12 @@ const (
 	Provisioning      Status = "allocating"
 	Running           Status = "running"
 	ProvisioningError Status = "provisioning error"
+
+	// ProvisioningProfileError represents a charm profile error when
+	// attemptting to apply a charm profile to a machine. The issue occurs when
+	// the LXD profile returns an error, but the LXC instance states that the
+	// container is running.
+	ProvisioningProfileError Status = "provisioning profile error"
 )
 
 const (
--- a/worker/instancepoller/machine_test.go
+++ b/worker/instancepoller/machine_test.go
@@ -110,6 +110,75 @@ func (s *machineSuite) testShortPoll(
 	clock.CheckCall(c, 1, "After", time.Duration(float64(ShortPoll)*ShortPollBackoff*ShortPollBackoff))
 }
 
+func (s *machineSuite) TestStatusWhenProvisioningProfileError(c *gc.C) {
+	context := &testMachineContext{
+		getInstanceInfo: func(id instance.Id) (instanceInfo, error) {
+			c.Check(id, gc.Equals, instance.Id("i1234"))
+			return instanceInfo{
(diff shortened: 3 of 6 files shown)


SimonRichardson:
Test timeout !!build!!

hmlanigan:
I'm thinking the juju status should be:
0        error  10.178.104.124  juju-be6237-0  bionic      cannot upgrade machine's lxd profile: ... 
that's what it's set to initially and when there is an upgrade failure, started being green makes you think everything is okay if looking quickly.

I haven't been able to test that juju upgrade-charm will clear the error state as I keep hitting other known issues.

SimonRichardson:
Closing this one as I'm going to supersede it.

PR #9722, opened 2019-02-06
Populate charm version for charms fetched from the charm store

achilleasa:
## Description of change

When deploying local charms that contain a `version` file, the charm version details are updated with the contents of that file. This was not the case when attempting to deploy charms from the charm store which results in the version details not being included in the `juju status` output.

This PR bumps the sha for the `charm.v6` dependency (see: https://github.com/juju/charm/pull/269) which provides a helper charm archive method for fetching the version of a charm and updates the relevant code to populate the charm version field prior to persisting its details to the store.

## QA steps

```console
$ juju deploy vault
$ juju status vault --format yaml | grep charm-version
    charm-version: 220f0ee
```

Note: you should get the same output when deploying a local charm that includes a `version` file.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1812227


achilleasa:
@SimonRichardson I think the status code will eventually read from mongo so I didn't add a test in there as the bug was caused by the code not persisting the version to the store. Happy to add an extra test if you think it's important.

achilleasa:
!!build!!

timClicks:
The current tests for status are quite brittle. They typically do exact string matching. So it's possible that this fix will actually break existing tests. because they are reliant on broken behaviour.

achilleasa:
@SimonRichardson I have pushed a new commit which hopefully should get all broken tests passing except these 2 that I am not really sure how to fix:

- https://github.com/juju/juju/blob/develop/cmd/juju/application/series_selector_test.go#L16 (most cases fail due to [this](https://github.com/juju/charm/blob/v6/charm.go#L54) check in `charm.v6`)
- https://github.com/juju/juju/blob/develop/cmd/juju/application/deploy_test.go#L720

Any ideas for fixing these ones?

achilleasa:
!!build!!

SimonRichardson:
@mitechie so when a charm is built it will inject a version file into the charm archive. The question is, if you create a charm without a build step the version file may or may not be included. If the version file is missing the status just hides the charm version, the charm creator can remedy this by putting a version file into the charm archive. Is this satisfactory?

achilleasa:
$$merge$$

achilleasa:
$$merge$$

achilleasa:
$$merge$$

PR #9730, opened 2019-02-10
worker/lease fix the handling of multiple concurrent errors

howbazaar:
This fix is for the observed failures we have seen in openstack deploys with solutions QA.

The uniter worker would get stuck setting up the initial remote state watcher. This watcher would try to claim the leadership, and the ClaimLeadership facade call would never return.

On the apiserver, the worker/lease was in a state where it was trying to restart, but waiting on the in flight wait group. If there were two concurrent claims, both of which hit errors, the first would be processed causing the worker/lease loop to exit. This triggered a defer on the wait group, which wasn't being decremented until the second in flight claim finished. However the second in flight one was blocked trying to send the error down the errors channel. This was no longer being selected on.

The fix is to not use an errors channel, but instead just kill the catacomb.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1815397

### DIFF
--- a/worker/lease/manager.go
+++ b/worker/lease/manager.go
@@ -76,7 +76,6 @@ func NewManager(config ManagerConfig) (*Manager, error) {
 		expireDone: make(chan struct{}),
 		pins:       make(chan pin),
 		unpins:     make(chan pin),
-		errors:     make(chan error),
 		logContext: logContext,
 	}
 	err := catacomb.Invoke(catacomb.Plan{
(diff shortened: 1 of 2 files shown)


babbageclunk:
$$merge$$

babbageclunk:
The blocked ClaimLeadership API request might explain the API server blocking when trying to restart in https://bugs.launchpad.net/juju/+bug/1813261

I'm not sure what the underlying error is - a timeout won't kill the manager. Once this fix is landed we'll potentially be able to see the manager crashing for some other reason.

PR #9731, opened 2019-02-11
juju status: don't pass on close errors to the user.

howbazaar:
Noticed on a solutions QA run that juju wait filed with an error trying to close the API.

There is nothing a user could do, and it is possible that the command otherwise worked fine.


howbazaar:
$$merge$$

PR #9732, opened 2019-02-11
Add tcp keep alive to client connections.

howbazaar:
This just adds keep alives at the tcp layer for client api connections.

We have observed in the wild situations where the client connection is trying to call the api, and is blocked in a low level tcp write. Somehow the tcp stack doesn't realise that the other side went away. TCP keep alives will help keep track of this.

### DIFF
--- a/api/apiclient.go
+++ b/api/apiclient.go
@@ -617,7 +617,9 @@ func gorillaDialWebsocket(ctx context.Context, urlStr string, tlsConfig *tls.Con
 	// TODO(rogpeppe) We'd like to set Deadline here
 	// but that would break lots of tests that rely on
 	// setting a zero timeout.
-	netDialer := net.Dialer{}
+	netDialer := net.Dialer{
+		KeepAlive: 30 * time.Second,
+	}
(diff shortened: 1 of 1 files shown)


howbazaar:
!!build!!

api/apiclient.go
```diff
@@ -617,7 +617,9 @@ func gorillaDialWebsocket(ctx context.Context, urlStr string, tlsConfig *tls.Con
 	// TODO(rogpeppe) We'd like to set Deadline here
 	// but that would break lots of tests that rely on
 	// setting a zero timeout.
-	netDialer := net.Dialer{}
+	netDialer := net.Dialer{
+		KeepAlive: 30 * time.Second,
```
wallyworld: Can we do this as a const at the top of the file where all the other such ones live.
Also, we have pingTimeout defined as 30s also. Maybe use the same value for keep alive?

wallyworld:
Also, should we use a keep alive for the raft transport worker dialer?

mitechie:
    Beep..Boop...

PR #9733, opened 2019-02-11
leases: Only run global-clock-updater if legacy-leases is set

babbageclunk:
## Description of change

If we're using raft leases there's no need to have the global-clock-updater writing to the db once a second.

## QA steps

* Spin up a controller.
* Check that global-clock-updater isn't running in the engine report.
* Switch to legacy-leases: `juju controller-config features="[legacy-leases]"`
* The engine report shows global-clock-updater is started.
* Switch back to raft leases: `juju controller-config features="[]"`
* global-clock-updater is stopped again.

## Documentation changes

None

## Bug reference

https://bugs.launchpad.net/juju/+bug/1813996

### DIFF
--- a/cmd/jujud/agent/machine/manifolds.go
+++ b/cmd/jujud/agent/machine/manifolds.go
@@ -30,6 +30,7 @@ import (
 	"github.com/juju/juju/core/machinelock"
 	"github.com/juju/juju/core/presence"
 	"github.com/juju/juju/core/raftlease"
+	"github.com/juju/juju/feature"
 	"github.com/juju/juju/state"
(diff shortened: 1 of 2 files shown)


babbageclunk:
$$merge$$

babbageclunk:
Well, there are tests for the dependencies (updated) and tests for the feature flag worker, so I don't *think* we need a test for this specific case?

babbageclunk:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #9735, opened 2019-02-11
Update mongo ulimits.

howbazaar:
Make limits support unlimited, and update mongo config based on mongo recommendations.

In order to support "unlimited", had to change the limit map in the service code to use strings.

### DIFF
--- a/service/agent.go
+++ b/service/agent.go
@@ -14,8 +14,6 @@ import (
 )
 
 const (
-	maxAgentFiles = 20000
-
 	agentServiceTimeout = 300 // 5 minutes
 )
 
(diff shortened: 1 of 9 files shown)


service/agent.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	maxAgentFiles = 20000
+	maxAgentFiles = "64000"
```
wallyworld: Please add comment to explain why not an int

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #9739, opened 2019-02-11
MongoVersion: Check BestAPIVersion before making a call

SimonRichardson:
## Description of change

The following ensures that we check if we can even make the call
before calling it. Also hide the mongo version if it's empty, cleaning
up the output from the API.

## QA steps

Bootstap 2.5 branch
```
juju bootstrap lxd old
```

Upgrade to juju develop/edge
```
juju show-controller
```

### DIFF
--- a/cmd/juju/controller/showcontroller.go
+++ b/cmd/juju/controller/showcontroller.go
@@ -90,6 +90,7 @@ type ControllerAccessAPI interface {
 	ModelStatus(models ...names.ModelTag) ([]base.ModelStatus, error)
 	AllModels() ([]base.UserModel, error)
 	MongoVersion() (string, error)
+	BestAPIVersion() int
 	Close() error
 }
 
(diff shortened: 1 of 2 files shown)


SimonRichardson:
https://trello.com/c/LLYpkOMy/1076-ensure-mongo-version-works-with-older-controllers

cmd/juju/controller/showcontroller.go
```diff
@@ -168,6 +169,23 @@ func (c *showControllerCommand) Run(ctx *cmd.Context) error {
...
+		if apiVersion := client.BestAPIVersion(); apiVersion >= 6 {
+			mongoVersion, err = client.MongoVersion()
+			if err != nil {
+				details.Errors = append(details.Errors, err.Error())
+				continue
+			}
+		}
+
+		// Update client store.
+		if err := c.SetControllerModels(c.store, controllerName, allModels); err != nil {
```
achilleasa: I think this block should be removed; it fails with a permission error if the user has no SuperuserAccess so I had that moved to L164 to keep things working for users with login (or with the add-model) permission.

SimonRichardson: You're right it's a rebase issue.

achilleasa:
LGTM. My only concern is that the minimum API version for the call is hardcoded at the call site **and** inside the api -> facade [bridge](https://github.com/juju/juju/blob/develop/api/controller/mongo.go) code which could prove to be problematic if we plan on calling this from different locations.

SimonRichardson:
$$merge$$

PR #9740, opened 2019-02-12
Fix cmr relations where there are 2 offers on the one app

wallyworld:
## Description of change

Creating 2 offers off the one app was failing when relations were processed. To fix, change the token generated on the offer side to be based on the offer name, not the app name.

Also, fix the offer status watcher so that status changes to an offering app's units also trigger a status update in the consuming model.

Note: existing single app offers will continue to work with this fix (it's backwards compatible).
But multi-offers on the same app will need to be redeployed.

## QA steps

juju bootstrap lxd test
juju switch controller
juju model-config logging-juju switch default
juju deploy memcached
juju deploy mariadb
juju add-model foo
juju deploy mediawiki
juju offer mediawiki:cache mw-cache
juju offer mediawiki:db mw-db
juju switch default
juju consume foo.mw-db
juju consume foo.mw-cache
juju relate mw-db mariadb:db
juju relate mw-cache memcached

Check controller logs for cmr relation errors.
Check workloads

juju remove-relation mariadb:db mw-db

Check SAAS status

Upgrade a 2.5.0 cmr deployment and check operations.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1813151
https://bugs.launchpad.net/juju/+bug/1815179

### DIFF
--- a/apiserver/facades/controller/crossmodelrelations/crossmodelrelations.go
+++ b/apiserver/facades/controller/crossmodelrelations/crossmodelrelations.go
@@ -271,14 +271,27 @@ func (api *CrossModelRelationsAPI) registerRemoteRelation(relation params.Regist
 	}
 	logger.Debugf("relation token %v exported for %v ", relation.RelationToken, localRel.Tag().Id())
 
-	// Export the local application from this model so we can tell the caller what the remote id is.
+	// Export the local offer from this model so we can tell the caller what the remote id is.
+	// The offer is exported as an application name since it models the behaviour of an application
+	// as far as the consuming side is concerned, and also needs to be unique.
+	// This allows > 1 offers off the one application to be made.
 	// NB we need to export the application last so that everything else is in place when the worker is
 	// woken up by the watcher.
-	token, err := api.st.ExportLocalEntity(names.NewApplicationTag(localApplicationName))
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return nil, errors.Annotatef(err, "exporting local application %v", localApplicationName)
+
+	// Juju 2.5.1 and earlier exported the local application name so for backwards compatibility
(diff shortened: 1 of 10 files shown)


wallyworld:
[cmr fixes](https://trello.com/c/8i9DQnCk/611-cmr-fixes)

apiserver/facades/controller/crossmodelrelations/crossmodelrelations.go
```diff
@@ -271,14 +271,27 @@ func (api *CrossModelRelationsAPI) registerRemoteRelation(relation params.Regist
...
+	// This allows > 1 offers off the one application to be made.
 	// NB we need to export the application last so that everything else is in place when the worker is
 	// woken up by the watcher.
-	token, err := api.st.ExportLocalEntity(names.NewApplicationTag(localApplicationName))
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return nil, errors.Annotatef(err, "exporting local application %v", localApplicationName)
+
+	// Juju 2.5.1 and earlier exported the local application name so for backwards compatibility
+	// use that if it's there.
+	token, err := api.st.GetToken(names.NewApplicationTag(localApplicationName))
```
babbageclunk: Will this do the right thing if there's an offer created  in 2.5.1 or earlier and then another offer for the same app is created in 2.5.2+? It seems like the new offer would also pick up the application-based token. Maybe converting back from token to offer and checking it's the one we expect would solve that?

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #9741, opened 2019-02-12
Update controller api to indicate configured identity provider

achilleasa:
## Description of change

This PR adds a new method to the controller API that returns the value for the `identity-url` config param that was specified while bootstrapping the controller or a blank value if no `identity-url` was specified. Contrary to other controller API methods, the implementation of `IdentityProviderURL` **does not** require `SuperuserAccess` for the caller as the identity URL is more or less public and known even to users with just login access.

The new method can be used by the UI to decide whether to display the _login with USSO_ link.

Finally, the `show-controller` implementation has been updated to include `identity-url` (if non-empty) into its output.

## QA steps

```console
$ juju bootstrap localhost lxd-no-ident
$ juju show-controller --format yaml | grep identity-url
# no output

$ juju bootstrap localhost lxd-with-ident --config identity-url=https://api.jujucharms.com/identity
$ juju show-controller --format yaml | grep identity-url
 identity-url: https://api.jujucharms.com/identity

# Check for non-admin users too
$ juju add-user foo
$ # set passwords and logout admin
$ juju login -c lxd-with-ident -u foo
$ juju show-controller --format yaml | grep identity-url
 identity-url: https://api.jujucharms.com/identity
```

## Bug reference
https://bugs.launchpad.net/juju/+bug/1771618

### DIFF
--- a/api/controller/identity_url.go
+++ b/api/controller/identity_url.go
@@ -1,0 +1,27 @@
+// Copyright 2012-2019 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package controller
+
+import (
+	"github.com/juju/errors"
+	"github.com/juju/juju/apiserver/params"
+)
+
+// IdentityProviderURL returns the URL of the configured external identity
+// provider for this controller or an empty string if no external identity
+// provider has been configured when the controller was bootstrapped.
+func (c *Client) IdentityProviderURL() (string, error) {
+	if c.BestAPIVersion() < 7 {
+		return "", errors.NotSupportedf("IdentityProviderURL not supported by this version of Juju")
+	}
+	var result params.StringResult
+	err := c.facade.FacadeCall("IdentityProviderURL", nil, &result)
+	if err != nil {
+		return "", errors.Trace(err)
+	}
+	if result.Error != nil {
+		return "", result.Error
+	}
+	return result.Result, nil
+}
--- a/api/firewaller/application_test.go
+++ b/api/firewaller/application_test.go
@@ -42,6 +42,8 @@ func (s *applicationSuite) TestTag(c *gc.C) {
 }
 
 func (s *applicationSuite) TestWatch(c *gc.C) {
+	s.WaitForModelWatchersIdle(c, s.Model.UUID())
+
 	w, err := s.apiApplication.Watch()
 	c.Assert(err, jc.ErrorIsNil)
 	wc := watchertest.NewNotifyWatcherC(c, w, s.BackingState.StartSync)
--- a/apiserver/facades/client/controller/controller.go
+++ b/apiserver/facades/client/controller/controller.go
@@ -21,6 +21,7 @@ import (
 	"github.com/juju/juju/apiserver/common/cloudspec"
 	"github.com/juju/juju/apiserver/facade"
 	"github.com/juju/juju/apiserver/params"
+	corecontroller "github.com/juju/juju/controller"
 	coremigration "github.com/juju/juju/core/migration"
 	"github.com/juju/juju/migration"
 	"github.com/juju/juju/permission"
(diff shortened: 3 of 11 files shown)


achilleasa:
!!build!!

achilleasa:
!!build!!

api/controller/identity_url.go
```diff
@@ -0,0 +1,27 @@
...
+package controller
+
+import (
+	"github.com/juju/errors"
+	"github.com/juju/juju/apiserver/params"
+)
+
+// IdentityProviderURL returns the URL of the configured external identity
+// provider for this controller or an empty string if no external identity
+// provider has been configured when the controller was bootstrapped.
```
SimonRichardson: Can you set the external identity url using controller-config?

api/firewaller/application_test.go
```diff
@@ -42,6 +42,8 @@ func (s *applicationSuite) TestTag(c *gc.C) {
 }
 
 func (s *applicationSuite) TestWatch(c *gc.C) {
+	s.WaitForModelWatchersIdle(c, s.Model.UUID())
```
SimonRichardson: Nice, hopefully this will fix the flaky test for the firewaller.

apiserver/facades/client/controller/controller.go
```diff
@@ -63,8 +70,8 @@ type ControllerAPIv3 struct {
 	*ControllerAPIv4
 }
 
-// NewControllerAPIv6 creates a new ControllerAPIv6
-func NewControllerAPIv6(ctx facade.Context) (*ControllerAPI, error) {
+// NewControllerAPIv7 creates a new ControllerAPIv7.
```
SimonRichardson: Can we add a note about what was added between v6 and v7. I find that useful when scanning the code.

cmd/juju/controller/showcontroller.go
```diff
@@ -167,19 +169,25 @@ func (c *showControllerCommand) Run(ctx *cmd.Context) error {
...
+				if err != nil {
+					details.Errors = append(details.Errors, err.Error())
+					continue
+				}
 			}
 		}
 
+		// Fetch identityURL if the apiserver supports it
+		identityURL, err := client.IdentityProviderURL()
+		if err != nil && !errors.IsNotSupported(err) {
```
SimonRichardson: I do find this much cleaner, maybe we should change the mongo version so it uses the same pattern?

api/controller/identity_url.go
achilleasa: AFAICT, you can only specify this when bootstrapping and then becomes read-only (see [here](https://github.com/juju/juju/blob/2fb13bbdc738665f08e6f98e7da5f692c1340cbc/controller/config.go#L279))

apiserver/facades/client/controller/controller.go
achilleasa: I followed the pattern of documenting updates at the struct def level (changes are documented [here](https://github.com/juju/juju/pull/9741/files/a448a3e843c48428e57a6e854ef3fdbbab147bbb#diff-b5ab52656a34e521b266f76098f4d8e9R49))

cmd/juju/controller/showcontroller.go
achilleasa: Sure, let me update my PR

apiserver/facades/client/controller/controller.go
SimonRichardson: Ah, I missed it, sorry.

api/controller/identity_url.go
SimonRichardson: TIL'd

cmd/juju/controller/showcontroller_test.go
```diff
@@ -524,6 +539,10 @@ func (c *fakeController) BestAPIVersion() int {
 	return c.bestAPIVersion
```
SimonRichardson: I think we can get rid of this now, including removing it from the interface.

achilleasa:
!!build!!

achilleasa:
!!build!!

achilleasa:
!!build!!

mitechie:
QA ok thank you!

achilleasa:
$$merge$$

PR #9754, opened 2019-02-15
Handle bounced pods with storage better

wallyworld:
## Description of change

A few storage related fixes. If a pod with storage bounces, do not delete it immediately from Juju. This allows the unit name to remain unchanged when the pod comes up again. Previously the unit name would get an new incremented id. The same applies to the storage id.

When detaching a volume, do not delete the PVC. This allows the volume to be re-attached later. The PVC is deleted when the volume is deleted by Juju.

Status is enhanced so that we a pod does terminate, if there are other alive pods the overall app status is still shown as active, not terminated. And we also exclude terminated units from the running count in status so the "Scale" information is correctly represented.

## QA steps

bootstrap k8s
deploy mariadb and scale out
manually kill a pod and see that the pod comes back and juju status reports the same unit and storage ids as before
remove mariadb and see the pvc and pv remain
deploy mariadb again and check that new storage and volumes are created
remove mariadb with --destroy-storage and see that the pvc and pv are deleted


apiserver/facades/controller/caasunitprovisioner/provisioner.go
```diff
@@ -796,8 +815,13 @@ func (a *Facade) updateStateUnits(app Application, unitInfo *updateStateUnitPara
 				if err != nil {
 					return errors.Trace(err)
 				}
-				if si.StorageName() != storageName {
-					continue
+				id := si.StorageTag().Id()
+				id = strings.Replace(id, "/", "", -1)
+				if id != storageId {
+					// Old k8s deployments used storage name.
```
ycliuhw: do we wanna log here?

wallyworld:
$$merge$$

wallyworld:
[removing a k8s application destroys storage](https://trello.com/c/PGXs1KUZ/614-removing-a-k8s-application-destroys-storage)

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #9757, opened 2019-02-15
Application charm profile

SimonRichardson:
## Description of change

The following exports the charm-profile if it's not empty. We
can't use valid, as that always passes back a profile even if it's
empty. So in this instance it's more important to know if it's
empty and if it's not empty then show the application-charm profile.

The driver behind this is so we can marry what is intended to be
applied vs what is actually been applied from just looking at the
status output.

## QA steps

```
juju bootstrap lxd
juju status --format=yaml
```

Expected output:
```yaml
model:
  name: default
  type: iaas
  controller: yyy
  cloud: lxd
  region: default
  version: 2.5.1.1
  model-status:
    current: available
    since: 15 Feb 2019 13:45:28Z
  sla: unsupported
machines:
  "0":
    juju-status:
      current: started
      since: 15 Feb 2019 13:48:05Z
      version: 2.5.1.1
    dns-name: 10.178.104.99
    ip-addresses:
    - 10.178.104.99
    instance-id: juju-e1e85b-0
    machine-status:
      current: running
      message: Running
      since: 15 Feb 2019 13:46:53Z
    series: bionic
    network-interfaces:
      eth0:
        ip-addresses:
        - 10.178.104.99
        mac-address: 00:16:3e:40:fd:93
        gateway: 10.178.104.1
        is-up: true
    hardware: arch=amd64 cores=0 mem=0M
    lxd-profiles:
      juju-default-lxd-profile-0:
        config:
          environment.http_proxy: ""
          linux.kernel_modules: openvswitch,nbd,ip_tables,ip6_tables
          security.nesting: "true"
          security.privileged: "true"
        description: lxd profile for testing, will pass validation
        devices:
          bdisk:
            source: /dev/loop0
            type: unix-block
          gpu:
            type: gpu
          sony:
            productid: 51da
            type: usb
            vendorid: 0fce
          tun:
            path: /dev/net/tun
            type: unix-char
applications:
  lxd-profile:
    charm: local:bionic/lxd-profile-0
    series: bionic
    os: ubuntu
    charm-origin: local
    charm-name: lxd-profile
    charm-rev: 0
    charm-profile: juju-default-lxd-profile-0
    exposed: false
    application-status:
      current: active
      message: 'load: 1.13, 1.42, 1.64'
      since: 15 Feb 2019 16:35:57Z
    units:
      lxd-profile/0:
        workload-status:
          current: active
          message: 'load: 1.13, 1.42, 1.64'
          since: 15 Feb 2019 16:35:57Z
        juju-status:
          current: idle
          since: 15 Feb 2019 13:48:10Z
          version: 2.5.1.1
        leader: true
        machine: "0"
        public-address: 10.178.104.99
    version: "18.04"
    endpoint-bindings:
      another: ""
      ubuntu: ""
storage: {}
controller:
  timestamp: 16:36:36Z
```

### DIFF
--- a/core/lxdprofile/validate.go
+++ b/core/lxdprofile/validate.go
@@ -46,12 +46,12 @@ func ValidateCharmInfoLXDProfile(info *apicharms.CharmInfo) error {
 	return nil
 }
 
-// IsEmpty will return true if the profiler containers a profile, that is
-// empty. If the profile isn't empty, we'll return false.
-// If there is no valid profile in the profiler, it will return true
-func IsEmpty(profiler charm.LXDProfiler) bool {
+// NotEmpty will return false if the profiler containers a profile, that is
+// empty. If the profile is empty, we'll return false.
+// If there is no valid profile in the profiler, it will return false
+func NotEmpty(profiler charm.LXDProfiler) bool {
 	if profile := profiler.LXDProfile(); profile != nil {
-		return len(profile.Devices) == 0 && len(profile.Config) == 0 && len(profile.Description) == 0
+		return !profile.Empty()
 	}
-	return true
+	return false
 }
--- a/core/lxdprofile/validate.go
+++ b/core/lxdprofile/validate.go
@@ -45,3 +45,13 @@ func ValidateCharmInfoLXDProfile(info *apicharms.CharmInfo) error {
 	}
 	return nil
 }
+
+// IsEmpty will return true if the profiler containers a profile, that is
+// empty. If the profile isn't empty, we'll return false.
+// If there is no valid profile in the profiler, it will return true
+func IsEmpty(profiler charm.LXDProfiler) bool {
+	if profile := profiler.LXDProfile(); profile != nil {
+		return len(profile.Devices) == 0 && len(profile.Config) == 0 && len(profile.Description) == 0
+	}
+	return true
+}
--- a/apiserver/facades/client/client/status.go
+++ b/apiserver/facades/client/client/status.go
@@ -1008,8 +1008,7 @@ func (context *statusContext) processApplication(application *state.Application)
 	}
 
 	var charmProfileName string
-	// Ensure the profile isn't empty.
-	if !lxdprofile.IsEmpty(applicationCharm) {
+	if lxdprofile.NotEmpty(applicationCharm) {
 		charmProfileName = lxdprofile.Name(context.model.Name(), application.Name(), applicationCharm.Revision())
 	}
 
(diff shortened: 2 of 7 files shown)


SimonRichardson:
https://trello.com/c/m7IodYEp/1097-1813044-follow-onexpose-charm-profile-name-on-the-application

core/lxdprofile/validate.go
```diff
@@ -45,3 +45,13 @@ func ValidateCharmInfoLXDProfile(info *apicharms.CharmInfo) error {
 	}
 	return nil
 }
+
+// IsEmpty will return true if the profiler containers a profile, that is
+// empty. If the profile isn't empty, we'll return false.
+// If there is no valid profile in the profiler, it will return true
+func IsEmpty(profiler charm.LXDProfiler) bool {
+	if profile := profiler.LXDProfile(); profile != nil {
+		return len(profile.Devices) == 0 && len(profile.Config) == 0 && len(profile.Description) == 0
```
hmlanigan: why not use profile.Empty() here?

core/lxdprofile/validate.go
```diff
@@ -45,3 +45,13 @@ func ValidateCharmInfoLXDProfile(info *apicharms.CharmInfo) error {
 	}
 	return nil
 }
+
+// IsEmpty will return true if the profiler containers a profile, that is
+// empty. If the profile isn't empty, we'll return false.
+// If there is no valid profile in the profiler, it will return true
+func IsEmpty(profiler charm.LXDProfiler) bool {
```
hmlanigan: Suggest rename to NotEmpty and reverse the logic.  We'll always be doing !lxdprofile.IsEmpty() I think.

This will be a nice addition.  Will have to update the other bits of code. I had to go searching for the line: if profile := profiler.LXDProfile(); profile != nil && !profile.Empty() the other day to remember the combo.

core/lxdprofile/validate.go
SimonRichardson: Because I wanted to be sure description was empty as well, tbh I could combine them...

core/lxdprofile/validate.go
SimonRichardson: Will do

core/lxdprofile/validate.go
SimonRichardson: I've thought about this more and we should be consistent with that `profile.Empty()` is. I'll update the tests.

SimonRichardson:
$$merge$$

PR #9766, opened 2019-02-20
Ignore storage failures when forcing machine removal.

anastasiamac:
## Description of change

We have noticed that the users could not remove a machine, even with --force, when there were storage failures.

In fact, both storage failures and a failure to set the option to keep cloud instance were not compliant with Juju users --force expectation.

This PR ensures that when --force is desired, all errors are ignored. 

## Bug reference

There are some bugs related to this failure but the root cause pointing at the storage related failures have not been identified.

### DIFF
--- a/apiserver/facades/client/machinemanager/machinemanager.go
+++ b/apiserver/facades/client/machinemanager/machinemanager.go
@@ -292,7 +292,7 @@ func (mm *MachineManagerAPI) destroyMachine(args params.Entities, force, keep bo
 				if !force {
 					return fail(err)
 				}
-				logger.Warningf("error destroying machine %v but keep instance: %v", machineTag.Id(), err)
+				result.Error = common.ServerError(err)
 			}
 		}
 		var info params.DestroyMachineInfo
--- a/apiserver/facades/client/machinemanager/machinemanager.go
+++ b/apiserver/facades/client/machinemanager/machinemanager.go
@@ -289,7 +289,10 @@ func (mm *MachineManagerAPI) destroyMachine(args params.Entities, force, keep bo
 		if keep {
 			logger.Infof("destroy machine %v but keep instance", machineTag.Id())
 			if err := machine.SetKeepInstance(keep); err != nil {
-				return fail(err)
+				if !force {
+					return fail(err)
+				}
+				logger.Warningf("error destroying machine %v but keep instance: %v", machineTag.Id(), err)
 			}
 		}
 		var info params.DestroyMachineInfo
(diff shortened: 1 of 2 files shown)


anastasiamac:
[destroy machine and -force when storage erred](https://trello.com/c/fh0qPOj7/617-destroy-machine-and-force-when-storage-erred)

anastasiamac:
mongo failure on github.com/juju/juju/worker/featureflag.TestPackage ?!
!!build!!

apiserver/facades/client/machinemanager/machinemanager.go
```diff
@@ -271,26 +271,41 @@ func (mm *MachineManagerAPI) destroyMachine(args params.Entities, force, keep bo
...
+			return fail(err)
 		}
 		if keep {
 			logger.Infof("destroy machine %v but keep instance", machineTag.Id())
 			if err := machine.SetKeepInstance(keep); err != nil {
-				return nil, err
+				if !force {
+					return fail(err)
+				}
+				result.Error = common.ServerError(err)
```
howbazaar: Surely if we are forcing we don't want an error in result.Error? That is what fail is doing right?

apiserver/facades/client/machinemanager/machinemanager.go
```diff
@@ -318,28 +334,34 @@ func (mm *MachineManagerAPI) destroyMachine(args params.Entities, force, keep bo
...
 			info.DestroyedStorage = append(info.DestroyedStorage, destroyed...)
 			info.DetachedStorage = append(info.DetachedStorage, detached...)
 		}
+
+		if len(storageErrors) != 0 {
+			all := params.ErrorResults{storageErrors}
+			if !force {
+				return fail(all.Combine())
+			}
+			result.Error = common.ServerError(all.Combine())
```
howbazaar: same here?

apiserver/facades/client/machinemanager/machinemanager.go
anastasiamac: This error does not prevent the actual destroy call to take place. This is just a returned result...

mitechie:
I just want to make sure we don't just ignore errors but at least output at
the end of the command all errors hit and note that the user might want to
check up on those aspects. e.g. if we hit storage issues we couldn't remove
we'd note that so the user can go manually check the storage situation and
do any manual cleanup, same with something like remove-machine for an
instance not found (e.g. someone removed it with cloud tools vs juju) so
they could identify/clean up the instance if there was an issue.

anastasiamac:
@wallyworld,
The errors are not put into the Result struct because the clients interpret the presence of errors as a failed operation.   I have initially put into a Result struct but @howbazaar pointed out that it will produce an unexpected behavior :D we'll force remove the machine successfully but will report that we did not.

So instead now, we'll log any encountered errors at a WARN and will keep going with the force as expected.

anastasiamac:
$$merge$$

anastasiamac:
unrelated, intermediate failure in github.com/juju/juju/worker/peergrouper.TestPackage

anastasiamac:
$$merge$$

PR #9767, opened 2019-02-20
When add-k8s is run, ensure default storage exists or is created

wallyworld:
## Description of change

When adding a k8s cluster, query the cluster to get metadata about the cloud/region and also any default storage class. We compare any default storage with what our opinionated defaults for the underlying substrate and if different, prompt the user to create a suitable storage class.

## QA steps

Run juju add-k8s on microk8s which does have a suitable storage class, as well as a CDK deployment on GCE, with and without a suitable storage class.

## Bug

https://bugs.launchpad.net/juju/+bug/1814678

### DIFF
--- a/caas/kubernetes/provider/init.go
+++ b/caas/kubernetes/provider/init.go
@@ -15,6 +15,7 @@ const (
 )
 
 var k8sCloudCheckers map[string]k8slabels.Selector
+var clusterPreferredWorkloadStorage map[string]caas.PreferredStorage
 
 func init() {
 	caas.RegisterContainerProvider(providerType, providerInstance)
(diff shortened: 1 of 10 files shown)


wallyworld:
[create opinionated k8s storage on cluster when add-k8s is run](https://trello.com/c/FfQwlfw0/619-create-opinionated-k8s-storage-on-cluster-when-add-k8s-is-run)

caas/kubernetes/provider/metadata.go
```diff
@@ -39,5 +46,98 @@ func getCloudProviderFromNodeMeta(node core.Node) string {
...
+	for _, n := range nodes.Items {
+		var cloudRegion, v string
+		var ok bool
+		if v = getCloudProviderFromNodeMeta(n); v == "" {
+			continue
+		}
+		cloudRegion += v
+		if v, ok = n.Labels[regionLabelName]; ok && v != "" {
+			cloudRegion += "/" + v
+		}
```
ycliuhw: So in `<cloud>/<region>`, both `<cloud>` and `<region>` were required. Do we allow here can only return `<cloud>` without `<region>` now?

caas/kubernetes/provider/metadata_test.go
```diff
@@ -4,11 +4,27 @@
 package provider_test
 
 import (
-	"github.com/juju/juju/caas/kubernetes/provider"
+	"os"
+
+	"github.com/golang/mock/gomock"
+	"github.com/juju/collections/set"
+	"github.com/juju/errors"
+	"github.com/juju/juju/caas"
```
ycliuhw: move me

caas/kubernetes/provider/metadata.go
wallyworld: Yes. Some clouds don't have regions yet, like microk8s. We only need the region for JAAS. But we always need the cloud so we can see what the storage needs to be.

caas/kubernetes/provider/metadata_test.go
wallyworld: done

caas/kubernetes/provider/metadata.go
ycliuhw: ic thx

caas/kubernetes/provider/init.go
```diff
@@ -15,13 +15,35 @@ const (
...
+	// clusterPreferredWorkloadStorage defines the opinionated storage
+	// that Juju requires to be available on supported clusters.
+	clusterPreferredWorkloadStorage = map[string]caas.PreferredStorage{
+		"microk8s": {
+			Name:        "hostpath",
+			Provisioner: "microk8s.io/hostpath",
+		},
+		"gce": {
+			Name:        "GCE Persistent Disk",
+			Provisioner: "kubernetes.io/gce-pd",
```
ycliuhw: we will need to add `Parameters` here

wallyworld:
$$merge$$

PR #9778, opened 2019-02-22
Fixes critical lp#1786140: Current controller should be determined correctly.

anastasiamac:
## Description of change

Previous behavior of environment variables had produces surprising for user results.

Juju used $JUJU_CONTROLLER environment variable for controller level commands and not take $JUJU_MODEL value into consideration, even if the controller was supplied there in the form of `<controller name>:<model name>`. 

At the same time, model level commands completely ignored $JUJU_CONTROLLER environment variable even if it was set.

This changes this behavior - current controller, whenever needed i.e. irrespective of the command level, is determined by looking at $JUJU_MODEL env var first, then $JUJU_CONTROLLER and lastly by the last outcome of running 'juju switch'.

The implicit fix here is also that if env var points to the controller that this client is not aware of (for example, 'juju controllers --refresh' needs to be run) Juju will now fail earlier.

In a follow-up PR, I'll be checking whether we always use $JUJU_MODEL correctly to determine current model. The logic will be similar.

## QA steps

1. bootstrap several controllers
2. run 'juju switch' to switch to one of them
3. set $JUJU_CONTROLLER to point to a different controller
4. all subesquent commands should be running against the controller set by env var

Repeat the same by specifying different controller in $JUJU_MODEL via <controller name>:<model name> format. All coomands should run against the controller specified in $JUJU_MODEL.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1786140
https://bugs.launchpad.net/juju/+bug/1813079

### DIFF
--- a/cmd/juju/backups/restore_test.go
+++ b/cmd/juju/backups/restore_test.go
@@ -47,7 +47,7 @@ func (s *restoreSuite) SetUpTest(c *gc.C) {
 
 	controllerName := "test-master"
 	s.store = jujuclient.NewMemStore()
-	s.store.Controllers["testing"] = jujuclient.ControllerDetails{
+	s.store.Controllers[controllerName] = jujuclient.ControllerDetails{
 		ControllerUUID: controllerUUID,
 		CACert:         testing.CACert,
 		Cloud:          "mycloud",
--- a/cmd/juju/cloud/showcredential.go
+++ b/cmd/juju/cloud/showcredential.go
@@ -102,7 +102,7 @@ type CredentialContentAPI interface {
 }
 
 func (c *showCredentialCommand) NewCredentialAPI() (CredentialContentAPI, error) {
-	currentController, err := c.store.CurrentController()
+	currentController, err := modelcmd.DetermineCurrentController(c.store)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return nil, errors.New("there is no active controller")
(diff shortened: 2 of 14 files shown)


anastasiamac:
[current controller saga](https://trello.com/c/yWRMkBM6/620-current-controller-saga)

anastasiamac:
win failed due to an unrelated, intermittent, old-non-favourite "Test left sockets in a dirty state"

anastasiamac:
!!build!!

anastasiamac:
win failures both "dirty socket" and [LOG] 0:02.194 DEBUG juju.state closed state without error
d:/payloads/1550826189/build/src/github.com/juju/juju/juju/testing/conn.go:489:
    c.Assert(err, jc.ErrorIsNil)
... value *errors.Err = &errors.unformatter{message:"", cause:(*errors.errorString)(0xc0003bc2f0), previous:(*errors.Err)(0xc0003f1360), file:"d:/payloads/1550826189/build/src/github.com/juju/juju/environs/bootstrap/bootstrap.go", line:536} ()
... error stack:
	Closed explicitly
	d:/payloads/1550826189/build/src/github.com/juju/juju/state/initialize.go:274: 
	d:/payloads/1550826189/build/src/github.com/juju/juju/environs/bootstrap/bootstrap.go:499: 
	d:/payloads/1550826189/build/src/github.com/juju/juju/environs/bootstrap/bootstrap.go:536:

anastasiamac:
!!build!!

anastasiamac:
$$merge$$

PR #9781, opened 2019-02-25
Report whether the model is the controller model

timClicks:
This commit adds "is-controller" into ModelInfo and ModelSummary.
enabling clients to determine which model is the controller model
directly.

User-facing impacts:

    juju list-models --format=yaml
    juju list-models --formay=json
    juju show-model <controller-name> --format=yaml
    juju show-model <controller-name> --format=json

The resulting reports will include an `is-controller` line for each model.


wallyworld:
$$merge$$

PR #9782, opened 2019-02-25
Backport Dmitrii's fix to 2.5.

jameinel:
## Description of change

Fix the network initialization code so that on Bionic we don't exit 0
after running netplan apply during network setup. That way users that
add their own bootcmd properties still get run.

This is just a backport of https://github.com/juju/juju/pull/9751.

## QA steps

You should be able to supply your own post-bootcmd and have it run on Bionic, rather than bootcmd exiting early once networking is set up.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1807615

### DIFF
--- a/cloudconfig/cloudinit/network_ubuntu.go
+++ b/cloudconfig/cloudinit/network_ubuntu.go
@@ -325,17 +325,17 @@ if [ ! -f /sbin/ifup ]; then
     echo "No /sbin/ifup, applying netplan configuration."
     netplan generate
     netplan apply
-    exit 0
-fi
-if [ -f /usr/bin/python ]; then
-    python %[1]s.py --interfaces-file %[1]s --output-file %[1]s.out
 else
-    python3 %[1]s.py --interfaces-file %[1]s --output-file %[1]s.out
+  if [ -f /usr/bin/python ]; then
+      python %[1]s.py --interfaces-file %[1]s --output-file %[1]s.out
+  else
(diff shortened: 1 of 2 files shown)


jameinel:
$$merge$$

dshcherb:
@jameinel We currently cannot change bootcmds via cloudinit-userdata.

https://github.com/juju/juju/blob/e16b7d8/cloudconfig/userdatacfg_unix.go#L395-L401
https://github.com/juju/juju/blob/e16b7d8/cloudconfig/userdatacfg_unix.go#L362-L371

I think it would be more precise to say that Juju's own bootcmds that run afterwards were not executed because of an early exit. It just happened to be that the proxy file rendering bootcmd was the last and only one specified.

howbazaar:
@dshcherb are you saying that the branch as it is now is broken? Or are you commenting about it being broken before?

dshcherb:
@howbazaar 

No, it's not broken, just saying that the fix was not about allowing user customizations to bootcmd to run (which are not possible via cloudinit-userdata model-config AFAIKS, we can only tune {pre,post}runcmd) - the bootcmd command that renders /etc/apt/apt.conf.d/95-juju-proxy-settings comes from the Juju code itself (https://github.com/juju/juju/blob/juju-2.5.0/cloudconfig/cloudinit/cloudinit_ubuntu.go#L302-L312) and the premature exit affected that specific command, not a user-specified command.

Sorry if it was not clear from the PR description - had to type in a hurry.

PR #9785, opened 2019-02-25
Prevent influxdb from rejecting points

achilleasa:
## Description of change

The influxdb client is quite picky when it comes to metric submission; unless an explicit `time_precision`  is specified,  influxdb treats the points as outside the retention policy boundaries and drops them. 

Furthermore, the client does not raise an exception when attempting to connect to a non-existing DB so the reporting code has been tweaked to always create the DB/retention policy (if they already exists then the calls behave as no-ops).

### DIFF
--- a/acceptancetests/assess_deploy_webscale.py
+++ b/acceptancetests/assess_deploy_webscale.py
@@ -13,9 +13,7 @@ import argparse
 import logging
 import sys
 import os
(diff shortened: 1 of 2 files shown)


achilleasa:
$$merge$$

achilleasa:
$$merge$$

PR #9786, opened 2019-02-25
Move the serving-info-setter manifold.

howbazaar:
Moves the serving info setter manifold into the worker directory.

The rationale for this is to be able to soon add a real worker as well that is responsible for updating agent config with the intent that we have one worker that updates the agent.conf information.

## QA steps

* bootstrapping a controller
* enable ha

### DIFF
--- a/cmd/jujud/agent/machine/manifolds.go
+++ b/cmd/jujud/agent/machine/manifolds.go
@@ -35,6 +35,7 @@ import (
 	proxyconfig "github.com/juju/juju/utils/proxy"
 	jworker "github.com/juju/juju/worker"
 	"github.com/juju/juju/worker/agent"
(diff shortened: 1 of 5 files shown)


cmd/jujud/agent/machine/manifolds.go
```diff
@@ -543,10 +544,10 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 			NewWorker:     singular.NewWorker,
 		})),
 
-		// The serving-info-setter manifold sets grabs the state
+		// The agent-config-updater manifold sets grabs the state
```
wallyworld: typo

howbazaar:
$$merge$$

PR #9787, opened 2019-02-25
2.5 cancel lease 1793276

jameinel:
## Description of change

Change the lease.Store interface so that all the methods take a 'stop' channel. Implement support for stopping early for Raft engine, and pass in catacomb/tomb.Dying() as the stop channel.

This should help ensure we don't end up in a situation (at least in tests), where the backend raft engine shuts down while a claim is still pending, which then hangs the front end because it waits 5 seconds before timing out the request.

## QA steps

```
$ go test -race -count=200 -failfast -check.v \
  -check.f TestDowngradeOnMasterWhenOtherControllerDoesntStartUpgrade
```

For me this fails without the patch reasonably quickly but passes with the patch.
The test also passes quite a bit faster. Without the patch it can take 7-10s to run (and the failing one took 13s). But with the patch I often see 2-7s passes.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1793276


jameinel:
$$merge$$

jameinel:
$$merge$$

PR #9788, opened 2019-02-26
raft leases: Group lease entries by (namespace, model) inside FSM

babbageclunk:
## Description of change

This is a refactoring to simplify getting all of the leaders for a specific model in a later change, so that we can replace the use of state.ApplicationLeaders with a call to the lease FSM. The snapshot format isn't changed to avoid needing to rewrite the snapshot in an upgrade step or handle being able to read different snapshot versions.

## QA steps

No behaviour change, unit tests still pass.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/core/raftlease/fsm.go
+++ b/core/raftlease/fsm.go
@@ -59,11 +59,26 @@ type FSMResponse interface {
 	Notify(NotifyTarget)
 }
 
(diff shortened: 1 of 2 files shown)


babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #9793, opened 2019-02-26
Attempt to cleanup models whether volumes exist or not.

jameinel:
## Description of change

Add a test that shows the issue. Use an $or so that we can handle when
the volumes and/or filesystems don't exist as they didn't on models
upgraded from 2.1.

## QA steps

See the new test that was added. When the change isn't applied it fails as in bug #1800872.

```
... error stack:
	state changing too quickly; try again soon
	github.com/juju/juju/state/model.go:1016: failed to destroy model
```

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1800872


jameinel:
$$merge$$

jameinel:
$$merge$$ failure was a TestWatch in FirewallerTest

PR #9795, opened 2019-02-26
Clean up lxdprofile core

SimonRichardson:
## Description of change

The following commit ensures that core/lxdprofile doesn't bring any
adverse import changes into the core package. We can test this
by using TestImports in the package_test.go.

The fallout from this is that we then need to ensure we convert
the various charm, charm archive and charm info into the right
type with various shims-structs.

 - Check against nil pointers


SimonRichardson:
!!build!!

> WARNING juju.testing mgo.go:212 failed to start mongo: mongod failed to listen on port 37730

SimonRichardson:
$$merge$$

PR #9799, opened 2019-02-27
Default model config is now per cloud

wallyworld:
## Description of change

Default model config used to be stored globally for all models in the controller, since all models shared the same cloud. Now that controllers support models in different clouds, the default config needs to be per cloud.

There's 4 commits:
- state changes
- facade changes
- CLI changes
- upgrade steps

## QA steps

bootstrap 2.5.1 with default cloud config
upgrade to this branch
check that the data in globalSettings collection has been upgraded
run juju model-defaults in various scenarios, checking that combinations of no args, args with cloud, args with cloud/region etc work


state/modelconfig.go
```diff
@@ -137,23 +137,32 @@ func (model *Model) modelConfigValues(modelCfg attrValues) (config.ConfigValues,
 }
 
 // UpdateModelConfigDefaultValues updates the inherited settings used when creating a new model.
-func (model *Model) UpdateModelConfigDefaultValues(attrs map[string]interface{}, removed []string, regionSpec *environs.RegionSpec) error {
+func (st *State) UpdateModelConfigDefaultValues(attrs map[string]interface{}, removed []string, regionSpec *environs.CloudRegionSpec) error {
 	var key string
 
 	if regionSpec != nil {
-		key = regionSettingsGlobalKey(regionSpec.Cloud, regionSpec.Region)
+		if regionSpec.Region != "" {
```
babbageclunk: Super minor nit - when you've got both an if and else clause I think it's more readable to order them so the condition is positive, so this would be 
```
if regionSpec.Region == "" {
    key = cloudGlobalKey(regionSpec.Cloud)
} else {
    key = regionSettingsGlobalKey(regionSpec.Cloud, regionSpec.Region)
}
```

(And the same for the outer if.)

state/modelconfig.go
```diff
@@ -397,21 +400,23 @@ func (st *State) defaultInheritedConfig() (attrValues, error) {
 
 // controllerInheritedConfig returns the inherited config values
 // sourced from the local cloud config.
-func (st *State) controllerInheritedConfig() (attrValues, error) {
-	settings, err := readSettings(st.db(), globalSettingsC, controllerInheritedSettingsGlobalKey)
-	if err != nil {
-		return nil, errors.Annotatef(err, "controller %q", st.ControllerUUID())
+func (st *State) controllerInheritedConfig(cloudName string) func() (attrValues, error) {
```
babbageclunk: Maybe mention that it returns a func now in the docstring?

apiserver/facades/client/cloud/cloud.go
```diff
@@ -28,6 +28,12 @@ import (
 
 var logger = loggo.GetLogger("juju.apiserver.cloud")
 
+// CloudV5 defines the methods on the cloud API facade, version 5.
```
babbageclunk: I think I get why we do it, but it's confusing that the interface for version N embeds version N-1 while the struct for version N embeds version N+1. Maybe it would be clearer for the interfaces to just be explicit and list all of the methods?

apiserver/facades/client/cloud/cloud.go
```diff
@@ -429,6 +451,33 @@ func (api *CloudAPI) DefaultCloud() (params.StringResult, error) {
...
+			result.Results[i].Error = common.ServerError(err)
+			continue
+		}
+		pm, err := api.pool.Get(modelTag.Id())
+		if err != nil {
+			result.Results[i].Error = common.ServerError(err)
+			continue
+		}
+		defer pm.Release()
+		m, err := pm.Model().Model()
```
babbageclunk: ugh

apiserver/facades/client/cloud/cloudV2_test.go
```diff
@@ -124,7 +124,7 @@ func (s *cloudSuiteV2) setTestAPIForUser(c *gc.C, user names.UserTag) {
 	}
 	client, err := cloudfacade.NewCloudAPI(s.backend, s.backend, s.statePool, s.authorizer, context.NewCloudCallContext())
 	c.Assert(err, jc.ErrorIsNil)
-	s.apiv2 = &cloudfacade.CloudAPIV2{&cloudfacade.CloudAPIV3{client}}
+	s.apiv2 = &cloudfacade.CloudAPIV2{&cloudfacade.CloudAPIV3{&cloudfacade.CloudAPIV4{client}}}
```
babbageclunk: heh

cmd/juju/model/defaultscommand.go
```diff
@@ -37,20 +36,28 @@ supplied key to the supplied value. This can be repeated for multiple keys.
 You can also specify a yaml file containing key values.
 By default, the model is the current model.
 
+Model default configuration settings are relevant to the cloud on which the model runs.
```
babbageclunk: I think "specific" rather than "relevant"?

cmd/juju/model/defaultscommand.go
```diff
@@ -338,23 +384,29 @@ func (c *defaultsCommand) parseArgsForRegion(args []string) ([]string, error) {
...
-		region = cr
+		// The arg may be either a cloud name or region or yaml file.
+		// We will see below if the value corresponds to a
+		// cloud on the controller, in which case it is interpretted
+		// as a cloud, else it is considered a region.
+		maybeRegion = cr
 	}
 
+	// Ensure we exit early when we have a settings file or key=value.
+	_, err := os.Stat(maybeRegion)
```
babbageclunk: Shouldn't you check for the file before treating it as cloud/region? Otherwise you'll do the wrong thing if someone passes dir/config.yaml - it'll get treated as `cloud=dir`, `maybeRegion=config.yaml` and it won't work - would be a confusing user experience.

state/upgrades.go
```diff
@@ -1843,3 +1843,44 @@ func SetEnableDiskUUIDOnVsphere(pool *StatePool) error {
 		return true, nil
 	}))
 }
+
+// UpdateInheritedControllerConfig migrates the existing global
+// settings doc ley on "controller" to be keyed on the cloud name.
```
babbageclunk: typo: ley

wallyworld:
[default model config based on cloud](https://trello.com/c/uBx68IZ2/636-default-model-config-based-on-cloud)

wallyworld:
$$merge$$

PR #9805, opened 2019-02-28
Remove Juju specific sysctl tweaks for LXD

mitechie:
Fixes #1817774

It was noted that we set a couple of sysctl values for LXD and since then LXD
has started providing their own default values. This pulls that from our
config.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -7,12 +7,3 @@ fi
 
 # copy bash completions to host system
 cp -a $SNAP/bash_completions/* /usr/share/bash-completion/completions/. || true
-
-# setup sysctl defaults for lxd
-mkdir -p /usr/lib/sysctl.d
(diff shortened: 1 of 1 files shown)


mitechie:
My preference is to have it available in the devel edge snap and make sure we don't see any fallout of expected behavior for a bit before making it available in 2.5. 

$$merge$$

mitechie:
$$merge$$

jameinel:
I certainly agree with testing it in develop/edge. Just noting that we want to keep thinking about this one, as it doesn't actually help the people in the bug until it is in a stable release that they can use.

jameinel:
$$merge$$

mitechie:
$$merge$$

jameinel:
4-hrs later it still doesn't seem to actually be running?
$$merge$$

PR #9809, opened 2019-02-28
Fix blocking shutdown race.

howbazaar:
There are some test timeouts for watcher tests where the test gets killed after 25 minutes.

This was due to a bare channel send and nothing listening.

This blocks the idleFunc in the hubwatcher, which stops it processing the main loop, which includes watching the dying channel, which matches where the test is hung, which is in the state.Close() method.


babbageclunk:
$$merge$$

howbazaar:
$$merge$$

PR #9811, opened 2019-02-28
Fix the race between the dial and the wait.

howbazaar:
apiclientSuite.TestOpenDialTimeoutDoesNotAffectLogin has a race where the dial isn't finished before the number of watchers was checked.

Fixed this by changing the unblock channel to be non-buffered so as to ensure that the api.Open goroutine has entered the Login method before continuing. We also change the test to actually advance the clock past the DialTimeout.


howbazaar:
$$merge$$

howbazaar:
$$merge$$

howbazaar:
$$merge$$

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #9813, opened 2019-02-28
LXD: fix broken juju-build

SimonRichardson:
## Description of change

The following adds the lxd via snap func for other architectures.


SimonRichardson:
$$merge$$

PR #9815, opened 2019-03-01
Add k8s deployment watcher to report scale changes

wallyworld:
## Description of change

Add a k8s watcher for stateful sets and deployments to report on scale changes.
This allows auto scaling in k8s to be reflected in the Juju model.

Part of the work was to take an existing multi notification watcher from the apiserver package and copy that to core so it can be used in more places.

## QA steps

Deploy a k8s model. Scale from the k8s side and observe the changes in the Juju model.
Scale from the Juju side as well to ensure things still work in that direction.

### DIFF
--- a/apiserver/facades/controller/caasunitprovisioner/state.go
+++ b/apiserver/facades/controller/caasunitprovisioner/state.go
@@ -67,6 +67,7 @@ type Model interface {
 // required by the CAAS unit provisioner facade.
 type Application interface {
 	GetScale() int
+	Scale(int) error
 	WatchScale() state.NotifyWatcher
 	ApplicationConfig() (application.ConfigAttributes, error)
 	AllUnits() (units []Unit, err error)
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -808,6 +808,32 @@ func (k *kubernetesClient) Service(appName string) (*caas.Service, error) {
 			Scope: network.ScopePublic,
 		})
 	}
+
+	deploymentName := k.deploymentName(appName)
+	statefulsets := k.AppsV1().StatefulSets(k.namespace)
+	ss, err := statefulsets.Get(deploymentName, v1.GetOptions{})
+	if err == nil {
+		scale := 0
+		if ss.Spec.Replicas != nil {
+			scale = int(*ss.Spec.Replicas)
+		}
+		result.Scale = scale
+		return &result, nil
+	}
+	if !k8serrors.IsNotFound(err) {
+		return nil, errors.Trace(err)
+	}
+
+	deployments := k.AppsV1().Deployments(k.namespace)
+	deployment, err := deployments.Get(deploymentName, v1.GetOptions{})
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	scale := 0
+	if deployment.Spec.Replicas != nil {
+		scale = int(*deployment.Spec.Replicas)
+	}
+	result.Scale = scale
 	return &result, nil
 }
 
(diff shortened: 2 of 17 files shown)


caas/kubernetes/provider/k8s.go
```diff
@@ -1640,6 +1666,41 @@ func (k *kubernetesClient) WatchUnits(appName string) (watcher.NotifyWatcher, er
...
+		Watch:         true,
+	})
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	w2, err := k.newWatcher(dwatcher, appName, k.clock)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
```
ycliuhw: I think, here could be either w1 or w2 NotFound.
And we should probably not return not found errors?

caas/kubernetes/provider/k8s.go
```diff
@@ -807,6 +807,32 @@ func (k *kubernetesClient) Service(appName string) (*caas.Service, error) {
...
+		result.Scale = scale
+		return &result, nil
+	}
+	if !k8serrors.IsNotFound(err) {
+		return nil, errors.Trace(err)
+	}
+
+	deployments := k.AppsV1().Deployments(k.namespace)
+	deployment, err := deployments.Get(deploymentName, v1.GetOptions{})
+	if err != nil {
```
ycliuhw: should we translate `k8serrors.NotFound` to `errors.NotFound` here?

caas/kubernetes/provider/k8s.go
```diff
@@ -1640,6 +1666,41 @@ func (k *kubernetesClient) WatchUnits(appName string) (watcher.NotifyWatcher, er
...
 
+// WatchService returns a watcher which notifies when there
+// are changes to the deployment of the specified application.
+func (k *kubernetesClient) WatchService(appName string) (watcher.NotifyWatcher, error) {
+	// Application may be a statefulset or deployment. It may not have
+	// been set up when the watcher is started so we don't know which it
+	// is ahead of time. So use a multi-watcher to cover both cases.
+	statefulsets := k.AppsV1().StatefulSets(k.namespace)
+	sswatcher, err := statefulsets.Watch(v1.ListOptions{
+		LabelSelector: applicationSelector(appName),
```
ycliuhw: probably we should include `IncludeUninitialized=true`?

caas/kubernetes/provider/k8s.go
```diff
@@ -1640,6 +1666,41 @@ func (k *kubernetesClient) WatchUnits(appName string) (watcher.NotifyWatcher, er
...
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	w1, err := k.newWatcher(sswatcher, appName, k.clock)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	deployments := k.AppsV1().Deployments(k.namespace)
+	dwatcher, err := deployments.Watch(v1.ListOptions{
```
ycliuhw: here as well

worker/caasunitprovisioner/application_worker.go
```diff
@@ -125,6 +130,16 @@ func (aw *applicationWorker) loop() error {
 				return errors.Annotatef(err, "failed to start operator watcher for %q", aw.application)
 			}
 		}
+		if appDeploymentWatcher == nil {
+			appDeploymentWatcher, err = aw.serviceBroker.WatchService(aw.application)
```
ycliuhw: `appServiceWatcher`?

wallyworld:
$$merge$$

apiserver/facades/controller/caasunitprovisioner/state.go
```diff
@@ -67,6 +67,7 @@ type Model interface {
 // required by the CAAS unit provisioner facade.
 type Application interface {
 	GetScale() int
+	Scale(int) error
```
ycliuhw: `SetScale` is a better name?

worker/caasunitprovisioner/application_worker.go
wallyworld: The Watch() api is on the serviceBroker interface along with a few other methods. There's not many so no real benefit to split up.

caas/kubernetes/provider/k8s.go
wallyworld: Not sure, we don't elsewhere for watching pods. If we do decide to change, should do so everyhwere else too.

caas/kubernetes/provider/k8s.go
wallyworld: theoretically should not get that error but we could so have added extra check

apiserver/facades/controller/caasunitprovisioner/state.go
wallyworld: Imported from existing state method so I changed the method nae across the stack

wallyworld:
$$merge$$

wallyworld:
[default model config based on cloud](https://trello.com/c/uBx68IZ2/636-default-model-config-based-on-cloud)

wallyworld:
[default model config based on cloud](https://trello.com/c/uBx68IZ2/636-default-model-config-based-on-cloud)

PR #9818, opened 2019-03-01
Golinters > Gometalinters 

SimonRichardson:
## Description of change

Moved to golangci-linters instead of the now deprecated gometalinters.

## QA steps

```
./scripts/golinters.sh
```

## Documentation changes

We now use golangci-linters instead of gometalinters for code checking


scripts/verify.bash
```diff
@@ -58,24 +64,18 @@ UserNotFoundf
...
    -composites=false \
    -printfuncs=$all_prints \
-    $FILES || [ -n "$IGNORE_VET_WARNINGS" ]
-
-echo "checking: copyright notices are in place ..."
-./scripts/copyright.bash
-
-echo "checking: dependency files ..."
-dep check
+    ./... || [ -n "$IGNORE_VET_WARNINGS" ]
```
jameinel: If we are doing this, is it just that it only supports a single argument? or is it that it only supports packages?
because we could do something like
DIRNAMES=$(dirname $FILES | sort -u)
Then we have the unique directories with files that we care about?

SimonRichardson:
https://trello.com/c/IiBbEovh/1125-golinters-and-go-vet-changes-sprint-left-overs

SimonRichardson:
$$merge$$

PR #9822, opened 2019-03-01
add-k8s is local by default with -c option

wallyworld:
## Description of change

juju add-k8s and remove-k8s is now only local by default.
It has an optional --controller argument to add/remove to a controller.

## QA steps

Run add-k8s standalone.
Run add-k8s with --controller
Run remove-k8s standalone.
Run remove-k8s with --controller


wallyworld:
$$merge$$

PR #9825, opened 2019-03-02
Add synchronisation to the test.

howbazaar:
The test was just using the clock waiters as a synchronisation point. The clock waiter happens outside the parallel try code. If the waiter was signalled before the parallel try goroutine actually started processing, we would get the 'try was stopped' error. This is because the context was cancelled before the try really started. The fix is to ensure that the dial function is entered before we advance time.

## QA steps

Run the tests with the stress.sh script.
Prior to the fix on my machine it failed more often than it succeeded.
After the fix it runs fine.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1815800

### DIFF
--- a/api/apiclient_test.go
+++ b/api/apiclient_test.go
@@ -790,7 +790,9 @@ func (s *apiclientSuite) TestOpenTimesOutOnLogin(c *gc.C) {
 }
 
 func (s *apiclientSuite) TestOpenTimeoutAffectsDial(c *gc.C) {
+	sync := make(chan struct{})
 	fakeDialer := func(ctx context.Context, urlStr string, tlsConfig *tls.Config, ipAddr string) (jsoncodec.JSONConn, error) {
+		close(sync)
 		<-ctx.Done()
 		return nil, ctx.Err()
 	}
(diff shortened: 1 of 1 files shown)


api/apiclient_test.go
```diff
@@ -790,7 +790,9 @@ func (s *apiclientSuite) TestOpenTimesOutOnLogin(c *gc.C) {
 }
 
 func (s *apiclientSuite) TestOpenTimeoutAffectsDial(c *gc.C) {
+	sync := make(chan struct{})
 	fakeDialer := func(ctx context.Context, urlStr string, tlsConfig *tls.Config, ipAddr string) (jsoncodec.JSONConn, error) {
+		close(sync)
 		<-ctx.Done()
```
jameinel: do we want a bare channel read here? or do we want to wrap it in a time.After(testing.LongWait)

howbazaar:
$$merge$$

jameinel:
I also tested with a patch that did:
```
--- a/api/apiclient.go
+++ b/api/apiclient.go
@@ -193,6 +193,7 @@ func Open(info *Info, opts DialOpts) (Connection, error) {
                defer cancel()
                dialCtx = ctx1
        }
+       time.Sleep(1*time.Millisecond)
        dialResult, err := dialAPI(dialCtx, info, opts)
        if err != nil {
                return nil, errors.Trace(err)
```

And can confirm that it failed very quickly for me, but with Tim's proposal it succeeded many times.

jameinel:
(never failed after 200 attempts, where without the PR it failed in 3 attempts)

api/apiclient_test.go
howbazaar: We know that the context is killed within the test, so it progresses.

PR #9827, opened 2019-03-02
Add synchronisation to the test.

jameinel:
## Description of change

Test would intermittently fail if the Dial was not reached by the time we go to check the assertion. This adds synchronization so we know we are in Dial if we get passed the sync point.

This is just Tim's patch (#9825) pointed at 2.5.

## QA steps

See pr #9825.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1815800


jameinel:
$$merge$$ approving Tim's PR against 2.5

PR #9830, opened 2019-03-06
pubsub: Fix lockups when a controller machine goes away

babbageclunk:
## Description of change
Investigating [this bug](https://bugs.launchpad.net/juju/+bug/1816533) revealed that the pubsub forwarder would get blocked if one of the controller machines was stopped (on MAAS - this didn't happen on LXD). That in turn meant that lease operations wouldn't be forwarded to the raft leader to be applied. There were two problems that caused this: 
* Writes to the forwarder's websocket for the dead machine would block indefinitely, although the remoteServer (in the local agent) still believed that it was connected to it.
* Once that was fixed a race in remoteServer.Publish could lead to it getting stuck trying waiting for the new-data notification to be received when the connection had been reset.

This changes the streams returned from `connection.ConnectStream` and `.ConnectControllerStream` to always set a 30s write deadline when calling `WriteJSON`, and fixes the race.

## QA steps
* Bootstrap an HA controller on MAAS.
* Deploy an application with several units (can all be on the same machine).
* Stop one of the controller machines.
* Remove the leader unit and see that a different unit becomes the leader after the previous leader's lease expires.
* Run juju_engine_report on the 2 running controller machines - it should finish, indicating that the pubsub forwarder isn't blocked.

## Documentation changes
None

## Bug reference
https://bugs.launchpad.net/juju/+bug/1816533

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/gorilla/websocket"
 	"github.com/juju/errors"
@@ -34,6 +35,10 @@ import (
 	"github.com/juju/juju/network"
 	"github.com/juju/juju/tools"
 )
+
+// websocketTimeout is how long we'll wait for a WriteJSON call before
+// timing it out.
+const websocketTimeout = 30 * time.Second
 
 // Client represents the client-accessible part of the state.
 type Client struct {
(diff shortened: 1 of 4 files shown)


babbageclunk:
I couldn't see any way to test the race fix.

I wanted to work out a way to test the SetWriteDeadline behaviour, but after trying for ages I couldn't get a websocket server and client in the test to fail with an i/o timeout. So I've fallen back to using a fake dialler and just checking that it has a deadline stream wrapped around it.

worker/pubsub/remoteserver.go
```diff
@@ -188,7 +189,11 @@ func (r *remoteServer) Publish(message *params.PubSubMessage) {
 		}
 		r.mutex.Unlock()
 		if notifyData {
-			r.data <- struct{}{}
+			select {
+			case r.data <- struct{}{}:
+			case <-r.connectionReset:
```
wallyworld: maybe log debug something?

api/client.go
```diff
@@ -654,7 +659,28 @@ func websocketDialWithErrors(dialer WebsocketDialer, urlStr string, requestHeade
...
+	Timeout time.Duration
+}
+
+// WriteJSON is part of base.Stream.
+func (s *DeadlineStream) WriteJSON(v interface{}) error {
+	// This uses a real clock rather than trying to use a clock passed
+	// in because the websocket will use a real clock to determine
+	// whether the deadline has passed anyway.
+	deadline := time.Now().Add(s.Timeout)
+	if err := s.Conn.SetWriteDeadline(deadline); err != nil {
```
wallyworld: this isn't threadsafe - can it be used by more than one thread?

worker/pubsub/remoteserver.go
babbageclunk: Good call.

api/client.go
babbageclunk: It can't be, no. I've gone through the uses of ConnectStream/ConnectControllerStream to see whether it is, these are the uses I can see:

* pubsub worker - only writes from one goroutine
* log forwarder worker - sends to syslog, only reads from the stream
* migrationmaster worker - only writes from one goroutine
* debug-log command - doesn't use WriteJSON, only reads
* logsender - only writes from one goroutine

Which obviously doesn't guard against someone writing to one from multiple goroutines in the future. I'll add a comment to the docstrings of Connect*Stream that writing should only be done from one goroutine because of the deadline.

babbageclunk:
> shouldn't this be targetting 2.5 ?

Oh yes it should! Good spotting!

babbageclunk:
I've tweaked the windows unit test job to fix the scp error it was reporting.
!!build!!

babbageclunk:
$$merge$$

PR #9833, opened 2019-03-07
Fix race in apiclient_test TestOpenDialTimeoutAffectsDial

howbazaar:
Same fix as for the TestOpenTimeoutAffectsDial in #9825.
Also checked the rest of the file for other time advance calls, and I'm pretty sure we are covered now.


howbazaar:
$$merge$$

PR #9835, opened 2019-03-07
Fix intermittent test failure in CAAS unit provisioner API

babbageclunk:
## Description of change
Removed unit ids would be reported out of order sometimes.

## QA steps

Test used to reliably fail within 10 runs under stress, now doesn't fail.

## Documentation changes

None

## Bug reference

None


apiserver/facades/controller/caasunitprovisioner/provisioner.go
```diff
@@ -660,10 +660,13 @@ func (a *Facade) updateUnitsFromCloud(app Application, unitUpdates []params.Appl
 
 	// If there are any extra provider ids left over after allocating all the cloud pods,
 	// then consider those state units as terminated.
+	terminatedIds := extraStateIds.Values()
```
howbazaar: set.Strings has a method called `SortedValues()`

wallyworld:
Already fixed in https://github.com/juju/juju/pull/9815

PR #9840, opened 2019-03-07
 Support running webscale tests against user-provided mongo snaps

achilleasa:
## Description of change

This PR adds the `--db-snap-path` argument to the webscale test script that allows the CI to override the mongo version that gets deployed to the test controller. This requires enabling the `mongodb-snap` feature flag which activates the `--db-snap` commandline option for `juju bootstrap`. 

The `--db-snap-path` argument to the webscale tests accepts either a local file or a URL. In the latter case, the snap will be downloaded and appropriately renamed to match the format expected by the `--db-snap` argument to `juju bootstrap`.

NOTE: this feature requires resolving of [this](https://bugs.launchpad.net/juju/+bug/1819004) issue before we can turn it on as part of the CI flow.

### DIFF
--- a/acceptancetests/assess_deploy_webscale.py
+++ b/acceptancetests/assess_deploy_webscale.py
@@ -16,6 +16,10 @@ import os
 import re
 import statistics
 import time
+import shutil
+
+from urllib.parse import urlparse
+from urllib.request import urlretrieve
 
 from deploy_stack import (
     BootstrapManager,
@@ -216,6 +220,13 @@ def parse_args(argv):
         help="Help the reporting metrics by supplying a target juju version",
         default="",
     )
+    parser.add_argument(
+        '--db-snap-path',
+        help="URL to a mongo snap to override the mongo version used for the "
+             "test controller; if a URL is specified, it will be download "
+             "locally before bootstrapping",
+        default="",
+    )
     add_basic_testing_arguments(parser, existing=False)
     # Override the default logging_config default value set by adding basic
     # testing arguments. This way we can have a default value for all tests,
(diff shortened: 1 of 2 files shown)


acceptancetests/assess_deploy_webscale.py
```diff
@@ -225,12 +235,43 @@ def parse_args(argv):
...
 
 
+def check_for_mongo_snap(args):
+    setattr(args, "db_snap_asserts_path", "/tmp/empty.asserts")
+    if args.db_snap_path:
+        # NOTE(achilleasa): ideally we should be using client.enable_feature()
+        # but it looks like the current implementation does not pass the
+        # enabled features to the backend; the backend however supports
+        # fetching the features through the JUJU_DEV_FEATURE_FLAGS envvar.
+        log.info("Enabling 'mongodb-snap' feature flag")
```
mitechie: let's log the value as well in the output here please.

mitechie: sorry in this I mean what the db_snap_path is set to.

acceptancetests/assess_deploy_webscale.py
```diff
@@ -225,12 +235,43 @@ def parse_args(argv):
...
     configure_logging(args.verbose)
+    args = check_for_mongo_snap(args)
+
     begin = time.time()
     bs_manager = BootstrapManager.from_args(args)
-    with bs_manager.booted_context(args.upload_tools):
+    with bs_manager.booted_context(
+        args.upload_tools,
+        db_snap_path=args.db_snap_path,
+        db_snap_asserts_path=args.db_snap_asserts_path
```
mitechie: So in general I'm usually a -1 on modifying args. I'd prefer if the function read args and then returned the result noting the db_snap name and the asserts path required. The issue is that it raises potential conflict for args in the future and makes it unclear if something is on the args object came from the user or the various code that modifies it. I tend to think of args like a user supplied chunk of data that is used but not modified so there's no doubt about the sources of things.

acceptancetests/jujupy/client.py
```diff
@@ -930,13 +934,16 @@ def update_user_name(self):
...
+                  no_gui=False, agent_version=None, db_snap_path=None,
+                  db_snap_asserts_path=None):
         """Bootstrap a controller."""
         self._check_bootstrap()
         with self._bootstrap_config() as config_filename:
             args = self.get_bootstrap_args(
                 upload_tools, config_filename, bootstrap_series, credential,
-                auto_upgrade, metadata_source, no_gui, agent_version)
+                auto_upgrade, metadata_source, no_gui, agent_version,
+                db_snap_path=db_snap_path,
```
mitechie: if we're going to modify to start using kwargs then let's go that route and update the rest to be kwargs as well please.

acceptancetests/assess_deploy_webscale.py
achilleasa: Now that I think about it there is really no need to modify `args`. In some of my earlier attempts I was passing the args to the bootstrap manager c-tor and I forgot to clean up my code. I will push an update for this.

acceptancetests/jujupy/client.py
achilleasa: Make sense

achilleasa: There was no need to have the last 2 as kwargs so I just made them regular args.

achilleasa:
$$merge$$

PR #9843, opened 2019-03-08
Fix issue reporting k8s scale

wallyworld:
## Description of change

There was a race condition reporting a k8s deployment scale which resulting in the juju model sometimes being updated with scale=0 by mistake.
As a driveby, also fix a typo in a volume lookup when reporting PV info. This prevented storage from being reported as attached.

## QA steps

Deploy a k8s charm with storage.
Check that storage shows as attached.

## Bug reference

https://bugs.launchpad.net/bugs/1818588


wallyworld:
[storage is not recorded as attached](https://trello.com/c/ncNlD0la/640-storage-is-not-recorded-as-attached)

wallyworld:
$$merge$$

PR #9847, opened 2019-03-11
Update times for test runs.

howbazaar:
Both arm64 and ppc64le state tests have been timing out on the state package for a while. Also the mongo oplog test sometimes just fails on s390x and ppc64le - no obvious reason why.


howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #9848, opened 2019-03-11
Sync before the watch.

howbazaar:
The test created the watcher then synced before checking values. The model needs to be idle before creating the watcher.


howbazaar:
$$merge$$

PR #9849, opened 2019-03-11
Add cloud api version to hook context

wallyworld:
## Description of change

The hook context now includes an env var to report on the underlying cloud api version.
CLOUD_API_VERSION

k8s charms need this because different k8s versions sometimes require slightly different yaml to be sent to configure workloads.

Note: the uniter client api was broken in that it hard coded the version as 9. As a drive by I fixed it to behave like all other facade clients.

## QA steps

deploy  a k8s charm which logs the CLOUD_API_VERSION from the hook context

### DIFF
--- a/api/uniter/uniter.go
+++ b/api/uniter/uniter.go
@@ -37,17 +37,14 @@ type State struct {
 	unitTag names.UnitTag
 }
 
-// newStateForVersion creates a new client-side Uniter facade for the
-// given version.
-func newStateForVersion(
+// NewState creates a new client-side Uniter facade.
+func NewState(
 	caller base.APICaller,
 	authTag names.UnitTag,
-	version int,
 ) *State {
-	facadeCaller := base.NewFacadeCallerForVersion(
+	facadeCaller := base.NewFacadeCaller(
 		caller,
 		uniterFacade,
-		version,
 	)
 	state := &State{
 		ModelWatcher:     common.NewModelWatcher(facadeCaller),
@@ -69,19 +66,6 @@ func newStateForVersion(
 	)
 	return state
 }
-
-func newStateForVersionFn(version int) func(base.APICaller, names.UnitTag) *State {
-	return func(caller base.APICaller, authTag names.UnitTag) *State {
-		return newStateForVersion(caller, authTag, version)
-	}
-}
-
-// newStateV9 creates a new client-side Uniter facade, version 9
-var newStateV9 = newStateForVersionFn(9)
-
-// NewState creates a new client-side Uniter facade.
-// Defined like this to allow patching during tests.
-var NewState = newStateV9
 
 // BestAPIVersion returns the API version that we were able to
 // determine is supported by both the client and the API Server.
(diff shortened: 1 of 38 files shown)


babbageclunk:
Why use a hook context var rather than model config or charm config?

wallyworld:
> Why use a hook context var rather than model config or charm config?

Charm config is a knob to tune how an application works and is meant to be understood and settable by the end user.
Model config is not exposed to the charm and is meant to control how Juju interacts with the cloud.
What we need here is a way to expose to the charm information about the capabilities of the cloud on which it is to be deployed. Different k8s versions have different requirements that can drive how a charm should generate yaml config and the like. The charm engineering team have encountered one such scenario where the yaml they need to generate in the charm differs for kubelet v1.14 (I think) vs earlier

apiserver/common/networkingcommon/shims.go
```diff
@@ -88,7 +88,7 @@ func NewStateShim(st *state.State) (*stateShim, error) {
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	return &stateShim{stateenvirons.EnvironConfigGetter{st, m}, st, m}, nil
+	return &stateShim{stateenvirons.EnvironConfigGetter{State: st, Model: m}, st, m}, nil
```
babbageclunk: Why include the field names in the EnvironConfigGetter but not the stateShim?

Oh, I see - because you've added a new field to EnvironConfigGetter? Ok cool.

apiserver/facades/agent/uniter/export_test.go
```diff
@@ -31,3 +32,7 @@ func NewStorageAPI(
 ) (*StorageAPI, error) {
 	return newStorageAPI(backend, storage, resources, accessUnit)
 }
+
+func SetNewContainerBrokerFucn(api *UniterAPI, newBroker caas.NewContainerBrokerFunc) {
```
babbageclunk: Should be `SetNewContainerBrokerFuck`.
I mean, `SetNewContainerBrokerFunc`.

caas/kubernetes/provider/k8s_test.go
```diff
@@ -339,6 +341,21 @@ type K8sBrokerSuite struct {
 
 var _ = gc.Suite(&K8sBrokerSuite{})
 
+func (s *K8sBrokerSuite) TestAPIVersion(c *gc.C) {
+	ctrl := s.setupBroker(c)
+	defer ctrl.Finish()
+
+	r := rest.NewRequest(nil, "get", &url.URL{Path: "/path/"}, "", rest.ContentConfig{}, rest.Serializers{}, nil, nil, 0)
+	gomock.InOrder(
```
babbageclunk: Do you need the InOrder here? There's only one expectation.

api/uniter/uniter.go
```diff
@@ -70,19 +67,6 @@ func newStateForVersion(
 	return state
 }
 
-func newStateForVersionFn(version int) func(base.APICaller, names.UnitTag) *State {
```
babbageclunk: This wasn't used from the tests somehow? Weird.

apiserver/facades/agent/uniter/export_test.go
wallyworld: lol, fixed

caas/kubernetes/provider/k8s_test.go
wallyworld: guess not

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
[hook env var for k8s version](https://trello.com/c/iqQraX3A/631-hook-env-var-for-k8s-version)

PR #9850, opened 2019-03-11
Wait for the model to be idle before starting the test.

howbazaar:
The meter status feature test needs to sync with the watchers before starting the tests.

Fixes intermitten test failures in featturetests/api_meterstatus_test.go TestWatchMeterStatus.


howbazaar:
$$merge$$

PR #9851, opened 2019-03-11
Add logging for a test that doesn't fail locally.

howbazaar:
Seems to fail intermittently on the merge bot, but can't make it fail locally, so lets get more logging when it does fail.


howbazaar:
$$merge$$

PR #9852, opened 2019-03-11
Display 'kubernetes' clouds as 'k8s'

wallyworld:
## Description of change

When listing or showing clouds, instead of "kubernetes", print "k8s".
Also a couple of drive by fixes: new make target to update microk8s operator image and a tweak to k8s api version reported in hook context.

## QA steps

juju clouds
juju clouds --format yaml
juju show-cloud foo

## Bug reference

https://bugs.launchpad.net/juju/+bug/1814670

### DIFF
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -303,7 +303,10 @@ func (k *kubernetesClient) APIVersion() (string, error) {
 	if err != nil {
 		return "", errors.Annotatef(err, "got '%s' querying API version", string(body))
 	}
-	return fmt.Sprintf("%s.%s", info.Major, info.Minor), nil
+	version := info.GitVersion
+	// git version is "vX.Y.Z", strip the "v"
+	version = strings.Trim(version, "v")
+	return version, nil
 }
 
 // Namespaces returns names of the namespaces on the cluster.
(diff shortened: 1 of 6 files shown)


caas/kubernetes/provider/k8s.go
```diff
@@ -303,7 +303,10 @@ func (k *kubernetesClient) APIVersion() (string, error) {
 	if err != nil {
 		return "", errors.Annotatef(err, "got '%s' querying API version", string(body))
 	}
-	return fmt.Sprintf("%s.%s", info.Major, info.Minor), nil
+	version := info.GitVersion
+	// git version is "vX.Y.Z", strip the "v"
+	version = strings.Trim(version, "v")
+	return version, nil
```
Veebers: So this has changed from returning "x.y" to now returning "x.y.z"? There was no test update needed?

wallyworld: There's a test that calling APIVersion() invokes the correct ReST URL but to weave in the return result would have required an extra 100s of lines of mocks and/or stubs seeing as the underlying code deals with a rest.Request struct containing an actual embedded http client.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #9855, opened 2019-03-12
Use real time for hubwatcher idle, and changes reset idle.

howbazaar:
Noticed a number of test failures in the state package. While reproducing noticed unexpected behaviour with the hubwatcher idle handling.

The changes channel from the txnwatcher wasn't resetting the idle time, as that was only being reset when an event was forwarded. Also we should be using wall clock time rather than whatever the hub watcher was created with as the test clock is often advanced all at once and this isn't conducive to iterative `.After` calls. The idle callback is only used in tests, so this isn't used at all in production code.

Both the unit and machine watchers in the state package were missing sync calls.


howbazaar:
$$merge$$

PR #9856, opened 2019-03-12
Restart machine agent if mongo profile changes.

howbazaar:
When the machine agent starts, it starts an agent config update worker. This worker is used to update the local agent.conf file. Ideally all updates to the file will go through this worker eventually, but for now, we are just changing the worker to get the agent to restart if the mongo memory profile is changed.

This is to record the mongo memory profile in the agent.conf file of any machines that become API servers. The value was written into the bootstrapped machine initially, but any new machines added, like with `enable-ha` didn't set the mongo memory profile.

When the agent restarts, the config file is read, now with the new mongo memory profile, and the mongo service script is updated and mongo restarted.

# QA steps
```
juju bootstrap lxd test  --config mongo-memory-profile=default
juju enable-ha
# wait for machines to start
juju ssh -m controller 0
sudo tail -2 /var/lib/juju/agents/machine-0/agent.conf
# there will be `mongomemoryprofile: default`
sudo grep mongod /etc/systemd/system/juju-db.service
# command line will not have --wiredTigerCacheSizeGB arg

# check the same for machine 1 and 2
juju ssh -m controller 1
sudo tail -2 /var/lib/juju/agents/machine-1/agent.conf
sudo grep mongod /etc/systemd/system/juju-db.service

juju ssh -m controller 2
sudo tail -2 /var/lib/juju/agents/machine-2/agent.conf
sudo grep mongod /etc/systemd/system/juju-db.service

```

### DIFF
--- a/worker/agentconfigupdater/manifold.go
+++ b/worker/agentconfigupdater/manifold.go
@@ -12,13 +12,22 @@ import (
 	coreagent "github.com/juju/juju/agent"
 	apiagent "github.com/juju/juju/api/agent"
 	"github.com/juju/juju/api/base"
+	"github.com/juju/juju/mongo"
+	jworker "github.com/juju/juju/worker"
 )
+
+// Logger defines the logging methods used by the worker.
+type Logger interface {
+	Infof(string, ...interface{})
+	Debugf(string, ...interface{})
+}
 
 // ManifoldConfig provides the dependencies for the
 // agent config updater manifold.
 type ManifoldConfig struct {
 	AgentName     string
 	APICallerName string
+	Logger        Logger
 }
 
 // Manifold defines a simple start function which
@@ -39,7 +48,8 @@ func Manifold(config ManifoldConfig) dependency.Manifold {
 			}
 
 			// Grab the tag and ensure that it's for a machine.
-			tag, ok := agent.CurrentConfig().Tag().(names.MachineTag)
+			currentConfig := agent.CurrentConfig()
+			tag, ok := currentConfig.Tag().(names.MachineTag)
 			if !ok {
 				return nil, errors.New("agent's tag is not a machine tag")
 			}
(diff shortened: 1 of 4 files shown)


howbazaar:
NOTE: this doesn't make mongo-memory-profile writable yet. That is coming in a follow up branch.

worker/agentconfigupdater/manifold.go
```diff
@@ -39,7 +48,8 @@ func Manifold(config ManifoldConfig) dependency.Manifold {
 			}
 
 			// Grab the tag and ensure that it's for a machine.
-			tag, ok := agent.CurrentConfig().Tag().(names.MachineTag)
+			currentConfig := agent.CurrentConfig()
```
wallyworld: What's wrong with inlining? That's the idiomatic Go way to do it.

worker/agentconfigupdater/manifold.go
```diff
@@ -81,11 +102,20 @@ func Manifold(config ManifoldConfig) dependency.Manifold {
 					info.PrivateKey = existing.PrivateKey
 				}
 				config.SetStateServingInfo(info)
+				if needsRestart {
```
wallyworld: "needsRestart" is a little misleading as it's really "mongoMemoryProfileChanged".
That change would make the code here make more sense since on the surface there's no relation between needing a restart (in general), and writing a new memory profile setting.

There could be a needsRestart variable as well for use below. Currently it would just be 
```
needsRestart := mongoMemoryProfileChanged
```
but maybe more conditions could be added later.

worker/agentconfigupdater/manifold.go
howbazaar: Because I use currentConfig further down the method.

worker/agentconfigupdater/manifold.go
howbazaar: Fair enough.

howbazaar:
$$merge$$

howbazaar:
$$merge$$

failure in MachineLegacyLeasesSuite.TestDyingMachine

PR #9857, opened 2019-03-12
Fix the state testing implementation of WaitForModelWatchersIdle.

howbazaar:
One fix and a few drive bys here. Using a logger for the output so we get timestamps in the output.

Also, the state testing implementation needed to wait for the txn watcher sync before creating the idle channel, otherwise we may get notified about idle while the txn watcher is processing but before it has sent any values.

This fixes a cause of intermittent failures in some state watcher tests.


howbazaar:
$$merge$$

PR #9859, opened 2019-03-12
Fix TestWatchApplication race

howbazaar:
There is a race in application_test.go:2855: ApplicationSuite.TestWatchApplication which causes a double event.

The destruction needs to have been processed by the txn watcher before the watcher in the test is started or the destroy notification may come through as an additional event.


howbazaar:
$$merge$$

howbazaar:
$$merge$$

howbazaar:
$$merge$$

weird bot

howbazaar:
$$merge$$

bot isn't starting merge job

PR #9860, opened 2019-03-12
Fix race in charmrevision worker test TestDelayedUpdateError

howbazaar:
Race in the test where clock is advancing before the worker is waiting.


howbazaar:
$$merge$$

PR #9869, opened 2019-03-13
Missing synchronisation in StateSuite.

howbazaar:
Found another failure during stressing the state watchers.

Went through the entire file updating watcher implementations with the `WaitForModelWatchersIdle` method.

NOTE: string watchers don't need this as they initially query the database for the set of information and change notifications cause them to reevaluate their state, but the string watchers only notify on changes of value.

During a pass through the file found that a test checking controller config created a machine unnecessarily, so I removed it.

### DIFF
--- a/state/state_test.go
+++ b/state/state_test.go
@@ -96,6 +96,7 @@ func (s *StateSuite) SetUpTest(c *gc.C) {
 	model, err := s.State.Model()
 	c.Assert(err, jc.ErrorIsNil)
 	s.model = model
+	s.WaitForModelWatchersIdle(c, s.Model.UUID())
 }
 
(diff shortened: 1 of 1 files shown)


howbazaar:
FAIL: state_test.go:3384: StateSuite.TestWatchCleanupsBulk
```
state_test.go:3405:
    wc.AssertOneChange()
testing/watcher.go:95:
    c.Fatalf("watcher sent unexpected change")
... Error: watcher sent unexpected change

```

howbazaar:
$$merge$$

timClicks:
Is this something to watch out for in application code/workers or is it an issue centered within tests?

PR #9874, opened 2019-03-13
Add webscale test arg to provide snap .asserts file

achilleasa:
## Description of change

This PR updates the webscale tests to accept a URI for fetching a snap asserts file that gets passed together with the mongo snap when bootstrapping the test controller.

### DIFF
--- a/acceptancetests/assess_deploy_webscale.py
+++ b/acceptancetests/assess_deploy_webscale.py
@@ -231,8 +231,16 @@ def parse_args(argv):
     parser.add_argument(
         '--db-snap-path',
         help="URL to a mongo snap to override the mongo version used for the "
-             "test controller; if a URL is specified, it will be download "
+             "test controller; if a URL is specified, it will be downloaded "
              "locally before bootstrapping",
+        default="",
+    )
+    parser.add_argument(
+        '--db-asserts-path',
+        help="URL to a mongo asserts file to be used in conjunction with a "
+             "provided mongo snap to bootstrap the test controller; if a URL "
+             "is specified, it will be downloaded locally before "
+             "bootstrapping",
         default="",
     )
     parser.add_argument(
@@ -250,7 +258,7 @@ def parse_args(argv):
 
 
 def mongo_snap_settings(args):
-    if not args.db_snap_path:
+    if not args.db_snap_path and not args.db_asserts_path:
         log.info("Using built-in mongo")
         return "", ""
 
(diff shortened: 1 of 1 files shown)


acceptancetests/assess_deploy_webscale.py
```diff
@@ -250,7 +258,7 @@ def parse_args(argv):
 
 
 def mongo_snap_settings(args):
-    if not args.db_snap_path:
+    if not args.db_snap_path and not args.db_asserts_path:
```
mitechie: should we error if you get one and not the other?

acceptancetests/assess_deploy_webscale.py
```diff
@@ -263,7 +271,7 @@ def mongo_snap_settings(args):
 
     # Fetch snap if using a remote URL. Juju expects the snap file to match
     # "juju-db_\d+.snap" so we should rename it accordingly.
-    dst_snap_path = "/tmp/juju-db_1.snap"
+    dst_snap_path = "/tmp/juju-db_6.snap"
```
mitechie: why is this now 6 vs 1?

acceptancetests/assess_deploy_webscale.py
achilleasa: I think no. I have rigged the CI jobs so that an empty --db-snap-path gets passed when running against the built-in mongo (see https://jenkins.juju.canonical.com/job/MongoWebscaleTests/)

acceptancetests/assess_deploy_webscale.py
achilleasa: No particular reason. I fetched the actual snap we are using internally from the edge channel and it has the `_6` suffix so I just copied that over; I think it doesn't really matter (AFAICT the asserts file does not include a reference to the snap version? suffix)

acceptancetests/assess_deploy_webscale.py
mitechie: Ok, fair enough. Maybe worth a comment in code then to make clear here?

acceptancetests/assess_deploy_webscale.py
mitechie: Ok, fair enough. So it's not something we have to continuously edit/keep up to date or the like?

achilleasa:
@mitechie it seems so; we will definitely know if the version has any significance when the mongodb-snap feature is fully implemented

acceptancetests/assess_deploy_webscale.py
achilleasa: Will do!

achilleasa: @mitechie Now that I think about it, if you don't provide both, then `juju bootstrap` will fail with an error when validating the arguments so I think we can actually get away without adding a comment here. What do you think?

mitechie:
works for me then, I'm just asking questions as a noob looking at the code pretending to be the person looking at this 6mo from now.

achilleasa:
$$merge$$

PR #9880, opened 2019-03-14
k8s storage is opinionated model config

wallyworld:
## Description of change

Juju now uses opinionated storage defaults when provisioning storage on k8s. The main changes:
- add-k8s queries the cluster and prompts for storage if no suitable storage class can be found
- the k8s provider gains config support for storing the workload and operator storage classes as model default attributes
- a bunch of code used to search for k8s storage classes is deleted

## QA steps

Use add-k8s to import microk8s
check that model-defaults shows hostpath storage for operator and workload
add a model
check that model config shows hostpath storage for operator and workload
deploy a charm with storage
create a storage pool and deploy a charm with pool storage

### DIFF
--- a/caas/kubernetes/provider/init.go
+++ b/caas/kubernetes/provider/init.go
@@ -11,8 +11,8 @@ import (
 )
 
 var k8sCloudCheckers map[string]k8slabels.Selector
-var clusterPreferredWorkloadStorage map[string]caas.PreferredStorage
-var clusterPreferredOperatorStorage map[string]caas.PreferredStorage
+var jujuPreferredWorkloadStorage map[string]caas.PreferredStorage
+var jujuPreferredOperatorStorage map[string]caas.PreferredStorage
 
 func init() {
 	caas.RegisterContainerProvider(CAASProviderType, providerInstance)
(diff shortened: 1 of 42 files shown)


caas/kubernetes/provider/init.go
```diff
@@ -40,6 +41,12 @@ func init() {
 			Provisioner: "kubernetes.io/aws-ebs",
 		},
 	}
+
+	// clusterPreferredOperatorStorage defines the opinionated storage
+	// that Juju requires to be available on supported clusters to
+	// provision storage for operators.
+	// TODO - support regional storage for GCE etc
+	clusterPreferredOperatorStorage = clusterPreferredWorkloadStorage
```
ycliuhw: we probably should rename `clusterPreferredOperatorStorage` to `jujuPreferredStorage` for both app `operator` and `controller` pod later.

wallyworld: yeah, ok, will do that before landing

wallyworld:
$$merge$$

wallyworld:
[extend cloud model to allow default storage class to be specified](https://trello.com/c/03FGMyWJ/645-extend-cloud-model-to-allow-default-storage-class-to-be-specified)

PR #9889, opened 2019-03-15
Cloud commands operate on controller by default

wallyworld:
## Description of change

The various cloud commands used to add/remove/list/update clouds are changed to operate by default on the controller rather than the local cache. After bootstrap, the most common use case is to manage clouds on the running controller, so the default command behaviour is now adjusted to cater for that.

If a new cloud is needed to be added to the local cache so another controller can be bootstrapped, there's a new --local option for that.

## QA steps

Run the various cloud commands with/without --local and check that the controller or local cache is updated as specified.

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -60,9 +60,14 @@ Adds a k8s endpoint and credential to Juju.`[1:]
 
 var usageAddCAASDetails = `
 Creates a user-defined cloud based on a k8s cluster.
+
 The new k8s cloud can then be used to bootstrap into, or it
-can be added to an existing controller with the --controller option.
-Specify non default kubeconfig file location using $KUBECONFIG
+can be added to an existing controller; the current controller
+is used unless the --controller option is specified. If you just
+want to update the local cache and not a running controller, use
+the --local option.
+
+Specify a non default kubeconfig file location using $KUBECONFIG
 environment variable or pipe in file content from stdin.
 
 The config file can contain definitions for different k8s clusters,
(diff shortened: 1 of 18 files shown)


mitechie:
-1 on changing the default as this seeds like something that would be a juju3 thing if we change how commands work like this.

wallyworld:
> -1 on changing the default as this seeds like something that would be a juju3 thing if we change how commands work like this.

As is there's significant confusion in the field which will only be made worse as we transition to multi-cloud controllers . The commands work the same when no controller is running, which is when local cache updates are required to set things up. We started work last cycle to fix the UX around cloud and credential management and this is part of that. Credential management is particularly problematic at the moment. We have a mandate to make the user experience on rails, even if sometimes there's a slight change in behaviour.

mitechie:
Right but I always have a jaas controller. I use it and go do other work. I
think it's a mistake to assume days later when I mess with clouds or
credentials I know and I tend to use it again this controller. It's be
different if controllers "times out" and logged me out or something so that
I don't end up accidentally badding things where I don't intend.

wallyworld:
Based on discussions with many external stakeholders, and internally, and the plan for credential CLI improvements geared towards a nice user experience for multi-cloud controller, we're going ahead with this change.

wallyworld:
$$merge$$

manadart:
From this patch onwards, the _nw-clouds-display_ functional test has been red.
http://10.125.0.203:8080/view/functional-tests-green/job/nw-clouds-display/1281/console

manadart:
@SimonRichardson has it in hand with https://github.com/juju/juju/pull/9978.

PR #9890, opened 2019-03-15
Restore backup from HA mode

SimonRichardson:
## Description of change

The following restores backup from HA mode, but will fail because
mongo 3.6.3 (which is what we're using) has a bug in it, that we
can't patch atm.

Anyway, once that patch/fix lands, we can turn on this test!

## QA steps

Setup acceptance tests correctly, then run the following command.

```
./assess_recovery.py --ha-backup
```

### DIFF
--- a/acceptancetests/assess_recovery.py
+++ b/acceptancetests/assess_recovery.py
@@ -64,6 +64,15 @@ def enable_ha(bs_manager, controller_client):
         controller_client, bs_manager.known_hosts)
     bs_manager.known_hosts = remote_machines
 
+def disable_ha(bs_manager, controller_client):
+    """Disable HA and wait for the controllers to be ready."""
+    log.info("Disabling HA.")
+    controller_client.wait_for(controller_client.remove_machines(["1", "2"], controller=True))
+    show_controller(controller_client)
+    remote_machines = get_remote_machines(
+        controller_client, bs_manager.known_hosts)
+    bs_manager.known_hosts = remote_machines
+
 def show_controller(client):
     controller_info = client.show_controller(format='yaml')
     log.info('Controller is:\n{}'.format(controller_info))
@@ -133,14 +142,12 @@ def assess_recovery(bs_manager, strategy, charm_series):
     if strategy == 'ha-backup':
         enable_ha(bs_manager, controller_client)
         backup_file = controller_client.backup()
-        # at the moment we can't currently restore a backup in HA without
-        # removing the controllers and replica set. This test will need to
-        # be fix to accommodate this setup.
+        disable_ha(bs_manager, controller_client)
     else:
         backup_file = controller_client.backup()
-        create_tmp_model(client)
-        restore_backup(bs_manager, controller_client, backup_file)
-    
+
+    create_tmp_model(client)
+    restore_backup(bs_manager, controller_client, backup_file)
     log.info("Test complete.")
 
 def main(argv):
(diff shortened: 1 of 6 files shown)


acceptancetests/assess_recovery.py
```diff
@@ -64,6 +64,15 @@ def enable_ha(bs_manager, controller_client):
         controller_client, bs_manager.known_hosts)
     bs_manager.known_hosts = remote_machines
 
+def disable_ha(bs_manager, controller_client):
+    """Disable HA and wait for the controllers to be ready."""
+    log.info("Disabling HA.")
+    controller_client.wait_for(controller_client.remove_machines(["1", "2"], controller=True))
+    show_controller(controller_client)
```
mitechie: what's the show_controller doing here?

acceptancetests/assess_recovery.py
```diff
@@ -64,6 +64,15 @@ def enable_ha(bs_manager, controller_client):
...
+    remote_machines = get_remote_machines(
+        controller_client, None)
+    bs_manager.known_hosts = remote_machines
```
mitechie: I think it'd be better for this to just return the remote_machines vs monkey with a bs_manager that's not directly tied to this.

acceptancetests/assess_recovery.py
```diff
@@ -133,14 +142,12 @@ def assess_recovery(bs_manager, strategy, charm_series):
     if strategy == 'ha-backup':
         enable_ha(bs_manager, controller_client)
         backup_file = controller_client.backup()
-        # at the moment we can't currently restore a backup in HA without
-        # removing the controllers and replica set. This test will need to
-        # be fix to accommodate this setup.
+        disable_ha(bs_manager, controller_client)
```
mitechie: this would then turn into 

    bs_manager.remote_machines = disable_ha(contrller_client)

Though we'd have to check for the failure cases (timeout/other failure to remove the HA units)

acceptancetests/jujupy/wait_condition.py
```diff
@@ -227,6 +227,38 @@ def do_raise(self, model_name, status):
                         self.machine)
 
 
+class WaitMachinesNotPresent(BaseCondition):
```
mitechie: I'd prefer we just update the other condition to take one or more machines.

acceptancetests/assess_recovery.py
SimonRichardson: It's really helpful when debugging to see what state the controller is in. Same for enable_ha.

acceptancetests/assess_recovery.py
SimonRichardson: So this is how enable_ha does it, so just following suit.

acceptancetests/jujupy/wait_condition.py
SimonRichardson: I'm not sure about this, I think it's preferable if the wait conditions just did one thing?

acceptancetests/assess_recovery.py
mitechie: maybe a pre-log line then that helps the log show that this is kicked off as log/tracking output?

acceptancetests/jujupy/wait_condition.py
mitechie: As a python dev I'd say it is doing one thing. Waiting for machines to go away. If you're doing one or 3 it's the same job with the same checks and the same process.

SimonRichardson:
$$merge$$

PR #9894, opened 2019-03-18
Add --aks param to add-k8s

Veebers:
## Description of change

Ability to add AKS cluster as K8s cloud using interactive prompts or command
arguments.

## QA steps

Create an AKS cluster (either via the web UI or the cli tool), run ```az login``` then run:
    ```juju add-k8s --aks``` (or use any of the provided arguments as per help text.

## Documentation changes

Help text updated.

## Bug reference

n/a

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -62,7 +62,7 @@ var usageAddCAASDetails = `
 Creates a user-defined cloud based on a k8s cluster.
 The new k8s cloud can then be used to bootstrap into, or it
 can be added to an existing controller with the --controller option.
-Speficify non default kubeconfig file location using $KUBECONFIG
+Specify non default kubeconfig file location using $KUBECONFIG
 environment variable or pipe in file content from stdin.
 
 The config file can contain definitions for different k8s clusters,
@@ -70,12 +70,13 @@ use --cluster-name to pick which one to use.
 It's also possible to select a context by name using --context-name.
 
 When running add-k8s the underlying cloud/region hosting the cluster need to be
-detected to enable storage to be correctly configured.if the cloud/region cannot
+detected to enable storage to be correctly configured. If the cloud/region cannot
 be detected automatically, use --region <cloudType/region> to specify the host
 cloud type and region.
 
-When adding a GKE cluster, you can use the --gke option to interactively be stepped
-through the registration process, or you can supply the necessary parameters directly.
+When adding a GKE or AKS cluster, you can use the --gke or --aks option to
+interactively be stepped through the registration process, or you can supply the
+necessary parameters directly.
 
 Examples:
     juju add-k8s myk8scloud
@@ -90,6 +91,10 @@ Examples:
     juju add-k8s --gke --project=myproject myk8scloud
     juju add-k8s --gke --credential=myaccount --project=myproject myk8scloud
     juju add-k8s --gke --credential=myaccount --project=myproject --region=someregion myk8scloud
+
+    juju add-k8s --aks myk8scloud
+    juju add-k8s --aks --cluster-name mycluster myk8scloud
+    juju add-k8s --aks --cluster-name mycluster --resource-group myrg myk8scloud
 
 See also:
     remove-k8s
@@ -123,6 +128,9 @@ type AddCAASCommand struct {
 	// credential is the credential to use when accessing the cluster.
 	credential string
 
+	// resourceGroup is the resource group name for the cluster.
+	resourceGroup string
+
 	// hostCloudRegion is the cloud region that the nodes of cluster (k8s) are running in.
 	// The format is <cloudType/region>.
 	hostCloudRegion string
@@ -134,6 +142,7 @@ type AddCAASCommand struct {
 	brokerGetter BrokerGetter
 
 	gke        bool
+	aks        bool
 	k8sCluster k8sCluster
 
 	cloudMetadataStore    CloudMetadataStore
@@ -185,7 +194,9 @@ func (c *AddCAASCommand) SetFlags(f *gnuflag.FlagSet) {
 	f.StringVar(&c.workloadStorage, "storage", "", "kubernetes storage class for workload storage")
 	f.StringVar(&c.project, "project", "", "project to which the cluster belongs")
 	f.StringVar(&c.credential, "credential", "", "the credential to use when accessing the cluster")
+	f.StringVar(&c.resourceGroup, "resource-group", "", "the Azure resource group of the AKS cluster")
 	f.BoolVar(&c.gke, "gke", false, "used when adding a GKE cluster")
+	f.BoolVar(&c.aks, "aks", false, "used when adding an AKS cluster")
 }
 
 // Init populates the command with the args from the command line.
@@ -203,7 +214,24 @@ func (c *AddCAASCommand) Init(args []string) (err error) {
 		if c.contextName != "" {
 			return errors.New("do not specify context name when adding a GKE cluster")
 		}
+		if c.k8sCluster == nil {
+			c.k8sCluster = newGKECluster()
+		}
+		if err := c.k8sCluster.ensureExecutable(); err != nil {
+			return errors.Trace(err)
+		}
 	} else {
+		if c.aks {
+			if c.contextName != "" {
+				return errors.New("do not specify context name when adding a AKS cluster")
+			}
+			if c.k8sCluster == nil {
+				c.k8sCluster = newAKSCluster()
+			}
+			if err := c.k8sCluster.ensureExecutable(); err != nil {
+				return errors.Trace(err)
+			}
+		}
 		if c.project != "" {
 			return errors.New("do not specify project unless adding a GKE cluster")
 		}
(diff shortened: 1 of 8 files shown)


cmd/juju/caas/add.go
```diff
@@ -61,20 +61,21 @@ var usageAddCAASDetails = `
...
 can be added to an existing controller with the --controller option.
-Speficify non default kubeconfig file location using $KUBECONFIG
+Specify non default kubeconfig file location using $KUBECONFIG
 environment variable or pipe in file content from stdin.
 
 The config file can contain definitions for different k8s clusters,
 use --cluster-name to pick which one to use.
 It's also possible to select a context by name using --context-name.
 
 When running add-k8s the underlying cloud/region hosting the cluster need to be
```
wallyworld: needs

cmd/juju/caas/add.go
```diff
@@ -202,7 +213,24 @@ func (c *AddCAASCommand) Init(args []string) (err error) {
 		if c.contextName != "" {
 			return errors.New("do not specify context name when adding a GKE cluster")
 		}
+		if c.k8sCluster == nil {
+			c.k8sCluster = newGKECluster()
+		}
+		if err := c.k8sCluster.ensureExecutable(); err != nil {
+			return errors.Trace(err)
+		}
 	} else {
```
wallyworld: put else if on one line

cmd/juju/caas/add.go
```diff
@@ -272,33 +300,52 @@ func (c *AddCAASCommand) newCloudCredentialFromKubeConfig(reader io.Reader, cont
...
+	return c.k8sCluster.getKubeConfig(p)
+}
+
+func (c *AddCAASCommand) getAKSKubeConfg(ctx *cmd.Context) (io.Reader, string, error) {
+	p := &clusterParams{
+		name:          c.clusterName,
+		resourceGroup: c.resourceGroup,
+	}
+
+	if p.name == "" || p.resourceGroup == "" {
```
wallyworld: // If any items are missing, prompt for them.

cmd/juju/caas/add.go
```diff
@@ -272,33 +300,52 @@ func (c *AddCAASCommand) newCloudCredentialFromKubeConfig(reader io.Reader, cont
...
+		var err error
+		p, err = c.k8sCluster.interactiveParams(ctx, p)
 		if err != nil {
 			return nil, "", errors.Trace(err)
 		}
 	}
 	c.clusterName = p.name
-	c.hostCloudRegion = cluster.cloud() + "/" + p.region
-	return cluster.getKubeConfig(p)
+	return c.k8sCluster.getKubeConfig(p)
```
wallyworld: c.hostCloudRegion = c.k8sCluster.cloud() + "/" + p.region

cmd/juju/caas/aks.go
```diff
@@ -0,0 +1,287 @@
...
+	CommandRunner
+	azExecName string
+}
+
+func newAKSCluster() k8sCluster {
+	return &aks{CommandRunner: &defaultRunner{}}
+}
+
+func (a *aks) cloud() string {
+	return "aks"
```
wallyworld: This has to be "azure", ie a recognised Juju cloud type.
I'm surprised this passed manual testing. I suspect the missing code to set cloud/region explains it.

cmd/juju/caas/aks.go
```diff
@@ -0,0 +1,287 @@
...
+	if a.azExecName != "" {
+		return nil
+	}
+
+	cmd := []string{"which", "az"}
+	err := collapseRunError(runCommand(a, cmd, ""))
+	if err != nil {
+		cmd = []string{"which", "azure-cli"}
+		err := collapseRunError(runCommand(a, cmd, ""))
+		errAnnotationMessage := "neither az or azure-cli (version > 2.0.52) found. Please install, login, and try again"
```
wallyworld: Instead of "Please install"....

Tell them which one to install and how, eg "snap install foo" or "apt install foo" or whatever

cmd/juju/caas/aks.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+func (a *aks) ensureExecutable() error {
+	if a.azExecName != "" {
+		return nil
+	}
+
+	cmd := []string{"which", "az"}
+	err := collapseRunError(runCommand(a, cmd, ""))
+	if err != nil {
```
wallyworld: Doing the simple == nil case first reads better
```
if err == nil {
    a.azExecName = "az"
} else {
}
```

cmd/juju/caas/aks.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+		a.azExecName = "azure-cli"
+	} else {
+		a.azExecName = "az"
+	}
+
+	// check that we are logged in, there is no way to provide login details to a separate command.
+	cmd = []string{a.azExecName, "account", "show"}
+	err = collapseRunError(runCommand(a, cmd, ""))
+	errAnnotationMessage := fmt.Sprintf("please run '%s login' to setup account", a.azExecName)
```
wallyworld: .. and try running add-k8s again

or something like that

cmd/juju/caas/cluster.go
```diff
@@ -32,6 +33,16 @@ func getEnv(key string) string {
 	return result
 }
 
+func collapseRunError(result *exec.ExecResponse, err error) error {
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if result.Code != 0 {
```
wallyworld: Should check for result == nil

cmd/juju/caas/aks_test.go
```diff
@@ -0,0 +1,573 @@
...
+	stdin := strings.NewReader("mycluster\n")
+	out := &bytes.Buffer{}
+	ctx := &cmd.Context{
+		Dir:    c.MkDir(),
+		Stdout: out,
+		Stderr: ioutil.Discard,
+		Stdin:  stdin,
+	}
+	expected := `
+Available Clusters In Resource Group TestRG
```
wallyworld: For top level titles, should be "TestRG" etc?? Here and elsewhere

cmd/juju/caas/add.go
Veebers: I'm not sure I follow this, this else includes checks for --aks and for when neither aks/gke is passed. Perhaps I need to be clearer here?

cmd/juju/caas/aks.go
Veebers: very true!

cmd/juju/caas/aks.go
Veebers: hah, are we ok with having urls to web pages for install? As it stands right now the snap is not suitable to be used (not in stable and an older version.) and the deb requires addition of a ppa + key etc. as per https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-apt?view=azure-cli-latest

cmd/juju/caas/aks.go
Veebers: You where right about that Ian.

mitechie:
Are we adding a unique --arg for each cloud then? Or should we have a more generic type=aks or cloud=aks or something? Is there a spec for this we can review?

mitechie:
I gave this a go and hit the following. 

    juju add-k8s --aks testaks
    ERROR neither az or azure-cli (version > 2.0.52) found. Please install, login, and try again:

- Should there be a colon at the end of the error message?

*Note* the gke one suggest installing a snap gcloud that's not found so I tried a snap here:


    % snap find azure-cli
    Name       Version  Publisher  Notes  Summary
    azure-cli  2.0.25   azpycli✓   -      Microsoft Azure CLI


    sudo snap install azure-cli --candidate
    juju add-k8s --aks testaks
    ERROR please run 'azure-cli login' to setup account: ERROR: Please run 'az login' to setup account.

- I've got one of the commands but I'm getting errors about logging in with two tools?


    azure-cli login --allow-no-subscriptions
    To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the 
    code EEEEEEE2 to authenticate.
    [
      {
        "cloudName": "AzureCloud",
        "id": "97e71113",
        "isDefault": true,
        "name": "N/A(tenant level account)",
        "state": "Enabled",
        "tenantId": "97e71113",
        "user": {
          "name": "rick.harding@canonical.com",
          "type": "user"
        }
      }
    ]
    juju add-k8s --aks testaks 
    ERROR ERROR: The subscription '97e71113' could not be found.

This wasn't clear but I went in and added a subscription to my account to test out. I have a subscription in my azure account:

    

Pay-As-You-Go | ba2c5fa1 | Account admin | Active

But I still get the same error:

    juju add-k8s --aks testaks
    ERROR ERROR: The subscription '97e71113' could not be found.


In summary a couple of notes on the error messaging here. I also think that adding cloud specific args is a poor user experience and brings up cases where things aren't handled well like 

    % juju add-k8s --aks --gke testaks
    ERROR gcloud command not found, please 'snap install gcloud' then try again:

If we get into choices of things like this it'd be great if it supported the normal add-cloud/add-credential interactive mode and for fallback from that was a more generic config path or worst case a single flag that you could specify a valid type of k8s you're looking to add for specific help.

wallyworld:
> Are we adding a unique --arg for each cloud then? Or should we have a more generic type=aks or cloud=aks or something? Is there a spec for this we can review?

The implementation is as per Mark's request in the realk meetings.

Veebers:
@mitechie I've cleaned up some things re: install etc. Unfortunately the snap that is available is neither stable nor a suitable version to use (the candidate snap generates invalid json for the creds (instead of the required yaml) for instance.
I've chucked in the URL for the install instructions as unfortunately it's not in universe. We have follow up cards to try reach out to whomever is responsible for snapping the cli tool to see if we can get a stable newer version sorted. 

The subscription issue may be a result of an out-of-date cli tool (as per the snap comments above). I had success with using 'az login' and interacting with the webpage that popped up.

Veebers:
$$merge$$

wallyworld:
[add-k8s --azure support](https://trello.com/c/WRjrfh0F/627-add-k8s-azure-support)

mitechie:
Since the azure-cli snap is unusable should we just focus on the az tool then? If we've tested that works and since we're outputting errors for both commands vs one or the other should we just focus on one we know and not lead users astray into worlds we know won't help them?

PR #9899, opened 2019-03-18
Issue error for setcharm

SimonRichardson:
## Description of change

The following will attempt to throw an error if your controller is
greater than 2.5.2, yet your agents are less than and equal too
2.5.2. This prevents a potential version miss-match of LXDProfile
code attempting to use old code vs the new code. From the very
outset we tell the operator that they should upgrade.

## QA steps

Check out `2.5`:

```
export JUJU_DEV_FEATURE_FLAGS=instance-mutater
juju bootstrap lxd test
```

Check out to this branch:

```
juju upgrade-model -m controller --build-agent=true
juju upgrade-charm lxd-profile --path ./testcharms/charm-repo/quantal/lxd-profile
```

This should return the following error:

```
Added charm "local:bionic/lxd-profile-1" to the model.
ERROR Unable to upgrade LXDProfile charms with the current model version. Please run juju upgrade-juju to upgrade the current model to match your controller.
```

To fix this, run the following:

```
juju upgrade-model
juju upgrade-charm lxd-profile --path ./testcharms/charm-repo/quantal/lxd-profile
```

This should succeed!

## Documentation changes

This will prevent upgrading of greater than 2.5.2 from units on 2.5.2
or less, to prevent conflicting code running.

### DIFF
--- a/apiserver/facades/client/application/application.go
+++ b/apiserver/facades/client/application/application.go
@@ -719,6 +719,21 @@ func parseSettingsCompatible(charmConfig *charm.Config, settings map[string]stri
 	return changes, nil
 }
 
+type setCharmParams struct {
+	AppName               string
+	Application           Application
+	Channel               csparams.Channel
+	ConfigSettingsStrings map[string]string
+	ConfigSettingsYAML    string
+	ResourceIDs           map[string]string
+	StorageConstraints    map[string]params.StorageConstraints
+	Force                 forceParams
+}
+
+type forceParams struct {
+	ForceSeries, ForceUnits, Force bool
+}
+
 // Update updates the application attributes, including charm URL,
 // minimum number of units, charm config and constraints.
 // All parameters in params.ApplicationUpdate except the application name are optional.
@@ -741,17 +756,17 @@ func (api *APIBase) Update(args params.ApplicationUpdate) error {
 		// TODO(ericsnow) Support it?
 		channel := app.Channel()
 		if err = api.updateCharm(
-			args.ApplicationName,
-			app,
+			setCharmParams{
+				AppName:     args.ApplicationName,
+				Application: app,
+				Channel:     channel,
+				Force: forceParams{
+					ForceSeries: args.ForceSeries,
+					ForceUnits:  args.ForceCharmURL,
+					Force:       args.Force,
+				},
+			},
 			args.CharmURL,
-			channel,
-			nil, // charm settings (strings map)
-			"",  // charm settings (YAML)
-			args.ForceSeries,
-			args.ForceCharmURL,
-			args.Force,
-			nil, // resource IDs
-			nil, // storage constraints
 		); err != nil {
 			return errors.Trace(err)
 		}
@@ -795,17 +810,8 @@ func (api *APIBase) Update(args params.ApplicationUpdate) error {
 // this is analogous to setCharmWithAgentValidation, minus the validation around
 // setting the profile charm.
 func (api *APIBase) updateCharm(
-	appName string,
-	application Application,
+	params setCharmParams,
 	url string,
-	channel csparams.Channel,
-	configSettingsStrings map[string]string,
-	configSettingsYAML string,
-	forceSeries,
-	forceUnits,
-	force bool,
-	resourceIDs map[string]string,
-	storageConstraints map[string]params.StorageConstraints,
 ) error {
 	curl, err := charm.ParseURL(url)
 	if err != nil {
@@ -815,11 +821,7 @@ func (api *APIBase) updateCharm(
 	if err != nil {
 		return errors.Trace(err)
 	}
-	return api.applicationSetCharm(appName, application, charm,
-		channel, configSettingsStrings, configSettingsYAML,
-		forceSeries, forceUnits, force, resourceIDs,
-		storageConstraints,
-	)
+	return api.applicationSetCharm(params, charm)
 }
 
 // UpdateApplicationSeries updates the application series. Series for
@@ -885,17 +887,21 @@ func (api *APIBase) SetCharm(args params.ApplicationSetCharm) error {
 	}
 	channel := csparams.Channel(args.Channel)
 	return api.setCharmWithAgentValidation(
-		args.ApplicationName,
-		application,
+		setCharmParams{
+			AppName:               args.ApplicationName,
+			Application:           application,
+			Channel:               channel,
+			ConfigSettingsStrings: args.ConfigSettings,
+			ConfigSettingsYAML:    args.ConfigSettingsYAML,
+			ResourceIDs:           args.ResourceIDs,
+			StorageConstraints:    args.StorageConstraints,
+			Force: forceParams{
+				ForceSeries: args.ForceSeries,
+				ForceUnits:  args.ForceUnits,
+				Force:       args.Force,
+			},
+		},
 		args.CharmURL,
-		channel,
-		args.ConfigSettings,
-		args.ConfigSettingsYAML,
-		args.ForceSeries,
-		args.ForceUnits,
-		args.Force,
-		args.ResourceIDs,
-		args.StorageConstraints,
 	)
 }
 
@@ -905,17 +911,8 @@ func (api *APIBase) SetCharm(args params.ApplicationSetCharm) error {
 // the correct and only work around is to upgrade the units to match the
 // controller.
 func (api *APIBase) setCharmWithAgentValidation(
-	appName string,
-	application Application,
+	params setCharmParams,
 	url string,
-	channel csparams.Channel,
-	configSettingsStrings map[string]string,
-	configSettingsYAML string,
-	forceSeries,
-	forceUnits,
-	force bool,
-	resourceIDs map[string]string,
-	storageConstraints map[string]params.StorageConstraints,
 ) error {
 	curl, err := charm.ParseURL(url)
 	if err != nil {
@@ -925,6 +922,7 @@ func (api *APIBase) setCharmWithAgentValidation(
 	if err != nil {
 		return errors.Trace(err)
 	}
+	application := params.Application
 	// Check if the controller agent tools version is greater than the
 	// version we support for the new LXD profiles.
 	// Then check all the units, to see what their agent tools versions is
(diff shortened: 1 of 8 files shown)


SimonRichardson:
https://trello.com/c/zHeBn09N/1195-setcharm-version-checking-to-prevent-old-code-being-triggered

state/application.go
```diff
@@ -1180,6 +1182,40 @@ func (a *Application) SetCharm(cfg SetCharmConfig) (err error) {
...
+					return errors.Annotate(err, "cannot retrieve units")
+				}
+				for _, unit := range units {
+					unitTools, err := unit.AgentTools()
+					if err != nil {
+						return errors.Annotate(err, "cannot retrieve unit agent tools")
+					}
+					if unitVer := unitTools.Version; unitVer.Compare(epoch) <= 0 {
+						return errors.Errorf(
+							"Unable to upgrade LXDProfile charms with the current version. "+
```
hmlanigan: Let's be more specific about what needs to be upgraded, perhaps "Unable to upgrade LXDProfile charms with the current model version.  Please run juju upgrade-juju to upgrade the current model to match your controller."

state/application.go
```diff
@@ -1180,6 +1182,40 @@ func (a *Application) SetCharm(cfg SetCharmConfig) (err error) {
...
 	if profile := cfg.Charm.LXDProfile(); profile != nil {
+		// Check if the controller agent tools version is greater than the
+		// version we support for the new LXD profiles.
+		// Then check all the units, to see what their agent tools versions is
+		// so that we can ensure that everyone is aligned. If the units version
+		// is to low (i.e. less than the 2.5.2 epoch), then show an error
+		// message that the operator should upgrade to receive the latest
+		// LXD Profile changes.
+		if featureflag.Enabled(feature.InstanceMutater) {
+			epoch := version.Number{Major: 2, Minor: 5, Patch: 2}
```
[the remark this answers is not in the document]
SimonRichardson: I was hoping you'd spot that, I was very unsure what the epoch should be.

state/application.go
```diff
@@ -1180,6 +1182,40 @@ func (a *Application) SetCharm(cfg SetCharmConfig) (err error) {
 	// we don't need to check that this is a charm.LXDProfiler, as we can
 	// state that the function exists.
 	if profile := cfg.Charm.LXDProfile(); profile != nil {
+		// Check if the controller agent tools version is greater than the
+		// version we support for the new LXD profiles.
+		// Then check all the units, to see what their agent tools versions is
+		// so that we can ensure that everyone is aligned. If the units version
+		// is to low (i.e. less than the 2.5.2 epoch), then show an error
```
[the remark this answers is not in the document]
SimonRichardson: nps.

state/application.go
SimonRichardson: Yeah, I was really struggling with the correct wording here.

SimonRichardson:
> ERROR cannot upgrade application "lxd-profile" to charm "local:bionic/lxd-profile-1": cannot retrieve agent tools: operator image metadata for application "lxd-profile" not found (not found)

I thought I solved that, but if we move up the stack to apiserver it would solve that anyway. I'll have a play at moving up to the apiserver.

apiserver/facades/client/application/application.go
```diff
@@ -847,6 +850,56 @@ func (api *APIBase) SetCharm(args params.ApplicationSetCharm) error {
...
+		epoch := version.Number{Major: 2, Minor: 6, Patch: 0}
+		tools, err := application.AgentTools()
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				return errors.Annotate(err, "cannot retrieve agent tools")
+			}
+			// When trying to access the application agent tools version
+			// number it might not be available, so we need to fallback to
+			// the model.LatestToolsVersion.
+			agentVer = api.model.LatestToolsVersion()
```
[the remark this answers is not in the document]
hmlanigan: So you could end up where the controller version is 2.5.4, the agent version is 2.5.3 and the latestavailabletools could be 2.6.0.

apiserver/facades/client/application/application_unit_test.go
```diff
@@ -282,6 +347,43 @@ postgresql:
...
+		Charm:          &state.Charm{},
+		ConfigSettings: charm.Settings{"stringOption": "value"},
+	})
+}
+
+func (s *ApplicationSuite) TestLXDProfileSetCharmWithOldAgentVersion(c *gc.C) {
+	s.SetFeatureFlags(feature.InstanceMutater)
+
+	err := s.api.SetCharm(params.ApplicationSetCharm{
+		ApplicationName: "redis",
```
hmlanigan: I'd expect the charm upgrade with redis to succeed in the real world.  Then a new test to fail with a charm having a lxd profile.

apiserver/facades/client/application/application.go
```diff
@@ -847,6 +850,56 @@ func (api *APIBase) SetCharm(args params.ApplicationSetCharm) error {
...
 		return errors.Trace(err)
 	}
+	// Check if the controller agent tools version is greater than the
+	// version we support for the new LXD profiles.
+	// Then check all the units, to see what their agent tools versions is
+	// so that we can ensure that everyone is aligned. If the units version
+	// is too low (i.e. less than the 2.6.0 epoch), then show an error
+	// message that the operator should upgrade to receive the latest
+	// LXD Profile changes.
+	if featureflag.Enabled(feature.InstanceMutater) {
```
hmlanigan: I'd expect this only if the charm has a profile.

hmlanigan:
The first part of qa worked: bootstrap 2.5.2, deploy lxd-profile, upgrade controller model, upgrade lxd-profile charm (expected error).  Then I upgraded the default model and tried to upgrade the lxd-profile charm:
$ juju upgrade-charm lxd-profile --path ./testcharms/charm-repo/quantal/lxd-profile
Added charm "local:bionic/lxd-profile-2" to the model.
ERROR unable to set charm profile: upgrade charm profile already in process for machine 0, profile from "local:bionic/lxd-profile-1"

However this PR is a stepping stone to the new lxd profile way.

apiserver/facades/client/application/backend.go
```diff
@@ -154,6 +158,7 @@ type Model interface {
 	Tag() names.Tag
 	Type() state.ModelType
 	ModelConfig() (*config.Config, error)
+	LatestToolsVersion() version.Number
```
hmlanigan: please remove, no longer used.

apiserver/facades/client/application/backend.go
```diff
@@ -5,7 +5,9 @@ package application
 
 import (
 	"github.com/juju/juju/core/model"
+	"github.com/juju/juju/tools"
```
hmlanigan: import stanzas are incorrect.

state/application_test.go
```diff
@@ -28,6 +27,7 @@ import (
 	"github.com/juju/juju/core/crossmodel"
 	"github.com/juju/juju/core/model"
 	"github.com/juju/juju/core/status"
+	"github.com/juju/juju/network"
```
hmlanigan: great!

SimonRichardson:
@hmlanigan I know what's going on, I'll fix it in the morning... It'll require some juggling of functions...

SimonRichardson:
@hmlanigan I've updated the Q&A to reflect the issue of `SetCharmProfile` not being idempotent.

apiserver/facades/client/application/application.go
```diff
@@ -2199,3 +2288,81 @@ func (api *APIBase) ApplicationsInfo(in params.Entities) (params.ApplicationInfo
...
+)
+
+func validateAgentVersions(application Application, cfgAgentVersion ModelConfig) error {
+	var agentVer version.Number
+	var noAgentVersion bool
+
+	// The epoch is set like this, because beta tags are less than release tags.
+	// So 2.6-beta1.1 < 2.6.0, even though the patch is greater than 0. To
+	// prevent the miss-match, we add the upper epoch limit.
+	epoch := version.Number{Major: 2, Minor: 5, Patch: math.MaxInt32}
```
jameinel: Does Version.Number.Compare(other.Number) not work?

If it doesn't work, this code should be factored out into something that just compares versions, rather than co-mingling the version comparison with loading what the versions/tools/agents are.

apiserver/facades/client/application/application.go
```diff
@@ -885,15 +921,68 @@ func (api *APIBase) applicationSetCharm(
...
 		return errors.Trace(err)
 	}
+	// Check if the controller agent tools version is greater than the
+	// version we support for the new LXD profiles.
+	// Then check all the units, to see what their agent tools versions is
+	// so that we can ensure that everyone is aligned. If the units version
+	// is too low (i.e. less than the 2.6.0 epoch), then show an error
+	// message that the operator should upgrade to receive the latest
+	// LXD Profile changes.
+	if featureflag.Enabled(feature.InstanceMutater) {
```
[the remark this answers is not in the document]
SimonRichardson: Correct.

apiserver/facades/client/application/application.go
```diff
@@ -2199,3 +2275,77 @@ func (api *APIBase) ApplicationsInfo(in params.Entities) (params.ApplicationInfo
...
+}
+
+func validateAgentVersions(application Application, modelConfig ModelConfig) error {
+	// The epoch is set like this, because beta tags are less than release tags.
+	// So 2.6-beta1.1 < 2.6.0, even though the patch is greater than 0. To
+	// prevent the miss-match, we add the upper epoch limit.
+	epoch := version.Number{Major: 2, Minor: 5, Patch: math.MaxInt32}
+
+	ver, err := getAgentToolsVersion(application)
+	if ver.Compare(epoch) >= 0 || errors.IsNotFound(err) {
```
jameinel: don't you need to check the error *before* you treat the version as valid?
we probably also want to trace the error if we get a failure that we don't understand, which seems like something:
    return errors.Trace(err)
}

I do believe application.AgentTools is only valid for CAAS models, because there you have 1 agent for the application, but no agents for the Units. However on IAAAS models, there is only the Unit.AgentTools or the Model agent tools. I guess maybe that is your IsNotFound check?

maybe just a comment explaining what we are looking for. Do we even need to bother with the Application version rather than just checking the ModelConfig desired agent version? (arguably you might want to check on the per-unit level, as that is the actual version of the agent that is running, but if you look at the ModelConfig, it is telling you the intent is to move to that version eventually, which seems sufficient for a pre-check like this.)

SimonRichardson: > maybe just a comment explaining what we are looking for. Do we even need to bother with the Application version rather than just checking the ModelConfig desired agent version? (arguably you might want to check on the per-unit level, as that is the actual version of the agent that is running, but if you look at the ModelConfig, it is telling you the intent is to move to that version eventually, which seems sufficient for a pre-check like this.)

This is exactly what I'm think and was hoping you'd agree....

apiserver/facades/client/application/application.go
```diff
@@ -2199,3 +2275,77 @@ func (api *APIBase) ApplicationsInfo(in params.Entities) (params.ApplicationInfo
...
+func getAgentToolsVersion(agentTools AgentTools) (version.Number, error) {
+	tools, err := agentTools.AgentTools()
+	if err != nil {
+		return version.Zero, err
+	}
+	return tools.Version.Number, nil
+}
+
+func getModelConfigVersion(modelConfig ModelConfig) (version.Number, error) {
+	agent, err := modelConfig.AgentVersion()
```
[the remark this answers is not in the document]
SimonRichardson: > Maybe its just the types/names that need updating? Instead of passing ModelConfig you would pass a modelVersions AgentVersion parameter?

I'll check.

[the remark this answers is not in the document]
SimonRichardson: Yes you're right, the naming of this is wrong, I'll update and amend.

SimonRichardson:
!!build!!

SimonRichardson:
!!build!!

> mgo failed to start.

SimonRichardson:
$$merge$$

PR #9901, opened 2019-03-19
Better support for choosing k8s storage for add-k8s

wallyworld:
## Description of change

When add-k8s selects the storage class to use, it now handles default classes with the beta annotation. It will also use the only matching storage class even if it is not annotated as the default. And if there's any matching storage classes for operator storage, the first will be used, preferring one annotated as the default.

## QA steps

juju add-k8s on various substrates.


wallyworld:
$$merge$$

wallyworld:
[prefer k8s cluster default sc if more than one exists, prompt if needed](https://trello.com/c/1olZaNbl/654-prefer-k8s-cluster-default-sc-if-more-than-one-exists-prompt-if-needed)

PR #9905, opened 2019-03-19
add-k8s --aks Remove use of quotes in interact Select.

Veebers:
## Description of change

Because the pollster Select uses html/template any quotes in the string are replaced with &#34;.

I was unable to make it display the quotes as expected, so I've removed them.
So instead of:
```Available Clusters In Resource Group &#34;someTestGroup&#34;```

we have the expected:
```Available Clusters In Resource Group someTestGroup```

## QA steps

Run ```juju add-k8s --aks testaks --storage default``` and observe the output.

## Documentation changes

n/a
## Bug reference

n/a


wallyworld:
$$merge$$

PR #9909, opened 2019-03-19
raftflag: Fix intermittent hang when raft is shutdown

babbageclunk:
## Description of change

Updates hashicorp/raft dependency to use our own fork to respect timeouts when sending RPC requests (at least until our PR is merged upstream). See https://github.com/hashicorp/raft/pull/313 for more detail.

## QA steps

Running the raftflag worker test under stress doesn't hang anymore.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -371,11 +371,12 @@
   version = "v0.5.0"
 
 [[projects]]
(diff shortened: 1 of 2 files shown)


babbageclunk:
$$merge$$

PR #9917, opened 2019-03-19
Add model.WatchMachines to model cache

hmlanigan:
## Description of change

Add a model.WatchMachines to model cache to be notified of new and removed machines in the model.  Functionality for the Instance Mutater.

## QA steps

Functionality not used by packages outside of model cache yet. Unit tests should pass.  No change to existing juju functionality.

### DIFF
--- a/core/cache/model.go
+++ b/core/cache/model.go
@@ -110,7 +110,7 @@ func (m *Model) WatchMachines() *ChangeWatcher {
 	}
 
 	w := newAddRemoveWatcher(machines...)
-	unsub := m.hub.Subscribe(m.modelTopic(modelMachineChange), w.changed)
+	unsub := m.hub.Subscribe(m.topic(modelMachineChange), w.changed)
 
 	w.tomb.Go(func() error {
 		<-w.tomb.Dying()
@@ -184,7 +184,7 @@ func (m *Model) updateMachine(ch MachineChange) {
 	if !found {
 		machine = newMachine(m.metrics, m.hub)
 		m.machines[ch.Id] = machine
-		m.hub.Publish(m.modelTopic(modelMachineChange), []string{ch.Id})
+		m.hub.Publish(m.topic(modelMachineChange), []string{ch.Id})
 	}
 	machine.setDetails(ch)
 
@@ -195,7 +195,7 @@ func (m *Model) updateMachine(ch MachineChange) {
 func (m *Model) removeMachine(ch RemoveMachine) {
 	m.mu.Lock()
 	delete(m.machines, ch.Id)
-	m.hub.Publish(m.modelTopic(modelMachineChange), []string{ch.Id})
+	m.hub.Publish(m.topic(modelMachineChange), []string{ch.Id})
 	m.mu.Unlock()
 }
 
--- a/core/cache/model.go
+++ b/core/cache/model.go
@@ -11,6 +11,7 @@ import (
 )
 
 const modelConfigChange = "model-config-change"
+const modelMachineChange = "model-machine-change"
 
 func newModel(metrics *ControllerGauges, hub *pubsub.SimpleHub) *Model {
 	m := &Model{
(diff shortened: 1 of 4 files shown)


hmlanigan:
The conflict will be resolved once PR 9916 lands

core/cache/model.go
```diff
@@ -11,6 +11,7 @@ import (
 )
 
 const modelConfigChange = "model-config-change"
+const modelMachineChange = "model-machine-change"
```
hmlanigan: Maybe this should be modelMachineAddRemove, since it doesn't cover changes to existing machines?

core/cache/model_test.go
```diff
@@ -135,6 +136,88 @@ func (s *ModelSuite) TestApplicationNotFoundError(c *gc.C) {
 	c.Assert(errors.IsNotFound(err), jc.IsTrue)
 }
 
+func (s *ModelSuite) TestMachineNotFoundError(c *gc.C) {
+	m := s.newModel(modelChange)
+	_, err := m.Machine("nope")
+	c.Assert(errors.IsNotFound(err), jc.IsTrue)
+}
+
+func (s *ControllerSuite) TestWatchMachineStops(c *gc.C) {
```
hmlanigan: These test work best with the ControllerSuite, but the piece tested isn't in controller.go... I debated about what file to place them in.  Or rewrite with new methods in export_test.go

core/cache/watcher.go
```diff
@@ -131,3 +131,98 @@ func (w *ConfigWatcher) configChanged(topic string, value interface{}) {
...
+	closed bool
+	mu     sync.Mutex
+}
+
+func newStringsWatcherBase(values ...string) *stringsWatcherBase {
+	// We use a single entry buffered channel for the changes.
+	// This allows the config changed handler to send a value when there
+	// is a change, but if that value hasn't been consumed before the
+	// next change, the second change is discarded.
+	ch := make(chan []string, 1)
```
hmlanigan: Should this be longer?  Right now, the code for sending changes, doesn't take into consideration that the prior change hasn't been grabbed yet.  However the end user of WatchMachines will require all changes.

SimonRichardson: So sending more values on the same buffered channel could potentially cause a deadlock, unless it's constantly grabbing the changes. If we know that the reads are congruent to the writes, this doesn't need buffering, it depends if you want to block the send.

core/cache/model.go
```diff
@@ -87,6 +91,42 @@ func (m *Model) Application(appName string) (*Application, error) {
...
+
+// WatchMachines creates a ChangeWatcher (strings watcher) to notify about
+// added and removed machines in the model.  The initial event contains
+// a slice of the current machine ids.
+func (m *Model) WatchMachines() *ChangeWatcher {
+	machines := make([]string, len(m.machines))
+	i := 0
+	for k := range m.machines {
+		machines[i] = k
+		i += 1
```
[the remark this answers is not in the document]
hmlanigan: To handle an index into the new slice as with range on a map, you don't get an index.   I could have done an append instead, though we know how big the new slice will be.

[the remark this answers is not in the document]
SimonRichardson: Ah right yeah, I was thinking `k` was a sequential int. It's not, it's a map. Ignore me :smile:

core/cache/watcher.go
manadart: I don't think we can implement this in the same way as the notify watcher (with a one-buffered channel). That works OK for notify, because we know that when the watch initiator peels the event off the channel and queries, it will get the data that caused any of the events that we discarded.

Here, discarding events means missing changes.

Perhaps we need to to implement a collect method to compress events while channel writes are blocked waiting for the other side to read, similar to the state-based strings watcher.

core/cache/watcher.go
```diff
@@ -144,3 +145,104 @@ func (w *ConfigWatcher) configChanged(topic string, value interface{}) {
...
+	w.mu.Lock()
+
+	if w.closed {
+		w.mu.Unlock()
+		return
+	}
+
+	select {
+	case w.changes <- values:
+	default:
```
manadart: Nice.

hmlanigan:
$$merge$$

PR #9920, opened 2019-03-20
Allow cloud changes to be watched for openstack

wallyworld:
## Description of change

The environ tracker for each model worker watches for changes to the model's cloud and updates the environ instance with the new cloud spec. Currently, only openstack has the capability to react to cloud changes.

There are 2 commits.

1. boiler plate

The changes are largely mechanical and add to the existing common cloud spec api facade plugin, using the existing coding conventions. The key components are:
- state watcher
- apiserver, api facade implementation
- environ tracker worker

2. openstack provider

Add support for reacting to cloud changes. Really just a simple re-org of the Open() logic.

## QA steps

bootstrap lxd
run juju update-cloud
observe via logging that the SetCloudSpec api is called on the lxd provider
todo: test on openstack

## Bug reference

https://bugs.launchpad.net/juju/+bug/1819456

### DIFF
--- a/api/common/cloudspec/cloudspec.go
+++ b/api/common/cloudspec/cloudspec.go
@@ -5,11 +5,13 @@ package cloudspec
 
 import (
 	"github.com/juju/errors"
-	names "gopkg.in/juju/names.v2"
+	"gopkg.in/juju/names.v2"
 
 	"github.com/juju/juju/api/base"
+	apiwatcher "github.com/juju/juju/api/watcher"
 	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/cloud"
+	"github.com/juju/juju/core/watcher"
 	"github.com/juju/juju/environs"
 )
 
@@ -24,6 +26,25 @@ type CloudSpecAPI struct {
 // FacadeCaller.
 func NewCloudSpecAPI(facade base.FacadeCaller, modelTag names.ModelTag) *CloudSpecAPI {
 	return &CloudSpecAPI{facade, modelTag}
+}
+
+// WatchCloudSpecChanges returns a NotifyWatcher waiting for the
+// model's cloud to change.
+func (api *CloudSpecAPI) WatchCloudSpecChanges() (watcher.NotifyWatcher, error) {
+	var results params.NotifyWatchResults
+	args := params.Entities{Entities: []params.Entity{{api.modelTag.String()}}}
+	err := api.facade.FacadeCall("WatchCloudSpecsChanges", args, &results)
+	if err != nil {
+		return nil, err
+	}
+	if n := len(results.Results); n != 1 {
+		return nil, errors.Errorf("expected 1 result, got %d", n)
+	}
+	result := results.Results[0]
+	if result.Error != nil {
+		return nil, errors.Annotate(result.Error, "API request failed")
+	}
+	return apiwatcher.NewNotifyWatcher(api.facade.RawAPICaller(), result), nil
 }
 
 // CloudSpec returns the cloud specification for the model associated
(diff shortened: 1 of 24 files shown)


environs/interface.go
```diff
@@ -292,6 +298,7 @@ type Bootstraper interface {
 type Configer interface {
 	ConfigGetter
 	ConfigSetter
+	CloudSpecSetter
```
SimonRichardson: Does this need to be added here? Could you not check if the environ implements the `CloudSpecSetter` and then call the method. That way we don't need the empty `SetCloudSpec` in every provider?

wallyworld:
[openstack cloud update issue](https://trello.com/c/S1zE5uDK/660-openstack-cloud-update-issue)

caas/kubernetes/provider/k8s.go
```diff
@@ -180,6 +180,12 @@ func (k *kubernetesClient) SetConfig(cfg *config.Config) error {
 	return nil
 }
 
+// SetCloudSpec is specified in the environs.Environ interface.
+func (k *kubernetesClient) SetCloudSpec(spec environs.CloudSpec) error {
```
jameinel: given that this seems optional, why implement a no-op version?
if we *didn't* implement it, couldn't we just then check if it was supported and if not, do the no-op in the caller?
That's what we do for the NetworkingEnviron code

worker/environ/environ.go
```diff
@@ -101,14 +113,30 @@ func (t *Tracker) loop() error {
...
+			logger.Debugf("reloading environ config")
+			modelConfig, err := t.config.Observer.ModelConfig()
+			if err != nil {
+				return errors.Annotate(err, "cannot read environ config")
+			}
+			if err = t.environ.SetConfig(modelConfig); err != nil {
+				return errors.Annotate(err, "cannot update environ config")
+			}
+		case _, ok := <-cloudWatcher.Changes():
+			if !ok {
```
jameinel: here you could do:
cloudSpecChanger, ok := t.environ.(CloudSpecSetter)
if !ok {
 logger.Warningf("cloud type %X doesn't support dynamic changing of cloud spec", t.environ)
 continue
}

That seems better than having it be an unrealized no op
I might slightly prefer to have all the Environs implement the interface, but have *them* log a message that it isn't actually supported, as it makes it easier to grep for what clouds providers need updating.

api/common/cloudspec/cloudspec.go
```diff
@@ -26,6 +28,25 @@ func NewCloudSpecAPI(facade base.FacadeCaller, modelTag names.ModelTag) *CloudSp
 	return &CloudSpecAPI{facade, modelTag}
 }
 
+// WatchCloudSpecChanges returns a NotifyWatcher waiting for the
+// model's cloud to change.
+func (api *CloudSpecAPI) WatchCloudSpecChanges() (watcher.NotifyWatcher, error) {
+	var results params.NotifyWatchResults
+	args := params.Entities{Entities: []params.Entity{{api.modelTag.String()}}}
+	err := api.facade.FacadeCall("WatchCloudSpecsChanges", args, &results)
```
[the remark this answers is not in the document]
wallyworld: Our naming convention for all api methods is to use singular on the api side and plural on the apiserver side (to reflect the bulk call aspect).

caas/kubernetes/provider/k8s.go
wallyworld: yeah. i should have done that and will fix

worker/environ/environ.go
wallyworld: yup, fixed

wallyworld:
$$merge$$

PR #9935, opened 2019-03-21
Change SetSSHHostKeys to use buildTxn.

jameinel:
## Description of change

Rather than having it always Insert+Update, now it reads to see which
one it needs to do. The old code worked with client-side transactions,
but doesn't feel great. So switch to something that will work with
server-side transactions, and still feels a bit better. (No need to
create Txns that don't do anything.)

## QA steps

The change should cause us to not write to the database on every startup of a machine agent. The easiest way to check this would be to look at the txns log, and restart various machine agents to see that there aren't transactions trying to update the host keys.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1821147

### DIFF
--- a/state/sshhostkeys.go
+++ b/state/sshhostkeys.go
@@ -4,6 +4,8 @@
 package state
 
 import (
+	"sort"
+
 	"github.com/juju/errors"
 	jujutxn "github.com/juju/txn"
 	"gopkg.in/juju/names.v2"
@@ -52,6 +54,10 @@ func keysEqual(a, b []string) bool {
 	if len(a) != len(b) {
 		return false
 	}
+	a = a[:]
+	b = b[:]
+	sort.Strings(a)
+	sort.Strings(b)
 	for i := range a {
 		if a[i] != b[i] {
 			return false
--- a/state/sshhostkeys.go
+++ b/state/sshhostkeys.go
@@ -5,6 +5,7 @@ package state
 
 import (
 	"github.com/juju/errors"
+	jujutxn "github.com/juju/txn"
 	"gopkg.in/juju/names.v2"
 	"gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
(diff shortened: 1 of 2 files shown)


state/sshhostkeys.go
```diff
@@ -45,26 +46,57 @@ func (st *State) GetSSHHostKeys(tag names.MachineTag) (SSHHostKeys, error) {
 	return SSHHostKeys(doc.Keys), nil
 }
 
+// keysEqual checks if the ssh host keys are the same between two sets.
+// we shouldn't care about the order of the keys.
```
SimonRichardson: Out of interest, why doesn't order matter in this case?

state/sshhostkeys.go
```diff
@@ -45,26 +46,57 @@ func (st *State) GetSSHHostKeys(tag names.MachineTag) (SSHHostKeys, error) {
...
+func keysEqual(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
```
achilleasa: Based on the function docs shouldn't you be sorting the lists first or using a map to implement exists-in-set semantics?

state/sshhostkeys.go
achilleasa: The way I read this comment is that since they are sets element order should not matter for equality checks. Isn't this the case?

jameinel: I don't think SSH host keys care whether it is (A, B) or (B, A) they just care what values are there. That said, I couldn't be bothered, so I just went with the easy way of comparing them in order.
I guess we could just do a sort.Strings() before the comparison.

I know we had problems like that with IP addresses, because we were putting them in a map and then iterating a map. Which meant we kept getting InstancePoller thinking they had changed. I'll just put in the sort and be done with it.

state/sshhostkeys.go
jameinel: we need them to be bi-directionally equal, I was just being lazy by not sorting. patch is updated.

jameinel:
$$merge$$

PR #9943, opened 2019-03-22
Fixes lp#1800056: Remove redundant --credential option - it was never operational.

anastasiamac:
## Description of change

'update-credential' command advertises a --credential option but it never catered for it. The command takes credential name from a positional argument and due to the logic that handles positional arguments, this command never worked with the option. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1800056


anastasiamac:
[remove redundant --credential option](https://trello.com/c/KdrKrcod/668-remove-redundant-credential-option)

anastasiamac:
$$merge$$

PR #9946, opened 2019-03-22
Fix MemoryProfile!='low' to properly handle WiredTigerCacheSize.

jameinel:
## Description of change

Using 1GB or 0.25GB memory are both considered "low" memory. We want
Mongo to decide how much memory to use rather than passing it in if you
use default memory profile.

When the code was refactored to support snaps, it accidentally broke how we handled low mem vs default mem with 3.4+.

## QA steps

Test suite passes, ``juju bootstrap`` should default to configure mongo without '--wiredTigerCacheSizeGB', and bootstrap with mongo-memory-profile=low should set it to 0.25GB on new mongo.

## Documentation changes

None.

## Bug reference

None.

### DIFF
--- a/mongo/mongo.go
+++ b/mongo/mongo.go
@@ -238,13 +238,20 @@ var (
 		Patch:         "",
 		StorageEngine: WiredTiger,
 	}
-	// Mongo406wt represents 'mongodb-server-core' at version 4.0.6 with WiredTiger
+	// Mongo406wt represents snap 'juju-db' at version 4.0.x with WiredTiger
+	// XXX: (jam) This is the only Mongo declaration that declares a Point
+	//  release. It doesn't seem the right place to do so. We currently have to,
+	//  because the K8s pod spec requires it but
+	//  a) It doesn't match the current snap version
+	//  b) It prevents our normal "update the package to get an updated mongo without
+	//     rebuilding Juju".
+	//  c) It will get *real* messy if we support a range of 4.0.* releases.
 	Mongo406wt = Version{Major: 4,
 		Minor:         0,
 		Point:         6,
 		StorageEngine: WiredTiger,
 	}
-	// MongoUpgrade represents a sepacial case where an upgrade is in
+	// MongoUpgrade represents a special case where an upgrade is in
 	// progress.
 	MongoUpgrade = Version{Major: 0,
 		Minor:         0,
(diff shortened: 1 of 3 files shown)


mongo/mongo.go
```diff
@@ -238,13 +238,20 @@ var (
 		Patch:         "",
 		StorageEngine: WiredTiger,
 	}
-	// Mongo406wt represents 'mongodb-server-core' at version 4.0.6 with WiredTiger
+	// Mongo406wt represents snap 'juju-db' at version 4.0.x with WiredTiger
```
wallyworld: This isn't used for the snap, just the docker image used for k8s.

mongo/mongo.go
```diff
@@ -238,13 +238,20 @@ var (
...
+	// XXX: (jam) This is the only Mongo declaration that declares a Point
+	//  release. It doesn't seem the right place to do so. We currently have to,
+	//  because the K8s pod spec requires it but
+	//  a) It doesn't match the current snap version
+	//  b) It prevents our normal "update the package to get an updated mongo without
```
wallyworld: That's by design so we get fully repeatable deploys.
But maybe that's too restrictive. Maybe we just want to tag upstream mongo images as juju-db:4.X and feel free to copy a point release across when needed.

mongo/service.go
```diff
@@ -459,11 +459,15 @@ func generateConfig(mongoPath string, dataDir string, statePort int, oplogSizeMB
 		//BindIP:                "127.0.0.1", // TODO(tsm): use machine's actual IP address via dialInfo
 	}
 
-	if usingMongo34orAbove && useLowMemory && usingWiredTiger {
-		mongoArgs.WiredTigerCacheSizeGB = Mongo34LowCacheSize
-	} else if usingWiredTiger {
-		mongoArgs.WiredTigerCacheSizeGB = LowCacheSize
-	} else {
+	if useLowMemory && usingWiredTiger {
```
wallyworld: I have a drive by for this in a current PR

mongo/service.go
```diff
@@ -459,11 +459,15 @@ func generateConfig(mongoPath string, dataDir string, statePort int, oplogSizeMB
...
+		if usingMongo34orAbove {
+			// Mongo 3.4 introduced the ability to have fractional GB cache size
```
wallyworld: I copied these extra comments to my PR

mongo/mongo.go
```diff
@@ -238,13 +238,20 @@ var (
...
+	//     rebuilding Juju".
+	//  c) It will get *real* messy if we support a range of 4.0.* releases.
```
wallyworld: I've updated my PR to omit the point version.
It will now just pull juju-db:4.0 (or whatever)
We can tag whatever upstream point version we want as 4.X

jameinel:
replaced my Ian's fix

PR #9951, opened 2019-03-26
vsphere: Fix root disk resizing

babbageclunk:
## Description of change

Creating a new VM with a non-default disk size would hang because after the extend disk task is successfully started, the API call to get the task status fails with a permission error (you can see the error in the commit message). This means the task waiter never returns. I haven't been able to work out the cause of the error requesting the task status - it seems like a bug in the VSphere API. The extend disk task itself succeeds, we just can't request the status.

To work around this, poll the size of the VM's disk to detect when it's been extended and return then.
If we timeout waiting for the disk to be the right size we return a ZoneIndependentError to make sure the provisioner doesn't retry in different resource groups.

## QA steps
* Bootstrap a vsphere controller 
* Deploy an application using a root-disk constraint `--constraints="root-disk=20G"`
* The machine should be created with a 20Gb root disk

## Documentation changes
None

## Bug reference
https://bugs.launchpad.net/juju/+bug/1790183

### DIFF
--- a/provider/vsphere/environ_availzones.go
+++ b/provider/vsphere/environ_availzones.go
@@ -73,6 +73,7 @@ func (env *sessionEnviron) AvailabilityZones(ctx context.ProviderCallContext) ([
 		for _, cr := range computeResources {
 			if cr.Summary.GetComputeResourceSummary().EffectiveCpu == 0 {
 				logger.Debugf("skipping empty compute resource %q", cr.Name)
+				continue
 			}
 			pools, err := env.client.ResourcePools(env.ctx, cr.Name+"/...")
 			if err != nil {
--- a/provider/vsphere/internal/vsphereclient/client.go
+++ b/provider/vsphere/internal/vsphereclient/client.go
@@ -8,7 +8,9 @@ import (
 	"net/url"
 	"path"
 	"strings"
+	"time"
 
+	"github.com/juju/clock"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	"github.com/vmware/govmomi"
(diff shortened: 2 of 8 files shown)


babbageclunk:
[vsphere: hang when resizing disk](https://trello.com/c/ORiG9rcj/658-vsphere-hang-when-resizing-disk)

provider/vsphere/environ_availzones.go
```diff
@@ -73,6 +73,7 @@ func (env *sessionEnviron) AvailabilityZones(ctx context.ProviderCallContext) ([
 		for _, cr := range computeResources {
 			if cr.Summary.GetComputeResourceSummary().EffectiveCpu == 0 {
 				logger.Debugf("skipping empty compute resource %q", cr.Name)
+				continue
```
wallyworld: oops, good catch

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -438,8 +475,31 @@ func (c *Client) cloneVM(
...
+	for _, dev := range mo.Config.Hardware.Device {
+		dev, ok := dev.(*types.VirtualDisk)
+		if !ok {
+			continue
+		}
+		backing, ok := dev.Backing.(types.BaseVirtualDeviceFileBackingInfo)
+		if !ok {
+			continue
+		}
+		return dev, backing, nil
```
wallyworld: so there can't be more than one?
we just return the first one?

babbageclunk: We create this machine from a template, I think there can only be one.

babbageclunk:
$$merge$$

PR #9952, opened 2019-03-26
Fix serialization of errors in status fields.

howbazaar:
Found during debugging of deploying a bundle problem in bug listed below.

The missing status for machine modifications was getting an error added to the DetailedStatus response. This should never have had 'error' as the type.

## QA steps

Reproduction steps listed in bug

## Documentation changes

None

## Bug reference

https://bugs.launchpad.net/juju/+bug/1821418

PR #9957, opened 2019-03-26
Modification Upgrade Step

SimonRichardson:
## Description of change

The following adds a upgrade step to add missing documents for
the modification status. We jump through a few hoops here, because
we don't want to remove existing modification statuses, we only
want to add a status document, if and when a document doesn't exist.
We can work that out by checking if the machine is Alive then also
verifying that the document doesn't already exist before inserting
it.

## QA steps

Firstly you need to check out the 2.4 branch
```
export JUJU_MAKE_GODEPS=true; make install
juju bootstrap lxd test --debug
juju deploy ubuntu
juju export-bundle > bundle24.yaml
juju deploy --dry-run ./bundle24.yaml # this is to verify that expected behaviour
```

Then move to this branch
```
make install
juju upgrade-model -m controller; sleep 300
juju deploy --dry-run ./bundle24.yaml # this previously broke
```

Also it's worth verifying 2.5 `export-bundle` as well
```
juju export-bundle > bundle25.yaml
juju deploy --dry-run ./bundle25.yaml # this previously broke
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1821418

### DIFF
--- a/state/upgrades.go
+++ b/state/upgrades.go
@@ -1916,7 +1916,7 @@ func EnsureDefaultModificationStatus(pool *StatePool) error {
 
 		// We only need to migrate machines that don't have a modification
 		// status document. So we need to first check if there is one, before
-		// creating an txn.Op for the missing document.
+		// creating a txn.Op for the missing document.
 		var status statusDoc
 		err := statusCol.Find(bson.D{{"_id", key}}).Select(bson.D{{"_id", 1}}).One(&status)
 		if err == mgo.ErrNotFound {
--- a/state/upgrades.go
+++ b/state/upgrades.go
@@ -23,6 +23,7 @@ import (
 	"github.com/juju/juju/cloud"
 	"github.com/juju/juju/controller"
 	corelease "github.com/juju/juju/core/lease"
+	"github.com/juju/juju/core/life"
 	"github.com/juju/juju/core/status"
 	"github.com/juju/juju/environs/config"
 	"github.com/juju/juju/mongo/utils"
@@ -1881,6 +1882,57 @@ func UpdateInheritedControllerConfig(pool *StatePool) error {
 	if len(ops) > 0 {
 		err = errors.Trace(st.runRawTransaction(ops))
 		return err
+	}
+	return nil
+}
+
+// EnsureDefaultModificationStatus ensures that there is a modification status
+// document for every machine in the statuses.
+func EnsureDefaultModificationStatus(pool *StatePool) error {
+	st := pool.SystemState()
+	db := st.db()
+
+	machineCol, machineCloser := db.GetRawCollection(machinesC)
+	defer machineCloser()
+	// Find all machines that are alive.
+	machineIter := machineCol.Find(bson.D{{"life", life.Alive}}).Iter()
+	defer machineIter.Close()
+
+	statusCol, statusCloser := db.GetRawCollection(statusesC)
+	defer statusCloser()
+
+	var ops []txn.Op
+	var machine machineDoc
+	for machineIter.Next(&machine) {
+		key := machineGlobalModificationKey(machine.Id)
+
+		rawDoc := statusDoc{
+			ModelUUID:  machine.ModelUUID,
+			Status:     status.Idle,
+			StatusInfo: "",
+			StatusData: make(map[string]interface{}),
+			Updated:    st.clock().Now().UnixNano(),
+		}
+
+		// We only need to migrate machines that don't have a modification
+		// status document. So we need to first check if there is one, before
+		// creating an txn.Op for the missing document.
+		var status statusDoc
+		err := statusCol.Find(bson.D{{"_id", key}}).Select(bson.D{{"_id", 1}}).One(&status)
+		if err == mgo.ErrNotFound {
+			ops = append(ops, txn.Op{
+				C:      statusesC,
+				Id:     key,
+				Assert: txn.DocMissing,
+				Insert: rawDoc,
+			})
+		}
+	}
+	if err := machineIter.Close(); err != nil {
+		return errors.Trace(err)
+	}
+	if len(ops) > 0 {
+		return errors.Trace(st.runRawTransaction(ops))
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


SimonRichardson:
https://trello.com/c/ax16Unzf/1230-setmodificationstatus-upgrade-missing-step

state/upgrades.go
```diff
@@ -1884,3 +1885,54 @@ func UpdateInheritedControllerConfig(pool *StatePool) error {
...
+			ModelUUID:  machine.ModelUUID,
+			Status:     status.Idle,
+			StatusInfo: "",
+			StatusData: make(map[string]interface{}),
+			Updated:    st.clock().Now().UnixNano(),
+		}
+
+		// We only need to migrate machines that don't have a modification
+		// status document. So we need to first check if there is one, before
+		// creating an txn.Op for the missing document.
```
SimonRichardson:
```suggestion
		// creating a txn.Op for the missing document.
```

state/upgrades.go
```diff
@@ -1884,3 +1885,54 @@ func UpdateInheritedControllerConfig(pool *StatePool) error {
...
+func EnsureDefaultModificationStatus(pool *StatePool) error {
+	st := pool.SystemState()
+	db := st.db()
+
+	machineCol, machineCloser := db.GetRawCollection(machinesC)
+	defer machineCloser()
+	// Find all machines that are alive.
+	machineIter := machineCol.Find(bson.D{{"life", life.Alive}}).Iter()
+	defer machineIter.Close()
+
```
jameinel: you really should check the returned error on Iter.Close. it is the only thing that tells u if the db went away or the query was invalid.

state/upgrades.go
```diff
@@ -1884,3 +1885,54 @@ func UpdateInheritedControllerConfig(pool *StatePool) error {
...
+			StatusData: make(map[string]interface{}),
+			Updated:    st.clock().Now().UnixNano(),
+		}
+
+		// We only need to migrate machines that don't have a modification
+		// status document. So we need to first check if there is one, before
+		// creating a txn.Op for the missing document.
+		var status statusDoc
+		err := statusCol.Find(bson.D{{"_id", key}}).Select(bson.D{{"_id", 1}}).One(&status)
+		if err == mgo.ErrNotFound {
```
jameinel: his would probably be easier to read if you move this query earlier and just have

if err == nil {
  continue
} else if err != mgo.ErrNotFound {
  return errors.Trace(err)
}

then you don't have to nest the logic and you don't have to create rawDoc if you aren't going to use it.

state/upgrades.go
```diff
@@ -1884,3 +1885,54 @@ func UpdateInheritedControllerConfig(pool *StatePool) error {
...
+// EnsureDefaultModificationStatus ensures that there is a modification status
+// document for every machine in the statuses.
+func EnsureDefaultModificationStatus(pool *StatePool) error {
+	st := pool.SystemState()
+	db := st.db()
+
+	machineCol, machineCloser := db.GetRawCollection(machinesC)
+	defer machineCloser()
+	// Find all machines that are alive.
+	machineIter := machineCol.Find(bson.D{{"life", life.Alive}}).Iter()
```
howbazaar: I'm not sure what the assertions are like for the removal of the status docs for the machine when it gets removed, so I'm not clear whether we should be adding it for all machines or just the alive ones.

state/upgrades.go
```diff
@@ -1884,3 +1885,54 @@ func UpdateInheritedControllerConfig(pool *StatePool) error {
...
+	defer statusCloser()
+
+	var ops []txn.Op
+	var machine machineDoc
+	for machineIter.Next(&machine) {
+		key := machineGlobalModificationKey(machine.Id)
+
+		rawDoc := statusDoc{
+			ModelUUID:  machine.ModelUUID,
+			Status:     status.Idle,
```
howbazaar: I'd skip the StatusInfo and StatusData for the rawDoc. Empty values are fine.

state/upgrades.go
howbazaar: The removeStatusOps doesn't have an assert, so it just tries to remove it. However it is more work both here and in the testing to make sure they don't get added for non-alive machines, so just take the approach of "add one for all machines".

state/upgrades.go
```diff
@@ -1884,3 +1885,54 @@ func UpdateInheritedControllerConfig(pool *StatePool) error {
...
+	machineIter := machineCol.Find(bson.D{{"life", life.Alive}}).Iter()
+	defer machineIter.Close()
+
+	statusCol, statusCloser := db.GetRawCollection(statusesC)
+	defer statusCloser()
+
+	var ops []txn.Op
+	var machine machineDoc
+	for machineIter.Next(&machine) {
+		key := machineGlobalModificationKey(machine.Id)
```
howbazaar: This is the wrong key as it doesn't include the model uuid. This is necessary because we are using raw collections.

Your QA procedures wouldn't have caught this because of the fix I landed yesterday in #9952. You'd need to look at the output of YAML status to ensure that the docs aren't missing.

state/upgrades_test.go
```diff
@@ -2926,6 +2931,43 @@ func (s *upgradesSuite) TestUpdateInheritedControllerConfig(c *gc.C) {
...
+		expectUpgradedData{coll, expected},
+	)
+}
+
+func (s *upgradesSuite) makeMachine(c *gc.C, life life.Value) {
+	coll, closer := s.state.db().GetRawCollection(machinesC)
+	defer closer()
+
+	err := coll.Insert(bson.M{
+		"machineid": "0",
```
howbazaar: This doesn't insert a valid machine, there is no model-uuid specified, and the resulting test above is showing that the "_id" and model-uuid are not set correctly.

state/upgrades.go
howbazaar: You are also checking against life.Alive, which is a string, which won't match state.Life which is a uint8. This wasn't caught because the test was inserting what you were looking for rather than inserting a machine doc.

PR #9959, opened 2019-03-26
upgrade step to add device constraints

wallyworld:
## Description of change

An upgrade step to add empty device constraint docs for applications created prior to 2.5

## QA steps

bootstrap juju 2.4
add an application
upgrade to 2.5
use mongo client to confirm deviceconstraints collection has a doc for each application

### DIFF
--- a/state/upgrades.go
+++ b/state/upgrades.go
@@ -1939,3 +1939,52 @@ func EnsureDefaultModificationStatus(pool *StatePool) error {
 	}
 	return nil
 }
+
+// EnsureApplicationDeviceConstraints ensures that there is a device
+// constraints document for every application.
+func EnsureApplicationDeviceConstraints(pool *StatePool) error {
+	st := pool.SystemState()
+	db := st.db()
+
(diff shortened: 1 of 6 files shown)


state/upgrades.go
```diff
@@ -1939,3 +1939,55 @@ func EnsureDefaultModificationStatus(pool *StatePool) error {
...
+		// creating a txn.Op for the missing document.
+		var doc statusDoc
+		err := constraintsCol.Find(bson.D{{"_id", key}}).Select(bson.D{{"_id", 1}}).One(&doc)
+		if err == nil {
+			continue
+		} else if err != mgo.ErrNotFound {
+			return errors.Trace(err)
+		}
+
+		rawDoc := deviceConstraintsDoc{
```
howbazaar: What happens if you just insert an empty deviceConstraintsDoc? And inline would be fine. I'm fairly sure it'll work.

wallyworld: ok

wallyworld:
$$merge$$

PR #9967, opened 2019-03-27
WatchApplications API AssertStop

SimonRichardson:
## Description of change

The following ensures that we have sent out all the notifies for the
watcher before finishing the test. This way we can be sure that the
first event is consumed by the facade and that we don't pass it
towards.


apiserver/facades/agent/instancemutater/instancemutater.go
```diff
@@ -223,6 +223,7 @@ func (api *InstanceMutaterAPI) watchOneEntityApplication(canAccess common.AuthFu
 		return result, err
 	}
 	watch := machine.WatchApplicationLXDProfiles()
+	// Consume the initial event and forward it to the result.
```
hmlanigan: Thought we weren't forwarding it?

SimonRichardson:
$$merge$$

PR #9969, opened 2019-03-28
add functions to model cache for instance mutater apiserver functionaltiy 

hmlanigan:
Based on PR9966 for change to cache.Machine constructer.

## Description of change

Added charm.LXDProfile(), unit.Application(), machine.Units(), and application.CharmURL() to model cache.

## QA steps

No change  to current juju usage.

### DIFF
--- a/core/cache/application.go
+++ b/core/cache/application.go
@@ -33,6 +33,11 @@ type Application struct {
 	hashCache  *hashCache
 }
 
+// CharmURL returns the charm url string for this application.
+func (a *Application) CharmURL() string {
+	return a.details.CharmURL
+}
+
 // Config returns a copy of the current application config.
 func (a *Application) Config() map[string]interface{} {
 	a.mu.Lock()
--- a/core/cache/charm.go
+++ b/core/cache/charm.go
@@ -7,6 +7,8 @@ import (
 	"sync"
 
 	"github.com/juju/pubsub"
+
+	"github.com/juju/juju/core/lxdprofile"
 )
 
 func newCharm(metrics *ControllerGauges, hub *pubsub.SimpleHub) *Charm {
@@ -25,6 +27,11 @@ type Charm struct {
 	mu      sync.Mutex
 
 	details CharmChange
+}
+
+// LXDProfile returns the lxd profile of this charm.
+func (c *Charm) LXDProfile() lxdprofile.Profile {
+	return c.details.LXDProfile
 }
 
 func (c *Charm) setDetails(details CharmChange) {
--- a/core/cache/export_test.go
+++ b/core/cache/export_test.go
@@ -18,3 +18,13 @@ func (m *Model) SetDetails(details ModelChange) {
 func (a *Application) SetDetails(details ApplicationChange) {
 	a.setDetails(details)
 }
+
+// Expose Update* for testing.
+
+func (m *Model) UpdateMachine(details MachineChange) {
+	m.updateMachine(details)
+}
+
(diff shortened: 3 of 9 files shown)


core/cache/watcher.go
```diff
@@ -146,6 +148,188 @@ func (w *ConfigWatcher) configChanged(topic string, value interface{}) {
...
+	if len(values) != 2 {
+		logger.Errorf("programming error, 2 values not provided")
+		return
+	}
+	appName, chURL := values[0], values[1]
+	info, ok := w.applications[appName]
+	if ok {
+		ch, err := w.getCharm(chURL)
+		if err != nil {
+
```
SimonRichardson: What should happen here? I guess return with the error?

core/cache/watcher.go
```diff
@@ -146,6 +148,188 @@ func (w *ConfigWatcher) configChanged(topic string, value interface{}) {
...
+func (w *MachineAppLXDProfileWatcher) addUnit(unit *Unit) bool {
+	_, ok := w.applications[unit.details.Application]
+	if !ok {
+		info := appInfo{
+			charmURL: unit.details.CharmURL,
+			units:    set.NewStrings(unit.details.Name),
+		}
+		ch, err := w.getCharm(unit.details.CharmURL)
+		if err != nil {
+
```
SimonRichardson: Missing handling error?

core/cache/watcher.go
```diff
@@ -146,6 +148,188 @@ func (w *ConfigWatcher) configChanged(topic string, value interface{}) {
...
+	} else {
+		w.applications[unit.details.Application].units.Add(unit.details.Name)
+	}
+	if w.applications[unit.details.Application].charmProfile != nil {
+		return true
+	}
+	return false
+}
+
+func (w *MachineAppLXDProfileWatcher) removeUnit(names []string) bool {
```
SimonRichardson: I'm not sure I like this, I think we should either do `[2]string`, `name struct{UnitName, AppName string}` or pass two arguments.

core/cache/watcher.go
```diff
@@ -146,6 +148,188 @@ func (w *ConfigWatcher) configChanged(topic string, value interface{}) {
...
+			w.notify()
+		}
+	}(&notify)
+
+	names, okString := value.([]string)
+	unit, okUnit := value.(*Unit)
+	switch {
+	case okString:
+		logger.Tracef("Stop watching %q", names)
+		notify = w.removeUnit(names)
```
SimonRichardson: I think this should be... and validate the logic here, as removeUnit becomes very cryptic further down in the PR.

```
notify = w.removeUnit(names[0], names[1])
```

core/cache/watcher.go
hmlanigan: resolved in pr 9966 lxdprofilewatcher.go

core/cache/watcher.go
hmlanigan: will be resolved in pr 9966 lxdprofilewatcher.go

core/cache/watcher.go
hmlanigan: passing two args, will be resolved in pr 9966 lxdprofilewatcher.go

core/cache/watcher.go
hmlanigan: will be resolved in pr 9966 lxdprofilewatcher.go

hmlanigan:
squashed two commits before:
$$merge$$

PR #9974, opened 2019-03-28
Adding instance mutater to machine agent

SimonRichardson:
## Description of change

The following adds the instance mutater to the machine agent, along
with additional broker manifold changes so that the instance mutater 
can then write the profiles to the container.

## QA steps

```
export JUJU_DEV_FEATURE_FLAGS=instance-mutater
juju bootstrap lxd test
juju deploy ./testcharms/charm-repo/quantal/lxd-profile --to=lxd
```


cmd/jujud/agent/machine/manifolds.go
```diff
@@ -928,7 +936,24 @@ func IAASManifolds(config ManifoldsConfig) dependency.Manifolds {
 			Clock:                        config.Clock,
 			NewCredentialValidatorFacade: common.NewCredentialInvalidatorFacade,
 		}))),
-	})
+	}
+
+	if utilsfeatureflag.Enabled(feature.InstanceMutater) {
```
hmlanigan: Does this worker get started on a controller?

apiserver/facades/agent/instancemutater/instancemutater.go
```diff
@@ -79,14 +79,16 @@ func NewInstanceMutaterAPI(st InstanceMutaterState,
...
 	return &InstanceMutaterAPI{
-		ModelMachinesWatcher: common.NewModelMachinesWatcher(st, resources, authorizer),
-		UnitsWatcher:         common.NewUnitsWatcher(st, resources, getAuthFunc),
-		LifeGetter:           common.NewLifeGetter(st, getAuthFunc),
-		st:                   st,
-		model:                model,
-		resources:            resources,
-		authorizer:           authorizer,
-		getAuthFunc:          getAuthFunc,
+		ModelMachinesWatcher: common.NewModelMachinesWatcherWithAuthFunc(st, resources, func() bool {
```
howbazaar: My problem with this is that it looks like a general machine could ask to watch all the machines in a model. Surely the instance mutater is only interested in its own machines?

cmd/jujud/agent/machine/manifolds.go
howbazaar: Why are we putting it behind a feature flag here? I thought this work replaced what we had in 2.5?

cmd/jujud/agent/machine/manifolds.go
```diff
@@ -928,7 +936,24 @@ func IAASManifolds(config ManifoldsConfig) dependency.Manifolds {
...
+		manifolds[environTrackerName] = ifNotMigrating(environ.Manifold(environ.ManifoldConfig{
+			APICallerName:  apiCallerName,
+			NewEnvironFunc: config.NewEnvironFunc,
```
howbazaar: Only the controller has access to the environment normally. What is this for?

cmd/jujud/agent/machine/manifolds.go
```diff
@@ -954,6 +979,14 @@ func clockManifold(clock clock.Clock) dependency.Manifold {
 	}
 }
 
+// ifResponsible wraps a manifold such that it only runs if the
+// responsibility flag is set.
+var ifResponsible = engine.Housing{
```
howbazaar: What is the purpose of this? For workers where we need one per model, they don't go in the machine manifolds, but in `model/manifolds.go`.

cmd/jujud/agent/machine/manifolds.go
SimonRichardson: @howbazaar yeah, but we'll rip out the feature flag once all done, so we don't ever end up like previously - I thought that was the plan we took away from the sprint?

cmd/jujud/agent/machine/manifolds.go
SimonRichardson: We need access to the LXD broker, this way allows us to get that, but is why I wanted feedback, as I'm not sure it's generally the best approach.

cmd/jujud/agent/machine/manifolds.go
SimonRichardson: Right, I didn't know what `ifResponsible` was and if I needed it. I think there is a bit of knowledge gap around these decorators...

apiserver/facades/agent/instancemutater/instancemutater.go
SimonRichardson: @howbazaar can you add a bit more clarification on this... I think I'm missing something here.

cmd/jujud/agent/machine/manifolds.go
howbazaar: Fair call.

cmd/jujud/agent/machine/manifolds.go
howbazaar: I think we need a call to sort this out. Not sure it'll work well in a review situation.

hmlanigan:
I can deploy lxd container with the current version, however: /var/log/juju/machine-3.log:2019-04-10 22:37:09 DEBUG juju.worker.dependency engine.go:583 "instance-mutater" manifold worker stopped: resource permanently unavailable

worker/broker/broker.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, errors.Annotate(err, "retrieving supported container types")
+	}
+	if len(instanceContainers) == 0 && !determined {
+		return nil, errors.Annotate(err, "no container types deterimined")
+	}
+	// we only work on LXD, so check for that.
+	for _, containerType := range instanceContainers {
+		if containerType != instance.LXD {
+			return nil, worker.ErrTerminateAgent
```
jameinel: I don't think you want TerminateAgent. I believe that means "stop jujud" entirely.
It may be that you just want to return nil error which says the "worker is done", no need to restart.

worker/instancemutater/manifold.go
```diff
@@ -53,17 +53,25 @@ func (config ManifoldConfig) Validate() error {
...
 	}
-
+	// If we don't have a LXDProfiler, we should uninstall the worker as quickly
+	// as possible.
+	broker, ok := environ.(environs.LXDProfiler)
+	if !ok {
+		// If we don't have an LXDProfiler broker, there is no need to
+		// run this worker.
+		config.Logger.Debugf("Uninstalling worker because the broker is not a LXDProfiler %T", environ)
+		return nil, dependency.ErrUninstall
```
jameinel: I don't think ErrUninstall is correct, we should confirm. I don't think this is "stop running the worker", I think it is "uninstall jujud from this machine".

worker/instancemutater/manifold.go
```diff
@@ -124,3 +132,117 @@ func EnvironAPIManifold(config EnvironAPIConfig, start EnvironAPIStartFunc) depe
...
+		return nil, errors.Trace(err)
+	}
+	// If we don't have a LXDProfiler, we should uninstall the worker as quickly
+	// as possible.
+	broker, ok := instanceBroker.(environs.LXDProfiler)
+	if !ok {
+		// If we don't have an LXDProfiler broker, there is no need to
+		// run this worker.
+		config.Logger.Debugf("Uninstalling worker because the broker is not a LXDProfiler %T", instanceBroker)
+		return nil, dependency.ErrUninstall
```
jameinel: confirm the error response.

worker/broker/broker.go
```diff
@@ -0,0 +1,163 @@
+// Copyright 2019 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+package broker
```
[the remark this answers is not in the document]
jameinel: or possibly even lxdbroker?

worker/instancemutater/manifold.go
SimonRichardson: This is probably what we want. https://github.com/juju/worker/blob/v1/dependency/interface.go#L126-L129

container/broker/instance-broker.go
```diff
@@ -0,0 +1,151 @@
...
+	broker, err := newBroker(prepareHost(config), config.APICaller, manager, config.AgentConfig)
+	if err != nil {
+		logger.Errorf("failed to create new %s broker", config.ContainerType)
+		return nil, errors.Trace(err)
+	}
+
+	return broker, nil
+}
+
+func prepareHost(config Config) PrepareHostFunc {
```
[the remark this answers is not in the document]
SimonRichardson: This will be done in a follow up PR.

container/broker/instance-broker.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+}
+
+type AvailabilityZoner interface {
+	AvailabilityZone() (string, error)
+}
+
+// ConfigureAvailabilityZone reads the availability zone from the machine and
+// adds the resulting information to the the manager config.
+func ConfigureAvailabilityZone(managerConfig container.ManagerConfig, machineZone AvailabilityZoner) (container.ManagerConfig, error) {
```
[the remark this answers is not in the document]
SimonRichardson: This will be done in a follow up PR.

cmd/jujud/agent/machine.go
```diff
@@ -849,7 +852,7 @@ func (a *MachineAgent) updateSupportedContainers(
 		return errors.Annotatef(err, "cannot load machine %s from state", tag)
 	}
 	if len(result) != 1 {
-		return errors.Annotatef(err, "expected 1 result, got %d", len(result))
+		return errors.Errorf("expected 1 result, got %d", len(result))
```
SimonRichardson: Drive by, this will return nil, if this ever happens, because we guard against `err` above.

container/broker/lxd-broker.go
```diff
@@ -244,3 +245,12 @@ func (broker *lxdBroker) ReplaceOrAddInstanceProfile(instId, oldProfile, newProf
 	}
 	return profileMgr.ReplaceOrAddInstanceProfile(instId, oldProfile, newProfile, put)
 }
+
+// AssignProfiles implements environs.LXDProfiler.
+func (broker *lxdBroker) AssignProfiles(instId string, profilesNames []string, profilePosts []lxdprofile.ProfilePost) ([]string, error) {
+	profileMgr, ok := broker.manager.(container.LXDProfileManager)
+	if !ok {
+		return []string{}, nil
```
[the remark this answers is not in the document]
SimonRichardson: This is being solved in another PR.

api/provisioner/machine.go
```diff
@@ -608,6 +611,29 @@ func (m *Machine) SupportsNoContainers() error {
 	return m.SetSupportedContainers([]instance.ContainerType{}...)
 }
 
+// SupportedContainers implements MachineProvisioner.SupportedContainers.
+func (m *Machine) SupportedContainers() ([]instance.ContainerType, bool, error) {
```
hmlanigan: Unit test?

SimonRichardson: Fixed.

SimonRichardson:
$$merge$$

PR #9976, opened 2019-03-29
Add CLI front end for upgrade-controller k8s

wallyworld:
## Description of change

Implement the CLI front end for upgrade-controller for k8s.
Instead of getting available agent versions using FindTools(), we instead query the docker repo hosting the operator images to see what tagged versions are available.

A fair chunk of the PR is refactoring to extract common logic used by both IAAS and CAAS. Also the existing upgrade tests are cleaned up to remove the patching of the API funcs and instead pass in stubs.

The next PR will wire up the backend changes, ie upgrade worker.

## QA steps

Ensure IAAS upgrade controller and model still works.

### DIFF
--- a/cmd/juju/commands/upgradecontroller.go
+++ b/cmd/juju/commands/upgradecontroller.go
@@ -4,15 +4,27 @@
 package commands
 
 import (
+	"fmt"
+
 	"github.com/juju/cmd"
 	"github.com/juju/errors"
 	"github.com/juju/gnuflag"
 	"github.com/juju/version"
+	"gopkg.in/juju/names.v2"
 
+	"github.com/juju/juju/api/modelconfig"
+	"github.com/juju/juju/cloudconfig/podcfg"
 	jujucmd "github.com/juju/juju/cmd"
-	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
+	"github.com/juju/juju/controller"
+	"github.com/juju/juju/core/model"
+	"github.com/juju/juju/docker"
+	"github.com/juju/juju/environs"
 	"github.com/juju/juju/environs/bootstrap"
+	"github.com/juju/juju/environs/config"
+	"github.com/juju/juju/jujuclient"
+	"github.com/juju/juju/tools"
+	jujuversion "github.com/juju/juju/version"
 )
 
 var usageUpgradeControllerSummary = `
@@ -49,9 +61,10 @@ func newUpgradeControllerCommand(options ...modelcmd.WrapControllerOption) cmd.C
 // upgradeControllerCommand upgrades the controller agents in a juju installation.
 type upgradeControllerCommand struct {
 	modelcmd.ControllerCommandBase
-	upgradeFlags
+	baseUpgradeCommand
 
-	rawArgs []string
+	upgradeJujuAPI upgradeJujuAPI
+	rawArgs        []string
 }
 
 func (c *upgradeControllerCommand) Info() *cmd.Info {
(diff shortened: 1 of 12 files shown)


cmd/juju/commands/upgradecontroller.go
```diff
@@ -64,41 +76,195 @@ func (c *upgradeControllerCommand) Info() *cmd.Info {
...
 	}
+	if details.ModelType == model.CAAS {
+		return c.upgradeCAASController(ctx)
+	}
 	return c.upgradeIAASController(ctx)
 }
 
+func (c *upgradeControllerCommand) upgradeCAASController(ctx *cmd.Context) error {
+	if c.BuildAgent {
+		return errors.New("--build-agent for k8s controller upgrades")
```
ycliuhw: better to raise notsupported error?

cmd/juju/commands/upgradecontroller.go
```diff
@@ -64,41 +76,195 @@ func (c *upgradeControllerCommand) Info() *cmd.Info {
...
+	if err := context.maybeChoosePackagedAgent(); err != nil {
+		ctx.Verbosef("%v", err)
+		return err
+	}
+
+	if err := context.validate(); err != nil {
+		return err
+	}
+	ctx.Verbosef("available agent images:\n%s", formatVersions(context.packagedAgents))
+	fmt.Fprintf(ctx.Stderr, "best version:\n    %v\n", context.chosen)
```
ycliuhw: do we want to print to stderr here?

cmd/juju/commands/upgradecontroller.go
```diff
@@ -64,41 +76,195 @@ func (c *upgradeControllerCommand) Info() *cmd.Info {
...
+	if warnCompat {
+		fmt.Fprintf(ctx.Stderr, "version %s incompatible with this client (%s)\n", context.chosen, jujuversion.Current)
+	}
+	if c.DryRun {
+		fmt.Fprintf(ctx.Stderr, "%s\n", c.upgradeMessage)
```
ycliuhw: here same

tools/list.go
```diff
@@ -86,14 +95,30 @@ func (src List) Newest() (version.Number, List) {
...
+func (src Versions) Newest() (version.Number, Versions) {
+	var result Versions
+	var best version.Number
+	for _, agent := range src {
+		if best.Compare(agent.AgentVersion()) < 0 {
+			// Found new best number; reset result list.
+			best = agent.AgentVersion()
+			result = append(result[:0], agent)
+		} else if agent.AgentVersion() == best {
+			result = append(result, agent)
```
ycliuhw: do we wanna append or overwrite like above here?

cmd/juju/commands/docker.go
```diff
@@ -0,0 +1,87 @@
...
+// The results are used when upgrading Juju to see what's available.
+func listOperatorImages(imagePath string) (tools.Versions, error) {
+	tagsURL := fmt.Sprintf("%s/%s/tags", baseRegistryURL, imagePath)
+	logger.Debugf("operater image tags URL: %v", tagsURL)
+	data, err := httpGet(tagsURL, 30*time.Second)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	type info struct {
```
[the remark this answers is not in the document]
wallyworld: We only need to unmarshall just the name attribute, so there's no need to pull in an entire docker struct just for one field?

cmd/juju/commands/upgradecontroller.go
wallyworld: this is what is done in the upgrade-model command so I wanted to keep it the same here
i am not sure why it was done like that

tools/list.go
wallyworld: this is an exact copy of the tools code and I am reluctant to change such a low level bit of logic which could have unintended consequences

cmd/juju/commands/docker.go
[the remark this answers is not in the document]
wallyworld: all the examples I can find, including a golang library, use the same http get under the hood

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #9980, opened 2019-03-29
Bump timeout for LXD bundle deploys

SimonRichardson:
## Description of change

The following bumps the deployment timeout to ensure we give enough
timeout for complex deploys.

## QA steps

It expects you to setup the following acceptance tests as [follows](https://github.com/juju/juju/tree/develop/acceptancetests#quick-run-using-lxd)

```
mkdir /tmp/test-run
mkdir /tmp/artifacts
export JUJU_HOME=/tmp/test-run
export JUJU_REPOSITORY=$GOPATH/src/github.com/juju/juju/acceptancetests/repository
vim $JUJU_HOME/environments.yaml
```

Local LXD environments.yaml
```yaml
environments:
    lxd:
        type: lxd
        test-mode: true
        default-series: bionic
```

Run the following (substitute the right arg):
```
cd acceptancetests
./assess_deploy_lxd_profile_bundle.py
```

### DIFF
--- a/acceptancetests/jujupy/client.py
+++ b/acceptancetests/jujupy/client.py
@@ -103,6 +103,7 @@ KVM_MACHINE = 'kvm'
 LXC_MACHINE = 'lxc'
 LXD_MACHINE = 'lxd'
 
+_DEFAULT_POLL_TIMEOUT = 2
 _DEFAULT_BUNDLE_TIMEOUT = 3600
 
 log = logging.getLogger("jujupy")
@@ -1439,7 +1440,7 @@ class ModelClient:
                             break
                         status.raise_highest_error(ignore_recoverable=True)
                         reporter.update(states)
-                        time.sleep(1)
+                        time.sleep(_DEFAULT_POLL_TIMEOUT)
                     else:
                         if status is not None:
                             log.error(status.status_text)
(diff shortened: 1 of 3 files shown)


acceptancetests/jujupy/client.py
```diff
@@ -103,6 +103,7 @@
 LXC_MACHINE = 'lxc'
 LXD_MACHINE = 'lxd'
 
+_DEFAULT_POLL_TIMEOUT = 2
```
achilleasa: Can we change this to 5 sec instead? It won't make a big difference in the grand scheme of things but seems a rounder number...

SimonRichardson:
$$merge$$

SimonRichardson:
I've noticed that is indeed a lot quicker for me locally, let's see if CI also becomes quicker because of the less number of status calls, so things are less likely to fail. But this does raise a concern about if we should do this long term and just make status work in a better way, or alternatively use different commands like `show-controllers` (etc) to be more light weight?

achilleasa:
I wonder if we could implement a watch (e.g `juju status -w`) type of option that somehow subscribes to status change notifications from the api server. kubectl has something like this (e.g `kubectl get pods -w`)

SimonRichardson:
@achilleasa yeah, i'd agree with that - send deltas, not the full thing!!

PR #9985, opened 2019-03-30
increase go linter deadline value

hmlanigan:
## Description of change

Execution times for the golinters.bash on macOS for juju vary from 1h48 to 5m+.  Increase golinter deadline to reduce unfound issues when pushing juju git hub juju branches from macOS.


hmlanigan:
$$merge$$

PR #9987, opened 2019-03-31
2.5 raft recovery 1822454

jameinel:
## Description of change

Change RaftBackstop so it handles 'no servers'.

If you wipe out the "raft/" directory, then this server is no longer
present in the list of available servers, and raft just never gets
going.

This changes it so that when we detect that case, we BootstrapCluster
with ourselves as the leader.

This was done so that restore-backup would be able to recover the HA cluster. (Where the newly bootstrapped controller's identity is replaced by the identity of an old machine.) However, it makes sense for other places as well, as we do need a way to recover if 'raft/' ever gets wiped.

## QA steps

```
$ juju bootstrap lxd lxd --model-default "logging-config=<root>=INFO;juju.worker.raft=DEBUG"
$ juju ssh -m controller 0
$$ sudo -i
# systemctl stop jujud-machine-0
# cd /var/lib/juju
# rm -rf raft
# mkdir raft
# systemctl start jujud-machine-0
# tail -f /var/log/juju/machine-0.log
2019-03-31 12:15:35 INFO juju.worker.apicaller connect.go:158 [122fa7] "machine-0" successfully connected to "localhost:17070"
2019-03-31 12:15:36 DEBUG juju.worker.raft log.go:172 [WARN] raft: no known peers, aborting election
2019-03-31 12:15:40 INFO juju.core.raftlease store.go:248 timeout
```

Without this patch, the machine's raft engine will be stuck looping, With this patch, there should end up with a line about "rebootstrapping raft configuration". And from then Raft should be happy.
```
2019-03-31 12:21:38 INFO juju.worker.raft.raftbackstop worker.go:246 rebootstrapping raft configuration: raft.Configuration{Servers:[]raft.Server{raft.Server{Suffrage:0, ID:"0", Address:"10.210.24.119:17070"}}}
```

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1822454

### DIFF
--- a/apiserver/facades/client/backups/restore.go
+++ b/apiserver/facades/client/backups/restore.go
@@ -116,6 +116,7 @@ func (a *API) Restore(p params.RestoreArgs) error {
 	// And passing os.Exit in wouldn't make this any better either,
 	// just using it subverts the expectations of *everything* else
 	// running in the process.
+	// XXX: We have ErrRestartAgent which should be capable of replacing this
 	os.Exit(1)
 	return nil
 }
--- a/cmd/juju/backups/backups.go
+++ b/cmd/juju/backups/backups.go
@@ -137,7 +137,7 @@ var getArchive = func(filename string) (rc ArchiveReader, metaResult *params.Bac
 	if err != nil {
 		return nil, nil, errors.Trace(err)
 	}
-	_, err = archive.Seek(0, os.SEEK_SET)
+	_, err = archive.Seek(0, io.SeekStart)
 	if err != nil {
 		return nil, nil, errors.Trace(err)
 	}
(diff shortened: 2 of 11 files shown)


jameinel:
!!build!!
error while writing out vendor tree: failed to write dep tree

jameinel:
The failed test is https://bugs.launchpad.net/juju/+bug/1780766 which I submitted 
https://github.com/juju/juju/pull/9992
$$merge$$

jameinel:
$$merge$$ known intermittent failure: https://bugs.launchpad.net/juju/+bug/1790076 (I've fixed my intermittent failure for today)

PR #9988, opened 2019-04-01
Part 1 of 3: force remove app / unit.

anastasiamac:
## Description of change

Providing functionality to force remove application or unit is a bit involved, so I am breaking the change into 3 separate proposals.
This PR, part 1, only contains weaving of 'force' into direct state calls and the logic persevering despite encountered errors. 
Part 2 contains changes to cleanup operations to react to 'force' with potential Part 3 containing errors propagating up to users despite removal proceeding as well as some common scenario tests.

### DIFF
--- a/state/application.go
+++ b/state/application.go
@@ -254,6 +254,10 @@ type DestroyApplicationOperation struct {
 	// are removed. If this is false, then the operation will
 	// fail if there are any offers remaining.
 	RemoveOffers bool
+
+	// Force controls whether or not the destruction of an application
+	// will be forced, i.e. ignore operational errors.
+	Force bool
 }
 
 // Build is part of the ModelOperation interface.
@@ -265,7 +269,7 @@ func (op *DestroyApplicationOperation) Build(attempt int) ([]txn.Op, error) {
 			return nil, err
 		}
 	}
-	ops, err := op.app.destroyOps(op.DestroyStorage, op.RemoveOffers)
+	ops, err := op.app.destroyOps(op.DestroyStorage, op.RemoveOffers, op.Force)
 	switch err {
 	case errRefresh:
 		return nil, jujutxn.ErrTransientFailure
@@ -285,13 +289,24 @@ func (op *DestroyApplicationOperation) Done(err error) error {
 // destroyOps returns the operations required to destroy the application. If it
 // returns errRefresh, the application should be refreshed and the destruction
 // operations recalculated.
-func (a *Application) destroyOps(destroyStorage, removeOffers bool) ([]txn.Op, error) {
+func (a *Application) destroyOps(destroyStorage, removeOffers, force bool) ([]txn.Op, error) {
 	if a.doc.Life == Dying {
+		// TODO (anastasiamac) ??? If unit is already dying but we are forcing, should we re-try
+		// removal all over again? i.e. don't throw errAlreadyDying but do everything below again?..
 		return nil, errAlreadyDying
 	}
+
+	// We want to collect errors when forcing in order to report
+	// these to the user.
+	// TODO (anastasiamac) report removal erros to the user when forcing.
+	errs := []string{}
 	rels, err := a.Relations()
 	if err != nil {
-		return nil, err
+		if !force {
+			return nil, err
+		}
+		// If we are forcing the removal, just proceed.
+		errs = append(errs, err.Error())
 	}
 	if len(rels) != a.doc.RelationCount {
 		// This is just an early bail out. The relations obtained may still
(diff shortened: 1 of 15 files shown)


anastasiamac:
[weave force into apiserver/state](https://trello.com/c/reC5NHl6/662-weave-force-into-apiserver-state)

state/relation.go
```diff
@@ -249,12 +249,25 @@ func (r *Relation) checkConsumePermission(offerUUID, userId string) (bool, error
 	return true, nil
 }
 
+// DestroyWithForce may force the destruction of the relation.
+func (r *Relation) DestroyWithForce(force bool) error {
```
babbageclunk: Since this is new, may as well make it return the errors up.

state/unit.go
```diff
@@ -891,27 +965,52 @@ func (u *Unit) EnsureDead() (err error) {
 // the application is Dying and no other references to it exist. It will fail if
 // the unit is not Dead.
 func (u *Unit) Remove() (err error) {
+	return u.RemoveWithForce(false)
+}
+
+// ForceRemove removes the unit from state similar to the unit.Remove() but
```
babbageclunk: Function renamed?

state/unit.go
```diff
@@ -891,27 +965,52 @@ func (u *Unit) EnsureDead() (err error) {
...
+				failRelations = err
+				continue
+			}
+			if err := ru.LeaveScope(); err != nil {
+				errs = append(errs, err)
+				failRelations = err
+				continue
+			}
+		}
+		if !force && failRelations != nil {
```
babbageclunk: Is this supposed to be inside the loop? Or are you intending to call LeaveScope on all of the relation units even in the non-force case?

state/application.go
```diff
@@ -354,17 +416,18 @@ func (a *Application) destroyOps(destroyStorage, removeOffers bool) ([]txn.Op, e
...
 	// being removed: the difference between 1 unit and 1000 is irrelevant.
-	if a.doc.UnitCount > 0 {
+	if op.app.doc.UnitCount > 0 {
 		cleanupOp := newCleanupOp(
 			cleanupUnitsForDyingApplication,
-			a.doc.Name,
-			destroyStorage,
+			op.app.doc.Name,
+			op.DestroyStorage,
+			op.Force,
```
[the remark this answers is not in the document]
babbageclunk: Oh no, looking at how cleanup marshalling works that doesn't really make sense.

state/application.go
```diff
@@ -285,63 +309,101 @@ func (op *DestroyApplicationOperation) Done(err error) error {
...
 		// asserts on relationcount and on each known relation, below.
 		return nil, errRefresh
 	}
-	ops := []txn.Op{minUnitsRemoveOp(a.st, a.doc.Name)}
+	ops := []txn.Op{minUnitsRemoveOp(op.app.st, op.app.doc.Name)}
 	removeCount := 0
+	failedRels := false
 	for _, rel := range rels {
-		relOps, isRemove, err := rel.destroyOps(a.doc.Name)
+		relOps, isRemove, opErrs, err := rel.destroyOps(op.app.doc.Name, op.Force)
```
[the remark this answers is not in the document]
anastasiamac: Completely agree and I started doing but it blew out this work to unreasonable. I will follow up once Part 2 lands with a dedicated relation remove operation PR.

state/application.go
```diff
@@ -796,8 +878,12 @@ func (a *Application) changeCharmOps(
 	// Drop the references to the old settings, storage constraints,
 	// and charm docs (if the refs actually exist yet).
 	if oldKey != nil {
-		decOps, err = appCharmDecRefOps(a.st, a.doc.Name, a.doc.CharmURL, true) // current charm
+		// Since we can force this now, let's.. There is no point hanging on
+		// to the old key.
+		decOps, _, err = appCharmDecRefOps(a.st, a.doc.Name, a.doc.CharmURL, true, true) // current charm
```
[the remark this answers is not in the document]
anastasiamac: I actually think that we do want to force removal here as we do really really want to get rid of the old references.
However, I agree that we could stop progression if we have encountered an error here and are not forcing, i.e. the same functional behavior as before.

state/relation.go
anastasiamac: Good point. Will adjust this and other, similar places \o/

state/unit.go
anastasiamac: doh :( Thank you!

state/unit.go
anastasiamac: I was going to keep going until I was done with all relations.. Why? Just because one failed, should not we try the rest?
If we are in non-force case and encountered any error, we'll still fall over with latest error after going through all of them.

anastasiamac: After an offline discussion, we'll leave as is :) Thank you for your time and for pondering over this \o/

anastasiamac:
$$merge$$

anastasiamac:
unrelated failure in github.com/juju/juju/worker/uniter.TestPackage

anastasiamac:
$$merge$$

PR #9990, opened 2019-04-01
vsphere: Allow specifying datastore with the root-disk-source constraint

babbageclunk:
## Description of change

Adds a new constraint root-disk-source. This will be the name of storage where the root disk for a machine should reside - it's currently only interpreted as the name of the datastore that shoud be used in vsphere. This allows units of different applications to be allocated in different datastores which is important for deployments of complex systems.

RootDiskSource is also added to instance.HardwareCharacteristics and machine instance data, so the selected datastore is stored for the created machine and can be used to filter machines to find one that matches the constraint. The new instance data and constraint attributes are included in model migrations.

## QA steps

* Bootstrap a vsphere controller.
* Deploy an application including the new constraint `--constraints="root-disk-source=datastore-name"`
* Verify that the new machine's root disk is created in the selected datastore
* Running `juju show-machine <new machine>` shows the root-disk-source in constraints and the hardware.
* Create a new machine without using the constraint.
* The hardware root-disk-source attribute is set to the default datastore for the new machine.

## Documentation changes

* The new root-disk-source constraint  and how it integrates with vSphere will need documentation.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1807960


instance/instance.go
```diff
@@ -72,6 +72,8 @@ type HardwareCharacteristics struct {
 
 	// RootDisk is the size of the disk in megabytes.
 	RootDisk *uint64 `json:"root-disk,omitempty" yaml:"rootdisk,omitempty"`
+	// RootDiskSource is where the disk storage resides.
```
wallyworld: newline

babbageclunk:
[vsphere: select datastore using root-disk-source](https://trello.com/c/9AVomEQf/672-vsphere-select-datastore-using-root-disk-source)

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -406,13 +402,16 @@ func (c *Client) selectDatastore(
...
+		c.logger.Debugf("desired datasource %q", dsName)
 		for _, ds := range datastores {
-			if ds.Name == args.Datastore {
+			c.logger.Debugf("seen %q", ds.Name)
+			if ds.Name == dsName {
 				return &ds, nil
 			}
 		}
-		return nil, errors.Errorf("could not find datastore %q", args.Datastore)
+		return nil, errors.Errorf("could not find datastore %q", dsName)
```
wallyworld: I think there's a pre-checks bit in the workflow - could be use that to fail early? Would mean that juju deploy exists with an error

state/state_test.go
```diff
@@ -1330,6 +1330,7 @@ func (s *StateSuite) TestInjectMachine(c *gc.C) {
 	arch := "amd64"
 	mem := uint64(1024)
 	disk := uint64(1024)
+	source := "loveshack"
```
wallyworld: tin roof..... rusty

provider/vsphere/internal/vsphereclient/createvm.go
babbageclunk: Good call - looking into that now.

provider/vsphere/environ_policy.go
```diff
@@ -23,8 +24,51 @@ func (env *environ) PrecheckInstance(ctx context.ProviderCallContext, args envir
 
 // PrecheckInstance is part of the environs.Environ interface.
 func (env *sessionEnviron) PrecheckInstance(ctx context.ProviderCallContext, args environs.PrecheckInstanceParams) error {
-	_, err := env.parsePlacement(ctx, args.Placement)
-	return err
+	if _, err := env.parsePlacement(ctx, args.Placement); err != nil {
```
wallyworld: args.Placement may be "" now right?
So we should check for that?
And in other providers too?

provider/vsphere/environ_policy.go
```diff
@@ -23,8 +24,51 @@ func (env *environ) PrecheckInstance(ctx context.ProviderCallContext, args envir
...
+		return errors.Trace(err)
+	}
+	if err := env.checkZones(ctx, args.Constraints.Zones); err != nil {
```
wallyworld: And exit early if constraints not set?

state/state.go
```diff
@@ -1468,54 +1507,31 @@ func (st *State) processIAASModelApplicationArgs(args *AddApplicationArgs) error
 					err, "cannot deploy to machine %s", m,
 				)
 			}
+			// We've checked the machine placement, there's not a
+			// provider-specific placement that needs checking
+			// (although the instance creation still needs to be
+			// checked).
+			placements.Add("")
```
wallyworld: This seems a bit icky. Shouldn't the check just skip over non-specified placement

babbageclunk: No, we should check the arguments as if there is no provider-relevant placement, so the provider gets a chance to check constraints and volumes.

provider/vsphere/environ_policy.go
babbageclunk: Well, we exit early above if both the constraints and the placement is empty. And parsePlacement handles empty placement correctly too (as checked by tests).

I'll go through the other providers to check they're not going to freak out being called with placement="".

provider/vsphere/environ_policy.go
babbageclunk: Constraints is a struct rather than a pointer, so it ends up being `args.Constraints.String() == ""` - do you think we need an early return here? Both of the checks return early if the zones/datastore are nil.

state/state.go
babbageclunk: Oh, do you mean we don't need to do this unless there're no other placement directives so we can just rely on the check at the end?

provider/vsphere/environ_policy.go
babbageclunk: Checking through the implementations of PrecheckInstance under `provider` most of them are clearly written with the expectation that they'll be called even when there's no placement directive (other than maas that only looks at placement). They all have code paths that deal with placement being "".

Hmm, the manual provider's one always fails. So potentially this change would break that. I'm going to check that a bit more.

state/state.go
babbageclunk: Ah, I think I understand your point now - for machine-placement, we're specifying an existing machine, so we don't need to check instance args because we're not going to create a new instance. Is that right?
That solves the problem with the manual provider as well - we should always be using machine placement to deploy in that case because the cloud can't create instances.

provider/vsphere/environ_policy.go
babbageclunk: I realised that the right thing to do here is not to do prechecking for machine placement because we're not creating instances.

babbageclunk:
$$merge$$

PR #9992, opened 2019-04-01
Change so that a delay=0 doesn't use a timer.

jameinel:
## Description of change

We don't really want to use real clock time.NewTimer anyway, so get rid
of synchronization via 'time.NewTimer(0)' and just treat delay=0 as
being no delay at all.

## QA steps

```
$ go test -failfast -count=1000 -race -check.v -check.f StatusSuite.TestRun
```

Without this patch, I got a failure from time to time with running count=100, with the patch, 1000 runs complete in 23s and have no failures.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1780766


jameinel:
$$merge$$

PR #9994, opened 2019-04-01
Make golinters optional

SimonRichardson:
## Description of change

The following PR changes the golinters to be optional, to prevent
timeouts from certain setups. Also people can turn them on easily
by setting the IGNORE_GOLINTERS=1.

## QA steps

If you've set up the git pre-push, this just runs `./scripts/verify.bash`
So to test it, you can run `./scripts/verify.bash` and it will ignore the
golinters, otherwise setting `IGNORE_GOLINTERS=1` and running
the command again, should run the golinters.

### DIFF
--- a/scripts/verify.bash
+++ b/scripts/verify.bash
@@ -72,7 +72,7 @@ go vet \
     $DIRNAMES || [ -n "$IGNORE_VET_WARNINGS" ]
 
 # Allow the ignoring of the golinters
-if [ -n "$IGNORE_GOLINTERS" ]; then
+if [ -n "$INCLUDE_GOLINTERS" ]; then
     echo "checking: golinters ..."
     ./scripts/golinters.bash
 else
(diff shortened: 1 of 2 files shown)


scripts/verify.bash
```diff
@@ -72,7 +72,7 @@ go vet \
     $DIRNAMES || [ -n "$IGNORE_VET_WARNINGS" ]
 
 # Allow the ignoring of the golinters
-if [ -z "$IGNORE_GOLINTERS" ]; then
+if [ -n "$IGNORE_GOLINTERS" ]; then
```
jameinel: I think we have to change the env var. Having "IGNORE_GOLINTERS=1" mean *don't* ignore, is confusing.
RUN_GOLINTERS
INCLUDE_GOLINTERS
DONT_IGNORE_GOLINTERS
I think I like INCLUDE the most.

SimonRichardson: 💯 agree, naming is hard and this makes sense

SimonRichardson:
https://trello.com/c/wompFlPA/1251-make-golinters-optional

SimonRichardson:
$$merge$$

PR #9995, opened 2019-04-01
Expose ca cert fingerprint in show controller output

achilleasa:
## Description of change

This PR updates the show-controller command to include a SHA-256 fingerprint for the controller's CA cert. It is part of the work required for implementing redirects for migrated models. The fingerprint format matches the output of `openssl x509 -noout -fingerprint -sha256 -inform pem -in CA.pem` 

## QA steps

```console
$ juju bootstrap lxd lxd-test
$ juju show-controller | grep ca-
 ca-cert: |
    ca-fingerprint: XX:YY:....:ZZ

# As an extra check verify that openssl output matches fingerprint
$ juju show-controller \
    | sed -n -e '/ca-cert:/,/  -----END/p; / -----END/q' \
    | sed -e 's/^[ ]*//' \
    | tail -n+2 \
    | openssl x509 -noout -fingerprint -sha256 -inform PEM
SHA256 Fingerprint=XX:YY:...:ZZ
```

## Documentation changes

@pmatulis Do we need to update the man pages to include the new attribute in the `show-controller` output?

### DIFF
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -4,7 +4,9 @@
 package cert
 
 import (
+	"crypto/sha256"
 	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"time"
 
@@ -63,6 +65,39 @@ func NewServer(caCertPEM, caKeyPEM string, expiry time.Time, hostnames []string)
 		ExtKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
 		KeyBits:     NewLeafKeyBits,
 	})
+}
+
+var hexAlphabet = [16]byte{'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F'}
+
+// Fingerprint returns a human-readable SHA-256 fingerprint for a certificate
+// stored in the PEM format. The returned fingerprint matches the output of:
+// openssl x509 -noout -fingerprint -sha256 -inform pem -in cert.pem
+func Fingerprint(certPEM string) (string, error) {
+	block, _ := pem.Decode([]byte(certPEM))
+	if block == nil {
+		return "", errors.New("input does not contain a valid certificate in PEM format")
+	}
+
+	if _, err := x509.ParseCertificate(block.Bytes); err != nil {
+		return "", errors.Annotate(err, "cannot calculate certificate fingerprint")
+	}
+
+	// fingerprint format is: XX:YY:...:ZZ
+	fingerprint := make([]byte, (sha256.Size*3)-1)
+	var index int
+	for _, fb := range sha256.Sum256(block.Bytes) {
+		if index != 0 {
+			fingerprint[index] = ':'
+			index++
+		}
+
+		// Encode each byte as two chars
+		fingerprint[index] = hexAlphabet[(fb>>4)&0xf]
+		fingerprint[index+1] = hexAlphabet[fb&0xf]
+		index += 2
+	}
+
+	return string(fingerprint), nil
 }
 
 // NewCA generates a CA certificate/key pair suitable for signing server
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -67,7 +67,7 @@ func NewServer(caCertPEM, caKeyPEM string, expiry time.Time, hostnames []string)
 	})
 }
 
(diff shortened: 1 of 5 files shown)


cert/cert.go
```diff
@@ -65,6 +67,39 @@ func NewServer(caCertPEM, caKeyPEM string, expiry time.Time, hostnames []string)
...
+// Fingerprint returns a human-readable SHA-256 fingerprint for a certificate
+// stored in the PEM format. The returned fingerprint matches the output of:
+// openssl x509 -noout -fingerprint -sha256 -inform pem -in cert.pem
+func Fingerprint(certPEM string) (string, error) {
+	block, _ := pem.Decode([]byte(certPEM))
+	if block == nil {
+		return "", errors.New("input does not contain a valid certificate in PEM format")
+	}
+
+	if _, err := x509.ParseCertificate(block.Bytes); err != nil {
```
SimonRichardson: What's the reason to match openssl directly, surely we can sum256 the cert?

```go
cert, err := x509.ParseCertificate(block.Bytes)
if err != nil {
    return "", errors.Annotate(err, "cannot calculate certificate fingerprint")
}
return fmt.Sprintf("%x", sha256.Sum256(cert.Raw)), nil
```

achilleasa: When we redirect users to the new controller for a migrated model we will also list the fingerprint for the new controller's CA. Then, when the user tries to `juju login` to the new controller we will display the fingerprint and ask them to trust it before proceeding.

As this is used in a purely client-facing scenario it made sense for me to format the fingerprint so it's easier to compare. I mentioned openssl in the comment more as an example; this format (or the variant with space as a delimiter) is pretty common out there.

pmatulis:
@achilleasa I don't think so, unless you see a docs use case. AFAICS, it's an internal blob of data.

cmd/juju/controller/showcontroller_test.go
```diff
@@ -468,6 +468,77 @@ func (s *ShowControllerSuite) TestShowControllerWithIdentityProvider(c *gc.C) {
...
+      azENMAsGA1UEChMEanVqdTEzMDEGA1UEAwwqanVqdS1nZW5lcmF0ZWQgQ0EgZm9y
+      IG1vZGVsICJqdWp1IHRlc3RpbmciMSUwIwYDVQQFExwxMjM0LUFCQ0QtSVMtTk9U
+      LUEtUkVBTC1VVUlEMFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAL+0X+1zl2vt1wI4
+      1Q+RnlltJyaJmtwCbHRhREXVGU7t0kTMMNERxqLnuNUyWRz90Rg8s9XvOtCqNYW7
+      mypGrFECAwEAAaNCMEAwDgYDVR0PAQH/BAQDAgKkMA8GA1UdEwEB/wQFMAMBAf8w
+      HQYDVR0OBBYEFHueMLZ1QJ/2sKiPIJ28TzjIMRENMA0GCSqGSIb3DQEBCwUAA0EA
+      ovZN0RbUHrO8q9Eazh0qPO4mwW9jbGTDz126uNrLoz1g3TyWxIas1wRJ8IbCgxLy
+      XUrBZO5UPZab66lJWXyseA==
+      -----END CERTIFICATE-----
+    ca-fingerprint: 93:D9:8E:B8:99:36:E8:8E:23:D5:95:5E:81:29:80:B2:D2:89:A7:38:20:7B:1B:BD:96:C8:D9:C1:03:88:55:70
```
jameinel: i wish ca-fingerprint was more legible, but it is the best that we have right now.

cert/cert.go
```diff
@@ -65,6 +67,39 @@ func NewServer(caCertPEM, caKeyPEM string, expiry time.Time, hostnames []string)
 	})
 }
 
+var hexAlphabet = [16]byte{'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F'}
```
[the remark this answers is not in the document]
achilleasa: Doh! That's why I shouldn't be pushing out PRs near the EOD :D
Let me push a commit to simplify this and address the other review points

cmd/juju/controller/showcontroller_test.go
achilleasa: How about renaming this to `ca-cert-fingerprint`?

achilleasa:
!!build!!

achilleasa:
$$merge$$

PR #9998, opened 2019-04-02
Feature/controller public dns

ycliuhw:
## Description of change

- make Juju CLI talking with controller using public DNS if possible (because the pod & service address 
 are all internally accessible and pod address might be changed very often) - mainly for k8s controller.
- this PR also enables us to bootstrap in public k8s clusters (like CDK on public clouds, GKE, AKS, EKS, etc)

**Note**: Juju CLI can talk to controllers running on public clusters via `public` service DNS now. But there are some other issues(not directly related with current PR) need to be solved to make the controller fully functioning.

## QA steps

- prepare a public k8s cluster, GKE for example;
- add cloud;

```bash
$ juju add-k8s <k8s-cloud> --local
```
- bootstrap controller;

```bash
$ juju bootstrap <k8s-cloud> <controller-name> --debug
```

- run any cmd to talk with k8s controller, juju status, juju add-model, etc for example;

```bash
$ juju add-model <model-name> <k8s-cloud>
$ juju create-storage-pool mariadb-pv kubernetes storage-class=microk8s-hostpath
$ juju create-storage-pool operator-storage kubernetes storage-class=microk8s-hostpath
$ juju deploy cs:~juju/mariadb-k8s-0  --debug --storage database=500M,mariadb-pv
$ juju deploy cs:~juju/gitlab-k8s-0  --debug  --config juju-external-hostname=127.0.0.1.xip.io
$ juju relate mariadb-k8s gitlab-k8s
$ juju expose gitlab-k8s
```

## Documentation changes

Not ready yet;

## Bug reference

None;


caas/kubernetes/provider/bootstrap.go
```diff
@@ -98,6 +104,36 @@ func DecideControllerNamespace(controllerName string) string {
 	return "controller-" + controllerName
 }
 
+// ReScheduler executes a function and keep retrying until timeout if the function returns an expected error.
+func ReScheduler(
```
wallyworld: We already have code for this in the juju/retry repo, eg

```
	err := retry.Call(retry.CallArgs{
		Attempts:    30,
		Delay:       time.Second,
		MaxDelay:    time.Minute, // because 2**29 seconds is beyond reasonable
		BackoffFunc: retry.DoubleDelay,
		Clock:       clock,
		Func: func() error {
			_, err := inst.DeleteSecurityGroup(group)
			if err == nil || isNotFoundError(err) {
				logger.Debugf("deleting security group %q", group.Name)
				return nil
			}
			return errors.Trace(maybeConvertCredentialError(err, ctx))
		},
		IsFatalError: func(err error) bool {
			return common.IsCredentialNotValid(err)
		},
		NotifyFunc: func(err error, attempt int) {
			logger.Debugf("deleting security group %q, attempt %d", group.Name, attempt)
		},
	})
```

caas/kubernetes/provider/constants.go
```diff
@@ -0,0 +1,12 @@
+// Copyright 2019 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package provider
+
+const (
```
wallyworld: There's consts already in caas/metadata.go

caas/kubernetes/provider/k8s.go
```diff
@@ -816,6 +816,31 @@ func (k *kubernetesClient) DeleteOperator(appName string) (err error) {
...
+	// has different way to report back LB address.
+	// This should cover most of case now.
+	lpAdd := svc.Spec.LoadBalancerIP
+	if lpAdd != "" {
+		return lpAdd
+	}
+
+	ing := svc.Status.LoadBalancer.Ingress
+	if len(ing) == 0 {
+		return lpAdd
```
wallyworld: This will be ""

caas/kubernetes/provider/k8s.go
```diff
@@ -1613,11 +1634,11 @@ func (k *kubernetesClient) configureService(
...
 		},
 	}
-	return k.ensureService(service)
+	return k.ensureK8sService(service)
 }
 
-// ensureService ensures a service resource.
-func (k *kubernetesClient) ensureService(spec *core.Service) error {
+// ensureK8sService ensures a k8s service resource.
+func (k *kubernetesClient) ensureK8sService(spec *core.Service) error {
```
wallyworld: why rename?

caas/kubernetes/provider/services.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2019 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package provider
+
+import (
+	core "k8s.io/api/core/v1"
+)
+
+var preferredControllerServiceTypes = map[string]core.ServiceType{
```
wallyworld: I think this can all go in caas/metadata.go as that's where we have started recording cluster specific preferences like storage etc

cmd/juju/common/controller.go
```diff
@@ -63,9 +66,17 @@ func tryAPI(c *modelcmd.ModelCommandBase) error {
...
+	ctx *cmd.Context,
+	c *modelcmd.ModelCommandBase,
+	isCAASController bool,
+	controllerName,
+	hostedModelName string,
+) (err error) {
 	// TODO(katco): 2016-08-09: lp:1611427
 	attempts := utils.AttemptStrategy{
+		// added total time to try for more stable of when we should give up;
+		Total: bootstrapReadyPollTotalTime,
```
wallyworld: not sure we need this

cmd/modelcmd/base.go
```diff
@@ -438,6 +439,8 @@ func newAPIConnectionParams(
 	}
 	dialOpts := api.DefaultDialOpts()
 	dialOpts.BakeryClient = bakery
+	// we should not use default timeout - 10 minutes.
+	dialOpts.Timeout = 15 * time.Second
```
wallyworld: The default should be changed in api.DefaultDialOpts()
10 minutes seems too long but was put there to allow controllers that are upgrading etc to finish.
Maybe a timeout of say 1 minute is better

environs/bootstrap/bootstrap.go
```diff
@@ -144,6 +144,10 @@ type BootstrapParams struct {
 	// CAPrivateKey is the controller's CA certificate private key.
 	CAPrivateKey string
 
+	// ControllerServiceType is the service type to use for a k8s controller.
+	// Optional: only for k8s controller.
+	ControllerServiceType string
```
wallyworld: We should be able to look this up from the cloud info on the server side.
I don't think there's a need to add a k8s specific attribute here

controller/config.go
```diff
@@ -529,6 +529,15 @@ func (c Config) AutocertDNSName() string {
 	return c.asString(AutocertDNSNameKey)
 }
 
+// SetAutocertDNSName sets controller public DNS.
+func (c Config) SetAutocertDNSName(controllerDNS string) error {
```
wallyworld: This is usually supplied by the user at bootstrap time via the --config argument.
Having bootstrap process overwrite it is problematic.
We could consider adding a cloudServiceDoc for the controller (and also cloudContainerDoc for the pods). This fits better with the data model and allows for when we will model controllers as applications

caas/kubernetes/provider/bootstrap.go
ycliuhw: got it, changed to use this

caas/kubernetes/provider/constants.go
ycliuhw: no, i didn't find it in metadata.go.
it's all using strings currently

caas/kubernetes/provider/k8s.go
ycliuhw: changed to ""

caas/kubernetes/provider/services.go
ycliuhw: moved to metadata.go

environs/bootstrap/bootstrap.go
ycliuhw: as discussed, unfortunately we have to pass through the whole process.
Because k8s cloud type is only known by add-k8s cmd then Juju will lose this knowledge forever.

cmd/juju/common/controller.go
ycliuhw: removed

apiserver/admin_test.go
```diff
@@ -1334,6 +1334,21 @@ func (s *macaroonLoginSuite) TestRemoteUserLoginToControllerLoginAccess(c *gc.C)
...
 	c.Check(result.UserInfo.ControllerAccess, gc.Equals, "login")
 	c.Check(result.UserInfo.ModelAccess, gc.Equals, "")
+	c.Check(result.Servers, gc.DeepEquals,
+		params.FromNetworkHostsPorts(
+			parseHostPortsFromAddress(c, info.Addrs...),
+		),
+	)
+}
+
+func parseHostPortsFromAddress(c *gc.C, addresses ...string) [][]network.HostPort {
```
wallyworld: Not sure there's sufficient test coverage here? Maybe it's too hard to create a k8s model and test the service address

caas/kubernetes/provider/bootstrap_test.go
```diff
@@ -54,7 +56,8 @@ func (s *bootstrapSuite) SetUpTest(c *gc.C) {
 	s.controllerUUID = "9bec388c-d264-4cde-8b29-3e675959157a"
 
 	s.controllerCfg = testing.FakeControllerConfig()
-	pcfg, err := podcfg.NewBootstrapControllerPodConfig(s.controllerCfg, controllerName, "bionic")
+	s.controllerCfg[controller.APIPort] = 17070
```
wallyworld: We try and avoid using 17070 in tests. Is there a reason not to stick with 17777 set up in FakeControllerConfig() ?

caas/metadata.go
```diff
@@ -7,11 +7,20 @@ import (
 	"fmt"
 
 	"github.com/juju/collections/set"
+	// core "k8s.io/api/core/v1"
```
wallyworld: delete me

cmd/juju/caas/add.go
```diff
@@ -487,6 +488,9 @@ func (c *AddCAASCommand) Run(ctx *cmd.Context) error {
 	if _, ok := newCloud.Config[provider.OperatorStorageKey]; !ok {
 		newCloud.Config[provider.OperatorStorageKey] = operatorStorageName
 	}
+	if _, ok := newCloud.Config[bootstrap.ControllerServiceTypeKey]; !ok {
```
wallyworld: We need to think about how to handle existing cloud definitions in clouds.yaml. Perhaps default to "LoadBalancer". The out of the box microk8s support will redo any micok8s config and correctly specify cluster ip.

cmd/juju/commands/bootstrap_test.go
```diff
@@ -1804,33 +1810,32 @@ func (s *BootstrapSuite) TestBootstrapPrintClouds(c *gc.C) {
...
-dummy-cloud-with-config                       
-dummy-cloud-with-region-config                
-dummy-cloud-without-regions                   
-many-credentials-no-auth-types                
-
-You will need to have a credential if you want to bootstrap on a cloud, see
-‘juju autoload-credentials’ and ‘juju add-credential’. The first credential
-listed is the default. Add more clouds with ‘juju add-cloud’.
-`[1:])
+	expectedOut := "You can bootstrap on these clouds. See ‘--regions <cloud>’ for all regions.\n" +
```
wallyworld: Why the change from `` to string concatenation? The original way is better and easier to edit

state/cloudservice.go
```diff
@@ -21,60 +21,78 @@ type CloudService interface {
...
-// Address implements CloudService.
-func (c *cloudService) Addresses() []network.Address {
+// Addresses implements CloudService.
+func (c *CloudService) Addresses() []network.Address {
 	return networkAddresses(c.doc.Addresses)
 }
 
-func (a *Application) cloudService() (*cloudServiceDoc, error) {
-	coll, closer := a.st.db().GetCollection(cloudServicesC)
+func (c *CloudService) cloudService() (cloudServiceDoc, error) {
```
wallyworld: can we continue to return a pointer *cloudServiceDoc

state/state.go
```diff
@@ -1118,6 +1118,47 @@ var (
 	errLocalApplicationExists          = errors.Errorf("application already exists")
 )
 
+type AddCloudServiceArgs struct {
+	// Id will be the application Name if it's a part of application,
+	// and will be controller UUID for k8s a controller(controller does Not have an application).
```
wallyworld: not

state/state.go
```diff
@@ -1118,6 +1118,47 @@ var (
...
+	Id         string
+	ProviderId string
+	Addresses  []network.Address
+}
+
+// AddCloudService creates a cloud service.
+func (st *State) AddCloudService(args AddCloudServiceArgs) (_ *CloudService, err error) {
```
wallyworld: This needs to be called "SaveCloudService" since the saveServiceOps() below does either an insert or update as needed. AddCloudService() would be a lie sometimes

state/state.go
```diff
@@ -1118,6 +1118,47 @@ var (
...
+	defer errors.DeferredAnnotatef(&err, "cannot add cloud service %q", args.ProviderId)
+
+	doc := cloudServiceDoc{
+		DocID:      args.Id,
+		ProviderId: args.ProviderId,
+		Addresses:  fromNetworkAddresses(args.Addresses, OriginProvider),
+	}
+	svc := newCloudService(st, &doc)
+
+	ops, err := svc.saveServiceOps(doc)
```
wallyworld: This needs to be part of a buildTxn loop, to cater for insert vs update races.
See for example (ec *externalControllers) Save() in state/externalcontroller.go

state/state.go
```diff
@@ -1118,6 +1118,47 @@ var (
...
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if err = st.db().RunTransaction(ops); err == nil {
+		if err = svc.Refresh(); err != nil {
```
wallyworld: This refresh after save pattern is not one we've used elsewhere have we?
Would prefer to stick to our standard save patterns

agent/agentbootstrap/bootstrap.go
```diff
@@ -391,6 +411,45 @@ func initBootstrapMachine(
...
+
+	env, err := getEnviron(isCAAS, args.ControllerModelConfig, args)
+	if err != nil {
+		return errors.Annotate(err, "getting environ")
+	}
+
+	broker, ok := env.(caas.ServiceGetterSetter)
+	if !ok {
+		// this should never happen.
+		return errors.New("CAAS environ is not a broker")
```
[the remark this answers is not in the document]
jameinel: Would it be better read as:
errors.Errorf("environ %T does not implement ServiceGetterSetter interface")
?

agent/agentbootstrap/bootstrap.go
```diff
@@ -391,6 +411,45 @@ func initBootstrapMachine(
...
+	}
+	svc, err := broker.GetService(k8sprovider.JujuControllerStackName)
```
jameinel: How does this affect things like CDK and microk8s where you may have a DNS name for services, but that DNS only resolves *inside* the K8s (eg, your laptop probably doesn't point to the K8s' DNS)

That is one of the main reason that Juju has historically preferred IP addresses, because DNS is only reliable if it is globally resolvable DNS (and things like MAAS DNS suffer from the same problem.

I know that we do have some support for controllers being advertised at DNS names (it is how we integrate with JAAS). And we do things like negotiate Let's Encrypt to sign our certificates rather than just self-signed.

caas/kubernetes/provider/base_test.go
```diff
@@ -120,7 +120,7 @@ func (s *BaseSuite) setupController(c *gc.C) *gomock.Controller {
 }
 
 func (s *BaseSuite) setupBroker(c *gc.C, ctrl *gomock.Controller, newK8sRestClientFunc provider.NewK8sClientFunc) *gomock.Controller {
-	s.clock = testclock.NewClock(time.Time{})
+	s.clock = testclock.NewClock(time.Now())
```
jameinel: Why?
This smells like something is wrong if we need a time that is close to real-time Now. The *point* of Clock is to make sure you don't depend on real time.

caas/kubernetes/provider/bootstrap_test.go
jameinel: shouldn't this be a dynamic port? or we know that this is never listened on or talked to so it is arbitrary?
I agree you want to avoid 17070 and have things "accidentally" work, and ignore the config

caas/kubernetes/provider/k8s.go
```diff
@@ -817,6 +817,31 @@ func (k *kubernetesClient) DeleteOperator(appName string) (err error) {
 	return errors.Trace(k.deleteDeployment(operatorName))
 }
 
+func getLoadBalancerAddress(svc *core.Service) string {
+	// reason we have this is because different cloud provider
+	// has different way to report back LB address.
+	// This should cover most of case now.
```
jameinel: different cloud providers have a different way to report back the Load Balancer address.
This covers the cases we know about so far.

cmd/juju/commands/bootstrap.go
```diff
@@ -750,36 +757,64 @@ See `[1:] + "`juju kill-controller`" + `.`)
...
-			AgentVersion:           agentVersion.String(),
-			CurrentHostPorts:       [][]network.HostPort{network.AddressesWithPort(addrs, config.controller.APIPort())},
-			PublicDNSName:          newStringIfNonEmpty(config.controller.AutocertDNSName()),
-			MachineCount:           newInt(1),
-			ControllerMachineCount: newInt(1),
-		}); err != nil {
-		return errors.Annotate(err, "saving bootstrap endpoint address")
+
+	retryCallArgs := retry.CallArgs{
+		Attempts:    20,
```
jameinel: we already have some retry logic for IAAS controllers because the time it takes to boot a machine can vary wildly, especially with MAAS. And we already have Config to allow users to set those things.
Setting a max retry timeout of 1minute seems very aggressive (our default timeout is 10min)
Is there a reason we need yet-another retry loop here that doesn't use our existing timeout configuration?

state/state.go
jameinel: we should be using st.db().Run(buildTxn) if we can, instead of RunTransaction, as it allows you to revalidate your assertions and give better error messages along with retrying if the db has changed vs memory.

state/state_test.go
```diff
@@ -1510,6 +1510,27 @@ func (s *StateSuite) TestAllRelations(c *gc.C) {
...
+	c.Assert(state.LocalID(s.State, svc.Id()), gc.Equals, "cloud-svc-ID")
+	c.Assert(svc.ProviderId(), gc.Equals, "dfsds-sdfsd-sdfds")
+	c.Assert(svc.Addresses(), gc.DeepEquals, network.NewAddresses("1.1.1.1"))
+
+	getResult, err := s.State.CloudService("cloud-svc-ID")
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(state.LocalID(s.State, getResult.Id()), gc.Equals, "cloud-svc-ID")
+	c.Assert(getResult.ProviderId(), gc.Equals, "dfsds-sdfsd-sdfds")
+	c.Assert(getResult.Addresses(), gc.DeepEquals, network.NewAddresses("1.1.1.1"))
+}
```
jameinel: Ideally we would add tests that use the transaction Hooks to ensure we are properly doing assertions against the database. (the hooks let you make a change right as the 'transaction' is starting, so that you can see that even though in-memory we think the content is Foo, we properly assert against the database, and notice that the content was changed by someone else to Bar.)

cmd/juju/commands/bootstrap.go
ycliuhw: ah, I have moved this retry into `provider.Bootstrap()`. I probably should remove this retry here.
The reason why we  need the 2nd retry strategy for k8s controller is we need to ensure LB provisioning has been finished (LB address is ready to use) and we need the LB address in bootstrap-state cmd or the cmd will fail.
So we wait and fail in juju CLI if we can't get the address to avoid failing in bootstrap-state in Jujud.

agent/agentbootstrap/bootstrap.go
```diff
@@ -391,6 +411,45 @@ func initBootstrapMachine(
...
+		return errors.Trace(err)
+	}
+	if len(svc.Addresses) == 0 {
+		// this should never happen because we have already checked in k8s controller bootstrap stacker.
+		return errors.NotProvisionedf("k8s controller service %q address", svc.Id)
+	}
+	svcId := args.ControllerConfig.ControllerUUID()
+	logger.Infof("creating cloud service for k8s controller %q", svcId)
+	cloudSvc, err := st.AddCloudService(state.AddCloudServiceArgs{
+		Id:         svcId,
```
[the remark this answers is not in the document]
ycliuhw: yeah, the arg.Id is wrapped with `applicationGlobalKey` in `state.AddCloudService` now.

apiserver/admin_test.go
ycliuhw: i was trying to add test for CAAS controller model, 
but i got some difficulty to setup a CAAS controller model because it's using `jujutesting.JujuConnSuite`..

caas/kubernetes/provider/bootstrap_test.go
ycliuhw: right, fixed!
now the expected spec is read from controllerCfg.
thanks

apiserver/admin_test.go
wallyworld: There's a pattern we use for that, search for 
// Set up a CAAS model to replace the IAAS one.

wallyworld:
> I think I saw that the default behavior is to use an IP if the service describes it, and then to fall back to a hostname if no IP is available. Which sounds likely to be the right behavior for most locations. But I'd like to confirm the intent.

We can't use hostname because the name resolution behaviour seems inconsistent across clusters, and on microk8s at the moment, doesn't seem to comply with the k8s docs. There's an internal cluster level endpoint we are looking to use in the future but for now the service external IP is all we have that will work across clusters

agent/agentbootstrap/bootstrap.go
wallyworld: The idea is we send the public service IP address to JUju clients (which is what this code is getting).
Internally we want to use something else, and ostensibly there's an internal cluster DNS and a naming scheme https://kubernetes.io/docs/concepts/services-networking/dns-pod-service. But experiments have shown that not everything behaves as documented at the moment.

state/state_test.go
ycliuhw: very handy, added 2 more tests using hooks for negative cases; thanks

agent/agentbootstrap/bootstrap.go
```diff
@@ -289,6 +283,32 @@ func ensureHostedModel(
 	return nil
 }
 
+func getEnviron(isCAAS bool, modelConfig *config.Config, args InitializeStateParams) (env environs.BootstrapEnviron, err error) {
```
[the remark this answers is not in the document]
ycliuhw: you are right, changed,

state/cloudservice.go
ycliuhw: changed

state/state.go
ycliuhw: yup, changed

caas/kubernetes/provider/base_test.go
ycliuhw: ah just changed to Now without thinking because I saw somewhere else uses `time.Now()`, 
yeah , u r right, it could be any time.

apiserver/admin_test.go
ycliuhw: I mean to setup a CAAS `controller` model which requires refactor a lot on `jujutesting.JujuConnSuite` itself.

cmd/juju/commands/bootstrap_test.go
ycliuhw: the original one has lots of trailing spaces on most of lines.
tests would be failed due to those spaces are automatically removed by linter..
it's quit annoying that I have to disable linter whenever i have to `ctrl+s` the file.

cmd/juju/caas/add.go
ycliuhw: yeah, the result uses LB now by default unless it can find related definition for this cloud.

ycliuhw:
!!build!!

ycliuhw:
$$merge$$

ycliuhw:
[introduce public DNS for Juju controller](https://trello.com/c/0P1rLr44/681-introduce-public-dns-for-juju-controller)

PR #9999, opened 2019-04-02
Add instance.id to CharmProfilingInfo

SimonRichardson:
## Description of change

The following changes adds the CharmProfilingInfo to the api and
apiserver. Additionally it also adds instance.Id to the machine
to ensure that we follow the same semantics (we're congruent) with
the state machine.

## QA steps

This is a mechanical change, and should be noted that running the unit
tests should ensure that the instance.Id is passed through.

Additionally you can alter the `worker/instancemutator/mutater.go` file to
log out the `InstanceId` from the `CharmProfilingInfo`.


core/cache/machine.go
```diff
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/juju/errors"
+	"github.com/juju/juju/core/instance"
```
jameinel: wrong section 'juju/juju' (this project) is its own section vs external non-stdlib section.

SimonRichardson:
> I don't see why this is needed yet, but the change by itself seems rather straightforward.

This is for the instance mutater worker...

core/cache/machine.go
SimonRichardson: I really need to get my stanza checker working as a pre-push for me :disappointed:

SimonRichardson: I've got it working now - https://youtu.be/EguALVNLwjM

apiserver/facades/agent/instancemutater/instancemutater.go
```diff
@@ -264,22 +272,33 @@ func (api *InstanceMutaterAPI) getMachine(canAccess common.AuthFunc, tag names.M
...
+		return empty, false, errors.Trace(err)
 	}
 	modelName := model.Name()
 
 	machineProfiles, err := m.CharmProfiles()
 	if err != nil {
-		return nil, machineProfiles, false, errors.Trace(err)
+		return empty, false, errors.Trace(err)
 	}
 	changeResults := make([]params.ProfileChangeResult, len(unitNames))
```
hmlanigan: You can do a machine.Units() from the model cache now.

The big difference with model cache vs state.  Is that the individual structs don't give you the thing, you have to get it from the model.  For instance, you get the application name from the model cache unit.  Then ask the model cache model for the application based on the name.

api/instancemutater/machine.go
```diff
@@ -142,6 +167,7 @@ func (m *Machine) WatchApplicationLXDProfiles() (watcher.NotifyWatcher, error) {
 
 type ProfileInfo struct {
```
hmlanigan: Something like:
type ProfileInfo struct {
   CurrentProfiles []string
   ModelName       string
   InstanceId      string
   ProfileUnits    []UnitProfileInfo
}

type UnitProfileInfo struct {
   ApplicationName string
   CharmRev        int
   Profile         *CharmLXDProfile
}

api/instancemutater/machine.go
```diff
@@ -201,7 +201,6 @@ func (m *Machine) CharmProfilingInfo() (*ProfileInfo, error) {
 	}
 	profileChanges := make([]ProfileChanges, len(result.ProfileChanges))
 	for i, change := range result.ProfileChanges {
-		profileChanges[i].Subordinate = change.Subordinate
 		if change.Profile != nil {
 			profileChanges[i].Profile = &CharmLXDProfile{
```
hmlanigan: There's an lxdprofile.Profile{} we can use now instead of the local CharmLXDProfile{}.  If so, can CharmLXDProfiles{} be removed?

api/instancemutater/machine.go
```diff
@@ -12,16 +12,21 @@ import (
...
 
 //go:generate mockgen -package mocks -destination mocks/caller_mock.go github.com/juju/juju/api/base APICaller,FacadeCaller
 //go:generate mockgen -package mocks -destination mocks/machinemutater_mock.go github.com/juju/juju/api/instancemutater MutaterMachine
 type MutaterMachine interface {
+
+	// InstanceId returns the provider specific instance id for this machine
+	InstanceId() (string, error)
+
 	// CharmProfilingInfo returns info to update lxd profiles on the machine
 	// based on the given unit names.
```
hmlanigan: "based on the given unit names." can be removed based on changes to CharmProfilingInfo()

api/instancemutater/machine.go
```diff
@@ -141,16 +166,15 @@ func (m *Machine) WatchApplicationLXDProfiles() (watcher.NotifyWatcher, error) {
...
 	CurrentProfiles []string
 }
 
 type ProfileChanges struct {
-	OldProfileName string
-	NewProfileName string
-	Profile        *CharmLXDProfile
-	Subordinate    bool
+	Profile     *CharmLXDProfile
+	Subordinate bool
```
hmlanigan: Subordinate can be removed now as unused.

api/instancemutater/machine.go
```diff
@@ -141,16 +166,15 @@ func (m *Machine) WatchApplicationLXDProfiles() (watcher.NotifyWatcher, error) {
...
 
 type ProfileInfo struct {
-	Changes         bool
+	ModelName       string
+	InstanceId      instance.Id
 	ProfileChanges  []ProfileChanges
 	CurrentProfiles []string
 }
 
 type ProfileChanges struct {
```
hmlanigan: Suggest renaming to something like UnitProfileInfo{}. We also need the application name and the charm revision data here.

There's an lxdprofile.Profile{} we can use now instead of the local CharmLXDProfile{}. If so, can CharmLXDProfiles{} be removed?

apiserver/facades/agent/instancemutater/instancemutater.go
```diff
@@ -93,29 +94,36 @@ func NewInstanceMutaterAPI(st InstanceMutaterState,
 // CharmProfilingInfo returns info to update lxd profiles on the machine
 // based on the given unit names.  If the machine is not provisioned,
```
hmlanigan: "based on the given unit names." can be removed based on changes to CharmProfilingInfo()

apiserver/facades/agent/instancemutater/interface.go
```diff
@@ -68,11 +67,33 @@ type LXDProfile interface {
...
+}
+
+// ModelCacheUnit represents a point of use Unit from the cache package.
+type ModelCacheUnit interface {
+	Application() string
+}
+
+// ModelCacheCharm represents point of use methods from the cache charm object
+type ModelCacheCharm interface {
+	LXDProfile() lxdprofile.Profile
```
hmlanigan: +1

core/cache/machine.go
```diff
@@ -35,6 +37,20 @@ func (m *Machine) Id() string {
 	return m.details.Id
 }
 
+// InstanceId returns the provider specific instance id for this machine and
+// returns not provisioned if instance id is empty
+func (m *Machine) InstanceId() (instance.Id, error) {
+	if m.details.InstanceId == "" && m.details.HardwareCharacteristics == nil {
```
hmlanigan: Should this use the machine lock, in case details is updated?

state/allwatcher.go
```diff
@@ -272,6 +272,7 @@ func (m *backingMachine) updated(st *State, store *multiwatcherStore, id string)
 		info.InstanceStatus = oldInfo.InstanceStatus
 		info.InstanceId = oldInfo.InstanceId
 		info.HardwareCharacteristics = oldInfo.HardwareCharacteristics
+		info.CharmProfiles = oldInfo.CharmProfiles
```
hmlanigan: Would we rather have no data or old data here?  No data might be better?  Then the worker can handle it.

SimonRichardson: So no data could be a no-op. OK, that makes sense

SimonRichardson:
$$merge$$

PR #10004, opened 2019-04-03
Add set-charms to instance-mutater API

SimonRichardson:
## Description of change

The following seems to be missing from the api, even though it's
there in the apiserver.

## QA steps

This is a mechanical change, and should be noted that running the unit
tests should ensure that the set charm profiles are passed through.

Additionally you can alter the `worker/instancemutator/mutater.go` file to
log out the `CurrentProfiles` from the `CharmProfilingInfo`.


SimonRichardson:
https://trello.com/c/SO9kae70/1149-api-apiserver-implementation-of-machinesetcharmprofiles

hmlanigan:
$$merge$$

PR #10005, opened 2019-04-03
adding AssignProfiles to LXDProfileManager interface for instance mutater worker

hmlanigan:
## Description of change

Adding AssignProfiles to LXDProfileManager and LXDProfiler interfaces.  Also added lxd server UpdateContainerProfiles() for use by the implementations of AssignProfiles.  These are needed by the Instance Mutater worker.

## QA steps

Not hooked up for use yet.  No change  to juju functionality should currently be seen.

### DIFF
--- a/container/lxd/manager.go
+++ b/container/lxd/manager.go
@@ -20,6 +20,7 @@ import (
 	"github.com/juju/juju/container"
 	"github.com/juju/juju/core/constraints"
 	"github.com/juju/juju/core/instance"
+	"github.com/juju/juju/core/lxdprofile"
 	"github.com/juju/juju/core/status"
 	"github.com/juju/juju/environs"
 	"github.com/juju/juju/environs/config"
@@ -290,6 +291,10 @@ func (m *containerManager) networkDevicesFromConfig(netConfig *container.Network
 	return nics, nil, errors.Trace(err)
 }
 
+// TODO: HML 2-apr-2019
+// When provisioner_task processProfileChanges() is
+// removed, maybe change to take an lxdprofile.ProfilePost as
+// an arg.
 // MaybeWriteLXDProfile implements container.LXDProfileManager.
 func (m *containerManager) MaybeWriteLXDProfile(pName string, put *charm.LXDProfile) error {
 	m.profileMutex.Lock()
@@ -318,6 +323,9 @@ func (m *containerManager) LXDProfileNames(containerName string) ([]string, erro
 	return m.server.GetContainerProfiles(containerName)
 }
 
+// TODO: HML 2-apr-2019
+// remove when provisioner_task processProfileChanges() is
+// removed.
 // ReplaceOrAddLXDProfile implements environs.LXDProfiler.
 func (m *containerManager) ReplaceOrAddInstanceProfile(instId, oldProfile, newProfile string, put *charm.LXDProfile) ([]string, error) {
 	if put != nil {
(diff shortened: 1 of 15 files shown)


provider/lxd/environ_test.go
```diff
@@ -377,36 +377,20 @@ type environProfileSuite struct {
...
-			ProfilePut: api.ProfilePut{
-				Config: map[string]string{
-					"security.nesting": "true",
-				},
-				Description: "test profile",
-			},
-		}).Return(nil),
-	)
+	profile := "testname"
+	s.expectMaybeWriteLXDProfile(false, profile)
```
hmlanigan: some noise in this file, i wanted to update to the pattern of expect*() for setup while here.

container/lxd/manager_test.go
```diff
@@ -119,31 +126,25 @@ func prepNetworkConfig() *container.NetworkConfig {
...
-	deleteOp := lxdtesting.NewMockOperation(ctrl)
-	deleteOp.EXPECT().Wait().Return(nil)
-
-	exp := cSvr.EXPECT()
+	exp := s.cSvr.EXPECT()
 
 	// Arrangements for the container creation.
-	expectCreateContainer(ctrl, cSvr, "juju/xenial/"+s.Arch(), "foo-target")
-	exp.UpdateContainerState(hostName, lxdapi.ContainerStatePut{Action: "start", Timeout: -1}, "").Return(startOp, nil)
+	s.expectCreateContainer(ctrl)
```
hmlanigan: some noise in this file, i wanted to update to the pattern of expect*() for setup while here.

container/lxd/manager.go
```diff
@@ -336,3 +344,44 @@ func (m *containerManager) ReplaceOrAddInstanceProfile(instId, oldProfile, newPr
...
+			var err2 error
+			current, err2 = m.LXDProfileNames(instId)
+			if err2 != nil {
+				logger.Errorf("secondary error, retrieving profile names: %s", err2)
+			}
+		}
+		return current, err
+	}()
+
+	deleteProfiles := []string{}
```
SimonRichardson: smallest of nits

```suggestion
	var deleteProfiles []string
```

container/lxd/manager.go
```diff
@@ -336,3 +344,44 @@ func (m *containerManager) ReplaceOrAddInstanceProfile(instId, oldProfile, newPr
 	}
 	return m.LXDProfileNames(instId)
 }
+
+// AssignProfiles implements environs.LXDProfiler.
+func (m *containerManager) AssignProfiles(instId string, profilesNames []string, profilePosts []lxdprofile.ProfilePost) (current []string, err error) {
+	defer func() ([]string, error) {
+		if err != nil {
```
SimonRichardson: I'm not entirely sure this works as expected, as all `err` below this use `:=`, so nothing will get past here.

I wonder if we can do this a better way and wrap all exits you want to pass this back in a function.

```go
report := func(err error) ([]string, error) {
    current, err1 := m.LXDProfileNames(instId)
    if err1 != nil {
          logger.Errorf("secondary error, retrieving profile names: %s", err1)
    }
    return current, err
}
...
return report(err)
```

container/lxd/server.go
```diff
@@ -225,6 +228,30 @@ func addRemoveReplaceProfileName(profiles []string, oldProfile, newProfile strin
...
+		return errors.Trace(errors.Annotatef(err, "failed to update %q with profiles", name))
+	}
+
+	op := resp.Get()
+	logger.Debugf("updated %q, waiting on %s", name, op.Description)
+	err = resp.Wait()
+	if err != nil {
+		logger.Tracef("updating %q failed on %q", name, err)
+	}
+	return errors.Trace(err)
```
SimonRichardson: I'm unsure we need the trace if we're already passing the error back

```suggestion
	if err := resp.Wait(); err != nil {
		logger.Tracef("updating %q failed on %q", name, err)
		return errors.Trace(err)
	}
	return nil
```

provider/lxd/environ.go
```diff
@@ -365,3 +373,45 @@ func (env *environ) ReplaceOrAddInstanceProfile(instId, oldProfile, newProfile s
 	}
 	return env.LXDProfileNames(instId)
 }
+
+// AssignProfiles implements environs.LXDProfiler.
+func (env *environ) AssignProfiles(instId string, profilesNames []string, profilePosts []lxdprofile.ProfilePost) (current []string, err error) {
+	defer func() ([]string, error) {
```
SimonRichardson: See previous comment about the defer being too cryptic for what this is doing.

provider/lxd/environ.go
```diff
@@ -365,3 +373,45 @@ func (env *environ) ReplaceOrAddInstanceProfile(instId, oldProfile, newProfile s
...
+	}
+
+	server := env.server()
+	if err := server.UpdateContainerProfiles(instId, profilesNames); err != nil {
+		return []string{}, errors.Trace(err)
+	}
+
+	for _, name := range deleteProfiles {
+		if err := server.DeleteProfile(name); err != nil {
+			// most likely the failure is because the profile is already in use
```
SimonRichardson: Orphan profiles although look bad from the point of view of `lxd profile list` isn't, unless it grows to 1000s I guess. We should make some sort of documentation around this.

container/lxd/manager.go
hmlanigan: done, this works better, ty for the idea.

container/lxd/manager.go
hmlanigan: done

container/lxd/server.go
hmlanigan: removed

provider/lxd/environ.go
hmlanigan: done

provider/lxd/environ.go
hmlanigan: The last profile on a machine destroying may not be deleted.  There might be a way with our new worker, to attempt to delete them. Will have to ponder.

hmlanigan:
$$merge$$

hmlanigan:
A mock lxdprofiler and the dummy provider required updates for their use to work.

$$merge$$

hmlanigan:
$$merge$$

hmlanigan:
"failed to start mongo: mongod failed to listen on port 38220"

$$merge$$

PR #10010, opened 2019-04-04
Support variable payloads in rpc request errors

achilleasa:
## Description of change

This PR is part of the work for supporting redirects for migrated models. I have split this work into a standalone PR to make reviewing easier. 

The main idea behind this PR is to make `params.Error` support variable `Info` payloads and to make it possible for them to be returned as part of an RPC `ResponseError` object. This change will allow us to support all those use-cases where we need to attach structured information to errors returned by the apiserver. 

One example of this (and relevant to the redirect work) is to be able to intercept calls to `Login` that attempt to login to a model that has been migrated return a `Redirect` error that includes the addresses/CA cert of the controller to connect to.

Half of the work in the PR updates the rpc/jsoncodec packages to support marshaling/unmarshaling of additional error information. As the new `ErrorInfo` field is entirely optional, I opted not to create V2 in/out messages and instead just add it to the existing V1 version. This shouldn't cause any problems with older juju clients/servers as this field will just be ignored.

The remainder of the work focuses around cleaning up `params.Errors` and the `ServerError` implementation in the apiserver/common packages.

### DIFF
--- a/apiserver/params/apierror.go
+++ b/apiserver/params/apierror.go
@@ -4,7 +4,9 @@
 package params
 
 import (
+	"encoding/json"
 	"fmt"
+	"reflect"
 
 	"github.com/juju/errors"
 	"gopkg.in/macaroon.v2-unstable"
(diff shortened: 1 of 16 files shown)


achilleasa:
@jameinel @howbazaar Can you guys take a look at this and let me know what you think?

apiserver/params/apierror.go
```diff
@@ -18,30 +20,9 @@ var MigrationInProgressError = errors.New(CodeMigrationInProgress)
...
-	// error code.
-	Macaroon *macaroon.Macaroon `json:"macaroon,omitempty"`
-
-	// MacaroonPath holds the URL path to be associated
-	// with the macaroon. The macaroon is potentially
-	// valid for all URLs under the given path.
-	// If it is empty, the macaroon will be associated with
-	// the original URL from which the error was returned.
-	MacaroonPath string `json:"macaroon-path,omitempty"`
+	Message string                 `json:"message"`
```
howbazaar: One of my initial concerns was how was this change going to impact the wire protocol.
From the look of this the existing expectations of non-juju clients (i.e. python-libjuju) is that the current implementation for deserializing the info portion will still match for the current implementation. The interesting part will be when we add an additional error info payload, like for the migrations. 
When we do this work, I'd like to see python-libjuju updated to handle the redirects. 

I have no concerns at this stage with how the changes are being sent.

apiserver/params/apierror.go
```diff
@@ -52,12 +33,71 @@ func (e Error) ErrorCode() string {
...
+	// valid for all URLs under the given path.
+	// If it is empty, the macaroon will be associated with
+	// the original URL from which the error was returned.
+	MacaroonPath string `json:"macaroon-path,omitempty"`
+}
+
+// AsMap encodes the error info as a map that can be attached to an Error.
+func (e DischargeRequiredErrorInfo) AsMap() map[string]interface{} {
+	out := make(map[string]interface{})
+	if e.Macaroon != nil {
```
howbazaar: Hmm... my primary concern here is the duplication of the map keys with the serialization directives.

Given that this isn't a frequent code path, I think I'd be happier if we did the same trick as above, serialize the struct to json, then back into a map. That way we could write a generic function for it and use it for the redirect type too when it comes, but also there is a nice symetry with how we handle the conversions.

rpc/rpc_test.go
```diff
@@ -572,13 +572,13 @@ func (root *Root) assertServerNotified(c *gc.C, p testCallParams, requestId uint
 		c.Assert(serverReply.body, gc.Equals, stringVal{p.request().Action + " ret"})
 	}
 	if p.retErr && p.testErr {
-		c.Assert(serverReply.hdr, gc.Equals, rpc.Header{
+		c.Assert(serverReply.hdr, gc.DeepEquals, rpc.Header{
```
howbazaar: The jc.DeepEquals gives nicer error output I believe.

apiserver/params/apierror.go
achilleasa: I opted for duplication vs json marshal/unmarshal to avoid masking json-related errors which might introduce subtle errors if for example someone decides to add a field that implements `MarshalJSON`. I could change the signature to return the error but given the way it is used in tests it would be a pain to refactor.

If we don't anticipate this to be an issue I can just change the implementation as you recommend. Thoughts?

apiserver/params/apierror.go
achilleasa: AFAICT from a quick look, python-libjuju [does not](https://github.com/juju/python-libjuju/blob/1ce8e0b5a650898b6df65cbcaaf18f532ee50456/juju/errors.py#L18-L22) currently process the `info` field. This is a bit confusing to me as `info` is only currently used for `DischargeRequiredError` and I would expect to see this in the python code (although there is a httpbakery.DischargeError in the [tests](https://github.com/juju/python-libjuju/blob/1ce8e0b5a650898b6df65cbcaaf18f532ee50456/tests/integration/test_macaroon_auth.py#L83)) 

Also, the [JujuRedirectException](https://github.com/juju/python-libjuju/blob/1ce8e0b5a650898b6df65cbcaaf18f532ee50456/juju/errors.py#L34) is currently built manually by making a second API call (works in JAAS but not implemented in upstream). This matches the implementation in juju cli. My next PR will allow this to happen as part of the login call so I can also open a PR against python-libjuju to ensure that it will work as expected.

rpc/rpc_test.go
achilleasa: Ah, didn't know that was a thing! I will push a commit to switch to that.

achilleasa:
!!build!!

apiserver/params/apierror.go
howbazaar: That would be awesome. There is an outstanding issue with python-libjuju around the discharge of macaroons.

https://github.com/juju/python-libjuju/pull/222

apiserver/params/apierror.go
howbazaar: In other places where we don't expect any error, and so don't have an error return, but internally there could be one (in theory), what we do is have something like:

`logger.Criticalf("programming error: %v", err)`

I do think that dancing through the JSON marshalling would be better here.

achilleasa:
$$merge$$

PR #10012, opened 2019-04-05
Disable replicaset config on k8s

wallyworld:
## Description of change

When a k8s controller pod restarts, it gets a new IP address. This breaks mongo replicaset config.
For now we disable mongo replicasets on k8s controllers, as HA is not yet supported anyway.

As a driveby remove an unused SupportsSpaces config item from the peergrouper worker.

## QA steps

bootstrap k8s controller
upgrade

bootstrap lxd
enable-ha

### DIFF
--- a/worker/peergrouper/worker.go
+++ b/worker/peergrouper/worker.go
@@ -125,10 +125,12 @@ type Config struct {
 	APIHostPortsSetter APIHostPortsSetter
 	MongoSession       MongoSession
 	Clock              clock.Clock
-	SupportsSpaces     bool
 	MongoPort          int
 	APIPort            int
 	ControllerAPIPort  int
+
+	// Kubernetes controllers do not support HA yet.
+	SupportsHA bool
 
 	// Hub is the central hub of the apiserver,
 	// and is used to publish the details of the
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -161,7 +161,7 @@ operator-image:
 push-operator-image: operator-image
 	docker push ${OPERATOR_IMAGE_PATH}
 
-microk8s-operator-update: operator-image
+microk8s-operator-update: install operator-image
```
ycliuhw: i think it's always already running install now by default coz  `OPERATOR_IMAGE_BUILD_SRC ?= true
`

worker/peergrouper/worker.go
```diff
@@ -591,7 +593,7 @@ func (w *pgWorker) updateReplicaSet() (map[string]*replicaset.Member, error) {
 	if err := setHasVote(added, true); err != nil {
 		return nil, errors.Annotate(err, "adding new voters")
 	}
-	if desired.isChanged {
+	if w.config.SupportsHA && desired.isChanged {
```
ycliuhw: It's probably good to have a comment to explain why we disable this for Non HA controller.

Makefile
wallyworld: ah right

worker/peergrouper/worker.go
wallyworld: ok

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10013, opened 2019-04-05
Add user messaging to cloud commands run locally

wallyworld:
## Description of change

When running commands like list-clouds, if there are no controllers at all that have been set up, operate on the local cache with suitable user facing messages. This allows for example a user to install juju and run list-clouds to see what they can bootstrap to.

## QA steps

Run the various cloud commands with/without controllers running.

### DIFF
--- a/cmd/juju/caas/remove.go
+++ b/cmd/juju/caas/remove.go
@@ -97,8 +97,8 @@ func (c *RemoveCAASCommand) Init(args []string) (err error) {
 // Run is defined on the Command interface.
 func (c *RemoveCAASCommand) Run(ctxt *cmd.Context) error {
 	if c.controllerName == "" && !c.Local {
-		ctxt.Stdout.Write(
-			[]byte("There are no controllers running.\nRemoving cluster from local cache. You will no longer be able to bootstrap on this cluster.\n"))
+		return errors.Errorf(
+			"There are no controllers running.\nTo remove cloud %q from the local cache, use the --local option.", c.cloudName)
 	}
 	if err := removeCloudFromLocal(c.cloudMetadataStore, c.cloudName); err != nil {
 		return errors.Annotatef(err, "cannot remove cloud from local cache")
--- a/cmd/juju/caas/remove.go
+++ b/cmd/juju/caas/remove.go
@@ -88,7 +88,7 @@ func (c *RemoveCAASCommand) Init(args []string) (err error) {
 	}
 	c.cloudName = args[0]
 	c.controllerName, err = c.ControllerNameFromArg()
-	if err != nil {
+	if err != nil && errors.Cause(err) != modelcmd.ErrNoControllersDefined {
 		return errors.Trace(err)
 	}
 	return cmd.CheckEmpty(args[1:])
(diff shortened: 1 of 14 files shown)


wallyworld:
Ok

cmd/juju/caas/remove.go
```diff
@@ -88,14 +88,18 @@ func (c *RemoveCAASCommand) Init(args []string) (err error) {
...
 		return errors.Trace(err)
 	}
 	return cmd.CheckEmpty(args[1:])
 }
 
 // Run is defined on the Command interface.
 func (c *RemoveCAASCommand) Run(ctxt *cmd.Context) error {
+	if c.controllerName == "" && !c.Local {
+		ctxt.Stdout.Write(
+			[]byte("There are no controllers running.\nRemoving cluster from local cache. You will no longer be able to bootstrap on this cluster.\n"))
```
howbazaar: Perhaps confirmation if it is removing locally without them explicitly saying?

cmd/juju/cloud/add.go
```diff
@@ -297,6 +297,10 @@ func (c *AddCloudCommand) Run(ctxt *cmd.Context) error {
 	}
 
 	if c.controllerName == "" {
+		if !c.Local {
+			ctxt.Stdout.Write(
+				[]byte("There are no controllers running.\nAdding cloud to local cache so you can use it to bootstrap a controller.\n"))
+		}
 		return addLocalCloud(c.cloudMetadataStore, *newCloud)
```
howbazaar: Can you also add general confirmation as to what is going on for the normal cases?
Some confirmation type line output with the context Info method (so could be hidden with --quiet).

Cloud 'my-k8s' added to controller 'local'.

cmd/juju/caas/remove.go
wallyworld: returned an error instead with a suitable message.
the use needs to type --local anyway so better to retain just that option rather than adding a --yes as well

cmd/juju/cloud/add.go
wallyworld: added for add/update/list

cmd/juju/cloud/add.go
```diff
@@ -298,8 +298,8 @@ func (c *AddCloudCommand) Run(ctxt *cmd.Context) error {
 
 	if c.controllerName == "" {
 		if !c.Local {
-			ctxt.Stdout.Write(
-				[]byte("There are no controllers running.\nAdding cloud to local cache so you can use it to bootstrap a controller.\n"))
+			ctxt.Infof(
```
howbazaar: I was happy with the initial message being written directly to stdout. I ummed and aahed when I first reviewed as to whether or not this should be written directly or using the Infof function. In the end I settled on writing directly because this part of the message we don't want hidden if they use --quiet as it actually impacts behaviour.

cmd/juju/cloud/list.go
```diff
@@ -163,8 +163,12 @@ func (c *listCloudsCommand) Run(ctxt *cmd.Context) error {
...
 	default:
 		if c.controllerName == "" && !c.Local {
-			ctxt.Stdout.Write(
-				[]byte("There are no controllers running.\nYou can bootstrap a new controller using one of these clouds:\n"))
+			ctxt.Infof(
+				"There are no controllers running.\nYou can bootstrap a new controller using one of these clouds:\n")
+		}
+		if c.controllerName != "" {
+			ctxt.Infof(
+				"Clouds on controller %q:\n\n", c.controllerName)
```
howbazaar: Nice addition

wallyworld:
$$merge$$

PR #10014, opened 2019-04-05
WIP: Part 2 of force remove application/unit - cleanup operations.

anastasiamac:
## Description of change

This is part 2 part of the implementation to allow users to force remove application or unit.

This caters for all asynchronous calls that are started up as part of the removal.

This is a WIP at the moment as it needs more comments.

### DIFF
--- a/state/cleanup.go
+++ b/state/cleanup.go
@@ -304,23 +304,21 @@ func (st *State) cleanupStorageForDyingModel(cleanupArgs []bson.Raw) (err error)
 	}
 
 	var force bool
-	switch n := len(cleanupArgs); n {
-	case 0:
-		// Old cleanups have no args, so follow the old
-		// behaviour: destroy the storage.
-	case 1:
-		if err := destroyStorageFromArg(); err != nil {
-			return err
+	// It's valid to have no args: old cleanups have no args, so follow the old behaviour.
+	if n := len(cleanupArgs); n > 0 {
+		if n > 2 {
+			return errors.Errorf("expected 0-2 arguments, got %d", n)
 		}
-	case 2:
-		if err := destroyStorageFromArg(); err != nil {
-			return err
+		if n >= 1 {
+			if err := destroyStorageFromArg(); err != nil {
+				return err
+			}
 		}
-		if err := cleanupArgs[1].Unmarshal(&force); err != nil {
-			return errors.Annotate(err, "unmarshalling cleanup arg 'force'")
+		if n >= 2 {
+			if err := cleanupArgs[1].Unmarshal(&force); err != nil {
+				return errors.Annotate(err, "unmarshalling cleanup arg 'force'")
+			}
 		}
-	default:
-		return errors.Errorf("expected 0-2 arguments, got %d", n)
 	}
 	storage, err := sb.AllStorageInstances()
 	if err != nil {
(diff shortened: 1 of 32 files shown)


anastasiamac:
[add force to cleanup operations](https://trello.com/c/wE7sx5bm/685-add-force-to-cleanup-operations)

state/unit.go
```diff
@@ -805,8 +817,23 @@ func (u *Unit) destroyHostOps(a *Application, force bool) (ops []txn.Op, errs []
 	machineUpdate := bson.D{{"$pull", bson.D{{"principals", u.doc.Name}}}}
 	var cleanupOps []txn.Op
 	if machineCheck && containerCheck {
-		machineUpdate = append(machineUpdate, bson.D{{"$set", bson.D{{"life", Dying}}}}...)
-		cleanupOps = []txn.Op{newCleanupOp(cleanupDyingMachine, m.doc.Id, force)}
+		err = m.advanceLifecycle(Dying, force)
```
babbageclunk: It's not valid to replace the ops with a call to advanceLifecycle here. The point of this code is to build a larger transaction that only advances the lifecycle if the rest of the transaction succeeds - with your change some other assertions could fail so the cleanup wouldn't be created, but the machine would still be set to dying. It makes it not an atomic transaction.

state/cleanup.go
```diff
@@ -292,28 +292,43 @@ func (st *State) cleanupStorageForDyingModel(cleanupArgs []bson.Raw) (err error)
...
-		var destroyStorageFlag bool
-		if err := cleanupArgs[0].Unmarshal(&destroyStorageFlag); err != nil {
-			return errors.Annotate(err, "unmarshalling cleanup args")
+		if err := destroyStorageFromArg(); err != nil {
+			return err
 		}
-		if !destroyStorageFlag {
-			destroyStorage = sb.ReleaseStorageInstance
+	case 2:
+		if err := destroyStorageFromArg(); err != nil {
```
babbageclunk: You could avoid duplicating the `destroyStorageFromArg` call by having case 2 first and then using `fallthrough` to process `arg[1]`

```go
case 2:
        if err := cleanupArgs[1].Unmarshal(&force); err != nil {
			return errors.Annotate(err, "unmarshalling cleanup arg 'force'")
        }
        fallthrough
case 1:
        if err := destroyStorageFromArg(); err != nil {
                return err
        }
}
```

babbageclunk: On second thought, that seems a bit gross. It would be better to just use multiple if statements:

```go
if len(cleanupArgs) >= 1 {
        if err := destroyStorageFromArg(); err != nil {
                return err
        }
}
if len(cleanupArgs) >= 2 {
        if err := cleanupArgs[1].Unmarshal(&force); err != nil {
			return errors.Annotate(err, "unmarshalling cleanup arg 'force'")
        }
}
```

state/cleanup.go
```diff
@@ -647,11 +730,15 @@ func (st *State) cleanupForceDestroyedMachine(machineId string) error {
 		return errors.Trace(err)
 	}
 	for _, unitName := range machine.doc.Principals {
-		if err := st.obliterateUnit(unitName); err != nil {
+		opErrs, err := st.obliterateUnit(unitName, true)
```
babbageclunk: `obliterateUnit` is such a good name. :)

state/unit.go
```diff
@@ -467,14 +467,24 @@ func (op *UpdateUnitOperation) Done(err error) error {
...
+// ignores errors.
+func (u *Unit) DestroyWithForce(force bool) (errs []error, err error) {
 	defer func() {
 		if err == nil {
 			// This is a white lie; the document might actually be removed.
 			u.doc.Life = Dying
 		}
 	}()
-	return u.st.ApplyOperation(u.DestroyOperation())
+	op := u.DestroyOperation()
```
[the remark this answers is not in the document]
anastasiamac: For 'force' to b a parameter to u.DestroyOperation(), I'd need to write another constructor  that will take it or duplicate what a constructor does and create DestroyOperation as struct by hand ignoring the constructor.. The way I have it now has less lines of code than either of these and is still readable :D

state/unit.go
anastasiamac: Ok.
There are a few other places where the effect is immediate but I am happy to adjust this to accumulate operations as before in order to land this PR. 
I do feel that what we want to have a call to advance lifecycle here as there is a bit more of nice logic performed as part of this call and we'd want consistency of operations but I am happy to come back to it later as a separate work.

state/cleanup.go
anastasiamac: true :) agreed! I like the 2nd version better - it's clean. Changed where appropriate and applicable.

state/cleanup.go
anastasiamac: It is :) and exactly what we are wishing for by the time we get here - 'just get rid of it already!' :D

state/unit.go
```diff
@@ -817,23 +817,12 @@ func (u *Unit) destroyHostOps(a *Application, force bool) (ops []txn.Op, errs []
 	machineUpdate := bson.D{{"$pull", bson.D{{"principals", u.doc.Name}}}}
 	var cleanupOps []txn.Op
 	if machineCheck && containerCheck {
-		err = m.advanceLifecycle(Dying, force)
-		if err != nil {
-			// TODO (anastasiamac) need to do more than log
-			logger.Warningf("could not advence machine %v state to Dying: %v", m.Id(), err)
-		}
-		if force {
+		machineUpdate = append(machineUpdate, bson.D{{"$set", bson.D{{"life", Dying}}}}...)
```
babbageclunk: Thanks, that looks better.

anastasiamac:
unreated failure- FAIL: client_macaroon_test.go:55: clientMacaroonSuite.TestAddLocalCharmSuccess

anastasiamac:
$$merge$$

PR #10019, opened 2019-04-09
Add support for upgrading k8s operators

wallyworld:
## Description of change

k8s application operators watch for agent version changes in their model and react. The existing k8s upgrader worker was changed to support this. A new API facade is introduced for the worker to call to request an upgrade instead of doing the upgrade inside the worker. This allows the upgrader worker to run in both the controller and the operator - the upgrade request to the controller then means that only the controller talks to the cluster.

## QA steps

This can't work properly until the controller API address which is published is changed.
For now, ensure that controller upgrades work as before and that operator agents log that the new worker is started and running and watcher for agent upgrades.

### DIFF
--- a/cmd/juju/commands/upgradecontroller.go
+++ b/cmd/juju/commands/upgradecontroller.go
@@ -186,7 +186,8 @@ func (c *upgradeControllerCommand) upgradeCAASController(ctx *cmd.Context) error
 		return err
 	}
 
-	context, err := c.initCAASVersions(controllerCfg, currentAgentVersion, warnCompat)
+	c.upgradeMessage = "upgrade to this version by running\n    juju upgrade-controller"
+	context, err := initCAASVersions(controllerCfg, c.Version, currentAgentVersion, warnCompat)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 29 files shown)


wallyworld:
[juju upgrade-model](https://trello.com/c/1sGpjziN/684-juju-upgrade-model)

cmd/juju/commands/upgradecontroller.go
```diff
@@ -215,17 +216,16 @@ func (c *upgradeControllerCommand) upgradeCAASController(ctx *cmd.Context) error
...
-	if c.Version == agentVersion {
+	if desiredVersion == agentVersion {
 		return nil, errUpToDate
 	}
-	c.upgradeMessage = "upgrade to this version by running\n    juju upgrade-controller"
 
 	filterVersion := jujuversion.Current
-	if c.Version != version.Zero {
-		filterVersion = c.Version
+	if desiredVersion != version.Zero {
```
ycliuhw: do we need raise errors if it's a downgrade?

wallyworld: This filter logic was copied across from the upgrade-juju command. Not sure exactly the reason for it but didn't want to mess with it as the upgrade stuff is complex and can easily break.

wallyworld:
$$merge$$

PR #10021, opened 2019-04-09
Remove Juju specific sysctl tweaks for LXD

mitechie:
Fixes #1817774

It was noted that we set a couple of sysctl values for LXD and since then LXD
has started providing their own default values. This pulls that from our
config.


mitechie:
This is a backport already approved and used in 2.6 (develop) self reviewing/landing. 

$$merge$$

PR #10027, opened 2019-04-10
Convert some destruction functions to use Operation.

anastasiamac:
## Description of change

Juju state has the ability to specify Operation (collection of database operations) to be applied at a later stage.
This is interesting and is important for the 'force' functionality since, under 'force', we want to let users know what operational errors where encountered. At the same time, we want to react or stop processing for fatal errors, say when we are applying these Operations to the database.

These PR converts the remaining destruction/removal related functions that can be 'forced' to use Operation: Relation.Destroy, applicationOffers.Remove, RemoteApplication.Destroy, RelationUnit.LeaveScope, Unit.Remove.

This conversion greatly simplifies errors collection. As a drive-by, addressed some styling warnings like using variable names that conflict with imported package names.

### DIFF
--- a/apiserver/facades/client/application/application_unit_test.go
+++ b/apiserver/facades/client/application/application_unit_test.go
@@ -507,7 +507,7 @@ func (s *ApplicationSuite) assertDestroyApplication(c *gc.C, force bool) {
 		"UnitStorageAttachments",
 		"ApplyOperation",
 	)
-	s.backend.CheckCall(c, 7, "ApplyOperation", &state.DestroyApplicationOperation{ForcedOperation: &state.ForcedOperation{Force: force}})
+	s.backend.CheckCall(c, 7, "ApplyOperation", &state.DestroyApplicationOperation{ForcedOperation: state.ForcedOperation{Force: force}})
 }
 
 func (s *ApplicationSuite) TestDestroyApplicationDestroyStorage(c *gc.C) {
@@ -541,8 +541,7 @@ func (s *ApplicationSuite) TestDestroyApplicationDestroyStorage(c *gc.C) {
 		"ApplyOperation",
 	)
 	s.backend.CheckCall(c, 5, "ApplyOperation", &state.DestroyApplicationOperation{
-		DestroyStorage:  true,
-		ForcedOperation: &state.ForcedOperation{},
+		DestroyStorage: true,
 	})
 }
 
@@ -639,10 +638,9 @@ func (s *ApplicationSuite) assertDestroyUnit(c *gc.C, force bool) {
 		"UnitStorageAttachments",
 		"ApplyOperation",
 	)
-	s.backend.CheckCall(c, 6, "ApplyOperation", &state.DestroyUnitOperation{ForcedOperation: &state.ForcedOperation{Force: force}})
+	s.backend.CheckCall(c, 6, "ApplyOperation", &state.DestroyUnitOperation{ForcedOperation: state.ForcedOperation{Force: force}})
 	s.backend.CheckCall(c, 9, "ApplyOperation", &state.DestroyUnitOperation{
-		DestroyStorage:  true,
-		ForcedOperation: &state.ForcedOperation{},
+		DestroyStorage: true,
 	})
 }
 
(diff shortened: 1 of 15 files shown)


anastasiamac:
[part of force saga:change some removals to use operation](https://trello.com/c/cKRmmGhT/687-part-of-force-sagachange-some-removals-to-use-operation)

anastasiamac:
unrelated peergrouper failure github.com/juju/juju/worker/peergrouper.TestPackage

!!build!!

apiserver/facades/client/application/application_unit_test.go
```diff
@@ -507,7 +507,7 @@ func (s *ApplicationSuite) assertDestroyApplication(c *gc.C, force bool) {
 		"UnitStorageAttachments",
 		"ApplyOperation",
 	)
-	s.backend.CheckCall(c, 7, "ApplyOperation", &state.DestroyApplicationOperation{Force: force})
+	s.backend.CheckCall(c, 7, "ApplyOperation", &state.DestroyApplicationOperation{ForcedOperation: &state.ForcedOperation{Force: force}})
```
babbageclunk: Is this right? I thought you changed it to not be a pointer. Maybe it works because of the way `.CheckCall` compares the call structs?

state/application.go
```diff
@@ -184,17 +184,17 @@ func (a *Application) SetAgentVersion(v version.Binary) (err error) {
 	if err = checkVersionValidity(v); err != nil {
 		return errors.Trace(err)
 	}
-	tools := &tools.Tools{Version: v}
+	result := &tools.Tools{Version: v}
```
babbageclunk: This confuses me - I'd expect result to be the name of the return value of the function.

state/application.go
```diff
@@ -332,8 +316,7 @@ func (op *DestroyApplicationOperation) destroyOps() ([]txn.Op, error) {
 		// relation as well as all operational errors encountered.
 		// If the 'force' is not set and the call came across some errors,
 		// these errors will be fatal and no operations will be returned.
-		relOps, isRemove, opErrs, err := rel.destroyOps(op.app.doc.Name, op.Force)
-		op.AddError(opErrs...)
+		relOps, isRemove, err := rel.destroyOps(op.app.doc.Name, &op.ForcedOperation)
```
babbageclunk: This is nice!

state/application.go
```diff
@@ -465,25 +447,24 @@ func removeResourcesOps(st *State, applicationID string) ([]txn.Op, error) {
...
 		Remove: true,
 	}}
 
-	errs := []error{}
 	// Remove application offers.
 	removeOfferOps, err := removeApplicationOffersOps(a.st, a.doc.Name)
 	if err != nil {
-		if !force {
-			return nil, errs, errors.Trace(err)
+		if !op.Force {
```
babbageclunk: A further refinement (which you don't need to do now) could be adding an IsFatal method to ForcedOperation - then this could be:
```go
if op.IsFatal(err) {
    return nil, errors.Trace(err)
}
```
And IsFatal would be:
```go
func (fo *ForcedOperation) IsFatal(err error) bool {
    if err == nil {
        return false
    }
    if fo.Force {
        return true
    }
    fo.AddError(err)
    return false
}
```

state/unit.go
```diff
@@ -272,17 +272,17 @@ func (u *Unit) SetAgentVersion(v version.Binary) (err error) {
 	if err = checkVersionValidity(v); err != nil {
 		return err
 	}
-	tools := &tools.Tools{Version: v}
+	result := &tools.Tools{Version: v}
```
babbageclunk: Same here - how about `newTools` instead?

state/unit_test.go
```diff
@@ -2015,6 +2015,8 @@ func (s *UnitSuite) TestRemovePathologicalWithBuggyUniter(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 	err = mysql0ru.EnterScope(nil)
 	c.Assert(err, jc.ErrorIsNil)
+	//c.Assert(mysql.Refresh(), jc.ErrorIsNil)
```
babbageclunk: ?

babbageclunk: Oh, you removed it.

state/application.go
anastasiamac: Love it! I'll change it here in this PR

state/unit.go
anastasiamac: changing to 'versionedTool' :D

state/application.go
anastasiamac: Actually, I might land this as is and then do it on one of the follow ups.. This has been open too long for my liking \o/

anastasiamac:
$$merge$$

PR #10028, opened 2019-04-11
Surface pod provisioning errors

wallyworld:
## Description of change

The pod spec and/or config such as securityContext may not allow any workload pods to be provisioned by the statefulset or deployment controller. These errors are now surfaced in juju status.

## QA steps

bootstrap k8s and deploy a charm with a bad pod spec
check juju status shows the error

## Bug reference

https://bugs.launchpad.net/juju/+bug/1814055

### DIFF
--- a/apiserver/params/params.go
+++ b/apiserver/params/params.go
@@ -375,6 +375,7 @@ type UpdateApplicationUnitArgs struct {
 type UpdateApplicationUnits struct {
 	ApplicationTag string                  `json:"application-tag"`
 	Scale          *int                    `json:"scale,omitempty"`
+	Status         EntityStatus            `json:"status,omitempty"`
 	Units          []ApplicationUnitParams `json:"units"`
 }
 
--- a/worker/caasunitprovisioner/application_worker.go
+++ b/worker/caasunitprovisioner/application_worker.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/juju/errors"
+	"github.com/juju/juju/caas"
 	"gopkg.in/juju/names.v2"
 	"gopkg.in/juju/worker.v1"
 	"gopkg.in/juju/worker.v1/catacomb"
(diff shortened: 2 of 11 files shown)


apiserver/params/params.go
```diff
@@ -375,6 +375,9 @@ type UpdateApplicationUnitArgs struct {
 type UpdateApplicationUnits struct {
 	ApplicationTag string                  `json:"application-tag"`
 	Scale          *int                    `json:"scale,omitempty"`
+	Status         string                  `json:"status,omitempty"`
```
ycliuhw: do we wanna have a status struct field instead here?

worker/caasunitprovisioner/application_worker.go
```diff
@@ -169,15 +171,19 @@ func (aw *applicationWorker) loop() error {
...
 			}
-			if scale == nil || *scale == lastReportedScale {
+			var scale *int
+			if service != nil {
+				scale = service.Scale
+			}
+			if scale != nil && *scale == lastReportedScale {
 				continue
 			}
 			lastReportedScale = *scale
```
ycliuhw: is it possible `scale == nil` here?

apiserver/params/params.go
wallyworld: Yeah, I was being consistent with other places, but will change.

worker/caasunitprovisioner/application_worker.go
wallyworld: Ah yes it is. We (now) need to continue to send status even if scale is nil but as you point out, I need to change the lastReportedScale logic.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
[caas models - better surfacing of pod errors](https://trello.com/c/YspDN1ok/689-caas-models-better-surfacing-of-pod-errors)

PR #10029, opened 2019-04-11
Extra checks for built-in MicroK8s

Veebers:
## Description of change

For builtin MicroK8s at bootstrap check that storage is enabled and fail if not (with message).
Also make sure we don't re-query for storage details if it is already defined for the cloud.
## QA steps

unit tests cover the changes.

## Documentation changes

n/a
## Bug reference

n/a

### DIFF
--- a/caas/kubernetes/provider/cloud.go
+++ b/caas/kubernetes/provider/cloud.go
@@ -240,10 +240,8 @@ func (p kubernetesEnvironProvider) FinalizeCloud(ctx environs.FinalizeCloudConte
 	}
 
 	// if storage is already defined there is no need to query the cluster
-	if opStorage, ok := cld.Config[OperatorStorageKey]; ok {
-		if opStorage != nil && opStorage != "" {
-			return cld, nil
-		}
+	if opStorage, ok := cld.Config[OperatorStorageKey]; ok && opStorage != "" {
+		return cld, nil
 	}
 
 	// Need the credentials, need to query for those details
@@ -321,6 +319,5 @@ func microK8sStatus(cmdRunner CommandRunner) (microk8sStatus, error) {
 }
 
 type microk8sStatus struct {
-	Microk8s map[string]string `yaml:"microk8s"`
-	Addons   map[string]string `yaml:"addons"`
+	Addons map[string]string `yaml:"addons"`
 }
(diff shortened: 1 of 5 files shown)


caas/kubernetes/provider/cloud.go
```diff
@@ -227,10 +229,17 @@ func BaseKubeCloudOpenParams(cloud jujucloud.Cloud, credential jujucloud.Credent
 
 // FinalizeCloud is part of the environs.CloudFinalizer interface.
 func (p kubernetesEnvironProvider) FinalizeCloud(ctx environs.FinalizeCloudContext, cld cloud.Cloud) (cloud.Cloud, error) {
-	cloudName := cld.Name
-	if cloudName != caas.K8sCloudMicrok8s {
+	if cld.Name != caas.K8sCloudMicrok8s {
 		return cld, nil
 	}
+
+	if cld.Config["operator-storage"] != nil && cld.Config["workload-storage"] != nil {
```
wallyworld: There are consts for these somewhere I'm sure

caas/kubernetes/provider/cloud.go
```diff
@@ -227,10 +229,17 @@ func BaseKubeCloudOpenParams(cloud jujucloud.Cloud, credential jujucloud.Credent
...
+		return cld, nil
+	}
+
+	if !microK8sStorageEnabled(p.cmdRunner) {
```
wallyworld: They need "dns" enabled too, would be awesome if we could check for that as well.
Also, these dns and storage enabled checks need to be before the cld.Config check since even if the cloud is configured etc, it will fail without storage enabled.

caas/kubernetes/provider/cloud.go
```diff
@@ -227,10 +229,23 @@ func BaseKubeCloudOpenParams(cloud jujucloud.Cloud, credential jujucloud.Credent
...
 		return cld, nil
 	}
+
+	if err := ensureMicroK8sSuitable(p.cmdRunner); err != nil {
+		return cld, errors.Trace(err)
+	}
+
+	// if storage is already defined there is no need to query the cluster
+	if opStorage, ok := cld.Config[OperatorStorageKey]; ok {
+		if opStorage != nil && opStorage != "" {
```
wallyworld: don't need the != nil, and can inline the != "" check

caas/kubernetes/provider/cloud.go
```diff
@@ -256,13 +271,56 @@ func (p kubernetesEnvironProvider) FinalizeCloud(ctx environs.FinalizeCloudConte
...
+
+	err = yaml.Unmarshal(result.Stdout, &status)
+	if err != nil {
+		return status, errors.Trace(err)
+	}
+	return status, nil
+}
+
+type microk8sStatus struct {
+	Microk8s map[string]string `yaml:"microk8s"`
```
wallyworld: why do we have this field, appears not used?

Veebers:
$$merge$$

Veebers:
$$merge$$

PR #10031, opened 2019-04-11
move charmstoreClientToTestcharmsClientShim to export_test

hmlanigan:
## Description of change

Move charmstoreClientToTestcharmsClientShim from shim.go to export_test.go to prevent panic on bootstrap.  Global variable in testcharms pkg led to os.stat of directory which doesn't exist on a juju machine which panics on error.

## QA steps

bootstrap.

### DIFF
--- a/cmd/juju/application/export_test.go
+++ b/cmd/juju/application/export_test.go
@@ -8,6 +8,8 @@ import (
 	"github.com/juju/errors"
 	"github.com/juju/romulus"
 	"gopkg.in/juju/charmrepo.v3"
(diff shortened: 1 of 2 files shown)


mitechie:
👍 for review/QA of a successful bootstrap. My only question would be do the tests depend on this behavior so watching the test run to see if they still pass with the change.

hmlanigan:
The tests are using the test shim, which was moved to export, which they can use.

hmlanigan:
$$merge$$

PR #10033, opened 2019-04-12
Skip don't fail intermittent zip upload error.

howbazaar:
We know that there is an intermittent failure in uploading charms in the tests. Exhaustive searching has gone on by many people and no one can find it. It doesn't seem to manifest in production, and often can't be reproduced locally.

This branch acknowledges the issue while generally keeping the test for local confirmation. If we do hit this particular issue it is marked as a skip rather than a fail.

### DIFF
--- a/api/client_macaroon_test.go
+++ b/api/client_macaroon_test.go
@@ -5,6 +5,7 @@ package api_test
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


howbazaar:
$$merge$$

PR #10035, opened 2019-04-12
Improve duplicate namespace message

wallyworld:
## Description of change

When adding a k8s model and the namespace already exists, improve the user facing error message.

## QA steps

bootstrap k8s
add same model twice


wallyworld:
$$merge$$

PR #10036, opened 2019-04-12
Fix race in featuretests/api_meterstatus_test.go

howbazaar:
There is a subtle small race condition that we tend to hit on the arm64 machines, and very hard to reproduce locally. When the `MetricsManager` is created for the first time it adds the initial document in. Then the `SetLastSuccessfulSend` call updates that document. This causes the potential for two events.

The fix is to sync with the internal watchers before asking for the next one.


howbazaar:
Example failure:
```
api_meterstatus_test.go:78:
    wc.AssertOneChange()
/var/lib/jenkins/workspace/RunUnittests-arm64/_build/src/github.com/juju/juju/core/watcher/watchertest/notify.go:95:
    c.Fatalf("watcher sent unexpected change: (_, %v)", ok)
... Error: watcher sent unexpected change: (_, true)
```

howbazaar:
$$merge$$

PR #10037, opened 2019-04-12
WIP Remove charmstore dep stage 4

timClicks:
Work in progress (tests not quite passing yet)
----

## Description of change

## QA steps

## Documentation changes

## Bug reference

### DIFF
--- a/charmstore/fakeclient.go
+++ b/charmstore/fakeclient.go
@@ -349,6 +349,9 @@ func NewRepository() *Repository {
 
 func (r *Repository) addRevision(ref *charm.URL) *charm.URL {
 	revision := r.revisions[r.channel][*ref]
+	if revision == 0 {
+		revision = 1
+	}
 	return ref.WithRevision(revision)
 }
 
@@ -401,6 +404,8 @@ func (r Repository) CharmInfo(charmURL string) (*charms.CharmInfo, error) {
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
+	logger.Infof("getting charm info for %+v (parsed as %v)", charmURL, charmId)
+	logger.Infof("uploaded charms: %+v", r.channel, r.charms)
 	charmDetails, err := r.Get(charmId)
 	if err != nil {
 		return nil, errors.Trace(err)
@@ -421,7 +426,7 @@ func (r Repository) CharmInfo(charmURL string) (*charms.CharmInfo, error) {
 //
 // Part of the charmrepo.Interface
 func (r Repository) Resolve(ref *charm.URL) (canonRef *charm.URL, supportedSeries []string, err error) {
-	return r.addRevision(ref), []string{"trusty", "wily", "quantal"}, nil
+	return r.addRevision(ref), []string{"trusty", "wily", "quantal", "kubernetes"}, nil
 }
 
 // ResolveWithChannel disambiguates a charm to a specific revision.
--- a/charmstore/fakeclient.go
+++ b/charmstore/fakeclient.go
@@ -347,10 +347,14 @@ func NewRepository() *Repository {
 	return &repo
 }
 
-func (r *Repository) addRevision(ref *charm.URL) *charm.URL {
+func (r *Repository) fillCharmDetails(ref *charm.URL) *charm.URL {
+	logger.Infof("filling details for %+V", ref)
 	revision := r.revisions[r.channel][*ref]
 	if revision == 0 {
 		revision = 1
+	}
+	if ref.User == "" && ref.Schema != "local" {
+		ref.User = "who"
 	}
 	return ref.WithRevision(revision)
 }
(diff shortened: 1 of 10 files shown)


charmstore/fakeclient.go
```diff
@@ -349,6 +349,9 @@ func NewRepository() *Repository {
 
 func (r *Repository) addRevision(ref *charm.URL) *charm.URL {
 	revision := r.revisions[r.channel][*ref]
+	if revision == 0 {
```
wallyworld: Shouldn't need to do this...

charmstore/fakeclient.go
```diff
@@ -421,7 +426,7 @@ func (r Repository) CharmInfo(charmURL string) (*charms.CharmInfo, error) {
 //
 // Part of the charmrepo.Interface
 func (r Repository) Resolve(ref *charm.URL) (canonRef *charm.URL, supportedSeries []string, err error) {
-	return r.addRevision(ref), []string{"trusty", "wily", "quantal"}, nil
+	return r.addRevision(ref), []string{"trusty", "wily", "quantal", "kubernetes"}, nil
```
wallyworld: Can't have charms operating as both k8s and non k8s, or WIndows and Ubuntu etc
All series for a given charm need to be the same OS

cmd/juju/application/deploy.go
```diff
@@ -88,13 +90,15 @@ type CharmDeployAPI interface {
 	CharmInfo(string) (*apicharms.CharmInfo, error)
 }
 
+type charmDeployClient struct {
```
wallyworld: This isn't used anywhere?

cmd/juju/application/deploy.go
```diff
@@ -137,11 +137,17 @@ type modelConfigClient struct {
 	*modelconfig.Client
 }
 
+type addCharmClient struct {
```
wallyworld: This looks like only used in export_test, so move it there. Is it needed?

cmd/juju/application/deploy_test.go
```diff
@@ -706,7 +708,9 @@ func (s *CAASDeploySuite) TestDevices(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 
 	_, ch := testcharms.UploadCharmWithSeries(c, s.client, "kubernetes/bitcoin-miner-1", "bitcoin-miner", "kubernetes")
-	err = runDeploy(c, "bitcoin-miner", "-m", m.Name(), "--device", "bitcoinminer=10,nvidia.com/gpu", "--series", "kubernetes")
+
+	logger.Infof("charmrepo state %#v", s.charmrepo)
```
wallyworld: Debugf

cmd/juju/application/deploy_test.go
```diff
@@ -1021,6 +1018,34 @@ func (s *charmstoreSuite) assertDeployedApplicationBindings(c *gc.C, info map[st
 	}
 }
 
+// assertRelationsEstablished checks that the given relations have been set.
+func (s *charmstoreSuite) assertRelationsEstablished(c *gc.C, relations ...string) {
```
wallyworld: Just check we need both of these new functions - we need one now, maybe we will need both eventually?

cmd/juju/application/deploy_test.go
```diff
@@ -1034,140 +1059,16 @@ func (s *charmstoreSuite) runDeploy(c *gc.C, args ...string) error {
...
-
-	err = runDeploy(c, "--channel", "edge", "~client-username/wordpress")
-	c.Assert(err, gc.IsNil)
-	s.assertCharmsUploaded(c, "cs:~client-username/precise/wordpress-0")
-	s.assertApplicationsDeployed(c, map[string]applicationInfo{
-		"wordpress": {charm: "cs:~client-username/precise/wordpress-0", config: ch.Config().DefaultSettings()},
-	})
-}
+// TODO
+//func (s *DeployCharmStoreSuite) TestDeployWithChannel(c *gc.C) {
```
wallyworld: Let's look to re-implement this as discussed live

timClicks:
This all sounds sensible

cmd/juju/application/deploy.go
timClicks: Was following the pattern that's already introduced in that file with everything being wrapped in a local `struct`. Probably isn't required,

cmd/juju/application/deploy_test.go
timClicks: These were copied verbatim from `legacyCharmStoreSuite`. They're there to maintain current test behaviour.

cmd/juju/application/deploy_test.go
timClicks: 100%

nammn:
How do we want to proceed with this?

PR #10039, opened 2019-04-12
enhancements on add-k8s

ycliuhw:
## Description of change

error message enhancements on `add-k8s`;
removed CDK as a k8s cloud in Juju because we only need to track the base cloud that CDK is running on;

## QA steps

1. run add-k8s for a non-existing/existing sc

```bash
$ cat ~/.kube/config | juju add-k8s cdk-aws1  --debug --storage k-sc --region ec2/ap-southeast-2 --local
```

## Documentation changes

None

## Bug reference

None


cmd/juju/caas/add.go
```diff
@@ -499,7 +499,7 @@ func (c *AddCAASCommand) validateCloudRegion(cloudRegion string) (_ string, err
 }
 
 func (c *AddCAASCommand) getClusterMetadataFunc(ctx *cmd.Context) provider.GetClusterMetadataFunc {
-	return func(broker caas.ClusterMetadataChecker) (*caas.ClusterMetadata, error) {
+	return func(storageClassName string, broker caas.ClusterMetadataChecker) (*caas.ClusterMetadata, error) {
```
wallyworld: This doesn't quite look right - we pass in storageClassName here but then below use c.workloadStorage and ignore the function param.
Let's discuss.

ycliuhw:
$$merge$$

ycliuhw:
$$merge$$

PR #10043, opened 2019-04-15
Fix data race in raftbackstop tests

babbageclunk:
## Description of change

TestRecoversClusterNoVoters was racing because mockConfigFuture.Configuration() was returning a raft.Configuration that shared the server slice with the original that was being changed in BootstrapCluster. Share the mutex with the mockConfigFuture so we can protect calling configuration.Clone() to return a proper copy.

## QA steps

Original test fails under stress-race on first try, changed version passes for as long as I could stand waiting (about 30 runs).

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/worker/raft/raftbackstop/worker_test.go
+++ b/worker/raft/raftbackstop/worker_test.go
@@ -404,6 +404,7 @@ func (r *mockRaft) setValues(state raft.RaftState, cf *mockConfigFuture) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	r.state = state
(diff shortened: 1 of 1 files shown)


babbageclunk:
Example failure here: https://jenkins.juju.canonical.com/view/Unit%20tests/job/RunUnittests-race-amd64/1379/testReport/github/com_juju_juju_worker_raft_raftbackstop/TestPackage/

babbageclunk:
$$merge$$

PR #10044, opened 2019-04-15
MAAS display name falls back to FQDN when hostname unavailable.

timClicks:
## Description of change

MAAS nodes sometimes can sometimes fail to return a hostname. This leads to their instance names being rendered with their (ugly) instance ID. The suggestion is that we fall back to the fully-qualified domain name first, before relying on the instance ID. This commit implements that suggestion.

## QA steps

TODO 

## Documentation changes

Perhaps add a note on how MAAS machines may change their display names in status. 

## Bug reference

- [lp#1825068](https://bugs.launchpad.net/juju/+bug/1825068)
- indirectly related to [lp#1792210](https://bugs.launchpad.net/juju/+bug/1792210)

### DIFF
--- a/provider/maas/instance.go
+++ b/provider/maas/instance.go
@@ -23,6 +23,7 @@ import (
 type maasInstance interface {
 	instances.Instance
 	zone() (string, error)
+	displayName() (string, error)
 	hostname() (string, error)
 	hardwareCharacteristics() (*instance.HardwareCharacteristics, error)
 	volumes(names.MachineTag, []names.VolumeTag) ([]storage.Volume, []storage.VolumeAttachment, error)
@@ -261,4 +262,12 @@ func (mi *maas1Instance) hardwareCharacteristics() (*instance.HardwareCharacteri
 func (mi *maas1Instance) hostname() (string, error) {
 	// A MAAS instance has its DNS name immediately.
 	return mi.maasObject.GetField("hostname")
+}
+
+func (mi *maas1Instance) displayName() (string, error) {
+	displayName, err := mi.hostname()
+	if err != nil {
+		logger.Infof("error detecting hostname for %s", mi)
+	}
+	return displayName, err
 }
--- a/provider/maas/instance.go
+++ b/provider/maas/instance.go
@@ -264,10 +264,20 @@ func (mi *maas1Instance) hostname() (string, error) {
 	return mi.maasObject.GetField("hostname")
 }
 
+func (mi *maas1Instance) fqdn() (string, error) {
+	return mi.maasObject.GetField("fqdn")
+}
+
 func (mi *maas1Instance) displayName() (string, error) {
 	displayName, err := mi.hostname()
 	if err != nil {
 		logger.Infof("error detecting hostname for %s", mi)
+	}
+	if displayName == "" {
+		displayName, err = mi.fqdn()
+	}
+	if err != nil {
+		logger.Infof("error detecting fqdn for %s", mi)
 	}
 	return displayName, err
 }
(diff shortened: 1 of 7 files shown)


provider/maas/instance.go
```diff
@@ -262,3 +263,11 @@ func (mi *maas1Instance) hostname() (string, error) {
 	// A MAAS instance has its DNS name immediately.
 	return mi.maasObject.GetField("hostname")
 }
+
+func (mi *maas1Instance) displayName() (string, error) {
+	displayName, err := mi.hostname()
```
wallyworld: We need to fall back to fqdn here as well.
Add a new method fqdn() to maas1Instance to get the "fqdn" attribute

provider/maas/maas2instance.go
```diff
@@ -47,8 +47,20 @@ func (mi *maas2Instance) hardwareCharacteristics() (*instance.HardwareCharacteri
 	return hc, nil
 }
 
+func (mi *maas2Instance) displayName() (string, error) {
+	hostname, err := mi.hostname()
```
wallyworld: use mi.machine.Hostname() to avoid the error check

provider/maas/maas2instance.go
```diff
@@ -47,8 +47,20 @@ func (mi *maas2Instance) hardwareCharacteristics() (*instance.HardwareCharacteri
...
+	}
+	if hostname != "" {
+		return hostname, nil
+	}
+	return mi.machine.FQDN(), nil
+}
+
 func (mi *maas2Instance) String() string {
-	return fmt.Sprintf("%s:%s", mi.machine.Hostname(), mi.machine.SystemID())
+	hostname, _ := mi.hostname()
```
wallyworld: why this change?

provider/maas/maas2instance_test.go
```diff
@@ -113,6 +113,22 @@ func (s *maas2InstanceSuite) TestHostname(c *gc.C) {
...
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(result, gc.Equals, "saul-goodman")
+}
+
+func (s *maas2InstanceSuite) TestDisplayNameFallsBackToFQDN(c *gc.C) {
+	machine := newFakeMachine("abc123", "amd64", "ok")
+	thing := &maas2Instance{machine: machine}
+	result, err := thing.displayName()
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(result, gc.Equals, thing.machine.FQDN()) //TODO(tsm) fix to use non-empty string
```
wallyworld: I don't understand the TODO??

timClicks: Oh, sorry this was a note to self that I've already fixed.

provider/maas/maas2instance.go
timClicks: Removes code duplication, but isn't necessary at all.

wallyworld: you don't need to ignore the error here.
just revert the change

provider/maas/maas2instance_test.go
wallyworld: remove the todo

timClicks:
> <wallyworld> timClicks: i'll look at gce pr now, maas one can land

$$merge$$

provider/maas/instance.go
```diff
@@ -262,3 +263,21 @@ func (mi *maas1Instance) hostname() (string, error) {
...
+	return mi.maasObject.GetField("fqdn")
+}
+
+func (mi *maas1Instance) displayName() (string, error) {
+	displayName, err := mi.hostname()
+	if err != nil {
+		logger.Infof("error detecting hostname for %s", mi)
+	}
+	if displayName == "" {
+		displayName, err = mi.fqdn()
```
jameinel: Isn't fqdn just fmt.Sprintf("%s.%s", hostname, domainname) ?

how is it that we we would have an FQDN but not a hostname?

howbazaar: Older versions of MAAS provided fqdn but not hostname. Like 2.3 - which much of bootstack is using.

PR #10046, opened 2019-04-15
Fix intermittent test failure in TestUnitRelations.

howbazaar:
On slow substrate machines like the arm64 or s390x build machines we'd intermittently see a test failure.
This was due to the uniter being killed before the init function was finished.

The error returns from init were not sufficiently handled for the various error types. During the discovery of this, I found another issue with CAAS uniter instances where the intent is to have a nil error return from the main loop if the unit becomes dead or is dead. However this was also returned from `init`, which the caller took as all good, and would cause a panic when a method on the operatorExecutor was called.
To address this we introduce a particular error type for the CAAS unit being dead, and handle that error type explicitly in the response places.

This lead to yet another issue where we were checking the model type before it had been initialized in the init function. To resolve this we pass the model type into the uniter as one of the UniterParams as when we start the uniter, we know what the model type is, so don't bother asking the model.

### DIFF
--- a/worker/caasoperator/caasoperator.go
+++ b/worker/caasoperator/caasoperator.go
@@ -24,6 +24,7 @@ import (
 	apiuniter "github.com/juju/juju/api/uniter"
 	"github.com/juju/juju/core/leadership"
 	"github.com/juju/juju/core/life"
+	"github.com/juju/juju/core/model"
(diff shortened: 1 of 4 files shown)


howbazaar:
$$merge$$

PR #10051, opened 2019-04-16
Add base-image-path model config attribute to gce provider

timClicks:
## Description of change

This functionality has been requested to facilitate nested virtualisation capabilities of Google Compute Engine (GCE).

> Juju can target GCP[sic] to stand up services, however, the images don't allow nested virtualisation.
>
> Google offers this by creating a copy of the disk image and applying a Google license tag to it: https://cloud.google.com/compute/docs/instances/enable-nested-virtualization-vm-instances
>
> [...]
>> If you did go and create your own images, I do not know if said images would still be available from the base path "projects/ubuntu-os-cloud/global/images/"? I suspect we'd also need to **introduce a model config item** to allow the base image path to be specified.

## QA steps

Default behaviour:

* Bootstrap to GCE
* Running `juju model-config base-image-path` prints out an empty line

Invalid Path

* `juju bootstrap --config base-image-path=/tmp/ google g1`
```
Creating Juju controller "g1" on google/us-east1
Looking for packaged Juju agent version 2.6-beta2 for amd64
No packaged binary found, preparing local Juju agent binary
Launching controller instance(s) on google/us-east1...
ERROR failed to bootstrap model: cannot start bootstrap instance in any availability zone (us-east1-b, us-east1-c, us-east1-d)
```

**Future improvement**: verify that `relative-image-path` exists at the start of the bootstrap command.

## Documentation changes

End-user documentation is required.

We need to explain that the `base-image-path` is used in conjunction with `default-image-id`.

Additionally, please note that when `base-image-path` and `image-stream` are specified `image-stream` is ignored.

## Bug reference

* [lp#1814633](https://bugs.launchpad.net/juju/+bug/1814633)

### DIFF
--- a/provider/gce/config.go
+++ b/provider/gce/config.go
@@ -11,17 +11,16 @@ import (
 	"github.com/juju/juju/environs/config"
 )
 
-// TODO(ericsnow) While not strictly config-related, we could use some
-// mechanism by which we can validate the values we've hard-coded in
-// this provider match up with the external authoritative sources. One
-// example of this is the data stored in instancetypes.go. Similarly
-// we should also ensure the cloud-images metadata is correct and
-// up-to-date, though that is more the responsibility of that team.
-// Regardless, it may be useful to include a tool somewhere in juju
-// that we can use to validate this provider's potentially out-of-date
-// data.
+const (
+	cfgBaseImagePath = "base-image-path"
+)
 
-var configSchema = environschema.Fields{}
+var configSchema = environschema.Fields{
+	cfgBaseImagePath: {
+		Description: "Base path to look for machine disk images.",
+		Type:        environschema.Tstring,
+	},
+}
 
 // configFields is the spec for each GCE config value's type.
 var configFields = func() schema.Fields {
@@ -34,7 +33,9 @@ var configFields = func() schema.Fields {
 
 var configImmutableFields = []string{}
 
-var configDefaults = schema.Defaults{}
+var configDefaults = schema.Defaults{
+	cfgBaseImagePath: schema.Omit,
+}
 
 type environConfig struct {
 	config *config.Config
@@ -78,4 +79,9 @@ func newConfig(cfg, old *config.Config) (*environConfig, error) {
 		attrs:  attrs,
 	}
 	return ecfg, nil
+}
+
+func (c *environConfig) baseImagePath() (string, bool) {
+	path, ok := c.attrs[cfgBaseImagePath].(string)
+	return path, ok
 }
--- a/provider/gce/environ_broker.go
+++ b/provider/gce/environ_broker.go
@@ -131,6 +131,28 @@ func (env *environ) findInstanceSpec(
 	return spec, errors.Trace(err)
 }
 
+func (env *environ) imageURLBase(os jujuos.OSType) (string, error) {
+	base, useCustomPath := env.ecfg.baseImagePath()
+	if useCustomPath {
+		return base, nil
+	}
+
+	switch os {
+	case jujuos.Ubuntu:
+		if env.Config().ImageStream() == "daily" {
+			base = ubuntuDailyImageBasePath
+		} else {
+			base = ubuntuImageBasePath
+		}
+	case jujuos.Windows:
+		base = windowsImageBasePath
+	default:
+		return "", errors.Errorf("os %s is not supported on the gce provider", os.String())
+	}
+
+	return base, nil
+}
+
 // newRawInstance is where the new physical instance is actually
 // provisioned, relative to the provided args and spec. Info for that
 // low-level instance is returned.
@@ -155,11 +177,16 @@ func (env *environ) newRawInstance(ctx context.ProviderCallContext, args environ
 		hostname,
 	}
 
+	imageURLBase, err := env.imageURLBase(os)
+	if err != nil {
+		return nil, common.ZoneIndependentError(err)
+	}
+
 	disks, err := getDisks(
 		spec, args.Constraints,
 		args.InstanceConfig.Series,
 		env.Config().UUID(),
-		env.Config().ImageStream() == "daily",
+		imageURLBase,
 	)
 	if err != nil {
 		return nil, common.ZoneIndependentError(err)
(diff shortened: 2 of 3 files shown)


provider/gce/config.go
```diff
@@ -11,17 +11,16 @@ import (
...
-// that we can use to validate this provider's potentially out-of-date
-// data.
+const (
+	RelativeImagePath = "relative-image-path"
+)
 
-var configSchema = environschema.Fields{}
+var configSchema = environschema.Fields{
+	RelativeImagePath: {
+		Description: "Base path to look for machine disk images. Setting this to a custom value in conjunction with setting the default-image-id allows you to provide custom virtual machine images.",
```
wallyworld: "default-image-id" is not an attribute that can be set for GCE providers, so I don't quite follow this text.

Perhaps all that's needed is "Base path to look for machine disk images"

provider/gce/config.go
```diff
@@ -11,17 +11,16 @@ import (
...
-// mechanism by which we can validate the values we've hard-coded in
-// this provider match up with the external authoritative sources. One
-// example of this is the data stored in instancetypes.go. Similarly
-// we should also ensure the cloud-images metadata is correct and
-// up-to-date, though that is more the responsibility of that team.
-// Regardless, it may be useful to include a tool somewhere in juju
-// that we can use to validate this provider's potentially out-of-date
-// data.
+const (
+	RelativeImagePath = "relative-image-path"
```
wallyworld: Perhaps, "base-image-path" (after reading the comments in the bug).

Also, should either unexport this const or give it a comment

timClicks: Have gone with `cfgBaseImagePath` as per the vsphere provider

provider/gce/config.go
```diff
@@ -79,3 +80,8 @@ func newConfig(cfg, old *config.Config) (*environConfig, error) {
 	}
 	return ecfg, nil
 }
+
+func (c *environConfig) baseImagePath() (path string, isSet bool) {
```
wallyworld: don't need named return values, just (string, bool)

provider/gce/config.go
```diff
@@ -11,17 +11,16 @@ import (
...
-// data.
+const (
+	cfgBaseImagePath = "base-image-path"
+)
 
-var configSchema = environschema.Fields{}
+var configSchema = environschema.Fields{
+	cfgBaseImagePath: {
+		Description: "Base path to look for machine disk images.",
+		Type:        environschema.Tstring,
```
timClicks: Some provider code uses `schema.String()`. Decided to use what openstack does.

provider/gce/environ_broker.go
```diff
@@ -231,32 +257,20 @@ func getMetadata(args environs.StartInstanceParams, os jujuos.OSType) (map[strin
...
-		}
-	case jujuos.Windows:
-		imageURL = windowsImageBasePath
-	default:
-		return nil, errors.Errorf("os %s is not supported on the gce provider", os.String())
+	if imageURLBase == "" {
+		return nil, errors.NotValidf("imageURLBase must be set")
 	}
+	imageURL := imageURLBase + spec.Image.Id
+	logger.Infof("fetching disk image from %v", imageURL)
```
timClicks: Decided to add some logging in case people ever got confused when they were trying different options.

provider/gce/environ_broker.go
```diff
@@ -131,6 +131,27 @@ func (env *environ) findInstanceSpec(
...
+
+	if os == jujuos.Ubuntu {
+		base = ubuntuImageBasePath
+	} else if os == jujuos.Windows {
+		return windowsImageBasePath, nil
+	} else {
+		return "", errors.Errorf("os %s is not supported on the gce provider", os.String())
+	}
+
+	if env.Config().ImageStream() == "daily" {
```
wallyworld: This is a bit messy - let's redo the if then else above as a switch and have all os related logic under ubuntu or windows

provider/gce/environ_broker.go
```diff
@@ -131,6 +131,27 @@ func (env *environ) findInstanceSpec(
 	return spec, errors.Trace(err)
 }
 
+func (env *environ) imageURLBase(os jujuos.OSType) (string, error) {
+	base, useCustomPath := env.ecfg.baseImagePath()
+	if useCustomPath {
+		return base, nil
```
timClicks: Am not worried about testing for the OS here, as it has already been tested within the `getMetadata()` call within `getDisks()` anyway

provider/gce/environ_broker_test.go
```diff
@@ -233,7 +234,7 @@ func (s *environBrokerSuite) TestGetDisks(c *gc.C) {
 		}
 	}
 
-	diskSpecs, err := gce.GetDisks(s.spec, s.StartInstArgs.Constraints, "trusty", "32f7d570-5bac-4b72-b169-250c24a94b2b", true)
+	diskSpecs, err := gce.GetDisks(s.spec, s.StartInstArgs.Constraints, "trusty", "32f7d570-5bac-4b72-b169-250c24a94b2b", gce.UbuntuDailyImageBasePath)
```
wallyworld: This now doesn't do anything - you need a test where the model config has stream set to "daily" and then check the expected image path

provider/gce/config.go
timClicks: Prefer to have the variables in there, so that the method is self-documenting. Even though we're using the value as-is, not using `ok` for its original purpose.

provider/gce/environ_broker_test.go
timClicks: Have added some tests. Could you please cast your eye over them and I'll fire off the merge?

provider/gce/environ_broker.go
timClicks: Fixed :hammer:

provider/gce/export_test.go
```diff
@@ -14,10 +14,12 @@ import (
 
 var (
 	Provider                 environs.EnvironProvider = providerInstance
+	NewConfig                                         = newConfig
 	NewInstance                                       = newInstance
 	CheckInstanceType                                 = checkInstanceType
 	GetMetadata                                       = getMetadata
 	GetDisks                                          = getDisks
+	CfgBaseImagePath                                  = cfgBaseImagePath
```
wallyworld: not needed?

timClicks: Good catch. This was churn from figuring out how the test suite managed config changes.

timClicks:
$$merge$$

timClicks:
[gce model config for path snippet](https://trello.com/c/ep1Tz8QT/693-gce-model-config-for-path-snippet)

PR #10052, opened 2019-04-16
'model-destroy' gains --force and max-wait options.

anastasiamac:
## Description of change

Model destroy can very soon be forced.
This PR adds --force option to the command and passes it through to api and apiserver. 
As most destructions and removals are a multi-step process, all these operations when forced will support user's ability to specify for how long each step will wait to execute cleanly. If a step waited for --max-wait duration and its following step did not kick off, Juju will force the next step.

This PR also adds --max-wait option from CLI and transports through to api and apiserver.
This required a facade version bump, since additional values were added to an existing api call parameter.

### DIFF
--- a/api/facadeversions.go
+++ b/api/facadeversions.go
@@ -76,7 +76,7 @@ var facadeVersions = map[string]int{
 	"MigrationTarget":              1,
 	"ModelConfig":                  2,
 	"ModelGeneration":              1,
-	"ModelManager":                 6,
+	"ModelManager":                 7,
 	"ModelUpgrader":                1,
 	"NotifyWatcher":                1,
 	"OfferStatusWatcher":           1,
--- a/api/modelmanager/modelmanager.go
+++ b/api/modelmanager/modelmanager.go
@@ -4,6 +4,8 @@
 package modelmanager
 
 import (
+	"time"
+
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	"gopkg.in/juju/names.v2"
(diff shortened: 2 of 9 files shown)


anastasiamac:
[add --force and --max-wait to destroy-model](https://trello.com/c/VjSXg4l2/691-add-force-and-max-wait-to-destroy-model)

anastasiamac:
$$merge$$

anastasiamac:
unrelated failure in lxdProfileWatcherSuite.TestMachineAppLXDProfileWatcherUnitChangeAppNotFound

$$merge$$

mitechie:
Is this --max-wait going to have some hints with it? I'm not sure how long
each step should take tbh and how many steps there are to judge a total
time to take. It seems an option where users can/would just toss random
stuff at it or 0 and go. I also think that if we hit the max-wait and are
timing out a step that we should help the user with output that instructs
them to do things like check their VMs for hung instances that couldn't be
removed or the like. The idea being that if we succeed then all's well, but
if we have to skip something how would the user manually go and garbage
collect the things left behind by Juju.

anastasiamac:
@mitechie,
There is a whole paragraph in command help about max-wait :)
Realistically, this is a crutch for a savy user that knows what they are doing. It is a way to say 'at each step of destruction, don't take longer than X'. Current Juju setup waits for about 30 sec and this option is a means to customise this interval. 
How many steps and what exactly is involved in destroying a model (or any other entity) depends on individual setup and deploy. It is not standard and could vary literally each time the destruction is desired. Having said that, we have started a brainstorm of how to provide visibility to what Juju does at each turn to ensure some form of observability is satisfied.
There is also work to ensure that the user get useful feedback if errors are encountered during forced destruction but the destruction was able to proceed. For example, there was a hook error on one of the units, machine agent was not responding, etc... There are provisions but due to the overall complexity of our product, it's a step-by-step process and will not be delivered as one PR.

PR #10053, opened 2019-04-16
Feature/bootstrapping2cdk aws

ycliuhw:
## Description of change

This PR introduces flexibility to customise k8s service definition for different base cloud;
For example, to support bootstrapping on CDK on AWS, we need define the LB on aws to use tcp to support websocket. 

Plus a small fix on status cmd: only show storage section if there are any to show, 

## QA steps

1. import cdk-aws cloud to local;

```bash
$ juju add-k8s <cloud-name> --local --cluster-name <gke-cluster-name> --debug
```

2. bootstrap controller;

```bash
$ juju bootstrap <cloud-name> <controller-name> --debug --config caas-image-repo=<ur-docker-ns>
```

3. deploy caas workload into the k8s controller;

Note: tested on `microk8s`, `GKE`, `CDK` on `AWS`;

## Documentation changes

None

## Bug reference

None


ycliuhw:
[bootstrapping to CDK on AWS support](https://trello.com/c/4A8JXK8b/698-bootstrapping-to-cdk-on-aws-support)

caas/kubernetes/provider/bootstrap.go
```diff
@@ -40,6 +42,39 @@ var (
...
 )
 
+type controllerServiceSpec struct {
+	// serviceType is required.
+	serviceType core.ServiceType
+	// annotations is optional.
+	annotations k8sannotations.Annotation
+}
+
+var controllerServiceSpecs = map[string]*controllerServiceSpec{
```
wallyworld: There should be a default as well for clouds we don't specifically enumerate below

caas/kubernetes/provider/bootstrap.go
```diff
@@ -294,8 +329,25 @@ func (c controllerStack) Deploy() (err error) {
...
 
+func (c controllerStack) getControllerSvcSpec(cloudType string) (*controllerServiceSpec, error) {
+	if spec, ok := controllerServiceSpecs[cloudType]; ok {
+		if spec.serviceType == "" {
+			// serviceType is required.
+			return nil, errors.NotValidf("serviceType is empty for %q", cloudType)
+		}
+		return spec, nil
+	}
+	return nil, errors.NotValidf("cloudType %q", cloudType)
```
wallyworld: We should just use a default as there are lots of cloud types, eg joyent, rackspace etc

ycliuhw: ah, right, added a default svc spec for any other cloud is not listed here

wallyworld:
Maybe we should add the CDK cloud labels as a drive by before landing (as per the IRC conversation). So that add-k8s will start working once CDK is updated.

ycliuhw:
$$merge$$

PR #10055, opened 2019-04-16
Be careful when initializing slices.

jameinel:
## Description of change

If the loop just does 'append' we need to make sure to set the capacity
and not the slice length.

## QA steps

```
$ JUJU_DEV_FEATURE_FLAGS="mongodb-snap" juju bootstrap lxd
$ juju ssh -m controller 0
$$ sudo cat /var/snap/juju-db/common/juju-db.config
```

Notice that with the change, you don't end up with a header, then 16 blank lines followed by 16 lines of content.

## Documentation changes

None.

## Bug reference

None.


jameinel:
$$merge$$

PR #10056, opened 2019-04-16
2.6 Provide the mongodb-sstxn feature flag

jameinel:
## Description of change

This wires up the server side transactions through to Juju so you can bootstrap with the feature flags and test them out. There are still some known transactions that are using old semantics that will need fixes, but getting this under a feature flag makes it much easier to find out what those are.

Interestingly, I did (accidentally) test upgrading from client-side transactions to server-side transactions, and nothing seem impacted by it. The logic should all be good and not trigger any problems. (As long as transactions are completed, I would expect it to be the same as if we just pruned transactions.)

## QA steps

```
$ export JUJU_DEV_FEATURE_FLAGS="mongodb-snap,mongodb-sstxn"
$ time juju bootstrap lxd lxd --debug --model-default "logging-config=<root>=INFO;juju.state.txn=TRACE;juju.txn=TRACE" --no-gui
```

You should see a line like:
```
2019-04-16 12:45:39 INFO juju.state open.go:179 using server-side transactions
```

## Documentation changes

@pmatulis We'll eventually want to document the feature flag, but we probably want to wait until we're ready for people to actually be bringing up controllers with it set.

## Bug reference

None.


state/open.go
```diff
@@ -165,11 +168,25 @@ func newState(
 		}
 	}()
 
+	mongodb := session.DB(jujuDB)
+	sstxn := featureflag.Enabled(feature.MongoDbSSTXN)
+	if sstxn {
+		if !txn.SupportsServerSideTransactions(mongodb) {
+			logger.Warningf("user requested server-side transactions, but they are not supported\n"+
+				" consider using the '%s' feature flag", feature.MongoDbSnap)
+			sstxn = false
```
achilleasa: You could add a `logger.Infof("using client-side transactions")` here for parity with the other info statements OR  alternatively convert the `else` block on L181 into a standalone `if !sstxn {...}` block so you hit it after you force sstxn to false on L177

jameinel:
$$merge$$

jameinel:
$$merge$$

jameinel:
$$merge$$

jameinel:
$$merge$$

jameinel:
$$merge$$

jameinel:
$$merge$$

PR #10058, opened 2019-04-17
Change max-wait to no-wait.

anastasiamac:
## Description of change

PR https://github.com/juju/juju/pull/10052 introduced --force and --max-wait to model-destroy command. Max-wait was provided to control how quickly model destruction could step through its necessary hoops. However, as per review comments, this could have also led to user confusion due to too much complexity. It was decided that the most valueable to user would be to say 'don't wait for each step to complete, just keep going'. So this PR changed the command option to 'no-wait'.

Users now will be able to specify:
1. 'destroy-model' - proceed with model destruction as per usual;
2. 'destroy-model --force' - destroy model ignoring all errors;
3. 'destroy-model --force --no-wait' - destroy model ignoring all errors as quickly as possible.

As a drive-by, renamed some test variables to avoid collision with the names of imported packages.

### DIFF
--- a/cmd/juju/model/destroy.go
+++ b/cmd/juju/model/destroy.go
@@ -66,8 +66,9 @@ type destroyCommand struct {
 	configAPI      ModelConfigAPI
 	storageAPI     StorageAPI
 
-	Force   bool
-	MaxWait time.Duration
+	Force  bool
+	NoWait bool
+	fs     *gnuflag.FlagSet
 }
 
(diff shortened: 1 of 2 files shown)


anastasiamac:
[change max-wait to no-wait](https://trello.com/c/gcnp3ORa/699-change-max-wait-to-no-wait)

anastasiamac:
$$merge$$

PR #10065, opened 2019-04-18
Ensure add-k8s uses existing storage if specified

wallyworld:
## Description of change

When running add-k8s on an unknown cluster with --storage, we should use the specified storage if it exists. We were previously not doing that and printing the wrong error message.

As a drive by, tweak the message when adding a duplicate model.

## QA steps

add-k8s --storage foo --region unknown/foo
should print storage-class "foo" not found

### DIFF
--- a/caas/kubernetes/provider/cloud.go
+++ b/caas/kubernetes/provider/cloud.go
@@ -92,7 +92,7 @@ func newCloudCredentialFromKubeConfig(reader io.Reader, cloudParams KubeCloudPar
 }
 
 // UpdateKubeCloudWithStorage updates the passed Cloud with storage details retrieved from the clouds' cluster.
-func UpdateKubeCloudWithStorage(k8sCloud *cloud.Cloud, credential cloud.Credential, storageParams KubeCloudStorageParams) (string, error) {
+func UpdateKubeCloudWithStorage(k8sCloud *cloud.Cloud, storageParams KubeCloudStorageParams) (string, error) {
 	fail := func(e error) (string, error) {
 		return "", e
 	}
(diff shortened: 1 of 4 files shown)


caas/kubernetes/provider/cloud.go
```diff
@@ -125,31 +125,44 @@ func UpdateKubeCloudWithStorage(k8sCloud *cloud.Cloud, credential cloud.Credenti
...
 	}
 
-	// If no storage class exists, we need to create one with the opinionated defaults.
+	// If no storage class exists, we need to create one with the opinionated defaults,
+	// or use an existing one.
 	var storageMsg string
-	if storageParams.WorkloadStorage != "" && caas.IsNonPreferredStorageError(err) {
-		preferredStorage := errors.Cause(err).(*caas.NonPreferredStorageError).PreferredStorage
+	if storageParams.WorkloadStorage != "" && maybeCustomStorage {
+		var (
```
ycliuhw: so does it mean if there is an existing sc named `--storage foo` found but which is not configured to Juju preferred provisioner, we will modify the foo to a Juju prefered config?
If so, we should probably not modify that.

wallyworld: If there is an existing sc, EnsureStorageProvisioner() will exit early. So we don't modify it.
If --storage is used, we will try and use the sc the user has specified. If that doesn't exist, and we can't query the cluster to get a preferred storage type, we will exit with not found.

ycliuhw:
```go
if storageParams.WorkloadStorage == ""{
	if errors.IsNotFound(err) {
		err = UnknownClusterError{CloudName: cloudType}
	}
	if caas.IsNonPreferredStorageError(err){
		npse := err.(*caas.NonPreferredStorageError)
		err = NoRecommendedStorageError{Message: err.Error(), ProviderName: npse.Name}
	}
}

if err != nil {
	return errors.Trace(err)
}

if storageParams.WorkloadStorage != "" {
	
}
```

caas/kubernetes/provider/cloud.go
```diff
@@ -125,31 +130,45 @@ func UpdateKubeCloudWithStorage(k8sCloud *cloud.Cloud, credential cloud.Credenti
...
+			return fail(UnknownClusterError{CloudName: cloudType})
+		}
+		if caas.IsNonPreferredStorageError(err) {
+			npse := err.(*caas.NonPreferredStorageError)
+			return fail(NoRecommendedStorageError{Message: err.Error(), ProviderName: npse.Name})
+		}
+		if err != nil {
+			return fail(errors.Trace(err))
+		}
+	}
```
ycliuhw: do we want to return any known errors for `storageParams.WorkloadStorage != ""` case?

wallyworld: workload storage != "" is handled below

wallyworld:
$$merge$$

PR #10066, opened 2019-04-18
Report but don't return k8s pod spec errors

wallyworld:
## Description of change

When creating a k8s service during application deploy, if there's an error in the pod spec, k8s will report that. We need to report it to Juju but not exit the worker with an error, to prevent the worker continually restarting over and over.

## QA steps

Deploy a k8s charm with an error in the pod spec.
Check that juju status shows the error, and that the debug-log shows the worker not restarting

### DIFF
--- a/caas/kubernetes/provider/errors.go
+++ b/caas/kubernetes/provider/errors.go
@@ -3,6 +3,11 @@
 
 package provider
 
+import (
+	"github.com/juju/errors"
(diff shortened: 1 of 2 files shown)


wallyworld:
[charm pod spec issues](https://trello.com/c/y2sRIn5s/700-charm-pod-spec-issues)

wallyworld:
$$merge$$

PR #10067, opened 2019-04-18
remove-unit --force: cleanups to remove a unit if the uniter isn't responding

babbageclunk:
## Description of change

Unit removals can stall if the uniter (in the unit agent) and deployer (in the machine agent) don't respond to the unit's lifecycle changes for some reason - a common example is a charm hook failing with an error.

To handle this we now have the `--force` flag on the `remove-unit` command. If the force flag is set when removing a unit we schedule a fallback cleanup from the dyingUnit cleanup. This cleanup performs the tasks the unit agent normally performs when a unit is dying:
* removing subordinates
* leaving the scope of any relations
* detaching storage
* marking the unit dead

Then there's another cleanup to remove the unit in case the deployer is stuck for some reason.

At the moment the fallback cleanups are scheduled immediately, but once the `--no-wait` flag is fully plumbed through from CLI -> API -> state -> cleanups it will be deferred for a minute (unless `--no-wait` is set) to allow the uniter & deployer to act.

## QA steps

* Deploy the [hook-error](https://private-fileshare.canonical.com/~xtian/hook-error.zip) charm to a model with some storage `juju deploy ~/path/to/hook-error -n 3 --storage pgdata=tmpfs,1G,1`.
* Add subordinates and other relations to it (it has a require-mysql endpoint).
* Once the units are ready, change the config so that the unit has an error processing the hook: `juju config hook-error crash=true`
* Running `juju remove-unit hook-error/0` will work but the unit will stick around.
* Running `juju remove-unit --force hook-error/0` will succeed - the subordinates and storage will be cleaned up, and the db relations will be left.
* The unit will be removed from status.
* Running `remove-application --force hook-error` will also not get blocked by the error units.

## Documentation changes

This doesn't require any docs changes (although it relies on the new `--force` flags to remove-unit, remove-application and destroy-model).

## Bug reference

None.

### DIFF
--- a/state/cleanup.go
+++ b/state/cleanup.go
@@ -20,11 +20,19 @@ import (
 type cleanupKind string
 
 const (
+	// forceTimeout is how far in the future the backstop force
+	// cleanup will be scheduled.
+	forceTimeout = time.Minute
+)
+
+const (
 	// SCHEMACHANGE: the names are expressive, the values not so much.
 	cleanupRelationSettings              cleanupKind = "settings"
 	cleanupUnitsForDyingApplication      cleanupKind = "units"
 	cleanupCharm                         cleanupKind = "charm"
 	cleanupDyingUnit                     cleanupKind = "dyingUnit"
+	cleanupForceDestroyedUnit            cleanupKind = "forceDestroyUnit"
+	cleanupForceRemoveUnit               cleanupKind = "forceRemoveUnit"
 	cleanupRemovedUnit                   cleanupKind = "removedUnit"
 	cleanupApplicationsForDyingModel     cleanupKind = "applications"
 	cleanupDyingMachine                  cleanupKind = "dyingMachine"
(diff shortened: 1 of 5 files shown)


state/cleanup.go
```diff
@@ -531,6 +574,123 @@ func (st *State) cleanupDyingUnit(name string, cleanupArgs []bson.Raw) error {
...
+		// the unit to dead until the subordinates and storage are
+		// gone, so we should give them time to be removed.
+		return err
+	} else if err != nil {
+		logger.Warningf("couldn't set unit %q dead: %v", unitId, err)
+	}
+
+	// Set up another cleanup to remove the unit in a minute if the
+	// deployer doesn't do it.
+	st.scheduleForceCleanup(cleanupForceRemoveUnit, unitId)
```
wallyworld: Maybe in this case we really to want to wait a bit to save thrashing on error in --force

babbageclunk: I think scheduling this immediately is ok - if the cleanup is failing because of storage or subordinates it will have exited before now. This will run at most once - if it happens, the cleanup will succeed and be removed, so it can't cause thrashing.

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10069, opened 2019-04-18
Advise user how to access migrated models

achilleasa:
## Description of change

This PR updates both the client and the server code so that clients (with sufficient access) attempting to connect to a model (basically, running any CLI command supporting a `-m` flag) that has been migrated to another controller will get back an error letting them know that the model has been migrated and providing advise on how to access it.

When a model gets successfully migrated to another controller, its relevant documents get deleted from the DB and the only thing that remains is a document that tracks the migration itself and details about the controller the model got migrated to. 

When users attempt to run a model-related command (e.g `juju status -m XXX`) the client attempts to establish a websocket connection and includes the target model UUID in the request URL. As the model document does not exist any more, the connection gets torn down with a NotFound error. Slightly O/T: this is interesting from a security standpoint as the current implementation provides an oracle for finding out if a model with a particular UUID exists or not; however this is probably not an issue as UUIDs don't convey any useful information anyway.

The server can figure out if a model has been migrated by checking whether a migration document exists for that particular model UUID. To make redirects work for our use-case we also need to know whether the user **used to have access** to the model before it got migrated. The first 2 commits from the PR update the migration document code to also track the users and their access levels before initiating the migration. Note that this information cannot be recovered for older controllers so we don't need an upgrade step for the extra field. Also, this information does not get migrated across controllers so presumably no migration step is required either.

With these 2 changes we have the means to detect if we should return a redirect error to the user or not. Unfortunately, we need to wait for the user to login first before we can decide. To make this flow work, the admin route handler has been modified to allow websocket connections to be established **if and only if** we are aware that a particular model has been migrated. This allows the users to invoke the `Login` RPC and to get back either a `RedirectError` or an `Unauthorized` error depending on whether they used to have access to the model or not.

On the client-side, things are also a bit more complicated due to the fact that we already have a codepath that deals with RedirectErrors! However, that path is used to implement **transparent** redirects for the JAAS use-case where JAAS appears as a controller front-end that can potentially dispatch requests to other controllers.

For _this_ particular use-case we **do not** want to transparently redirect users but would rather abort the command with an error advising the users how to connect to the new controller (either via a `juju login $controller_ip` or a `juju switch XXX` if the destination controller is already known). The reason for this is that we always want users to be aware on which controller they are executing commands.

To ensure that both cases work as expected, the client-side `RedirectError` has been augmented with a `FollowRedirect` attribute (set to true for JAAS-like cases) which allows us to provide a user-friendly error message for redirected models in the non-JAAS use-case.

NOTE: The changes from this PR only affect model-related commands (CLI commands using modelcmd/base.go). Commands that login to the controller like `juju show-model XXX` will just get back a `permission denied` error when targeting a migrated model. If we need to support redirects for this type of commands we can do it via a separate PR. 

### Migrate model to locally known controller 

```console
$ juju bootstrap lxd src
$ juju change-user-password 

$ juju bootstrap lxd dst
$ juju change-user-password

$ juju switch src:controller
$ juju add-model migrate

# At this point make a copy of your ~/.local/share/juju/models.yaml
# as the model uuid will probably be removed from src after migrating
$ juju migrate src:migrate dst

# wait a few seconds for the migration to complete and ensure that 
# the source controller in your local models.yaml still contains the 
# uuid for the model that just got migrated 

# Try any model-related command, e.g:
$ juju status -m migrate
ERROR Model "migrate" has been migrated to controller "dst".
To access it run 'juju switch dst:migrate'.
```

### Migrate model to controller not locally known

```console
# Run these steps after trying out the previous QA step
$ juju unregister dst

# Try any model-related command, e.g:
$ juju status -m migrate
ERROR Model "migrate" has been migrated to another controller.
To access it run one of the following commands:
  'juju login 10.65.47.40:17070 -c new-controller'

New controller fingerprint [93:73:88:C9:9A:AA:EC:C8:85:AE:D1:33:E5:92:CE:95:0F:B6:00:82:21:CB:8C:A0:42:16:29:77:CF:6D:B6:D4]
```

### Different client/server versions

```
| client | server | juju status -m $model                         |
|--------|--------|-----------------------------------------------|
| 2.5    | 2.5    | model not found                               |
| 2.5    | 2.6    | cannot get redirect addresses: not redirected |
| 2.6    | 2.5    | model not found                               |
```

## Documentation changes

@pmatulis we probably need to document this work-flow.

### DIFF
--- a/api/apiclient.go
+++ b/api/apiclient.go
@@ -164,6 +164,10 @@ type RedirectError struct {
 
 	// CACert holds the certificate of the remote server.
 	CACert string
+
+	// FollowRedirect is set to true for cases like JAAS where the client
+	// needs to automatically follow the redirect to the new controller.
+	FollowRedirect bool
 }
 
 func (e *RedirectError) Error() string {
--- a/api/apiclient_test.go
+++ b/api/apiclient_test.go
@@ -692,8 +692,9 @@ func (s *apiclientSuite) TestOpenWithRedirect(c *gc.C) {
 
 	hps, _ := network.ParseHostPorts(redirectToHosts...)
 	c.Assert(errors.Cause(err), jc.DeepEquals, &api.RedirectError{
-		Servers: [][]network.HostPort{hps},
-		CACert:  redirectToCACert,
+		Servers:        [][]network.HostPort{hps},
+		CACert:         redirectToCACert,
+		FollowRedirect: true,
 	})
 }
 
--- a/api/apiclient_test.go
+++ b/api/apiclient_test.go
@@ -179,14 +179,14 @@ func (s *apiclientSuite) TestVerifyCA(c *gc.C) {
 			// proceeds with the connection to the servers. This
 			// would be the case where we connect to an older juju
 			// controller.
-			expConnCount: 6,
+			expConnCount: 2,
 			errRegex:     `unable to connect to API: .*`,
 		},
 		{
 			descr:      "no VerifyCA provided",
 			serverCert: serverCertWithSelfSignedCA,
 			// Dial connects to all servers
-			expConnCount: 3,
+			expConnCount: 1,
 			errRegex:     `unable to connect to API: .*`,
 		},
 		{
@@ -196,7 +196,7 @@ func (s *apiclientSuite) TestVerifyCA(c *gc.C) {
 				return errors.New("CA not trusted")
 			},
 			// Dial aborts after fetching CAs
-			expConnCount: 3,
+			expConnCount: 1,
 			errRegex:     "CA not trusted",
 		},
 		{
@@ -206,7 +206,7 @@ func (s *apiclientSuite) TestVerifyCA(c *gc.C) {
 				return nil
 			},
 			// Dial fetches CAs and then proceeds with the connection to the servers
-			expConnCount: 6,
+			expConnCount: 2,
 			errRegex:     `unable to connect to API: .*`,
 		},
 	}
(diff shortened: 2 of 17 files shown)


pmatulis:
The QA steps work perfectly for me. All 2.6, so no testing of permutations of client and server versions.

mitechie:
One bit of feedback is the message for logginginto the new controller. When we do a juju register we give a hint for the controller name to match what was used by the admin.

    juju register MGwTBHJ...
    Enter a new password:
    Confirm password:
    Enter a name for this controller [src]:

It'd be great if we could crib that so that the message here turns into

    $ juju status -m migrate
    ERROR Model "migrate" has been migrated to another controller.
    To access it run one of the following commands:
      'juju login 10.65.47.40:17070 -c src'

mitechie:
2.5 client on 2.6 server gets:

    ERROR invalid entity name or password (unauthorized access)

After the model is moved and it attempts to read it from the src controller

    juju status src:admin/moveme

Note I cheated by creating a second user account, granting that user superuser access, and having that user on both controllers. In this way I didn't need to monkey with local files and just used the two accounts to test. 

    % juju whoami
    Controller:  src
    Model:       admin/moveme
    User:        rick
    % juju status -m src:admin/moveme
    ERROR invalid entity name or password (unauthorized access)
    % juju status -m dst:admin/moveme
    Model   Controller  Cloud/Region     Version      SLA          Timestamp
    moveme  dst         google/us-east1  2.6-beta2.1  unsupported  14:12:02-04:00
    
    Model "admin/moveme" is empty.


When I attempt from the 2.6 (this PR) client I do get the solid message:

    # as original admin user
    % juju status -m src:moveme
    ERROR Model "admin/moveme" has been migrated to controller "dst".
    To access it run 'juju switch dst:admin/moveme'.

    % juju status -m dst:moveme
    Model   Controller  Cloud/Region     Version      SLA          Timestamp
    moveme  dst         google/us-east1  2.6-beta2.1  unsupported  14:10:15-04:00
    
    Model "moveme" is empty.

mitechie:
I wondered if I would get the notice about redirect if the 2.5 client talking to the 2.6 server was the admin user to I logged out and back in as the admin user on the 2.5 client and got:

     juju status -m src:moveme
    ERROR model src:admin/moveme not found

So there's definitely a difference with the user account and the error that comes back even though the superuser permission is common between them.

mitechie:
Validated that the 2.6 client against the 2.5 controller (moved model) just shows the error about the model not being found. 

    % juju status -m admin/moveme
    ERROR model src:admin/moveme not found

achilleasa:
> I wondered if I would get the notice about redirect if the 2.5 client talking to the 2.6 server was the admin user to I logged out and back in as the admin user on the 2.5 client and got:
> 
> ```
>  juju status -m src:moveme
> ERROR model src:admin/moveme not found
> ```
> So there's definitely a difference with the user account and the error that comes back even though the superuser permission is common between them.

I suspect you got this because your local models.yaml does not contain the model UUID. When using the 2.5 client it filters out models.yaml because it cannot find the model. You can edit it, manually copy the relevant entry from the `dst` controller and try the command again to hit the redirect.

achilleasa:
> One bit of feedback is the message for logging into the new controller. When we do a juju register we give a hint for the controller name to match what was used by the admin.
> 
> ```
> juju register MGwTBHJ...
> Enter a new password:
> Confirm password:
> Enter a name for this controller [src]:
> ```
> It'd be great if we could crib that so that the message here turns into
> 
> ```
> $ juju status -m migrate
> ERROR Model "migrate" has been migrated to another controller.
> To access it run one of the following commands:
>   'juju login 10.65.47.40:17070 -c src'
> ```
> That can be looked at as follow up polish and isn't a -1 on this PR.

This was my initial thought but then I realized that the controller label is just a local alias for the controller UUID. AFAICT the server is not aware of the controller name (i.e I can refer to it as `foo` on my  machine and you can refer to it as `bar` on yours) so we cannot expose it to the client.

The only exception is when we can match the endpoint that we get back from the redirect to an existing local entry. In that case you will get the model has been migrated to controller "XXX" message.

mitechie:
This was my initial thought but then I realized that the controller label is just a local alias for the controller UUID. 

How does the register command do it though? Since I'm on a different machine with no local reference to the controller the register command has no right to know the name it's suggesting unless register itself is backing it into the hash information that's used as part of the register command?

achilleasa:
> ```
> This was my initial thought but then I realized that the controller label is just a local alias for the controller UUID. 
> ```
> How does the register command do it though? Since I'm on a different machine with no local reference to the controller the register command has no right to know the name it's suggesting unless register itself is backing it into the hash information that's used as part of the register command?

The register command packs it in the ASN1-encoded [hash](https://github.com/juju/juju/blob/develop/cmd/juju/controller/register.go#L428) data. We could theoretically push the local controller name to the remote endpoint when we initiate the model migration and store it in the migration document so we can send it back to the user as part of the RedirectError payload (falling back to a generic name suggestion if it is not filled in). Not sure if it's worth the development effort but happy to spike it next week if it's something that would improve UX.

mitechie:
Not sure if it's worth the development effort but happy to spike it next week if it's something that would improve UX.

Agreed, it's something for follow up as polish.

achilleasa:
$$merge$$

PR #10072, opened 2019-04-20
Remove InstanceMutater feature flag and old update charm lxd profile code path

hmlanigan:
## Description of change

Remove InstanceMutater feature flag and fix a few tests failing due to
change in machine and model manifolds, and uniter worker.

Remove old code to upgrade a charm lxd profile which has been replaced
with the instance mutater worker. Including the instanceCharmProfileData
collection.

## QA steps

Deploy and upgrade charms with and without lxd profiles.  Try bundles
with more than 5 machines, subordinates, containers: with and without 
the lxd provider, local and charm store charms.

Bootstrap 2.5.4.  Deploy a few charms with and without lxd profiles.  
Upgrade the juju controller but not model to this code.  Upgrade charms,
should failure with error message to upgrade the model on charms 
with lxd profiles.  Upgrade the model and try upgrade charms again
without error.

charms with an lxd profile:
./testcharms/charm-repo/quantal/lxd-profile-alt
./testcharms/charm-repo/quantal/lxd-profile-subordinate

juju deploy cs:nova-compute
juju deploy cs:neutron-openvswitch-252
juju add-relation nova-compute:neutron-plugin neutron-openvswitch:neutron-plugin
let settle, will be waiting for other relations
juju upgrade-charm neutron-openvswitch  <-- adds an lxd profile


apiserver/facades/agent/instancemutater/shim.go
```diff
@@ -4,113 +4,14 @@
...
 	"github.com/juju/juju/core/cache"
-	"github.com/juju/juju/core/lxdprofile"
 	"github.com/juju/juju/state"
 )
 
 type instanceMutaterStateShim struct {
 	*state.State
 }
 
-func (i *instanceMutaterStateShim) Model() (Model, error) {
```
hmlanigan: Removed: these were left over from when state was used instead of the model cache.

apiserver/facades/agent/instancemutater/interface.go
```diff
@@ -16,46 +16,15 @@ import (
 type InstanceMutaterState interface {
 	state.EntityFinder
 
-	WatchModelMachines() state.StringsWatcher
```
hmlanigan: Removed: these were left over from when state was used instead of the model cache.

cmd/jujud/agent/util_test.go
```diff
@@ -418,3 +421,19 @@ func (e *minModelWorkersEnviron) AllInstances(context.ProviderCallContext) ([]in
 func (e *minModelWorkersEnviron) Instances(ctx context.ProviderCallContext, ids []instance.Id) ([]instances.Instance, error) {
 	return nil, nil
 }
+
+func (env *minModelWorkersEnviron) MaybeWriteLXDProfile(pName string, put *charm.LXDProfile) error {
```
hmlanigan: needed for the instance-mutated worker to start and stay started with the manifold related tests

state/watcher.go
```diff
@@ -2102,8 +1853,8 @@ func watchInstanceCharmProfileCompatibilityData(backend modelBackend, watchDocId
 	}
 	extract := func(data map[string]interface{}) (string, error) {
 		if value, ok := data["charmurl"]; ok {
-			if url, ok := value.(*charm.URL); ok {
-				return url.String(), nil
+			if url, ok := value.(string); ok {
```
hmlanigan: This fixes a bug where the uniter was printing an error of "invalid applicationDoc type"

worker/provisioner/provisioner_task.go
```diff
@@ -339,191 +324,6 @@ func (task *provisionerTask) processMachines(ids []string) error {
 	return task.startMachines(pending)
 }
 
-// processProfileChanges adds, removes, or updates lxc profiles changes to
```
hmlanigan: this block of code is replaced by the instancemutater worker.

worker/uniter/util_test.go
```diff
@@ -414,12 +416,14 @@ type createUniter struct {
 func (s createUniter) step(c *gc.C, ctx *context) {
 	step(c, ctx, ensureStateWorker{})
 	step(c, ctx, createApplicationAndUnit{})
+	//step(c, ctx, createApplicationAndUnit{applicationName: "wordpress"})
```
hmlanigan: I'll remove this

worker/uniter/util_test.go
```diff
@@ -396,6 +396,8 @@ func (csau createApplicationAndUnit) step(c *gc.C, ctx *context) {
 	app := ctx.s.AddTestingApplicationWithStorage(c, csau.applicationName, sch, csau.storage)
 	unit, err := app.AddUnit(state.AddUnitParams{})
 	c.Assert(err, jc.ErrorIsNil)
+	err = unit.SetCharmURL(curl(0))
```
hmlanigan: Allow the watchInstanceCharmProfileCompatibilityData to work with the test.

hmlanigan:
I apologize in advance, 4019 lines are removed and 203 lines added.  My recommendation 
is to review the apiserver changes where some calls are now old and not needed in the 
current versions; review the new code (i'll point it out); and test it very well.

apiserver/facades/agent/provisioner/provisioner.go
```diff
@@ -314,106 +313,30 @@ func (p *ProvisionerAPI) WatchAllContainers(args params.WatchContainers) (params
...
-// WatchContainersCharmProfiles starts a StringsWatcher to  notifies when
+// WatchContainersCharmProfiles starts a StringsWatcher to notify when
 // the provisioner should update the charm profiles used by a container on
 // the given machine.
+//
+// Remove in juju 3, this is for compatibility of provisioner using APIV7.
+// It's required for the worker to start, and for the former
+// processProfileChanges, which will no longer be used in APIv9.
+// Code in apiserver application SetCharm() will prevent APIv8
+// models from upgrading charm with profiles and causing issues.
```
wallyworld: Because ProvisionerAPIV7 embeds ProvisionerAPIV8, you don't need to duplicate this WatchContainersCharmProfiles() method here on V7.

apiserver/facades/agent/provisioner/provisioner.go
```diff
@@ -1504,34 +1418,8 @@ func (p *ProvisionerAPIV7) ModificationStatus(_, _ struct{}) {}
...
-			statusInfo, err = machine.ModificationStatus()
-			result.Results[i].Status = statusInfo.Status.String()
-			result.Results[i].Info = statusInfo.Message
-			result.Results[i].Data = statusInfo.Data
-			result.Results[i].Since = statusInfo.Since
-		}
-		result.Results[i].Error = common.ServerError(err)
-	}
-	return result, nil
+func (p *ProvisionerAPIV8) ModificationStatus(args params.Entities) (params.StatusResults, error) {
```
wallyworld: So older provisioner clients don't call this anymore? Not being fully across the work to date, I'm not sure why we need to keep v8 WatchModelMachinesCharmProfiles() etc but not this v8 method as well.

apiserver/facades/agent/provisioner/provisioner.go
```diff
@@ -1561,7 +1449,7 @@ func (p *ProvisionerAPI) SetModificationStatus(args params.SetStatus) (params.Er
 }
 
 func (p *ProvisionerAPI) setOneModificationStatus(canAccess common.AuthFunc, arg params.EntityStatusArgs) error {
-	logger.Tracef("SetInstanceStatus called with: %#v", arg)
+	logger.Tracef("SetModificationStatus called with: %#v", arg)
```
wallyworld: So the ModificationStatus() method above raises "NotSupported" and yet we've retained this method it used to call but which appears no longer used now?

state/machine.go
```diff
@@ -2210,483 +2208,6 @@ func (m *Machine) VerifyUnitsSeries(unitNames []string, series string, force boo
 	return results, nil
 }
 
-// instanceCharmProfileDataC holds attributes relevant to a lxd charm profile
-// data that's on the machine
-type instanceCharmProfileData struct {
```
wallyworld: Does there need to be an upgrade step to delete any existing data that we now no longer use?

state/watcher.go
jameinel: The type of the object should actually be a map[string]interface{} since we should have a Document type here, and we deserialized into a map[string]interface{} so nested documents should have that same type.

apiserver/facades/agent/provisioner/provisioner.go
jameinel: I chatted with Heather about this. We have explicitly chosen for the Profile work not *not* be backwards compatible. And instead, we have code that checks you can't upgrade a charm to a version that has a Profile *if* you are running in mixed mode. (This was done, because the old code was watching for documents that we just aren't creating now.)

We don't want those agents to bounce when they call in, we just don't have any work for them to do (which we've ensured with the 'upgrade-charm' code.)

I think this is ok. We'd *love* to return NotSupported from the other ones to make it clear that the particular version of the API isn't available anymore, but that causes the workers to bounce. This doesn't have a valid return, and as long as the Watcher code doesn't give them a reason to look at this (since it is disabled), there shouldn't be a problem.


We should document the API compatibility break in a comment.

hmlanigan: It appears this is dead code and was never called in 2.5.2, 2.5.3, 2.5.4, nor tip of 2.5.  It was added in when SetModificationStatus() was added in 2.5.2.

apiserver/facades/agent/provisioner/provisioner.go
hmlanigan: setOneModificationStatus() is called by SetModificationStatus() which is being kept, not ModificationStatus().

state/machine.go
hmlanigan: Any docs in this collection are transient and "should" not exist at the time of an upgrade.  However an upgrade step won't hurt, to be sure.  Will add.

apiserver/facades/agent/provisioner/provisioner.go
SimonRichardson: If this is dead code, can we not remove it?

apiserver/facades/agent/uniter/lxdprofile.go
```diff
@@ -227,21 +226,15 @@ func (u *LXDProfileAPI) RemoveUpgradeCharmProfileData(args params.Entities) (par
...
-			result.Results[i].Error = common.ServerError(err)
-			continue
-		}
-		unit, err := u.getUnit(tag)
-		if err != nil {
-			result.Results[i].Error = common.ServerError(err)
-			continue
-		}
-		err = machine.RemoveUpgradeCharmProfileData(unit.Name())
+		_, err = u.getMachine(tag)
```
SimonRichardson: Why continue to call this, are we attempting to find out if it's `not-provisioned` or `not-found`?

state/allcollections.go
```diff
@@ -559,7 +558,6 @@ const (
 	guimetadataC               = "guimetadata"
 	guisettingsC               = "guisettings"
 	instanceDataC              = "instanceData"
-	instanceCharmProfileDataC  = "instanceCharmProfileData"
```
SimonRichardson: Nice!

state/upgrades.go
```diff
@@ -2086,3 +2086,18 @@ func EnsureApplicationDeviceConstraints(pool *StatePool) error {
 	}
 	return nil
 }
+
+// RemoveInstanceCharmProfileDataCollection removes the
+// instanceCharmProfileData collection on upgrade.
+func RemoveInstanceCharmProfileDataCollection(pool *StatePool) error {
```
SimonRichardson: This is such a good idea, I didn't think about this... this should ensure that no existing upgrades cause erroneous errors.

state/watcher.go
SimonRichardson: see https://github.com/juju/juju/pull/10081

worker/uniter/util_test.go
SimonRichardson: ??

apiserver/facades/agent/uniter/lxdprofile.go
hmlanigan: This will be pulled out in pr fixing api/apiserver for uniter, was trying to break this into pieces.  Should have just returned empty results no errors.

worker/uniter/util_test.go
hmlanigan: darn gremlins, i remember removing this.  ty, will check that the change got pushed up.

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #10073, opened 2019-04-22
Tweak status for k8s apps without pods

wallyworld:
## Description of change

Some k8s charms will not start any pods. So the "waiting for container" status message is misleading. And more so when the charm has set status to Active but Juju still displays it as Waiting because there are no pods yet.
This PR tweaks the k8s unit/app status so that if not pod spec has been set, it instead displays as "agent initializing". If the charm does set a status, that is still preferred.

Writing tests showed up a serious, latent bug in setting the params.ApplicationStatus error. The type was wrong and needed to be params.Error not error.
Also, the test help to clear status times now also clears empty error data to make things easier.

Also a driveby Makefile fix to enable jujud to be compiled with debug info left in.

## QA steps

Deploy a "normal" k8s app and check status history.
Deploy a k8s app which does not set a pod spec and check status history.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,15 @@ else
 	CHECK_ARGS =
 endif
 
+# Compile with debug flags if requested.
+ifeq ($(DEBUG_JUJU), 1)
+    COMPILE_FLAGS = -gcflags "all=-N -l"
+    LINK_FLAGS =
+else
+    COMPILE_FLAGS =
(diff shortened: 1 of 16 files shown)


wallyworld:
[k8s status for charms without pods](https://trello.com/c/KYj7wotN/703-k8s-status-for-charms-without-pods)

wallyworld:
$$merge$$

PR #10076, opened 2019-04-23
Add --no-wait for remove-application and -unit CLI and api.

anastasiamac:
## Description of change

Unit or application removal is a multi-step process. Currently, Juju will not proceed to the next step until a previous step in the process is completed successfully. This may block the removal process - next steps are not called and the removal does not complete. For example, Juju may not be able to contact machine agent to do some unit cleanup jobs and unit removal will seem to hang. 

This PR is part of the change that will support force removal despite operational errors. Force removal will call the next step in the process regardless of where the current step is or if it has succeeded. Generally, we'd wait for a set period of time before calling this next step. However, with 'no-wait' command option specified, Juju will forgo this wait and will call the next step is soon as possible. This means that the whole force removal process will be faster.

### DIFF
--- a/apiserver/params/params.go
+++ b/apiserver/params/params.go
@@ -414,13 +414,18 @@ type DestroyUnitParams struct {
 
 	// Force controls whether or not the destruction of an application
 	// will be forced, i.e. ignore operational errors.
-	Force bool
+	Force bool `json:"force"`
 
 	// Errors contains errors encountered while applying this operation.
 	// Generally, these are non-fatal errors that have been encountered
 	// during, say, force. They may not have prevented the operation from being
 	// aborted but the user might still want to know about them.
-	Errors []error
+	Errors []error `json:"errors,omitempty"`
+
+	// MaxWait specifies the amount of time that each step in unit removal
+	// will wait before forcing the next step to kick-off. This parameter
+	// only makes sense in combination with 'force' set to 'true'.
+	MaxWait *time.Duration `json:"max-wait,omitempty"`
 }
 
 // Creds holds credentials for identifying an entity.
--- a/cmd/juju/application/removeunit.go
+++ b/cmd/juju/application/removeunit.go
@@ -273,9 +273,6 @@ func (c *removeUnitCommand) removeUnits(ctx *cmd.Context, client RemoveApplicati
 		if c.NoWait {
 			zeroSec := 0 * time.Second
 			maxWait = &zeroSec
-		} else {
-			oneMin := 1 * time.Minute
-			maxWait = &oneMin
 		}
 	}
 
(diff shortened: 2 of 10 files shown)


anastasiamac:
[add --no-wait to remove -unit and -application](https://trello.com/c/kvjeQsQm/692-add-no-wait-to-remove-unit-and-application)

apiserver/params/params.go
```diff
@@ -414,13 +414,18 @@ type DestroyUnitParams struct {
...
 	// will be forced, i.e. ignore operational errors.
-	Force bool
+	Force bool `json:"force"`
 
 	// Errors contains errors encountered while applying this operation.
 	// Generally, these are non-fatal errors that have been encountered
 	// during, say, force. They may not have prevented the operation from being
 	// aborted but the user might still want to know about them.
-	Errors []error
+	Errors []error `json:"errors,omitempty"`
```
wallyworld: I am pretty sure this needs to be of type []params.Error or else serialisation across the wire will fail spectacularly.

cmd/juju/application/removeunit.go
```diff
@@ -206,6 +219,19 @@ func (c *removeUnitCommand) unitsHaveStorage(unitNames []string) (bool, error) {
 // Run connects to the environment specified on the command line and destroys
 // units therein.
 func (c *removeUnitCommand) Run(ctx *cmd.Context) error {
+	noWaitSet := false
+	forceSet := false
+	c.fs.Visit(func(flag *gnuflag.Flag) {
```
wallyworld: This should be done in Init()
Generally, unless there's a good reason, arg checking is done as early as possible.

cmd/juju/application/removeunit.go
```diff
@@ -242,10 +268,22 @@ func (c *removeUnitCommand) removeUnitsDeprecated(ctx *cmd.Context, client Remov
 }
 
 func (c *removeUnitCommand) removeUnits(ctx *cmd.Context, client RemoveApplicationAPI) error {
+	var maxWait *time.Duration
+	if c.Force {
+		if c.NoWait {
+			zeroSec := 0 * time.Second
+			maxWait = &zeroSec
+		} else {
+			oneMin := 1 * time.Minute
```
wallyworld: Would prefer that the wait time in params is a pointer. If not set, the server gets to decide what to use. Save all the different CLI clients from needing to specify it. And there's no common const used anyway.

cmd/juju/application/removeunit.go
anastasiamac: No, it cannot because we need to have a flagset populated to know if the users have provided the options or not.

cmd/juju/application/removeunit.go
anastasiamac: It is a pointer here but I'll change to pass nil when --no-wait is false.

anastasiamac:
$$merge$$

PR #10079, opened 2019-04-23
Fix another case of removing a wedged application

wallyworld:
## Description of change

An application can get stuck when being removed if the unit and relation count get to 0 and the db becomes inconsistent with that. remove --force should work but it doesn't. This PR has a small fix for that, based on what was done elsewhere but also needed to be done here. If --force, alter the assert used to check the app life.

Hard to unit test because we can't easily get the db into the bad state, and --force remove app is not really unit tested in the state package at the moment anyway.

## QA steps

Run assess nw health CI test, use this PR to force remove the wedged app.

### DIFF
--- a/state/application.go
+++ b/state/application.go
@@ -369,12 +369,19 @@ func (op *DestroyApplicationOperation) destroyOps() ([]txn.Op, error) {
 	// If the application has no units, and all its known relations will be
 	// removed, the application can also be removed.
 	if op.app.doc.UnitCount == 0 && op.app.doc.RelationCount == removeCount {
-		hasLastRefs := bson.D{{"life", Alive}, {"unitcount", 0}, {"relationcount", removeCount}}
+		// If we're forcing destruction the assertion shouldn't be that
+		// life is alive, but that it's what we think it is now.
+		assertion := isAliveDoc
+		if op.Force {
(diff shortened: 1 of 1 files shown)


state/application.go
```diff
@@ -369,12 +369,19 @@ func (op *DestroyApplicationOperation) destroyOps() ([]txn.Op, error) {
 	// If the application has no units, and all its known relations will be
 	// removed, the application can also be removed.
 	if op.app.doc.UnitCount == 0 && op.app.doc.RelationCount == removeCount {
-		hasLastRefs := bson.D{{"life", Alive}, {"unitcount", 0}, {"relationcount", removeCount}}
+		// If we're forcing destruction the assertion shouldn't be that
+		// life is alive, but that it's what we think it is now.
+		assertion := isAliveDoc
+		if op.Force {
+			assertion = bson.D{{"life", op.app.doc.Life}}
```
anastasiamac: I wonder if we need to assert on life when forcing at all... Maybe it should be just that the doc exists?

wallyworld: Maybe. We can discuss with xtian and fix in all places this pattern is used.

wallyworld:
$$merge$$

PR #10081, opened 2019-04-23
Watcher cleanup

SimonRichardson:
## Description of change

Typed extraction

The following changes ensure that we're better typed for extracting
the type rather than falling down to a `map[string]interface{}`. This
should clean up any potential issues when trying to understand the
underlying type.

Example: https://github.com/juju/juju/pull/10081#discussion_r277629465

## QA steps

Tests should pass.

### DIFF
--- a/state/watcher.go
+++ b/state/watcher.go
@@ -1848,22 +1848,20 @@ func watchInstanceCharmProfileCompatibilityData(backend modelBackend, watchDocId
 	initial := ""
 	members := bson.D{{"_id", watchDocId}}
 	collection := applicationsC
-	document := func() interface{} {
-		return applicationDoc{}
-	}
 	filter := func(id interface{}) bool {
 		return id.(string) == watchDocId
 	}
-	extract := func(data interface{}) (string, error) {
-		if doc, ok := data.(applicationDoc); ok {
-			return doc.CharmURL.String(), nil
+	extract := func(query documentFieldWatcherQuery) (string, error) {
+		var doc applicationDoc
+		if err := query.One(&doc); err != nil {
+			return "", err
 		}
-		return initial, errors.NotValidf("applicationDoc type")
+		return doc.CharmURL.String(), nil
 	}
 	transform := func(value string) string {
 		return lxdprofile.NotRequiredStatus
 	}
-	return newDocumentFieldWatcher(backend, collection, document, members, initial, filter, extract, transform)
+	return newDocumentFieldWatcher(backend, collection, members, initial, filter, extract, transform)
 }
 
 // *Deprecated* Although this watcher seems fairly admirable in terms of what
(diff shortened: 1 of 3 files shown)


state/watcher_profilecharm_test.go
```diff
@@ -206,12 +206,12 @@ func (s *instanceCharmProfileWatcherSuite) expectLoopCollectionFilterAndNotify(c
...
 
-func (s *instanceCharmProfileWatcherSuite) expectMergeCollectionInstanceField(state string) func() {
+func (s *instanceCharmProfileWatcherSuite) expectMergeCollectionInstanceField(value string) func() {
 	return func() {
 		s.database.EXPECT().GetCollection("instanceCharmProfileData").Return(s.collection, noop)
 		s.collection.EXPECT().Find(bson.D{{"_id", "1"}}).Return(s.query)
-		s.query.EXPECT().One(gomock.Any()).SetArg(0, map[string]interface{}{
-			"upgradecharmprofilecomplete": state,
+		s.query.EXPECT().One(gomock.Any()).SetArg(0, state.InstanceCharmProfileDataDoc{
+			UpgradeCharmProfileComplete: value,
```
SimonRichardson: This makes it worth it, as we no longer have to guess what the string value should be, esp. if somebody changes the name of the field.

SimonRichardson:
https://trello.com/c/IWlrt81O/1314-clean-up-watcher-john-feedback

state/watcher.go
```diff
@@ -2168,26 +2178,28 @@ func (w *documentFieldWatcher) initial() error {
...
 
 	logger.Tracef("Started watching %s for %v: %q", w.collection, w.members, field)
 	return nil
 }
 
 func (w *documentFieldWatcher) merge(change watcher.Change) (bool, error) {
-	if change.Revno == -1 {
+	// we care about change.Revno equalling -1 as we want to know about
+	// documents being deleted.
+	if w.known == nil {
```
jameinel: I think this fix is the wrong way around.
'w.known == nil' means that we think the document is deleted.
I think you want something more like:
```
if change.Revno == -1 {
 // treat this as the document being deleted
 if w.known != nil {
   w.known = nil
   return true, nil
 }
 return false, nil
}
```

state/watcher.go
```diff
@@ -2202,8 +2214,8 @@ func (w *documentFieldWatcher) merge(change watcher.Change) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	if w.known != currentField {
-		w.known = currentField
+	if *w.known != currentField {
```
jameinel: this would become:
if w.known == nil || *w.known != currentField {
...

state/watcher.go
```diff
@@ -2168,26 +2178,28 @@ func (w *documentFieldWatcher) initial() error {
...
 		return false, nil
 	}
 	col, closer := w.db.GetCollection(w.collection)
 	defer closer()
 
-	var data map[string]interface{}
+	data := w.document()
 	if err := col.Find(w.members).One(&data); err != nil {
 		if err != mgo.ErrNotFound {
 			logger.Debugf("%s NOT mgo err not found", w.collection)
```
jameinel: This should be treated the same as ch.Revno == -1

state/watcher.go
SimonRichardson: ah right, nps.

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

PR #10082, opened 2019-04-23
Add --no-wait for remove-machine CLI and api.

anastasiamac:
## Description of change

Machine removal is a multi-step process. Currently, Juju will not proceed to the next step until a previous step in the process is completed successfully. This may block the removal process - next steps are not called and the removal does not complete.

Force removal will call the next step in the process regardless of where the current step is or if it has succeeded. Generally, we'd wait for a set period of time before calling this next step. However, with 'no-wait' command option specified, Juju will forgo this wait and will call the next step is soon as possible. This means that the whole force removal process will be faster.

This PR has a MachineManager facade bump.

### DIFF
--- a/api/facadeversions.go
+++ b/api/facadeversions.go
@@ -62,7 +62,7 @@ var facadeVersions = map[string]int{
 	"LogForwarding":                1,
 	"Logger":                       1,
 	"MachineActions":               1,
-	"MachineManager":               5,
+	"MachineManager":               6,
 	"MachineUndertaker":            1,
 	"Machiner":                     1,
 	"MeterStatus":                  1,
--- a/api/machinemanager/machinemanager.go
+++ b/api/machinemanager/machinemanager.go
@@ -4,6 +4,8 @@
 package machinemanager
 
 import (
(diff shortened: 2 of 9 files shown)


anastasiamac:
[add --no-wait to remove-machine](https://trello.com/c/2sxs05Ps/704-add-no-wait-to-remove-machine)

anastasiamac:
@wallyworld,
So various reasons for having a separate struct - most importantly: it's a different method and we do not want it to be affected if the other method requires the struct to gain new parameters. To avoid this we have separate param struct dedicated to a call. I could throw out there a 'single responsibility' principal is a good thing to have in mind but am sure yo are aware.

To be honest, my concern when being in this code was 'why do we have a separate ForceDestroy call when we have DestroyWithParams'? As you said, it does everything we want and better :)

anastasiamac:
@wallyworld,
As per our discussion, I've changed the logic to use DestroyWithParams instead of ForceDestroy.

anastasiamac:
$$merge$$

PR #10084, opened 2019-04-23
Remove charm LXD Profile knowledge from uniter

hmlanigan:
## Description of change

Remove LXD Profile code from worker/uniter, 2nd part of InstanceMutater feature flag removal..  Requires PR 10072 to land first.

## QA steps

Deploy and upgrade charms with and without lxd profiles. Try bundles
with more than 5 machines, subordinates, containers: with and without
the lxd provider, local and charm store charms.

charms with an lxd profile:
./testcharms/charm-repo/quantal/lxd-profile-alt
./testcharms/charm-repo/quantal/lxd-profile-subordinate

juju deploy cs:nova-compute
juju deploy cs:neutron-openvswitch-252
juju add-relation nova-compute:neutron-plugin neutron-openvswitch:neutron-plugin
let settle, will be waiting for other relations
juju upgrade-charm neutron-openvswitch <-- adds an lxd profile


hmlanigan:
A follow on to this PR, remove from uniter apiserver for 2.6, why keeping compatibility for 2.5.x:
- LXDProfileAPI 
- RemoveUpgradeCharmProfileData
- WatchLXDProfileUpgradeNotifications

hmlanigan:
While in the process of pushing up another change to remove the uniter api pieces, my internet went down...  will upload when I can. The change doesn’t impact what’s currently up.

worker/uniter/resolver.go
```diff
@@ -69,14 +68,6 @@ func (s *uniterResolver) NextOp(
 		if localState.Conflicted {
 			return s.nextOpConflicted(localState, remoteState, opFactory)
 		}
-		op, err = s.config.UpgradeCharmProfile.NextOp(localState, remoteState, opFactory)
```
SimonRichardson: This is basically fine right, because you can never have a controller behind an agent. So new controllers would never emit the change event, so new agents never need this code as it'll be dead code anyway. That's my understanding of this PR?

hmlanigan: yes.  we only need to keep the watcher in state and compatibility in the apiserver layer for 2.5.x uniters to work with 2.6 controllers until those models can be upgraded.

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #10085, opened 2019-04-23
Add metrics and add life to unit multiwatcher.

howbazaar:
Adds metrics for number of machines, applications, and units.
As part of this it became clear that the multiwatcher delta for units wasn't passing the life information along. This was added to better match the expectations and consistency with the other entities that have life, like models, machines and applications.


howbazaar:
$$merge$$

PR #10094, opened 2019-04-25
Snapcraft: ignore errors, wait for microk8s

Veebers:
## Description of change

Updates the snap config with the latest changes:
  - Ignore errors when caching images (don't want to block install)
  - Give microk8s a moment to be ready (if it was installed just before Juju etc.)


Veebers:
$$merge$$

PR #10097, opened 2019-04-26
Weaving force into model-destroy backend.

anastasiamac:
## Description of change

Force has been added as an option to 'model-destroy' command in previous work. This PR ensures that backend is aware of it.

As a drive-by addressed some var names that collided with import names.

### DIFF
--- a/state/model.go
+++ b/state/model.go
@@ -1198,7 +1198,7 @@ func (m *Model) destroyOps(
 				if errors.IsNotFound(err) {
 					continue
 				}
-				// TODO (anastasiamac 2019-4-24) we should not break out here but
+				// TODO (force 2019-4-24) we should not break out here but
 				// continue with other models.
 				return nil, errors.Trace(err)
 			}
@@ -1206,7 +1206,7 @@ func (m *Model) destroyOps(
 
 			model, err := newSt.Model()
 			if err != nil {
-				// TODO (anastasiamac 2019-4-24) we should not break out here but
+				// TODO (force 2019-4-24) we should not break out here but
 				// continue with other models.
 				return nil, errors.Trace(err)
 			}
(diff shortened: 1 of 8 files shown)


anastasiamac:
[destroy-model --force wired in the backend](https://trello.com/c/TFmyNY65/713-destroy-model-force-wired-in-the-backend)

state/model.go
```diff
@@ -1214,6 +1234,8 @@ func (m *Model) destroyOps(
 				aliveEmpty++
 			default:
 				if !IsModelNotEmptyError(err) {
+					// TODO (anastasiamac 2019-4-24) we should not break out here but
```
wallyworld: could we make the TODOs something like TODO(force) so we can grep for all related TODOs for a given unit of work regardless of who wrote the code?

state/model.go
```diff
@@ -1242,8 +1264,10 @@ func (m *Model) destroyOps(
 
 	var ops []txn.Op
 	modelOp := txn.Op{
-		C:      modelsC,
-		Id:     modelUUID,
+		C:  modelsC,
+		Id: modelUUID,
+		// TODO (anastasiamac 2019-4-24) when forcing, we might just consider if the document exists
+		//  instead of checking Life.
 		Assert: isAliveDoc,
```
wallyworld: This can fail with --force.
If --force is used, the assert needs to be that the life equals the current life (as was done in places for force destroying units)

state/model.go
anastasiamac: This violates our TODO writing convention but I like it better :) I'll change

state/model.go
anastasiamac: Although, truthfully most of these TODOs are not TODOs but a reminder that maybe we should not get out of the loop if encountering an error but keep going with other elements.

anastasiamac:
$$merge$$

anastasiamac:
$$merge$$

PR #10098, opened 2019-04-26
Feature flags mutate

SimonRichardson:
## Description of change

The following allows feature flags to mutate in some tests. The
following changes[1] unfortunately prevented this from happening
as the code used a frozenset which is an immutable set, rather
than a mutable set.

 1. https://github.com/juju/juju/commit/3fc87c5ff6a8e1da7e8416ffcf0b26d150f5c891#diff-4a7d0528501759a68ec4f4bf568c2ecc

## QA steps

Setup acceptance tests running the following:

```
./assess_model_migration_versions.py
```

### DIFF
--- a/acceptancetests/jujupy/client.py
+++ b/acceptancetests/jujupy/client.py
@@ -715,7 +715,7 @@ class ModelClient:
         """
         self.env = env
         if _backend is None:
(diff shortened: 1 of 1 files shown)


SimonRichardson:
$$merge$$

PR #10099, opened 2019-04-26
Keep track of migration targets to show to users who need to login

achilleasa:
## Description of change

This PR contributes a UX enhancement to the changes introduced by #10069. When attempting to run any model-related command against a model that has been migrated to a different controller that is **not** locally known, juju would display an error message like `juju login $ip -c new-controller`. 

While this works, it is not the most user-friendly approach. For example, if the user runs the command _as-is_ twice for a different controller IP or they already have a controller named `new-controller` they will get an error.

To improve the UX for this use-case, this PR threads the controller name, the local `alias` used by the user initiating the migration, through the `InitiateMigration` API call and ensures it gets stored in the migration document for the model. When the server emits a `RedirectError` it includes the alias in the error payload which allows the client to display a more sensible value for the `-c` argument to `juju login`.

The alias itself is optional (older controllers do not keep track of it); the redirect error handling code will fall-back to the default `new-controller` value for the controller name if no alias is available. To avoid user confusion, the error wording has been tweaked slightly to indicate that the user is free to swap the `-c` argument value with any other value that makes sense for the user.

One could argue that the controller alias may have different meaning for the user that initiating the model migration vs the user who tries to run a command against the migrated model. However, we already do something similar for the [juju register](https://github.com/juju/juju/blob/develop/jujuclient/registration.go#L23) command.

## QA steps

```console
$ juju bootstrap lxd src
$ juju change-user-password 

$ juju bootstrap lxd dst
$ juju change-user-password

$ juju switch src:controller
$ juju add-model migrate

# At this point make a copy of your ~/.local/share/juju/models.yaml
# as the model uuid will probably be removed from src after migrating
$ juju migrate src:migrate dst

# wait a few seconds for the migration to complete and ensure that 
# the source controller in your local models.yaml still contains the 
# uuid for the model that just got migrated 

$ juju unregister dst

# Try any model-related command, e.g:
$ juju status -m migrate
ERROR Model "migrate" has been migrated to another controller.
To access it run one of the following commands (you can replace the -c argument with your own preferred controller name):
  'juju login 10.65.47.40:17070 -c dst'

New controller fingerprint [93:73:88:C9:9A:AA:EC:C8:85:AE:D1:33:E5:92:CE:95:0F:B6:00:82:21:CB:8C:A0:42:16:29:77:CF:6D:B6:D4]
```

## Documentation changes

@pmatulis there is a slight change in the error output (the bit in parentheses); we should probably update the docs to reflect the change.

### DIFF
--- a/api/apiclient.go
+++ b/api/apiclient.go
@@ -168,6 +168,11 @@ type RedirectError struct {
 	// FollowRedirect is set to true for cases like JAAS where the client
 	// needs to automatically follow the redirect to the new controller.
 	FollowRedirect bool
+
+	// An optional alias for the controller the model got redirected to. It
+	// can be used by the client to present the user with a more meaningful
+	// juju login -c XYZ command
+	ControllerAlias string
 }
 
 func (e *RedirectError) Error() string {
--- a/api/controller/controller.go
+++ b/api/controller/controller.go
@@ -260,13 +260,14 @@ func (c *Client) ConfigSet(values map[string]interface{}) error {
 // MigrationSpec holds the details required to start the migration of
 // a single model.
 type MigrationSpec struct {
-	ModelUUID            string
-	TargetControllerUUID string
-	TargetAddrs          []string
-	TargetCACert         string
-	TargetUser           string
-	TargetPassword       string
-	TargetMacaroons      []macaroon.Slice
+	ModelUUID             string
+	TargetControllerUUID  string
+	TargetControllerAlias string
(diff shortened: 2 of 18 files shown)


achilleasa:
!!build!!

mitechie:
QA ok, thanks for the update. 

One note, the multiple login commands is good because we don't know what IP the client needs to use, but wish we could do better.

achilleasa:
$$merge$$

PR #10101, opened 2019-04-28
Wait longer for expected events.

howbazaar:
On slower machines we don't always get the expected events in 50ms. This is why as a general rule, we wait for a `LongWait` for events we expect, even if we expect them quickly.

### DIFF
--- a/core/watcher/notify_test.go
+++ b/core/watcher/notify_test.go
@@ -190,12 +190,12 @@ func waitForTimeout(c *gc.C, ch <-chan error, timeout time.Duration) error {
 	return nil
 }
 
-func waitShort(c *gc.C, w worker.Worker) error {
+func waitForWorkerStopped(c *gc.C, w worker.Worker) error {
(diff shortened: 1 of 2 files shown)


howbazaar:
Fixes this failure we saw in the race tests:
```
FAIL: notify_test.go:290: notifyWorkerSuite.TestHandleErrorStopsWorkerAndWatcher

notify_test.go:295:
    err := waitShort(c, w)
notify_test.go:188:
    c.Errorf("timed out waiting to receive a change after %s", timeout)
... Error: timed out waiting to receive a change after 50ms

notify_test.go:296:
    c.Check(err, gc.ErrorMatches, "my handling error")
... value = nil
... regex string = "my handling error"
... Error value is nil
```

howbazaar:
$$merge$$

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #10102, opened 2019-04-28
Leave scope with force when forcing.

anastasiamac:
## Description of change

Force cleanup can and should call LeaveScopeWithForce(..)


anastasiamac:
$$merge$$

PR #10103, opened 2019-04-29
Actions ignore blocks

timClicks:
## Description of change

User feedback indicates that blocks should be ignored for `run-action`.

## QA steps

Deploying a charm that takes actions, run `juju disable-commands` and then `run-action`. Success is indicated by the lack of an error. An example session:

**`juju bootstrap localhost ignore-blocks`**
...
Initial model "default" added.

**`juju deploy git`**
Located charm "cs:trusty/git-1".
Deploying charm "cs:trusty/git-1".

**`juju run-action git/0 add-repo repo=https://github.com/juju/juju.git`**
Action queued with id: 7c1cf3f7-16dd-46d0-850c-71464267a01

**`juju disable-command all`**

**`juju run-action git/0 add-repo repo=https://github.com/timClicks/juju.git`**
Action queued with id: 174a8942-61c8-4e64-8d4c-960643858d77

## Documentation changes

Will be required. CLI is now more powerful, as actions now have their guard rails removed.

## Bug reference
I think this are related bugs:
https://bugs.launchpad.net/juju/+bug/1822017
https://bugs.launchpad.net/juju/+bug/1794891

### DIFF
--- a/apiserver/facades/client/action/action.go
+++ b/apiserver/facades/client/action/action.go
@@ -209,10 +209,6 @@ func (a *ActionAPI) Enqueue(arg params.Actions) (params.ActionResults, error) {
 		return params.ActionResults{}, errors.Trace(err)
 	}
 
-	if err := a.check.ChangeAllowed(); err != nil {
-		return params.ActionResults{}, errors.Trace(err)
-	}
-
 	var leaders map[string]string
 	getLeader := func(appName string) (string, error) {
 		if leaders == nil {
@@ -305,10 +301,6 @@ func (a *ActionAPI) ListCompleted(arg params.Entities) (params.ActionsByReceiver
 // Cancel attempts to cancel enqueued Actions from running.
 func (a *ActionAPI) Cancel(arg params.Entities) (params.ActionResults, error) {
 	if err := a.checkCanWrite(); err != nil {
-		return params.ActionResults{}, errors.Trace(err)
-	}
-
-	if err := a.check.ChangeAllowed(); err != nil {
 		return params.ActionResults{}, errors.Trace(err)
 	}
 
--- a/apiserver/facades/client/action/action_test.go
+++ b/apiserver/facades/client/action/action_test.go
@@ -8,7 +8,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/juju/errors"
 	jc "github.com/juju/testing/checkers"
 	gc "gopkg.in/check.v1"
 	"gopkg.in/juju/names.v2"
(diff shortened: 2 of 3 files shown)


apiserver/facades/client/action/action.go
```diff
@@ -308,10 +302,6 @@ func (a *ActionAPI) Cancel(arg params.Entities) (params.ActionResults, error) {
 		return params.ActionResults{}, errors.Trace(err)
 	}
 
-	if err := a.check.ChangeAllowed(); err != nil {
```
anastasiamac: I am not sure that we want to allow cancelling the action when the change block is on. I think we might want to have this check.

timClicks: @wallyworld thoughts?

howbazaar: Yes we do. Starting and cancelling actions are orchestration and devops tasks. Not model changes.

apiserver/facades/client/action/run.go
```diff
@@ -81,9 +81,6 @@ func (a *ActionAPI) Run(run params.RunParams) (results params.ActionResults, err
 	if err := a.checkCanAdmin(); err != nil {
 		return results, err
 	}
-	if err := a.check.ChangeAllowed(); err != nil {
```
howbazaar: `juju run` is different. We should still block this. We only want to open up actions.

timClicks: Fixed

apiserver/facades/client/action/action.go
timClicks: Per discussion in IRC, retaining change.

apiserver/facades/client/action/action_test.go
```diff
@@ -102,26 +101,18 @@ func (s *actionSuite) SetUpTest(c *gc.C) {
...
-		Code:    "operation is blocked",
-	})
-}
-
 func (s *actionSuite) TestBlockEnqueue(c *gc.C) {
 	// block all changes
 	s.BlockAllChanges(c, "Enqueue")
 	_, err := s.action.Enqueue(params.Actions{})
-	s.AssertBlocked(c, err, "Enqueue")
+	c.Assert(err, jc.ErrorIsNil)
```
wallyworld: Strictly speaking, you should actually test that the Enqueue has happened - this test as written will pass but does not guarantee correct behaviour. You can simply add the BlockAllChanges() call to the existing Enqueue test and delete this test. Same with the other test

timClicks: Makes perfect sense.

wallyworld:
[setting blocked should still allow run-action and juju-run](https://trello.com/c/qBDLjZn6/694-setting-blocked-should-still-allow-run-action-and-juju-run)

timClicks:
[Intermittent failure](http://ci.jujucharms.com/job/make-check-juju/4723/)

> 2019-04-29 03:32:48 WARNING juju.testing mgo.go:216 failed to start mongo: mongod failed to listen on port 39696

!!build!!

timClicks:
$$merge$$

timClicks:
$$merge$$

PR #10105, opened 2019-04-29
Fix/operator agentversion never updated

ycliuhw:
## Description of change

Fix: operators' `upgradedToVersion` in `agent.cfg` is not updated after `uprade-model`;

## QA steps

- prepare caas model;
- deploy workloads;
- upgrade-controller;
- upgrade-model;
- check `upgradedToVersion`;

```bash
$ mkubectl exec gitlab-k8s-operator-0  -n t1 -it cat /var/lib/juju/agents/application-gitlab-k8s/agent.conf | grep upgradedToVersion
```

## Documentation changes

None;

## Bug reference

No, not released yet;

### DIFF
--- a/worker/upgradesteps/manifold.go
+++ b/worker/upgradesteps/manifold.go
@@ -13,7 +13,6 @@ import (
 	"github.com/juju/juju/api"
 	apiagent "github.com/juju/juju/api/agent"
 	"github.com/juju/juju/state"
-	"github.com/juju/juju/state/multiwatcher"
 	"github.com/juju/juju/worker/gate"
 )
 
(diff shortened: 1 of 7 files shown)


ycliuhw:
[operator agent.conf targetAgentVersion has to be updated after model upgrade](https://trello.com/c/VEIdcqvI/709-operator-agentconf-targetagentversion-has-to-be-updated-after-model-upgrade)

worker/upgradesteps/manifold.go
```diff
@@ -56,32 +58,39 @@ func Manifold(config ManifoldConfig) dependency.Manifold {
...
-			if err != nil {
-				return nil, err
-			}
-			entity, err := agentFacade.Entity(agent.CurrentConfig().Tag())
-			if err != nil {
-				return nil, err
+			var jobs []multiwatcher.MachineJob
+			if names.IsValidApplication(agent.CurrentConfig().Tag().String()) {
+				// application tag for CAAS operator.
+				jobs = []multiwatcher.MachineJob{}
```
wallyworld: We don't need an empty slice do we?
Can't we just leave it as nil? If not let's fix it so we can and avoid the multiwatcher import.

worker/upgradesteps/manifold.go
```diff
@@ -56,32 +58,39 @@ func Manifold(config ManifoldConfig) dependency.Manifold {
...
 			// TODO(fwereade): use appropriate facade!
-			agentFacade, err := apiagent.NewState(apiConn)
-			if err != nil {
-				return nil, err
-			}
-			entity, err := agentFacade.Entity(agent.CurrentConfig().Tag())
-			if err != nil {
-				return nil, err
+			var jobs []multiwatcher.MachineJob
+			if names.IsValidApplication(agent.CurrentConfig().Tag().String()) {
```
wallyworld: You can just check Tag().Kind(), less verbose. Here and below

worker/upgradesteps/worker.go
```diff
@@ -209,6 +209,9 @@ func (w *upgradesteps) run() error {
 			return errors.Trace(err)
 		}
 		w.isCaas = model.Type() == state.ModelTypeCAAS
+	} else {
+		// application tag for CAAS operator.
+		w.isCaas = names.IsValidApplication(w.agent.CurrentConfig().Tag().String())
```
wallyworld: If we are doing htis can we just have one check for isCass that can be used for controller or not.
Having 2 checks is a bit clunky.

worker/upgradesteps/manifold.go
ycliuhw: sure, fixed,

worker/upgradesteps/manifold.go
ycliuhw: yup, now it's using Kind to check

worker/upgradesteps/worker.go
ycliuhw: yup, moved this to manifold then pass in

ycliuhw:
$$merge$$

PR #10109, opened 2019-04-29
2.6 application cleanup 1794602

jameinel:
## Description of change

This cleans up application destruction. We have a bug in 2.6 when removing the last unit. If it removes the final relation at the same time that it removes the final unit, it ends up failing to destroy the application. Rather than try to be extra careful about each removal operation in case *this* one needs to do the removal, we now queue up a cleanup step that just checks if it is time to destroy the application.

This also changes app.Destroy so that it is no longer a no-op if the application is already Dying. We already had removed the no-op in the case that we supplied Force. But we can do it always. We are already checking all the references have been removed before we actually remove it, so it is still safe to do. And it means the Cleanup step is just a call to Destroy, which already has all the logic of how to do a clean teardown.

## QA steps

The network health CI tests were failing regularly. The included Unit Test was also failing to remove the application when you remove an app that has 1 unit and 1 relation.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1794602


mitechie:
pre-check error:

    11:10:54 state/cleanup.go:410: Logger.Tracef format %s has arg app.UnitCount() of wrong type int

state/cleanup.go
```diff
@@ -385,6 +388,46 @@ func (st *State) cleanupStorageForDyingModel(cleanupArgs []bson.Raw) (err error)
 	return nil
 }
 
+// cleanupApplication checks if all references to a dying application have been removed,
+// and if so, removes the application.
+func (st *State) cleanupApplication(applicationname string, cleanupArgs []bson.Raw) (err error) {
+	app, err := st.Application(applicationname)
```
howbazaar: I like the tracing you have added.

state/cleanup.go
```diff
@@ -385,6 +388,46 @@ func (st *State) cleanupStorageForDyingModel(cleanupArgs []bson.Raw) (err error)
...
+			logger.Tracef("cleanupApplication(%s): application already gone", applicationname)
+			return nil
+		}
+		return errors.Trace(err)
+	}
+	if app.Life() == Alive {
+		return errors.BadRequestf("cleanupApplication requested for an application (%s) that is still alive", applicationname)
+	}
+	// We know the app is at least Dying, so check if the unit/relation counts are no longer referencing this application.
+	if app.UnitCount() > 0 || app.RelationCount() > 0 {
```
howbazaar: It seems that here we are caring about relation count being > 0, but we only ever seem to add the cleanup job when removing a unit. To me this feels like we have either got our logic wrong in the expectation that relation count can be non-zero when units are zero, or that we need to also add an application cleanup job when we decrement the relation count.

state/cleanup_test.go
```diff
@@ -43,6 +43,12 @@ func (s *CleanupSuite) SetUpTest(c *gc.C) {
 
 }
 
+func (s *CleanupSuite) TestCleanupDyingApplicationNoUnits(c *gc.C) {
+	mysql := s.AddTestingApplication(c, "mysql", s.AddTestingCharm(c, "mysql"))
+	c.Assert(mysql.Destroy(), jc.ErrorIsNil)
```
howbazaar: Who is running the cleanup here? Is there a potential race?

state/relation.go
```diff
@@ -513,36 +514,32 @@ func (r *Relation) removeLocalEndpointOps(ep Endpoint, departingUnitName string,
...
-			// If the 'force' is not set, any error will be fatal and no operations will be returned.
-			return app.removeOps(hasLastRef, op)
-		} else if err != mgo.ErrNotFound {
-			if !op.Force {
-				return nil, err
+		asserts = append(hasRelation)
+		var appDoc applicationDoc
+		if err := applications.FindId(ep.ApplicationName).One(&appDoc); err == nil {
+			if appDoc.Life != Alive {
+				cleanupOps = append(cleanupOps, newCleanupOp(
```
howbazaar: Oh... here we go. Happy now.

state/cleanup.go
jameinel: I'm pretty sure we do add a cleanup when we decrement the relation count. If you grep the diff you can see 2 calls to "newCleanupOp". One is done during RemoveUnitOperation.removeOps (which ultimately calls Application.removeUnitOps) and the other is done during Relation.removeLocalEndpointOps.

state/cleanup_test.go
jameinel: so this is the code that fast-paths Destroy to immediately destroy the application if there are no units and/or relations. Nothing calls cleanup. Ultimately Cleanup effectively just calls app.Destroy under that same circumstance.

jameinel:
$$merge$$

jameinel:
$$merge$$

jameinel:
The failure was:
```
----------------------------------------------------------------------
FAIL: worker_test.go:314: workerEnvironSuite.TestCharmProfilingInfoError

TRACE received tag "machine-0"
TRACE watching change on MutaterMachine 0
worker_test.go:326:
    c.Assert(err, jc.Satisfies, params.IsCodeNotSupported)
... obtained = nil
... func(T) bool func(error) bool = (func(error) bool)(0x8dda20)
```

I can get that to fail here. I'm guessing it is intermittent?

$$merge$$

jameinel:
$$merge$$

jameinel:
$$merge$$

jameinel:
$$merge$$

PR #10110, opened 2019-04-29
Refactor most of network into core/network.

hmlanigan:
## Description of change

Follow on from PR 9916: Refactor network, iptables, netplan, debinterfaces, and ssh packages into core.  containerizer was left in place, as it has imports which should not be dragged into core.

## QA steps

All tests etc should work as normal.


core/network/debinterfaces/activate.go
```diff
@@ -11,12 +11,13 @@ import (
 	"time"
 
 	"github.com/juju/clock"
-	"github.com/juju/juju/utils/scriptrunner"
+	"github.com/juju/errors"
 	"github.com/juju/loggo"
-	"github.com/pkg/errors"
+
+	"github.com/juju/juju/utils/scriptrunner"
```
howbazaar: This is the bit that I'm hesitant on. But perhaps I shouldn't be?

hmlanigan:
John requested a different approach to what is refactored from network to core/network.  Will start over with a new PR.

PR #10113, opened 2019-04-30
Remove old state metrics.

howbazaar:
The old state metrics were very slow as they quieried many documents. The cache now has these metrics.


howbazaar:
$$merge$$

PR #10116, opened 2019-04-30
K8s resource storage fixes

wallyworld:
## Description of change

There are 2 fixes here in 2 commits.
1. Add support for OCI docker images in the juju attach-resource command.
All of the backend APIs are already in place; the change is to the command itself. It makes an additional
API call to get the charm resource metadata and properly handles file resources vs OCI image resources.
There's a file rename which bloats the diff a bit.

2. Fix k8s storage provisioning - storage other than k8s (eg tmpfs, rootfs) was broken. Also, fix support for scaling applications to 0, as it had been disabled since storage relied on having on unit. Now when storage is provisioned, we don't rely on looking at unit storage metadata but get the storage info directly from the application metadata. This is much cleaner and less complicated.

## QA steps

bootstrap k8s
deploy mariadb-k8s
juju scale-application mariadb-k8s 0
(check k8s cluster that pods are gone, and also units gone)
juju scale-application mariadb-k8s 1
(check that a new unit and pod are created)
juju deploy mariadb-k8s --storage database=tmpfs
(kubectl describe pod to see that emptyDir storage is used)
juju attach-resource mariadb-k8s mysql_image=mariadb
(check mongo db that resource has been updated)


wallyworld:
[k8s tmpfs/rootfs provisioning broken](https://trello.com/c/3YG1eRo3/730-k8s-tmpfs-rootfs-provisioning-broken)

wallyworld:
[juju attach-resource for k8s](https://trello.com/c/RUQxOLoH/727-juju-attach-resource-for-k8s)

apiserver/facades/controller/caasunitprovisioner/provisioner.go
```diff
@@ -254,47 +259,25 @@ func (f *Facade) provisioningInfo(model Model, tagString string) (*params.Kubern
...
 
-		// The juju-storage-owner tag is set to the unit. We use it as a label on the CAAS volume.
-		// Since we used an arbitrary unit to get the info, reset the tag to the application.
-		for _, fsp := range filesystemParams {
-			fsp.Tags[tags.JujuStorageOwner] = appTag.Id()
-		}
+	// The juju-storage-owner tag is set to the application.
+	// We use it as a label on the CAAS volume.
+	for _, fsp := range filesystemParams {
+		fsp.Tags[tags.JujuStorageOwner] = appTag.Id()
```
ycliuhw: it's already done in `func filesystemParams`
```
filesystemTags[tags.JujuStorageOwner] = app.Name()
```

ycliuhw:
and the test failed

apiserver/facades/controller/caasunitprovisioner/provisioner.go
wallyworld: ah yes, thanks

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10118, opened 2019-04-30
Not provisioned instance mutater

SimonRichardson:
## Description of change

The machine that we're currently watching may not be provisioned,
if that's the case we need to loop until more changes are sent.
Unfortunately we may actually end up with missing the changes that
are being sent and so the machine might end up being provisioned
incorrectly. We need to think/solve this.

## QA steps

TBA

### DIFF
--- a/core/cache/lxdprofilewatcher.go
+++ b/core/cache/lxdprofilewatcher.go
@@ -40,14 +40,15 @@ type appInfo struct {
 }
 
 type MachineAppLXDProfileConfig struct {
-	appTopic        string
-	unitAddTopic    string
-	unitRemoveTopic string
-	machine         *Machine
-	modeler         MachineAppModeler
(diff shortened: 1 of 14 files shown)


core/cache/lxdprofilewatcher.go
```diff
@@ -75,21 +77,21 @@ func newMachineAppLXDProfileWatcher(config MachineAppLXDProfileConfig) (*Machine
 
 	err := w.init(config.machine)
 	if err != nil {
-		// Stop the watcher, which unsubcribes above.
+		// Stop the watcher, which unsubscribes above.
```
SimonRichardson: nice drive by

SimonRichardson:
QA'd and did a code review.

hmlanigan:
$$merge$$

PR #10119, opened 2019-04-30
Server-side migration pre-checks to ensure model users are present in destination before migrating

achilleasa:
## Description of change

This PR implements the server-side equivalent of the changes introduced by #10092 because as we all know: _you can never trust the client_.

Side-note: the facade tests for the InitiateMigration RPC tend to patch the pre-check code so it returns a nil value. As the introduced code runs as part of that pre-check and it is also expected to make extra calls to the target controller I have added an internal (in the `controller` package) test-suite so I could just check the user comparison logic without having to set up elaborate multi-controller tests.

## QA steps

**IMPORTANT**: you must bootstrap the controllers with this branch but when running juju commands you need to use an **older** client (e.g 2.5). 

### Happy path 1: shared local users

```console
$ juju bootstrap lxd src
$ juju add-model mig
$ juju add-user foo 
$ juju grant foo read mig

$ juju bootstrap lxd dst
$ juju add-user foo 

$ juju migrate src:mig dst
 Migration started with ID "45967f61-1991-4340-8c14-f3cdf2682d35:0"
```

### Happy path 2: shared local users and same identity providers 

```console
$ juju bootstrap lxd src --config identity-url=https://api.jujucharms.com/identity --config allow-model-access=true
$ juju add-model mig
$ juju add-user foo 
$ juju grant foo read mig
$ juju grant me@external write mig

$ juju bootstrap lxd dst --config identity-url=https://api.jujucharms.com/identity --config allow-model-access=true
$ juju add-user foo 

$ juju migrate src:mig dst
Migration started with ID "fe066fbe-cec4-4efd-8e46-e77465e08c22:0"
```
### Local user missing from target

```console
$ juju bootstrap lxd src
$ juju add-model mig
$ juju add-user foo 
$ juju grant foo read mig

$ juju bootstrap lxd dst

$ juju migrate src:mig dst
ERROR cannot initiate migration as the users granted access to the model do not exist
on the destination controller. To resolve this issue you can add the following
users to the destination controller or remove them from the current model:
  - foo
```

### Shared local users, different identity providers and granted access to external user

```console
$ juju bootstrap lxd src --config identity-url=https://api.jujucharms.com/identity --config allow-model-access=true
$ juju add-model mig
$ juju add-user foo 
$ juju grant foo read mig
$ juju grant me@external write mig

$ juju bootstrap lxd dst --config identity-url=https://candid.provider/identity --config allow-model-access=true
$ juju add-user foo

$ juju migrate src:mig dst
ERROR cannot initiate migration as external users have been granted access to the model
and the two controllers have different identity provider configurations. To resolve
this issue you can remove the following users from the current model:
  - me@external
```

### Not shared local users, different identity providers and granted access to external user

```console
$ juju bootstrap lxd src --config identity-url=https://api.jujucharms.com/identity --config allow-model-access=true
$ juju add-model mig
$ juju add-user foo 
$ juju grant foo read mig
$ juju grant me@external write mig

$ juju bootstrap lxd dst --config identity-url=https://candid.provider/identity --config allow-model-access=true

$ juju migrate src:mig dst
ERROR cannot initiate migration as external users have been granted access to the model
and the two controllers have different identity provider configurations. To resolve
this issue you need to remove the following users from the current model:
  - me@external

and add the following users to the destination controller or remove them from
the current model:
  - foo
```

### DIFF
--- a/apiserver/facades/client/controller/controller.go
+++ b/apiserver/facades/client/controller/controller.go
@@ -8,15 +8,19 @@ package controller
 import (
 	"encoding/json"
 	"sort"
+	"strings"
 
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	"github.com/juju/txn"
+	"github.com/juju/utils/set"
 	"gopkg.in/juju/names.v2"
 	"gopkg.in/macaroon.v2-unstable"
 
 	"github.com/juju/juju/api"
+	controllerclient "github.com/juju/juju/api/controller"
 	"github.com/juju/juju/api/migrationtarget"
+	"github.com/juju/juju/api/usermanager"
 	"github.com/juju/juju/apiserver/common"
 	"github.com/juju/juju/apiserver/common/cloudspec"
 	"github.com/juju/juju/apiserver/facade"
@@ -656,8 +660,15 @@ var runMigrationPrechecks = func(st, ctlrSt *state.State, targetInfo *coremigrat
 		return errors.Annotate(err, "connect to target controller")
 	}
 	defer conn.Close()
-	modelInfo, err := makeModelInfo(st, ctlrSt)
+	modelInfo, srcUserList, err := makeModelInfo(st, ctlrSt)
 	if err != nil {
+		return errors.Trace(err)
+	}
+	dstUserList, err := getTargetControllerUsers(conn)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if err = srcUserList.checkCompatibilityWith(dstUserList); err != nil {
 		return errors.Trace(err)
 	}
 	client := migrationtarget.NewClient(conn)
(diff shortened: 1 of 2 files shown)


apiserver/facades/client/controller/controller.go
```diff
@@ -676,39 +687,172 @@ var runMigrationPrechecks = func(st, ctlrSt *state.State, targetInfo *coremigrat
...
 		Name:                   model.Name(),
 		Owner:                  model.Owner(),
 		AgentVersion:           agentVersion,
 		ControllerAgentVersion: controllerVersion,
-	}, nil
+	}, ul, nil
+}
+
+func getTargetControllerUsers(controllerUUID, controllerAlias string, conn api.Connection) (userList, error) {
+	var ul = userList{controllerUUID: controllerUUID}
```
SimonRichardson:
```suggestion
	ul := userList{controllerUUID: controllerUUID}
```

mitechie:
QA 

happy path 1
---
works as intended

Local user missing from target
---
    ERROR cannot initiate migration of model "5a6c2fc8-6662-4396-89fe-b3c656d85772" to
controller "513c0adb-8b15-4eff-82da-05419ef7b24e" as some of the model's users
do not exist in the target controller. To resolve this issue you can either add
the following list of users to "513c0adb-8b15-4eff-82da-05419ef7b24e" or remove
them from model "5a6c2fc8-6662-4396-89fe-b3c656d85772":
    - bar

Maybe we can leave out the model uiuds if we can just say "Cannot initial model migration as the following users are missing on the destination controller: bar, baz, ..." and leave out the controller/model uuid business? The user's specified the names of the controller/model in the command they entered so I don't think there's a lack of context that the error message needs to help with aside from the user names that need updating.

mitechie:
Along those lines, suggested error messaging updates:

    ERROR cannot initiate migration of model "f3daca2f-c047-4a64-895f-ebef85544b30" to
    controller "016c46e6-b64f-49f9-8fec-c0e62e7fcc03" as external users
    have been granted access to the model and the two controllers have different
    identity provider configurations. To resolve this issue you can remove the
    following list of external users from model "f3daca2f-c047-4a64-895f-ebef85544b30":
    - me@external

vs: 

        ERROR cannot initiate migration as the users granted access to the model do not exist on the destination controller. 
    To resolve this issue you can add the users to the destination controller or remove them from the current model. 
    Users: me@external

this would be generic and reusable for either case. I wouldn't worry about splitting the user lists as the command to remove a grant would be the same and the external users have their @xxxx extension on them. 

    ERROR cannot initiate migration of model "b573aef9-de92-43b7-8979-4aead46bdba8" to
    target controller "016c46e6-b64f-49f9-8fec-c0e62e7fcc03" as external users
    have been granted access to the model and the two controllers have different
    identity provider configurations. Additionally, some of the model's local users
    do not exist in the target controller. To resolve this issue you need to remove
    the following external users from model "b573aef9-de92-43b7-8979-4aead46bdba8":
      - me@external
    
    and either add the following list of local users to "016c46e6-b64f-49f9-8fec-c0e62e7fcc03"
    or remove them from model "b573aef9-de92-43b7-8979-4aead46bdba8":
      - foo

    ERROR cannot initiate migration as the users granted access to the model do not exist on the destination controller. 
    To resolve this issue you can add the users to the destination controller or remove them from the current model. 
    Users: me@external, foo

So I'd suggest in all error cases just having that common error with the list of users.

achilleasa:
@mitechie That wording seems much better for a server-side error. 

What about the case where the two controllers have **different** identity providers and the model grants access to an external user? The "add users to the destination controller" bit would be misleading in this case as the pre-checks would still block the migration (me@external may be a different user with the same name)

achilleasa:
@mitechie Also, are we happy with the wording for the client-side errors remaining as-is (see: #10092) or do they also need to be changed for consistency reasons?

mitechie:
Ideally t would be consistent but can be updated after the fact. I think
the different identity URL case is a corner exception. Let's go with the
update message and let that case be a bit odd.

achilleasa:
Pushed https://github.com/juju/juju/pull/10119/commits/031bc8823699f70aa49832fd1970c0c316a5f37d to address the above concerns re: error wording. Error messages are now UUID-free!

achilleasa:
$$merge$$

achilleasa:
$$merge$$

PR #10123, opened 2019-05-01
Add remove-machine acceptance tests

timClicks:
## Description of change

Add acceptance tests for removing machines. The intention is to ensure that `--force` is able to actually  remove machines that get stuck in the pending state.
 
## QA steps

* `cd $GOPATH/acceptancetests`
* `./assess remove_machine`
* ..wait for success

## Documentation changes

n/a

## Bug reference

n/a

### DIFF
--- a/acceptancetests/assess_remove_machine.py
+++ b/acceptancetests/assess_remove_machine.py
@@ -1,0 +1,83 @@
+#! /usr/bin/env python2
+
+import argparse
+import logging
+import typing
+import yaml
+
+from deploy_stack import BootstrapManager
+from utility import add_basic_testing_arguments, add_arg_juju_bin, configure_logging
+
+Optional = typing.Optional
+log = logging.getLogger("remove_machine")
+
+def list_machines(client):
+    machines = yaml.safe_load(
+        client.get_juju_output(
+            'list-machines', '--format=yaml'))['machines']
+    return machines
+
+def assess_remove_machine(client, series):
+    """
+    assess_remove_machine checks to see whether
+    juju can (add and) remove machines.
+
+    :param client: The jujupy client
+    :type client: jujupy.Client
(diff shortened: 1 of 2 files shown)


acceptancetests/utility.py
```diff
@@ -322,10 +322,17 @@ def add_basic_testing_arguments(
 add_basic_testing_arguments.__test__ = False
 
 
-def configure_logging(log_level):
+def configure_logging(log_level, logger=None):
```
timClicks: This change enables us to alter specific loggers, which are used within some of the `assess_*.py` files.

acceptancetests/assess_remove_machine.py
```diff
@@ -0,0 +1,85 @@
+import argparse
```
SimonRichardson: Missing hash bang, although not strictly required, we should include them, until all of the acceptance tests are migrated to python3.

acceptancetests/assess_remove_machine.py
```diff
@@ -0,0 +1,85 @@
...
+import typing
+import yaml
+
+from deploy_stack import BootstrapManager
+from utility import add_basic_testing_arguments, add_arg_juju_bin, configure_logging
+
+Optional = typing.Optional
+log = logging.getLogger("remove_machine")
+
+if __name__ != "__main__":
```
SimonRichardson: Why change from the common main entry point?

timClicks: Sorry, should have used the project's conventions. I like to abort early because it avoids wasted work when the file is imported.

acceptancetests/assess_remove_machine.py
timClicks: Good point, have added.

timClicks:
$$merge$$

PR #10127, opened 2019-05-02
Fixes intermittent failure in CAASProvisionerSuite.TestProvisioningInfo.

anastasiamac:
## Description of change

Map order is not guaranteed and some tests can fail intermittently because of it. This especially affects tests where we are checking more than 1 item.

This PR fixes a failure observed in http://ci.jujucharms.com/job/make-check-juju/4853/testReport/junit/github/com_juju_juju_apiserver_facades_controller_caasunitprovisioner/TestAll/

With this PR 71 successful runs under stress.

### DIFF
--- a/apiserver/facades/controller/caasunitprovisioner/provisioner_test.go
+++ b/apiserver/facades/controller/caasunitprovisioner/provisioner_test.go
@@ -190,62 +190,75 @@ func (s *CAASProvisionerSuite) TestProvisioningInfo(c *gc.C) {
 		},
 	})
 	c.Assert(err, jc.ErrorIsNil)
-	c.Assert(results, jc.DeepEquals, params.KubernetesProvisioningInfoResults{
-		Results: []params.KubernetesProvisioningInfoResult{{
-			Result: &params.KubernetesProvisioningInfo{
-				PodSpec: "spec(gitlab)",
-				DeploymentInfo: &params.KubernetesDeploymentInfo{
-					DeploymentType: "stateful",
-					ServiceType:    "loadbalancer",
-				},
-				Filesystems: []params.KubernetesFilesystemParams{{
-					StorageName: "data",
-					Provider:    string(provider.K8s_ProviderType),
-					Size:        100,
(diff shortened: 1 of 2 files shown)


anastasiamac:
[CAASProvisionerSuite.TestProvisioningInfo](https://trello.com/c/vgJkc9gc/89-caasprovisionersuitetestprovisioninginfo)

apiserver/facades/controller/caasunitprovisioner/provisioner_test.go
```diff
@@ -190,62 +190,75 @@ func (s *CAASProvisionerSuite) TestProvisioningInfo(c *gc.C) {
...
+	obtained := results.Results[0].Result
+	c.Assert(obtained.PodSpec, jc.DeepEquals, expectedResult.PodSpec)
+	c.Assert(obtained.DeploymentInfo, jc.DeepEquals, expectedResult.DeploymentInfo)
+	c.Assert(len(obtained.Filesystems), gc.Equals, len(expectedFileSystems))
+	for _, fs := range obtained.Filesystems {
+		c.Assert(fs, gc.DeepEquals, expectedFileSystems[fs.StorageName])
+	}
+	c.Assert(obtained.Devices, jc.DeepEquals, expectedResult.Devices)
+	c.Assert(obtained.Constraints, jc.DeepEquals, expectedResult.Constraints)
+	for tagKey := range obtained.Tags {
```
babbageclunk: You can just do a DeepEquals on the two maps here - it does an order-insensitive comparison in that case. Otherwise you'd need a length check as well to make sure that there aren't any keys missing in obtained.Tags.

anastasiamac:
Actually looking at the production code, i think i want to ensure that storage order is always the same and order it at https://github.com/juju/juju/blob/develop/apiserver/facades/controller/caasunitprovisioner/provisioner.go#L366

The test has to change anyway since there is a map of Tags...

anastasiamac:
!!build!!

anastasiamac:
$$merge$$

PR #10128, opened 2019-05-02
Longer wait for workerSuite.TestRemovePrimaryValidSecondaries.

anastasiamac:
## Description of change

Prior to this PR, this test failed frequently. In fact, locally on my machine, it has failed only after 7 runs under stress.
By increasing the wait from Short to Long, I have not seen a failure locally after 147 runs.

Drive-by: naming collision in test between variable name and an import.


anastasiamac:
[Under stress workerSuite.TestRemovePrimaryValidSecondaries fails](https://trello.com/c/ty9bM2bZ/90-under-stress-workersuitetestremoveprimaryvalidsecondaries-fails)

anastasiamac:
$$merge$$

PR #10139, opened 2019-05-06
Fix machine cleanup with manual machines

wallyworld:
## Description of change

Models with manual machines were not being destroyed properly - the loop which iterated over machines to clean up incorrectly skipped any non manual machines once the first manual machine was encountered.

## QA steps

create a model and add a manual machine
deploy an application (creates a non-manual machine)
destroy-model
should complete (previously would have got stuck on machines)

## Bug reference

https://bugs.launchpad.net/juju/+bug/1827285


wallyworld:
[manual machine cleanup](https://trello.com/c/696y6acC/733-manual-machine-cleanup)

wallyworld:
$$merge$$

PR #10147, opened 2019-05-07
Add make test target

wallyworld:
## Description of change

Add a new `make test ` target which is like `make check` but does not run the linters.


wallyworld:
retarget to 2.5

PR #10156, opened 2019-05-08
Increase default teardown timeout

timClicks:
## Description of change

A small number of our CI jobs are failing because the machine or controller teardown takes longer than a pre-defined 10 minutes. This commit increases the default limit to 15 minutes. It doesn't affect the global timeout for the job that is set by Jenkins Job Builder.

## QA steps

n/a

## Documentation changes

n/a

## Bug reference

n/a

### DIFF
--- a/acceptancetests/jujupy/client.py
+++ b/acceptancetests/jujupy/client.py
@@ -116,7 +116,7 @@ def get_teardown_timeout(client):
     elif client.env.provider == 'gce':
         return 1200
     else:
-        return 600
+        return 900
 
 
 def parse_new_state_server_from_error(error):


timClicks:
> I'd rather not keep extending the timeout for destroying controllers or anything in CI (I've done it for webscale, but that is an oddity), as we end up masking a problem. Instead we should either workout why it's taking longer and then see if there is a fix for it or admit it does take longer and think of another strategy in CI (unknown atm!).

The problem is that this timeout is causing tests that pass to be marked as failing. The tests run and complete, and then this internal timeout causes a non-zero exit code to be generated. Why does a CI test care about the time that it takes to destroy the machine(s)?

I would prefer that we just blow away the cloud/LXD instance rather than bothering about `--destroy-controller`, but that's a much bigger change.

Let's start a thread on discourse that is a brainstorm of different strategies.

timClicks:
$$merge$$

mitechie:
> The problem is that this timeout is causing tests that pass to be marked as failing. The tests run and complete, and then this internal timeout causes a non-zero exit code to be generated. Why does a CI test care about the time that it takes to destroy the machine(s)?

To clarify, it's not that extending where needed is bad. It's that this, being a default value, extends everything by default. I don't think anyone has issue for extending timeouts where there's a demonstrated need to do so. Having a larger default means that it also causes resources to be tied up for longer in situations that are in fact an issue. 


> Let's start a thread on discourse that is a brainstorm of different strategies.

Thanks for offering to start the conversation and I'm sure we can look into changes that help improve things together.

SimonRichardson:
My thoughts are here https://discourse.jujucharms.com/t/the-tear-down-problem-in-acceptance-tests/1467

PR #10158, opened 2019-05-09
Support more acceptance tests from ./assess helper

timClicks:
## Description of change

Juju's acceptance tests use several variants in their command-line arguments. This change teaches the `acceptancetests/assess` helper script about almost all of them.  `assess_heterogenous_control.py` is particularly idiosyncratic and has been deferred to a future commit.

## QA steps

n/a (developer-facing only)

## Documentation changes

n/a (developer-facing only)

## Bug reference

None.

### DIFF
--- a/acceptancetests/assess
+++ b/acceptancetests/assess
@@ -65,7 +65,9 @@ def list_tests():
     test_files = glob.glob(test_pattern)
     test_files = (f.split('/')[-1] for f in test_files)
     test_files = (f[7:-3] for f in test_files)
-    return sorted(test_files)
+    test_files = sorted(test_files)
+    test_files.remove("heterogeneous_control")  # TODO(tsm) fixme
+    return test_files
 
 
 def default_juju_bin():
@@ -154,6 +156,8 @@ def parse_args():
     pass_through.add_argument('--region', default=None, help='Override environment region.')
     pass_through.add_argument('--to', default=None,
                               help='Place the controller at a location.')
+    pass_through.add_argument("--agent", help="Which agent to test (Only relevant for log_rotation test)",
+                              choices=["machine", "unit"])
     pass_through.add_argument('--agent-url', action='store', default=None,
                               help='URL for retrieving agent binaries.')
     pass_through.add_argument('--agent-stream', action='store', default=None,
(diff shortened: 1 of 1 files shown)


acceptancetests/assess
```diff
@@ -162,14 +166,114 @@ def parse_args():
...
+                              help="Path to directory with pre-loaded metadata. (Revelant only for the bootstrap test)")
+    pass_through.add_argument("--remote-metadata-source",
+                              help="URL to image metadata (Relevant only for the mixed_images test)")
+    pass_through.add_argument("--export-to",
+                              help="Part to where an exported bundle will be written to. (Relevant only for the bundle_export test)")
+    pass_through.add_argument("--caas-provider",
+                              help="Provider for CaaS models. (Relevant only for caas_deploy_charms test) [default: microk8s]",
+                              choices=["microk8s", "k8s"])
+    pass_through.add_argument('--caas-image-repo',
+                              help="Docker image repo to use. (Relevant only for caas_deploy_charms test) [default: jujuqabot]",
```
wallyworld: This is relevant for any caas test. As is --caas-provider etc. We only have one test now but are adding more. Can we reword to be more generic.
Not sure if there's other examples also that are currently a little too specific.

acceptancetests/assess
```diff
@@ -263,14 +381,151 @@ def main():
...
+        testrun_argv += ["--to", args.to]
     if args.agent_url:
-        testrun_argv.extend(["--agent-url", args.agent_url])
+        testrun_argv += ["--agent-url", args.agent_url]
     if args.agent_stream:
-        testrun_argv.extend(["--agent-stream", args.agent_stream])
-
+        testrun_argv += ["--agent-stream", args.agent_stream]
+
+    assess = args.assess
```
wallyworld: Can we add a big comment that the below test specific behaviour is only for tests which use bespoke args for that test; tests which just use generic args are not listed etc etc

acceptancetests/assess
timClicks: Have created a new section.

acceptancetests/assess
timClicks: Sure, done.

timClicks:
$$merge$$

PR #10159, opened 2019-05-09
For 2.6.1: Force destroy life asserts should take current life.

anastasiamac:
## Description of change

When we are forcing a removal or a destruction of the entity, we should assert on it's current life.

### DIFF
--- a/state/machine.go
+++ b/state/machine.go
@@ -601,11 +601,12 @@ func (m *Machine) forceDestroyOps(maxWait time.Duration) ([]txn.Op, error) {
 		// We set the machine to Dying if it isn't already dead.
 		var machineOp txn.Op
 		if m.Life() < Dead {
-			// Make sure we don't want the vote, and we are queued to be Dying
+			// Make sure we don't want the vote, and we are queued to be Dying.
+			// Since we are force deleting, life assert should be current machine's life.
 			machineOp = txn.Op{
 				C:      machinesC,
 				Id:     m.doc.DocID,
-				Assert: bson.D{{"life", bson.D{{"$in", []Life{Alive, Dying}}}}},
+				Assert: bson.D{{"life", m.doc.Life}},
 				Update: bson.D{{"$set", bson.D{{"novote", true}, {"life", Dying}}}},
 			}
 		} else {
--- a/state/machine.go
+++ b/state/machine.go
@@ -606,7 +606,7 @@ func (m *Machine) forceDestroyOps(maxWait time.Duration) ([]txn.Op, error) {
 			machineOp = txn.Op{
 				C:      machinesC,
 				Id:     m.doc.DocID,
-				Assert: bson.D{{"life", m.doc.Life}},
+				Assert: bson.D{{"life", bson.D{{"$in", []Life{Alive, Dying}}}}},
 				Update: bson.D{{"$set", bson.D{{"novote", true}, {"life", Dying}}}},
 			}
 		} else {
--- a/state/relation.go
+++ b/state/relation.go
@@ -427,10 +427,16 @@ func (r *Relation) destroyOps(ignoreApplication string, op *ForcedOperation) (op
 		}
 		return removeOps, true, nil
 	}
+	lifeAssert := isAliveDoc
+	if op.Force {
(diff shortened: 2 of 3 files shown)


anastasiamac:
[check destroy cleanup asserts](https://trello.com/c/Jpf1UgTh/712-check-destroy-cleanup-asserts)

state/machine.go
```diff
@@ -601,11 +601,12 @@ func (m *Machine) forceDestroyOps(maxWait time.Duration) ([]txn.Op, error) {
...
 		var machineOp txn.Op
 		if m.Life() < Dead {
-			// Make sure we don't want the vote, and we are queued to be Dying
+			// Make sure we don't want the vote, and we are queued to be Dying.
+			// Since we are force deleting, life assert should be current machine's life.
 			machineOp = txn.Op{
 				C:      machinesC,
 				Id:     m.doc.DocID,
-				Assert: bson.D{{"life", bson.D{{"$in", []Life{Alive, Dying}}}}},
+				Assert: bson.D{{"life", m.doc.Life}},
```
wallyworld: The original assertion is actually *more* permissive than this one. I'd be hesitant to tighten it up unless we're 1000% sure that it's never an expected outcome that the life could go from alive to dying as this transaction is being composed.

state/relation.go
```diff
@@ -427,10 +427,16 @@ func (r *Relation) destroyOps(ignoreApplication string, op *ForcedOperation) (op
...
+	if op.Force {
+		// Since we are force destroying, life assert should be current relation's life.
+		lifeAssert = bson.D{{"life", r.doc.Life}}
+	}
+
 	return []txn.Op{{
 		C:      relationsC,
 		Id:     r.doc.DocID,
-		Assert: bson.D{{"life", Alive}, {"unitcount", bson.D{{"$gt", 0}}}},
+		Assert: append(bson.D{{"unitcount", bson.D{{"$gt", 0}}}}, lifeAssert...),
```
wallyworld: append(lifeAssert, {"unitcount", bson.D{{"$gt", 0}}})

state/storage.go
```diff
@@ -466,10 +471,7 @@ func (sb *storageBackend) destroyStorageInstanceOps(
 	// There are still attachments: the storage instance will be removed
 	// when the last attachment is removed. We schedule a cleanup to destroy
 	// attachments.
-	notLastRefs := bson.D{
-		{"life", Alive},
-		{"attachmentcount", bson.D{{"$gt", 0}}},
-	}
+	notLastRefs := append(bson.D{{"attachmentcount", bson.D{{"$gt", 0}}}}, lifeAssert...)
```
wallyworld: ditto

state/machine.go
anastasiamac: I am not sure why you think it is more permissive? Previously it was 'either alive or dying', now it is 'whatever is current live' :)

state/relation.go
anastasiamac: I am pretty sure either way will work and will achieve the same result but I'll change :)

anastasiamac: Actually the reverse does not work as you'd get an error similar to "cannot use bson.D{...} as type DocElement"

The correct syntax, if you wanted to change the order would have been "append(lifeAssert, bson.D{{"unitcount", bson.D{{"$gt", 0}}}}...)" which does not give you much of a difference from what I have written.

The only way that it could work is if lifeAssert was declared as bson.DocElement but it would not work because our isAliveDoc is bson.D already.

So no change is needed here.

state/machine.go
wallyworld: Before, the current life could be "alive", but if the life were to change to "dying" by the time the txn was run, it would still satisfy the assert. With the change in this PR, that is no longer the case.

anastasiamac: I see.
Well, the life is kind of irrelevant during force... Maybe it should not be part of the assert at all...
Meanwhile I've reverted this change.

anastasiamac:
unrelated failure in github.com/juju/juju/worker/firewaller.TestAll

anastasiamac:
!!build!!

anastasiamac:
$$merge$$

PR #10161, opened 2019-05-09
2.5.5 release

SimonRichardson:
## Description of change

We need a 2.5.5 build for the 2.5.5 release. Because we burned 2.5.5
and then didn't burn it, we moved the version back to 2.5.4, which
meant that aws s3 buckets didn't have the correct version number in
the path of the files. It had 2.5.4 instead of 2.5.5.

The following commit that bumped it to 2.5.6 then meant we never had
a 2.5.5 build in s3, so we could never do a build from it. This
fixes this, but after this we'll need to bump it back to 2.5.6!


SimonRichardson:
$$merge$$

jameinel:
$$merge$$

PR #10165, opened 2019-05-10
Allow lxd profile test to run on aws

SimonRichardson:
## Description of change

Some of the lxd profile tests will fail because of the assertions
with in the test on what profiles should be where. This PR fixes
this, so that we don't care what profiles land where.

It's a better fix because we use a object for instance removal,
rather than a counter, so we can visually inspect the object to
see what's left to be applied rather than just knowing something
has to be applied.

## QA steps

Setup the acceptance tests correctly for aws, then run the 
following:

```
./assess_deploy_lxd_profile_bundle.py myaws --charm-bundle=cs:~juju-qa/bundle/lxd-profile-bundle-sub-1
```

### DIFF
--- a/acceptancetests/assess_deploy_lxd_profile_bundle.py
+++ b/acceptancetests/assess_deploy_lxd_profile_bundle.py
@@ -71,9 +71,23 @@ def assess_profile_machines(client):
                 else:
                     machines = application_machines_from_app_info(info)
                 machine_profiles.append((charm_profile, machines))
-    if len(machine_profiles) > 0:
-        aligned_machine_profiles = align_machine_profiles(machine_profiles)
+    
+    # if the provider type is not lxd, then we should drop all machine profiles
+    # as profiles will not be applied 
+    if client.env.provider_type is not 'lxd':
+        machine_profiles = select_containers(machine_profiles)
+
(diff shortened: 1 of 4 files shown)


SimonRichardson:
https://trello.com/c/qcIq9jOY/1372-ci-lxd-subordinate-to-aws

acceptancetests/assess_deploy_lxd_profile_bundle.py
```diff
@@ -71,10 +71,24 @@ def assess_profile_machines(client):
...
+        machine_profiles = select_containers(machine_profiles)
+
+    aligned_machine_profiles = align_machine_profiles(machine_profiles)
+    if len(aligned_machine_profiles) > 0:
         client.wait_for(WaitForLXDProfilesConditions(aligned_machine_profiles))
 
+def select_containers(profiles):
+    result = []
+    for tup in profiles:
+        m = [ machine for machine in tup[1] if machine.find("/") >= 0 ]
```
hmlanigan: Won't this catch kvm containers too?  Suggest "/lxd/" instead.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #10171, opened 2019-05-13
Fix intermittent test failure in caas provisioner

SimonRichardson:
## Description of change

The following ensures the order of the filesystem in the results,
ordering by the storage name

## QA steps

Run the following without my branch, then do the same with my branch

```
go test -v ./apiserver/facades/controller/caasunitprovisioner/... -check.v -check.f=CAASProvisionerSuite.TestProvisioningInfo -count=40
```


SimonRichardson:
$$merge$$

PR #10178, opened 2019-05-14
Add --force and --no-wait to juju remove-saas

timClicks:
## Description of change

This commit adds `--force` and `--no-wait` parameters to the `remove-saas` CLI command. The first prevents the removal process from aborting if there is an error detected during an intermediate step. The second argument instructs Juju to proceed as quickly as possible. By default, Juju does one thing at a time and waits before proceeding to the next step.

## QA steps

- Create a remote/consumed application in accordance with the CMR steps
- Use `juju remove-saas --force`

## Documentation changes

New parameters have been added to the `remove-saas` command.

Error messages are reported when `--no-wait` is provided without `--force` and when either `--force` or `--no-wait` is provided for controllers that do not support those options.

## Bug reference

* [lp:1822050](https://bugs.launchpad.net/juju/+bug/1822050)
* [lp:1813886](https://bugs.launchpad.net/juju/+bug/1813886)

### DIFF
--- a/api/application/client.go
+++ b/api/application/client.go
@@ -604,14 +604,44 @@ func (c *Client) DestroyApplications(in DestroyApplicationsParams) ([]params.Des
 	return allResults, nil
 }
 
+type DestroyConsumedApplicationParams struct {
+	// SaasNames holds the names of the consumed applications
+	// that are being destroyed
+	SaasNames []string
+
+	// Force controls whether or not the removal of applications
+	// will be forced, i.e. ignore removal errors.
+	Force bool
+
+	// MaxWait specifies the amount of time that each step in application removal
+	// will wait before forcing the next step to kick-off. This parameter
+	// only makes sense in combination with 'force' set to 'true'.
+	MaxWait *time.Duration
+}
+
 // DestroyConsumedApplication destroys the given consumed (remote) applications.
-func (c *Client) DestroyConsumedApplication(saasNames ...string) ([]params.ErrorResult, error) {
+func (c *Client) DestroyConsumedApplication(in DestroyConsumedApplicationParams) ([]params.ErrorResult, error) {
+	apiVersion := c.BestAPIVersion()
 	args := params.DestroyConsumedApplicationsParams{
-		Applications: make([]params.DestroyConsumedApplicationParams, 0, len(saasNames)),
+		Applications: make([]params.DestroyConsumedApplicationParams, 0, len(in.SaasNames)),
 	}
-	allResults := make([]params.ErrorResult, len(saasNames))
-	index := make([]int, 0, len(saasNames))
-	for i, name := range saasNames {
+
+	if apiVersion > 9 {
+		if in.MaxWait != nil && !in.Force {
+			return nil, errors.New("--force is required when --max-wait is provided")
+		}
+	} else {
+		if in.Force {
+			return nil, errors.New("this controller does not support --force")
+		}
+		if in.MaxWait != nil {
+			return nil, errors.New("this controller does not support --no-wait")
+		}
+	}
+
+	allResults := make([]params.ErrorResult, len(in.SaasNames))
+	index := make([]int, 0, len(in.SaasNames))
+	for i, name := range in.SaasNames {
 		if !names.IsValidApplication(name) {
 			allResults[i].Error = &params.Error{
 				Message: errors.NotValidf("SAAS application name %q", name).Error(),
@@ -619,9 +649,14 @@ func (c *Client) DestroyConsumedApplication(saasNames ...string) ([]params.Error
 			continue
 		}
 		index = append(index, i)
-		args.Applications = append(args.Applications, params.DestroyConsumedApplicationParams{
+		appParams := params.DestroyConsumedApplicationParams{
 			ApplicationTag: names.NewApplicationTag(name).String(),
-		})
+		}
+		if apiVersion > 9 {
+			appParams.Force = &in.Force
+			appParams.MaxWait = in.MaxWait
+		}
+		args.Applications = append(args.Applications, appParams)
 	}
 
 	var result params.ErrorResults
--- a/api/application/client_test.go
+++ b/api/application/client_test.go
@@ -444,7 +444,10 @@ func (s *applicationSuite) TestDestroyConsumedApplications(c *gc.C) {
 		*out = params.ErrorResults{expectedResults}
 		return nil
 	})
-	results, err := client.DestroyConsumedApplication("foo", "bar")
+	destroyParams := application.DestroyConsumedApplicationParams{
+		[]string{"foo", "bar"}, false, nil,
+	}
+	results, err := client.DestroyConsumedApplication(destroyParams)
 	c.Assert(err, jc.ErrorIsNil)
 	c.Assert(results, jc.DeepEquals, expectedResults)
 }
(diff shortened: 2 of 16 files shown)


apiserver/facades/client/application/application.go
```diff
@@ -1565,10 +1565,14 @@ func (api *APIBase) DestroyConsumedApplications(args params.DestroyConsumedAppli
...
-		// see https://bugs.launchpad.net/juju/+bug/1822050
-		err = app.Destroy()
+
+		op := app.DestroyOperation(arg.Force)
+		if arg.Force {
+			op.MaxWait = common.MaxWait(arg.MaxWait)
+		}
+		err = api.backend.ApplyOperation(op)
 		if err != nil {
+			logger.Warningf("operational error encountered destroying application %v: %v", appTag.Id(), op.Errors)
```
anastasiamac: This should b printed in a separate if-, if there are any :)

apiserver/params/applications.go
```diff
@@ -314,6 +314,15 @@ type DestroyConsumedApplicationsParams struct {
 // RemoteApplication.Destroy call.
 type DestroyConsumedApplicationParams struct {
 	ApplicationTag string `json:"application-tag"`
+
+	// Force controls whether or not the destruction process ignores
+	// operational errors. When true, the process will ignore them.
+	Force bool `json:"force"`
```
anastasiamac: I think this should be *bool with omitempty in json tag... to cater for backward compatibility.

state/remoteapplication.go
```diff
@@ -249,6 +249,11 @@ func copyAttributes(values attributeMap) attributeMap {
 	return result
 }
 
+// DestroyOperation returns a model operation to destroy this remote application.
+func (s *RemoteApplication) DestroyOperation(force bool) *DestroyRemoteApplicationOperation {
```
anastasiamac: I do not think you need this one... Just use the one below.

(or rename the existing one)

timClicks: I like the idea of renaming, as RemoteApplication is implied as it's a method.

api/application/client.go
```diff
@@ -621,6 +621,8 @@ func (c *Client) DestroyConsumedApplication(saasNames ...string) ([]params.Error
 		index = append(index, i)
 		args.Applications = append(args.Applications, params.DestroyConsumedApplicationParams{
 			ApplicationTag: names.NewApplicationTag(name).String(),
+			Force: force,
```
anastasiamac: This will need to be facade version aware since older facades would not support force, etc..

cmd/juju/application/removesaas.go
```diff
@@ -71,11 +77,18 @@ func (c *removeSaasCommand) Init(args []string) error {
 	return nil
 }
 
+func (c *removeSaasCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.ModelCommandBase.SetFlags(f)
+	f.BoolVar(&c.Force, "force", false, "Completely remove an application and all its dependencies")
```
anastasiamac: Are we talking about 'applications' in remove-saas?.. I wonder if it makes sense here or if there s another term that is more natural?

timClicks: I think that the term "applications" is most natural. It's included in `juju help remove-saas`:

> Summary:
> Remove consumed applications (SAAS) from the model.

> Details:
> Removing a consumed (SAAS) application will terminate any relations that
> application has, potentially leaving any related local applications
> in a non-functional state.

apiserver/params/applications.go
timClicks: Done

apiserver/facades/client/application/application.go
timClicks: Done. Could you check that this is what you intended?

apiserver/params/applications.go
timClicks: Done

state/remoteapplication.go
timClicks: Renamed.

apiserver/facades/client/application/application.go
anastasiamac: Yes, thank you! But could we please add to the message "consumed application" as in "destroying consumed application" otherwise it'll be indistinguishable from  normal application destroy.

api/application/client.go
```diff
@@ -604,14 +604,41 @@ func (c *Client) DestroyApplications(in DestroyApplicationsParams) ([]params.Des
 	return allResults, nil
 }
 
+type DestroyConsumedApplicationParams struct {
```
anastasiamac: Nice!

api/application/client.go
```diff
@@ -621,6 +648,8 @@ func (c *Client) DestroyConsumedApplication(saasNames ...string) ([]params.Error
 		index = append(index, i)
 		args.Applications = append(args.Applications, params.DestroyConsumedApplicationParams{
 			ApplicationTag: names.NewApplicationTag(name).String(),
+			Force: in.Force,
```
anastasiamac: Please only add Force and MaxWait if the BestAPIVersion is > 9. Previous versions do not need them.

apiserver/params/applications.go
```diff
@@ -314,6 +314,15 @@ type DestroyConsumedApplicationsParams struct {
 // RemoteApplication.Destroy call.
 type DestroyConsumedApplicationParams struct {
 	ApplicationTag string `json:"application-tag"`
+
+	// Force controls whether or not the destruction process ignores
+	// operational errors. When true, the process will ignore them.
+	Force bool `json:"force,omitempty"`
```
anastasiamac: Let's make this a *bool, please.

cmd/juju/application/removesaas.go
```diff
@@ -6,7 +6,9 @@ package application
 import (
 	"github.com/juju/cmd"
 	"github.com/juju/errors"
+	"github.com/juju/gnuflag"
 	"gopkg.in/juju/names.v2"
+	"time"
```
anastasiamac: move me

api/application/client.go
timClicks: Will do. I left this in because they would default to 0-values, so didn't think it was important.

api/application/client.go
```diff
@@ -604,24 +604,57 @@ func (c *Client) DestroyApplications(in DestroyApplicationsParams) ([]params.Des
...
 			}
 			continue
 		}
 		index = append(index, i)
-		args.Applications = append(args.Applications, params.DestroyConsumedApplicationParams{
+		appParams := params.DestroyConsumedApplicationParams{
 			ApplicationTag: names.NewApplicationTag(name).String(),
-		})
+		}
+		if apiVersion >= 10 {
```
anastasiamac: This would work beautifully, just as '> 9' could have.

anastasiamac:
[add force to deletion of consumed application](https://trello.com/c/I8iQtlXw/683-add-force-to-deletion-of-consumed-application)

api/application/client.go
timClicks: I used `>=` because I've included `10` above, and wanted it to be easiest for people to looking at the code to see what's happening.

api/application/client_test.go
```diff
@@ -453,10 +456,84 @@ func (s *applicationSuite) TestDestroyConsumedApplicationsArity(c *gc.C) {
...
 	c.Assert(err, gc.ErrorMatches, `expected 1 result\(s\), got 0`)
 }
 
+func (s *applicationSuite) TestDestroyConsumedApplicationsForcev9(c *gc.C) {
+	var called bool
+	client := application.NewClient(basetesting.BestVersionCaller{
+		APICallerFunc: basetesting.APICallerFunc(
+			func(objType string, version int, id, request string, a, response interface{}) error {
+				called = true
+				return nil
```
anastasiamac: Yes, this is great and this is where we'd want to check that force and maxwait were not passed in.

anastasiamac: And to add the test that if the force and maxwait is passed in to an older version, we get the message that they are not supported.

timClicks: Sounds good, thanks for taking another look :)

api/application/client_test.go
```diff
@@ -453,10 +456,84 @@ func (s *applicationSuite) TestDestroyConsumedApplicationsArity(c *gc.C) {
...
+			},
+		),
+		BestVersion: 9, // v9 does not support --force or --no-wait
+	})
+	destroyParams := application.DestroyConsumedApplicationParams{
+		[]string{"foo"}, true, nil,
+	}
+	results, err := client.DestroyConsumedApplication(destroyParams)
+	c.Check(err, gc.ErrorMatches, "this controller does not support --force")
```
timClicks: @anastasiamac is this the check that you're looking for?

api/application/client.go
```diff
@@ -629,7 +629,7 @@ func (c *Client) DestroyConsumedApplication(in DestroyConsumedApplicationParams)
 
 	if apiVersion > 9 {
 		if in.MaxWait != nil && !in.Force {
-			return nil, errors.New("--force is required when --max-wait is provided")
+			return nil, errors.New("--force is required when --no-wait is provided")
```
anastasiamac: I think this should be max-wait as no-wait is a cli layer term.

timClicks: Makes sense, I'll revert this change

timClicks:
$$merge$$

timClicks:
$$merge$$

timClicks:
[Transient error](http://ci.jujucharms.com/job/unit-tests-win2012/3438/console) relating to networking on the Windows test machine

$$merge$$

PR #10181, opened 2019-05-14
Capture TPS metric when running webscale tests

achilleasa:
The number of transactions is not constant between runs of the webscale tests. For example, 
when server-side txns are enabled we can perform more status update transactions compared 
to running with server-side txns off. 

For this reason, the transaction count is not a meaningful metric to compare txn-related metrics across runs. Since influxdb does not allow us to query across measurements, this PR updates the webscale metrics python file to also export a transactions per second metric for our dashboards.


achilleasa:
$$merge$$

PR #10187, opened 2019-05-14
Intermittent test failure for error case

SimonRichardson:
## Description of change

The following fixes an error case where the test would call kill
on the worker before the actual work calls kill itself. It's a race
to see whom calls kill. The substantial changes ensure that we
wait until the worker mutater context has been killed with an error
first, then we can move on.

We have had to expose some machinery, but it's not as bad as first
thought.

## QA steps

Run the following before and after the PR changes. Note you might need to run it
a couple of times.

```
go test -v ./worker/instancemutater/... -check.v -check.f=TestCharmProfilingInfoError -count=500
```


SimonRichardson:
https://trello.com/c/hAq3Nm7E/1394-intermittent-worker-failure-instance-mutater

SimonRichardson:
$$merge$$

PR #10189, opened 2019-05-14
2.6 nil fixes

jameinel:
## Description of change

This cherry picks the proposed nil pointer fixes that had landed in the 2.6 branch before being reverted, and adds one for the typed nil LXDProfile interface bug.

## QA steps

For this bug the fix is to
```
$ juju-2.4 bootstrap
$ juju-2.4 deploy cs:~jameinel/ubuntu-lite
$ juju-2.6 upgrade-juju 
$ juju status
```

Without this fix, the 2.4 Juju will create a charm document that doesn't have an LXDProfile attribute. Upgrading it to 2.6 will create a Charm object in memory that has a nil pointer for lxdprofile. And the code that thought it was trapping a nil pointer was failing because of Go's [typed-nil issue](https://dave.cheney.net/2017/08/09/typed-nils-in-go-2).

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1829014

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -701,7 +701,11 @@ func (p lxdCharmProfiler) LXDProfile() lxdprofile.LXDProfile {
 		return nil
 	}
 	if profiler, ok := p.Charm.(charm.LXDProfiler); ok {
-		return profiler.LXDProfile()
+		profile := profiler.LXDProfile()
+		if profile == nil {
+			return nil
+		}
+		return profile
 	}
 	return nil
 }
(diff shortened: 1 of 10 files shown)


jameinel:
$$merge$$

api/client.go
```diff
@@ -701,7 +701,11 @@ func (p lxdCharmProfiler) LXDProfile() lxdprofile.LXDProfile {
 		return nil
 	}
 	if profiler, ok := p.Charm.(charm.LXDProfiler); ok {
-		return profiler.LXDProfile()
+		profile := profiler.LXDProfile()
+		if profile == nil {
```
4a6f656c: This dance seems rather fragile - as it is, it is repeated in five places in this PR. It will only take one more use of profiler.LXDProfile to be added, or someone to unintentionally remove this code on the basis that profiler.LXDProfile returns nil and the bug comes back.

Either adding an IsNil (or IsValid) function to the interface, or allowing its functions to be safely callable on nil would be more robust. Another option is to have an isNilProfile function that takes the interface knows about "concrete" types, which can then type switch accordingly.

PR #10193, opened 2019-05-15
Don't uninstall machine agent when a fatal connection error occurs in the machine agent manifold

achilleasa:
## Description of change

There are two codepaths that eventually call `agent.SetCanUninstall`:

- A filter func for the `api-caller` worker in the machine agent manifold invokes it when a `ErrConnectImpossible` error is encountered; the error is logged and replaced with `ErrTerminateAgent`.
- The machiner, after ensuring that a machine is dead.

In both cases, when the machine agent processes the `ErrTerminateAgent` error it attempts to uninstall the agent. This attempt will only succeed if `agent.SetCanUninstall` is invoked first (it writes out an uninstall file that the agent tries to stat before continuing).

This PR patches the code for the **first** case so an uninstall is never triggered. NOTE: Dealing with the latter case will be tackled via a separate PR as it has implications for users working with the manual provider.

### DIFF
--- a/cmd/jujud/agent/machine/manifolds.go
+++ b/cmd/jujud/agent/machine/manifolds.go
@@ -262,16 +262,11 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 	// connectFilter exists:
 	//  1) to let us retry api connections immediately on password change,
 	//     rather than causing the dependency engine to wait for a while;
-	//  2) to ensure that certain connection failures correctly trigger
-	//     complete agent removal. (It's not safe to let any agent other
-	//     than the machine mess around with SetCanUninstall).
+	//  2) to decide how to deal with fatal, non-recoverable errors
+	//     e.g apicaller.ErrConnectImpossible.
 	connectFilter := func(err error) error {
 		cause := errors.Cause(err)
 		if cause == apicaller.ErrConnectImpossible {
-			err2 := coreagent.SetCanUninstall(config.Agent)
-			if err2 != nil {
-				return errors.Trace(err2)
-			}
 			return jworker.ErrTerminateAgent
 		} else if cause == apicaller.ErrChangedPassword {
 			return dependency.ErrBounce
--- a/cmd/jujud/agent/machine/manifolds_test.go
+++ b/cmd/jujud/agent/machine/manifolds_test.go
@@ -4,8 +4,6 @@
 package machine_test
 
 import (
-	"io/ioutil"
-	"os"
 	"sort"
 
 	"github.com/juju/collections/set"
@@ -228,13 +226,9 @@ func (*ManifoldsSuite) TestSingularGuardsUsed(c *gc.C) {
 }
 
 func (*ManifoldsSuite) TestAPICallerNonRecoverableErrorHandling(c *gc.C) {
-	tmpDir, err := ioutil.TempDir("", "agent-data-dir")
-	c.Assert(err, jc.ErrorIsNil)
-	defer func() { _ = os.RemoveAll(tmpDir) }()
-
 	ag := &mockAgent{
 		conf: mockConfig{
-			dataPath: tmpDir,
+			dataPath: c.MkDir(),
 		},
 	}
 	manifolds := machine.Manifolds(machine.ManifoldsConfig{
@@ -247,7 +241,7 @@ func (*ManifoldsSuite) TestAPICallerNonRecoverableErrorHandling(c *gc.C) {
 	// Check that when the api-caller maps non-recoverable errors to
 	// ErrTerminateAgent and that it does not create an uninstall file for
 	// the agent.
-	err = apiCaller.Filter(apicaller.ErrConnectImpossible)
+	err := apiCaller.Filter(apicaller.ErrConnectImpossible)
 	c.Assert(err, gc.Equals, jworker.ErrTerminateAgent)
 	c.Assert(agent.CanUninstall(ag), gc.Equals, false)
 }
--- a/cmd/jujud/agent/machine/manifolds_test.go
+++ b/cmd/jujud/agent/machine/manifolds_test.go
@@ -4,6 +4,8 @@
 package machine_test
 
 import (
+	"io/ioutil"
+	"os"
 	"sort"
 
 	"github.com/juju/collections/set"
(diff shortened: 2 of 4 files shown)


cmd/jujud/agent/machine/manifolds_test.go
```diff
@@ -223,6 +227,31 @@ func (*ManifoldsSuite) TestSingularGuardsUsed(c *gc.C) {
 	}
 }
 
+func (*ManifoldsSuite) TestAPICallerNonRecoverableErrorHandling(c *gc.C) {
+	tmpDir, err := ioutil.TempDir("", "agent-data-dir")
```
jameinel: This should be c.Mkdir() as that will be cleaned up by the test infrastructure.

cmd/jujud/agent/machine/manifolds.go
```diff
@@ -262,16 +262,11 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
...
-	//     than the machine mess around with SetCanUninstall).
+	//  2) to decide how to deal with fatal, non-recoverable errors
+	//     e.g apicaller.ErrConnectImpossible.
 	connectFilter := func(err error) error {
 		cause := errors.Cause(err)
 		if cause == apicaller.ErrConnectImpossible {
-			err2 := coreagent.SetCanUninstall(config.Agent)
-			if err2 != nil {
-				return errors.Trace(err2)
-			}
```
jameinel: One discussion we had is potentially having us remove the Systemd file (so they agent doesn't continually start, fail to login, then die and get restarted by systemd), but *not* remove all of the /var/lib/juju/* content. (Thus in the event of triggering this accidentally, you can still recover)
That said, this is a more conservative option (just don't uninstall anything), which might be a good first step.

cmd/jujud/agent/machine/manifolds_test.go
achilleasa: Pushed https://github.com/juju/juju/pull/10193/commits/96777d5d16ae76ec1c4279a71cbb0bea2625aceb to switch to c.MkDir()

achilleasa:
!!build!!

Clock panic: `github.com/juju/juju/vendor/github.com/juju/clock/testclock/clock.go:144: got 1 timers added after waiting 50ms: wanted 2`

achilleasa:
!!build!!

achilleasa:
After considering the comments on the bug page by @jameinel and digging a bit further with @SimonRichardson it seems like in this particular case it was the machiner and not the filter func that triggered the agent removal.

Since the machiner will invoke `EnsureDead` before invoking the `NotifyMachineDead` callback we can probably assume that the call to `EnsureDead` returned without an error. 

Taking a look at the [advanceLifecycle](https://github.com/juju/juju/blob/2.5/state/machine.go#L740) implementation in state/machine.go there are a few places where we may get `ErrNoOperations` out of the txn builder code: 

- if the machine is **not** found
- if it's already transitioned to a `Dead` state

In any case we need to dig further into the logs to figure out what actually happened here. The proposed changes will mask the problem but I am not sure if they will cause any issues when using 
the **manual** provider.

worker/machiner/manifold.go
```diff
@@ -91,7 +91,13 @@ func newWorker(a agent.Agent, apiCaller base.APICaller) (worker.Worker, error) {
 		Tag:                          tag.(names.MachineTag),
 		ClearMachineAddressesOnStart: ignoreMachineAddresses,
 		NotifyMachineDead: func() error {
-			return agent.SetCanUninstall(a)
+			// achilleasa: uninstalling the agent is probably not
```
jameinel: We've decided not to include this one as it has "normal use case" implications. So lets back out this change and then land your patch into 2.5 and beyond.

And then we'll do a bigger spec around "if we never want to uninstall ourselves how do we recover a manual machine", and when we have the answer worked out, we can land this part of the change.

Also, fwiw, we should just set NotifyMachineDead: nil, not a func() that does nothing.

achilleasa: Totally agree about setting it to nil. I just left it there for the commented line to be visible.

I will drop this commit and proceed as you suggest.

achilleasa:
The removal of the uninstall call from the manifold filter function causes some tests in the machine lease test suite to fail:

- https://github.com/juju/juju/blob/2.5/cmd/jujud/agent/machine_test.go#L225
- https://github.com/juju/juju/blob/2.5/cmd/jujud/agent/machine_test.go#L239
- https://github.com/juju/juju/blob/2.5/cmd/jujud/agent/machine_test.go#L800 (and L805)

All of these tests seem to rely on the fact that when the agent tries to login, it receives back an unauthorized error which gets replaced with ErrConnectImpossible in ScaryConnect. In the previous implementation, this caused the filter func in the worker manifold to flag the agent as ready to be uninstalled and the uninstall process to run before the worker terminates. After my changes only the machiner can trigger the uninstall but that doesn't seem to ever start as the api-caller fails.

@jameinel any ideas how to fix these (or whether having those predicates makes sense after my changes)? We could potentially omit the check for the first two checks but the third one explicitly tests clean-ups.

jameinel:
- https://github.com/juju/juju/blob/2.5/cmd/jujud/agent/machine_test.go#L225
- https://github.com/juju/juju/blob/2.5/cmd/jujud/agent/machine_test.go#L239
- https://github.com/juju/juju/blob/2.5/cmd/jujud/agent/machine_test.go#L800 (and L805)

These are interesting. We are running the agent after we have called "EnsureDead" on the machine object. I'm pretty sure that means it sees ErrConnectImpossible rather than seeing that it is a dying machine/should clean up soon. We should check what code path it is going by.  One thing that Tim and I discussed is possibly having the worker just stop the executable rather than uninstall.

I wonder if we could do the test with the agent running, and *then* call EnsureDead so the agent sees the machine object die rather than start dead.

achilleasa:
@jameinel https://github.com/juju/juju/pull/10193/commits/1fa885a44f05cfa79521cf3fdbedcd559ddad717 is my proposal for fixing the tests. Can I somehow make the agent spin up faster? It takes about 30sec for the test to run :-(

achilleasa:
!!build!!

(windows test run failed)

achilleasa:
!!build!!

achilleasa:
!!build!!

achilleasa:
!!build!!

(rebooted windows boxes)

achilleasa:
$$merge$$

PR #10195, opened 2019-05-16
Change k8s storage config upgrade step to skip non-live models

babbageclunk:
## Description of change

We've had users with dying models that they can't remove (because the k8s has gone away and we're trying to talk to it before removing the model). Those models would prevent the k8s storage config step from succeeding and block the upgrade. Skip dying models in this upgrade step.

Also drive-by fix to handle the error if the cloud config can't be retrieved.

## QA steps
* Bootstrap a 2.5.7 controller and add a k8s model to it.
* Stop the k8s instance and destroy the model.
* Cancel the destroy-model command, it won't finish.
* Upgrade the controller to this version.
* The upgrade should complete without changing the storage configuration of the dying model.

## Documentation changes
None

## Bug reference
None

### DIFF
--- a/state/upgrades.go
+++ b/state/upgrades.go
@@ -1930,6 +1930,11 @@ func updateKubernetesStorageConfig(st *State) error {
 	if err != nil || model.Type() == ModelTypeIAAS {
 		return errors.Trace(err)
 	}
+	if model.Life() != Alive {
+		// No need to update models that are going away; they may no
+		// longer have settings to update.
+		return nil
+	}
 	cred, ok := model.CloudCredential()
 	if !ok {
 		return nil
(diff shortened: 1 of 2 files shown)


state/upgrades.go
```diff
@@ -1930,6 +1930,11 @@ func updateKubernetesStorageConfig(st *State) error {
 	if err != nil || model.Type() == ModelTypeIAAS {
 		return errors.Trace(err)
 	}
+	if model.Life() != Alive {
```
anastasiamac: I think that this is where we want to be overall in the long run... However, there are some IAAS models that are already dying on older Juju versions and can only be removed if they are upgraded to newer Juju versions... I wonder if we should be only putting this here for CAAS models.

state/upgrades.go
```diff
@@ -1940,6 +1945,9 @@ func updateKubernetesStorageConfig(st *State) error {
 	}
 
 	defaults, err := st.controllerInheritedConfig(model.Cloud())()
+	if err != nil {
```
anastasiamac: Good pick up! I am not sure why anyone would want to swallow an error :(

state/upgrades.go
babbageclunk: Oh, this whole step is only for caas models - it exits early for iaas ones.

babbageclunk:
$$merge$$

PR #10197, opened 2019-05-16
Put together a CLI script that can run Mongo transactions.

jameinel:
## Description of change

This just uses the TXN infrastructure to run any Operation that you want
with the normal TXN logic. Initial tests seem to show it able to hit
assertions and properly update documents.

I want to have this in the Juju directory just to keep the helper script in a place where people can find it, and have easy access to it.

## QA steps

You can see the examples in the directory. You should be able to update records in a Juju database by crafting a JSON file that looks a lot like our normal txn.Op structure, and then run
```
$ agent=$(cd /var/lib/juju/agents; echo machine-*)
$ pw=$(sudo awk '$1 ~ "^statepassword:" {print $2}' /var/lib/juju/agents/${agent}/agent.conf)
$ ./mgo-run-txn --user ${agent} --password ${pw} ./transaction.json
```

## Documentation changes

None.

## Bug reference

None.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ vendor/
 .vscode/
 juju-backup-*.tar.gz
 .editorconfig
+scripts/mgo-run-txn/mgo-run-txn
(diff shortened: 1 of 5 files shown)


jameinel:
$$merge$$

PR #10201, opened 2019-05-17
All cloud instances.

anastasiamac:
## Description of change

Each provider has to implement an InstanceBroker interface. One of the major calls of this interface is AllInstances(). The call is made whenever Juju needs to retrieve known cloud instances, be it for provisioning, for adopting resources, for querying instance states, destroying instances, etc.

The original intent was to *always* list all instances. However, it became apparent that some providers had limitations on what a user can do with an instance in a particular state, so the AllInstance method was inconsistently re-defined to mean *all running, usable instances*. This especially affected public clouds like aws. This new definition works for most of cases, except for destroying an instance - Juju does not really care what state the instance is in as long as we knew about and at the time when we want to terminate, we can. In addition, some providers did not need to make this distinction either because their underlying substrate only returned *running* instances or because the substrate dealt with instances in different state itself.

This means that we needed to rename current AllInstances to mean AllRunnningInstances and re-implement new AllInstances to return all instances except the ones that are already terminated.
 This PR does just that.
  
This is a largely mechanical PR. Most of fun is in environ broker where a new AllRunningInstances call is added and in providers that filter on status like ec2 and gce.

### DIFF
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -1153,7 +1153,7 @@ func (e *environ) subnetsForVPC(ctx context.ProviderCallContext) (resp *ec2.Subn
 // AdoptResources is part of the Environ interface.
 func (e *environ) AdoptResources(ctx context.ProviderCallContext, controllerUUID string, fromVersion version.Number) error {
 	// Gather resource ids for instances, volumes and security groups tagged with this model.
-	instances, err := e.AllInstances(ctx)
+	instances, err := e.AllRunningInstances(ctx)
 	if err != nil {
 		return errors.Trace(err)
 	}
@@ -1182,6 +1182,14 @@ func (e *environ) AdoptResources(ctx context.ProviderCallContext, controllerUUID
 
 // AllInstances is part of the environs.InstanceBroker interface.
 func (e *environ) AllInstances(ctx context.ProviderCallContext) ([]instances.Instance, error) {
+	// We want to return everything we find here except for instances that are
+	// "shutting-down" - they are on the way to be terminated - or already "terminated".
+	// From https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html
+	return e.AllInstancesByState(ctx, "rebooting", "pending", "running", "stopping", "stopped")
+}
+
+// AllRunningInstances is part of the environs.InstanceBroker interface.
+func (e *environ) AllRunningInstances(ctx context.ProviderCallContext) ([]instances.Instance, error) {
 	return e.AllInstancesByState(ctx, "pending", "running")
 }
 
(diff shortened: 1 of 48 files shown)


anastasiamac:
[cloud instances statuses when destroying (all instances)](https://trello.com/c/DhsjHvhc/688-cloud-instances-statuses-when-destroying-all-instances)

anastasiamac:
!!build!!

anastasiamac:
ahhh... 09:59:19 go install github.com/juju/juju/acceptancetests/repository/charms/fill-logs/actions: build output "/home/ubuntu/go/bin/actions" already exists and is not an object file

anastasiamac:
!!build!!

provider/ec2/environ.go
```diff
@@ -1153,7 +1153,7 @@ func (e *environ) subnetsForVPC(ctx context.ProviderCallContext) (resp *ec2.Subn
 // AdoptResources is part of the Environ interface.
 func (e *environ) AdoptResources(ctx context.ProviderCallContext, controllerUUID string, fromVersion version.Number) error {
 	// Gather resource ids for instances, volumes and security groups tagged with this model.
-	instances, err := e.AllInstances(ctx)
+	instances, err := e.AllRunningInstances(ctx)
```
babbageclunk: Eventually this should definitely be AllInstances, but I guess you're still doing the mechanical transformation in this pass?

provider/lxd/environ_broker.go
```diff
@@ -324,6 +324,12 @@ func (env *environ) AllInstances(ctx context.ProviderCallContext) ([]instances.I
 	return instances, errors.Trace(err)
 }
 
+// AllRunningInstances implements environs.InstanceBroker.
+func (env *environ) AllRunningInstances(ctx context.ProviderCallContext) ([]instances.Instance, error) {
+	// We can only get Alive containers from lxd api which means that "all" is the same as "running".
```
babbageclunk: This comment isn't quite right - the AliveContainers method in container/lxd.Server includes stopping and stopped containers - there's nothing stopping us from adding another one which doesn't.

https://github.com/juju/juju/blob/develop/container/lxd/container.go#L148

provider/openstack/provider.go
```diff
@@ -1662,6 +1662,13 @@ func (e *Environ) AllInstances(ctx context.ProviderCallContext) ([]instances.Ins
 	return instances, nil
 }
 
+// AllRunningInstances returns all running, available instances in this environment.
+func (e *Environ) AllRunningInstances(ctx context.ProviderCallContext) ([]instances.Instance, error) {
```
babbageclunk: Even openstack does this?

provider/ec2/environ.go
anastasiamac: Yes, I'd rather we changed the calls when the need actually arises. However, if you are sure that AdoptResources should be dealing with all cloud instances irrespective of their state (I guess it does since this is essentially what we need for migration), I am happy to do a follow-up.

anastasiamac:
$$merge$$

PR #10202, opened 2019-05-17
Use actions for tokens

SimonRichardson:
## Description of change

The following uses actions on a charm, rather than using config
changes. Although the config changes ensure that a relation is set
and that the token can be read when the relation changes, the way
it sets the status and then read means that it's problematic when
reading the value in the tests. This is more evident when running
against a cloud like aws/gce, where latency comes in.

To improve this, we instead run an action for the charm to ensure
that the charm is still alive and that we can read the value from
the action result. This replicates some of the functionality of
the config setup, but the difference being that it's reliable to
work on the clouds.

## QA steps

Ensure you've setup the acceptance tests as per the readme.

```
cd acceptancetests
./assess_recovery.py myaws
```

### DIFF
--- a/acceptancetests/jujupy/client.py
+++ b/acceptancetests/jujupy/client.py
@@ -857,9 +857,9 @@ class ModelClient:
         """
         options = ()
         if force:
-            options = ('--force',)
+            options = options + ('--force',)
         if controller:
-            options = ('-m', 'controller',)
+            options = options + ('-m', 'controller',)
         self.juju('remove-machine', options + tuple(machine_ids))
         return self.make_remove_machine_condition(machine_ids)
 
@@ -1693,7 +1693,7 @@ class ModelClient:
         """Return the controller-member-status of the machine if it exists."""
         return info_dict.get('controller-member-status')
 
-    def wait_for_ha(self, timeout=1200, start=None):
+    def wait_for_ha(self, timeout=1200, start=None, quorum=3):
         """Wait for voiting to be enabled.
 
         May only be called on a controller client."""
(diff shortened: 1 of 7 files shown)


SimonRichardson:
https://trello.com/c/OdRAIdvq/1369-ci-backup-restore-on-aws-gce

acceptancetests/jujupy/client.py
```diff
@@ -857,9 +857,9 @@ def remove_machine(self, machine_ids, force=False, controller=False):
         """
         options = ()
         if force:
-            options = ('--force',)
+            options = options + ('--force',)
```
SimonRichardson: This was broken before if both force and controller was used.

acceptancetests/jujupy/client.py
```diff
@@ -1462,6 +1462,10 @@ def _get_substrate_constraints(self):
             # For now only maas support spaces in a meaningful way.
             return 'spaces={}'.format(','.join(
                 '^' + space for space in sorted(self.excluded_spaces)))
+        elif self.env.lxd:
+            # LXD should be constrained by memory when running in HA, otherwise
+            # mongo just eats everything.
+            return 'mem=6G'
```
SimonRichardson: @hmlanigan we need to set an upper limit.

SimonRichardson:
Keep note, the update to the test does work, but unfortunately it depends on when the restore happens if the machine is in position 0 once the other down machines have been restored. If it happens that it isn't then the test will fail. The idea behind this PR is to get us incrementally in a better place, at the moment nothing is covered for ha-backup and ha-restore.

The following output[1] is a case where the backup was restored successfully

I believe there is a `slice -> map -> slice` conversion somewhere when applying a backup during a restore, as the instances end up with the correct data, but in the wrong order and the order is never guaranteed, which feels like a map read issue.

1. https://pastebin.canonical.com/p/dpsndPjYKx/

acceptancetests/jujupy/client.py
hmlanigan: +1

SimonRichardson:
$$merge$$

PR #10203, opened 2019-05-17
Remove default memory constrainst for bootstrap in CI tests.

hmlanigan:
## Description of change

CI tests shouldn't have a mem=2G constraint when the default for a juju controller is 3.5G.  A follow on for the same change for LXD controllers.

# QA

Run an acceptance test using a non lxd environment.

### DIFF
--- a/acceptancetests/jujupy/client.py
+++ b/acceptancetests/jujupy/client.py
@@ -1457,24 +1457,19 @@ class ModelClient:
     def _get_substrate_constraints(self):
         if self.env.joyent:
             # Only accept kvm packages by requiring >1 cpu core, see lp:1446264
-            return 'mem=2G cpu-cores=1'
+            return 'cores=1'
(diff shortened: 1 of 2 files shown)


acceptancetests/jujupy/client.py
```diff
@@ -1457,24 +1457,19 @@ def _maas_spaces_enabled():
     def _get_substrate_constraints(self):
         if self.env.joyent:
             # Only accept kvm packages by requiring >1 cpu core, see lp:1446264
-            return 'mem=2G cpu-cores=1'
+            return 'cores=1'
```
SimonRichardson: is it `cpu-cores` or `cores`?

hmlanigan: should be cores.  i saw a warning during bootstrap of joyent: 
Warning: constraint "cpu-cores" is deprecated in favor of "cores".

hmlanigan:
$$merge$$

PR #10204, opened 2019-05-20
Avoid repetition in juju register warnings/errors

achilleasa:
## Description of change

When invoking a `juju register` command twice, you would get a warning message advising you that the controller is already known and how to connect to it, followed by an _error_ like `controller is already registered as "blah"`. 

This PR fixes the above stuttering issue and as a drive-by updates the logic behind `juju register $ip` to work similarly to the `juju login $ip` command, i.e when you try to register an endpoint IP for an already locally known controller you will get the same error like calling `juju register` twice.

## QA steps

```console

# Create a new user and grab the registration command
$ juju add-user bob
User "bob" added
Please send this command to bob:
    juju register MEMTA2JvYjAUExIxMC42NS40Ny4xMTU6MTcwNzAEIP4g-Nnh4plv7E11qLsOPK0bSKlZDkQhy4RHqdIODGTQEwR0ZXN0

"bob" has not been granted access to any models. You can use "juju grant" to grant access.

# Edit ~/.local/share/juju/controllers.yaml and change the controller entry key to "test1"
# Then, run the registration command (when prompted for a controller name use the default value: "test")
$ juju register MEMTA2JvYjAUExIxMC42NS40Ny4xMTU6MTcwNzAEIP4g-Nnh4plv7E11qLsOPK0bSKlZDkQhy4RHqdIODGTQEwR0ZXN0

Enter a new password:
Confirm password:
Enter a name for this controller [test]:
Initial password successfully set for bob.
ERROR This controller has already been registered on this client as "test1".
To login user "bob" run 'juju login -u bob -c test1'.
To update controller details and login as user "bob":
    1. run 'juju unregister bob'
    2. request from your controller admin another registration string, i.e
       output from 'juju change-user-password bob --reset'
    3. re-run 'juju register' with the registration from (2) above.

# Next, try to register the controller using the endpoint of the test controller (see controllers.yaml)
$ juju register 10.65.47.115:17070
ERROR This controller has already been registered on this client as "test1"
To login run 'juju login -c test1'.

# Finally, edit the controllers.yaml file and change the controller key back to "test".
# Then, run the registration command one more time
$ juju register 10.65.47.115:17070
ERROR This controller has already been registered on this client as "test".
To login user "admin" run 'juju login -u admin -c test'.
To update controller details and login as user "admin":
    1. run 'juju unregister admin'
    2. request from your controller admin another registration string, i.e
       output from 'juju change-user-password admin --reset'
    3. re-run 'juju register' with the registration from (2) above.
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1825166

### DIFF
--- a/cmd/juju/controller/register.go
+++ b/cmd/juju/controller/register.go
@@ -138,6 +138,15 @@ func (c *registerCommand) run(ctx *cmd.Context) error {
 	if err != nil {
 		return errors.Trace(err)
 	}
+
+	// Check if user is trying to register an already known controller by
+	// by providing the IP of one of its endpoints.
+	if registrationParams.publicHost != "" {
+		if err := ensureNotKnownEndpoint(c.store, registrationParams.publicHost); err != nil {
+			return errors.Trace(err)
+		}
+	}
+
 	controllerName, err := c.promptControllerName(registrationParams.defaultControllerName, ctx.Stderr, ctx.Stdin)
 	if err != nil {
 		return errors.Trace(err)
@@ -202,6 +211,7 @@ func (c *registerCommand) publicControllerDetails(host, controllerName string) (
 	if !strings.Contains(apiAddr, ":") {
 		apiAddr += ":443"
 	}
+
 	// Make a direct API connection because we don't yet know the
 	// controller UUID so can't store the thus-incomplete controller
 	// details to make a conventional connection.
@@ -621,4 +631,41 @@ type byteAtATimeReader struct {
 
 func (r byteAtATimeReader) Read(out []byte) (int, error) {
 	return r.Reader.Read(out[:1])
+}
+
+// ensureNotKnownEndpoint checks whether any controllers in the local client
+// cache contain the provided endpoint and returns an error if that is the
+// case.
+func ensureNotKnownEndpoint(store jujuclient.ClientStore, endpoint string) error {
+	existingDetails, existingName, err := store.ControllerByAPIEndpoints(endpoint)
+	if err != nil && !errors.IsNotFound(err) {
+		return errors.Trace(err)
+	}
+
+	if existingDetails == nil {
+		return nil
+	}
+
+	// Check if we know the username for this controller
+	accountDetails, err := store.AccountDetails(existingName)
+	if err != nil && !errors.IsNotFound(err) {
+		return errors.Trace(err)
+	}
+
+	if accountDetails != nil {
+		var buf bytes.Buffer
+		if err := alreadyRegisteredMessageT.Execute(
+			&buf,
+			map[string]interface{}{
+				"ControllerName": existingName,
+				"UserName":       accountDetails.User,
+			},
+		); err != nil {
+			return err
+		}
+		return errors.New(buf.String())
+	}
+
+	return errors.Errorf(`This controller has already been registered on this client as %q.
+To login run 'juju login -c %s'.`, existingName, existingName)
 }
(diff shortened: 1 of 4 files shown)


cmd/juju/controller/register.go
```diff
@@ -356,7 +365,7 @@ func (c *registerCommand) updateController(
 }
 
 var alreadyRegisteredMessageT = template.Must(template.New("").Parse(`
-This controller has already been registered on this client as "{{.ControllerName}}."
+This controller has already been registered on this client as "{{.ControllerName}}".
```
SimonRichardson: Nice driveby.

cmd/juju/controller/register.go
```diff
@@ -623,3 +632,40 @@ type byteAtATimeReader struct {
...
+// ensureNotKnownEndpoint checks whether any controllers in the local client
+// cache contain the provided endpoint and returns an error if that is the
+// case.
+func ensureNotKnownEndpoint(store jujuclient.ClientStore, endpoint string) error {
+	existingDetails, existingName, err := store.ControllerByAPIEndpoints(endpoint)
+	if err != nil && !errors.IsNotFound(err) {
+		return errors.Trace(err)
+	}
+
+	if existingDetails == nil {
```
SimonRichardson: I wonder if it would be good to also do this check?

```suggestion
	if existingDetails == nil || existingName == "" {
```

cmd/juju/controller/register.go
```diff
@@ -623,3 +632,40 @@ type byteAtATimeReader struct {
...
+	accountDetails, err := store.AccountDetails(existingName)
+	if err != nil && !errors.IsNotFound(err) {
+		return errors.Trace(err)
+	}
+
+	if accountDetails != nil {
+		var buf bytes.Buffer
+		if err := alreadyRegisteredMessageT.Execute(
+			&buf,
+			map[string]interface{}{
```
SimonRichardson: I'd rather we used a struct here, so we keep type some type safety by removing the stingified keys in the map.

cmd/juju/controller/register.go
achilleasa: If no local controller with this endpoint exists, L640 (`store.ControllerByAPIEndpoints`) will return a NotFound error and details will be nil (existingName will be blank too). So we don't need the extra check here. I don't think you can have a nameless controller, can you?

SimonRichardson: I didn't know hence why i thought the extra test might be worth doing, but if we can't get coverage then it's pointless having it.

cmd/juju/controller/register.go
achilleasa: Pushed https://github.com/juju/juju/pull/10204/commits/db81428327442f0f21af8286f45bac1f6abe9dfe to address this comment.

achilleasa:
!!build!!

(Failed due to `ProvisionerSuite.TestAvailabilityZoneMachinesStartMachinesAZFailuresWithDG`)

achilleasa:
!!build!!

(mongo failed to start)

achilleasa:
@mitechie Before I merge this, can you please take a look at the output from the QA steps and let me know if this is the correct `juju register` behavior or whether I broke something?

achilleasa:
@pmatulis https://github.com/juju/juju/pull/10204/commits/da3f7fb7724c3da9e1a99c97108aad629ce9446e fixes a typo in a pre-existing error message. Can you please check if we reference its content somewhere in the docs?

achilleasa:
$$merge$$

achilleasa:
$$merge$$

(failed because of `workerEnvironSuite.TestCharmProfilingInfoError`)

PR #10213, opened 2019-05-22
Raise error when timeout reached while result pending

timClicks:
**don't merge until manual testing completed**

When users currently run `juju show-action-output`  with the `--wait` option enabled, and that deadline is reached without output, Juju does not indicate that an error has occurred.

## QA steps

[Manual testing still required]

## Documentation changes

None. This change should make Juju match the current behaviour.

## Bug reference

* [lp:1827251](https://bugs.launchpad.net/juju/+bug/1827251)


timClicks:
Superseded by #10214

PR #10216, opened 2019-05-22
Improve help add-credential message for GCE

timClicks:
For GCE especially, it is be unclear what is intended by the options to `add-credential`. `jsonfile` and `oauth2` are not particularly intuitive. This PR includes two changes that aim to remedy this. 

One makes the interactive prompt more explicit about what it's looking for: 

```
$ juju add-credential google
...
Enter path to the .json file containing a service account key for your project
(detailed instructions available at https://discourse.jujucharms.com/t/1508)
Path: 
```

The other also expands our `juju help add-credential` help documentation.

## QA steps

None.

## Documentation changes

None.

## Bug reference

* [lp:1576301](https://bugs.launchpad.net/juju/+bug/1576301)

### DIFF
--- a/cmd/juju/cloud/addcredential.go
+++ b/cmd/juju/cloud/addcredential.go
@@ -26,47 +26,86 @@ var usageAddCredentialSummary = `
 Adds or replaces credentials for a cloud, stored locally on this client.`[1:]
 
 var usageAddCredentialDetails = `
-The user is prompted to add credentials interactively if a YAML-formatted
-credentials file is not specified. Here is a sample credentials file:
+The juju add-credential command operates in two modes.
 
-credentials:
-  aws:
-    <credential name>:
-      auth-type: access-key
-      access-key: <key>
-      secret-key: <key>
-  azure:
-    <credential name>:
-      auth-type: service-principal-secret
-      application-id: <uuid1>
-      application-password: <password>
-      subscription-id: <uuid2>
-  lxd:
-    <credential name>:
-      auth-type: interactive
-      trust-password: <password>
+When called with only the <cloud name> argument, ` + "`juju add-credential`" + ` will 
+take you through an interactive prompt to add a credential specific to 
+the cloud provider.
 
-A "credential name" is arbitrary and is used solely to represent a set of
-credentials, of which there may be multiple per cloud.
-The ` + "`--replace`" + ` option is required if credential information for the named
-cloud already exists locally. All such information will be overwritten.
-This command does not set default regions nor default credentials. Note
-that if only one credential name exists, it will become the effective
-default credential.
-For credentials which are already in use by tools other than Juju, ` + "`juju \nautoload-credentials`" + ` may be used.
-When Juju needs credentials for a cloud, i) if there are multiple
-available; ii) there's no set default; iii) and one is not specified ('--
-credential'), an error will be emitted.
+Providing the ` + "`-f <credentials.yaml>` " + `option switches to the 
+non-interactive mode. <credentials.yaml> must be a path to a correctly 
+formatted YAML-formatted file. Details of the format are provided at 
+"About credentials.yaml" below.
+
+The ` + "`--replace`" + ` option is required if credential information 
+for the named cloud already exists locally. All such information will be 
+overwritten.
+
+About credentials.yaml:
+Here is a sample credentials.yaml showing four credentials being stored 
+against three clouds:
+
+  credentials:
+    aws:
+      <credential-name>:
+        auth-type: access-key
+        access-key: <key>
+        secret-key: <key>
+    azure:
+      <credential-name>:
+        auth-type: service-principal-secret
+        application-id: <uuid>
+        application-password: <password>
+        subscription-id: <uuid>
+    lxd:
+      <credential-name>:
+        auth-type: interactive
+        trust-password: <password>
+      <credential-name>:
+        auth-type: interactive
+        trust-password: <password>
+
+More generally, here is a loosely defined grammar for credentials.yaml:
+
+  credentials:
+    <cloud-name>:
+      <credential-name>:
+        auth-type: <auth-type>
+        <auth-type-key>: <auth-type-value>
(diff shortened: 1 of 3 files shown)


pmatulis:
Documentation efforts have shied away from tracking web site instructions where possible. They simply become outdated too quickly and are hard to track. The Juju [GCE documentation page](https://docs.jujucharms.com/gce-cloud) uses the CLI tools. The tools can also accomplish other necessary tasks such as enabling APIs and linking a project to billing.

I think *some* extra information can be included in the help but not detailed information. I would also recommend a link be included that points to the above.

I think "cloud provider" should be considered an internal Juju term. I prefer the term "cloud type" instead in user-facing text. It's what they see in the output to `juju clouds`.

"2 options" should be "Two options".

Specifying two options but not explaining why a user would choose one over the other can lead to frustration on the part of the user.

I would like to see this:

> The `--replace` option is required if credential information for the named
cloud already exists locally. All such information will be overwritten.

changed to

> The `--replace` option is required if a credential name is being re-used
(i.e. the name already exists locally for the cloud in question). This option will
cause its current information to be overwritten.

timClicks:
These are very helpful suggestions, Peter. I'll amend the commit.

cmd/juju/cloud/addcredential.go
```diff
@@ -62,8 +65,26 @@ Examples:
     juju add-credential google
     juju add-credential aws -f ~/credentials.yaml
 
+Cloud-specific instructions:
```
wallyworld: The google instructions seem very specific and could easily become outdated if google changes the workflow on their end. Could we make the doc more generic. We just need to tell the user to obtain a json file or oauth2 token using the GCE web console.

cmd/juju/cloud/addcredential.go
```diff
@@ -62,8 +65,26 @@ Examples:
...
+               account" that has the "Compute Engine > Compute Admin" 
+               role.
+
+    oauth2     Click "Create credentials", then "OAuth client ID". 
+               Following the dialogue, Google will provide you with 
+               a client ID and secret ID.
+
 See also: 
-    credentials
+    credentialss
```
wallyworld: typo

cmd/juju/cloud/addcredential.go
timClicks: We can't really leave it as "obtain a json file or oauth2 token using the GCE web console". There are several methods for obtaining both. As indicated from the bug report, one of Juju's most experienced users wasn't able to figure that out.

I would like to be more helpful, even if there is a risk that things become stale. It's very easy to update documentation. It's very frustrating being a new user to Juju and getting stuck with credential issues.

Adding credentials should be as seemless as possible. At this stage, a new user is least likely to report any issues. They're nervous and undecided about whether Juju is a good tool.

My thinking after @pmatulis's comment was to remove the "how to" and keep things as "what is required". Juju's `jsonfile` means service account key in Google's terminology. Removing the steps to get one, but leaving in terms that are searchable might work.

timClicks: I've actually remove that addition to the help docs. I've changed the message in the interactive prompt so that it's clearer what's being asked for.

provider/gce/credentials.go
```diff
@@ -52,7 +52,7 @@ func (environProviderCredentials) CredentialSchemas() map[cloud.AuthType]cloud.C
 		cloud.JSONFileAuthType: {{
 			Name: credAttrFile,
 			CredentialAttr: cloud.CredentialAttr{
-				Description: "path to the .json file containing your Google Compute Engine project credentials",
+				Description: "path to a service account key",
```
wallyworld: can we keep the bit about it being a .json file, eg

"path to the .json file containing a service account key for your project"

cmd/juju/cloud/addcredential.go
```diff
@@ -26,41 +26,74 @@ var usageAddCredentialSummary = `
...
     juju add-credential aws -f ~/credentials.yaml
+    juju add-credential aws --replace -f ~/credentials.yaml
+
+Notes:
+If you setting up Juju for the first time, consider running
+` + "`juju autoload-credentials`" + `. This may help you skip adding credentials
+manually.
+
+This command does not set default regions nor default credentials for the
+cloud. The commands ` + "`juju set-default-region`" + ` and `+ "`juju set-default-region`" + `
```
wallyworld: These should be "See also" as well, plus looks like there's a type (dup command)

provider/gce/credentials.go
wallyworld: Maybe we can add a discourse link in the description as well

cmd/juju/cloud/addcredential.go
```diff
@@ -26,41 +26,74 @@ var usageAddCredentialSummary = `
...
+be unique within each <cloud-name>. This allows allow each cloud to store 
+multiple credentials.
 
 Examples:
     juju add-credential google
     juju add-credential aws -f ~/credentials.yaml
+    juju add-credential aws --replace -f ~/credentials.yaml
+
+Notes:
+If you setting up Juju for the first time, consider running
```
pmatulis: If you *are* setting up...

timClicks: Fixed, thanks for the close read :eyes:

cmd/juju/cloud/addcredential.go
timClicks: Good point. Have included it.

provider/gce/credentials.go
timClicks: I've added a link to discourse thread that I'll flesh out before 2.6 is released.

timClicks:
!!build!!

timClicks:
$$merge$$

timClicks:
[Improve add-credential error message](https://trello.com/c/pwrv7Pk1/778-improve-add-credential-error-message)

timClicks:
[Improve add-credential error message](https://trello.com/c/pwrv7Pk1/778-improve-add-credential-error-message)

PR #10219, opened 2019-05-23
Remove package level logger for provisioner and pass one in.

howbazaar:
A precursor to work to have workers run in controllers on behalf of other models go into the model logs. To do this, we need to remove the package level loggers.

Doing the provisioner first as it is one of the most asked for informational pieces for models.

There is currently no change in behaviour (apart from the removal of one call site because I couldn't thread the logger through easily).


howbazaar:
$$merge$$

PR #10226, opened 2019-05-24
Fix dupe k8s model detection

wallyworld:
## Description of change

Duplicate model detection for k8s models used to include the cloud name as well as the owner in the selection criteria. This was wrong as it allowed 2 models of the same name in state model.

There was no unit test for the functionality that was removed that used a different cloud name- there's an existing unit test for the standard dupe case.

## QA steps

Add a k8s model with the same name to different clouds and get an error.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1768857

### DIFF
--- a/state/model.go
+++ b/state/model.go
@@ -330,17 +330,15 @@ func (ctlr *Controller) NewModel(args ModelArgs) (_ *Model, _ *State, err error)
 
 	var prereqOps []txn.Op
 
-	if args.Type == ModelTypeIAAS {
-		// If no region specified and the cloud only has one, default to that.
-		if args.CloudRegion == "" && len(modelCloud.Regions) == 1 {
-			args.CloudRegion = modelCloud.Regions[0].Name
-		}
-		assertCloudRegionOp, err := validateCloudRegion(modelCloud, args.CloudRegion)
-		if err != nil {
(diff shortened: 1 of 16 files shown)


wallyworld:
[It's possible to add two models with same name on different k8s](https://trello.com/c/BmHGYHmJ/775-its-possible-to-add-two-models-with-same-name-on-different-k8s)

state/model.go
```diff
@@ -414,24 +414,16 @@ func (ctlr *Controller) NewModel(args ModelArgs) (_ *Model, _ *State, err error)
 		// the same "owner" and "name" in the collection. If the txn is
 		// aborted, check if it is due to the unique key restriction.
 		name := args.Config.Name()
-		qualifierTerm := bson.DocElem{"owner", owner.Id()}
-		if args.Type == ModelTypeCAAS {
```
ycliuhw: so for caas, we have owner as well, right?

ycliuhw:
test failed

state/model.go
wallyworld: yeah, all models have owners

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10228, opened 2019-05-24
Show effect of credential validity in model status.

anastasiamac:
## Description of change

Occasionally, cloud credentials used by the model become invalid.
This PR allows user to find out that this happened via 'juju models' or 'juju status'.
We now will tell the user that a model is suspended due to the credential being invalid. Once credential becomes valid again, the model status is reverted to whatever it was beforehand.

Model credential can become valid again if it is either updated with a new content or  a different, new credential is set for the model. 

If we cannot figure out what model status was prior to the model being suspended (for example, if status history for the model has been pruned), we will set model to 'available'.

## QA steps
1. bootstrap
2. add a model with a different credential
[at this stage, juju output would look similar to this]

```
$ juju models
Controller: mycontroller

Model       Cloud/Region   Type  Status     Machines  Cores  Access  Last connection
controller  aws/us-east-1  ec2   available         1      4  admin   just now
default     aws/us-east-1  ec2   available         0      -  admin   12 minutes ago
trial*      aws/us-east-1  ec2   available         0      -  admin   never connected

$ juju status
Model  Controller    Cloud/Region   Version  SLA          Timestamp
trial  mycontroller  aws/us-east-1  2.6.3.1  unsupported  17:01:53+10:00

Model "admin/trial" is empty.
```
3. disable second credential and try to deploy something on the model
[at this stage, juju output will change to something like this]

```
$ juju models
Controller: mycontroller

Model       Cloud/Region   Type  Status     Machines  Cores  Access  Last connection
controller  aws/us-east-1  ec2   available         1      4  admin   just now
default     aws/us-east-1  ec2   available         0      -  admin   6 minutes ago
trial*      aws/us-east-1  ec2   suspended         0      -  admin   8 seconds ago

$ juju status
Model  Controller    Cloud/Region   Version  SLA          Timestamp       Notes
trial  mycontroller  aws/us-east-1  2.6.3.1  unsupported  17:00:31+10:00  suspended since cloud credential is not valid

Model "admin/trial" is empty.
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1822117

### DIFF
--- a/core/status/status.go
+++ b/core/status/status.go
@@ -332,6 +332,7 @@ func ValidModelStatus(status Status) bool {
 		Available,
 		Busy,
 		Destroying,
+		Suspended, // For  model, this means that its cloud credential is invalid and model will not be doing any cloud calls.
 		Error:
 		return true
 	default:
--- a/core/status/status_test.go
+++ b/core/status/status_test.go
@@ -3,6 +3,7 @@
 package status_test
 
 import (
+	jc "github.com/juju/testing/checkers"
 	gc "gopkg.in/check.v1"
 
 	"github.com/juju/juju/core/status"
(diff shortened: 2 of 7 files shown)


anastasiamac:
[credential validity needs to be surfaced to the user](https://trello.com/c/lOhGKseo/751-credential-validity-needs-to-be-surfaced-to-the-user)

anastasiamac:
Looks like a windows problem finding dependencies...
Errors similar to 
17:55:58 helpers.go:11:2: cannot find package "github.com/juju/cmd" in any of:...

Might be a one of thing...
!!build!!

anastasiamac:
Closing in favor of different approach.

PR #10231, opened 2019-05-27
Add a timeout for lease manager stopping

babbageclunk:
## Description of change

Debugging logs generated when the http-server worker times out shutting down suggest that the lease manager is failing to stop in some cases. Add a timeout to the shutdown so we don't wait indefinitely if a background claim or expire doesn't finish in a timely manner.

Track how many outstanding claims and expires there are in flight and report these in a debug log dumped on timeout, as well as in the introspection report. Include goroutines since (if this is the only thing stopping the httpserver from stopping) the httpserver won't dump them when we timeout here.

The chain from httpserver to lease manager is:
* httpserver is waiting for the mux
* the mux is waiting for api-server to call ClientDone
* the api-server is waiting for a request to finish releasing a state
  back to the pool
* the release is stuck because the state in question is waiting for
  its workers to stop
* the state worker goroutine counts indicate that there's one more
  lease manager than all of the other workers (hub watcher, presence
  watcher, ping batcher)

## QA steps

* Bootstrap a controller.
* Running juju_engine_report shows the outstanding claims and expires for each lease manager.
* I don't have a way to induce the shutdown problem (otherwise I'd just fix it rather than adding more specific debugging), but to check the output of the log I commented out the early return leg of the select in `waitForGoroutines` and upgraded twice. The second time dumped the debug logging.

## Documentation changes
 None

## Bug reference
https://bugs.launchpad.net/bugs/1829635

### DIFF
--- a/cmd/jujud/agent/machine/manifolds.go
+++ b/cmd/jujud/agent/machine/manifolds.go
@@ -883,6 +883,7 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 			FSM:                  leaseFSM,
 			RequestTopic:         leaseRequestTopic,
 			Logger:               loggo.GetLogger("juju.worker.lease.raft"),
+			LogDir:               agentConfig.LogDir(),
 			PrometheusRegisterer: config.PrometheusRegisterer,
 			NewWorker:            leasemanager.NewWorker,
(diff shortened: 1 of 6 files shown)


babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #10232, opened 2019-05-27
Pass in logger and logging context to the logger worker.

howbazaar:
This branch changes the logger worker so the logging context and logger are passed in rather than depending on the package level logger and the default loggo context.

This is needed for the move to logging the workers run by the controller on behalf of a model to the model's log collection rather than the the controller's log collection.

## QA steps

Run a controller, logs still work.

## Documentation changes

No user facing impact.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -584,14 +584,14 @@
   revision = "a0ef8b74ebcffeeff9fc374854deb4af388f037e"
 
 [[projects]]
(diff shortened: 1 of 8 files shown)


howbazaar:
Yes, I thought about that too when I was writing it. Perhaps I will add for later.

howbazaar:
$$merge$$

PR #10233, opened 2019-05-27
Add useful error message when vSphere datastore not specified

timClicks:
We currently produce a confusing error message when we attempt to bootstrap without including a datastore config option. 

> ERROR failed to bootstrap model: cannot start bootstrap instance in any availability zone (aron.internal, aron.internal/Low, aron.internal/High, aron.internal/High/Child)

Looking into the logs, the actual error is attempting to find a datastore that matches an empty string:

> 13:37:15 DEBUG juju.provider.common bootstrap.go:216 failed to start instance in availability zone "aron.internal": could not find datastore ""

## QA steps

Manual bootstrap into vSphere in 3 configurations:

- single datastore (ok)
- multiple datastores available, none selected (error: config required)
- multiple datastores available, one selected (ok)
- multiple datastores available, missing selected (error: unable to find datastore)

## Documentation changes

Business logic to mention during bootstrap process for vSphere:

- when there are no datastores available, `juju bootstrap` fails
- if there's one available, Juju will use that
- when there are multiple datastores,  `--config bootstrap=<datastore>` is required

The error message repeats 4 times, as Juju retries a few times before stopping.

```plain
ERROR no datastore provided, select one from testing, ds1, ds2
ERROR no datastore provided, select one from testing, ds1, ds2
ERROR no datastore provided, select one from testing, ds1, ds2
ERROR no datastore provided, select one from testing, ds1, ds2
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1830452

### DIFF
--- a/provider/vsphere/environ_broker.go
+++ b/provider/vsphere/environ_broker.go
@@ -192,7 +192,7 @@ func (env *sessionEnviron) newRawInstance(
 		cons.RootDisk = &minRootDisk
 	}
 	defaultDatastore := env.ecfg.datastore()
-	if cons.RootDiskSource == nil || *cons.RootDiskSource == "" {
+	if (cons.RootDiskSource == nil || *cons.RootDiskSource == "") && defaultDatastore != "" {
 		cons.RootDiskSource = &defaultDatastore
 	}
 
--- a/provider/vsphere/internal/vsphereclient/createvm.go
+++ b/provider/vsphere/internal/vsphereclient/createvm.go
@@ -392,8 +392,9 @@ func (c *Client) selectDatastore(
 	ctx context.Context,
 	args CreateVirtualMachineParams,
 ) (*mo.Datastore, error) {
-	// Select a datastore. If the user specified one, use that; otherwise
-	// choose the first one in the list that is accessible.
+	// Select a datastore. If the user specified one, use that. When no datastore
+	// is provided and there is only datastore accessible, use that. Otherwise return
+	// an error and ask for guidance.
 	refs := make([]types.ManagedObjectReference, len(args.ComputeResource.Datastore))
 	for i, ds := range args.ComputeResource.Datastore {
 		refs[i] = ds.Reference()
@@ -402,23 +403,47 @@ func (c *Client) selectDatastore(
 	if err := c.client.Retrieve(ctx, refs, nil, &datastores); err != nil {
 		return nil, errors.Annotate(err, "retrieving datastore details")
 	}
+
+	var accessibleDatastores []mo.Datastore
+	var datastoreNames []string
+	for _, ds := range datastores {
+		if ds.Summary.Accessible {
+			accessibleDatastores = append(accessibleDatastores, ds)
+			datastoreNames = append(datastoreNames, ds.Name)
+		} else {
+			c.logger.Debugf("datastore %s is inaccessible", ds.Name)
+		}
+	}
+
+	if len(accessibleDatastores) == 0 {
+		return nil, errors.New("no accessible datastores available")
+	}
+
 	if args.Constraints.RootDiskSource != nil {
 		dsName := *args.Constraints.RootDiskSource
 		c.logger.Debugf("desired datasource %q", dsName)
+		c.logger.Debugf("accessible datasources %q", datastoreNames)
 		for _, ds := range datastores {
-			c.logger.Debugf("seen %q", ds.Name)
 			if ds.Name == dsName {
+				c.logger.Infof("selecting datastore %s", ds.Name)
 				return &ds, nil
 			}
 		}
-		return nil, errors.Errorf("could not find datastore %q", dsName)
+		datastoreNamesMsg := fmt.Sprintf("%q", datastoreNames)
+		datastoreNamesMsg = strings.Trim(datastoreNamesMsg, "[]")
+		datastoreNames = strings.Split(datastoreNamesMsg, " ")
+		datastoreNamesMsg = strings.Join(datastoreNames, ", ")
+		return nil, errors.Errorf("could not find datastore %q, datastore(s) accessible: %s", dsName, datastoreNamesMsg)
 	}
-	for _, ds := range datastores {
-		if ds.Summary.Accessible {
-			c.logger.Debugf("using datastore %q", ds.Name)
-			return &ds, nil
-		}
+
+	if len(accessibleDatastores) == 1 {
+		ds := accessibleDatastores[0]
+		c.logger.Infof("selecting datastore %s", ds.Name)
+		return &ds, nil
+	} else if len(accessibleDatastores) > 1 {
+		return nil, errors.Errorf("no datastore provided and multiple available: %q", strings.Join(datastoreNames, ", "))
 	}
+
 	return nil, errors.New("could not find an accessible datastore")
 }
 
--- a/provider/vsphere/environ_broker_test.go
+++ b/provider/vsphere/environ_broker_test.go
@@ -467,3 +467,30 @@ func (s *environBrokerSuite) TestStopInstancesPermissionError(c *gc.C) {
 		return s.env.StopInstances(ctx, "vm-0")
 	})
 }
+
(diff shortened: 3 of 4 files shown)


provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -402,24 +402,48 @@ func (c *Client) selectDatastore(
...
+	if len(accessibleDatastores) == 1 {
+		ds := accessibleDatastores[0]
+		c.logger.Infof("selecting datastore %s", ds.Name)
+		return &ds, nil
+	} else if len(accessibleDatastores) > 1 {
+		var datastoreNames []string
+		for _, ds :=range accessibleDatastores {
+			datastoreNames = append(datastoreNames, ds.Name)
+		}
+		c.logger.Errorf("no datastore provided, select one from %q", strings.Join(datastoreNames, ", "))
```
babbageclunk: Why not put this in the returned error?

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -402,24 +402,48 @@ func (c *Client) selectDatastore(
...
 	}
-	return nil, errors.New("could not find an accessible datastore")
+
+	if len(accessibleDatastores) == 1 {
+		ds := accessibleDatastores[0]
+		c.logger.Infof("selecting datastore %s", ds.Name)
+		return &ds, nil
+	} else if len(accessibleDatastores) > 1 {
+		var datastoreNames []string
+		for _, ds :=range accessibleDatastores {
```
babbageclunk: I'm probably paranoid, but this might be unpleasant if there are hundreds of datastores. Maybe cap it or don't collect the names if there are more than 10?

provider/vsphere/internal/vsphereclient/createvm.go
timClicks: Mainly because I see log messages as for the end user and errors for developers, but perhaps that's not entirely correct.

provider/vsphere/internal/vsphereclient/createvm.go
timClicks: Good point. I guess the point is that we the user to pick one. We could even just provide a number of datastores, rather than the names.

provider/vsphere/internal/vsphereclient/createvm.go
babbageclunk: Well, the error will be what the user sees if it stops the bootstrap. In fact, we shouldn't be logging this here, otherwise they'll see the message twice.

provider/vsphere/internal/vsphereclient/createvm.go
babbageclunk: Oh, I definitely think it's more user-friendly to tell the user about the datastores we can see rather than just making them go and work it out themselves. I'm just worried about the extreme cases.

provider/vsphere/environ_broker.go
```diff
@@ -192,7 +192,7 @@ func (env *sessionEnviron) newRawInstance(
 		cons.RootDisk = &minRootDisk
 	}
 	defaultDatastore := env.ecfg.datastore()
-	if cons.RootDiskSource == nil || *cons.RootDiskSource == "" {
+	if (cons.RootDiskSource == nil || *cons.RootDiskSource == "") && defaultDatastore != "" {
```
babbageclunk: I wrote a test for this (the easy bit, sorry!)

```go

func (s *environBrokerSuite) TestStartInstanceNoDatastoreSetting(c *gc.C) {
	startInstArgs := s.createStartInstanceArgs(c)
	res, err := s.env.StartInstance(s.callCtx, startInstArgs)
	c.Assert(err, jc.ErrorIsNil)

	s.client.CheckCallNames(c, "ComputeResources", "ResourcePools", "ResourcePools", "CreateVirtualMachine", "Close")
	call := s.client.Calls()[3]
	c.Assert(call.Args, gc.HasLen, 2)
	c.Assert(call.Args[0], gc.Implements, new(context.Context))
	c.Assert(call.Args[1], gc.FitsTypeOf, vsphereclient.CreateVirtualMachineParams{})

	createVMArgs := call.Args[1].(vsphereclient.CreateVirtualMachineParams)

	var expected *string
	c.Assert(createVMArgs.Constraints.RootDiskSource, gc.Equals, expected)

	var (
		arch     = "amd64"
		rootDisk = common.MinRootDiskSizeGiB("trusty") * 1024
	)
	c.Assert(res.Hardware, jc.DeepEquals, &instance.HardwareCharacteristics{
		Arch:           &arch,
		RootDisk:       &rootDisk,
		RootDiskSource: nil,
	})
}

```

provider/vsphere/internal/vsphereclient/createvm.go
wallyworld: > Mainly because I see log messages as for the end user and errors for developers, but perhaps that's not entirely correct.

It's the exact opposite - errors are for end users because that's what they see. Log messages are for developers because they are the ones that trawl through logs.

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -402,24 +402,48 @@ func (c *Client) selectDatastore(
 	if err := c.client.Retrieve(ctx, refs, nil, &datastores); err != nil {
 		return nil, errors.Annotate(err, "retrieving datastore details")
 	}
+
+	if len(datastores) == 0 {
+		c.logger.Errorf("no datastores available")
```
wallyworld: No need to log and return an error. The error will be logged.
Also, the message should include the requested datastore, eg
cannot use requested datastore "foo" because no datastores are available

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -402,24 +402,48 @@ func (c *Client) selectDatastore(
...
+		return nil, errors.New("no datastores available")
+	}
+
 	if args.Constraints.RootDiskSource != nil {
 		dsName := *args.Constraints.RootDiskSource
 		c.logger.Debugf("desired datasource %q", dsName)
 		for _, ds := range datastores {
-			c.logger.Debugf("seen %q", ds.Name)
+			c.logger.Debugf("seen %s", ds.Name)
```
wallyworld: There's not enough context in this debug message.
It should probably be Tracef as well. What really needs doing is that we record the names in a slice and include the names in the error message below.
could not find datastore %q from %q, dsName, availaableDatastores

timClicks: This is in a loop, so I don't know if we want to print `accessibleDatastores` every time. Also this opens us up to the concern that @babbageclunk raised about cases where there might be hundreds of datastores.

provider/vsphere/internal/vsphereclient/createvm.go
timClicks: At this stage, we don't know whether the user has specified a desired datastore. I can add a test here and format accordingly, but it might make the code more complicated.

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -402,22 +403,45 @@ func (c *Client) selectDatastore(
...
 			}
 		}
-		return nil, errors.Errorf("could not find datastore %q", dsName)
-	}
-	for _, ds := range datastores {
-		if ds.Summary.Accessible {
-			c.logger.Debugf("using datastore %q", ds.Name)
-			return &ds, nil
-		}
+		datastoreNamesMsg := fmt.Sprintf("%q", datastoreNames)
```
timClicks: This string wrangling results in a "well", "quoted", "list"

timClicks:
$$merge$$

timClicks:
[vSphere deployment fails without specifying a datastore when there are multiple](https://trello.com/c/XDAbtyLI/783-vsphere-deployment-fails-without-specifying-a-datastore-when-there-are-multiple)

timClicks:
$$merge$$

PR #10237, opened 2019-05-28
Credential validity visibility on model status.

anastasiamac:
## Description of change

Occasionally, cloud credentials used by the model become invalid.
This PR allows user to find out that this happened via 'juju models' or 'juju status'.
We now will tell the user that a model is suspended due to the credential being invalid. Once credential becomes valid again, the model status is reverted to whatever it was beforehand.

Model credential can become valid again if it is either updated with a new content or a different, new credential is set for the model.

## QA steps

1.    bootstrap
2.     add a model with a different credential
    [at this stage, juju output would look similar to this]
```
$ juju models
Controller: mycontroller

Model       Cloud/Region   Type  Status     Machines  Cores  Access  Last connection
controller  aws/us-east-1  ec2   available         1      4  admin   just now
default     aws/us-east-1  ec2   available         0      -  admin   12 minutes ago
trial*      aws/us-east-1  ec2   available         0      -  admin   never connected

$ juju status
Model  Controller    Cloud/Region   Version  SLA          Timestamp
trial  mycontroller  aws/us-east-1  2.6.3.1  unsupported  17:01:53+10:00

Model "admin/trial" is empty.
```
3.    disable second credential and try to deploy something on the model
    [at this stage, juju output will change to something like this]
```
$ juju models
Controller: mycontroller

Model       Cloud/Region   Type  Status     Machines  Cores  Access  Last connection
controller  aws/us-east-1  ec2   available         1      4  admin   just now
default     aws/us-east-1  ec2   available         0      -  admin   6 minutes ago
trial*      aws/us-east-1  ec2   suspended         0      -  admin   8 seconds ago

$ juju status
Model  Controller    Cloud/Region   Version  SLA          Timestamp       Notes
trial  mycontroller  aws/us-east-1  2.6.3.1  unsupported  17:00:31+10:00  suspended since cloud credential is not valid

Model "admin/trial" is empty.
```
4. re-enabling credential reverts model status and message to the original [shown in 2.]

## Bug reference

https://bugs.launchpad.net/juju/+bug/1822117

### DIFF
--- a/state/cloudcredentials.go
+++ b/state/cloudcredentials.go
@@ -15,6 +15,7 @@ import (
 	"gopkg.in/mgo.v2/txn"
 
 	"github.com/juju/juju/cloud"
+	"github.com/juju/juju/core/status"
 	"github.com/juju/juju/permission"
 )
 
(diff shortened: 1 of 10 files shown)


anastasiamac:
[credential validity needs to be surfaced to the user](https://trello.com/c/lOhGKseo/751-credential-validity-needs-to-be-surfaced-to-the-user)

state/cloudcredentials_test.go
```diff
@@ -112,6 +114,118 @@ func (s *CloudCredentialsSuite) TestUpdateCloudCredentialsExisting(c *gc.C) {
...
+	// 3. update credential to be invalid and check model is suspended
+	cred := cloud.NewCredential(cloud.AccessKeyAuthType, map[string]string{
+		"foo": "foo val",
+		"bar": "bar val",
+	})
+	cred.Invalid = true
+	cred.InvalidReason = "because it is really really invalid"
+	err := testSuite.State.UpdateCloudCredential(credentialTag, cred)
+	c.Assert(err, jc.ErrorIsNil)
+	assertModelStatus(c, testSuite.StatePool, modelUUID, status.Suspended)
```
wallyworld: can we also check (here or elsewhere) the suspended status info string as well?

state/cloudcredentials.go
```diff
@@ -116,21 +117,61 @@ func (st *State) UpdateCloudCredential(tag names.CloudCredentialTag, credential
...
 	annotationMsg := "updating cloud credentials"
+
+	existing, err := st.CloudCredential(tag)
+	if err != nil && !errors.IsNotFound(err) {
+		return errors.Annotatef(err, "fetching cloud credentials")
+	}
+	exists := err == nil
+	var revert []*Model
+	if exists {
+		// Existing credential will become valid after this call, and
```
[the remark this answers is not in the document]
anastasiamac: I do not think it makes that much difference to keep it as is or to change as per suggestion. What's the value in changing?

state/cloudcredentials_test.go
anastasiamac: Yes

anastasiamac:
I am actually missing one more commit from here since my previous work set model status and everything worked fine. However, now that we are only adding histories and massaging the actual model status, when running 'juju models' which does not go through model struct but rather has a direct database query, the 'juu models' does not reflect 'suspension'. I need to add another logic massage for the part that parses query result into a param struct (by-passing state.Model entirely).

anastasiamac:
unrelated, intermittent failures in github.com/juju/juju/state/presence.TestPackage

anastasiamac:
!!build!!

anastasiamac:
unrelated intermittent windows failures

anastasiamac:
!!build!!

anastasiamac:
$$merge$$

anastasiamac:
$$merge$$

PR #10239, opened 2019-05-28
Update log db storage to not use tags.

howbazaar:
Once we have a log entry we never use it to retrieve the underlying entity. We just spend time getting the value in and out of a tag again on the server side. None of these changes impact the apiserver streaming format nor how the values are actually stored in the database.

## QA steps

Bootstrap LXD
Deploy ubuntu-lite
Check both debug-log command, and SSH into the machines to check the logs on disk. Nothing there changes.

## Documentation changes

No user facing change at all.


howbazaar:
$$merge$$

PR #10240, opened 2019-05-28
Bring in worker.v1 dependency changes.

jameinel:
## Description of change

Change the BackoffResetTime to 1 minute so that any errors within 1
minute of starting will be treated as things that should be retried, but
not on a fast cadence.

## QA steps

```
$ juju bootstrap lxd lxd
// Trigger some sort of error in the workers. An obvious on is 'out of disk space'.
$ juju debug-log
// Watch that the workers are being restarted, but each time they take a bit 
// longer to come back up again.
```

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1827664

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1540,7 +1540,7 @@
   revision = "fd59336b4621bc2a70bf96d9e2f49954115ad19b"
 
 [[projects]]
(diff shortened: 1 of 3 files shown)


jameinel:
[update 2.5/2.6 for backoff patch](https://trello.com/c/kpqP7Xwg/1431-update-25-26-for-backoff-patch)

wallyworld:
$$merge$$

PR #10245, opened 2019-05-29
undertaker: Remove model if cloud-destroy fails and --force is specified

babbageclunk:
## Description of change
There are various situations that might cause the undertaker's Destroy() call to fail - examples we've seen include kubernetes not being available or problems with other clouds. If the user has specified `--force` on the `destroy-model` command we should remove the model in spite of those errors.

## QA steps

* Bootstrap a controller and add a k8s to it.
* Create a k8s model.
* Stop the k8s instance.
* Destroy the model - the model will hang around and the undertaker will log errors about not being able to destroy it in k8x.
* Force-destroy the model - it should get removed successfully.

## Documentation changes
None

## Bug reference
https://bugs.launchpad.net/juju/+bug/1830216
https://bugs.launchpad.net/juju/+bug/1828870
https://bugs.launchpad.net/juju/+bug/1829423

### DIFF
--- a/apiserver/facades/controller/undertaker/mock_test.go
+++ b/apiserver/facades/controller/undertaker/mock_test.go
@@ -98,11 +98,12 @@ func (m *mockState) ModelUUID() string {
 // mockModel implements Model interface and allows inspection of called
 // methods.
 type mockModel struct {
-	tod   time.Time
-	owner names.UserTag
-	life  state.Life
-	name  string
-	uuid  string
+	tod    time.Time
+	owner  names.UserTag
+	life   state.Life
+	name   string
+	uuid   string
+	forced bool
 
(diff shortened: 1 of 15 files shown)


babbageclunk:
!!build!!
Build job couldn't get to github.

babbageclunk:
I can confirm that `destroy-model` (no `--force`) and then `destroy-model --force` works.

babbageclunk:
$$merge$$

babbageclunk:
The new modelDoc attribute broke an upgrade test - made it omitempty.

$$merge$$

babbageclunk:
Looks like the worker/instancemutater tests have an intermittent failure - fails under stress-race after 16 runs on my machine.

$$merge$$

PR #10247, opened 2019-05-29
Fix issue with upgrading k8s charms with hook errors

wallyworld:
## Description of change

The k8s operator was incorrectly trying to use IAAS deploy operations instead of no-op deploys in some circumstances. An extra check is added to the deployer operation to error if it is constructed from a caas model.

## QA steps

Deploy a k8s charm with a hook error
Fix the hook error
upgrade-charm --force-units, ensure the hook is rerun and error is gone

Also check that upgrading the charm with the hook error still in place allows juju resolved to cause the hook to be rerun.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1830259

### DIFF
--- a/worker/caasoperator/caasoperator.go
+++ b/worker/caasoperator/caasoperator.go
@@ -404,7 +404,9 @@ func (op *caasOperator) loop() (err error) {
 						return err
 					}
 				} else {
-					aliveUnits[unitId] = make(chan struct{})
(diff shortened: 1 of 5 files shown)


wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10250, opened 2019-05-29
Fix/caas cmr egress ingress

ycliuhw:
## Description of change

Fix CMR on caas models.
Now relation data bad can return correct `ingress_address` and `egress_subnets`;

## QA steps

1. deploy mariadb on caas model 1;

```bash
$ juju deploy cs:~juju/mariadb-k8s-0 --debug 
```

2. offer db;

```bash
$ juju offer $MODEL1.mariadb-k8s:server
```

3. deploy mediawiki on caas model 2;

```bash
$ juju deploy cs:~juju/mediawiki-k8s-3 --debug  --config kubernetes-service-type=LoadBalancer
```

4. relate;

```bahs
$ juju relate mediawiki-k8s:db $MODEL1.mariadb-k8s
```

## Documentation changes

No(bug fix);

## Bug reference

https://bugs.launchpad.net/juju/+bug/1830252

### DIFF
--- a/cmd/juju/caas/gke.go
+++ b/cmd/juju/caas/gke.go
@@ -185,9 +185,6 @@ func (g *gke) listClusters(account, project, region string) (map[string]cluster,
 	cmd := []string{
 		"gcloud", "container", "clusters", "list", "--filter", "status:RUNNING", "--account", account, "--project", project, "--format", "value\\(name,zone\\)",
 	}
-	if region != "" {
-		cmd = append(cmd, "--region", region)
-	}
 	result, err := runCommand(g, cmd, "")
 	if err != nil {
 		return nil, errors.Trace(err)
@@ -203,7 +200,7 @@ func (g *gke) listClusters(account, project, region string) (map[string]cluster,
 		if len(parts) == 0 {
 			continue
 		}
-		c := cluster{name: parts[0]}
+		c := cluster{name: parts[0], region: region}
 		if len(parts) > 1 {
 			c.zone = parts[1]
 			region := regexp.MustCompile(`([a-z]+-[a-z0-9]+)`).FindStringSubmatch(c.zone)
--- a/cmd/juju/caas/gke.go
+++ b/cmd/juju/caas/gke.go
@@ -4,6 +4,7 @@
 package caas
 
 import (
+	"regexp"
 	"fmt"
 	"io"
 	"os"
@@ -51,9 +52,9 @@ func (g *gke) getKubeConfig(p *clusterParams) (io.ReadCloser, string, error) {
 		cmd = append(cmd, "--project", p.project)
 		qualifiedClusterName += p.project + "_"
 	}
-	if p.region != "" {
-		cmd = append(cmd, "--region", p.region)
-		qualifiedClusterName += p.region + "_"
+	if p.zone != "" {
+		cmd = append(cmd, "--zone", p.zone)
+		qualifiedClusterName += p.zone + "_"
 	}
 	qualifiedClusterName += p.name
 
@@ -89,7 +90,7 @@ func (g *gke) interactiveParams(ctxt *cmd.Context, p *clusterParams) (*clusterPa
 	}
 
 	if p.name == "" {
-		p.name, p.region, err = g.queryCluster(pollster, p.credential, p.project, p.region)
+		p.name, p.region, p.zone, err = g.queryCluster(pollster, p.credential, p.project, p.region)
 		if err != nil {
 			return nil, errors.Trace(err)
 		}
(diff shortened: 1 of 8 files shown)


cmd/juju/caas/gke.go
```diff
@@ -202,26 +200,30 @@ func (g *gke) listClusters(account, project, region string) (map[string]cluster,
 		if len(parts) == 0 {
 			continue
 		}
-		c := cluster{name: parts[0]}
+		c := cluster{name: parts[0], region: region}
 		if len(parts) > 1 {
-			c.region = parts[1]
+			c.zone = parts[1]
+			region := regexp.MustCompile(`([a-z]+-[a-z0-9]+)`).FindStringSubmatch(c.zone)
```
wallyworld: regexp compile should be outside the loop, as a module var

cmd/juju/caas/gke.go
```diff
@@ -184,9 +185,6 @@ func (g *gke) listClusters(account, project, region string) (map[string]cluster,
 	cmd := []string{
 		"gcloud", "container", "clusters", "list", "--filter", "status:RUNNING", "--account", account, "--project", project, "--format", "value\\(name,zone\\)",
 	}
-	if region != "" {
```
wallyworld: removing this means that we are no longer filtering only those clusters in the specified region; we will get all clusters in any region

state/relationunit.go
```diff
@@ -613,34 +613,29 @@ func NetworksForRelation(
 	// is bound to the default space, we need to look up the ingress
 	// address info which is aware of cross model relations.
 	if boundSpace == environs.DefaultSpaceName || err != nil {
-		_, crossmodel, err := rel.RemoteApplication()
```
wallyworld: We need the cross model check.
If not cross model, we must not return the public address

state/unit.go
```diff
@@ -1267,11 +1266,9 @@ func (u *Unit) serviceAddress(scope string) (network.Address, error) {
 	if len(addresses) == 0 {
 		return network.Address{}, network.NoAddressError(scope)
 	}
-
 	getStrictPublicAddr := func(addresses []network.Address) (network.Address, bool) {
 		addr, ok := network.SelectPublicAddress(addresses)
-		ok = ok && addr.Scope == network.ScopePublic
-		return addr, ok
+		return addr, ok && (addr.Scope == network.ScopePublic || addr.Scope == network.ScopeCloudLocal)
```
wallyworld: getStrictPublicAddr is not the correct name anymore

worker/caasunitprovisioner/application_worker.go
```diff
@@ -175,6 +175,12 @@ func (aw *applicationWorker) loop() error {
 			if err != nil && !errors.IsNotFound(err) {
 				return errors.Trace(err)
 			}
+			// update svc info (addresses etc.) cloudservices.
+			if err = updateApplicationService(
```
wallyworld: This shouldn't be done here - we call clusterChanged) below.
Doing it here means we are making 2 API calls when we only need one

state/relationunit.go
```diff
@@ -652,36 +652,28 @@ func NetworksForRelation(
...
-			} else {
-				network.SortAddresses(addr)
-				for _, a := range addr {
-					ingress = append(ingress, a.Value)
-				}
 			}
 		}
 	}
+	if len(ingress) == 0 {
+		address, err := fetchAddr(unit.PrivateAddress)
```
wallyworld: This also should be up in the if crossmodel{} block to replace address, err = unit.PrivateAddress(). We don't want to delete that logic.
We should retain the if unit.ShouldBeAssigned() {} else {} logic and have this in the else{} bit.

state/unit.go
wallyworld: After the latest discussion, it needs to stay as StrictPublic.
Since we will fall back to Private() which will return the CLoudLocal

worker/caasunitprovisioner/deployment_worker.go
```diff
@@ -183,15 +183,25 @@ func (w *deploymentWorker) loop() error {
...
+			); err != nil {
 				return errors.Trace(err)
 			}
 			serviceUpdated = true
 		}
 	}
 }
+
+func updateApplicationService(appTag names.ApplicationTag, svc *caas.Service, updater ApplicationUpdater) error {
+	if svc.Id == "" {
```
ycliuhw: without this, an empty doc will be created then assertion failed in later `update` ops because we are using `svc.Id` for unq ID `provider-id`

ycliuhw:
$$merge$$

mitechie:
!!build!!

PR #10261, opened 2019-05-30
Improve model destruction messaging.

anastasiamac:

Destroying a model can take a while. Previously, our messaging would repeat 
```
Waiting on model to be removed...
Waiting on model to be removed...
```
This PR improves it as per suggestion in the bug below.

As drive-by, this PR abort attempts at destroying a model with an invalid cloud credential if '--force' was not used.

In addition, previously when destroying models, if there were no application and machines left, we were not reporting to the user that they may have volumes and/or filesystems left. 

## QA steps

1. bootstrap
2. add model with a different credential and invalidate it.
3. destroy model (without force) should give a similar output:
```
$ juju destroy-model trial
WARNING! This command will destroy the "trial" model.
This includes all machines, applications, data and other resources.

Continue [y/N]? y
Destroying model
ERROR failed to destroy model "trial"
ERROR cannot destroy model: invalid cloud credential, use --force
```
4. long running destroy model will now look similar to:
```
$ juju destroy-model trial -y
Destroying model
Waiting for model to be removed............
Model destroyed.
```

Long lines are now wrapped at 80 characters. So, the above example could look like this [my live testing output...]:
```
$ juju destroy-model trial -y
Destroying model
Waiting for model to be removed.................................................
.................
Model destroyed.
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1830935

### DIFF
--- a/cmd/juju/model/destroy.go
+++ b/cmd/juju/model/destroy.go
@@ -394,6 +394,14 @@ type modelData struct {
 	errorCount       int
 }
 
+func (data *modelData) isEmpty() bool {
+	return data.errorCount == 0 &&
+		data.machineCount == 0 &&
+		data.applicationCount == 0 &&
+		data.volumeCount == 0 &&
+		data.filesystemCount == 0
+}
+
 func waitForModelDestroyed(
 	ctx *cmd.Context,
 	api DestroyModelAPI,
@@ -417,6 +425,7 @@ func waitForModelDestroyed(
 	// no wait for 1st time.
 	intervalSeconds := 0 * time.Second
 	timeoutAfter := clock.After(timeout)
+	printedOnce := false
 	for {
 		select {
 		case <-interrupted:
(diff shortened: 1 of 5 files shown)


anastasiamac:
[destroy-model improvements](https://trello.com/c/wiDptDft/785-destroy-model-improvements)

cmd/juju/model/destroy.go
```diff
@@ -432,7 +441,13 @@ func waitForModelDestroyed(
 				// model has been destroyed successfully.
 				return nil
 			}
-			ctx.Infof(formatDestroyModelInfo(data) + "...")
+			msg := formatDestroyModelInfo(data)
+			if reported == msg {
+				fmt.Fprint(ctx.Stderr, ".")
```
SimonRichardson: Out of interest, does this auto wrap if this get's too long?

pmatulis: Yeah, I noticed when I removed a CDK model recently that there were hundreds of repeated lines.

anastasiamac: Interesting question... I am happy to wrap but it does not seem like I can easily determine terminal width or any other parameters to customize wrapping based on the user client. This means that I'd have to arbitrary wrap at a reasonable string length... I am picking 80 \o/ At 80 characters per line, I'd be wrapping to the next one.

anastasiamac:
unrelated failure pylibjuju in test_application_annotations

anastasiamac:
$$merge$$

PR #10266, opened 2019-05-31
Use service addresses for k8s charms

wallyworld:
## Description of change

The addresses provided to k8s charms are tweaked.

Key changes:
- k8s unit AllAddresses() returns only the service addresses, unless there's no service in which case the pod addresses are returned
- the address watcher used by the uniter watches the services addresses, not pod addresses
- when entering scope, only cloud local addresses are used (if any)
- for network-get calls, public addresses are polled if the k8s service type is load balancer or external ip

## QA steps

Deploy both cmr and non-cmr k8s scenarios and ensure relations are correctly created.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1830252

### DIFF
--- a/apiserver/facades/agent/uniter/uniter.go
+++ b/apiserver/facades/agent/uniter/uniter.go
@@ -14,6 +14,7 @@ import (
 	"github.com/juju/loggo"
 	"gopkg.in/juju/charm.v6"
 	"gopkg.in/juju/names.v2"
+	k8score "k8s.io/api/core/v1"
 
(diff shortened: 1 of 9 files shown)


wallyworld:
[k8s charm sees no egress-subnets on its own relation 2](https://trello.com/c/Mg76lmtj/787-k8s-charm-sees-no-egress-subnets-on-its-own-relation-2)

wallyworld:
$$merge$$

PR #10268, opened 2019-05-31
Multi-cloud CLI defaults to --local unless feature flag set

wallyworld:
## Description of change

The various cloud CLI commands default to --local unless the "multi-cloud" feature flag is set.
You can still explicitly specify a controller with --controller even without the feature flag.

## QA steps

Without feature flag, juju clouds shows all the out of the box clouds and requires --controller to show controller clouds.
With feature flag, juju clouds requires --local to show out of the box clouds.
Same with other cloud commands.

## Documentation changes

Doc based on previous release notes should be amended.

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -27,6 +27,7 @@ import (
 	jujucmdcloud "github.com/juju/juju/cmd/juju/cloud"
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
+	"github.com/juju/juju/feature"
 	"github.com/juju/juju/jujuclient"
 )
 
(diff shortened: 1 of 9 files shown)


cmd/juju/caas/add.go
```diff
@@ -160,9 +161,12 @@ type AddCAASCommand struct {
 func NewAddCAASCommand(cloudMetadataStore CloudMetadataStore) cmd.Command {
 	store := jujuclient.NewFileClientStore()
 	cmd := &AddCAASCommand{
-		OptionalControllerCommand: modelcmd.OptionalControllerCommand{Store: store},
-		cloudMetadataStore:        cloudMetadataStore,
-		store:                     store,
+		OptionalControllerCommand: modelcmd.OptionalControllerCommand{
+			Store:       store,
+			EnabledFlag: feature.MultiCloud,
```
howbazaar: Is this part of the add-k8s command? If so, I don't think we need the multicloud feature flag on this right?

wallyworld:
$$merge$$

PR #10271, opened 2019-06-01
remove-machine: Add forceRemoveMachine fallback cleanup

babbageclunk:
## Description of change

We could see situations where `remove-machine --force` would still leave the machine in the model and unable to be removed, which could also prevent destroying the model. A specific case was when the model's cloud credential is invalid, so the provisioner can't run.

Now `remove-machine --force` will schedule a fallback cleanup that will remove the machine after a timeout if it hasn't already been removed by the provisioner.

## QA steps

* Bootstrap a controller with a credential that you can disable.
* Deploy a unit.
* Disable the credential.
* Remove the machine with `--force --no-wait` - it should be removed from the model.

## Documentation changes
@pmatulis It should be made clear in the documentation for `remove-machine --force` that it can leave cloud instances behind and that they should be cleaned up manually.

## Bug reference
None

### DIFF
--- a/state/cleanup.go
+++ b/state/cleanup.go
@@ -331,8 +331,6 @@ func (st *State) cleanupMachinesForDyingModel(cleanupArgs []bson.Raw) (err error
 			}
 			continue
 		}
-		// TODO (force 2019-04-26) Should this always be ForceDestroy
-		// or only when 'destroy-model --force' is specified?...
 		if err := m.ForceDestroy(args.MaxWait); err != nil {
 			err = errors.Annotatef(err, "while destroying machine %v is", m.Id())
 			// TODO (force 2019-4-24) we should not break out here but continue with other machines.
--- a/state/cleanup.go
+++ b/state/cleanup.go
@@ -41,6 +41,7 @@ const (
 	cleanupApplicationsForDyingModel     cleanupKind = "applications"
 	cleanupDyingMachine                  cleanupKind = "dyingMachine"
 	cleanupForceDestroyedMachine         cleanupKind = "machine"
+	cleanupForceRemoveMachine            cleanupKind = "forceRemoveMachine"
 	cleanupAttachmentsForDyingStorage    cleanupKind = "storageAttachments"
 	cleanupAttachmentsForDyingVolume     cleanupKind = "volumeAttachments"
 	cleanupAttachmentsForDyingFilesystem cleanupKind = "filesystemAttachments"
@@ -175,6 +176,8 @@ func (st *State) Cleanup() (err error) {
 			err = st.cleanupDyingMachine(doc.Prefix, args)
 		case cleanupForceDestroyedMachine:
 			err = st.cleanupForceDestroyedMachine(doc.Prefix, args)
+		case cleanupForceRemoveMachine:
+			err = st.cleanupForceRemoveMachine(doc.Prefix, args)
 		case cleanupAttachmentsForDyingStorage:
 			err = st.cleanupAttachmentsForDyingStorage(doc.Prefix, args)
 		case cleanupAttachmentsForDyingVolume:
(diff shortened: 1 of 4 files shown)


babbageclunk:
[fallback cleanup for remove-machine --force](https://trello.com/c/6FVxo9bN/793-fallback-cleanup-for-remove-machine-force)

pmatulis:
@babbageclunk I'm confused. These statements seem contradictory:

> It should be made clear in the documentation for remove-machine --force that it can leave cloud instances behind and that they should be cleaned up manually.

> Now remove-machine --force will schedule a fallback cleanup that will remove the machine after a timeout if it hasn't already been removed by the provisioner.

anastasiamac:
@pmatulis,
"[Juju] can leave cloud instances behind" under normal circumstances - for example, if you machines are manual or if harvest provisioning mode is not set. However, here @babbageclunk refers to circumstances where, for example, we cannot reach the cloud instance. So, it is possible that the machine was removed form our records but cloud instance is till there. This needs to be highlighted in documentation.

anastasiamac:
!!build!!

anastasiamac:
Another state test needs to be adjusted with this change...

Currently model_test.go:721: ModelSuite.TestDestroyControllerAndHostedModelsWithResources fails with:

[LOG] 0:00.637 DEBUG juju.state model 0193e8 cleanup: dyingMachine("0")
model_test.go:770:
    assertCleanupCount(c, otherSt, 3)
model_test.go:1803:
    c.Assert(actual, jc.IsFalse)
... obtained bool = true

state/cleanup.go
```diff
@@ -328,8 +331,8 @@ func (st *State) cleanupMachinesForDyingModel(cleanupArgs []bson.Raw) (err error
 			}
 			continue
 		}
-		// TODO (force 2019-04-26) Should this always be ForceDestroy or only when
-		// 'destroy-model --force' is specified?...
+		// TODO (force 2019-04-26) Should this always be ForceDestroy
+		// or only when 'destroy-model --force' is specified?...
```
anastasiamac: This TODO needs to go - I have put it in before realizing that for dying models we always want to force destroy non-manual machines. It's redundant now.

state/cleanup_test.go
```diff
@@ -395,9 +395,9 @@ func (s *CleanupSuite) TestCleanupForceDestroyedControllerMachine(c *gc.C) {
 	s.assertCleanupRuns(c)
 	c.Assert(machine.SetHasVote(false), jc.ErrorIsNil)
 	// However, if we remove the vote, it can be cleaned up.
-	// ForceDestroy sets up a cleanupForceDestroyedMachine, which calls EnsureDead which sets up a cleanupDyingMachine
-	// so it takes 2 cleanup runs to run clear
-	s.assertCleanupCount(c, 2)
+	// ForceDestroy sets up a cleanupForceDestroyedMachine, which calls EnsureDead which sets up a cleanupDyingMachine, which in turn creates a delayed cleanupForceRemoveMachine.
```
anastasiamac: EnsureDead call has been removed... Looks like this comment needs updating.

state/cleanup_test.go
```diff
@@ -1186,17 +1221,33 @@ func (s *CleanupSuite) assertCleanupCount(c *gc.C, count int) {
...
 }
 
-func assertUnitLife(c *gc.C, unit *state.Unit, expected state.Life) {
-	err := unit.Refresh()
+type lifeChecker interface {
+	Refresh() error
+	Life() state.Life
+}
+
+func assertLifeIs(c *gc.C, thing lifeChecker, expected state.Life) {
```
anastasiamac: 'thing' :)

state/cleanup.go
babbageclunk: Ok done!

state/cleanup_test.go
babbageclunk: Yeah, it's changed to a call to `advanceLifecycle` so that I could pass maxWait down. Updated.

babbageclunk:
$$merge$$

babbageclunk:
MachineLegacyLeasesSuite.TestDyingMachine failed again - taking a look at it. :(
$$merge$$

anastasiamac:
Oh thanks! I've seen it recently too :(

PR #10272, opened 2019-06-01
undertaker: Enable force-destroying model with invalid cloud credential

babbageclunk:
## Description of change

If the model's cloud credential was invalid the model couldn't be destroyed, even with `--force`. This was because the undertaker was guarded by `ifCredentialValid`. 

Removing the undertaker's guard wasn't sufficient, because the environ tracker is also guarded, so the undertaker's manifold is changed so it uses a custom environment destroyer if the environ tracker is unavailable (rather than just stopping). The custom destroyer always fails (because it can't talk to the environment at the moment) but the changes in #10245 mean that `destroy-model --force` will still be able to remove the model from the controller.

## QA steps

* Bootstrap a controller against a cloud where you can disable the credential used.
* Add models m1 and m2.
* Deploy an application in m1.
* Disable the credential.
* Add a unit to m1 to trigger suspension.
* Destroying m2 fails because the credential is invalid.
* Destroying m2 with `--force` succeeds, while logging an error that the cloud environment couldn't be torn down.

## Documentation changes

@pmatulis it should be made clear in the destroy-model docs that if a model is force-destroyed then resources might not be cleaned up in the underlying cloud.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1827988


babbageclunk:
[force destroy models with invalid cloud credential](https://trello.com/c/ofZohWLb/737-force-destroy-models-with-invalid-cloud-credential)

anastasiamac:
!!build!!

cmd/jujud/agent/model/manifolds.go
```diff
@@ -253,15 +253,13 @@ func commonManifolds(config ManifoldsConfig) dependency.Manifolds {
 		})),
 		statusHistoryPrunerName: ifNotMigrating(pruner.Manifold(pruner.ManifoldConfig{
 			APICallerName: apiCallerName,
-			EnvironName:   environTrackerName,
```
anastasiamac: Thank you \o/

cmd/jujud/agent/model/manifolds_test.go
```diff
@@ -361,7 +359,6 @@ var expectedCAASModelManifoldsWithDependencies = map[string][]string{
 		"environ-upgrade-gate",
 		"environ-upgraded-flag",
 		"not-alive-flag",
-		"valid-credential-flag",
```
anastasiamac: It's strange that you are removing ifCredentialValid wrap from CAAS dependencies for undertaker but not IAAS? Unles it's just a test that you've missed... Because I can still see it as expected on line 613.

worker/undertaker/manifold.go
```diff
@@ -32,7 +33,12 @@ func (config ManifoldConfig) start(context dependency.Context) (worker.Worker, e
 		return nil, errors.Trace(err)
 	}
 	var destroyer environs.CloudDestroyer
-	if err := context.Get(config.CloudDestroyerName, &destroyer); err != nil {
+	if err := context.Get(config.CloudDestroyerName, &destroyer); isErrMissing(err) {
```
anastasiamac: Nice \o/

worker/undertaker/manifold.go
```diff
@@ -69,3 +75,18 @@ func Manifold(config ManifoldConfig) dependency.Manifold {
...
+
+// unavailableDestroyer is an environs.CloudDestroyer that always
+// fails to destroy. We use it when the real environ isn't available
+// because the cloud credentials are invalid so that the undertaker
+// can still remove the model if destruction is forced.
+type unavailableDestroyer struct{}
+
+// Destroy is part of environs.CloudDestroyer.
+func (*unavailableDestroyer) Destroy(ctx envcontext.ProviderCallContext) error {
+	return errors.New("cloud environment unavailable")
```
anastasiamac: I am not sure that we want to return an error here...

At the very least, I would have expected to a change on the worker side - how do we handle the error in the actual undertaker worker? Currently, we'd err out in the run() ... But since it is valid for us now to have an undertaker that does not communicate with the cloud, maybe u.config.Destroyer.Destroy() call should not err?

cmd/jujud/agent/model/manifolds_test.go
babbageclunk: This is a transitive dependency from environ tracker - the IAAS undertaker depends on environ tracker and that's guarded by `ifCredentialValid`.

worker/undertaker/manifold.go
babbageclunk: #10245 changed the worker to log the error and continue if `destroy-model --force` is run.

If we don't return an error here then destroy-model will work without force when the credential is invalid, which we definitely don't want - that could lead to people not realising they've got instances still running in the cloud. As long as we make them specify `--force` then they (should be) aware that things might be left.

babbageclunk:
!!build!!

babbageclunk:
$$merge$$

babbageclunk:
Timed out getting GPG key?
$$merge$$

babbageclunk:
Failed to start mongo. I'm getting all the weirdos today!
$$merge$$

PR #10284, opened 2019-06-04
Local credentials tolerate per-cloud malformed credentials.

hpidcock:
## Description of change

Malformed credential.yaml files produce unhelpful errors and prevent the user
from moving forward despite the command not requiring cloud credential that is malformed.

For example `juju bootstrap --clouds`, `juju remove-credential`, `juju list-credentials` etc that depend on a cloud credential from the local credential store will fail regardless of if they need the credential.

The change allows the local file credential store via CredentialForCloud to access credentials without requiring the whole credentials.yaml file to be validate. Instead validation is performed lazily when the
particular credential is accessed.

Both `juju bootstrap --clouds` and `juju list-credentials` treat individual cloud credential entries independently and warn instead of returning an error.

## QA steps

Alter an entry in credentials.yaml and change the field name `auth-type` to `auth_type` to force validation to fail.

Run `juju bootstrap --clouds` and it should WARN with the credential error, but should still list available clouds to bootstrap.

Run `juju list-credentials` and it should WARN with the credential error, but should still list available clouds and their information (except the invalid ones).

Run `juju bootstrap localhost` and it should deploy successfully and mutate the `credentials.yaml` file leaving the malformed entry.

## Documentation changes

Credentials that don't map to a cloud name no longer list in `juju list-credentials`

## Bug reference

[https://bugs.launchpad.net/juju/+bug/1616523](https://bugs.launchpad.net/juju/+bug/1616523)

### DIFF
--- a/cloud/credentials.go
+++ b/cloud/credentials.go
@@ -464,7 +464,7 @@ func ParseCredentialCollection(data []byte) (*CredentialCollection, error) {
 	return &collection, nil
 }
 
-// CloudCredential returns a copy of the CloudCredential for the specifed cloud or
+// CloudCredential returns a copy of the CloudCredential for the specified cloud or
 // an error when the CloudCredential was not found or failed to pass validation.
 func (c *CredentialCollection) CloudCredential(cloudName string) (*CloudCredential, error) {
 	credentialValue, ok := c.Credentials[cloudName]
--- a/cloud/credentials.go
+++ b/cloud/credentials.go
@@ -400,14 +400,14 @@ func (c cloudCredentialValueChecker) Coerce(v interface{}, path []string) (inter
 // ParseCredentials parses the given yaml bytes into Credentials, but does
 // not validate the credential attributes.
 func ParseCredentials(data []byte) (map[string]CloudCredential, error) {
-	credentialContainer, err := ParseCredentialContainer(data)
+	credentialCollection, err := ParseCredentialCollection(data)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	cloudNames := credentialContainer.CloudNames()
+	cloudNames := credentialCollection.CloudNames()
 	credentials := make(map[string]CloudCredential)
 	for _, cloud := range cloudNames {
-		v, err := credentialContainer.CloudCredential(cloud)
+		v, err := credentialCollection.CloudCredential(cloud)
 		if err != nil {
 			return nil, errors.Trace(err)
 		}
@@ -434,26 +434,26 @@ func RemoveSecrets(
 	return &Credential{authType: credential.authType, attributes: redactedAttrs}, nil
 }
 
-// CredentialContainer holds CloudCredential(s) that are lazily validated.
-type CredentialContainer struct {
+// CredentialCollection holds CloudCredential(s) that are lazily validated.
+type CredentialCollection struct {
 	Credentials map[string]interface{} `yaml:"credentials"`
 }
 
-// ParseCredentialContainer parses YAML bytes for the credential
-func ParseCredentialContainer(data []byte) (*CredentialContainer, error) {
-	container := CredentialContainer{}
-	err := yaml.Unmarshal(data, &container)
+// ParseCredentialCollection parses YAML bytes for the credential
+func ParseCredentialCollection(data []byte) (*CredentialCollection, error) {
+	collection := CredentialCollection{}
+	err := yaml.Unmarshal(data, &collection)
 	if err != nil {
 		return nil, errors.Annotate(err, "cannot unmarshal yaml credentials")
 	}
-	return &container, nil
+	return &collection, nil
 }
 
 // CloudCredential returns a copy of the CloudCredential for the specifed cloud or
 // an error when the CloudCredential was not found or failed to pass validation.
-func (c *CredentialContainer) CloudCredential(cloudName string) (*CloudCredential, error) {
+func (c *CredentialCollection) CloudCredential(cloudName string) (*CloudCredential, error) {
 	credentialValue, ok := c.Credentials[cloudName]
-	if ok == false {
+	if !ok {
 		return nil, errors.NotFoundf("credentials for cloud %s", cloudName)
 	}
 	if credential, ok := credentialValue.(CloudCredential); ok {
@@ -483,8 +483,8 @@ func (c *CredentialContainer) CloudCredential(cloudName string) (*CloudCredentia
 	return &credential, nil
 }
 
-// CloudNames returns all the names of CloudCredentials inside the CredentialContainer.
-func (c *CredentialContainer) CloudNames() []string {
+// CloudNames returns all the names of CloudCredentials inside the CredentialCollection.
+func (c *CredentialCollection) CloudNames() []string {
 	var cloudNames []string
 	for k := range c.Credentials {
 		cloudNames = append(cloudNames, k)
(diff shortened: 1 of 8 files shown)


hpidcock:
[one badly formed credential makes bootstrap fail](https://trello.com/c/f83Iqd78/794-one-badly-formed-credential-makes-bootstrap-fail)

cloud/credentials.go
```diff
@@ -437,3 +433,86 @@ func RemoveSecrets(
...
+		return nil, errors.Annotate(err, "cannot unmarshal yaml credentials")
+	}
+	return &container, nil
+}
+
+// CloudCredential returns a copy of the CloudCredential for the specifed cloud or
+// an error when the CloudCredential was not found or failed to pass validation.
+func (c *CredentialContainer) CloudCredential(cloudName string) (*CloudCredential, error) {
+	credentialValue, ok := c.Credentials[cloudName]
+	if ok == false {
```
anastasiamac: Could be !ok..

cmd/juju/commands/bootstrap_clouds.go
```diff
@@ -45,7 +45,9 @@ func formatCloudDetailsTabular(clouds cloudList, credStore jujuclient.Credential
 		for _, name := range cloudNames {
 			cred, err := credStore.CredentialForCloud(name)
 			if err != nil && !errors.IsNotFound(err) {
-				return errors.Annotatef(err, "error loading credential for cloud %q", name)
+				ctx.Warningf("error loading credential for cloud %v: %v", name, err)
+				p(name, "ERROR", "")
```
anastasiamac: I am no sure that we need to have both the error printed as ctx.Warningf as well as a record in the tabular output. ctx.Warningf  is enough I think...

It might be worthwhile adding tests where several credentials will have troubles loading in different order to make sure the output is as expected.

cmd/juju/cloud/listcredentials.go
```diff
@@ -156,35 +156,64 @@ func (c *listCredentialsCommand) personalClouds() (map[string]jujucloud.Cloud, e
...
+	return builtin, nil
+}
+
+func (c *listCredentialsCommand) Run(ctxt *cmd.Context) error {
+	var cloudNames []string
+
+	if c.cloudName != "" {
+		cloudNames = []string{c.cloudName}
+	} else {
+		// Find cloud names.
```
anastasiamac: I am not sure how I feel about this whole long stanza here but this area will be changed shortly to also list credential from the controller so maybe it is ok.

cloud/credentials.go
```diff
@@ -437,3 +433,86 @@ func RemoveSecrets(
...
+			if name == details.DefaultCredential {
+				stillHaveDefault = true
+				break
+			}
+		}
+		if !stillHaveDefault {
+			details.DefaultCredential = ""
+		}
+	}
+	if len(details.AuthCredentials) == 0 {
```
wallyworld: We could do this len == 0 check first up and exit early

cmd/juju/cloud/listcredentials.go
wallyworld: Consider breaking out the gathering of cloud names to a helper func
Then no real need to bother with the older "one line" helper funcs for builtinClouds(), publicClouds() etc

cmd/juju/commands/bootstrap_clouds.go
wallyworld: I think INVALID might be better
I'm not sure how logging message will interact with printing the tabular output. You normally see log messages when --show-log is used.
We need new unit tests for tabular (and yaml/json) output of list-credentials or bootstrap --cloud etc etc when there's a bad credential.

hpidcock: I guess in this case it makes sense not to list the failed cloud, because then it isn't really "available" to bootstrap to.

cmd/juju/cloud/listcredentials.go
hpidcock: Yep, I regretted it immediately. My alternate was to alter the CredentialGetter interface to return the names for all the entries, was trying to avoid changing that interface as it then requires the memory store and file store to be updated + unit tests for the additional function. Possibly would be better to add a method to aggregate all the available clouds (personal, public and builtin), as this is almost verbatim to what is in the boostrap --clouds command.

cloud/credentials.go
hpidcock: Force of habit :)

cloud/credentials.go
```diff
@@ -437,3 +433,86 @@ func RemoveSecrets(
...
+// UpdateCloudCredential stores CloudCredential for a specific cloud.
+func (c *CredentialCollection) UpdateCloudCredential(cloudName string, details CloudCredential) {
+	if len(details.AuthCredentials) == 0 {
+		delete(c.Credentials, cloudName)
+		return
+	}
+	if c.Credentials == nil {
+		c.Credentials = make(map[string]interface{})
+	}
+	// Validate default credential actually exists in the CloudCredential
```
wallyworld: This block of code could be broken out as a helper as it's also used in CloudCredential(cloudName string)

cmd/juju/cloud/listcredentials.go
```diff
@@ -156,35 +156,54 @@ func (c *listCredentialsCommand) personalClouds() (map[string]jujucloud.Cloud, e
...
 
 	displayCredentials := make(map[string]CloudCredential)
 	var missingClouds []string
-	for cloudName, cred := range credentials {
+	for _, cloudName := range cloudNames {
+		cred, err := c.store.CredentialForCloud(cloudName)
+		if errors.IsNotFound(err) {
+			continue
+		} else if err != nil {
+			ctxt.Infof("error loading credential for cloud %v: %v", cloudName, err)
```
wallyworld: Should probably be a Warning as it is an issue the user should look at fixing

cmd/juju/cloud/listcredentials_test.go
```diff
@@ -6,6 +6,9 @@ package cloud_test
 import (
 	"strings"
 
+	"github.com/juju/juju/jujuclient/jujuclienttesting"
```
wallyworld: this import belongs below with the other juju/juju imports

cmd/juju/cloud/listcredentials_test.go
```diff
@@ -123,6 +127,27 @@ mycloud  me
...
+	}
+	ctx := s.listCredentialsWithStore(c, store)
+	c.Assert(cmdtesting.Stdout(ctx), gc.Equals, `
+Cloud   Credentials
+aws     down*, bob
+azure   azhja
+google  default
+
+`[1:])
+	c.Assert(cmdtesting.Stderr(ctx), gc.Equals, `
```
wallyworld: This implies the return code with be > 0 right?
Not sure that we want to have a non-success error code when the command ran successfully. We could add text to the output after the main tabular listing to print any bad credentials but maybe that's overkill.

cmd/juju/commands/bootstrap_clouds.go
```diff
@@ -45,7 +45,8 @@ func formatCloudDetailsTabular(clouds cloudList, credStore jujuclient.Credential
 		for _, name := range cloudNames {
 			cred, err := credStore.CredentialForCloud(name)
 			if err != nil && !errors.IsNotFound(err) {
-				return errors.Annotatef(err, "error loading credential for cloud %q", name)
+				ctx.Infof("error loading credential for cloud %v: %v", name, err)
```
wallyworld: Warning?

cloud/credentials.go
```diff
@@ -437,3 +433,86 @@ func RemoveSecrets(
...
+		}
+		if !stillHaveDefault {
+			credential.DefaultCredential = ""
+		}
+	}
+	c.Credentials[cloudName] = credential
+	return &credential, nil
+}
+
+// CloudNames returns all the names of CloudCredentials inside the CredentialCollection.
```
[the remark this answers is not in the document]
hpidcock: Sorry, I think by CloudCredential I mean the struct CloudCredential which itself contains a map of Credential.

hpidcock:
$$merge$$

PR #10285, opened 2019-06-04
Support trust flag when deploying bundles

achilleasa:
## Description of change

This PR enables operators to deploy a bundle containing one or more applications that require trusted access to cloud credentials while at the same time granting them access to said credentials. This is achieved by passing the `--trust` flag to `juju deploy` calls. 

NOTE: the deploy command already accepted a `--trust` flag but prior to this PR this was only valid for deploying individual charms.

When a bundle is deployed, juju calls out to the [bundlechanges](https://github.com/juju/bundlechanges) package which basically scans the bundle data and returns back a list of changes that we need to apply in order to deploy the bundle. Each change type is modeled using a different Go type. The one we are interested in is called `AddApplicationChange` and as you can probably guess by the name, it requests the creation of a new application. The orchestration and execution of any changes that are returned by the bundlechanges code is handled by the `bundleHandler` type which lives in `cmd/juju/application/bundle.go`.

As the indicator of whether an application is trusted or not is nothing more than a simple configuration setting (`juju trust` manipulates its value through a facade) the solution implemented by this PR patches the `bundleHandler` code to inject the trust-related configuration option (if the operator provided `--trust`) to the application before calling the `Deploy` API.

## QA steps

```console 

$ export JUJU_DEV_FEATURE_FLAGS=trusted-bundles

# Force-deploy a bundle with a single app requiring trust
$ juju deploy testcharms/charm-repo/bundle/aws-integrator-trust-single/bundle.yaml --force

$ juju config aws-integrator | grep -A 6 application-config
application-config:
  trust:
    default: false
    description: Does this application have access to trusted credentials
    source: user
    type: bool
    value: false    <- you need to run "juju trust aws-integrator"

# Clean up
$ juju remove-application aws-integrator --force
$ juju remove-application ubuntu-lite --force

# IMPORTANT: wait till everything gets torn down before running the next step!

$ juju deploy testcharms/charm-repo/bundle/aws-integrator-trust-single/bundle.yaml --trust

$ juju config aws-integrator | grep -A 6 application-config
application-config:
  trust:
    default: false
    description: Does this application have access to trusted credentials
    source: user
    type: bool
    value: true    <- all set!
```

## Documentation changes

@pmatulis This would be a good time to update the docs for `deploy --trust` so we are ready when its time to remove the feature flag

### DIFF
--- a/cmd/juju/application/bundle.go
+++ b/cmd/juju/application/bundle.go
@@ -106,44 +106,41 @@ func verifyBundle(data *charm.BundleData, bundleDir string) error {
 	return errors.Trace(verifyError)
 }
 
+type bundleDeploySpec struct {
+	ctx *cmd.Context
+
+	dryRun bool
+	force  bool
+	trust  bool
+
+	bundleData        *charm.BundleData
+	bundleDir         string
+	bundleURL         *charm.URL
+	bundleOverlayFile []string
+	channel           csparams.Channel
+
+	apiRoot DeployAPI
+
+	useExistingMachines bool
+	bundleMachines      map[string]string
+	bundleStorage       map[string]map[string]storage.Constraints
+	bundleDevices       map[string]map[string]devices.Constraints
+}
+
 // deployBundle deploys the given bundle data using the given API client and
 // charm store client. The deployment is not transactional, and its progress is
 // notified using the given deployment logger.
-func deployBundle(
-	bundleDir string,
-	data *charm.BundleData,
-	bundleURL *charm.URL,
-	bundleOverlayFile []string,
-	channel csparams.Channel,
-	apiRoot DeployAPI,
-	ctx *cmd.Context,
-	bundleStorage map[string]map[string]storage.Constraints,
-	bundleDevices map[string]map[string]devices.Constraints,
-	dryRun bool,
-	force bool,
-	trust bool,
-	useExistingMachines bool,
-	bundleMachines map[string]string,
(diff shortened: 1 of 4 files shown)


cmd/juju/application/bundle.go
```diff
@@ -29,6 +30,7 @@ import (
 
 	"github.com/juju/juju/api"
 	"github.com/juju/juju/api/application"
+	app "github.com/juju/juju/apiserver/facades/client/application"
```
achilleasa: I don't really like this alias but it's there to maintain consistency with `deploy.go` (lives in same package)

cmd/juju/application/bundle.go
```diff
@@ -119,6 +121,7 @@ func deployBundle(
 	bundleDevices map[string]map[string]devices.Constraints,
 	dryRun bool,
 	force bool,
+	trust bool,
 	useExistingMachines bool,
 	bundleMachines map[string]string,
 ) (map[*charm.URL]*macaroon.Macaroon, error) {
```
SimonRichardson: So having spoken to @jameinel recently about this, we should make this series of arguments a struct. It's becoming clear that this is overloaded.

achilleasa: I also thought about this when I was going through the code. I will push an extra commit to address this.

achilleasa:
!!build!!

(intermittent test failure: MachineLegacyLeasesSuite.TestManageModelRunsCleaner)

achilleasa:
$$merge$$

PR #10286, opened 2019-06-05
Working towards removing the clock worker.

howbazaar:
The manifold config has a clock defined. The workers can just use that rather than start a worker to hold the clock value and then have the workers have to ask it for the clock.

This work it to simplify the worker dependencies.

## QA steps

Run tests and start a controller. All workers start as before.

## Documentation changes

No user impact.


howbazaar:
$$merge$$

PR #10289, opened 2019-06-05
Expose trust key when exporting bundles

achilleasa:
## Description of change

This PR ensures that we correctly set the `trust` key for each application that the operator has explicitly trusted when exporting bundles via `juju export-bundle`. This PR is split in two parts.
 
The first part patches the server API so that we populate the `RequiresTrust` field of the `BundleData` type when exporting the bundle. It also bumps the Bundle API to V3. This change is needed to allow the client code to figure out if it can expect the trust flag to be included by the call to `ExportBundle` or whether it needs to use a fall-back approach (see below).

In order to be backwards-compatible with older controllers, the second part of the PR patches the client so that when it connects to an older controller it will first grab the exported bundle and then locally patch it to set the `trust` flag for the applications. To achieve this, the client will first unmarshal the yaml blob returned by the server and then proceed to iterate the list of applications included in the bundle. For each application, the client attempts to fetch its configuration from the server and checks whether the appropriate trust key is set. If that happens to be the case it will set the `RequiresTrust` key in the appropriate `ApplicationSpec` entry. Finally, the patched `BundleData` is marshaled back to yaml and returned back to the operator as usual.

## QA steps

```console 

$ export JUJU_DEV_FEATURE_FLAGS=trusted-bundles

# Deploy a bundle with a single app requiring trust
$ juju deploy testcharms/charm-repo/bundle/aws-integrator-trust-single/bundle.yaml --trust

$ juju export-bundle
applications:
  aws-integrator:
    charm: cs:~containers/aws-integrator-10
    num_units: 1
    to:
    - "2"
    trust: true      <- this is what you are looking for
  ubuntu-lite:
    charm: cs:~jameinel/ubuntu-lite-7
    num_units: 1
    to:
    - "3"
machines:
  "2": {}
  "3": {}
series: bionic

## Try against an older controller (2.5, 2.6 etc)
$ unset JUJU_DEV_FEATURE_FLAGS
$ juju bootstrap lxd older-controller --no-gui

# NOTE: older deploy does not support --trust
$ juju deploy testcharms/charm-repo/bundle/aws-integrator-trust-single/bundle.yaml
$ juju trust aws-integrator

# The following invocation will cause the juju client to locally patch the exported bundle
# NOTE: make sure to use the juju client from this PR
$ juju export-bundle
applications:
  aws-integrator:
    charm: cs:~containers/aws-integrator-10
    num_units: 1
    to:
    - "2"
    trust: true      <- this is what you are looking for
  ubuntu-lite:
    charm: cs:~jameinel/ubuntu-lite-7
    num_units: 1
    to:
    - "3"
machines:
  "2": {}
  "3": {}
series: bionic
```

### DIFF
--- a/api/facadeversions.go
+++ b/api/facadeversions.go
@@ -23,7 +23,7 @@ var facadeVersions = map[string]int{
 	"ApplicationScaler":            1,
 	"Backups":                      2,
 	"Block":                        2,
-	"Bundle":                       2,
+	"Bundle":                       3,
 	"CAASAgent":                    1,
 	"CAASFirewaller":               1,
 	"CAASOperator":                 1,
--- a/apiserver/allfacades.go
+++ b/apiserver/allfacades.go
@@ -158,6 +158,7 @@ func AllFacades() *facade.Registry {
 	reg("Block", 2, block.NewAPI)
 	reg("Bundle", 1, bundle.NewFacadeV1)
 	reg("Bundle", 2, bundle.NewFacadeV2)
+	reg("Bundle", 3, bundle.NewFacadeV3)
 	reg("CharmRevisionUpdater", 2, charmrevisionupdater.NewCharmRevisionUpdaterAPI)
 	reg("Charms", 2, charms.NewFacade)
 	reg("Cleaner", 2, cleaner.NewCleanerAPI)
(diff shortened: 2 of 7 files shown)


achilleasa:
$$merge$$

PR #10262, opened 2019-05-30
Display error when deploying bundles containing apps that require trust

achilleasa:
## Description of change

This PR updates the deploy command logic to emit an error when the operator attempts to deploy 
bundles that contain one or more apps that require trusted access to cloud credentials . Each application/service within a bundle can request trust by:

- definining `trust: true` at the application specification block level, OR
- by including a `trust: true` config option

The extra logic introduced by this PR is hidden behind the `trusted-bundles` feature flag as trust is still a work-in-progress feature.

NOTE: This PR only implements the error-emitting logic. A separate PR will introduce a mechanism for the operator to indicate that they indeed trust the apps requiring access to credentials.

## QA steps

```console
$ juju bootstrap lxd test --no-gui

$ export JUJU_DEV_FEATURE_FLAGS=trusted-bundles

# Deploy a bundle with a single app requiring trust
$ juju deploy testcharms/charm-repo/bundle/aws-integrator-trust-single/bundle.yaml
ERROR Bundle cannot be deployed without trusting applications with your cloud credentials.
Please repeat the deploy command with the --trust argument if you consent to trust the following application(s):
  - aws-integrator

# Deploy bundle with multiple apps requiring trust
$ juju deploy testcharms/charm-repo/bundle/aws-integrator-trust-multi/bundle.yaml
ERROR Bundle cannot be deployed without trusting applications with your cloud credentials.
Please repeat the deploy command with the --trust argument if you consent to trust the following application(s):
  - aws-integrator
  - gcp-integrator

# Finally, deploy a bundle with a single app requiring trust via a config option
$ juju deploy testcharms/charm-repo/bundle/aws-integrator-trust-conf-param/bundle.yaml
ERROR Bundle cannot be deployed without trusting applications with your cloud credentials.
Please repeat the deploy command with the --trust argument if you consent to trust the following application(s):
  - aws-integrator
```

PR #10290, opened 2019-06-05
Remove trusted-bundles feature flag

achilleasa:
## Description of change

Now that all the moving parts of the trusted bundles feature are in place and we cannot break any future release, we can safely remove the feature flag introduced by #10262.

NOTE: this PR should not be merged before #10289 lands.

## Documentation changes

@pmatulis  this change will allow us to ship this feature with the next 2.6.x release so we should ensure that any relevant docs are updated before that time.


achilleasa:
$$merge$$

PR #10292, opened 2019-06-06
Fix panic in metrics gathering.

howbazaar:
Due to missing test coverage, we didn't notice that the metrics collector would panic if there were any units. This is clearly bad.

## QA steps

Bootstrap and deploy a unit. Then ssh into the controller and run juju_metrics.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1831836


howbazaar:
$$merge$$

PR #10295, opened 2019-06-07
Start making HA/peergrouper independent of machine

wallyworld:
## Description of change

Currently, we stuff HA related information into machine entities, and store in state as part of the machine doc. For k8s, we don't have machines and so want to rework the model so that HA info is managed separately to the concept of a machine. There will ultimately be a HA node entity for each controller. This ensures correct separation of cencerns etc.

The PR simply starts out by renaming a bunch of machine related structs/interfaces/methods etc in preparation for further work.

## QA steps

bootstrap and deploy a k8s app
bootstrap and enable HA on a vm cloud

### DIFF
--- a/worker/peergrouper/desired.go
+++ b/worker/peergrouper/desired.go
@@ -17,20 +17,20 @@ import (
 	"github.com/juju/juju/network"
 )
 
-// jujuMachineKey is the key for the tag where we save a member's machine id.
-const jujuMachineKey = "juju-machine-id"
+// jujuNodeKey is the key for the tag where we save a member's node id.
+const jujuNodeKey = "juju-machine-id"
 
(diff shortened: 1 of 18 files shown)


worker/peergrouper/desired.go
```diff
@@ -17,20 +17,20 @@ import (
 	"github.com/juju/juju/network"
 )
 
-// jujuMachineKey is the key for the tag where we save a member's machine id.
-const jujuMachineKey = "juju-machine-id"
```
ycliuhw: Is it ok to change this `value` without having any compatibility issue?
It is used as the key of Tag in `mongo`.

wallyworld: Ah thanks, I didn't intend to change this yet

wallyworld:
$$merge$$

PR #10296, opened 2019-06-07
Allow OpenStack provider to use Cinder block devices as root disk

hpidcock:
## Description of change

Current behaviour of the OpenStack provider is to use local (ephemeral) block devices as the
root disk. This change adds constraints root-disk-source=volume to OpenStack to bind a new
cinder block device as the root disk. The device is created with the size passed in via root-disk constraint or uses the default root-disk size for the instance type selected.

Constraint validation is limited and therefore you can't add an instance-type constraint with root-disk-source constraint, because the mutual exclusivity of root-disk and instance-type must be maintained (for local disks).

NOTE: The cinder block device is destroyed on machine termination which follows the current behaviour of local disks on machine termination.

## QA steps

Using an OpenStack cloud:
`juju add-machine --constraints "root-disk-source=volume"`
and
`juju add-machine --constraints "root-disk-source=volume root-disk=10G"`

After the machine has started, you can validate by running `nova volume-list`
```
+--------------------------------------+--------+--------------+------+-------------+--------------------------------------+
| ID                                   | Status | Display Name | Size | Volume Type | Attached to                          |
+--------------------------------------+--------+--------------+------+-------------+--------------------------------------+
| 1ace235e-506e-4583-a7b0-77cdcaa8a82c | in-use |              | 10   | None        | ce48e628-4fd4-48a8-b699-e24d7a69c1a7 |
+--------------------------------------+--------+--------------+------+-------------+--------------------------------------+
```

## Documentation changes

Current user workflows are unchanged when root-disk-source is not set.

## Bug reference

[https://bugs.launchpad.net/juju/+bug/1746073](https://bugs.launchpad.net/juju/+bug/1746073)

### DIFF
--- a/provider/openstack/image.go
+++ b/provider/openstack/image.go
@@ -35,7 +35,7 @@ func AcceptAllFlavors(nova.FlavorDetail) bool {
 // The instance type comes from querying the flavors supported by the deployment.
 func findInstanceSpec(
 	e *Environ,
-	ic *instances.InstanceConstraint,
+	ic instances.InstanceConstraint,
 	imageMetadata []*imagemetadata.ImageMetadata,
 ) (*instances.InstanceSpec, error) {
 	// First construct all available instance types from the supported flavors.
@@ -44,6 +44,15 @@ func findInstanceSpec(
 	if err != nil {
 		return nil, err
 	}
+
+	if ic.Constraints.HasRootDiskSource() && *ic.Constraints.RootDiskSource == "volume" {
+		// When the root disk is a volume (i.e. cinder block volume)
+		// we don't want to match on RootDisk size. If an instance requires
+		// a very large root disk we don't want to select a larger instance type
+		// to fit a disk that won't be local to the instance.
+		ic.Constraints.RootDisk = nil
+	}
+
 	// Not all needed information is available in flavors,
 	// for e.g. architectures or virtualisation types.
 	// For these properties, we assume that all instance types support
@@ -62,14 +71,6 @@ func findInstanceSpec(
 			RootDisk: uint64(flavor.Disk * 1024),
 			// tags not currently supported on openstack
 		}
-		if ic.Constraints.HasRootDiskSource() &&
-			*ic.Constraints.RootDiskSource == "volume" {
-			// When the root disk is a volume (i.e. cinder block volume)
-			// we don't want to match on RootDisk size. If an instance requires
-			// a very large root disk we don't want to select a larger instance type
-			// to fit a disk that won't be local to the instance.
-			instanceType.RootDisk = 0
-		}
 		if ic.Constraints.HasVirtType() {
 			// Instance Type virtual type depends on the virtual type of the selected image, i.e.
 			// picking an image with a virt type gives a machine with this virt type.
@@ -79,7 +80,7 @@ func findInstanceSpec(
 	}
 
 	images := instances.ImageMetadataToImages(imageMetadata)
-	spec, err := instances.FindInstanceSpec(images, ic, allInstanceTypes)
+	spec, err := instances.FindInstanceSpec(images, &ic, allInstanceTypes)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


provider/openstack/image.go
```diff
@@ -44,6 +44,15 @@ func findInstanceSpec(
 	if err != nil {
 		return nil, err
 	}
+
+	if ic.Constraints.HasRootDiskSource() && *ic.Constraints.RootDiskSource == "volume" {
```
wallyworld: let's use a const for "volume" as we use it here and a few other places

provider/openstack/image.go
```diff
@@ -44,6 +44,15 @@ func findInstanceSpec(
...
+		// When the root disk is a volume (i.e. cinder block volume)
+		// we don't want to match on RootDisk size. If an instance requires
+		// a very large root disk we don't want to select a larger instance type
+		// to fit a disk that won't be local to the instance.
+		ic.Constraints.RootDisk = nil
```
wallyworld: Instead of just overwriting a potential user specified root disk size we need to use constraint validation to explicitly disallow root-disk-source="volume" and root-disk != nil for openstack
See InstancePrechecker inferface

provider/openstack/provider.go
```diff
@@ -1338,6 +1345,40 @@ func (e *Environ) StartInstance(ctx context.ProviderCallContext, args environs.S
...
+		// block storage service.
+		DestinationType:     rootDiskSource,
+		DeleteOnTermination: true,
+	}
+	switch rootDiskSource {
+	case "local":
+		runOpts.ImageId = spec.Image.Id
+	case "volume":
+		size := spec.InstanceType.RootDisk
+		if args.Constraints.HasRootDisk() {
```
wallyworld: don't we set root disk constraint to nil if root disk source == "volume" elsewhere?

hpidcock: RootDisk is only set to nil in the findInstanceSpec to prevent matching instance flavors on RootDisk, due to RootDisk having a different meaning when RootDiskSource=volume. Otherwise when someone specifies RootDiskSource=volume and RootDisk=1TB, we don't want to match a instance type that has 1TB of local storage, just we want the root disk cinder volume to have 1TB.

provider/openstack/image.go
hpidcock: RootDisk is only set to nil for findInstanceSpec and only affects the matching of an instanceSpec based on constraints. In the case of RootDiskSource=volume, we want to change the meaning of RootDisk to the size of the cinder volume not the size of the local storage on the instance. Possibly we need to rethink this. I'll have a look at InstancePrechecker and how I can use it.

provider/openstack/local_test.go
```diff
@@ -1434,6 +1434,13 @@ func (s *localServerSuite) TestPrecheckInstanceInvalidInstanceType(c *gc.C) {
 	c.Assert(err, gc.ErrorMatches, `invalid Openstack flavour "m1.large" specified`)
 }
 
+func (s *localServerSuite) TestPrecheckInstanceInvalidRootDiskConstraint(c *gc.C) {
+	env := s.Open(c, s.env.Config())
+	cons := constraints.MustParse("instance-type=m1.small root-disk=10G")
+	err := env.PrecheckInstance(s.callCtx, environs.PrecheckInstanceParams{Series: supportedversion.SupportedLTS(), Constraints: cons})
+	c.Assert(err, gc.ErrorMatches, `constraint root-disk cannot be specified with instance-type unless when constraint root-disk-source=volume`)
```
wallyworld: s/when//

provider/openstack/local_test.go
```diff
@@ -2384,6 +2391,34 @@ func (t *localServerSuite) TestStartInstanceVolumeAttachmentsAvailZoneConflictsP
...
+	c.Assert(err, jc.ErrorIsNil)
+
+	cons, err := constraints.Parse("root-disk-source=volume")
+	c.Assert(err, jc.ErrorIsNil)
+
+	_, err = testing.StartInstanceWithParams(t.env, t.callCtx, "1", environs.StartInstanceParams{
+		ControllerUUID: t.ControllerUUID,
+		Constraints:    cons,
+	})
+	c.Assert(err, jc.ErrorIsNil)
```
wallyworld: Here and in test that follows we need to check that the root disk params passed in are correctly used.

hpidcock:
$$merge$$

hpidcock:
$$merge$$

PR #10297, opened 2019-06-07
Bundle trust acceptance test

achilleasa:
## Description of change

The test deploys (using --trust) a bundle that contains an application requiring trust and verifies that the application will reach an active state. 

The application uses the `trust-checker` charm : a dummy charm that transitions from blocked to active (and vice-versa) depending on whether it can access cloud credentials.


achilleasa:
$$merge$$

PR #10305, opened 2019-06-10
Describe API

SimonRichardson:
## Description of change

In order to provide a clean workflow for python-libjuju (pylibjuju), we 
need to find a way to create a schema (JSON) for pylibjuju so that
when generating the new set of client definitions in pylibjuju we're not
required to jump through hoops at updating and change schemagen.

Schemagen tags juju at a given hash, which means recording all the 
correct dependencies from juju along with pylibjuju as well. This isn't
trivial as we also need to work with different branches and can become
confusing when explain the setup.

Instead the work flow will be, juju will always contain a schema.json
and then when making pylibjuju we can just consume it.

## QA steps

```console
make rebuild-schema
```

Then run to verify:
```console
./scripts/verify.sh
```


SimonRichardson:
!!build!!

SimonRichardson:
https://trello.com/c/kKYz6ih1/1487-move-schemagen-into-juju-so-we-can-use-the-facades-schema-directly

generate/schemagen/schemagen.go
```diff
@@ -0,0 +1,39 @@
...
+		os.Exit(1)
+	}
+
+	jsonSchema, err := json.MarshalIndent(result, "", "    ")
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	err = ioutil.WriteFile(os.Args[1], jsonSchema, 0644)
```
jameinel: can we do a little bit to handle if someone doesn't supply enough args so they don't get a panic/out-of-index error, but instead a hint as to how it should be used?

scripts/schema.bash
```diff
@@ -0,0 +1,13 @@
...
+function schema_sha {
+    echo `cat ./apiserver/facades/schema.json | sha256sum | awk '{ print $1 }'`
+}
+
+current_schema_sha=$(schema_sha)
+make rebuild-schema
+new_schema_sha=$(schema_sha)
+if [ $current_schema_sha != $new_schema_sha ]; then
+    (>&2 echo "Error: facades schema is not in sync. Run `make rebuild-schema` and commit source.")
+fi
```
jameinel: missing newline at end of file.
Also, I'm not super happy with checking the content on disk before you check the content on disk. (consider verify.bash being used pre-push). Doesn't it need to verify against the value of the last commit?
I think you could use "git show HEAD -- apiserver/facades/schema.json" or something like that to get the last committed version. Thoughts?
I would have thought there was a way to just get the sha from git but I don't see it (AIUI git commit hashes have more than just content.)

SimonRichardson: The check in how I envisioned it was, "it looks like you're attempting push the schema.json which is out of line with what the go code states". The problem is that if we use the last commit of git to test against, what are we trying to determine; if you changed a file, what was the reason, did you change the API server facades? The only true way to really know if something changed is to ask the facades themselves...

We could do something along the lines of the golinters and ignore them for local pre-push (unless you set the argument) and enable them for CI, that way you don't have to take the hit of rebuild the schema for every push.

SimonRichardson: > Doesn't it need to verify against the value of the last commit?

So the schema.json will be the last commit if it's done on in pre-push, but I can see that we could be sure and use `git show HEAD:./apiserver/facades/schema.json`, as we then know that what you're pushing is in fact what we need to check.

SimonRichardson: I've address the issues in https://github.com/juju/juju/pull/10305/commits/0328e7276a101144a77b9b0c92bbac94109b211a

Makefile
```diff
@@ -116,9 +116,12 @@ rebuild-dependencies:
 
 rebuild-schema:
 	@echo "Generating facade schema..."
+ifdef SCHEMA_PATH
+	@go run ./generate/schemagen/schemagen.go "$(SCHEMA_PATH)"
+else
```
SimonRichardson: Overload rebuild-schema target, so we reference the schemagen in one place.

SimonRichardson:
#10336 supersedes this PR, so to target 2.6 rather than develop (2.7)

PR #10306, opened 2019-06-10
leases: Fix leadership flapping for multiple models with the same application

babbageclunk:
## Description of change

The bound lease reader wasn't filtering leases by namespace and model before adding them to the map of lease -> holder, so if you have two models with the same application but different units are the leader
in each, status would show the leadership flapping for them (although the actual leadership wasn't changing).

Filtering by the namespace and model passed in fixes the flapping, but a better fix is to use the changed structure of the FSM (to group leases by namespace and model) to get the leaders directly. That will follow in a later PR.

## QA steps

* Bootstrap a controller and add two models.
* Deploy the same application to each model with multiple units.
```
juju bootstrap localhost E --build-agent --verbose --model-default='logging-config="<root>=DEBUG"' --debug  && juju add-model m1 && juju deploy -n4 ~jameinel/ubuntu-lite && juju add-model m2 && juju deploy -n4 ~jameinel/ubuntu-lite
```
* Since unit 0 will very likely be the leader in each model, remove unit 0 from one and wait for the lease to expire.
```
```
* Watching status for each model should show distinct consistent leaders in each model.

## Documentation changes
None

## Bug reference
None

### DIFF
--- a/worker/lease/manager.go
+++ b/worker/lease/manager.go
@@ -627,7 +627,9 @@ func (manager *Manager) pinned(namespace, modelUUID string) map[string][]string 
 func (manager *Manager) leases(namespace, modelUUID string) map[string]string {
 	leases := make(map[string]string)
 	for key, lease := range manager.config.Store.Leases() {
-		leases[key.Lease] = lease.Holder
+		if key.Namespace == namespace && key.ModelUUID == modelUUID {
+			leases[key.Lease] = lease.Holder
+		}
 	}
 	return leases
 }
--- a/acceptancetests/repository/charms/mysql/README.md
+++ b/acceptancetests/repository/charms/mysql/README.md
@@ -12,7 +12,7 @@ To deploy a MySQL service:
 
     juju deploy mysql
 
-Once deployed, you can retrive the MySQL root user password by logging in to the machine via `juju ssh` and readin the `/var/lib/mysql/mysql.passwd` file. To log in as root MySQL User at the MySQL console you can issue the following:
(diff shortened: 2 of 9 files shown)


worker/lease/manager.go
```diff
@@ -627,7 +627,9 @@ func (manager *Manager) pinned(namespace, modelUUID string) map[string][]string
 func (manager *Manager) leases(namespace, modelUUID string) map[string]string {
 	leases := make(map[string]string)
 	for key, lease := range manager.config.Store.Leases() {
-		leases[key.Lease] = lease.Holder
+		if key.Namespace == namespace && key.ModelUUID == modelUUID {
+			leases[key.Lease] = lease.Holder
+		}
 	}
 	return leases
```
timClicks: Do you want to return an `ErrNotFound` if it's empty?

babbageclunk: No, ErrNotFound would suggest that something's gone wrong (you've passed a bad ID or something), but it would be entirely normal for the leadership leases map to be empty for a model before you've deployed anything.

howbazaar: An empty map is sufficient.

babbageclunk:
$$merge$$

babbageclunk:
Driveby fix for typos in messages.

$$merge$$

babbageclunk:
$$merge$$

Failure in caas/kubernetes/provider:bootstrapSuite.TestBootstrap - I can't reproduce this locally.

babbageclunk:
spurious failure in worker/instancemutater:workerEnvironSuite.TestCharmProfilingInfoError

$$merge$$

babbageclunk:
[lease flapping for same application across models](https://trello.com/c/UgOEbquo/802-lease-flapping-for-same-application-across-models)

PR #10315, opened 2019-06-12
Bring in worker backoff fix.

howbazaar:
This brings in the dependency engine fix that among other things fixes the log spew when a machine runs out of disk space. When a machine is out of disk the metrics-sender worker fails to start because it can't create a directory. After 120 failures, the backoff value that was cast to a time.Duration becomes negative and caues the worker to restart immediately. This results in hundreds of thousands of logs per minute.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1827664


howbazaar:
$$merge$$

PR #10316, opened 2019-06-12
raft leases: use fast path to get leadership for a model

babbageclunk:
## Description of change

The raft lease FSM stores leases grouped by namespace+model uuid so there's no need to get all leases and then filter them by model. This can mean a lot more work when there are a lot of models. Add a lease.Store.LeaseGroup interface method to get a group of leases, and add it to all of the store implementations.

## QA steps

* Bootstrap a controller, add two models and deploy an application with multiple units in each model.
```
juju bootstrap localhost E --build-agent --verbose --model-default='logging-config="<root>=DEBUG"' --model-default="enable-os-upgrade=false" --debug  && juju add-model m1 && juju deploy -n4 ~jameinel/ubuntu-lite && juju add-model m2 && juju deploy -n4 ~jameinel/ubuntu-lite
```
* Once all the units are ready remove unit 0 in one of the models (to see the leadership change).
```
```
* Status in each model shows different leaders.
* Removing the leader shows leadership being claimed by another unit once the lease expires.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/core/raftlease/fsm.go
+++ b/core/raftlease/fsm.go
@@ -583,9 +583,6 @@ func (c *Command) Validate() error {
 			return errors.NotValidf("setTime with model UUID")
 		}
 		if c.Lease != "" {
-			if c.Holder == "" {
-				return errors.NotValidf("%s with empty holder", c.Operation)
-			}
 			return errors.NotValidf("setTime with lease")
 		}
 		if c.PinEntity != "" {
--- a/core/raftlease/fsm.go
+++ b/core/raftlease/fsm.go
@@ -258,6 +258,25 @@ func (f *FSM) infoFromEntry(localTime time.Time, key lease.Key, entry *entry) le
 	}
 }
 
+// LeaseGroup returns all leases matching the namespace and model -
+// when there are many models this is more efficient than getting all
+// the leases and filtering by model.
+func (f *FSM) LeaseGroup(getLocalTime func() time.Time, namespace, modelUUID string) map[lease.Key]lease.Info {
+	f.mu.Lock()
+	defer f.mu.Unlock()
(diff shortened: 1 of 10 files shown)


core/raftlease/fsm.go
```diff
@@ -583,9 +602,6 @@ func (c *Command) Validate() error {
 			return errors.NotValidf("setTime with model UUID")
 		}
 		if c.Lease != "" {
-			if c.Holder == "" {
```
howbazaar: Why did you remove this?

core/raftlease/store_test.go
```diff
@@ -356,6 +356,42 @@ func (s *storeSuite) TestLeasesFilter(c *gc.C) {
 	c.Check(s.fsm.Calls()[0].Args[1], jc.SameContents, []lease.Key{lease1, lease2})
 }
 
+func (s *storeSuite) TestLeaseGroup(c *gc.C) {
+	in5Seconds := s.clock.Now().Add(5 * time.Second)
```
howbazaar: Is the clock a fake clock here?

core/raftlease/fsm.go
babbageclunk: I'm pretty sure it's a rogue cut'n'pasto - it doesn't make any sense to have a holder on a setTime. I've checked with @manadart but haven't heard back yet.

core/raftlease/store_test.go
babbageclunk: Yup!

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #10318, opened 2019-06-12
raft leases: use fast path to get leadership for a model (2.6 forward merge)

babbageclunk:
## Description of change

The raft lease FSM stores leases grouped by namespace+model uuid so there's no need to get all leases and then filter them by model. This can mean a lot more work when there are a lot of models. Add a lease.Store.LeaseGroup interface method to get a group of leases, and add it to all of the store implementations.

This is a trivial forward-merge of #10316 from the 2.5 branch.

## QA steps

* Bootstrap a controller, add two models and deploy an application with multiple units in each model.
```
juju bootstrap localhost E --build-agent --verbose --model-default='logging-config="<root>=DEBUG"' --model-default="enable-os-upgrade=false" --debug  && juju add-model m1 && juju deploy -n4 ~jameinel/ubuntu-lite && juju add-model m2 && juju deploy -n4 ~jameinel/ubuntu-lite
```
* Once all the units are ready remove unit 0 in one of the models (to see the leadership change).
```
```
* Status in each model shows different leaders.
* Removing the leader shows leadership being claimed by another unit once the lease expires.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/core/lease/store.go
+++ b/core/lease/store.go
@@ -37,6 +37,13 @@ type Store interface {
 	// Supplying any lease keys will filter the return for those requested.
 	Leases(keys ...Key) map[Key]Info
 
+	// LeaseGroup returns a snapshot of all of the leases for a
+	// particular namespace/model combination. This is useful for
+	// reporting holder information for a model, and can often be
+	// implemented more efficiently than getting all leases when there
+	// are many models.
+	LeaseGroup(namespace, modelUUID string) map[Key]Info
+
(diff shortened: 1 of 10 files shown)


babbageclunk:
This doesn't require a review.
$$merge$$

babbageclunk:
Starting mongo failed in github.com/juju/juju/worker/certupdater tests.
$$merge$$

PR #10329, opened 2019-06-14
Model controller HA

wallyworld:
## Description of change

Commit 1 - state/model changes to introduce controller node entities
Commit 2 - upgrade steps to create controller nodes from existing machines

We create new controller node docs in the controllers collection; these represent machines in the mongo replicaset and their voting status. The idea is that we will remove "novote" and "hasvote" from the machine doc, and all associated HA watching and status etc will be drive off the controller nodes. This properly separates concerns and allows k8s deployments to avoid having to create machine docs.

This PR is the first of several - next steps are to move watchers across, refactor peer grouper worker, then delete the HA stuff off machine.

Several of our unit tests were broken (and so is our current HA modelling on machine). The tests set "wants vote" to true even though a machine is not a controller machine (ie it is missing the ManageMachines job). The new controller node modelling enforces correct behaviour so tests neede dto be fixed to account for that.

## QA steps

bootstrap lxd
use mongo client to check controllers collection that there's a single controller node
enable-ha
check that there's now 3 controller nodes and their contents
check juju show-controllers for correct info
remove a controller machine
repeat the above checks (2 nodes now)
add a new machine
enable-ha --to <id>
repeat the above checks (3 nodes again)

bootstrap 2.6
enable-ha
upgrade to this version
check the controller nodes have been created


wallyworld:
[add new controller HA entities to mirror machine wants-vote/has-vote](https://trello.com/c/yLFCoERl/810-add-new-controller-ha-entities-to-mirror-machine-wants-vote-has-vote)

state/watcher.go
```diff
@@ -1645,8 +1645,20 @@ func (m *Machine) WatchHardwareCharacteristics() NotifyWatcher {
 }
 
 // WatchControllerInfo returns a NotifyWatcher for the controllers collection
-func (st *State) WatchControllerInfo() NotifyWatcher {
-	return newEntityWatcher(st, controllersC, modelGlobalKey)
+func (st *State) WatchControllerInfo() StringsWatcher {
+	return newCollectionWatcher(st, colWCfg{
```
babbageclunk: What does this change do? Is it that it used to be a model-specific collection and now it isn't, and the ids yielded from the watcher should still be model-local?

state/watcher.go
```diff
@@ -2863,7 +2875,7 @@ func (st *State) watchEnqueuedActionsFilteredBy(receivers ...ActionReceiver) Str
 }
 
 // WatchControllerStatusChanges starts and returns a StringsWatcher that
-// notifies when the status of a controller machine changes.
+// notifies when the status of a controller nodes changes.
```
babbageclunk: node

state/watcher.go
babbageclunk: Oh duh - missed that it had changed from a notify watcher to a strings watcher.

state/upgrades.go
```diff
@@ -2159,3 +2159,55 @@ func UpdateK8sModelNameIndex(pool *StatePool) error {
...
+		n, err := controllers.FindId(expectedId).Count()
+		if err != nil {
+			return errors.Trace(err)
+		}
+		if n > 0 {
+			continue
+		}
+
+		doc := &controllerNodeDoc{
+			DocID:   ensureModelUUID(modelUUID, controllerNodeGlobalKey(mid)),
```
babbageclunk: Reuse `expectedId` here?

state/upgrades_test.go
```diff
@@ -3212,6 +3212,77 @@ func (s *upgradesSuite) TestUpdateK8sModelNameIndex(c *gc.C) {
...
+	})
+	c.Assert(err, jc.ErrorIsNil)
+
+	controllerSettings := bson.M{}
+	for k, v := range coretesting.FakeControllerConfig() {
+		controllerSettings[k] = v
+	}
+	expected := bsonMById{
+		{
+			"_id": "apiHostPorts", "apihostports": []interface{}{},
```
babbageclunk: Wow, I didn't realise how much random stuff there was in the controllers collection! Wouldn't it be easier to use a different collection for controller nodes, rather than jamming them into controllersC?

state/watcher.go
babbageclunk: This would be simpler if it didn't have to filter out other unrelated rows from the collection, wouldn't it?

state/machine.go
```diff
@@ -407,11 +407,20 @@ func (m *Machine) setHasVoteOps(hasVote bool) ([]txn.Op, error) {
...
+	var asserts interface{}
+	if hasVote {
+		asserts = txn.DocExists
+	}
 	ops := []txn.Op{{
 		C:      machinesC,
 		Id:     m.doc.DocID,
 		Assert: notDeadDoc,
 		Update: bson.D{{"$set", bson.D{{"hasvote", hasVote}}}},
+	}, {
```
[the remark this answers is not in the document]
wallyworld: done

state/watcher.go
wallyworld: ok

state/watcher.go
wallyworld: fixed

state/upgrades.go
wallyworld: moot now

state/upgrades_test.go
wallyworld: ok

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10331, opened 2019-06-14
Intermittent failure CharmProfilingInfo

SimonRichardson:
## Description of change

The following changes ensure that we wait until the
context.KillWithError(err) is called before attempting to check
that worker is killed. This is a logical race, where the mocks
expect that CharmProfilingInfo is called and returns an error, but
it collapses the wait groups. This in turns causes context.Kill(nil)
to be called at the same time. This leads to a race between the
two context kill methods. The tests expect KillWithError(err) to
happen before Kill(nil).

By using the shim we can make the wait group collapse AFTER
KillWithError(err) is called. This removes the logical race between
the two.

## QA steps

```
go test -v ./worker/instancemutater/... -check.v -check.f=workerEnvironSuite.TestCharmProfilingInfoError -count=1111
```

### DIFF
--- a/worker/instancemutater/export_test.go
+++ b/worker/instancemutater/export_test.go
@@ -4,9 +4,11 @@
 package instancemutater
 
 import (
+	"github.com/juju/errors"
 	"github.com/juju/juju/api/instancemutater"
 	"github.com/juju/juju/core/lxdprofile"
 	"github.com/juju/juju/environs"
+	names "gopkg.in/juju/names.v2"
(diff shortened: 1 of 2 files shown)


SimonRichardson:
https://trello.com/c/kfWlSoLa/1529-testcharmprofilinginfoerror-intermittent-test-failure

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

PR #10332, opened 2019-06-16
Added a headless service for stateful sets to expose pod DNS

hpidcock:
## Description of change

When an application is deployed via a stateful set, the stateful set will now have a headless service named <app-name>-endpoints which doesn't expose the pods as a service, but rather provides dns names to resolve the pod names within the cluster.

The stateful set then is created with a back-pointer to the headless service.

## QA steps

Bootstrap a k8s controller using an operator image built from this tree.
`juju deploy cs:~juju/mariadb-k8s-1`

The application should have a service called mariadb-k8s-endpoints
```
~$ kubectl describe svc -n controller-mycontroller mariadb-k8s-endpoints
Name:              mariadb-k8s-endpoints
Namespace:         controller-mycontroller
Labels:            juju-app=mariadb-k8s
Annotations:       <none>
Selector:          juju-app=mariadb-k8s
Type:              ClusterIP
IP:                None
Session Affinity:  None
Events:            <none>
```
You should be able to contact the pod within the cluster via `mariadb-k8s-0.mariadb-k8s-endpoints.controller-mycontroller.svc.cluster.local`
```
~$ kubectl run -i --tty busybox --image=busybox --restart=Never --rm -- ping -c 3 mariadb-k8s-0.mariadb-k8s-endpoints.controller-mycontroller.svc.cluster.local
64 bytes from 10.56.0.13: seq=1 ttl=64 time=0.123 ms
64 bytes from 10.56.0.13: seq=2 ttl=64 time=0.095 ms
--- mariadb-k8s-0.mariadb-k8s-endpoints.controller-mycontroller.svc.cluster.local ping statistics ---
3 packets transmitted, 3 packets received, 0% packet loss
round-trip min/avg/max = 0.095/0.108/0.123 ms
pod "busybox" deleted
```
We can also check it resolved to the correct PodIP
```
~$ kubectl -n controller-mycontroller get pod mariadb-k8s-0 -o custom-columns=:.status.podIP

10.56.0.13
```

## Documentation changes

Applications using the stateful set deployment strategy will always create a service called `<app-name>-endpoints` that does not expose any ports (no IP), instead the headless service provides pod names into kubedns. Names like `mariadb-k8s-0.mariadb-k8s-endpoints.controller-test.svc.cluster.local` which resolve to the pod named `mariadb-k8s-0`.

## Bug reference

[https://bugs.launchpad.net/juju/+bug/1832608](https://bugs.launchpad.net/juju/+bug/1832608)

### DIFF
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -998,6 +998,9 @@ func (k *kubernetesClient) DeleteService(appName string) (err error) {
 	if err := k.deleteStatefulSet(deploymentName); err != nil {
 		return errors.Trace(err)
 	}
+	if err := k.deleteService(headlessServiceName(deploymentName)); err != nil {
+		return errors.Trace(err)
+	}
 	if err := k.deleteDeployment(deploymentName); err != nil {
 		return errors.Trace(err)
 	}
@@ -1285,7 +1288,11 @@ func (k *kubernetesClient) EnsureService(
 
 	numPods := int32(numUnits)
 	if useStatefulSet {
-		if err := k.configureStatefulSet(appName, deploymentName, randPrefix, annotations.Copy(), unitSpec, params.PodSpec.Containers, &numPods, params.Filesystems, hasService); err != nil {
+		if err := k.configureHeadlessService(appName, deploymentName, annotations.Copy()); err != nil {
+			return errors.Annotate(err, "creating or updating headless service")
+		}
+		cleanups = append(cleanups, func() { k.deleteService(headlessServiceName(deploymentName)) })
+		if err := k.configureStatefulSet(appName, deploymentName, randPrefix, annotations.Copy(), unitSpec, params.PodSpec.Containers, &numPods, params.Filesystems); err != nil {
 			return errors.Annotate(err, "creating or updating StatefulSet")
 		}
 		cleanups = append(cleanups, func() { k.deleteDeployment(appName) })
@@ -1584,7 +1591,6 @@ func (k *kubernetesClient) deleteDeployment(name string) error {
 func (k *kubernetesClient) configureStatefulSet(
 	appName, deploymentName, randPrefix string, annotations k8sannotations.Annotation, unitSpec *unitSpec,
 	containers []caas.ContainerSpec, replicas *int32, filesystems []storage.KubernetesFilesystemParams,
-	hasService bool,
 ) error {
 	logger.Debugf("creating/updating stateful set for %s", appName)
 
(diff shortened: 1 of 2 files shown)


hpidcock:
[DNS for pods in statefulset](https://trello.com/c/0Uu8drCj/804-dns-for-pods-in-statefulset)

hpidcock:
Waiting on 2.6.4 to release.

jameinel:
So is the issue that you can't add ServiceName to a StatefulSet?
How does this react to existing deployments (if I have a stable deployment with 2.6.3, what happens when I upgrade to 2.6.4 and the topology wants to change?) I didn't see any sort of upgrade step for this.
It seems odd that you'd access your service as "mariadb-k8s-0.mariadb-k8s.controller-mycontroller.svc.cluster.local" unless it is a stateful set and then it becomes "mariadb-k8s-0.mariadb-k8s-endpoints.controller-mycontroller.svc.cluster.local".
Am I understanding that correctly? (charms that are stateful sets map their DNS name differently?)

hpidcock:
> So is the issue that you can't add ServiceName to a StatefulSet?
> How does this react to existing deployments (if I have a stable deployment with 2.6.3, what happens when I upgrade to 2.6.4 and the topology wants to change?) I didn't see any sort of upgrade step for this.

Deployments are not affected. Old StatefulSets won't get the functionality since it appears to require setting Spec.ServiceName on the StatefulSet, which is unfortunately immutable.
To allow this behaviour on old StatefulSets would require them to be re-created, which isn't desired.

> It seems odd that you'd access your service as "mariadb-k8s-0.mariadb-k8s.controller-mycontroller.svc.cluster.local" unless it is a stateful set and then it becomes "mariadb-k8s-0.mariadb-k8s-endpoints.controller-mycontroller.svc.cluster.local".
> Am I understanding that correctly? (charms that are stateful sets map their DNS name differently?)

`mariadb-k8s.controller-mycontroller.svc.cluster.local` is how consumers of the mariadb cluster would consume it, or just mariadb-k8s inside the namespace. The difference is when you want to connect to the specific pod mariadb-k8s-0, you can connect via the endpoint mariadb-k8s-0.mariadb-k8s-endponts(.controller-mycontroller.svc.cluster.local)

For example:
```
kubectl -n controller-test-localhost run busybox -i --tty --image=busybox --restart=Never --rm -- sh
/ # ping mariadb-k8s-0
ping: bad address 'mariadb-k8s-0'
/ # ping mariadb-k8s-0.mariadb-k8s-endpoints
PING mariadb-k8s-0.mariadb-k8s-endpoints (10.1.1.14): 56 data bytes
64 bytes from 10.1.1.14: seq=0 ttl=64 time=0.183 ms
```

The added benefits also include being able to discover the published pod endpoints, so that something like a clustered database could be seeded.
```
# dig mariadb-k8s-endpoints.controller-test-localhost.svc.cluster.local. SRV

;; QUESTION SECTION:
;mariadb-k8s-endpoints.controller-test-localhost.svc.cluster.local. IN SRV

;; ANSWER SECTION:
mariadb-k8s-endpoints.controller-test-localhost.svc.cluster.local. 30 IN SRV 10 100 0 mariadb-k8s-0.mariadb-k8s-endpoints.controller-test-localhost.svc.cluster.local.

;; ADDITIONAL SECTION:
mariadb-k8s-0.mariadb-k8s-endpoints.controller-test-localhost.svc.cluster.local. 30 IN A 10.1.1.14
```

hpidcock:
$$merge$$

PR #10335, opened 2019-06-17
CMR: leave scope on all relation units for terminated remote application

babbageclunk:
## Description of change

When an offered application with a consumer is force-removed, it would leave the consuming model data in a bad state that prevented removing the application from consuming model with `remove-saas`.

To avoid this, leave scope for all the remote application's relation units when we see it go to Terminated state.

## QA steps

* Bootstrap a controller with two models.
* In model 1 deploy mariadb and offer the db endpoint.
* In model 2 deploy mediawiki and relate it to the mariadb offer.
* Once the connection is established and settled, remove the offer with --force.
* Removing the consumed application with remove-saas (no force) should remove it from model 2.

## Documentation changes
None

## Bug reference
None

### DIFF
--- a/state/remoteapplication.go
+++ b/state/remoteapplication.go
@@ -488,6 +488,7 @@ func (s *RemoteApplication) SetStatus(info status.StatusInfo) error {
 	if !info.Status.KnownWorkloadStatus() {
 		return errors.Errorf("cannot set invalid status %q", info.Status)
 	}
+
 	return setStatus(s.st.db(), setStatusParams{
 		badge:     "remote application",
 		globalKey: s.globalKey(),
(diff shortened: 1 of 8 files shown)


state/remoteapplication.go
```diff
@@ -488,16 +488,50 @@ func (s *RemoteApplication) SetStatus(info status.StatusInfo) error {
 	if !info.Status.KnownWorkloadStatus() {
 		return errors.Errorf("cannot set invalid status %q", info.Status)
 	}
+
+	// If the remote application's been terminated then we aren't
+	// going to see any more state changes for its units - ensure that
+	// they've left scope on all of their relations.
+	var extraOps func() ([]txn.Op, error)
```
wallyworld: This I think is the wrong layer to do this in.
Ideally, there'd be a RemoteApplicationTerminated() API call the worker would invoke (which would set status to terminated and do the units thing).
As a compromise here though, to avoid a new facade call, we would move this logic up to the SetRemoteApplicationsStatus() apiserver call. And we'd introduce a ModelOperation implementation to do the status set and unit thing which would be invoked by the apiserver call. Just a few extra lines over what's in the PR now but much cleaner.

babbageclunk: Checking I understand what you mean: in SetRemoteApplicationsStatus, if the target status is Terminated we'd construct a TerminateRemoteApplicationOperation which would do the status setting and leave scope on the relation units (so they still happen in the same txn).  If the status was anything else we just call remoteApp.SetStatus as normal. Is that right?

wallyworld: yup

state/status.go
```diff
@@ -442,7 +442,7 @@ func setStatus(db Database, params setStatusParams) (err error) {
 		// If this status is not new (i.e. it is exactly the same as
 		// our last status), there is no need to update the record.
 		// Update here will only reset the 'Since' field.
-		return err
+		return nil
```
babbageclunk: Clearer to return nil here, since err hasn't been assigned to at this point.

state/remoteapplication.go
```diff
@@ -498,6 +499,65 @@ func (s *RemoteApplication) SetStatus(info status.StatusInfo) error {
 	})
 }
 
+// TerminateOperation return a ModelOperation that will terminate this
```
wallyworld: returns

state/remoteapplication.go
```diff
@@ -498,6 +499,65 @@ func (s *RemoteApplication) SetStatus(info status.StatusInfo) error {
...
+	for _, rel := range rels {
+		remoteUnits, err := rel.AllRemoteUnits(name)
+		if err != nil {
+			return nil, errors.Annotatef(err, "getting remote units for %q", name)
+		}
+		for _, ru := range remoteUnits {
+			leaveOperation := ru.LeaveScopeOperation(false)
+			// Don't pass attempt down since the relationunit is
+			// always fresh.
+			leaveOps, err := leaveOperation.Build(0)
```
wallyworld: I think it's safer to pass attempt because if the LeaveScopeOperation changes implementation, it will be robust. Passing 0 always isn't robust to such changes.

state/remoteapplication.go
babbageclunk: good spot!

state/remoteapplication.go
babbageclunk: Yeah, that makes sense - it seems wasteful since they were just loaded, but you're right that it could be doing something else with attempt.

babbageclunk:
$$merge$$

babbageclunk:
Mongod failed to start.

$$merge$$

PR #10336, opened 2019-06-17
Schema gen

SimonRichardson:
## Description of change

In order to provide a clean workflow for python-libjuju (pylibjuju), we 
need to find a way to create a schema (JSON) for pylibjuju so that
when generating the new set of client definitions in pylibjuju we're not
required to jump through hoops at updating and change schemagen.

Schemagen tags juju at a given hash, which means recording all the 
correct dependencies from juju along with pylibjuju as well. This isn't
trivial as we also need to work with different branches and can become
confusing when explain the setup.

Instead the work flow will be, juju will always contain a schema.json
and then when making pylibjuju we can just consume it.

## QA steps

```console
make rebuild-schema
```

Then run to verify:
```console
./scripts/verify.sh
```

Note: this supersedes #10305


SimonRichardson:
https://trello.com/c/kKYz6ih1/1487-move-schemagen-into-juju-so-we-can-use-the-facades-schema-directly

SimonRichardson:
$$merge$$

babbageclunk:
Awesome! Love a 38kloc PR. :D

babbageclunk:
I tried to look at the output json and github told me to sod off.

PR #10337, opened 2019-06-17
Missing exit

SimonRichardson:
## Description of change

Missing exit from schema.bash prevents the pre-push from failing.


SimonRichardson:
$$merge$$

PR #10341, opened 2019-06-18
1833155: Tie the watcher life cycle to the catacomb

SimonRichardson:
## Description of change

The machines profile watcher was missing the ability to tie itself
to the provisioner task life-cycle, which means that when a
provisioner task was killed or restarted then the watcher was still
spinning and never cleaned up, leaking as a resource.

There is another problem that @jameinel spotted in the creation of
the provisioner task, is that if the provisioner errored out on
the construction of the provisioner task, then we never killed the
other watchers and they could also be leaked. We should look at
those in the near future to ensure that any watchers that die on
construction are also cleaned up.

## QA steps

Ensure that the following test passes:

```console
go test -v ./worker/provisioner/... -check.v -check.f=ProvisionerTaskSuite.TestStartStop
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1833155

### DIFF
--- a/worker/provisioner/provisioner_task.go
+++ b/worker/provisioner/provisioner_task.go
@@ -89,12 +89,16 @@ func NewProvisionerTask(
 ) (ProvisionerTask, error) {
 	machineChanges := machineWatcher.Changes()
 	workers := []worker.Worker{machineWatcher}
+
 	var retryChanges watcher.NotifyChannel
 	if retryWatcher != nil {
 		retryChanges = retryWatcher.Changes()
 		workers = append(workers, retryWatcher)
 	}
+
 	profileChanges := profileWatcher.Changes()
+	workers = append(workers, profileWatcher)
+
 	task := &provisionerTask{
 		controllerUUID:             controllerUUID,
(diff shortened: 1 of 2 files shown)


SimonRichardson:
https://trello.com/c/ynAaAc7X/1539-1833155-gce-memory-faults

SimonRichardson:
@jameinel can you approve so I can land it?

worker/provisioner/provisioner_task.go
```diff
@@ -89,12 +89,16 @@ func NewProvisionerTask(
 ) (ProvisionerTask, error) {
 	machineChanges := machineWatcher.Changes()
```
howbazaar: Should we also be adding in the machineWatcher?

howbazaar: It seems weird that the task would take ownership of two of the three watchers passed in.

howbazaar: So sorry, am blind.

howbazaar:
$$merge$$

PR #10350, opened 2019-06-19
2.6 capped logs

howbazaar:
This work changes the logs collections to be capped collections. This will stop the logs database growing without bound when we are writing logs faster than we can delete them.

The capping and sizing thereof is done when the database is opened.

This means we can remove the log pruner. We need to keep the controller config values for max-logs-age and max-logs-size as these values are required for earlier 2.6 releases and the later 2.6 client should still be able to bootstrap earlier 2.6 versions, and without these values, the apiserver panics.
 
## QA steps
```
* bootstrap a controller with debug logging
* deploy some things into the default model
* go into the mongo shell and check the collStats of the logs databases
* change the model-logs-size controller config value
* restart the controllers
* go into the mongo shell and check the collStats of the logs databases
```

```
* bootstrap a 2.6.3 controller with debug logging
* deploy some things into the default model
* upgrade the controller
* go into the mongo shell and check the collStats of the logs databases
```

## Documentation changes

Will need to update the controller config values.


controller/config.go
```diff
@@ -694,15 +682,13 @@ func Validate(c Config) error {
...
+		if err != nil {
+			return errors.Annotate(err, "invalid model logs size in configuration")
 		}
-	}
-
-	if v, ok := c[MaxLogsSize].(string); ok {
-		if _, err := utils.ParseSize(v); err != nil {
-			return errors.Annotate(err, "invalid max logs size in configuration")
+		if mb < 1 {
+			return errors.NotValidf("model logs size must be at least 1 MB")
```
manadart:
```go
errors.NotValidf("model logs size less than 1 MB")
```
will read better when printed.

state/logs.go
```diff
@@ -74,14 +77,128 @@ func logCollectionName(modelUUID string) string {
...
+				// If the value is there we know it fits in an int without wrapping.
+				return int(size), nil
+			}
+		}
+	}
+	return controller.DefaultModelLogsSizeMB, nil
+}
+
+// modelUUIDs returns the UUIDs of all models currently stored in the database.
+// This function is called very early in the opening of the databaes, so it uses
```
manadart: "database"

state/logs.go
```diff
@@ -74,14 +77,128 @@ func logCollectionName(modelUUID string) string {
...
+		return nil, err
+	}
+	var result []string
+	for _, doc := range docs {
+		result = append(result, doc.UUID)
+	}
+	return result, nil
+}
+
 // InitDbLogs sets up the indexes for the logs collection. It should
```
manadart: Comment update for method renaming.

howbazaar:
[max db log collection size for models](https://trello.com/c/1jN13AkC/29-max-db-log-collection-size-for-models)

controller/config.go
howbazaar: updated.

state/logs.go
howbazaar: fixed

state/logs.go
howbazaar: Updated.

state/logs.go
```diff
@@ -74,14 +76,130 @@ func logCollectionName(modelUUID string) string {
...
+	var doc settingsDoc
+	err := session.DB(jujuDB).C(controllersC).FindId(controllerSettingsGlobalKey).One(&doc)
+	if err != nil {
+		return 0, errors.Trace(err)
+	}
+	// During initial migration there is no guarantee that the value exists
+	// in the settings document.
+	if value, ok := doc.Settings[controller.ModelLogsSize]; ok {
+		if s, ok := value.(string); ok {
+			size, _ := utils.ParseSize(s)
```
manadart: I guess we have already validated by this time.

howbazaar: Yep.

howbazaar:
$$merge$$

howbazaar:
Peer grouper test race $$merge$$

PR #10354, opened 2019-06-20
Fixes lp#1831527: misleading user errors.

anastasiamac:
## Description of change

Occasionally, juju does not corectly check provider 'not found" errors which get propagated and are surfacing to the users causing confusion.

This PR improves error checks in provider code. I have examined all providers and am fairly certain that I have covered all occurrences.
  
## QA steps

1. bootstrap on gce
2. deploy something sizable.. I've used ```juju deploy cs:canonical-kubernetes```
3. once things settle, run ```juju kill-controller --debug -y -t0 <your controller> --show-log``` and watch the logs...

Before this PR, you'd get a few ```ERROR juju.provider.gce.gceapi conn_instance.go:131 while removing instance "juju-71b3b8-7": googleapi: Error 404: The resource 'projects/juju-239814/zones/us-east1-b/instances/juju-71b3b8-7' was not found, notFound```
After this PR, you will not get any ERROR (if killing is successful)

## Bug reference

https://bugs.launchpad.net/juju/+bug/1831527

### DIFF
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -583,7 +583,7 @@ func (e *environ) StartInstance(ctx context.ProviderCallContext, args environs.S
 	}
 
 	switch {
-	case subnetErr != nil && (errors.IsNotFound(subnetErr) || isNotFoundError(err)):
+	case isNotFoundError(subnetErr):
 		return nil, errors.Trace(subnetErr)
 	case subnetErr != nil:
 		return nil, errors.Annotatef(maybeConvertCredentialError(subnetErr, ctx), "getting subnets for zone %q", availabilityZone)
@@ -1654,7 +1654,6 @@ func (e *environ) deleteSecurityGroupsForInstances(ctx context.ProviderCallConte
 // SecurityGroupCleaner defines provider instance methods needed to delete
 // a security group.
 type SecurityGroupCleaner interface {
-
 	// DeleteSecurityGroup deletes security group on the provider.
 	DeleteSecurityGroup(group ec2.SecurityGroup) (resp *ec2.SimpleResp, err error)
 }
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -583,7 +583,7 @@ func (e *environ) StartInstance(ctx context.ProviderCallContext, args environs.S
 	}
 
 	switch {
-	case subnetErr != nil && errors.IsNotFound(subnetErr):
+	case subnetErr != nil && (errors.IsNotFound(subnetErr) || isNotFoundError(err)):
 		return nil, errors.Trace(subnetErr)
 	case subnetErr != nil:
 		return nil, errors.Annotatef(maybeConvertCredentialError(subnetErr, ctx), "getting subnets for zone %q", availabilityZone)
--- a/provider/gce/google/conn_instance.go
+++ b/provider/gce/google/conn_instance.go
@@ -90,7 +90,7 @@ func (gce *Connection) Instances(prefix string, statuses ...string) ([]Instance,
 func (gce *Connection) removeInstance(id, zone string) error {
 	err := gce.raw.RemoveInstance(gce.projectID, zone, id)
 	if err != nil {
-		if errors.IsNotFound(err) || IsNotFound(err) {
+		if IsNotFound(err) {
 			return nil
 		}
 		// TODO(ericsnow) Try removing the firewall anyway?
@@ -100,7 +100,7 @@ func (gce *Connection) removeInstance(id, zone string) error {
 	fwname := id
 	err = gce.raw.RemoveFirewall(gce.projectID, fwname)
 	if err != nil {
-		if errors.IsNotFound(err) || IsNotFound(err) {
+		if IsNotFound(err) {
 			return nil
 		}
 		return errors.Trace(err)
--- a/provider/gce/google/conn_instance.go
+++ b/provider/gce/google/conn_instance.go
@@ -90,16 +90,19 @@ func (gce *Connection) Instances(prefix string, statuses ...string) ([]Instance,
 func (gce *Connection) removeInstance(id, zone string) error {
 	err := gce.raw.RemoveInstance(gce.projectID, zone, id)
 	if err != nil {
+		if errors.IsNotFound(err) || IsNotFound(err) {
+			return nil
+		}
 		// TODO(ericsnow) Try removing the firewall anyway?
 		return errors.Trace(err)
 	}
 
 	fwname := id
 	err = gce.raw.RemoveFirewall(gce.projectID, fwname)
-	if errors.IsNotFound(err) {
-		return nil
-	}
 	if err != nil {
+		if errors.IsNotFound(err) || IsNotFound(err) {
+			return nil
+		}
 		return errors.Trace(err)
 	}
 	return nil
--- a/provider/gce/google/errors.go
+++ b/provider/gce/google/errors.go
@@ -91,29 +91,24 @@ func HasDenialStatusCode(err error) bool {
 		return false
 	}
 
-	// http/url.Error is constructed with status code in mind and, at the time of writing for go-1.10,
-	// contains response status code and description in error.Error.
-	// We have to examine the error message to determine whether the error is related to authentication failure.
-	if cause, ok := errors.Cause(err).(*url.Error); ok {
-		for code, descs := range AuthorisationFailureStatusCodes {
-			for _, desc := range descs {
-				if strings.Contains(cause.Error(), fmt.Sprintf(": %v %v", code, desc)) {
-					return true
-				}
-			}
-		}
+	var cause error
+	switch e := errors.Cause(err).(type) {
+	case *url.Error:
+		cause = e
+	case *googleapi.Error:
+		cause = e
+	default:
+		return false
 	}
-	if cause, ok := errors.Cause(err).(*googleapi.Error); ok {
-		for code, descs := range AuthorisationFailureStatusCodes {
-			for _, desc := range descs {
-				if strings.Contains(cause.Error(), fmt.Sprintf(": %v %v", code, desc)) {
-					return true
-				}
+
+	for code, descs := range AuthorisationFailureStatusCodes {
+		for _, desc := range descs {
+			if strings.Contains(cause.Error(), fmt.Sprintf(": %v %v", code, desc)) {
+				return true
 			}
 		}
 	}
 	return false
-
 }
 
 // AuthorisationFailureStatusCodes contains http status code and
@@ -139,5 +134,5 @@ func IsNotFound(err error) bool {
 	if gerr, ok := errors.Cause(err).(*googleapi.Error); ok {
 		return gerr.Code == http.StatusNotFound
 	}
-	return errors.IsNotFound(err)
+	return errors.IsNotFound(errors.Cause(err))
 }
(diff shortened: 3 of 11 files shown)


anastasiamac:
[lp#1831527 correct provider 'not found' error check](https://trello.com/c/GA5KNAu2/822-lp1831527-correct-provider-not-found-error-check)

provider/openstack/cinder.go
```diff
@@ -595,7 +595,7 @@ func cinderToJujuVolumeInfo(volume *cinder.Volume) storage.VolumeInfo {
 
 func detachVolume(instanceId, volumeId string, storageAdapter OpenstackStorage) error {
 	err := storageAdapter.DetachVolume(instanceId, volumeId)
-	if err != nil && !errors.IsNotFound(err) {
+	if err != nil && !(errors.IsNotFound(err) || gooseerrors.IsNotFound(err)) {
```
hpidcock: This could be expressed as `err != nil && !errors.IsNotFound(err) && !gooseerrors.IsNotFound(err)`

provider/ec2/environ.go
```diff
@@ -583,7 +583,7 @@ func (e *environ) StartInstance(ctx context.ProviderCallContext, args environs.S
 	}
 
 	switch {
-	case subnetErr != nil && errors.IsNotFound(subnetErr):
+	case subnetErr != nil && (errors.IsNotFound(subnetErr) || isNotFoundError(err)):
```
babbageclunk: Shouldn't this be `isNotFoundError(subnetErr)`? Also, looking at the definition of `isNotFoundError` it already includes checking for `errors.IsNotFound`, so you should just be able to use that.

provider/gce/google/conn_instance.go
```diff
@@ -90,16 +90,19 @@ func (gce *Connection) Instances(prefix string, statuses ...string) ([]Instance,
 func (gce *Connection) removeInstance(id, zone string) error {
 	err := gce.raw.RemoveInstance(gce.projectID, zone, id)
 	if err != nil {
+		if errors.IsNotFound(err) || IsNotFound(err) {
```
babbageclunk: These would be simplified if you make your new `IsNotFound` function handle `errors.NotFound` errors as well.

provider/gce/google/errors.go
```diff
@@ -120,3 +120,11 @@ var AuthorisationFailureStatusCodes = map[int][]string{
 	// https://tools.ietf.org/html/rfc6749#section-5.2
 	http.StatusBadRequest: {"Bad Request"},
 }
+
+// IsNotFound reports whether err contains `not found'.
+func IsNotFound(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), " not found")
```
babbageclunk: This is pretty rough - is there anything more specific than this you can use to detect errors from `RemoveInstance`, `RemoveFirewall` and `GetFirewalls`?

provider/openstack/cinder.go
```diff
@@ -107,7 +107,7 @@ var newOpenstackStorage = func(env *Environ) (OpenstackStorage, error) {
 	client := env.clientUnlocked
 	if env.volumeURL == nil {
 		url, err := getVolumeEndpointURL(client, env.cloudUnlocked.Region)
-		if errors.IsNotFound(err) {
+		if errors.IsNotFound(err) || gooseerrors.IsNotFound(err) {
```
babbageclunk: Same here - can you make a function that returns true for both juju/errors NotFound and goose NotFound?

provider/gce/google/errors.go
```diff
@@ -120,3 +120,11 @@ var AuthorisationFailureStatusCodes = map[int][]string{
 	// https://tools.ietf.org/html/rfc6749#section-5.2
 	http.StatusBadRequest: {"Bad Request"},
 }
+
+// IsNotFound reports whether err contains `not found'.
+func IsNotFound(err error) bool {
```
hpidcock: From memory the best way to do this is:
```
if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == http.StatusNotFound {
}
```

anastasiamac: Yes, that covers some errors although ```errors.Cause(err).(*googleapi.Error)``` is a better.

However, google provider is notoriously inconsistent with errors and the above does not cover for example ```juju.provider.gce.gceapi raw.go:368 GCE operation error: (RESOURCE_NOT_FOUND) The resource 'projects/gothic-list-89514/zones/us-east1-b/instances/juju-f22d51-8' was not found```

I am trying to test with this + examining the gerr.Error() but so far the quickest check that covers all cases is my original one. In fact, since errors.IsNotFound is essentially checking for 'not found'  suffix as well, my original code covers that case too.

provider/gce/google/errors.go
anastasiamac: As per my reply to Harry's comment, it actually covers all cases - both scenarios that come from google and the errors.IsNotFound...

I am running  last test with a more precise implementation:
```
	if gerr, ok := errors.Cause(err).(*googleapi.Error); ok {
		return gerr.Code == http.StatusNotFound || strings.Contains(gerr.Error(), "not found")
	}
	return errors.IsNotFound(err)
```
but so far, my original one is the cleanest and quickest.

provider/openstack/cinder.go
anastasiamac: I went with Christian's suggestion of putting errors.IsNotFound as well as gooseerrors.IsNotFound into one provider method. It does make reading code easier.

provider/gce/google/errors.go
anastasiamac: So after further tests, the answer is a no. This covers all the cases.

provider/gce/google/errors.go
babbageclunk: errors.IsNotFound isn't checking for "not found", it looks for a specific type. I'd rather the check here looked for 3 specific types (if it turns out that RemoveInstance, RemoveFirewall and GetFirewalls all return different ones), rather than checking the string value for "not found". Using the string value seems both overly broad (might match unexpected errors) and fragile - it'll fail if they change the message to something more user-friendly like "couldn't be found" while the type stays the same.

anastasiamac: Ah, it's *compute.OperationErrorErrors... I'll see if I can add it as a 3rd check

anastasiamac: and of course compute.OperationErrorErrors does not implement error :( so i'd still need to check for "not found" string in this case...

At this stage, I do not see the value of type asserting. It's easier to just have a check for 'not found' sting than errors.NotFound, googleapi.Error and still check the string.

babbageclunk: Intriguingly, looking at the code I don't think the actual error that will be returned in that case includes the message "not found". 

The error's being logged from provider/gce/google/raw.go:368, but the error that will be returned is just `waitError{op, nil}`, so the not found error will have been shown to the user and then we'll show a generic "GCE operation blah failed" because it won't match the "not found" check higher up. I think the best way to fix this would be to assume that in the vast majority of cases op.Error.Errors will have 1 error in it, so pull out the first one, log any others and return the first one in the waitError. Then add a .Cause method to waitError so you can pull the internal error out with errors.Cause for your not found check higher up.

What do you think?

anastasiamac: Yes, so I have decided to handle these operational compute 'errors'. There is a func that preserves current behavior for majority of cases, i.e. just log these errors. There is another one to inspect them and if any contain the code I am encountering, then actually convert non-error struct into an error we all recognise and love.
I thin it's a cleaner approach. Have a look - this change is in raw.go. It also works a treat - just tested it live.

We can expand on what we do with different errors or add to the list we consider 'not found' as the need arises.

provider/gce/google/errors.go
```diff
@@ -102,6 +103,15 @@ func HasDenialStatusCode(err error) bool {
 			}
 		}
 	}
+	if cause, ok := errors.Cause(err).(*googleapi.Error); ok {
```
babbageclunk: You can merge these two loops together - they just need cause to be an `error`.

Something like:
```go
var cause error
switch e := errors.Cause(err).(type) {
case *urlError:
    cause = e
case *googleapi.Error:
    cause := e
default: 
    return false
}
for code, descs := range AuthorisationFailureStatusCodes {
     ...
}
return false
```

provider/gce/google/raw.go
```diff
@@ -152,19 +161,54 @@ func (rc *rawConn) UpdateFirewall(projectID, name string, firewall *compute.Fire
...
 }
 
+type handleOperationErrors func(operation *compute.Operation) error
+
+func returnNotFoundOperationErrors(operation *compute.Operation) error {
+	hadOtherErrors := false
+	if operation.Error != nil {
+		for _, err := range operation.Error.Errors {
+			if err.Code == "RESOURCE_NOT_FOUND" {
+				return errors.NotFoundf("resource", err.Message)
```
babbageclunk: This'll hide subsequent errors, won't it? I think my suggestion of stashing the error in the waitError and exposing it with .Cause() is cleaner. It also means that you don't need both returnNotFound and log variants, so you don't need to pass them into waitOperation.

provider/gce/google/raw.go
```diff
@@ -152,19 +161,54 @@ func (rc *rawConn) UpdateFirewall(projectID, name string, firewall *compute.Fire
...
+	return nil
+}
+
 func (rc *rawConn) RemoveFirewall(projectID, name string) error {
 	call := rc.Firewalls.Delete(projectID, name)
 	operation, err := call.Do()
 	if err != nil {
 		return errors.Trace(convertRawAPIError(err))
 	}
+	if err := returnNotFoundOperationErrors(operation); err != nil {
```
babbageclunk: I don't understand why you need to call this here *and* pass it into waitOperation - why isn't just calling it from waitOperation enough? (Same for the other places with this pattern.)

anastasiamac: Because both call.Do() and rc.waitOperation()... run operation... The first run actually happens before the wait call... Essentially, all the wait call does is to check on operation progress on cloud side. 
Since there is a chance that we'd have errors on operation already, we want to process them both after the .Do() cll and in wait...

provider/gce/google/raw.go
anastasiamac: I think you need both but I've updated to use waitError.cause better

provider/gce/google/raw.go
anastasiamac: Actually, looking at the code, we want to give several chances to the operation to succeed. Consequently, we ignore errors until all attempts have been exhausted.

I'll change the code to only inspect once as before.

provider/gce/google/errors.go
babbageclunk: ?

anastasiamac:
On 21/6/19 10:01 am, Christian Muirhead wrote:
>
> *@babbageclunk* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In provider/gce/google/errors.go 
> <https://github.com/juju/juju/pull/10354#discussion_r296055202>:
>
> > @@ -102,6 +103,15 @@ func HasDenialStatusCode(err error) bool {
>   			}
>   		}
>   	}
> +	if cause, ok := errors.Cause(err).(*googleapi.Error); ok {
>
> ?
>
Did you refresh? I do not see this in the code...

anastasiamac:
!!build!!

anastasiamac:
$$merge$$

PR #10369, opened 2019-06-24
Ensure user store uses lowercased username for document id.

hpidcock:
## Description of change

Not all user store operations correctly formatted the mongo document id.
This fixes the behaviour in all affected user state functions and add tests to ensure this behaviour works.

Previously `juju register` with a secret token failed because the update password didn't behave correctly when a username contained uppercase.

## QA steps

Adding a user and registering that user should succeed when the username contains uppercase.
```
juju bootstrap localhost
juju add-user TestName
XDG_DATA_HOME=`mktemp -d` juju register <secret token from add-user>`
```

## Documentation changes

N/A

## Bug reference

[https://bugs.launchpad.net/juju/+bug/1833449](https://bugs.launchpad.net/juju/+bug/1833449)

### DIFF
--- a/state/user.go
+++ b/state/user.go
@@ -37,12 +37,14 @@ func userIDFromGlobalKey(key string) string {
 }
 
 func (st *State) checkUserExists(name string) (bool, error) {
+	lowercaseName := strings.ToLower(name)
+
 	users, closer := st.db().GetCollection(usersC)
 	defer closer()
 
 	var count int
 	var err error
-	if count, err = users.FindId(name).Count(); err != nil {
+	if count, err = users.FindId(lowercaseName).Count(); err != nil {
 		return false, err
 	}
 	return count > 0, nil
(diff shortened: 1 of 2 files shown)


hpidcock:
[juju register fails with usernames that contain uppercase letters](https://trello.com/c/NqBxdEsJ/824-juju-register-fails-with-usernames-that-contain-uppercase-letters)

state/user.go
```diff
@@ -37,12 +37,14 @@ func userIDFromGlobalKey(key string) string {
 }
 
 func (st *State) checkUserExists(name string) (bool, error) {
+	nameToLower := strings.ToLower(name)
```
timClicks: Personal preference.. but I don't like variables that sound like they're about actions in the future, when they're about the past. imo we're representing a `nameLowerCase` rather than a `nameToLower`.

hpidcock: Yeah agreed, there was already code see addUser which named it this way. So I was going for consistency. I don't mind changing them all.

hpidcock:
> General question only - do we know if our documentation specifies that user names are case insensitive?

Not sure. The user name validation allows upper case explicitly in it's regex and the add user logic already supports this (hence why the mongo ids are lowercased)

state/user.go
timClicks: It looks like the variables in those private methods were [created 4 years ago](https://github.com/juju/juju/commit/6c720319d758d216921bd2ea29865d8e80a7f3d2). We probably won't hurt anyone's feelings by changing them now.

Maybe add to PR?

hpidcock:
$$merge$$

PR #10377, opened 2019-06-25
Fixes lp#1564168: Secrets can be shown in json and yaml - warn in tabular.

anastasiamac:
## Description of change

When listing credentials, users may ask to see secrets (if they have access to them). However, secrets could be very big and, thus, are not suitable for a tabular output.

This has always been the case, this PR just makes this very explicit by mentioning this fact in command help, --show-secrets option help output as well as informing the user when an option is used with a tabular output.

## QA steps

1. list credentials with --show-credential in tabular format should inform user that "secrets are not shown in tabular format"

## Bug reference

https://bugs.launchpad.net/juju/+bug/1564168

### DIFF
--- a/cmd/juju/cloud/listcredentials.go
+++ b/cmd/juju/cloud/listcredentials.go
@@ -34,7 +34,7 @@ Note that there can be multiple sets of credentials and, thus, multiple
 names.
 
 Actual authentication material is exposed with the '--show-secrets' 
-option.
+option in json or yaml formats. Secrets are not shown in tabular format.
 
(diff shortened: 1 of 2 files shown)


anastasiamac:
['credentials' show-secrets does not show secrets in tabular format](https://trello.com/c/x2WSRa1q/835-credentials-show-secrets-does-not-show-secrets-in-tabular-format)

anastasiamac:
$$merge$$

anastasiamac:
$$merge$$

PR #10386, opened 2019-06-26
add generation.Abort() to state

hmlanigan:
## Description of change

The first piece of adding the juju abort command for branches to juju: implement generation.Abort(username) in state and add related tests.  Only check that no 
units have been assigned to the branch before aborting.  Checking on charm 
upgrades is a TODO for now as the feature will go on ice after config part is 
implemented.

## QA steps

Abort is not called yet, so no impact to current operations should be felt.  Unit tests
should pass.


hmlanigan:
$$merge$$

PR #10391, opened 2019-06-27
Fixes lp#1642226: validate credential name before using.

anastasiamac:
## Description of change

We validate credential name almost everywhere except during bootstrap process, we validate it very late on the server side.

This PR checks credential name before kicking off bootstrap as per linked bug.

## QA steps

Have an client side credential with a space in the name. For example, I've used "new one":
```
$ juju bootstrap google mycontroller --credential "new one"
ERROR credential name "new one" not valid

$ juju bootstrap google mycontroller
ERROR credential name "new one" not valid
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1642226

### DIFF
--- a/cmd/juju/common/cloudcredential.go
+++ b/cmd/juju/common/cloudcredential.go
@@ -68,10 +68,18 @@ func GetOrDetectCredential(
 	store jujuclient.CredentialGetter,
 	provider environs.EnvironProvider,
 	args modelcmd.GetCredentialsParams,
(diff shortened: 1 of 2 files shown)


anastasiamac:
[credential names with spaces are invalid](https://trello.com/c/uu2fmiGh/838-credential-names-with-spaces-are-invalid)

anastasiamac:
$$merge$$

PR #10403, opened 2019-06-28
Backport PR10203: Remove default memory constrainst for bootstrap in CI tests.

hmlanigan:
## Description of change

Backport PR10203: CI tests shouldn't have a mem=2G constraint when the default for a juju controller is 3.5G. A follow on for the same change for LXD controllers.


hmlanigan:
$$merge$$

PR #10404, opened 2019-06-28
Backport pr10236: remove mem=2G constraints from test bundles and add centos7 to charms

hmlanigan:
## Description of change

Backport pr10236: remove mem=2G constraints from test bundles and add centos7 to charms


hmlanigan:
$$merge$$

PR #10405, opened 2019-06-28
Excessivewarnings on controller with ci runs on GCE

hmlanigan:
## Description of change

The config value "project-id" from the cloud-city environments.yaml should not be added to the model-config during ci test runs.  It's used by the GCE credentials, not provider and causes excessive warning messages in the controller log.

### DIFF
--- a/acceptancetests/jujupy/client.py
+++ b/acceptancetests/jujupy/client.py
@@ -912,6 +912,7 @@ class ModelClient:
             'name',
             'password',
             'private-key',
+            'project-id',
             'region',
             'sdc-key-id',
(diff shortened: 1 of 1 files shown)


hmlanigan:
WARNING juju.environs.config config.go:1570 unknown config field "project-id"

http://10.125.0.203:8080/job/nw-deploy-client-macos/1201/artifact/artifacts/controller/machine-0/machine-0.log.gz

hmlanigan:
mongo not available error, return build

!!build!!

hmlanigan:
$$merge$$

PR #10406, opened 2019-07-01
De-dupe store in list cloud.

anastasiamac:
## Description of change

While working on 'credentials', I've noticed that 'clouds' has 2 references to client store one directly on the command, one in the super command. Only one reference is needed and considering that both are implementation of the file store, consequences of having two instances could be surprising.

This PR de-dupe the reference, only keeping the one on the super-command.


anastasiamac:
[de-dupe store in 'clouds'](https://trello.com/c/7aBUFjVg/843-de-dupe-store-in-clouds)

anastasiamac:
$$merge$$

PR #10409, opened 2019-07-01
fix RootBlockDevice unit tests on non amd64 boxes

hmlanigan:
## Description of change

Prevent RootBlockDevice tests from failing if not run on amd64
machines due the image id, which becomes the BlockDeviceMapping
UUID, being different based on platform.  Setup the config of
these tests to always be amd64.

## QA steps

Run the OpenStack provider unit tests on platforms other than 
amd64.


hpidcock:
!!build!!

hmlanigan:
$$merge$$

PR #10418, opened 2019-07-03
LXD certificates are hidden attributes.

anastasiamac:
## Description of change

All cloud authentication mechanisms can specify attributes that are hidden by default - password, secret keys, etc.
LXD certificate auth type was missing Hidden property settings for such secret properties.
These properties can only be visible to  the owner of the credential when a show-secret option is provided.

```
$ juju credentials localhost --format yaml
local-credentials:
  localhost:
    default-credential: localhost
    localhost:
      auth-type: certificate
```
not

```
$ juju credentials localhost --format yaml
local-credentials:
  localhost:
    default-credential: localhost
    localhost:
      auth-type: certificate
      client-cert: <REDACTED>
      client-key: <REDACTED>
      server-cert: <REDACTED>
```

### DIFF
--- a/provider/lxd/credentials.go
+++ b/provider/lxd/credentials.go
@@ -55,7 +55,6 @@ type CertificateGenerator interface {
 
 // NetLookup groups methods for looking up hosts and interface addresses.
 type NetLookup interface {
-
 	// LookupHost looks up the given host using the local resolver.
 	// It returns a slice of that host's addresses.
 	LookupHost(string) ([]string, error)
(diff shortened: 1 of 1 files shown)


anastasiamac:
[hide lxd certificate auth type secret properties](https://trello.com/c/4HMTxJnR/845-hide-lxd-certificate-auth-type-secret-properties)

anastasiamac:
$$merge$$

SimonRichardson:
I think I did this so you could easily override them. I'm a bit foggy on this situation as it's well over a year ago and the use case might have dissolved as we got more of the LXD cluster setup working. Originally that was the idea.

SimonRichardson:
Note: I'm fine with this if we didn't break any LXD cluster setup, which I doubt we have.

anastasiamac:
intermittent, unrelated failure in github.com/juju/juju/worker/uniter/relation.TestPackage

anastasiamac:
$$merge$$

PR #10424, opened 2019-07-04
Use spec name for zone name

SimonRichardson:
## Description of change

When not in clustered mode and trying to use a different name for
the LXD profile in clouds.yaml, the server name will be wrongly
be reported by lxd as the host name. With recent changes to the
cloud spec and credentials, this will cause LXD to fail bootstrapping.

The bug in question refers to 2.5.9, but in fact a lot of reworking of
credentials have made this fix non-trivial to back port it and make it
work easily.

By over writing the server name with the spec name in non-clustered
environments we should then be allowed to bootstrap at will.

## QA steps

If you change your `~/.local/share/juju/clouds.yaml` to copy and
paste the "lxd" cloud you have existing, but change the name (in
this instance to "local"):

```yaml
clouds:
  lxd:
    type: lxd
    auth-types: [certificate]
    endpoint: https://192.168.1.201:8443
    regions:
      default:
        endpoint: https://192.168.1.201:8443
  local:
    type: lxd
    auth-types: [certificate]
    endpoint: https://192.168.1.201:8443
    regions:
      default:
        endpoint: https://192.168.1.201:8443
```

Then run bootstrap

```console
juju bootstrap local test --no-gui
```

The following should bootstrap fine.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1813650

### DIFF
--- a/container/lxd/server.go
+++ b/container/lxd/server.go
@@ -85,6 +85,11 @@ func NewRemoteServer(spec ServerSpec) (*Server, error) {
 		return nil, errors.Trace(err)
 	}
 	svr, err := NewServer(cSvr)
+	// If we're not clustered we should set the name to the name of the cloud
+	// spec name, so that we can use that name for the zones.
+	if err == nil && !svr.IsClustered() {
+		svr.name = spec.Name
+	}
 	return svr, err
 }
 


SimonRichardson:
https://trello.com/c/1kkH6t43/1620-1813650-lxd-bootstrap-fails-with-credentials-not-found-for-local-cloud-26

hpidcock:
@SimonRichardson would be good to see a unit test added to test this behaviour.

hmlanigan:
interesting, i can only reproduce the issue with 2.5, not 2.6 nor develop. was looking at qa for this pr.

SimonRichardson:
@hmlanigan you're right... i'll see if i can fix it again with 2.5, otherwise the work around might be move to 2.6 or rename your local name to lxd.

PR #10427, opened 2019-07-05
Allow more than 11 EBS volumes to be allocated to an AWS EC2 instances

timClicks:
## Description of change

Juju currently puts on an artificial limit on the number of volumes that may be attached to an instance. From the bug report:

> The EBS volume provider currently limits each machine to a maximum of 
> 11 volume attachments, because it follows the AWS recommendation of 
> naming block devices in the range xvdf-xvdp. It is permissible to use 
> xvdf-xvdz, and it's not clear why those names are not recommended for use.
> 
> We should not artificially limit them, and just let the AWS APIs complain if there's an issue.

## QA steps

* bootstrap to AWS
* provision >11 storage volumes
* check `juju status --storage`, all storage units should have the status "attached"

## Documentation changes

None. 

## Bug reference

* https://bugs.launchpad.net/juju/+bug/1517474


timClicks:
$$merge$$

timClicks:
$$merge$$

PR #10434, opened 2019-07-10
Report timeout rather than misleading error message

timClicks:
While logging into a controller via OAuth, we present an error message that users can't do much with after 1 minute:

```
ERROR cannot get discharge from "https://api.jujucharms.com/identity": cannot unmarshal wait error response: invalid character '<' looking for beginning of value
```

This commit changes the error message to this:

```
ERROR cannot log into "azure": cannot log in: context deadline exceeded
```

## QA steps

- Run `juju login azure`, then wait for approx. 1 minute

## Documentation changes

-

## Bug reference

* https://bugs.launchpad.net/juju/+bug/1673926

### DIFF
--- a/cmd/juju/user/login.go
+++ b/cmd/juju/user/login.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/juju/cmd"
 	"github.com/juju/collections/set"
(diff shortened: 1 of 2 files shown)


timClicks:
$$merge$$

timClicks:
$$merge$$

PR #10447, opened 2019-07-19
'add-credential' also uploads it to a given controller when possible.

anastasiamac:
## Description of change

'add-credential' only added it to local client cache. As per linked bug, this was confusing to the users and, in fact, was also restrictive as the only way to add your own credential to the controller was either by bootstrapping it or by creating a model. 

That also led to difficulties in scenario where users wanted to replace credential that a model uses ('set-model-credential') but said credential was not on the controller yet.

This PR ensures that a credential when added to client is also uploaded to a current or given controller. If controller is not provided/found, previous behavior is preserved, i.e. credential is only added locally. If only local addition is desired, i.e. no upload, a --local flag is provided as per similar commands. 

## QA steps

1. add credential without controller will only add locally
2. when a current controller exists or a different controller is provided, a credential is also uploaded to that controller [ verify with 'juju credentials -c <controller used>']

## Bug reference

https://bugs.launchpad.net/juju/+bug/1814395 (FP2)

### DIFF
--- a/cmd/juju/cloud/addcredential.go
+++ b/cmd/juju/cloud/addcredential.go
@@ -14,10 +14,12 @@ import (
 	"github.com/juju/gnuflag"
 	"gopkg.in/juju/names.v2"
 
+	apicloud "github.com/juju/juju/api/cloud"
 	jujucloud "github.com/juju/juju/cloud"
 	jujucmd "github.com/juju/juju/cmd"
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/juju/interact"
+	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/environs"
 	"github.com/juju/juju/jujuclient"
 )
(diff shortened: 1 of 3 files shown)


anastasiamac:
[add-credentials uploads remote credentials](https://trello.com/c/1FvBGgWN/854-add-credentials-uploads-remote-credentials)

cmd/juju/cloud/addcredential.go
```diff
@@ -166,9 +194,19 @@ func (c *addCredentialCommand) Run(ctxt *cmd.Context) error {
 	}
 	// Check that the supplied cloud is valid.
 	var err error
-	if c.cloud, err = common.CloudOrProvider(c.CloudName, c.cloudByNameFunc); err != nil {
-		if !errors.IsNotFound(err) {
-			return err
+	if !c.Local {
+		if err := c.maybeRemoteCloud(ctxt); err != nil {
```
ycliuhw: not sure if we need to return error if no remote cloud found when c.Local == false

cmd/juju/cloud/addcredential.go
```diff
@@ -538,6 +584,60 @@ func (c *addCredentialCommand) promptFieldValue(p *interact.Pollster, attr jujuc
 	}
 }
 
+func (c *addCredentialCommand) credentialsAPI() (CredentialAPI, error) {
+	var err error
```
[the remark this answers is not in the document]
anastasiamac: ah yes thank you ;)

cmd/juju/cloud/addcredential.go
anastasiamac: Good point...
The value of c.Local does not matter. However, if we could not find the remote cloud, we should not try to upload but guide the user instead.

PR #10448, opened 2019-07-19
De-mystify controller configuration keys.

anastasiamac:
## Description of change

This PR takes the guessing work out of what configuration can be changed on the running controller.
Just like 'juju help model-config' exposes keys for configuration properties that the command can modify, 'juju help controller-config' exposes editable items' keys.

## QA steps

'juju help controller-config' contains this list https://pastebin.ubuntu.com/p/sxbvq4PQyV/

## Bug reference

https://bugs.launchpad.net/juju/+bug/1836815

### DIFF
--- a/cmd/juju/common/cloud.go
+++ b/cmd/juju/common/cloud.go
@@ -4,6 +4,9 @@
 package common
 
 import (
+	"bytes"
+
+	"github.com/juju/cmd"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	"gopkg.in/juju/environschema.v1"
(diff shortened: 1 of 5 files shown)


anastasiamac:
[controller-config help needs properties list](https://trello.com/c/VbiZbwG1/857-controller-config-help-needs-properties-list)

cmd/juju/common/cloud.go
```diff
@@ -157,3 +160,12 @@ type PrintConfigSchema struct {
 	Type        string `yaml:"type,omitempty" json:"type,omitempty"`
 	Description string `yaml:"description,omitempty" json:"description,omitempty"`
 }
+
+func FormatConfigSchema(values interface{}) (string, error) {
```
hpidcock: Would be awesome to format this nicer, but seems future work. Yaml is easy to read, but it doesn't fit well with the --help output for controller-config.

anastasiamac: Yes, but at the moment we are after consistency and this is what model-config does (as well as other output that prints attibutes' help).

PR #10450, opened 2019-07-22
'autoload-credentials' uploads credentials to the controller.

anastasiamac:
## Description of change

'autoload-credentials' scans known locations on local client (device) for cloud supplied files that may contain cloud credentials that a user wants Juju  to be aware of. Previously this command only had ability to add credentials to Juju client locally. This was confusing for the users and there were reports where operators would say 'I have just autoloaded credentials but Juju still cannot find them'.

This PR adds the ability to upload these discovered credentials to the controller to avoid confusion.

As a drive-by, fix for a linked bug to allow users to quit at 2nd prompt line too.

The command now also supports built-in clouds. For example, previously it could detect lxd credentials but not able to load them.

## QA steps

1. 'autoload-credentials' works locally without a current, bootstrapped controller
2.  'autoload-credentials' works with a current or specified controller
3. User can quit at a 'cloud name' prompt. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1837009

### DIFF
--- a/cmd/juju/cloud/detectcredentials.go
+++ b/cmd/juju/cloud/detectcredentials.go
@@ -466,6 +466,12 @@ func (c *detectCredentialsCommand) addRemoteCredentials(ctxt *cmd.Context, cloud
 		return err
 	}
 
+	client, err := c.credentialAPIFunc()
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
 	var results []params.UpdateCredentialResult
 	moreCloudInfoNeeded := false
 	for cloud, byCloud := range all {
@@ -480,11 +486,6 @@ func (c *detectCredentialsCommand) addRemoteCredentials(ctxt *cmd.Context, cloud
 			if len(verified) == 0 {
 				return erred
 			}
-			client, err := c.credentialAPIFunc()
-			if err != nil {
-				return err
-			}
-			defer client.Close()
 			result, err := client.UpdateCloudsCredentials(verified)
 			if err != nil {
 				logger.Errorf("%v", err)
--- a/cmd/juju/cloud/detectcredentials.go
+++ b/cmd/juju/cloud/detectcredentials.go
@@ -14,20 +14,25 @@ import (
 	"github.com/juju/cmd"
 	"github.com/juju/collections/set"
 	"github.com/juju/errors"
+	"github.com/juju/gnuflag"
+	"github.com/juju/naturalsort"
 
+	apicloud "github.com/juju/juju/api/cloud"
+	"github.com/juju/juju/apiserver/params"
 	jujucloud "github.com/juju/juju/cloud"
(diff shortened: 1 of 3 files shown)


anastasiamac:
[autoload-credentials: quitting at 'enter cloud name' is ignored.](https://trello.com/c/McZ6WITp/859-autoload-credentials-quitting-at-enter-cloud-name-is-ignored)

anastasiamac:
[autoload-credentials should upload loaded credentials remotely too](https://trello.com/c/S4folGaN/861-autoload-credentials-should-upload-loaded-credentials-remotely-too)

cmd/juju/cloud/detectcredentials.go
```diff
@@ -122,18 +154,54 @@ type discoveredCredential struct {
...
+		}
+		defer client.Close()
+
+		remoteUserClouds, err := client.Clouds()
+		if err != nil {
+			return nil, err
+		}
+		c.remoteClouds = map[string]jujucloud.Cloud{}
+		for k, v := range remoteUserClouds {
+			clouds[k.Id()] = v
```
ycliuhw: do we want to warn if the personal cloud(same name but different content) gets overwritten by remote cloud?

cmd/juju/cloud/detectcredentials.go
```diff
@@ -309,18 +387,30 @@ func (c *detectCredentialsCommand) interactiveCredentialsUpdate(ctxt *cmd.Contex
...
 			fmt.Fprintln(ctxt.Stderr, "No cloud name entered.")
 			continue
 		}
+		cloud, err := common.CloudOrProvider(cloudName, c.cloudByNameFunc)
+		if err != nil {
+			fmt.Fprintln(ctxt.Stderr, err.Error())
+			continue
+		}
+		if cloud.Type != cred.cloudType {
+			fmt.Fprintln(ctxt.Stderr, errors.Errorf("chosen credentials not compatible with a %s cloud", cloud.Type))
```
ycliuhw: do we want to return error or continue rather than go to next line?

anastasiamac: Yes, thank you :) good pickup, fixing now :D

cmd/juju/cloud/detectcredentials.go
anastasiamac: Not in this case. Here we do want to be consistent with all other places where we are dealing with both local and remote clouds. To say anything additional here would be too noisy.

However, in our f-2-f discussion, this does bring an in interesting point about what happens when local and remote clouds mismatch? what happens when remote clouds get updated? What happens if it gets updated with  outdated information? All interesting questions that need to be answered separately (outside of this PR) and consistently (across all functionalities that deal with local/remote clouds as well as deal with cloud management).

PR #10452, opened 2019-07-22
Improve user facing messaging when adding cloud/model.

anastasiamac:
## Description of change

Adding a cloud as well as adding a model can be simple and exhilarating but when things go wrong and starts do not quiet align, Juju can be better at helping users fixing the problems.

This PR improves user facing messaging that occur due to credentials misalignment - either local/remote clouds do not work, credentials does not exist remotely on the controller, etc.

For example, when adding a cloud remotely that does not have local credentials to upload, the user will be presented with some guidance:
```
$ juju add-cloud manymore -c mycontroller
Cloud "manymore" added to controller "mycontroller".
ERROR loading credentials: credentials for cloud manymore not found
To upload credentials to the controller for cloud "manymore", use 
* 'add-model' with --credential option or
* 'add-credential -c manymore'.
```

When adding a duplicate cloud to the controller previously, Juju will not say anything and act as if it was a successful operation. This PR reports duplicate name and error. 

When model is being added to the cloud that does not exist on the controller, users are also guided as to how to add that cloud.

### DIFF
--- a/cmd/juju/cloud/add.go
+++ b/cmd/juju/cloud/add.go
@@ -161,7 +161,6 @@ type AddCloudCommand struct {
 	// These attributes are used when adding a cloud to a controller.
 	controllerName  string
 	credentialName  string
-	store           jujuclient.ClientStore
 	addCloudAPIFunc func() (AddCloudAPI, error)
 }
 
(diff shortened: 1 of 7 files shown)


hpidcock:
LGTM

anastasiamac:
[provide more guidance when adding models and clouds re: local vs remote](https://trello.com/c/8p0E2gE0/862-provide-more-guidance-when-adding-models-and-clouds-re-local-vs-remote)

PR #10453, opened 2019-07-22
Malformed error

SimonRichardson:
## Description of change

When attempting to contact HTTPS url using a HTTP endpoint, in
go version < 1.12 it would return with a malformed garbage error.
In go version >= 1.12 the error is no longer returned and instead
a status code of 400 is returned.

In order to handle this is the correct manor for both past and future
runtimes we should hide this fact and just get prior errors report
that a bad request was sent.

## QA steps

Tests go green.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1837357

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -114,7 +114,7 @@
   name = "github.com/juju/romulus"
 
 [[constraint]]
-  revision = "e25f034181e5ae6fa6cd7ba86c804f48388f1ff7"
+  revision = "fe4dde6a8d8a5c36350355052d6e04c427c6efcf"
   name = "github.com/juju/testing"
   source = "github.com/SimonRichardson/testing"
 
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -114,9 +114,8 @@
   name = "github.com/juju/romulus"
 
 [[constraint]]
-  revision = "fe4dde6a8d8a5c36350355052d6e04c427c6efcf"
+  revision = "ce30eb24acd2c45a1bc42208835468ca619685a9"
   name = "github.com/juju/testing"
-  source = "github.com/SimonRichardson/testing"
 
 [[constraint]]
   revision = "bf9cc5bdd62dabc40b7f634b39a5e2dc44d44c45"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -114,8 +114,9 @@
   name = "github.com/juju/romulus"
 
 [[constraint]]
-  revision = "6570bd8f8541d15ce654437084f8edcd7b76637b"
+  revision = "e25f034181e5ae6fa6cd7ba86c804f48388f1ff7"
   name = "github.com/juju/testing"
+  source = "github.com/SimonRichardson/testing"
 
 [[constraint]]
   revision = "bf9cc5bdd62dabc40b7f634b39a5e2dc44d44c45"
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ endif
 
 PROJECT := github.com/juju/juju
 PROJECT_DIR := $(shell go list -e -f '{{.Dir}}' $(PROJECT))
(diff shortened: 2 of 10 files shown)


SimonRichardson:
https://trello.com/c/zH1JmAfj/1656-bug-1837357-s390x-fix-26-issues-around-tls-errors-for-golang-111

jameinel:
!!build!!

Gopkg.toml
```diff
@@ -114,8 +114,9 @@
   name = "github.com/juju/romulus"
 
 [[constraint]]
-  revision = "6570bd8f8541d15ce654437084f8edcd7b76637b"
+  revision = "fe4dde6a8d8a5c36350355052d6e04c427c6efcf"
   name = "github.com/juju/testing"
+  source = "github.com/SimonRichardson/testing"
```
jameinel: If we need a testing patch, we probably just want to update juju/testing first?

worker/httpserver/worker_test.go
```diff
@@ -243,7 +243,7 @@ func (s *WorkerSuite) TestMinTLSVersion(c *gc.C) {
 	tlsConfig.MaxVersion = tls.VersionSSL30
 
 	conn, err := tls.Dial("tcp", parsed.Host, tlsConfig)
-	c.Assert(err, gc.ErrorMatches, ".*protocol version not supported")
+	c.Assert(err, gc.ErrorMatches, "tls: no supported versions satisfy MinVersion and MaxVersion")
```
jameinel: I assume the reason we can actually change this is hidden in the testing package change.

jameinel:
14:31:03 E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/p/python2.7/libpython2.7-minimal_2.7.15-4ubuntu4~18.04_amd64.deb  Temporary failure resolving 'archive.ubuntu.com'

!!build!!

worker/httpserver/worker_test.go
jameinel: I don't see how this would work with go-1.11 since it gives "tls: protocol version not supported" there.
This looks more like you tested using go-1.12 on the 2.6 codebase.

Gopkg.toml
SimonRichardson: So I do this normally to get ok'd from reviewers that this is the right way to do this, so I don't have rev the testing lib and once I've got signed off with that, I land testing repo then update this repo.

worker/httpserver/worker_test.go
SimonRichardson: I tested both go-1.11 and go-1.12.

SimonRichardson: I've tweaked the error message to ensure we do handle 1.11 and 1.12

SimonRichardson:
!!build!!

Makefile
```diff
@@ -7,7 +7,7 @@ endif
 
 PROJECT := github.com/juju/juju
 PROJECT_DIR := $(shell go list -e -f '{{.Dir}}' $(PROJECT))
-PROJECT_PACKAGES := $(shell go list $(PROJECT)/... | grep -v /vendor/)
+PROJECT_PACKAGES := $(shell go list $(PROJECT)/... | grep -v /vendor/ | grep -v /acceptancetests/)
```
jameinel: I'm curious why this was needed. Did we end up with an acceptance test in go that is causing problems?

SimonRichardson: We end up attempting to install everything in acceptance tests, which we shouldn't do to begin with. Also CI also has an [issue](https://jenkins.juju.canonical.com/job/github-make-check-juju/90/console) with the following error:

```console
15:07:41 go install github.com/juju/juju/acceptancetests/repository/trusty/fill-logs/actions: build output "/home/ubuntu/go/bin/actions" already exists and is not an object file
15:07:56 Makefile:82: recipe for target 'go-install' failed
```

SimonRichardson:
$$merge$$

PR #10461, opened 2019-07-24
Enable Juju to be deployed on vSAN-backed storage volumes

timClicks:
This change enables Juju to bootstrap itself and deploy charms/bundles on vSphere instances that are backed by vSAN.

## QA steps

* juju bootstrap vsphere
* juju deploy ubuntu

## Documentation changes

Required 

## Bug reference

* https://bugs.launchpad.net/juju/+bug/1800940


timClicks:
!!build!!

provider/vsphere/environ_broker.go
```diff
@@ -289,9 +290,13 @@ func (env *sessionEnviron) AllInstances(ctx context.ProviderCallContext) ([]inst
 
 	// Turn mo.VirtualMachine values into *environInstance values,
 	// whether or not we got an error.
-	results := make([]instances.Instance, len(vms))
-	for i, vm := range vms {
-		results[i] = newInstance(vm, env.environ)
+	results := make([]instances.Instance, 0, len(vms))
```
wallyworld: Overkill for a small list I think.
Can just be:
`var results []instances.Instance`

There should be a unit test update also

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -456,23 +458,65 @@ func (c *Client) destroyVM(
...
+		configSpec.MemoryMB = int64(*args.Constraints.Mem)
+	}
+	if args.Constraints.HasCpuPower() {
+		cpuPower := int64(*args.Constraints.CpuPower)
+		configSpec.CpuAllocation = &types.ResourceAllocationInfo{
+			Limit:       &cpuPower,
+			Reservation: &cpuPower,
+		}
+	}
+	if configSpec.Flags == nil {
```
wallyworld: Flags is always nil

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -456,23 +458,65 @@ func (c *Client) destroyVM(
...
+	args CreateVirtualMachineParams,
 	srcVM *object.VirtualMachine,
 	dstName string,
 	vmFolder *object.Folder,
+	datacenter *object.Datacenter,
 	taskWaiter *taskWaiter,
 ) (*object.VirtualMachine, error) {
+	configSpec := &types.VirtualMachineConfigSpec{}
+
+	if args.Constraints.HasCpuCores() {
```
wallyworld: So we were ignoring constraints?

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -456,23 +458,65 @@ func (c *Client) destroyVM(
...
-	return object.NewVirtualMachine(c.client.Client, info.Result.(types.ManagedObjectReference)), nil
+
+	vm := object.NewVirtualMachine(c.client.Client, info.Result.(types.ManagedObjectReference))
+
+	if args.Constraints.RootDisk != nil {
+		// The user specified a root disk, so extend the VM's
+		// disk before powering the VM on.
+		args.UpdateProgress(fmt.Sprintf(
+			"extending disk to %s",
+			humanize.IBytes(*args.Constraints.RootDisk*1024*1024),
```
wallyworld: I think there's a helper we can use that doesn't require 1024 * 1024

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -106,30 +108,28 @@ type CreateVirtualMachineParams struct {
...
-//      with the Juju machine.
-//   5. If the user specified a root-disk constraint, extend the VMDK if its
-//      capacity is less than the specified constraint.
-//   6. Power on the virtual machine.
+// Example:
+//   vmTemplatePath("juju-abc123-1")
+//   "juju-abc123-template"
+func vmTemplatePath(vmpath string) string {
+	parts := strings.Split(vmpath, "-")
+	templateNameParts := []string{
```
wallyworld: We should validate input to be sure that len(parts) >= 2 or else we might get a bounds error

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -106,30 +108,28 @@ type CreateVirtualMachineParams struct {
...
+		parts[0],
+		parts[1],
+		"template",
+	}
+	return strings.Join(templateNameParts, "-")
+}
+
+// CreateVirtualMachine creates and powers on a new VM.
```
wallyworld: Should we keep the detailed explanation?

timClicks: Happy to add one, but I did want to make sure that we weren't keeping the stale comment around.

provider/vsphere/internal/vsphereclient/createvm.go
timClicks: Wondered about this, but figured since juju sets the names, we're probably safe. Have changed.

FrederikNS:
Thanks for working on this @timClicks, however I'm having a bit of trouble when actually trying it out.

I have tried creating the controller using both a regular datastore and a vsanDatastore, and neither have worked for me. I got two different error messages during the bootstrap:

    $ juju bootstrap cloud-vsphere-a9c juju-controller --config datastore=datastore1 --config "primary-network=K8S(161)"
    Creating Juju controller "juju-controller" on cloud-vsphere-a9c/K8s Datacenter
    Looking for packaged Juju agent version 2.7-beta1 for amd64
    No packaged binary found, preparing local Juju agent binary
    Launching controller instance(s) on cloud-vsphere-a9c/K8s Datacenter...
    - creating template VM: streaming http://cloud-images.ubuntu.com/releases/server/releases/bionic/release-20190722.1/ubuntu-18.04-server-cloudimg-amd64.ova to https://vmhost27.cphdERROR failed to bootstrap model: cannot start bootstrap instance in availability zone "K8s Cluster 00": creating template VM: streaming http://cloud-images.ubuntu.com/releases/server/releases/bionic/release-20190722.1/ubuntu-18.04-server-cloudimg-amd64.ova to https://vmhost27.cphdomain.com/nfc/52d11d0b-bf93-0f5f-ce7f-fbcb06509e41/disk-0.vmdk: Post https://vmhost27.cphdomain.com/nfc/52d11d0b-bf93-0f5f-ce7f-fbcb06509e41/disk-0.vmdk: EOF

    $ juju bootstrap cloud-vsphere-a9c juju-controller --config datastore=vsanDatastore --config "primary-network=K8S(161)"
    Creating Juju controller "juju-controller" on cloud-vsphere-a9c/K8s Datacenter
    Looking for packaged Juju agent version 2.7-beta1 for amd64
    No packaged binary found, preparing local Juju agent binary
    Launching controller instance(s) on cloud-vsphere-a9c/K8s Datacenter...
    - creating template VM: streaming http://cloud-images.ubuntu.com/releases/server/releases/bionic/release-20190722.1/ubuntu-18.04-server-cloudimg-amd64.ova to https://vmhost28.cphdERROR failed to bootstrap model: cannot start bootstrap instance in availability zone "K8s Cluster 00": creating template VM: streaming http://cloud-images.ubuntu.com/releases/server/releases/bionic/release-20190722.1/ubuntu-18.04-server-cloudimg-amd64.ova to https://vmhost28.cphdomain.com/nfc/524f4823-6a16-cb1a-dbeb-ef70da55b9ce/disk-0.vmdk: 500 Internal Server Error

I am watching the progress in the vSphere meanwhile, and can see that the upload of the OVF just stops part way. I have let it sit for multiple minutes, but nothing changed.

image

I tried canceling by hitting ctrl-c, but that hung as well. Finally if I cancel the task directly in vSphere, juju finishes.

I did re-try using juju 2.6.5, and that works correctly when using a normal datastore.

Let me know if I can help out with anything else in relation to this.

timClicks:
@FrederikNS Thanks so much for testing this out! That 500 error is being generated by your vSphere instance. Is it possible to retrieve any logs?

timClicks:
After some further testing, it looks as though this change introduces a regression in the `kill-controller` command.

> $ JUJU_LOGGING_CONFIG="<root>=DEBUG;juju.provider.vmware=TRACE" $GOPATH/bin/juju kill-controller -y c-testing
Unable to open API: open connection timed out
Unable to connect to the API server, destroying through provider
ERROR A specified parameter was not correct: ConfigSpec.files.vmPathName
ERROR destroying instances: failed to stop instance juju-ea0863-template: failed to remove instances: A specified parameter was not correct: ConfigSpec.files.vmPathName

timClicks:
@FrederikNS could you please try the following command and then post your output to https://paste.ubuntu.com? 

```bash
$GOPATH/bin/juju bootstrap cloud-vsphere-a9c juju-controller --config datastore=vsanDatastore --config "primary-network=K8S(161)" --show-log --logging-config="<root>=DEBUG;juju.provider.vmware=TRACE"
```

FrederikNS:
@timClicks Here's the logs from the bootstrap. It's probably most interesting around the 09:57:44 timestamp, as that is where the template upload terminates prematurely, and juju seems to try to continue to do it's thing, even though the template hasn't completed uploading.

https://paste.ubuntu.com/p/YkrkTX8d5n/

I have not been able to get any logs from VMware yet, but I'm talking to my IT department to get access to some logs.

timClicks:
Thanks for taking the time to send in extra information. Will take a good look within 24h. I am sure that I will be able to get this working for you.

FrederikNS:
It would seem that I was actually experiencing some network related issues, as I have now created a virtual machine on the vSphere from where I run the juju commands, and bootstrapping the controller now works correctly. It does however seem that juju is very sensitive to any network flickering.

However I'm still having trouble with deploying charmed-kubernetes. At first it spins up 3 VMs in turn, which seems to be working fine. But right after that it spins up the remaining 6 VMs all at once, which fails. I'm not sure it's related to this pull request though.

timClicks:
@FrederikNS really great to hear that you've been able to bootstrap. I'll investigate the behaviour with that charmed-kubernetes, as we're very interested in making sure that this is a seamless experience on vSphere.

timClicks:
> After some further testing, it looks as though this change introduces a regression in the `kill-controller` command.

This has proven to not be a problem for this PR. vSphere's own console returned the same error. It looks like I put the VM into an inconsistent state while playing around with kill controller and aborting prematurely.

provider/vsphere/internal/vsphereclient/client.go
hpidcock: Should be humanize.IBytes(*args.Constraints.RootDisk * humanize.GiByte)

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -159,87 +177,131 @@ func (c *Client) CreateVirtualMachine(
...
+	}
+	defer ovaReadCloser.Close()
+	ovaTarReader := tar.NewReader(ovaReadCloser)
+	for {
+		header, err := ovaTarReader.Next()
+		if err != nil {
+			return nil, errors.Annotate(err, "reading OVA")
+		}
+		if strings.HasSuffix(header.Name, ".vmdk") {
+			item := info.Items[0]
```
hpidcock: If the OVA has multiple vmdk's, how do we select the right one? It appears we just upload the first one.

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -159,87 +177,131 @@ func (c *Client) CreateVirtualMachine(
...
-		}
-	}()
-	if _, err := c.detachDisk(ctx, tempVM, taskWaiter); err != nil {
+	templateVM, err := c.ensureTemplateVM(ctx, args)
+	if err != nil {
+		return nil, errors.Annotate(err, "creating template VM")
+	}
+
+	args.UpdateProgress("cloning template")
+	vm, err2 := c.cloneVM(ctx, args, templateVM, vmFolder)
```
hpidcock: I'm confused why there is an err2? If this error needs to be used elsewhere, perhaps give it a better name, but I don't see it used elsewhere.

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -343,55 +379,53 @@ func (c *Client) createImportSpec(
...
-						Datastore: &ds,
-					},
-				},
-			},
-		}
-		deviceConfigSpec.Device = disk
-		deviceConfigSpec.FileOperation = "" // attach existing disk
+	datastore *object.Datastore,
+) (*types.OvfCreateImportSpecResult, error) {
+	c.logger.Debugf("Creating import spec")
```
hpidcock: These logs could have more useful information, or just be removed.

provider/vsphere/environ_broker.go
timClicks: Cleaned up, thanks.

provider/vsphere/internal/vsphereclient/client.go
timClicks: No. They didn't appear here as they were being dealt with in a different area of the code.

provider/vsphere/internal/vsphereclient/client.go
timClicks: Sorry, tunnel vision there I think. Fixed.

provider/vsphere/internal/vsphereclient/client.go
timClicks: Thanks @hpidcock. Updated.

provider/vsphere/internal/vsphereclient/createvm.go
timClicks: This should be filed as a new bug. We're mirroring the old business logic here. This hasn't been an issue for Juju thus far as we choose OVAs from a known source that only has a single VMDK.

provider/vsphere/internal/vsphereclient/createvm.go
timClicks: Oh good point. It's a wart from a previous iteration. Will fix.

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -149,8 +143,32 @@ func (c *Client) CreateVirtualMachine(
...
+// that instances can be created from.
+func (c *Client) ensureTemplateVM(
+	ctx context.Context,
+	args CreateVirtualMachineParams,
+) (*object.VirtualMachine, error) {
+	finder, datacenter, err := c.finder(ctx)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	templateVM, err := finder.VirtualMachine(ctx, vmTemplateName(args.Name))
```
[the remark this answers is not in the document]
timClicks: Fixed. Template naming is now based on the SHA256 value of the .ova.

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -441,38 +443,94 @@ func (c *Client) DeleteDatastoreFile(ctx context.Context, datastorePath string)
...
-	taskWaiter *taskWaiter,
-) error {
-	task, err := vm.Destroy(ctx)
-	if err != nil {
-		return errors.Trace(err)
-	}
-	_, err = taskWaiter.waitTask(ctx, task, "destroying VM")
-	return errors.Trace(err)
+// rootDiskB converts the root disk constraint (which uses megabytes for its unit) to bytes
+func rootDiskB(diskMiB uint64) uint64 {
```
babbageclunk: These might be better named for what they do, rather than that they're used on the root disk size - megabytesToB and megabytesToKiB?

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -106,51 +111,40 @@ type CreateVirtualMachineParams struct {
...
+
+// ensureTemplateVM returns a vSphere template VM
+// that instances can be created from.
+func (c *Client) ensureTemplateVM(
 	ctx context.Context,
 	args CreateVirtualMachineParams,
-) (_ *mo.VirtualMachine, resultErr error) {
-
-	// Locate the folder in which to create the VM.
+) (*object.VirtualMachine, error) {
```
babbageclunk: You need the error return named, otherwise the use of `err` in the deferred func to abort the lease won't necessarily correspond to what's being returned - it'll be the local variable `err`, rather than the return error. For example, the return on line 253:
```go
    return nil, errors.New("SHA-256 hash mismatch for OVA")
```
That doesn't set `err`, so the func will see `err == nil` and not abort the lease, but if the name `err` in the func refers to the named return it will see it.

Basically you need to replace the err local var with the named return parameter err (which the code in the method can still assign to in the same way). Hopefully that makes sense? (This is about the fiddliest that scoping in Go gets.)

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -441,38 +443,94 @@ func (c *Client) DeleteDatastoreFile(ctx context.Context, datastorePath string)
...
+			return nil, errors.Trace(err)
+		}
+	}
+
+	vmConfigSpec := types.VirtualMachineConfigSpec{}
+	err = c.buildConfigSpec(ctx, args, &vmConfigSpec)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	task, err = vm.Reconfigure(ctx, vmConfigSpec)
```
babbageclunk: Aren't you setting all of the same values in the config spec that you pass to `srcVM.Clone`? It seems like you'd only need one or the other. I'd call `buildConfigSpec` to build the spec on line 471.

provider/vsphere/internal/vsphereclient/createvm.go
timClicks: Good catch, I think I get it.

timClicks: Switched to using named return values, which should resolve this issue.

provider/vsphere/internal/vsphereclient/client.go
timClicks: Good point. This second config spec isn't needed as we're not dealing with temporary VMs and root disks.

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -277,64 +339,38 @@ func (c *Client) extendVMRootDisk(
 	))
 }
 
-func (c *Client) createImportSpec(
-	ctx context.Context,
+func (c *Client) buildConfigSpec(ctx context.Context,
```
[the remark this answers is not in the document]
timClicks: Done

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -159,87 +177,131 @@ func (c *Client) CreateVirtualMachine(
...
 	}
-	info, err := lease.Wait(ctx, nil)
+	info, err := lease.Wait(ctx, spec.FileItem)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
+
+	updater := lease.StartUpdater(ctx, info)
+	defer updater.Done()
+
```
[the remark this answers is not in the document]
timClicks: Fixed, ty :)

provider/vsphere/internal/vsphereclient/createvm.go
timClicks: Fixed by using the OVA's SHA256 digest in the template name

provider/vsphere/internal/vsphereclient/createvm.go
timClicks: Added a fuller comment in.

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -549,52 +586,10 @@ func (c *Client) extendDisk(
 	return errors.Trace(ErrExtendDisk)
 }
 
-func (c *Client) detachDisk(
```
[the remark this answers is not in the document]
timClicks: Phew!

provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -149,8 +143,32 @@ func (c *Client) CreateVirtualMachine(
...
+	}
+	templateVM, err := finder.VirtualMachine(ctx, vmTemplateName(args.Name))
+	if err == nil {
+		return templateVM, nil
+	}
+	if _, ok := err.(*find.NotFoundError); !ok {
+		return nil, errors.Trace(nil)
+	}
 
 	// Select the datastore.
```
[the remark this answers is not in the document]
timClicks: Deleted

provider/vsphere/internal/vsphereclient/client.go
timClicks: Makes sense, fixed.

timClicks:
Urgh mucked up the rebase.

timClicks:
Changing the base branch resulted in this PR bringing in hundreds of extra commits.

galgalesh:
It seems that this PR broke vSphere support on our cluster.

The PR causes Juju to contact the individual hosts in order to upload images. The individual hosts of our vSphere cluster, however, cannot be contacted by the virtual machines in it (and thus the Juju controller). Only access using the API is allowed. It seems to be a best-practice to firewall off this access. 


More info here: https://discourse.jujucharms.com/t/vsphere-clusters-cant-create-new-machines-after-upgrade-to-2-7-0/2529?u=merlijn-sebrechts

PR #10463, opened 2019-07-25
2.6 fix intermittent login failure

howbazaar:
This branch fixes two different intermittent failures due to the model not being found in the model cache.
This happens when the apiserver code is creating the facade context.

The fix is to ensure that the model exists in the cache before continuing with the test. A helper method has been added to the JujuConnSuite as this is the one that triggers it most.

Additionally the apiserver is now not created in the engine until the model cache has been initialized. This will help avoid login issues to the apiserver if there are many models.

## QA steps

* bootstrap a controller
* deploy something to ensure the components are hooked up

## Documentation changes

No user facing impact.

### DIFF
--- a/juju/testing/conn.go
+++ b/juju/testing/conn.go
@@ -258,19 +258,21 @@ func (s *JujuConnSuite) WaitForModelWatchersIdle(c *gc.C, modelUUID string) {
 // want to do things with those models where the actions may touch
 // the model cache.
 func (s *JujuConnSuite) EnsureCachedModel(c *gc.C, uuid string) {
-	start := time.Now()
+	timeout := time.After(testing.LongWait)
+	retry := time.After(0)
 	for {
-		_, err := s.Controller.Model(uuid)
-		if err == nil {
-			break
-		}
-		if errors.IsNotFound(err) {
-			time.Sleep(testing.ShortWait)
-		} else {
-			c.Errorf("problem getting model from cache: %v", err)
-		}
-		if time.Now().Sub(start) > testing.LongWait {
-			c.Errorf("model %v not seen in cache", uuid)
+		select {
+		case <-retry:
+			_, err := s.Controller.Model(uuid)
+			if err == nil {
+				return
+			}
+			if !errors.IsNotFound(err) {
+				c.Fatalf("problem getting model from cache: %v", err)
+			}
+			retry = time.After(testing.ShortWait)
+		case <-timeout:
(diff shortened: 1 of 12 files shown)


howbazaar:
Address already in use !!build!!

worker/modelcache/worker.go
```diff
@@ -29,8 +29,14 @@ type BackingWatcher interface {
 	Stop() error
 }
 
+// Unlocker is used to indicate that the model cache is ready to be used.
+type Unlocker interface {
+	Unlock()
```
timClicks: Can `Unlock()` fail?

worker/modelcache/worker.go
```diff
@@ -29,8 +29,14 @@ type BackingWatcher interface {
...
+}
+
 // Config describes the necessary fields for NewWorker.
 type Config struct {
+	InitializedGate      Unlocker
```
timClicks: Will tend to stutter as variables are called `gate`.

worker/modelcache/worker.go
howbazaar: No.

worker/modelcache/worker.go
howbazaar: Actually none of the variables are called `gate`. But `initialized` or `unlocker`. The later of which is a bit odd IMO.

juju/testing/conn.go
```diff
@@ -253,6 +253,28 @@ func (s *JujuConnSuite) WaitForModelWatchersIdle(c *gc.C, modelUUID string) {
 	}
 }
 
+// EnsureCachedModel is used to ensure that the model specified is
+// in the model cache. This is used when tests create models and then
+// want to do things with those models where the actions may touch
+// the model cache.
+func (s *JujuConnSuite) EnsureCachedModel(c *gc.C, uuid string) {
+	start := time.Now()
```
jameinel: Is this clearer as a

```
stop := time.After(testing.LongWait)
retry := time.After(0)
for {
   select {
   case <-retry:
    _, err := s.Controller.Model(uuid)
    if err == nil {
     return
   }
   timeout = time.After(testing.ShortWait)
   case <-stop:
    c.Errorf("model %v not seen in cache after %v", uuid, testing.LongWait)
    return
 }
}
```

howbazaar: Yes, I think something with a select is better.

howbazaar:
$$merge$$

PR #10467, opened 2019-07-29
Schema gen changes

SimonRichardson:
## Description of change

The following ensures that all properties get flattened to one struct
because that's how go actually manages them when calling json.

This has huge problems for pylibjuju and in turn will cause lots of
issues when it comes to building a new client. I expect to see a lot
of churn.

## QA steps

Ensure that rebuilding a schema is idempotent.

```
make dep
make rebuild-schema
cat ./apiserver/facades/schema.json | shasum -a 1 | awk '{ print $1 }'
make rebuild-schema
cat ./apiserver/facades/schema.json | shasum -a 1 | awk '{ print $1 }'
```

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -537,5 +537,6 @@
   revision = "4d05e43ad98c1edcf2f52291685bd9bbc12fd2cd"
 
 [[constraint]]
-  revision = "66506e52db051b5e238aa63ec4dad0a790fbc236"
+  revision = "6c0c47f99db1d6789ea71b126fd6951df11de2c4"
   name = "github.com/bcsaller/jsonschema"
+  source = "github.com/SimonRichardson/jsonschema"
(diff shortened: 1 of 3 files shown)


Gopkg.toml
```diff
@@ -537,5 +537,6 @@
   revision = "4d05e43ad98c1edcf2f52291685bd9bbc12fd2cd"
 
 [[constraint]]
-  revision = "66506e52db051b5e238aa63ec4dad0a790fbc236"
+  revision = "6c0c47f99db1d6789ea71b126fd6951df11de2c4"
   name = "github.com/bcsaller/jsonschema"
+  source = "github.com/SimonRichardson/jsonschema"
```
SimonRichardson: We should fork `jsonschema` to `juju`, but there exists one already https://github.com/juju/jsonschema

SimonRichardson:
!!build!!

> fatal: unable to access 'https://github.com/golangci/golangci-lint/': Could not resolve host: github.com

SimonRichardson:
closing this as a merge forward from 2.6 should cover this one.

PR #10471, opened 2019-07-30
Schema gen changes

SimonRichardson:
## Description of change

The following ensures that all properties get flattened to one struct
because that's how go actually manages them when calling json.

This has huge problems for pylibjuju and in turn will cause lots of
issues when it comes to building a new client. I expect to see a lot
of churn.

 - This is the 2.6 version of https://github.com/juju/juju/pull/10467

## QA steps

Ensure that rebuilding a schema is idempotent.

```
make dep
make rebuild-schema
cat ./apiserver/facades/schema.json | shasum -a 1 | awk '{ print $1 }'
make rebuild-schema
cat ./apiserver/facades/schema.json | shasum -a 1 | awk '{ print $1 }'
```


SimonRichardson:
$$merge$$

PR #10410, opened 2019-07-01
Deploy a SAAS from another local controller

SimonRichardson:
## Description of change

The following changes how we consume details for a controller that
is found in the local store, that isn't the same controller as our
own. This is useful for cross controller model relations.

## QA steps

Note: I've skipped over export bundle command in this QA, to make it
easier to test.

```console
juju bootstrap lxd test
juju deploy mysql
juju offer mysql:db
juju status --relations
```

Save the following to a `bundle.yaml`

```yaml
series: trusty
saas:
  mysql:
    url: test:admin/default.mysql
applications:
  wordpress:
    charm: cs:trusty/wordpress-5
    num_units: 1
    to:
    - "0"
machines:
  "0": {}
relations:
- - wordpress:db
  - mysql:db
```

Then do the following:

```console
juju bootstrap lxd other
juju deploy ./bundle.yaml
```

PR #10408, opened 2019-07-01
Export bundle with SAAS block

SimonRichardson:
## Description of change

The following just exports remote applications (SAAS) in the export
bundle command, if you've got a SAAS offering setup.

## QA steps

Setup and deploy mysql

```console
juju bootstrap lxd test
juju deploy mysql
juju offer mysql:db
juju status --relations
```

Create a new model and export the bundle

```console
juju add-model other
juju deploy wordpress
juju add-relation wordpress:db admin/default.mysql
juju export-bundle
```

The following should export the SAAS block with mysql in it.

```yaml
series: trusty
saas:
  mysql:
    url: test:admin/default.mysql
applications:
  wordpress:
    charm: cs:trusty/wordpress-5
    num_units: 1
    to:
    - "0"
machines:
  "0": {}
relations:
- - wordpress:db
  - mysql:db
```

## Documentation changes

Export bundle now includes a SAAS block if you've set on up.

PR #10473, opened 2019-07-31
Revert bundle cmr changes from develop and forward-merge 2.6

achilleasa:
This PR first reverts the cmr for bundles changes (#10408, #10410) that landed on develop and then merges the current 2.6 HEAD which has all the cmr for bundles changes removed via #10474.

This ensures that both 2.6 and develop do not contain any bundle-related changes and allows us to create a separate clean PR to re-introduce the changes just to develop.


SimonRichardson:
$$merge$$

PR #10475, opened 2019-07-31
fix intermittent failures in statusUnitTestSuite.TestMigrationInProgress

hmlanigan:
## Description of change

Only use the model cache model in apiserver client if behind 
the generations feature flag. Causing intermittent failures in 
TestMigrationInProgress, and might cause issues for users too.
Changing use of feature flag after bootstrap is possible, as it's
found in the controller config.

TestMigrationInProgress uses a sync to wait for the model to 
show up in the case, remove Skip().

## QA steps

1. juju bootstrap localhost tuesday
4. juju deploy ubuntu 
3. juju status -- show no branch info
3. juju destroy-controller --destroy-all-models --destroy-storage --yes tuesday
1. export JUJU_DEV_FEATURE_FLAGS=generations
1. juju bootstrap localhost tuesday
4. juju deploy ubuntu 
5. juju add-branch apple
6. juju track apple ubuntu
3. juju status  -- shows the apple branch

### DIFF
--- a/apiserver/facades/client/client/client.go
+++ b/apiserver/facades/client/client/client.go
@@ -11,6 +11,7 @@ import (
 	"github.com/juju/loggo"
 	"github.com/juju/os"
 	"github.com/juju/os/series"
+	"github.com/juju/utils/featureflag"
 	"gopkg.in/juju/names.v2"
 
 	"github.com/juju/juju/apiserver/common"
@@ -27,6 +28,7 @@ import (
 	"github.com/juju/juju/environs/context"
 	"github.com/juju/juju/environs/manual/sshprovisioner"
 	"github.com/juju/juju/environs/manual/winrmprovisioner"
+	"github.com/juju/juju/feature"
 	"github.com/juju/juju/network"
 	"github.com/juju/juju/permission"
 	"github.com/juju/juju/state"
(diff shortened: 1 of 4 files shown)


apiserver/facades/client/client/client.go
```diff
@@ -180,9 +182,20 @@ func newFacade(ctx facade.Context) (*Client, error) {
 		return nil, errors.Trace(err)
 	}
 
-	modelCache, err := ctx.Controller().Model(modelUUID)
-	if err != nil {
-		return nil, errors.Trace(err)
+	var modelCache *cache.Model
+	if featureflag.Enabled(feature.Generations) {
```
babbageclunk: It might be better to make this check against the feature flags in controller config - that way the flag could be toggled without rebootstrapping. See `apiserver/facades/client/cloud/cloud.go` for an example (search for `feature.MultiCloud`).

hmlanigan: Updated and tested.

hmlanigan:
@babbageclunk, yes I tested both ways.  Updated the QA steps to reflect this.

hmlanigan:
amended commit to set feature flag to enable StatusSuite.TestStatusAllFormats() to pass.

apiserver/facades/client/client/status.go
```diff
@@ -806,6 +806,12 @@ func fetchRelations(st Backend) (map[string][]*state.Relation, map[int]*state.Re
 }
 
 func fetchBranches(m *cache.Model) map[string]cache.Branch {
+	// Unless you're using the generations feature flag,
+	// the model cache model will be nil.  See note in
+	// newFacade().
+	if m == nil {
+		return map[string]cache.Branch{}
```
SimonRichardson: tiny nit:

```suggestion
		return make(map[string]cache.Branch)
```

hmlanigan: updated

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #10478, opened 2019-08-01
Expose subnet.IsPublic() and add to subnet migrations

hmlanigan:
## Description of change

Add IsPublic to SubnetInfo{}.  Use when creating a new subnet.  This is prep work.  
Export subnet IsPublic() and include in migrations.

## QA steps

No changes to current juju subnet behavior should be seen.

No provider currently sets the IsPublic flag on a subnet?  The create-subnet command 
behind the "post-net-cli-mvp" feature flag currently panics.  The only way to test 
migration would be to bootstrap a provider with networks.  Edit the db to set 
"is-public" to true on one. Do a juju migration and check the db values again.  Or rely
on the unit test.

### DIFF
--- a/core/network/subnet.go
+++ b/core/network/subnet.go
@@ -57,6 +57,9 @@ type SubnetInfo struct {
 	// It may be empty if this is not a fan subnet,
 	// or if this subnet information comes from a provider.
 	FanInfo *FanCIDRs
+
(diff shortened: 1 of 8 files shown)


hmlanigan:
$$merge$$

PR #10476, opened 2019-08-01
Re-introduce bundle-cmr changes to develop branch

achilleasa:
## Description of change

This PR composes the changes of the following PRs that implement the server and client-side bits for supporting offers/saas blocks in bundles as well as the changes for multi-document bundles (a bundle that contains additional overlay documents following the base bundle):

Originally landed on **2.6**:
- include offers in exported bundles #10353
- apply create offer changes when deploying bundles #10356
- include offer acl in export bundle output #10380
- consume offer #10382
- apply grant offer access changes when deploying bundles #10394
- support for multi-doc bundles and overlays #10460

Originally landed on **develop**:
- export bundle with SAAS block #10408 
- deploy a SAAS from another local controller #10410

All of the above changes where removed from 2.6 (via #10474) and develop (via #10473) so they can be re-introduced just on develop in order to avoid the risk of shipping potential bugs in a 2.6.x release. 

### Expose offers

```console
$ export JUJU_DEV_FEATURE_FLAGS=bundle-cmr
$ juju bootstrap lxd test --no-gui

# Deploy a bundle with an offer
$ juju add-user bar
$ juju deploy ./testcharms/charm-repo/bundle/apache2-with-offers
Resolving charm: cs:apache2-26
Executing changes:
- upload charm cs:apache2-26 for series bionic
- deploy application apache2 on bionic using cs:apache2-26
- create offer my-offer using apache2:apache-website,website-cache
- grant user admin admin access to offer my-offer
- grant user bar consume access to offer my-offer
- create offer my-other-offer using apache2:apache-website
Deploy of bundle completed.

# Verify that export-bundle generates a multi-doc bundle
$ juju export-bundle
series: bionic
applications:
  apache2:
    charm: cs:apache2-26
--- # overlay.yaml
applications:
  apache2:
    offers:
      my-offer:
        endpoints:
        - apache-website
        - website-cache
        acl:
          admin: admin
          bar: consume
      my-other-offer:
        endpoints:
        - apache-website
        acl:
          admin: admin

# Deploy another bundle with overlays
$ juju add-model test1

$ less testcharms/charm-repo/bundle/multi-doc-overlays/bundle.yaml
series: trusty
applications:
  apache2:
    charm: cs:apache2-26
    num_units: 1
--- # overlay 1
applications:
  apache2:
    offers:
      my-offer:
        endpoints:
        - apache-website
        acl:
          admin: admin
--- # overlay 2
applications:
  wordpress:
    charm: cs:trusty/wordpress-5
[...]

PR #10487, opened 2019-08-06
dev - resolve local charms relative to bundle base path

achilleasa:
## Description of change

This PR updates the develop branch to use the latest `charm.v6` version which includes an improved fix for resolving relative charm paths when deploying bundles (compared to the one that landed on the 2.6 branch via #10484).

The main problem with the 2.6 fix is that it was applied against the composed (with all overlays already merged in) bundle and that all relative charm paths were resolved using the base path of the bundle. The 2.6 fix does not work properly if an overlay living at a different location than the bundle specifies a relative charm path.

However, the develop branch uses a different code path (introduced by #10476) and merges overlays via helper functions provided by the `charm.v6` package. The new way of doing things enables us to properly resolve relative paths even for overlays.

Consequently, the original fix has now been removed but its related test has been kept so we can avoid any regressions in the future.

## QA steps

Exactly the same as with #10484 with the exception that the **last** QA step that involves an overlay **must work** as expected with the juju CLI from this PR.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1476,7 +1476,7 @@
   revision = "51fa6e26128d74e445c72d3a91af555151cc3654"
 
 [[projects]]
-  digest = "1:f33ecd7fc165d99a08d28f3559396da2426a35c9877443dadfa230aef2560054"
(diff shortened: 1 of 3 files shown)


achilleasa:
$$merge$$

PR #10496, opened 2019-08-07
Add id to subnetDoc

hmlanigan:
## Description of change

Add ID to subnetDoc.  Change subnetDoc.DocID to be new ID instead of CIDR.  Add related migration code.  Update most closely related tests.  

AddSpace and machine_linklayer code updated to check db for subnet id rather than CIDR.

More related changes are coming.  A number of tests have been marked Skip in the mean time to facilitate commits to the feature branch.

## QA 

bootstrap a cloud with subnet info, maas, openstack... 
Test with juju migrate and juju upgrade.
The subnet docID's in the juju db should be model-uuid:<number> rather than CIDR.

### DIFF
--- a/state/subnets.go
+++ b/state/subnets.go
@@ -4,8 +4,6 @@
 package state
 
 import (
-	"net"
-
 	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 	jujutxn "github.com/juju/txn"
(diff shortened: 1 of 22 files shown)


state/subnets.go
```diff
@@ -174,24 +176,6 @@ func (s *Subnet) ProviderNetworkId() network.Id {
 	return network.Id(s.doc.ProviderNetworkId)
 }
 
-// Validate validates the subnet, checking the CIDR, and VLANTag, if present.
-func (s *Subnet) Validate() error {
```
hmlanigan: After reworking AddSubnet(), this validation fits better on SubnetInfo{}, moved.

hmlanigan:
!!build!!

state/subnets.go
```diff
@@ -327,55 +334,43 @@ func (st *State) AddSubnet(args network.SubnetInfo) (subnet *Subnet, err error)
...
+		}
+		subnet = &Subnet{st: st, doc: subDoc}
+		ops = append(ops, assertModelActiveOp(st.ModelUUID()))
 		return ops, nil
 	}
 	err = st.db().Run(buildTxn)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
+	if err := subnet.setSpace(nil); err != nil {
```
achilleasa: Why do we set this to nil? Does that mean "use the default space"?

achilleasa: oh wait, this is a collection?

hmlanigan: The arg is nil because there isn't a mongo collection available pass.  Some places that call setSpace() already have the subnet collection available, others do not.  Therefore setSpace() will only get the subnet collection if it's not already available.

hmlanigan:
$$merge$$

PR #10497, opened 2019-08-08
Fix CSI provisioner selection and fallback

hpidcock:
## Fix CSI provisioner selection and fallback

- Introduced new dependency "k8s.io/csi-api/..."
- Interface with CSI CRD api to fetch CSIDrivers.
- Allow multiple preferred provisioners per cloud type.
- Filter out unavailable CSI preferred provisioners.
- Fix bug with non-deterministic selection of explicit and default
storage classes.
- Refactor storage class selection logic to be clearer and but more verbose.

## QA steps

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1836077


hpidcock:
!!build!!

PR #10498, opened 2019-08-08
Add LANG to hook execution environment for linux.

howbazaar:
Instead of forcing all charm developers to set the LANG in their scripts, provide a sensible default.

## QA steps

* bootstrap a model
* deploy the ubuntu-lite charm
* juju run --unit ubuntu-lite/0 'echo $LANG'

## Documentation changes

We will want to update the environment variables that are set as part of the hook execution environment, if we actually have this documented.
 
## Bug reference

https://bugs.launchpad.net/juju/+bug/1334482


howbazaar:
$$merge$$

PR #10499, opened 2019-08-08
Remove machine hasvote attribute

wallyworld:
## Description of change

Previously a new controller node entity was added which has the "has vote" and "wants vote" attributes. This PR removes the now defunct "has vote" attribute from the machine entity, which was just mirroring the controller node value.

The second commit removes obsolete and broken code which was added 4 years ago to allow mongo to be upgraded prior to juju 2.x. The code was untested and the facade api calls weren't wired up correctly in once case due to a typo.

## QA steps

```
bootstrap lxd
enable-ha
wait for ha to be set up, watching show-controller and mongo rs.config()
remove machine 0
check show-controller and mongo rs.config()
add a new machine
enable-ha --to X
wait for ha to be set up, watching show-controller and mongo rs.config()
enable-ha -n 5
wait for ha to be set up, watching show-controller and mongo rs.config()
```


wallyworld:
[remove hasvote from machine](https://trello.com/c/SrSZvQ65/813-remove-hasvote-from-machine)

api/highavailability/client.go
```diff
@@ -5,13 +5,10 @@ package highavailability
...
 import (
 	"github.com/juju/errors"
-	"github.com/juju/loggo"
-	"github.com/juju/replicaset"
-
 	"github.com/juju/juju/api/base"
 	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/core/constraints"
-	"github.com/juju/juju/mongo"
+	"github.com/juju/loggo"
```
babbageclunk: This should be up with errors.

core/cache/cachetest/state.go
```diff
@@ -122,7 +122,8 @@ func MachineChange(c *gc.C, modelUUID string, machine *state.Machine) cache.Mach
 		SupportedContainersKnown: scKnown,
 		HardwareCharacteristics:  hwc,
 		CharmProfiles:            chProf,
-		HasVote:                  machine.HasVote(),
+		HasVote:                  true,
```
babbageclunk: Shouldn't this be removed?

wallyworld: We still expose has/wants vote info to api callers via the MachineInfo struct (composed from a few sources of info). The cache entity reflects those values. Maybe we can remove from cache but need to verify first.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

core/cache/cachetest/state.go
manadart: This isn't used AFAIK, but correctness would have it determined the same way as it is in backing state for the multi-watcher. Something like:
```go
	wantsVote := false
	hasVote := false
	node, err := st.ControllerNode(m.Id)
		return errors.Trace(err)
	}
	wantsVote = err == nil && node.WantsVote()
	hasVote = err == nil && node.HasVote()
	...
```

PR #10501, opened 2019-08-08
add me-south-1 ec2 region.  add ec2 product family "compute instance (bare metal)"

hmlanigan:
## Description of change

Add me-south-1 ec2 region.  Add ec2 product family "compute instance (bare metal)"

## QA steps

1. juju regions aws
2. Verify endpoints here: https://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region
3. attempt bootstrap or charm deployment with i3.metal instance type.
juju bootstrap aws/us-east-1 --constraints instance-type="i3.metal"

## Bug reference

https://bugs.launchpad.net/juju/+bug/1839496


mitechie:
QA ok

https://pastebin.canonical.com/p/spt8mGsbk7/

I did hit an issue when I tried to use the us-east-1 that the instance-type was invalid. I checked the console and it showed if I manually launch an instance. I then changed to us-east-2 and it worked. So there's something in that we have poor error messaging if the issue was that aws refused for some reason. e.g. the instance type is valid but maybe none are available?

mitechie:
also checked new region shows up in the regions output

provider/ec2/internal/ec2instancetypes/process_cost_data.go
```diff
@@ -169,7 +169,8 @@ func process(in io.Reader) (map[string][]instanceType, metadata, error) {
 	}
 	for _, sku := range skus.SortedValues() {
 		productInfo := index.Products[sku]
-		if productInfo.ProductFamily != "Compute Instance" {
+		if productInfo.ProductFamily != "Compute Instance" &&
+			productInfo.ProductFamily != "Compute Instance (bare metal)" {
```
SimonRichardson:
```suggestion
			if strings.HasPrefix(productInfo.ProductFamily) != "Compute Instance" {
```

hmlanigan: In general that is a good suggestion.  In this case, I think we want to know when these change to be on the lookout for potential issues moving forward.

hmlanigan:
$$merge$$

PR #10502, opened 2019-08-08
Fix the match to handle slower race tests.

howbazaar:
When the test is running under the race flag, it sometimes takes longer to create the model, so the since value needs to be more permissive than "just now". Took the same approach as the other tests in this suite.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1839566


howbazaar:
$$merge$$

PR #10504, opened 2019-08-09
Increment juju to 2.7-beta-xtian3

jujubot:


babbageclunk:
Testing a change to the release process.

PR #10506, opened 2019-08-09
Fix race in TestUniterRelations

howbazaar:
There is a race in the construction of the remote state watcher that gets tickled by this test.

If the uniter notices that the endpoint has gone away before the 'verifyRunning' step, then there is the potential for an error. If the API call to get the relation succeeds, and then the underlying database object is removed, the follow up call in the uniter to watch the relation will fail with permission denied. The loop handles the case where the permission denied is returned from the first call, but not the second. This is a relatively small window, but we seem to hit this race quite often.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1812111


wallyworld:
$$merge$$

PR #10508, opened 2019-08-09
Generating docs from API facade schema

SimonRichardson:
## Description of change

In order to fully understand the schema you either have to read a lot of go
code or json code. Sometimes the go doc tool doesn't really give you a clear
vision of what's what in terms of the _schema_. So this little go tool can generate
a doc from the **actual** schema, which makes working and understanding the
schema a lot easier to grok.

Run the schemadocs using:

```
go run ./scripts/schemadocs/schemadocs.go ./apiserver/facades/schema.json > output.html
```

See the following screen shot for an idea of looks.

Screenshot from 2019-08-16 16-58-05

Note: the reason this isn't a plugin is because we need the location of the schema and
if you're not a developer of juju it's actually hard to locate where the schema.json is without
trying to download it.


timClicks:
Really cool idea Simon. Makes perfect sense .

SimonRichardson:
Ah good shout, I'll add a make target

SimonRichardson:
$$merge$$

SimonRichardson:
https://trello.com/c/l9EKYNve/1743-schema-docs

PR #10510, opened 2019-08-12
peergrouper watches k8s controller service instead of machine

wallyworld:
## Description of change

Essentially a cleanup branch to prepare for removing machines from k8s models.
The peergrouper watches the k8s controller service (not machine) to get the api addresses.
Also fix cloud service id as it was incorrectly exposing the model uuid.
And remove some unused code.

Bring in new names.v2 dep to get controller agents.
As a driveby, also update the gomaasapi dep.

## QA steps

bootstrap on k8s and lxd
enable-ha on lxd

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -544,11 +544,11 @@
   revision = "e1ad140384f254c82f89450d9a7c8dd38a632838"
 
 [[projects]]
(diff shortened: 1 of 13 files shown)


wallyworld:
$$merge$$

PR #10511, opened 2019-08-12
add-k8s UX cloud/region as positional and other polish.

anastasiamac:
## Description of change

'add-k8s' command is coming out of prototype stage and is becoming  a 1st class citizen in Juju CLI. As such it required some alignment with Juju CLI standard:

* cloud/region can be specified as positional arguments as per 'add-model' and 'bootstrap
 commands precedence. As such options for --cloud and --region has been deprecated. Until Juju 3 makes these options obsolete and they are removed from the codebase, Juju will warn users not to use them and will accept provided values as previously;

* although there were some provisions in the code to cater for additions to a remote store, it was not complete. For example, cloud validation only validated against local clouds not remote ones. Since all this code has already been written for 'cloud' package, this PR re-factors and re-uses common code.

* Since the command's code was written, credential validation has matured but was not used here. This PR adjusts that. As a consequence, some test setups needed tweaking.

As a drive-by, this PR improves validation messages for supplied region for various types of clouds - with regions, without regions, etc.

As a drive-by, improve add-cloud help text - we have originally mentioned that Juju does not support multi-cloud world. This is now in the past. 

## QA steps

Use add-k8s in vigor.

## Bug reference

Most of this work does not have bugs and was discovered by chasing cloud validation when user supplied non-existent cloud - https://bugs.launchpad.net/juju/+bug/1830949

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/juju/cmd"
@@ -16,7 +17,6 @@ import (
 	"github.com/juju/gnuflag"
 	"github.com/juju/loggo"
 	"golang.org/x/crypto/ssh/terminal"
-	"gopkg.in/juju/names.v2"
 
 	cloudapi "github.com/juju/juju/api/cloud"
 	"github.com/juju/juju/caas"
@@ -41,9 +41,12 @@ type CloudMetadataStore interface {
 	WritePersonalCloudMetadata(cloudsMap map[string]jujucloud.Cloud) error
 }
 
-// AddCloudAPI - Implemented by cloudapi.Client.
-type AddCloudAPI interface {
-	AddCloud(jujucloud.Cloud) error
+// CloudAPI - Implemented by cloudapi.Client.
+type CloudAPI interface {
+	common.CloudAPI
+	common.UploadAPI
+
+	// AddCredential uploads credential to the controller
 	AddCredential(tag string, credential jujucloud.Credential) error
 	Close() error
 }
@@ -72,8 +75,9 @@ It's also possible to select a context by name using --context-name.
 
 When running add-k8s the underlying cloud/region hosting the cluster needs to be
 detected to enable storage to be correctly configured. If the cloud/region cannot
-be detected automatically, use --region <cloudType|cloudName>/<someregion> to specify the host
-cloud type and region.
+be detected automatically, user has to specify desired host cloud and/or region
+via a positional argument. If a region is specified without a cloud qualifier, 
+then it is assumed to be in the same cloud as the controller model.
 
 When adding a GKE or AKS cluster, you can use the --gke or --aks option to
 interactively be stepped through the registration process, or you can supply the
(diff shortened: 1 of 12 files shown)


anastasiamac:
[add-k8s UX alignment](https://trello.com/c/nadc13HZ/865-add-k8s-ux-alignment)

cmd/juju/caas/add.go
```diff
@@ -135,12 +137,16 @@ type AddCAASCommand struct {
 	// resourceGroup is the resource group name for the cluster.
 	resourceGroup string
 
-	// hostCloudRegion is the cloud region that the nodes of cluster (k8s) are running in.
-	// The format is <cloudType/region>.
+	// hostCloudRegion stores user specified value of a region option.
```
wallyworld: The original comment is still valid - the point of this var is to record where the cluster nodes are running. We try and query it but require the user to specify if it is unable to be determined automatically.

cmd/juju/caas/add.go
```diff
@@ -199,19 +206,20 @@ func (c *AddCAASCommand) SetFlags(f *gnuflag.FlagSet) {
...
-	f.StringVar(&c.resourceGroup, "resource-group", "", "the Azure resource group of the AKS cluster")
-	f.BoolVar(&c.gke, "gke", false, "used when adding a GKE cluster")
-	f.BoolVar(&c.aks, "aks", false, "used when adding an AKS cluster")
+	f.StringVar(&c.hostCloudRegion, "region", "", "DEPRECATED kubernetes cluster cloud and/or region")
+	f.StringVar(&c.cloud, "cloud", "", "Kubernetes cluster cloud and/or region")
+	f.StringVar(&c.workloadStorage, "storage", "", "Kubernetes storage class for workload storage")
+	f.StringVar(&c.project, "project", "", "Project to which the cluster belongs")
+	f.StringVar(&c.credential, "credential", "", "The credential to use when accessing the cluster")
+	f.StringVar(&c.resourceGroup, "resource-group", "", "The Azure resource group of the AKS cluster")
+	f.BoolVar(&c.gke, "gke", false, "Type of kubernetes cluster to add, here GKE")
```
wallyworld: I prefer the original wording

cmd/juju/caas/add.go
```diff
@@ -223,13 +231,30 @@ func (c *AddCAASCommand) Init(args []string) (err error) {
...
+			if strings.Contains(c.cloud, "/") || strings.Contains(c.hostCloudRegion, "/") {
+				// <cloud>/<region> definitions may conflict.
+				return errors.New("provide either --region or --cloud, not both")
+			}
+			c.cloudRegion = jujucloud.BuildHostCloudRegion(c.cloud, c.hostCloudRegion)
+			goto next
 		}
+		c.cloudRegion = c.hostCloudRegion
+	} else {
+		c.cloudRegion = c.cloud
```
wallyworld: It really is hostCloudRegion rather than cloudRegion as the intent is to record where the host nodes are running.

cmd/juju/caas/add.go
```diff
@@ -363,10 +388,45 @@ var noRecommendedStorageErrMsg = `
 
 // Run is defined on the Command interface.
 func (c *AddCAASCommand) Run(ctx *cmd.Context) (err error) {
+	if c.hostCloudRegion != "" {
```
wallyworld: See previous comment about retaining this attribute name.
Can we just print the deprecation warning when the flags are parsed? Why here as well?

cmd/juju/caas/add.go
```diff
@@ -363,10 +388,45 @@ var noRecommendedStorageErrMsg = `
...
+		ctx.Infof("region flag is DEPRECATED. Use 'add-k8s name cloud/region' instead")
+	}
+	if c.cloud != "" {
+		ctx.Infof("cloud flag is DEPRECATED. Use 'add-k8s name cloud/region' instead")
+	}
 	if err := c.verifyName(c.caasName); err != nil {
 		return errors.Trace(err)
 	}
 
+	// validate provided cloud/region
```
wallyworld: The cloud and region used here to record where the cluster is running does not have to be known to the controller necessarily. We could add a k8s cluster running via CDK on openstack and that openstack does not have to be managed by the same controller as that which we want to use to run the k8s workloads. The cloud and region are used to ensure juju checks the cluster for opinionated storage choices.

cmd/juju/caas/add.go
```diff
@@ -363,10 +388,45 @@ var noRecommendedStorageErrMsg = `
...
+			UseDefaultRegion:  true,
+		}
+		var cloud jujucloud.Cloud
+		var err error
+		_, cloud, cloudRegion, err = common.ParseCloudRegionMaybeDefaultRegion(cloudRegionParams)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		// TODO (anastasiamac 2019-04-24) For some reason everywhere after this point,
+		// this changes from <cloud name>/<region> to <cloud type>/<region>..
```
wallyworld: Technically it is supposed to be cloud type.
But we allow users to specify cloud name as well and that is translated to cloud type

anastasiamac:
Closing in favor of different approach.

PR #10515, opened 2019-08-13
Fix useage of k8s charm deployment override

wallyworld:
## Description of change

k8s charms can specify a deployment type to override juju's default choice.
This was not being honoured due to a typo in the code.
The fix was one line - the rest is to account for adding a new test.

## QA steps

deploy a k8s charm without storage and ensure that the deployment can be forced to be stateful

## Bug reference
https://bugs.launchpad.net/juju/+bug/1839918


wallyworld:
$$merge$$

PR #10518, opened 2019-08-14
Add new Azure South Africa North/West and France South regions

wallyworld:
## Description of change

Add the new Azure South Africa North , South Africa West, and France South regions.
However, instance lookup was broken on several regions. We were using a deprecated API so I had to re-engineer using a different API.

## QA steps

juju bootstrap azure/southafricanorth

## Bug reference

https://bugs.launchpad.net/juju/+bug/1839569


wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10522, opened 2019-08-14
Better error messages

SimonRichardson:
## Description of change

The following changes how Plugin call back is called. Instead of
terminating inside the Plugin callback, we instead allow more of a
middleware approach, that each callback can be composed. This allows
us to do some very useful things with the return values of the
failed callback.

For example we can use the subcommand name to return a helpful
message like:

```
ERROR juju: "controllerss" is not a juju command. See "juju --help".

Did you mean this?
	controllers
```

See: https://github.com/juju/cmd/pull/65

## QA steps

```
juju controllerss
```

Should return:

```
ERROR juju: "controllerss" is not a juju command. See "juju --help".

Did you mean this?
	controllers
```

### DIFF
--- a/cmd/juju/commands/main.go
+++ b/cmd/juju/commands/main.go
@@ -224,12 +224,12 @@ func NewJujuCommand(ctx *cmd.Context) cmd.Command {
 		Doc:  jujuDoc,
 		MissingCallback: RunPlugin(func(ctx *cmd.Context, subcommand string, args []string) error {
 			if cmdName, _, ok := jcmd.FindClosestSubCommand(subcommand); ok {
-				return NotFoundCommand{
+				return &NotFoundCommand{
 					ArgName: subcommand,
 					CmdName: cmdName,
 				}
 			}
-			return &cmd.UnrecognizedCommand{Name: subcommand}
+			return cmd.DefaultUnrecognizedCommand(subcommand)
 		}),
 		UserAliasesFilename: osenv.JujuXDGDataHomePath("aliases"),
 		FlagKnownAs:         "option",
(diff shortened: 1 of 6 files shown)


cmd/juju/commands/main.go
```diff
@@ -218,17 +218,40 @@ func juju2xConfigDataExists() bool {
...
+// could be available if a mistake around the subcommand name is given.
+type NotFoundCommand struct {
+	ArgName string
+	CmdName string
+}
+
+func (c NotFoundCommand) Error() string {
+	return fmt.Sprintf(`juju: %q is not a juju command. See "juju --help".
+
+Did you mean this?
```
mitechie: Let's just do 

    Did you mean:

mitechie:
Looks like legit test fallout that need updating

    13:29:08 test 0: juju help foo doesn't exist

SimonRichardson:
> Looks like legit test fallout that need updating

Yeah, I wasn't sure the fallout, so I got CI to do a full sweep for me. I was expecting it to be worse tbh!

cmd/juju/commands/main.go
```diff
@@ -218,17 +218,40 @@ func juju2xConfigDataExists() bool {
...
-	jcmd := jujucmd.NewSuperCommand(cmd.SuperCommandParams{
-		Name:                "juju",
-		Doc:                 jujuDoc,
-		MissingCallback:     RunPlugin,
+	var jcmd *cmd.SuperCommand
+	jcmd = jujucmd.NewSuperCommand(cmd.SuperCommandParams{
+		Name: "juju",
+		Doc:  jujuDoc,
+		MissingCallback: RunPlugin(func(ctx *cmd.Context, subcommand string, args []string) error {
+			if cmdName, _, ok := jcmd.FindClosestSubCommand(subcommand); ok {
```
howbazaar: Nice. Does this use levenshtein distance to find the closest command?

SimonRichardson: Indeed it does, but we cap the result so that matches with large deviation don't result in large mismatches.

SimonRichardson:
$$merge$$

PR #10523, opened 2019-08-14
Fixes lp#1830949: add-k8s validation for user supplied cloud/region

anastasiamac:
## Description of change

'add-k8s' tries to detect cloud/region. However, for the cases when the command cannot detect the cloud, it relies on user input.

Previously, if the user supplied cloud/region but the command successfully detected what cloud/region to use, user input was ignored and not even validated. 

This PR checks user input when cloud/region is detected and if there is a mismatch, the user is prompted.

As a drive-by:
* Ensure that the command no longer ignores --local
* When cloud region validation fails, we list available regions. However, some clouds may not have regions (see lp#1819409) and displaying empty list is not  helpful. This PR displays a better message.  

## Bug reference

https://bugs.launchpad.net/juju/+bug/1830949
https://bugs.launchpad.net/juju/+bug/1839898

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -480,11 +480,11 @@ func checkCloudRegion(given, detected string) error {
 		if givenRegion != "" || givenCloud != detectedRegion {
 			// If givenRegion is empty, then givenCloud may be a region.
 			// Check that it is not a region.
-			return errors.Errorf("given cloud %q was different to the detected cloud %q: re-run the command without specifying the cloud", givenCloud, detectedCloud)
+			return errors.Errorf("specified cloud %q was different to the detected cloud %q: re-run the command without specifying the cloud", givenCloud, detectedCloud)
 		}
 	}
 	if givenRegion != "" && givenRegion != detectedRegion {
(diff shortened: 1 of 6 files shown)


anastasiamac:
[add-k8s UX alignment](https://trello.com/c/nadc13HZ/865-add-k8s-ux-alignment)

cmd/juju/caas/add.go
```diff
@@ -458,12 +469,32 @@ func (c *AddCAASCommand) Run(ctx *cmd.Context) (err error) {
...
+		// User provided no host cloud/region information.
+		return nil
+	}
+	givenCloud, givenRegion, _ := jujucloud.SplitHostCloudRegion(given)
+	detectedCloud, detectedRegion, _ := jujucloud.SplitHostCloudRegion(detected)
+	if givenCloud != "" && givenCloud != detectedCloud {
+		if givenRegion != "" || givenCloud != detectedRegion {
+			// If givenRegion is empty, then givenCloud may be a region.
+			// Check that it is not a region.
+			return errors.Errorf("given cloud %q was different to the detected cloud %q: re-run the command without specifying the cloud", givenCloud, detectedCloud)
```
wallyworld: s/given/specified I think

anastasiamac:
$$merge$$

PR #10524, opened 2019-08-15
vsphere: Apply constraints at clone time, not template VM import

babbageclunk:
## Description of change

There's no point in building a config spec for the template VM because the constraints we have are the ones for a controller machine, so we need to apply the right ones to reset all of the config when we clone the template VM anyway.

`buildConfigSpec` is now only used from cloning which means it can be simplified a bit.

## QA steps

* Bootstrap a controller on vsphere. Ensure that the controller VM has our normal constraints (3.5G ram), and the template VM doesn't.
* Deploy applications with different memory, cpu and disk constraints - check that the VMs all have the right configuration once they're running.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/provider/vsphere/internal/vsphereclient/client.go
+++ b/provider/vsphere/internal/vsphereclient/client.go
@@ -469,16 +469,10 @@ func (c *Client) cloneVM(
 		args.UpdateProgressInterval,
 	}
 
-	vmConfigSpec := &types.VirtualMachineConfigSpec{}
-	err = c.buildConfigSpec(ctx, args, vmConfigSpec)
+	vmConfigSpec, err := c.buildConfigSpec(ctx, args, srcVM)
 	if err != nil {
 		return nil, errors.Annotate(err, "building clone VM config")
 	}
-	newVAppConfig, err := customiseVAppConfig(ctx, srcVM, args)
-	if err != nil {
-		return nil, errors.Annotate(err, "changing VApp config")
-	}
-	vmConfigSpec.VAppConfig = newVAppConfig
 
 	task, err := srcVM.Clone(ctx, vmFolder, args.Name, types.VirtualMachineCloneSpec{
 		Config: vmConfigSpec,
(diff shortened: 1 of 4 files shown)


babbageclunk:
!!build!!

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -469,16 +469,10 @@ func (c *Client) cloneVM(
 		args.UpdateProgressInterval,
 	}
 
-	vmConfigSpec := &types.VirtualMachineConfigSpec{}
-	err = c.buildConfigSpec(ctx, args, vmConfigSpec)
+	vmConfigSpec, err := c.buildConfigSpec(ctx, args, srcVM)
```
timClicks: Good decision to clean up how this is called.

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -515,6 +509,67 @@ func (c *Client) cloneVM(
...
+		spec.MemoryMB = int64(*args.Constraints.Mem)
+	}
+	if args.Constraints.HasCpuPower() {
+		cpuPower := int64(*args.Constraints.CpuPower)
+		spec.CpuAllocation = &types.ResourceAllocationInfo{
+			Limit:       &cpuPower,
+			Reservation: &cpuPower,
+		}
+	}
+	if spec.Flags == nil {
```
timClicks: Conditional's not required I think (this looks familiar, and I thought that I had removed it earlier - sorry about that)

babbageclunk:
!!build!!

provider/vsphere/internal/vsphereclient/client.go
babbageclunk: Oh right, removed!

babbageclunk:
$$merge$$

babbageclunk:
Merge job couldn't reach archive.ubuntu.com.

$$merge$$

PR #10527, opened 2019-08-15
Build support for slim provider specific binaries

hpidcock:
## Build support for slim provider specific binaries.

This is needed to create a build jobs to output a smaller `juju` binary for microk8s.

- Broke all.go into provider specific units.
- Added build tags to filter which providers are imported.
- Docker image now only supports kubernetes provider.
- Makefile support BUILD_TAGS variable.

## QA steps

Binaries created by `BUILD_TAGS="minimal provider_kubernetes" make install`  should be smaller than `make install`.

## Documentation changes

N/A

## Bug reference

N/A


hpidcock:
$$merge$$

PR #10530, opened 2019-08-15
vsphere: Clean up the VM if we return an error after creating it

babbageclunk:
## Description of change

If an error occurred after cloning the VM we would return the error from CreateVirtualMachine which would leave the created VM around with no way to clean it up. Moved extending the root disk from cloneVM to CreateVirtualMachine so we can just use one cleanup handler.

## QA steps

* Bootstrap a vsphere controller and deploy an application with a root disk bigger than the space available on the target datastore. 
* Check that the cloned VM is removed when the extend error is reported.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/provider/vsphere/internal/vsphereclient/client.go
+++ b/provider/vsphere/internal/vsphereclient/client.go
@@ -5,13 +5,11 @@ package vsphereclient
 
 import (
 	"context"
-	"fmt"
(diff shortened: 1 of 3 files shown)


babbageclunk:
$$merge$$

PR #10533, opened 2019-08-19
Fixes lp#1840205: show default region from client store.

anastasiamac:
## Description of change

By default, Juju will pick the first region for the clouds that have more than one region. Users can change that to suit their needs and preferences using 'juju default-region' command. This will be the region used by all commands - bootstrap, add-model, credential validations, etc - that care about region if the user does not explicitly specify a  region using a --region option at CLI.

Whilst Juju was mostly correctly using the default region when user set it, Juju was not correctly reporting it in 'juju clouds', 'juju show-cloud', etc.

This PR fixes this.

As a drive-by some renamed some variables in tests that were colliding with imported packages names.

## QA steps

1. set default region
2. make sure it is shown correctly where needed

## Bug reference

https://bugs.launchpad.net/juju/+bug/1840205

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -596,8 +596,8 @@ func (c *AddCAASCommand) validateCloudRegion(ctx *cmd.Context, cloudRegion strin
 				return details.CloudType, nil
 			}
 			if region == "" && details.DefaultRegion != "" {
-				logger.Debugf("cloud region not provided, using client default %q", details.DefaultRegion)
-				return jujucloud.BuildHostCloudRegion(details.CloudType, details.DefaultRegion), nil
+				logger.Debugf("cloud region not provided by user, using client default %q", details.DefaultRegion)
+				region = details.DefaultRegion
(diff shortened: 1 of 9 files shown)


anastasiamac:
[default region should be shown and used correctly](https://trello.com/c/j8xPUM8R/885-default-region-should-be-shown-and-used-correctly)

anastasiamac:
!!build!!

cmd/juju/caas/add.go
```diff
@@ -595,6 +595,10 @@ func (c *AddCAASCommand) validateCloudRegion(ctx *cmd.Context, cloudRegion strin
 				}
 				return details.CloudType, nil
 			}
+			if region == "" && details.DefaultRegion != "" {
```
wallyworld: For safety we should still check that details.DefaultRegion is in the details.RegionsMap
ie if region is "", set it to default region can continue as before

anastasiamac:
$$merge$$

PR #10534, opened 2019-08-19
Update README to focus on project benefits

timClicks:
## Description of change

Non-functional change to the project's README. Push all compile/build instructions into CONTIBUTING.md. 

## QA steps

n/a

## Documentation changes

n/a

## Bug reference

-


mitechie:
Thanks for kicking this off. I've got a few comments and so I've started up and shared a gdoc to hash over the material.

README.md
```diff
@@ -1,212 +1,41 @@
...
+## Why Juju?
 
-Juju enables you to use [Charms](https://docs.jujucharms.com/applications-and-charms) to deploy your
-application architectures to EC2, OpenStack, Azure, GCE, your data center, and
-even your own Ubuntu based laptop.  Moving between models is simple giving you
-the flexibility to switch hosts whenever you want — for free.
+* Reduce complexity  
+* Enable repeatability  
+* Codify operations knowledge  
+* Simplify day two
```
jtroup: Suggest 'Simplify day two and beyond' or something similar.  Day two itself is less interesting than days 2-N.

README.md
```diff
@@ -1,212 +1,41 @@
...
 
-Juju enables you to use [Charms](https://docs.jujucharms.com/applications-and-charms) to deploy your
-application architectures to EC2, OpenStack, Azure, GCE, your data center, and
-even your own Ubuntu based laptop.  Moving between models is simple giving you
-the flexibility to switch hosts whenever you want — for free.
+* Reduce complexity  
+* Enable repeatability  
+* Codify operations knowledge  
+* Simplify day two 
+* Maintain portability
```
jtroup: I suspect this'll be confusing to folks not familiar with the idea of substrates.  'Maintain portability' makes me think of Java's "write once, run anywhere" stuff.  Perhaps "Run your applications on anything from bare metal, private cloud to any public cloud" or something?

README.md
timClicks: +1 that's a really helpful suggestion.

We were thinking about heading away from the leading bullets because it seems to be near-impossible to get the wording right.

Do you think that they're useful?

README.md
timClicks: Yes, you're probably right here. Portability isn't portability in the CPU architecture sense. It would also be helpful to have a few keywords to aid searching. 

In an attempt to stick to a 2 or 3 word budget, what about "Cloud agnostic"?

We're also considering having a deep link to a fuller explanation of what we mean by each of the bullets.

CONTRIBUTING.md
```diff
@@ -1,11 +1,15 @@
...
 Contents
 ========
 
-1. [Getting started](#getting-started)
-2. [Dependency management](#dependency-management)
-3. [Code formatting](#code-formatting)
-4. [Workflow](#workflow)
-5. [Community](#community)
+1. [Building Juju](#building-juju)
+2. [Getting started](#getting-started)
```
howbazaar: Looks like we have lost this heading now.

timClicks: Good catch, fixed.

timClicks:
$$merge$$

PR #10539, opened 2019-08-19
application data bag: Add application data to relations in state

babbageclunk:
## Description of change

The state changes to support application-level data bags on relations. Adds `Relation.ApplicationSettings`, `.WatchApplicationSettings` and `.UpdateApplicationSettings`, as well as an upgrade step to create application settings for each endpoint of existing relations.

Part of the implementation of [Application relation data bag](https://docs.google.com/document/d/19Uh-Cy0IU5-9sXSLQjsqBvRvuOi4ZEQ0PRetIBJKhPU/edit)

## QA steps

* Unit tests only - no API to call these yet.

## Documentation changes

We will need charming documentation for application relation data bags when the API and uniter changes are done.

## Bug reference

None

### DIFF
--- a/state/relation.go
+++ b/state/relation.go
@@ -782,8 +782,8 @@ func (change relationSettingsCleanupChange) Prepare(db Database) ([]txn.Op, erro
 
 }
 
-func relationApplicationSettingsKey(id int, ep Endpoint) string {
-	return fmt.Sprintf("%s#%s#%s", relationGlobalScope(id), ep.Role, ep.ApplicationName)
+func relationApplicationSettingsKey(id int, application string) string {
+	return fmt.Sprintf("%s#%s", relationGlobalScope(id), application)
 }
 
 // ApplicationSettings returns the application-level settings for the
@@ -793,7 +793,7 @@ func (r *Relation) ApplicationSettings(app *Application) (map[string]interface{}
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	applicationKey := relationApplicationSettingsKey(r.Id(), ep)
+	applicationKey := relationApplicationSettingsKey(r.Id(), ep.ApplicationName)
 	s, err := readSettings(r.st.db(), settingsC, applicationKey)
 	if err != nil {
 		return nil, errors.Annotatef(err, "relation %q application %q", r.String(), app.Name())
@@ -811,7 +811,7 @@ func (r *Relation) UpdateApplicationSettings(app *Application, token leadership.
 	if err != nil {
 		return errors.Trace(err)
 	}
-	key := relationApplicationSettingsKey(r.Id(), ep)
+	key := relationApplicationSettingsKey(r.Id(), ep.ApplicationName)
 	err = updateLeaderSettings(r.st.db(), token, key, updates)
 	if errors.IsNotFound(err) {
 		return errors.NotFoundf("relation %q application %q", r, app.Name())
@@ -828,7 +828,7 @@ func (r *Relation) WatchApplicationSettings(app *Application) (NotifyWatcher, er
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	key := relationApplicationSettingsKey(r.Id(), ep)
+	key := relationApplicationSettingsKey(r.Id(), ep.ApplicationName)
 	watcher := newEntityWatcher(r.st, settingsC, r.st.docID(key))
 	return watcher, nil
 }
--- a/state/relation.go
+++ b/state/relation.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/juju/juju/core/leadership"
 	"github.com/juju/juju/core/status"
-	mgoutils "github.com/juju/juju/mongo/utils"
 	"github.com/juju/juju/permission"
 )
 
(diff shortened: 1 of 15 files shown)


babbageclunk:
!!build!!

babbageclunk:
Fixing the migration failures. The api/uniter one seems like an intermittent one? (Although the location certainly seems suspicious so I'll keep an eye on it.)

state/relation.go
```diff
@@ -780,3 +781,54 @@ func (change relationSettingsCleanupChange) Prepare(db Database) ([]txn.Op, erro
...
+func (r *Relation) UpdateApplicationSettings(app *Application, token leadership.Token, updates map[string]interface{}) error {
+	// We can calculate the actual update ahead of time; it's not dependent
+	// upon the current state of the document. (*Writing* it should depend
+	// on document state, but that's handled below.)
+	ep, err := r.Endpoint(app.Name())
+	if err != nil {
+		return errors.Trace(err)
+	}
+	key := relationApplicationSettingsKey(r.Id(), ep)
+	err = updateLeaderSettings(r.st.db(), token, key, updates)
```
wallyworld: the other place where we call updateLeadersettings in this PR we first take a copy of the updates

state/relation.go
```diff
@@ -780,3 +781,54 @@ func (change relationSettingsCleanupChange) Prepare(db Database) ([]txn.Op, erro
 	return ops, nil
 
 }
+
+func relationApplicationSettingsKey(id int, ep Endpoint) string {
+	return fmt.Sprintf("%s#%s#%s", relationGlobalScope(id), ep.Role, ep.ApplicationName)
```
wallyworld: i don't think we need ep.Role

state/state.go
```diff
@@ -1105,12 +1105,16 @@ func (st *State) addPeerRelationsOps(applicationname string, peers map[string]ch
...
-		}, createStatusOp(st, relationGlobalScope(relId), relationStatusDoc))
+		ops = append(ops,
+			txn.Op{
+				C:      relationsC,
+				Id:     relDoc.DocID,
+				Assert: txn.DocMissing,
+				Insert: relDoc,
+			},
+			createStatusOp(st, relationGlobalScope(relId), relationStatusDoc),
+			createSettingsOp(settingsC, relationApplicationSettingsKey(relId, eps[0]), nil),
```
wallyworld: instead of nil,
emptySettings := make(map[string]interface{})
createSettingsOp(settingsC, relationApplicationSettingsKey(relId, eps[0]), emptySettings)

state/upgrades.go
```diff
@@ -2543,3 +2543,52 @@ func ReplacePortsDocSubnetIDCIDR(pool *StatePool) (err error) {
...
+		}
+
+		relations, err := st.AllRelations()
+		if err != nil {
+			return errors.Trace(err)
+		}
+
+		var ops []txn.Op
+		for _, rel := range relations {
+			if rel.Life() != Alive {
```
wallyworld: Are we sure that not having settings for Dying relations is ok everywhere? 
Isn't it safer to create always and let the relation remove code get rid of them when the relation is eventually deleted?

state/upgrades_test.go
```diff
@@ -331,19 +331,16 @@ func (s *upgradesSuite) TestStripLocalUserDomainLastConnection(c *gc.C) {
...
 	coll     *mgo.Collection
 	expected []bson.M
+	filter   bson.D
 }
 
 func (s *upgradesSuite) assertUpgradedData(c *gc.C, upgrade func(*StatePool) error, expect ...expectUpgradedData) {
-	s.assertUpgradedDataWithFilter(c, upgrade, nil, expect...)
-}
-
-func (s *upgradesSuite) assertUpgradedDataWithFilter(c *gc.C, upgrade func(*StatePool) error, filter bson.D, expect ...expectUpgradedData) {
```
wallyworld: why remove this? it's a nice helper and the changes require "nil" to be used everywhere and reading the code at the various call points it's no longer "obvious" what's being done

babbageclunk: It had filtering at the wrong level - the filter is dependent on the specific collection the data comes from, but this has one filter and applies it to all of the collections. 
I can re-add it if you want but it seems like having a filter per expectUpgradedData is a more sensible way to do it.

state/relation.go
babbageclunk: That's because it has a `map[string]string` but `updateLeaderSettings` takes a `map[string]interface{}` (for consistency with relation unit settings). It's not copying to avoid mutation, just type conversion.

state/relation.go
babbageclunk: Ok, removed.

state/state.go
babbageclunk: Changed.

state/upgrades.go
babbageclunk: I have this weird feeling that there was a reason that I added it, but I can't remember it now. :( I'll remove it, hopefully whatever the reason was becomes apparent.

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #10543, opened 2019-08-20
Use whitelist when adding clouds.

anastasiamac:
## Description of change

In multi-cloud Juju world, it is easy to add clouds to a bootstrapped controller. However, some clouds are less compatible with each other. This PR introduces a whitelist of the clouds that Juju can support easily out of the box.

With all other cloud combinations, users need to be absolutely sure that they know what they are doing (networks and connectivity needs to be configured outside of Juju) and can use --force option to force cloud addition.

'add-cloud' help also provides whitelist now - https://pastebin.ubuntu.com/p/WXc9rnTv5J/

'add-model' already prevents users from adding models on a clouds that are not yet known to the controller (https://pastebin.ubuntu.com/p/5h496RkqYw/) so there is no need to provide a --force option here.

As a drive-by, this PR renames some variables that have been conflicting with imports names.

## QA steps
1. bootstrap a controller on a non-whitelisted cloud and enable multi-cloud feature for it. I used aws
2. Try adding cloud. You should be stopped -  https://pastebin.ubuntu.com/p/Gc6KRZPp3Z/
3. Using --force allows operation to proceed
4. Bootstrap a controller on a whitelisted cloud and enable multi-cloud feature for it. I used maas
5. Try adding non-whitelisted cloud. You should be stopped. I used lxd

## Bug reference

https://bugs.launchpad.net/juju/+bug/1825889

### DIFF
--- a/api/cloud/cloud_test.go
+++ b/api/cloud/cloud_test.go
@@ -22,9 +22,16 @@ import (
 
 type cloudSuite struct {
 	gitjujutesting.IsolationSuite
+
+	called bool
 }
 
 var _ = gc.Suite(&cloudSuite{})
+
+func (s *cloudSuite) SetUpTest(c *gc.C) {
+	s.IsolationSuite.SetUpTest(c)
+	s.called = false
+}
 
 func (s *cloudSuite) TestCloud(c *gc.C) {
 	apiCaller := basetesting.APICallerFunc(
@@ -48,6 +55,7 @@ func (s *cloudSuite) TestCloud(c *gc.C) {
 					Regions:   []params.CloudRegion{{Name: "nether", Endpoint: "endpoint"}},
 				},
 			})
+			s.called = true
 			return nil
 		},
 	)
@@ -61,6 +69,7 @@ func (s *cloudSuite) TestCloud(c *gc.C) {
 		AuthTypes: []cloud.AuthType{cloud.EmptyAuthType, cloud.UserPassAuthType},
 		Regions:   []cloud.Region{{Name: "nether", Endpoint: "endpoint"}},
 	})
+	c.Assert(s.called, jc.IsTrue)
 }
 
 func (s *cloudSuite) TestClouds(c *gc.C) {
@@ -86,6 +95,7 @@ func (s *cloudSuite) TestClouds(c *gc.C) {
 					Regions:   []params.CloudRegion{{Name: "nether", Endpoint: "endpoint"}},
 				},
 			}
+			s.called = true
 			return nil
 		},
 	)
@@ -105,6 +115,7 @@ func (s *cloudSuite) TestClouds(c *gc.C) {
 			Regions:   []cloud.Region{{Name: "nether", Endpoint: "endpoint"}},
 		},
 	})
+	c.Assert(s.called, jc.IsTrue)
 }
 
 func (s *cloudSuite) TestUserCredentials(c *gc.C) {
@@ -130,6 +141,7 @@ func (s *cloudSuite) TestUserCredentials(c *gc.C) {
 					},
 				}},
 			}
+			s.called = true
 			return nil
 		},
 	)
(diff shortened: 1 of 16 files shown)


anastasiamac:
[whitelist clouds for multi-cloud world](https://trello.com/c/1lPUHL0q/886-whitelist-clouds-for-multi-cloud-world)

cloud/whitelist.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	sort.Strings(sorted)
+	result := []string{}
+	for _, one := range sorted {
+		result = append(result, fmt.Sprintf(" - controller cloud type %q supports %v", one, w.matrix[one].SortedValues()))
+	}
+	return strings.Join(result, "\n")
+}
+
+// Check will err out if 'exiting' controller cloud type is
```
wallyworld: existing

cmd/juju/cloud/add.go
```diff
@@ -112,8 +112,16 @@ do that. If there's only one credential for the cloud it will be
 uploaded to the controller. If the cloud has multiple local credentials
 you can specify which to upload with the --credential option.
 
+When adding clouds to a controller, some clouds are whitelisted and can be easily added:
+%v
+
+Other cloud combinations can only be force added as the user must consider
+network routability and other consideration that are outside of Juju concerns.
```
wallyworld: considerations

apiserver/facades/client/cloud/cloud.go
```diff
@@ -1012,6 +1008,23 @@ func (api *CloudAPI) AddCloud(cloudArgs params.AddCloudArgs) error {
...
+		if err != nil {
+			return errors.Trace(err)
+		}
+		controllerCloud, err := api.backend.Cloud(controllerInfo.CloudName)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		if err := cloud.CurrentWhiteList().Check(controllerCloud.Type, cloudArgs.Cloud.Type); err != nil {
+			if cloudArgs.Force == nil || !*cloudArgs.Force {
+				return errors.Trace(err)
```
wallyworld: We need to return a typed error and have the caller (eg CLI) interpret that error and print the desired message relevant to the call context. Like is done for CodeAlreadyExists. The message can be taylored by the CLI to mention the use of --force etc

cloud/whitelist.go
```diff
@@ -0,0 +1,60 @@
...
+// not compatible with a 'new' cloud type according to this white list.
+func (w *WhiteList) Check(existing, new string) error {
+	if list, ok := w.matrix[existing]; ok {
+		if !list.Contains(new) {
+			return errors.Errorf("cloud type %q is not whitelisted for controller cloud type %q, current whitelist: %v", new, existing, list.SortedValues())
```
wallyworld: "whitelist" might be too technical for some users.
Fine to return this message here though, and the final end user message should be rendered by the caller (see other comment), something like

"adding a cloud of type %q might not function correctly on a controller bootstrapped on a %q cloud; if you really want to do this use --force"

anastasiamac:
!!build!!

api/cloud/cloud_test.go
```diff
@@ -22,10 +22,17 @@ import (
 
 type cloudSuite struct {
 	gitjujutesting.IsolationSuite
+
+	called bool
 }
 
 var _ = gc.Suite(&cloudSuite{})
 
+func (s *cloudSuite) SetUpTest(c *gc.C) {
```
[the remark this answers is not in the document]
anastasiamac: I had to make it as part of test to ensure that I can create a re-usable api caller. We actually use both per test and test wide variables for the 'called' check, so there is no consistency to worry about.

cloud/whitelist.go
anastasiamac: I will have to change the wording here since apiserver call returns error not typed error.

This is fine since the actual whitelist is still visible in command help.

anastasiamac: Actually, params.Error can be returned since it complies with error interface. I'll change the code as per your suggestion. However, even though I like the message suggestion, it'll have to change slightly since we do not have bootstrapped controller's cloud at cmd here.

anastasiamac:
win tests?!

anastasiamac:
!!build!!

anastasiamac:
fatal: unable to access 'https://github.com/nu7hatch/gouuid/': Failed to connect to github.com port 443: Connection timed out

anastasiamac:
!!build!!

anastasiamac:
$$merge$$

PR #10544, opened 2019-08-21
Fix the rewriting of the mongo config on agent start.

howbazaar:
During the 2.6 refactor of the mongo service code, the rewriting of the mongod service config was mistakenly removed.

This branch reinstates the 2.5 logic around the checking of the service. Also adds a test and makes sure that the arguments for the config are output in a deterministic order.

## QA steps

```
juju bootstrap lxd test
juju model-config -m controller logging-config=juju=info:juju.mongo=debug
juju run -m controller --machine 0 'cat /lib/systemd/system/juju-db/juju-db.service'
# Note no --wiredTigerCacheSizeGB flag
juju controller-config mongo-memory-profile=low
# Wait a few seconds
juju run -m controller --machine 0 'cat /lib/systemd/system/juju-db/juju-db.service'
# Note that the flag turned up.
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1840731

### DIFF
--- a/mongo/mongo.go
+++ b/mongo/mongo.go
@@ -588,8 +588,8 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string) 
 		return zeroVersion, errors.Trace(err)
 	}
 	if installed {
-		// Exists does a check against the contents of the service config file.
-		// The return value is true iff the contest is the same.
+		// Exists() does a check against the contents of the service config file.
+		// The return value is true iff the content is the same.
 		exists, err := svc.Exists()
 		if err != nil {
 			return zeroVersion, errors.Trace(err)
--- a/mongo/mongo.go
+++ b/mongo/mongo.go
@@ -582,20 +582,44 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string) 
 		return mongodVersion, nil
 	}
 
-	running, err := svc.Running()
+	// Installed tells us if there exists a service of the right name.
+	installed, err := svc.Installed()
 	if err != nil {
 		return zeroVersion, errors.Trace(err)
 	}
-	if running {
-		return mongodVersion, nil
+	if installed {
+		// Exists does a check against the contents of the service config file.
+		// The return value is true iff the contest is the same.
+		exists, err := svc.Exists()
+		if err != nil {
+			return zeroVersion, errors.Trace(err)
+		}
+		if exists {
+			logger.Debugf("mongo exists as expected")
+			running, err := svc.Running()
+			if err != nil {
+				return zeroVersion, errors.Trace(err)
+			}
+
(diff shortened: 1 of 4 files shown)


mongo/mongo.go
```diff
@@ -582,20 +582,44 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string)
...
+	// Installed tells us if there exists a service of the right name.
+	installed, err := svc.Installed()
 	if err != nil {
 		return zeroVersion, errors.Trace(err)
 	}
-	if running {
-		return mongodVersion, nil
+	if installed {
+		// Exists does a check against the contents of the service config file.
+		// The return value is true iff the contest is the same.
```
jtroup: Typo: s/contest/content/

mongo/mongo.go
```diff
@@ -582,20 +582,44 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string)
...
-	running, err := svc.Running()
+	// Installed tells us if there exists a service of the right name.
+	installed, err := svc.Installed()
 	if err != nil {
 		return zeroVersion, errors.Trace(err)
 	}
-	if running {
-		return mongodVersion, nil
+	if installed {
+		// Exists does a check against the contents of the service config file.
```
jtroup: It might be easier to read as: 'Exists()' rather than just 'Exists'.

mongo/mongo.go
howbazaar: Updated.

mongo/mongo.go
howbazaar: Seems reasonable to me. Updated.

mongo/mongo.go
```diff
@@ -582,20 +582,44 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string)
...
 
+	// We want to write or rewrite the contents of the service.
+	// Stop is a no-op if the service doesn't exist or isn't running.
+	if err := svc.Stop(); err != nil {
+		return zeroVersion, errors.Annotatef(err, "failed to stop mongo")
+	}
 	dbDir := DbDir(args.DataDir)
 	if err := makeJournalDirs(dbDir); err != nil {
-		return zeroVersion, errors.Errorf("error creating journal directories: %v", err)
+		return zeroVersion, fmt.Errorf("error creating journal directories: %v", err)
```
wallyworld: why this change?

howbazaar: Because I copied the block of code from 2.5 branch. Will update.

howbazaar:
$$merge$$

howbazaar:
$$merge$$ intermittent failure, relationunit_test.go:243: relationUnitSuite.TestWatchRelationUnits

PR #10548, opened 2019-08-22
ControllerInfo machineids -> controllerids

wallyworld:
## Description of change

The ControllerInfo entity contains a list of ids representing controller nodes. The mongo doc field is renamed from "machineids" to "controller-ids". The ids were accessed by first loading the controller info doc and getting the ids attribute - this has been abstracted behind a new ControllerIds() method to allow future optimisations. With this PR, separation of controller HA related info from machines is done, and k8s deployments have no machine related info in the model.

## QA steps

bootstrap 2.6 lxd
enable ha
upgrade
check the controller info doc in mongo
enable ha -n 5

bootstrap 2.6 k8s
upgrade
check the controller info doc in mongo

### DIFF
--- a/state/enableha.go
+++ b/state/enableha.go
@@ -34,38 +34,31 @@ func isController(mdoc *machineDoc) bool {
 
 var errControllerNotAllowed = errors.New("controller jobs specified but not allowed")
 
-func (st *State) getVotingMachineCount(info *ControllerInfo) (int, error) {
+func (st *State) getVotingControllerCount() (int, error) {
 	controllerNodesColl, closer := st.db().GetCollection(controllerNodesC)
 	defer closer()
 
-	return controllerNodesColl.Find(
-		bson.M{
-			"_id":        bson.M{"$in": info.MachineIds},
-			"wants-vote": true,
-		},
-	).Count()
+	return controllerNodesColl.Find(bson.M{"wants-vote": true}).Count()
 }
 
 // maintainControllersOps returns a set of operations that will maintain
-// the controller information when the given machine documents
-// are added to the machines collection. If currentInfo is nil,
-// there can be only one machine document and it must have
-// id 0 (this is a special case to allow adding the bootstrap machine)
-func (st *State) maintainControllersOps(newIds []string, currentInfo *ControllerInfo) ([]txn.Op, error) {
+// the controller information when controllers with the given ids
+// are added. If bootstrapOnly is true, there can be only one id = 0;
+// (this is a special case to allow adding the bootstrap node).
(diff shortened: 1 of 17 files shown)


state/upgrades.go
```diff
@@ -2266,10 +2266,15 @@ func AddControllerNodeDocs(pool *StatePool) error {
 			Insert: doc,
 		})
 	}
-	if len(ops) > 0 {
-		return errors.Trace(st.runRawTransaction(ops))
-	}
-	return nil
+
+	ops = append(ops, txn.Op{
```
babbageclunk: This will always run the update, even if there's nothing to do - is it idempotent? I guess it probably is since if the machineids field is renamed then it does nothing.

state/enableha.go
```diff
@@ -34,51 +34,48 @@ func isController(mdoc *machineDoc) bool {
...
+func (st *State) maintainControllersOps(newIds []string, bootstrapOnly bool) ([]txn.Op, error) {
 	if len(newIds) == 0 {
 		return nil, nil
 	}
-	if currentInfo == nil {
+	currentControllerIds, err := st.ControllerIds()
+	if !bootstrapOnly && errors.IsNotFound(err) {
+		return nil, errors.Trace(err)
+	}
+	if err != nil {
```
babbageclunk: Won't this raise an error if `bootstrapOnly` is true and `errors.IsNotFound(err)`? It seems like it should continue in that case?

state/upgrades.go
wallyworld: yup, is idempotent, and the test runs the step twice

state/enableha.go
```diff
@@ -34,51 +34,48 @@ func isController(mdoc *machineDoc) bool {
...
-			"_id":        bson.M{"$in": info.MachineIds},
-			"wants-vote": true,
-		},
-	).Count()
+	return controllerNodesColl.Find(bson.M{"wants-vote": true}).Count()
 }
 
 // maintainControllersOps returns a set of operations that will maintain
 // the controller information when the given machine documents
 // are added to the machines collection. If currentInfo is nil,
```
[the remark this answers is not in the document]
wallyworld: derp

state/enableha.go
wallyworld: turns out i didn't need the error check as the controller doc is always there

wallyworld:
$$merge$$

wallyworld:
[rename controller info "machineids"](https://trello.com/c/IWrqWLTU/882-rename-controller-info-machineids)

PR #10553, opened 2019-08-23
[dev] refactor installation of lxd, kvm and mongo deps

achilleasa:
## Description of change

This PR originally began as a tentative fix for the attached bug. The bug manifests itself when trying to create an lxd container on a host running **cosmic** or **disco** that does not come with lxd pre-installed. While all disco images seem to come with lxd pre-installed (as a snap vs an apt pkg for bionic and earlier), this is not the case with images for the s390 arch.  By default, we try to install lxd via **apt**. 

The apt repos contain a transition package that actually performs `snap install lxd`. However, the package which is currently available cannot be installed in non-interactive mode: it displays a prompt for the operator to select which lxd version to install. Unfortunately, the prompt cannot be bypassed causing apt-get to block forever.

The solution for this issue is to install via snap on cosmic and onwards which essentially means 
we need to resolve to series-sniffing to decide which package manager (and package names) to use (the juju/packaging repo now includes a snap-based package manager via https://github.com/juju/packaging/pull/4 and https://github.com/juju/packaging/pull/5)

However, as I went through the code base I saw that we use the same pattern in more places:
- to install mongo dependencies when bootstrapping controllers
- to install kvm dependencies. The latter only works on ubuntu and only happens to work across different series because the required package names have not been renamed.

Evidently, the old approach where we first consulted the series to select a list of packages and then asked the `juju/packaging` repo to give us a suitable package manager cannot cover systems where multiple package managers can co-exist (e.g. apt/snap). Additionally, there are cases where we would want to install some packages from snap and others from snap (i.e. they are not available as snaps yet, we don't want them to auto-update etc.).

The solution proposed by this PR is to create an abstraction for a `Dependency`. Each dependency contains the package-selection logic for the series that we support and can be queried to obtain a list of packages **and** preferred package manager for installing each one. I have isolated the dependency requirements for lxd, kvm and mongo into standalone objects that live in the `packaging/dependency` juju pkg. 

The `packaging` package provides a helper for installing dependencies and handles all the low-level details for instantiating the appropriate package managers and injecting additional options (e.g. backports, cloud-tools) where required.

Finally, as an application of the boy-scout rule, I noticed that the `mongo` package contains centos-specific code (and tests) for bootstrapping purposes. Given that we only support bootstrapping on ubuntu (centos is fine for running workloads though), that code is never going to be used so it has been removed. As an extra layer of safety, the dependency object for mongo will return a `NotImplemented` error if asked to resolve the mongo packages on centos.

**Notes:**
To reduce the scope of changes for this PR, I have intentionally left out the changes required to configure proxies for systems with dual package managers (e.g. snap/apt on disco). This will be addressed via a separate PR

## QA steps

To simulate an image where lxd is not pre-installed, we will add two machines (based on bionic and disco), uninstall lxd and then try to add lxd containers on each. This should force a re-install of lxd using he appropriate package manager for each distro.

```console
$ juju add-machine --series=bionic
$ juju add-machine --series=disco

# Wait for the machines to become available and run the following commands
$ juju run --machine 0 'sudo apt-get --option=Dpkg::Options::=--force-confold --option=Dpkg::options::=--force-unsafe-io --assume-yes --quiet purge lxd'
...
Removing lxd (3.0.3-0ubuntu1~18.04.1) ...
...

$ juju run --machine 1 'sudo snap remove lxd'
lxd removed

# Now create two lxd containers and check that juju re-install lxd using the correct package manager
$ juju add-machine lxd:0
created container 0/lxd/0

$ juju add-machine lxd:1
created container 1/lxd/0
 
# Wait for containers to become available and check lxd installations on host machines
$ juju run --machine 0 'which lxd'
/usr/bin/lxd       <-- apt version on bionic host

$ juju run --machine 1 'which lxd'
/snap/bin/lxd      <-- snap version on disco host
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1835187

### DIFF
--- a/container/lxd/initialisation_test.go
+++ b/container/lxd/initialisation_test.go
@@ -164,6 +164,21 @@ func (s *InitialiserSuite) TestNoSeriesPackages(c *gc.C) {
 	})
 }
 
+func (s *InitialiserSuite) TestInstallViaSnap(c *gc.C) {
+	PatchLXDViaSnap(s, false)
+
+	PatchHostSeries(s, "disco")
+
+	paccmder := commands.NewSnapPackageCommander()
+
+	err := NewContainerInitialiser().Initialise()
+	c.Assert(err, jc.ErrorIsNil)
+
+	c.Assert(s.calledCmds, gc.DeepEquals, []string{
+		paccmder.InstallCmd("--classic lxd"),
+	})
+}
+
 func (s *InitialiserSuite) TestLXDInitBionic(c *gc.C) {
 	s.patchDF100GB()
 	PatchHostSeries(s, "bionic")
--- a/mongo/mongo_test.go
+++ b/mongo/mongo_test.go
@@ -11,8 +11,6 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
-	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/juju/errors"
@@ -22,7 +20,6 @@ import (
 	jc "github.com/juju/testing/checkers"
 	gc "gopkg.in/check.v1"
 
-	"github.com/juju/juju/controller"
 	"github.com/juju/juju/core/network"
 	"github.com/juju/juju/mongo"
 	"github.com/juju/juju/service/common"
@@ -54,46 +51,14 @@ var testInfo = struct {
 }
 
 var expectedArgs = struct {
-	MongoInstall []jc.SimpleMessage
-	YumBase      []string
-	AptGetBase   []string
-	Semanage     []string
-	Chcon        []string
+	AptGetBase []string
 }{
-	MongoInstall: []jc.SimpleMessage{
-		{loggo.INFO, "Ensuring mongo server is running; data directory.*"},
-		{loggo.INFO, regexp.QuoteMeta(`installing "epel-release" via "yum"`)},
-		{loggo.INFO, "Running: yum --assumeyes --debuglevel=1 install epel-release"},
-		{loggo.INFO, regexp.QuoteMeta(`installing "mongodb-server" via "yum"`)},
-		{loggo.INFO, "Running: yum --assumeyes --debuglevel=1 install mongodb-server"},
-	},
-	YumBase: []string{
-		"--assumeyes",
-		"--debuglevel=1",
-		"install",
-	},
 	AptGetBase: []string{
 		"--option=Dpkg::Options::=--force-confold",
 		"--option=Dpkg::Options::=--force-unsafe-io",
 		"--assume-yes",
 		"--quiet",
 		"install",
-	},
-	Semanage: []string{
-		"port",
-		"-a",
-		"-t",
-		"mongod_port_t",
-		"-p",
-		"tcp",
-		strconv.Itoa(controller.DefaultStatePort),
-	},
-	Chcon: []string{
-		"-R",
-		"-v",
-		"-t",
-		"mongod_var_lib_t",
-		"/var/lib/juju/",
 	},
 }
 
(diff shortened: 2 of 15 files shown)


container/lxd/initialisation_test.go
```diff
@@ -164,6 +164,22 @@ func (s *InitialiserSuite) TestNoSeriesPackages(c *gc.C) {
 	})
 }
 
+func (s *InitialiserSuite) TestInstallViaSnap(c *gc.C) {
+	PatchLXDViaSnap(s, false)
+
+	// Cosmic and beyond use snap as the default package-manager
```
jameinel: "snap as the default package-manager for lxd" ?
I would be surprised if it was the "default" package manager for everything.

mongo/mongo_test.go
```diff
@@ -385,204 +350,43 @@ func (s *MongoSuite) testEnsureServerNUMACtl(c *gc.C, setNUMAPolicy bool) string
...
+		pkgs    []string
+		expOpts []string
 	}
 
 	tests := []installs{
-		{"precise", "mongodb-server"},
-		{"trusty", "juju-mongodb"},
-		{"xenial", "juju-mongodb3.2"},
-		{"bionic", "mongodb-server-core"},
+		{"precise", []string{"mongodb-server"}, []string{"--target-release", "precise-updates/cloud-tools"}},
```
jameinel: We probably don't need to test for precise anymore, we could arguably even drop trusty for 2.7.

packaging/dependency/kvm.go
```diff
@@ -0,0 +1,47 @@
...
+// PackageList implements packaging.Dependency.
+func (dep kvmDependency) PackageList(series string) ([]packaging.Package, error) {
+	var pkgList []string
+
+	switch series {
+	case "centos7", "opensuseleap":
+		return nil, errors.NotImplementedf("installing kvm on %s", series)
+	default:
+		if dep.arch == arch.ARM64 {
+			// ARM64 doesn't support legacy BIOS so it requires Extensible Firmware
```
jameinel: I wonder if we should always install efi, it seems relevant for being forward looking. But not something we should change as part of this patch.

packaging/dependency/lxd.go
```diff
@@ -0,0 +1,36 @@
...
+type lxdDependency struct{}
+
+// PackageList implements packaging.Dependency.
+func (lxdDependency) PackageList(series string) ([]packaging.Package, error) {
+	var pkg packaging.Package
+
+	switch series {
+	case "centos7", "opensuseleap":
+		return nil, errors.NotImplementedf("installing lxd on %s", series)
+	case "precice", "trusty", "xenial", "zesty", "artful",
```
jameinel: misspelled "precise" but we probably don't need it anyway.
we don't need zesty or artful either as those are definitely not support.d

packaging/dependency/lxd.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"github.com/juju/errors"
+	"github.com/juju/juju/packaging"
+)
+
+const blankSeries = ""
+
+// LXD is a dependency instance for installing lxd.
+var LXD packaging.Dependency = lxdDependency{}
```
jameinel: Is it better to have these as declared, or would it be better to have this as a function like we do for Mongo? I feel like it would be good to be consistent.

packaging/dependency/mongo.go
```diff
@@ -0,0 +1,45 @@
...
+	var pm = packaging.AptPackageManager
+
+	if dep.useNUMA {
+		pkgList = append(pkgList, "numactl")
+	}
+
+	switch series {
+	case "centos7", "opensuseleap":
+		return nil, errors.NotImplementedf("installing mongo on %s", series)
+	case "precise":
```
jameinel: similarly drop 'precise' and 'zesty' 'artful', I don't know if we need an explicit 'disco' but I don't think we do.

packaging/manager.go
```diff
@@ -0,0 +1,148 @@
...
+
+var logger = loggo.GetLogger("juju.packaging")
+
+// PackageManagerName describes a package manager.
+type PackageManagerName string
+
+// The list of supported package managers.
+const (
+	AptPackageManager    PackageManagerName = "apt"
+	YumPackageManager                       = "yum"
```
jameinel: Do they all need PackageManagerName, aren't the later consts just 'string' types?

packaging/manager.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		if pkgConfer != nil {
+			if config.SeriesRequiresCloudArchiveTools(series) &&
+				pkgConfer.IsCloudArchivePackage(pkg.Name) {
+				extraOpts := pkgConfer.ApplyCloudArchiveTarget(pkg.Name)
+				pkg.appendInstallOptions(extraOpts[:len(extraOpts)-1]...)
+			}
+
+			if config.RequiresBackports(series, pkg.Name) {
```
jameinel: This should be moved into the Dependency, shouldn't it?

packaging/dependency/lxd.go
```diff
@@ -0,0 +1,36 @@
...
+
+type lxdDependency struct{}
+
+// PackageList implements packaging.Dependency.
+func (lxdDependency) PackageList(series string) ([]packaging.Package, error) {
+	var pkg packaging.Package
+
+	switch series {
+	case "centos7", "opensuseleap":
+		return nil, errors.NotImplementedf("installing lxd on %s", series)
```
jameinel: With all the abstractions around what package managers we use, we don't actually install anything but Ubuntu? (at least I can't track an install path from Dependency to, eg, Zypper). How is that?

container/lxd/initialisation_test.go
achilleasa: Whops. Typo from the previous iteration...

packaging/dependency/lxd.go
achilleasa: I will remove these.

packaging/dependency/lxd.go
achilleasa: No objection there. I will change this to match the others.

packaging/manager.go
achilleasa: I prefer typed strings as a hint to anyone working with the dependency code that they need to use one of the pre-defined package manager names instead of returning a plain string which could cause issues if misspelled or if we decide to change the package manager name (e.g. apt -> apt-get).

If you think this is overkill, I don't mind changing it to use strings instead.

packaging/manager.go
achilleasa: Well... the current approach that we already use is a bit ugly. These config bits live in the juju/packaging repo and work on a per-package basis (see: https://github.com/juju/packaging/blob/aa6c1d0ba66b1630ed07f0345039fe13ba11f28a/config/ubuntu_constants.go#L90 for an example).

As the mongo/lxd/kvm install code included calls to the configurators it made sense to extract the logic and add it to a single place (the installer) instead of repeating it for each dependency.

Ideally, we should refactor (even better: rewrite) the configuration stuff but I didn't want to expand the scope of this PR further.

packaging/dependency/lxd.go
achilleasa: Given that the recommended install procedure for lxd on centos7 is basically "install snap and snap install lxd" I would be surprised if that ever worked for non-ubuntu machines...

So just in case we ever do reach this code in the future (I think you can do that if you add-machine --series centos7 and juju add-machine lxd:X) I added the not implemented bit so the error makes more sense to the user.

packaging/dependency/lxd.go
achilleasa: Should I move "precise" to L23 next to centos so we get a not-implemented error?

achilleasa: Perhaps we need to add something like this for precise (and remove the [existing](https://github.com/juju/juju/blob/develop/container/lxd/initialisation_linux.go#L264-L266) check from ensureDependencies)

```go
case "precise":
     return errors.NotSupportedf("LXD containers on series %q", series)
```

packaging/dependency/mongo.go
achilleasa: I don't think we need a 'disco' one we should hit the default block for everything else. If at some point we decide to use snaps for mongo we can add a case for bionic to use apt and make the default block use snap.

packaging/dependency/lxd.go
achilleasa: I have updated this to return NotSupported for centos/opensuseleap/precise. Let me know if that's fine or whether I should remove precise altogether.

packaging/dependency/lxd.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+type lxdDependency struct{}
+
+// PackageList implements packaging.Dependency.
+func (lxdDependency) PackageList(series string) ([]packaging.Package, error) {
+	var pkg packaging.Package
+
+	switch series {
+	case "centos7", "opensuseleap", "precise":
```
SimonRichardson: I'm unsure if this should be done here, esp. considering if we have `os/series`?

SimonRichardson: Essentially I'm unsure if "precise" should be here!

packaging/manager_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (s *DependencyManagerTestSuite) TestInstallWithAptOnBionic(c *gc.C) {
+	s.assertInstallCallsCorrectBinary(c, assertParams{
+		series:       "bionic",
+		pkg:          "lxd",
+		pm:           packaging.AptPackageManager,
+		expPkgBinary: "apt-get",
+		expArgs: []string{
+			"--option=Dpkg::Options::=--force-confold",
+			"--option=Dpkg::options::=--force-unsafe-io",
```
SimonRichardson: I know you mentioned this last week, but is `Dpkg::Options::` case insensitive?

packaging/manager.go
jameinel: I actually agree, my point was YumPackageManager is just a 'string' type, isn't it?
https://play.golang.org/p/G1pQnXnO22q

packaging/manager.go
jameinel: I believe this is only used for Precise, which makes me think we can just drop it altogether.

packaging/dependency/lxd.go
jameinel: I'm asking, "are there any packages we *actually* install from a package manager outside of Ubuntu"? All the Dependencies that you implemented seem to only support Ubuntu, were there others that you just didn't get to yet?

packaging/dependency/lxd.go
achilleasa: It seems to me that the pattern we are currently using is specify cases for exceptions and default for everything else. If we did it the other way round, we would need to patch this code for every new ubuntu release, right? While it would still break if you pass centos8 you should still be able to bootstrap something like eoan 

I *could* make the switch work with string.HasPrefix() to catch centosX stuff though...

packaging/manager.go
achilleasa: Doh!!! I somehow thought that they would inherit the type like when using iota...
But obviously that's not the case in general. Nice catch! 

I would have expected the compiler to catch this when assigning to the Package field instead of auto-casting because under the hood it's the same type: https://play.golang.org/p/punXuLV3eAC

packaging/dependency/lxd.go
achilleasa: We install series/os-specific packages in the cloud-init stuff (e.g. https://github.com/juju/juju/blob/a5ab92ec9b7f5da3678d9ac603fe52d45af24412/cloudconfig/cloudinit/cloudinit_centos.go#L34). We could create a dependency for those as well.

jameinel:
I don't believe we need cloud-utils since Juju 1.* because it was used for caching lxc images. We should certainly try pulling them out and see if everything continues to work as expected, but it should be decoupled from this work.

packaging/manager.go
achilleasa: Ok then, I will update the PR and remove that.

achilleasa: Removed the cloud tools check which are precise-only but kept the backports check as it is used by trusty. Once we stop supporting trusty I can revisit this and remove the configuration bits altogether.

packaging/dependency/lxd.go
jameinel: we do, but having that done as part of when we make a release that "now we support eoan" would be a reasonable time to do so. As you say, we haven't done it so far, so I don't think we have to right now. I was just contemplating what makes the most sense from a sanity check pov.

packaging/dependency/mongo.go
jameinel: I would follow the same pattern as above so it is predictable. If we decided to do explicit disco/eon then we'd want to do it in both places.

packaging/manager_test.go
jameinel: we should make it consistent, at least

achilleasa: That comes from juju/packaging. I will submit a patch to nomalize this and update this PR to use the patched dependency.

achilleasa:
$$merge$$

PR #10554, opened 2019-08-23
 Validation of series during bootstrap

SimonRichardson:
## Description of change

The following PR updates how bootstrap and deploying a series is
validated. When bootstrapping a controller we can guarantee that
we would ever only want to bootstrap to a ubuntu release (bionic,
disco, eoan, etc) and for deploying a charm or adding a machine,
we allow for a wider spread of series (ubuntu releases, centos, 
windows, etc) variations.

The change in this PR no longer reads in the distro-info csv file
from the ubuntu host to update the release window, instead we hard
code the release per juju. Each juju is now versioned locked to the
release that juju has been coded to. This is so we can provide better
support for those versions if we know that 2.7.x has been coded to
and tested to bionic, disco and cosmic (even though the later is EOL).

Most of the mechanical changes in the PR are to ensure that the
version locking nature of bootstrapping and deploying are a fixed
constant, to hopefully prevent issues when a new release is updated
in os/series package.

The introduction of force at bootstrap allows operators to override
what bootstrap series are on offer, but if no agent binaries are found
no amount of forcing will allow you to deploy a series we don't have
binaries for. The use of force in this scenario is for past and future
releases we have to offer (precise or eoan for example).

#### Scenarios that should just work without errors:

```shell
juju bootstrap lxd test
juju bootstrap lxd test --config default-series=bionic
juju bootstrap lxd test --bootstrap-series=bionic
```

#### Scenarios that force should be used:

```shell
juju bootstrap lxd test --bootstrap-series=precise --force
```

#### Scenarios that will fail:

```shell
juju bootstrap lxd test --bootstrap-series=spock
juju bootstrap lxd test --bootstrap-series=spock --force
```

#### Special case:

Once the binaries have been built for the eoan release, then the
following should also work without errors.

```shell
juju bootstrap lxd test --bootstrap-series=eoan --force
```

### Deployment

Deployment of charms should work the same way, but cosmic
should no longer be supported as it's now EOL.

```shell
juju deploy --series=cosmic
```

Using force should ensure it's allowed though

```shell
juju deploy --series=cosmic --force
```


Gopkg.toml
```diff
@@ -90,8 +90,9 @@
   name = "github.com/juju/naturalsort"
 
 [[constraint]]
-  revision = "88a4c6ac59c1950f118f058d920204c9182b9927"
+  branch = "workload-crtl-series"
   name = "github.com/juju/os"
+  source = "github.com/SimonRichardson/os"
```
SimonRichardson: This is only used during development and will alter once everything has landed and settled.

cmd/juju/application/deploy.go
```diff
@@ -1702,16 +1702,6 @@ func (c *DeployCommand) validateCharmSeriesWithName(series, name string) error {
...
-			// output the warning information to help track down the error.
-			// see: lp:1833763
-			if warningInfo := series.SeriesWarningInfo(charmSeries); len(warningInfo) > 0 {
-				logger.Warningf(
-					"Parsing issues occurred when extracting the distro-info.\n\n"+
-						"  - %s\n"+
-						"\nYou may want to try updating your distro-info using `apt-get update distro-info`\n",
-					strings.Join(warningInfo, "\n  - "),
-				)
-			}
```
SimonRichardson: As we never parse the distro-info for dates any more, we don't need the warning info.

SimonRichardson:
https://trello.com/c/Z6Fb2NQh/1762-validate-bootstrap-series-when-attempting-to-bootstrap

juju/testing/conn.go
```diff
@@ -67,7 +67,16 @@ import (
 	jujuversion "github.com/juju/juju/version"
 )
 
-const ControllerName = "kontroll"
+const (
+	ControllerName = "kontroll"
+
+	kubernetesSeriesName = "kubernetes"
```
achilleasa: I wonder if we could get this const from juju/os; maybe something like a `(Must)GetOSFromSeries("kubernetes")`

provider/common/bootstrap_test.go
```diff
@@ -189,19 +193,204 @@ func (s *BootstrapSuite) TestBootstrapSeries(c *gc.C) {
...
+		*instance.HardwareCharacteristics,
+		[]network.InterfaceInfo,
+		error,
+	) {
+		return &mockInstance{id: checkInstanceId}, &checkHardware, nil, nil
+	}
+	var mocksConfig = minimalConfig(c)
+	var numGetConfigCalled int
+	getConfig := func() *config.Config {
+		numGetConfigCalled++
```
achilleasa: This does not seem to be checked anywhere inside the test.

On a side-note, why do we need to set up so many things to get back an invalid series error which normally would be an early check?

provider/common/bootstrap_test.go
```diff
@@ -189,19 +193,204 @@ func (s *BootstrapSuite) TestBootstrapSeries(c *gc.C) {
...
```
achilleasa: Same comment here; numGetConfigCalled is not checked

provider/common/bootstrap_test.go
```diff
@@ -189,19 +193,204 @@ func (s *BootstrapSuite) TestBootstrapSeries(c *gc.C) {
...
+		[]network.InterfaceInfo,
+		error,
+	) {
+		return &mockInstance{id: checkInstanceId}, &checkHardware, nil, nil
+	}
+	// We want an invalid fallback to trigger the not valid bootstrap series.
+	var mocksConfig = minimalConfigWithSeries(c, "")
+	var numGetConfigCalled int
+	getConfig := func() *config.Config {
+		numGetConfigCalled++
```
achilleasa: Same comment as above

achilleasa:
All QA steps except `juju bootstrap lxd test --bootstrap-series=precise --force` (which hangs while cloud-init gets stuck trying to locate ubuntu-fan). Is this expected?

SimonRichardson:
@achilleasa yes sorry, was meant to mention that :+1:

provider/common/bootstrap_test.go
SimonRichardson: I'll fix. ta

juju/testing/conn.go
SimonRichardson: Ah, yes, let's make that change. Would clean up a lot of things.

provider/common/bootstrap_test.go
SimonRichardson: I'll see if I can pair it down.

achilleasa:
Also, is it possible to catch the unknown series error earlier so we don't ask users to try again with ---force just to get the not supported error?

```
juju bootstrap lxd test --bootstrap-series=spock
juju bootstrap lxd test --bootstrap-series=spock --force
```

SimonRichardson:
@achilleasa yeah, it should be actually, let me look.

SimonRichardson:
$$merge$$

PR #10556, opened 2019-08-26
Ensure that the cached model exists during facade creation.

howbazaar:
More of the facades require the cached model to exist. Instead of pushing the necessity for each facade to call the CachedModel method, have this done further up the stack. This addresses one of the issues with model migration validation for unit agents, where the model is imported and then the units are asked to validate, but the uniter facade fails looking for the model from the cache.

The pubsub integration test has outlived its usefulness. With the raft leadership using it as a core component, we will notice if something goes wrong elsewhere. 

The alternative to removing it is to move it into the feature tests and use a JujuConnSuite. Which is why I leaned on the side of just removing it.

## QA steps

*  unit tests keep passing

### DIFF
--- a/api/pubsub/pubsub_test.go
+++ b/api/pubsub/pubsub_test.go
@@ -9,21 +9,12 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/juju/loggo"
-	"github.com/juju/pubsub"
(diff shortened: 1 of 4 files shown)


howbazaar:
$$merge$$

PR #10557, opened 2019-08-26
Fix upgrade CI tests

wallyworld:
## Description of change

The upgrade CI tests were generating agent metadata for "proposed" but not setting agent-stream or agent-version when requesting an upgrade. It should be "devel" and we need to specify the stream and agent version.

## QA steps

./assess upgrade


wallyworld:
$$merge$$

PR #10560, opened 2019-08-26
Get a test FAIL if the package doesn't build.

jameinel:
It seems that when running `go test ./...`
if a subdirectory fails to build, but we have no tests in
that directory the failure we get is:
```
./schemadocs.go:294:10: undefined: strings.ReplaceAll
```

However there isn't anything that clearly indicates a failure. By adding
a dummy Test function we get:
```
./schemadocs.go:294:10: undefined: strings.ReplaceAll
FAIL    github.com/juju/juju/scripts/schemadocs [build failed]
```

Which at least means we can find the issue in the test run output.

### DIFF
--- a/scripts/schemadocs/package_test.go
+++ b/scripts/schemadocs/package_test.go
@@ -5,8 +5,10 @@ package main_test
 
 import (
 	"testing"
+
+	gc "gopkg.in/check.v1"
 )
 
-// TestPackage just ensures that this package will build
-func TestPackage(t *testing.T) {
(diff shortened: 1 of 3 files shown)


jameinel:
$$merge$$

anastasiamac:
intermittent: FAIL: status_test.go:968: filteringBranchesSuite.TestFullStatusBranchFilterTwoBranchesSubordinateUnit

anastasiamac:
$$merge$$

PR #10562, opened 2019-08-26
migrations: Retry validation in the migration minion

babbageclunk:
## Description of change

In CI tests we were seeing errors in migration where the minions in unit agents would fail validation (and abort the migration) because the target controller wasn't ready - in this case because the model wasn't yet in the model cache. Have the minions in the various agents retry just in case.

## QA steps

* Bootstrap two controllers A and B.
* Add a model named m in A and deploy some things into it.
* Migrate m to B.
* Check that it works (not sure whether it's worth using iptables to force an agent to have to retry - tha's checked in the unit tests)

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/cmd/jujud/agent/unit/manifolds.go
+++ b/cmd/jujud/agent/unit/manifolds.go
@@ -233,6 +233,7 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 			AgentName:         agentName,
 			APICallerName:     apiCallerName,
 			FortressName:      migrationFortressName,
+			Clock:             clock.WallClock,
 			APIOpen:           api.Open,
 			ValidateMigration: config.ValidateMigration,
 			NewFacade:         migrationminion.NewFacade,
--- a/cmd/jujud/agent/unit/manifolds.go
+++ b/cmd/jujud/agent/unit/manifolds.go
@@ -99,6 +99,9 @@ type ManifoldsConfig struct {
 	// This is used by a number of workers to ensure serialisation of actions
 	// across the machine.
 	MachineLock machinelock.Lock
+
+	// Clock supplies timekeeping services to various workers.
+	Clock clock.Clock
 }
 
 // Manifolds returns a set of co-configured manifolds covering the various
@@ -233,7 +236,7 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 			AgentName:         agentName,
 			APICallerName:     apiCallerName,
 			FortressName:      migrationFortressName,
-			Clock:             clock.WallClock,
+			Clock:             config.Clock,
 			APIOpen:           api.Open,
 			ValidateMigration: config.ValidateMigration,
 			NewFacade:         migrationminion.NewFacade,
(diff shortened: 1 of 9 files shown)


cmd/jujud/agent/unit/manifolds.go
```diff
@@ -233,6 +233,7 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 			AgentName:         agentName,
 			APICallerName:     apiCallerName,
 			FortressName:      migrationFortressName,
+			Clock:             clock.WallClock,
```
howbazaar: There should be a clock passed in as monifold config to use.

worker/migrationminion/worker.go
```diff
@@ -149,7 +172,25 @@ func (w *Worker) doQUIESCE(status watcher.MigrationStatus) error {
...
+		w.catacomb.Dying(),
+	)
+	var err error
+	for attempt.Next() {
+		err = w.validate(status)
+		if err == nil {
+			break
+		}
+		if attempt.More() {
+			logger.Debugf("validation failed (retrying): %v", err)
```
howbazaar: Info perhaps?

worker/migrationminion/worker_test.go
```diff
@@ -204,13 +208,30 @@ func (s *Suite) TestVALIDATIONCantConnect(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 	defer workertest.CleanKill(c, w)
 
+	// Advance time enough for all of the retries to be exhausted.
+	for i := 0; i < 10; i++ {
+		// Don't worry about whether there are any waiters - we'll
+		// get them next time.
+		s.clock.Advance(time.Second)
+		time.Sleep(coretesting.ShortWait)
```
howbazaar: This does leave the potential for a race on slow substrates as we are assuming the worker is working at a fast enough speed.

Would be better to use wait advance the correct number of times IMO, and even have a channel that is closed by the APIOpen func above after 10 attempts. That gives a definitive thing to wait on that will ensure that the test passes.

worker/migrationminion/worker_test.go
```diff
@@ -226,15 +247,61 @@ func (s *Suite) TestVALIDATIONFail(c *gc.C) {
...
```
howbazaar: something similar here

cmd/jujud/agent/unit/manifolds.go
babbageclunk: Not in the unit agent, but there are for the machine and caasoperator manifolds. I've changed those - do you want me to introduce one for the unit agent?

worker/migrationminion/worker.go
babbageclunk: My thinking was that if you've got logging set to info and something eventually passes you probably don't want the preceding failed attempts logged.

worker/migrationminion/worker_test.go
babbageclunk: Yeah, will change this - it was basically just laziness to avoid working out the actual sleep times, but I guess I can have the computer calculate those for me!

cmd/jujud/agent/unit/manifolds.go
howbazaar: Not at this stage. Unless there is at least one other place that uses a wall clock.

worker/migrationminion/worker.go
howbazaar: Fair enough.

cmd/jujud/agent/unit/manifolds.go
howbazaar: Just looked, there are three other clock.WallClock usages in that file. We should probably introduce a clock in the config like the others.

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #10563, opened 2019-08-27
Introduce v3 command names for run-action and run

wallyworld:
## Description of change

Add a "juju-v3" feaure flag and start introducing new commands for run-action and run.
juju-action -> juju run
juju run -> juju exec

With flag not enabled, juju run is aliased to juju exec

## QA steps
export JUJU_DEV_FEATURE_FLAGS=juju-v3
juju run ....
juju exec ...

### DIFF
--- a/cmd/juju/action/run.go
+++ b/cmd/juju/action/run.go
@@ -11,6 +11,7 @@ import (
 	"github.com/juju/cmd"
 	"github.com/juju/errors"
 	"github.com/juju/gnuflag"
+	"github.com/juju/utils/featureflag"
 	"gopkg.in/juju/charm.v6"
 	"gopkg.in/juju/names.v3"
 	"gopkg.in/yaml.v2"
@@ -20,6 +21,7 @@ import (
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/cmd/output"
+	"github.com/juju/juju/feature"
 )
 
 // leaderSnippet is a regular expression for unit ID-like syntax that is used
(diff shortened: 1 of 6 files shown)


wallyworld:
[new command syntax behind feature flag](https://trello.com/c/K60ktehE/893-new-command-syntax-behind-feature-flag)

cmd/juju/action/run.go
```diff
@@ -97,12 +99,17 @@ func (c *runCommand) SetFlags(f *gnuflag.FlagSet) {
...
-	return jujucmd.Info(&cmd.Info{
+	info := jujucmd.Info(&cmd.Info{
 		Name:    "run-action",
 		Args:    "<unit> [<unit> ...] <action name> [key.key.key...=value]",
 		Purpose: "Queue an action for execution.",
 		Doc:     runDoc,
 	})
+	if featureflag.Enabled(feature.JujuV3) {
+		info.Name = "run"
+		info.Doc = strings.Replace(info.Doc, "run-action", "run", -1)
```
hpidcock: I think it would be good to append a note to the doc like "NOTE: pre-3.0 juju run functionality is provided by juju exec"

cmd/juju/commands/exec.go
```diff
@@ -246,8 +250,8 @@ func ConvertActionResults(result params.ActionResult, query actionQuery) map[str
 	return values
 }
 
-func (c *runCommand) Run(ctx *cmd.Context) error {
-	client, err := getRunAPIClient(c)
+func (c *execCommand) Run(ctx *cmd.Context) error {
```
hpidcock: Doesn't look possible with juju/cmd, but ideally we should be able to warn when a deprecated alias is used. Something like "WARN: juju exec will soon supersede juju run".

wallyworld: Yeah, not really possible. We'll highlight in the juju 3 release notes

cmd/juju/action/run.go
wallyworld: I don't thin we need to do this because this v3 behaviour is strictly opt-in - the user has to enable a feature flag to see it. They are therefore already aware of what they are doing in that sense and what the command names are etc.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10564, opened 2019-08-27
Wait for controller idle when waiting for model idle.

howbazaar:
Some tests are getting information from the controller. This will happen more as we use the cache controller for juju status.

This branch adds an idle func for the cache package in a similar way to the state watcher package. This allows the JujuConnSuite to wait for the model cache to be idle after processing txn watcher events. 

Among other things, this should fix the intermittent test failure of filteringBranchesSuite.TestFullStatusBranchNoFilter in the apiserver/facades/client/client package.


howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #10565, opened 2019-08-27
[dev] Fix panic when deploying bundles where maps have nil values

achilleasa:
## Description of change

The charm.v6 code that checks base bundles for the presence of overlay-specific options had a bug where the validation code would crash if the base bundle provided map entries (e.g. in options) with **nil** values.

This PR simply bumps the charm.v6 version to bring in the fix from https://github.com/juju/charm/pull/290

Note: this bug only affects dev releases and does **not** need to be back-ported to 2.6 as 2.6 does not use the same code path.

## QA steps

Try to deploy the following bundle with the latest edge snap (with --dry-run)
to verify it crashes. Then deploy the same with a juju binary built from 
this PR to verify that it does not crash (you will get unknown option errors as 
the options are bogus)

```yaml
# ssl_ca is left uninitialized so it resolves to nil
ssl_ca: &ssl_ca

applications:
  apache2:
    charm: apache2
    options:
      foo: bar
      ssl_ca: *ssl_ca
series: bionic
```

## Bug reference
https://bugs.launchpad.net/juju/+bug/1841105

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1494,7 +1494,7 @@
   revision = "51fa6e26128d74e445c72d3a91af555151cc3654"
 
 [[projects]]
-  digest = "1:ed246f348bdaa738a066035e2551a826e966f55a5528d6f44f9e8270c04a3888"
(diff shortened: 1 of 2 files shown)


achilleasa:
$$merge$$

PR #10567, opened 2019-08-27
Add show-action command

wallyworld:
## Description of change

Add a show-action command.
Prints the metadata about the specified action.
Reuses existing test infrastructure from list-action in the unit tests.

Driveby fix for adjusting list-action help when "juju-v3" feature flag in enabled.

## QA steps

deploy a charm with actions
juju show-action app action

### DIFF
--- a/cmd/juju/action/list.go
+++ b/cmd/juju/action/list.go
@@ -12,12 +12,14 @@ import (
 	"github.com/juju/errors"
 	"github.com/juju/gnuflag"
 	"github.com/juju/naturalsort"
+	"github.com/juju/utils/featureflag"
 	"gopkg.in/juju/names.v3"
 
 	"github.com/juju/juju/apiserver/params"
 	jujucmd "github.com/juju/juju/cmd"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/cmd/output"
+	"github.com/juju/juju/feature"
 )
 
 func NewListCommand() cmd.Command {
@@ -36,7 +38,14 @@ const listDoc = `
 List the actions available to run on the target application, with a short
 description.  To show the full schema for the actions, use --schema.
 
-For more information, see also the 'run-action' command, which executes actions.
+Examples:
+    juju list-actions postgresql
+    juju list-actions postgresql --format yaml
+    juju list-actions postgresql --schema
+
+See also:
+    run-action
+    show-action
 `
 
 // Set up the output.
@@ -63,13 +72,17 @@ func (c *listCommand) SetFlags(f *gnuflag.FlagSet) {
 }
 
 func (c *listCommand) Info() *cmd.Info {
-	return jujucmd.Info(&cmd.Info{
+	info := jujucmd.Info(&cmd.Info{
 		Name:    "actions",
 		Args:    "<application name>",
 		Purpose: "List actions defined for an application.",
 		Doc:     listDoc,
 		Aliases: []string{"list-actions"},
 	})
+	if featureflag.Enabled(feature.JujuV3) {
+		info.Doc = strings.Replace(info.Doc, "run-action", "run", -1)
+	}
+	return info
 }
 
 // Init validates the application name and any other options.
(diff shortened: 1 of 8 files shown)


anastasiamac:
So glad you are doing this \o/ While I am reviewing it, there is also a report asking for show-action https://bugs.launchpad.net/juju/+bug/1596906 that I think you are doing here :D

cmd/juju/action/list.go
```diff
@@ -81,11 +94,11 @@ func (c *listCommand) Init(args []string) error {
 	case 0:
 		return errors.New("no application name specified")
 	case 1:
-		svcName := args[0]
-		if !names.IsValidApplication(svcName) {
-			return errors.Errorf("invalid application name %q", svcName)
+		appName := args[0]
```
anastasiamac: Wow, what a rare find - 'svc' :D Thank you for updating!

cmd/juju/action/show.go
```diff
@@ -0,0 +1,125 @@
...
+		Doc:     showActionDoc,
+	})
+	if featureflag.Enabled(feature.JujuV3) {
+		info.Doc = strings.Replace(info.Doc, "run-action", "run", -1)
+	}
+	return info
+}
+
+func (c *showCommand) Run(ctx *cmd.Context) error {
+	api, err := c.NewActionAPIClient()
```
anastasiamac: It's a lot easier to test if the api client is retrieved from a command method, traditionally something along the lines of `newClientF`. This way you could also specify in interface what api methods you command needs :)

cmd/juju/action/show.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return err
+	}
+	info, ok := actions[c.actionName]
+	if !ok {
+		fmt.Fprintf(ctx.Stderr, "unknown action %q\n", c.actionName)
+		return cmd.ErrSilent
+	}
+
+	fmt.Fprintln(ctx.Stdout, info.Description+"\n\nArguments")
```
anastasiamac: You have access to command's out. Best to use it? 

You can also declare the formatter as yaml and this way the marshalling is done for you. See for example, ../storage/show.go

Something along the lines of 
```
func (c *showCommand) SetFlags(f *gnuflag.FlagSet) {
	c.out.AddFlags(f, "yaml", output.DefaultFormatters)
}
```

This will allow you to write in Run
```
c.out.Write(ctx, output)
```

cmd/juju/commands/main_test.go
```diff
@@ -574,6 +574,7 @@ var commandNames = []string{
 	"set-plan",
 	"set-series",
 	"set-wallet",
+	"show-action",
```
anastasiamac: Wouldn't this be conditional based on whether a feature flag is on or not?

wallyworld: No, this command is not behind a feature flag, it's not reliant on juju 3

cmd/juju/action/show.go
wallyworld: The issue is the output is a mix of straight text (the description and "Arguments" header) and YAML for the actual parameters. 
I've added a formatter for the actual parameters data but kept the Fprintf() for the other free text bits.

cmd/juju/action/show.go
wallyworld: Yeah, this method is on the base command. Didn't want to change all the existing commands in this PR.

wallyworld:
[show-action command](https://trello.com/c/dMVQR30K/894-show-action-command)

cmd/juju/action/show.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	defer api.Close()
+
+	actions, err := api.ApplicationCharmActions(params.Entity{Tag: c.applicationTag.String()})
+	if err != nil {
+		return err
+	}
+	info, ok := actions[c.actionName]
+	if !ok {
+		fmt.Fprintf(ctx.Stderr, "unknown action %q\n", c.actionName)
```
anastasiamac: This is better written as ```ctx.Infof()```

cmd/juju/action/show.go
```diff
@@ -0,0 +1,127 @@
...
+				continue
+			}
+			args[argName] = actionArg{
+				Type:        infoMap["type"],
+				Description: infoMap["description"],
+			}
+		}
+	}
+
+	c.out.Write(ctx, args)
```
anastasiamac: This swallows an error from out.Write. I think you just want to ```return c.out.Write(ctx, args)```.
There is also no need for the extra newline below.

cmd/juju/action/show.go
anastasiamac: This is easily resolved by putting the properties into a struct, similar to 

```
type actionArgs struct {
    Args map[string]actionArg `yaml:"Arguments"`
}
```

This way when you write the struct to c.out, everything will be taken care of.

cmd/juju/action/show.go
wallyworld: done

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10568, opened 2019-08-27
Copy run-action command to run with feature flag

wallyworld:
## Description of change

To prepare for implementing the new juju call (action), copy the existing juju run-action command so that the new version can be easily changed, leaving the legacy run-action version untouched. The changes will be significant and easier to do separately in a new command rather than adding feature flag logic to the existing run-action command.

The runaction(_test).go and call(_test).go files are straight copies of run(_test).go

## QA steps

ensure juju run-action works still

### DIFF
--- a/cmd/juju/action/call.go
+++ b/cmd/juju/action/call.go
@@ -31,13 +31,13 @@ var validLeader = regexp.MustCompile("^" + leaderSnippet + "$")
 // nameRule describes the name format of an action or keyName must match to be valid.
 var nameRule = charm.GetActionNameRule()
 
-func NewRunCommand() cmd.Command {
-	return modelcmd.Wrap(&runCommand{})
+func NewCallCommand() cmd.Command {
+	return modelcmd.Wrap(&callCommand{})
 }
 
-// runCommand enqueues an Action for running on the given unit with given
+// callCommand enqueues an Action for running on the given unit with given
 // params
-type runCommand struct {
+type callCommand struct {
 	ActionCommandBase
 	api           APIClient
 	unitReceivers []string
@@ -50,7 +50,7 @@ type runCommand struct {
 	args          [][]string
 }
 
-const runDoc = `
+const callDoc = `
 Queue an Action for execution on a given unit, with a given set of params.
 The Action ID is returned for use with 'juju show-action-output <ID>' or
 'juju show-action-status <ID>'.
@@ -76,19 +76,19 @@ explicit arguments will override the parameter file.
 
 Examples:
 
-    juju run mysql/3 backup --wait
-    juju run mysql/3 backup
-    juju run mysql/leader backup
+    juju call mysql/3 backup --wait
+    juju call mysql/3 backup
+    juju call mysql/leader backup
     juju show-action-output <ID>
-    juju run mysql/3 backup --params parameters.yml
-    juju run mysql/3 backup out=out.tar.bz2 file.kind=xz file.quality=high
-    juju run mysql/3 backup --params p.yml file.kind=xz file.quality=high
-    juju run sleeper/0 pause time=1000
-    juju run sleeper/0 pause --string-args time=1000
+    juju call mysql/3 backup --params parameters.yml
+    juju call mysql/3 backup out=out.tar.bz2 file.kind=xz file.quality=high
+    juju call mysql/3 backup --params p.yml file.kind=xz file.quality=high
+    juju call sleeper/0 pause time=1000
+    juju call sleeper/0 pause --string-args time=1000
 `
 
 // SetFlags offers an option for YAML output.
-func (c *runCommand) SetFlags(f *gnuflag.FlagSet) {
+func (c *callCommand) SetFlags(f *gnuflag.FlagSet) {
 	c.ActionCommandBase.SetFlags(f)
 	c.out.AddFlags(f, "yaml", output.DefaultFormatters)
 	f.Var(&c.paramsYAML, "params", "Path to yaml-formatted params file")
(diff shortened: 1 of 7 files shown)


cmd/juju/action/call.go
```diff
@@ -0,0 +1,310 @@
...
+var validLeader = regexp.MustCompile("^" + leaderSnippet + "$")
+
+// nameRule describes the name format of an action or keyName must match to be valid.
+var nameRule = charm.GetActionNameRule()
+
+func NewCallCommand() cmd.Command {
+	return modelcmd.Wrap(&callCommand{})
+}
+
+// runCommand enqueues an Action for running on the given unit with given
```
anastasiamac: callCommand...

cmd/juju/action/call.go
```diff
@@ -0,0 +1,310 @@
...
+
+If --params is passed, along with key.key...=value explicit arguments, the
+explicit arguments will override the parameter file.
+
+Examples:
+
+    juju call mysql/3 backup --wait
+    juju call mysql/3 backup
+    juju call mysql/leader backup
+    juju show-action-output <ID>
```
anastasiamac: I am not sure that you need 'show-action-output example among 'call' command examples :D

cmd/juju/action/call.go
```diff
@@ -0,0 +1,310 @@
...
+	if c.actionName == "" {
+		return errors.New("no action specified")
+	}
+
+	// Parse CLI key-value args if they exist.
+	c.args = make([][]string, 0)
+	for _, arg := range args[len(c.unitReceivers)+1:] {
+		thisArg := strings.SplitN(arg, "=", 2)
+		if len(thisArg) != 2 {
+			return errors.Errorf("argument %q must be of the form key...=value", arg)
```
anastasiamac: Maybe best to say ```key.key.key...=value``` as per Info().Args?

cmd/juju/action/call.go
```diff
@@ -0,0 +1,310 @@
...
+		}
+		keySlice := strings.Split(thisArg[0], ".")
+		// check each key for validity
+		for _, key := range keySlice {
+			if valid := nameRule.MatchString(key); !valid {
+				return errors.Errorf("key %q must start and end with lowercase alphanumeric, "+
+					"and contain only lowercase alphanumeric and hyphens", key)
+			}
+		}
+		// c.args={..., [key, key, key, key, value]}
```
anastasiamac: remove me? or re-phrase me :D

cmd/juju/action/call.go
```diff
@@ -0,0 +1,310 @@
...
+		}
+		if result.Action == nil {
+			return errors.Errorf("action failed to enqueue on %q", result.Action.Receiver)
+		}
+		tag, err := names.ParseActionTag(result.Action.Tag)
+		if err != nil {
+			return err
+		}
+
+		// Legacy Juju 1.25 output format for a single unit, no wait.
```
anastasiamac: OMG, are we still supporting this? Originally we said 'only for the life of trusty'... well, it's past EOL.

cmd/juju/action/call.go
```diff
@@ -0,0 +1,310 @@
...
+			}
+		}
+		// c.args={..., [key, key, key, key, value]}
+		c.args = append(c.args, append(keySlice, thisArg[1]))
+	}
+	return nil
+}
+
+func (c *callCommand) Run(ctx *cmd.Context) error {
+	if err := c.ensureAPI(); err != nil {
```
anastasiamac: This is kind of a problem since we close api later on but this method can still return it. I do prefer commands having function to obtain new API rather than caching the api.

cmd/juju/action/call.go
```diff
@@ -0,0 +1,310 @@
...
+		}
+
+		// Legacy Juju 1.25 output format for a single unit, no wait.
+		if !c.wait.forever && c.wait.d.Nanoseconds() <= 0 && len(results.Results) == 1 {
+			out := map[string]string{"Action queued with id": tag.Id()}
+			return c.out.Write(ctx, out)
+		}
+	}
+
+	out := make(map[string]interface{}, len(results.Results))
```
anastasiamac: This name is so close to c.out... Can it be named something else, say 'results'?

cmd/juju/action/export_test.go
```diff
@@ -29,27 +29,51 @@ type CancelCommand struct {
 	*cancelCommand
 }
 
-type RunCommand struct {
-	*runCommand
```
anastasiamac: Don't we need runCommand anymore in tests?

cmd/juju/commands/main_test.go
```diff
@@ -645,6 +645,7 @@ func (s *MainSuite) TestHelpCommands(c *gc.C) {
 	cmdSet := set.NewStrings(commandNames...)
 	if !featureflag.Enabled(feature.JujuV3) {
 		cmdSet.Add("run-action")
+		cmdSet.Add("run")
```
anastasiamac: "call"

cmd/juju/action/export_test.go
anastasiamac: Oh you've renamed. All good!

cmd/juju/action/call.go
wallyworld: show-action-output is still a valid command to see the result of calling an action.
It will be renamed however to show-operation in the next PR

cmd/juju/action/call.go
wallyworld: Existing code, removed in next PR but will do so here instead

cmd/juju/action/call.go
wallyworld: Removed in next PR

cmd/juju/action/call.go
wallyworld: Yeah, this is all existing, just a straight copy. Will try to clean up as work progresses

cmd/juju/commands/main_test.go
wallyworld: "run" is what's needed here.
If feature flag is *not* enabled, there is a run command

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10571, opened 2019-08-28
Avoid doing unnecessary work when cleaning up a model.

howbazaar:
This branch cleans up some error messages out of the state package so they don't throw away extra useful content.

Also when removing items from the database for a model, don't run transactions where the only op is the dead assertion.

This means we don't hit the race where the model is actually deleted underneath us in the feature tests.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -257,7 +257,7 @@ func (st *State) RemoveDyingModel() error {
 	}
 	err = st.removeAllModelDocs(bson.D{{"life", Dead}})
 	if errors.Cause(err) == txn.ErrAborted {
-		return errors.New("can't remove model: model not dead")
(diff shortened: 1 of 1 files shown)


howbazaar:
This fixes an intermittent failure that looks like this:
```
cmd_juju_controller_test.go:345:
    c.Check(err2, jc.ErrorIsNil)
... value *errors.Err = &errors.unformatter{message:"", cause:(*errors.errorString)(0xc000b13b00), previous:(*errors.Err)(0xc000601040), file:"github.com/juju/juju/state/state.go", line:262} ()
... error stack:
        failed to reload transaction: not found
        github.com/juju/juju/state/state.go:308: removing from collection "units"
        github.com/juju/juju/state/state.go:262: 
```

howbazaar:
$$merge$$

PR #10573, opened 2019-08-29
application data bag: Fix racy WatchApplicationSettings tests

babbageclunk:
## Description of change

The WatchApplicationSettings tests would fail intermittently.
Example: https://jenkins.juju.canonical.com/view/Unit%20tests/job/RunUnittests-race-amd64/1764/testReport/github/com_juju_juju_state/TestPackage/

They're getting two notifications on the watcher instead of the expected one about half the time under stress-race. Adding a StartSync fixes this.

## QA steps

* With the change the tests don't fail under stress-race.

## Documentation changes
None

## Bug reference
None

### DIFF
--- a/state/relation_test.go
+++ b/state/relation_test.go
@@ -947,6 +947,7 @@ func (s *RelationSuite) TestWatchApplicationSettings(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 	relation, err := s.State.AddRelation(eps...)
 	c.Assert(err, jc.ErrorIsNil)
+	s.WaitForModelWatchersIdle(c, s.Model.UUID())
 
 	w, err := relation.WatchApplicationSettings(mysql)
 	c.Assert(err, jc.ErrorIsNil)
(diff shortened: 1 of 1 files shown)


state/relation_test.go
```diff
@@ -947,6 +947,7 @@ func (s *RelationSuite) TestWatchApplicationSettings(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 	relation, err := s.State.AddRelation(eps...)
 	c.Assert(err, jc.ErrorIsNil)
+	s.State.StartSync()
```
howbazaar: You should instead use `WaitForModelWatchersIdle(c, s.State.ModelUUID())`

The reason it fails intermittently is that the state watcher will send down an initial event, but the events that are in flight from the construction of the relation may not have fully flowed through, so you end up with an additional notification.

This is why we have the WaitForModelWatcherIdle method.

babbageclunk: Ok, using that instead - thanks.

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #10581, opened 2019-09-01
Fix the intermittent failure of TestRemovePrimaryValidSecondaries

howbazaar:
There were two different causes of the intermittent failure. Both had to do with the expectations of the worker main loop, and things actually happening differently.

Given that all the inputs to the worker were controlled, we should be able to be deterministic.

The first failure case was when passing through the loop we had a success then a failure to write. This would give two waiters. However if we had failure followed by success, there would be only one waiter. This was fixed by the change in logic that says if there was a pevious failure when we get a success, update the wait time to be the poll interval.

The second failure case was during the initialize function call where the call would return before the time.After channel had been signalled. There was the assumption that init wasn't complete until there had been a poll happen, but it was possible to have this return before the updateChan was signalled. This caused the rest of the test to progress before things were ready.

## QA steps

 * stress-race the test for ages

## Bug reference

https://bugs.launchpad.net/juju/+bug/1802590

### DIFF
--- a/worker/peergrouper/worker.go
+++ b/worker/peergrouper/worker.go
@@ -144,6 +144,10 @@ type Config struct {
 	// and is used to publish the details of the
 	// API servers.
 	Hub Hub
+
+	// UpdateNotify is called when the update channel is signalled.
+	// Used solely for test synchronization.
+	UpdateNotify func()
 }
 
 // Validate validates the worker configuration.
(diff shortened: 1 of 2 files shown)


worker/peergrouper/worker.go
```diff
@@ -303,7 +310,11 @@ func (w *pgWorker) loop() error {
 			// Update the replica set members occasionally to keep them up to
 			// date with the current replica-set member statuses.
 			logger.Tracef("succeeded, waking up after: %v", pollInterval)
-			if updateChan == nil {
+			// If we had previously failed to update to update the replicaset,
```
anastasiamac: only need one 'to update'

howbazaar:
$$merge$$

PR #10582, opened 2019-09-02
Increase instance size for GKE functional test

hpidcock:
## Description of change

Increase instance size for GKE functional test due controller pod requiring too much memory for an n1-standard-1 instance.


hpidcock:
$$merge$$

PR #10584, opened 2019-09-02
worker/raft TestNoLeaderTimeout had a couple small bugs

jameinel:
## Description of change

This has an intermittent failure, which I haven't been able to
reproduce, but it does have a couple issues that I've tried to address:

a) "wait until no longer the leader" was comparing the Leader() string
   to "localhost", but the value was actually "123". So the wait always
   immediately returned.
b) The "WaitAdvance" updated the clock exactly 10s, which is exactly the
   length of time that we would wait to decide that we are having
   NoLeader. Instead advance more. Note that I *can* reproduce the issue
   every time if I change the WaitAdvance to 9.999s. But the logic
   should all be integer math in nanoseconds. Hopefully 10.001s does the
   trick.

## QA steps

I wasn't able to personally reproduce the issue. But it triggered here:
https://jenkins.juju.canonical.com/job/RunUnittests-race-amd64/1778/testReport/junit/github/com_juju_juju_worker_raft/TestPackage/
I *was* able to reproduce the symptoms by setting the WaitAdvance less than 10s.

## Documentation changes

None.

## Bug reference

* https://bugs.launchpad.net/bugs/1775737

### DIFF
--- a/worker/raft/worker_test.go
+++ b/worker/raft/worker_test.go
@@ -292,6 +292,8 @@ func (s *WorkerSuite) TestNoLeaderTimeout(c *gc.C) {
 	f2 := raft0.AddVoter("2", transport2.LocalAddr(), 0, 0)
 	c.Assert(f1.Error(), jc.ErrorIsNil)
 	c.Assert(f2.Error(), jc.ErrorIsNil)
+	// Now that we are leader, check that we are listed as such
+	initialLeader := raft0.Leader()
 
 	rafttest.CheckConfiguration(c, raft0, []coreraft.Server{{
 		ID:       "123",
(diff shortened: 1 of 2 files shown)


worker/raft/worker_test.go
```diff
@@ -329,7 +337,9 @@ func (s *WorkerSuite) TestNoLeaderTimeout(c *gc.C) {
 	// waits when we advance:
 	// * the loop timeout wait from starting the worker
 	// * the no leader timeout check in loop.
-	c.Assert(s.clock.WaitAdvance(10*time.Second, coretesting.LongWait, 2), jc.ErrorIsNil)
+	// The NoLeaderTimeout waits for 10s, so we increment the clock just slightly
+	// longer than that.
+	c.Assert(s.clock.WaitAdvance(10001*time.Millisecond, coretesting.LongWait, 2), jc.ErrorIsNil)
```
howbazaar: Given all the other cases in the code where we advance exactly the right amount, I don't think this is the issue.

babbageclunk: I can see your point (essentially this changes the test from >= to >), but I think it's worth seeing if it does actually change the behaviour given that we can't reproduce it locally.

jameinel: So this was a red herring. Which we followed up with *another* red herring. (Namely, inside the code we use raft.LastContact() which uses a real time.Now() call.) That *could* be the failure, but it would only trigger if the Clock time and the time.Now() time differ by 6 seconds (no leader timeout is 4s, clock.WaitAdvance is 10s.)

However, it turned out to really be the failed "am I still the leader" check.

If you put a 1ms sleep for the node to step down after getting the Demote request. Then the request succeeds (f3.Error() is nil). But you can get to the call WaitAdvance(10s) *before* the node actually steps down as the leader. Which means that when the Worker loop checks "r.State() == raft.Leader" it gets True. And if *I* am the leader, than clearly there we aren't  missing a leader, and we shouldn't trigger NoLeaderTimeout.

So the fix really should be the "fix the loop to wait until I'm not the leader", and I've reverted the change around 10s vs 10.001s.

babbageclunk: Curses, Tim was right!

jameinel:
$$merge$$

PR #10586, opened 2019-09-03
Fixes lp#1841885: clear models credential when it's removed.

anastasiamac:
## Description of change

Under normal circumstances, Juju will not allow to delete a credential that is being used by existing models. However, when using an API and when the credential removal is forced, there is a way to by-pass this restriction and delete a credential that is currently in-use. It leaves the system in a bit of a twist as Juju will see that the model has a credential set but not the actual value of the credential. This condition is currently treated the same way as an invalid credential, i.e. all models that use deleted credential will be suspended. However, it's cleaner to clear out these references.

This PR deletes credential references when credential is deleted preserving referential integrity.  

## Bug reference

https://bugs.launchpad.net/juju/+bug/1841885

### DIFF
--- a/state/cloudcredentials.go
+++ b/state/cloudcredentials.go
@@ -479,35 +479,39 @@ func (st *State) CredentialModels(tag names.CloudCredentialTag) (map[string]stri
 
 // RemoveModelsCredential clears out given credential reference from all models that have it.
 func (st *State) RemoveModelsCredential(tag names.CloudCredentialTag) error {
-	coll, cleanup := st.db().GetCollection(modelsC)
-	defer cleanup()
+	buildTxn := func(attempt int) ([]txn.Op, error) {
+		logger.Tracef("creating operations to remove models credential, attempt %d", attempt)
+		coll, cleanup := st.db().GetCollection(modelsC)
+		defer cleanup()
 
-	sel := bson.D{
-		{"cloud-credential", tag.Id()},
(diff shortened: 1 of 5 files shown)


anastasiamac:
[delete model cred refs when cred is force removed](https://trello.com/c/uGlny3ch/897-delete-model-cred-refs-when-cred-is-force-removed)

anastasiamac:
intermittent failure in github.com/juju/juju/worker/raft/raftflag.TestPackage

anastasiamac:
!!build!!

state/cloudcredentials.go
```diff
@@ -477,6 +477,39 @@ func (st *State) CredentialModels(tag names.CloudCredentialTag) (map[string]stri
...
+	for iter.Next(&doc) {
+		id, ok := doc["_id"]
+		if !ok {
+			return errors.New("no id found in model doc")
+		}
+
+		ops = append(ops, txn.Op{
+			C:      modelsC,
+			Id:     id,
+			Assert: txn.DocExists,
```
wallyworld: Need to assert $ne Dead here too based on the initial query. Also, if the model is deleted before the txn runs, it will error out, so need to include in a buildTxn loop. Either that or remove the DocExists assert since if the model is gone, who cares that the credential cannot be cleared.

Based on this issue, there should be a test for such a race condition using BeforeHook

anastasiamac:
$$merge$$

PR #10587, opened 2019-09-03
Don't create a second state pool in the JujuConnSuite.

howbazaar:
The JujuConnSuite was creating a second state pool. Both the state pool instances were calling the same callback function to tell the suite that the txnwatcher and hubwatchers were idle. This meant that sometimes when we asked for a model to be idle, one of the watchers would be, but the other may still be busy, leading to weird intermittent failures.


howbazaar:
$$merge$$

howbazaar:
$$merge$$

howbazaar:
$$merge$$ - port in use

PR #10588, opened 2019-09-03
Add a skip_state_tests build tag and apply tags in `make test`

babbageclunk:
## Description of change

This will enable some test jobs (particularly the arm64 one) to skip running the state tests as part of the normal build, since they take too long.

The state tests can be skipped using `go test --tags skip_state_tests` or `make test BUILD_TAGS=skip_state_tests`.

## QA steps

* Ran tests with and without the skip build tag, saw the state tests being skipped.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,8 @@ pre-check:
 check: dep pre-check test
 
 test: dep
-	go test $(CHECK_ARGS) -test.timeout=$(TEST_TIMEOUT) $(PROJECT_PACKAGES) -check.v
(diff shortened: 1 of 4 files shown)


mitechie:
Should this just be a different target? e.g. make test and make test-no-state ?

babbageclunk:
Using the BUILD_TAGS variable that was already there makes more sense to me. Adding another target seems like overkill given that we're only anticipating needing it in one job. I guess it could be useful for people who want to run all the tests except the state ones, but it doesn't seem like we actually want that to be easy, do we?

babbageclunk:
$$merge$$

PR #10589, opened 2019-09-03
Enable multipass and LXD-based builds

dshcherb:
## Description of change

Enable multipass and LXD-based builds
    
If a base is not specified in snapcraft.yaml then snapcraft tries to use
the host system to build a snap (equivalent to --destructive-mode), see

* LP: #1842371;
* https://snapcraft.io/docs/base-snaps.

This is why core18 is specified as a base with this change.

Also, the dep plugin present in the Juju repository tries to append the
"required" key to the json schema taken from the base class under
snapcraft/_baseplugin.py which results in the following error:

  File "/root/project/snap/plugins/x-dep.py", line 64, in schema
    schema["required"].append("go-importpath")
KeyError: 'required'

As BasePlugin's 'schema' class method does not include the "required"
key it needs to be added.

Note: using --use-lxd might be a better idea considering LP: #1842376 -
Juju needs a lot of memory to be built which is more than 2 GiB
allocated for multipass VMs at the time of writing.

## QA steps

Old behavior:

snapcraft --destructive-mode

Multipass (expected to fail unless SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY is
set to something 4G - 6G, see LP: #1842376):

snapcraft

LXD:

snapcraft --use-lxd

## Documentation changes

CI build scripts will need to pass --destructive mode if there is an
expectation that the execution environment will be used as is.

### DIFF
--- a/snap/plugins/x-dep.py
+++ b/snap/plugins/x-dep.py
@@ -61,7 +61,7 @@ class DepPlugin(snapcraft.BasePlugin):
         }
 
         # The import path must be specified.
-        schema["required"].append("go-importpath")
+        schema["required"] = ["go-importpath"]
 
         return schema
 
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -4,6 +4,7 @@ summary: juju client
 description: Through the use of charms, juju provides you with shareable, re-usable, and repeatable expressions of devops best practices.
 confinement: classic
 grade: devel
+base: core18
 
 apps:
   juju:


timClicks:
!!build!!

dshcherb:
I tried installing a built snap with core18 in a Xenial LXD container with a fresh cloud image and it worked:

```
snap list
Name    Version    Rev   Tracking  Publisher   Notes
core18  20190723   1074  stable    canonical✓  base
juju    2.7-beta1  x1    -         -           classic
snapd   2.40       4022  stable    canonical✓  snapd
```

I think we should make a release note that the snap will be switched to core18.

I also tried using 'core' instead of 'core18' and got a build failure:

```
Pulling juju 
Cloning into '/root/parts/juju/src'...
done.
Fetching dep...
go get github.com/golang/dep/cmd/dep
package context: unrecognized import path "context" (import path does not begin with hostname)
Failed to run 'go get github.com/golang/dep/cmd/dep' for 'juju': Exited with code 1.
Verify that the part is using the correct parameters and try again.
```

Have not investigated further yet - JFYI.

howbazaar:
What base does the juju snap have right now?

howbazaar:
And can we just have it update to core18?

timClicks:
1) The [core](https://github.com/snapcore/core) snap is the default,  based on Ubuntu 16.04 
2) Yes, my understanding is if the snap builds, then it works. We can certainly do some double checking though.

dshcherb:
Found this https://bugs.launchpad.net/snapd/+bug/1819875 but we do not have one of the preconditions to trigger it:

> Preconditions:
>  Snap updated base from core16 to core18
>  Snap has one or more services

Since the Juju snap does not define any services we will not be affected by this.

As far as using core18 on a 16.04 system I think we need snapd to be at least >= 2.35 which was released in August 2018.

https://github.com/snapcore/snapd/blob/bfe4456cc1f6d1e3f117b99a551a67092892efb4/packaging/ubuntu-16.04/changelog#L2141 (2.35: "cmd/snap-confine: fix snaps running on core18")
https://github.com/snapcore/snapd/releases/tag/2.32.4

dshcherb:
🤔

howbazaar:
$$merge$$

hpidcock:
$$merge$$

PR #10591, opened 2019-09-03
Merge pull request #10586 from anastasiamac/lp1841885

anastasiamac:
https://github.com/juju/juju/pull/10586

## Description of change

Under normal circumstances, Juju will not allow to delete a credential that is being used by existing models. However, when using an API and when the credential removal is forced, there is a way to by-pass this restriction and delete a credential that is currently in-use. It leaves the system in a bit of a twist as Juju will see that the model has a credential set but not the actual value of the credential. This condition is currently treated the same way as an invalid credential, i.e. all models that use deleted credential will be suspended. However, it's cleaner to clear out these references.

This PR deletes credential references when credential is deleted preserving referential integrity. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1841885

----

## Description of change

## QA steps

## Documentation changes

## Bug reference


anastasiamac:
[delete model cred refs when cred is force removed](https://trello.com/c/uGlny3ch/897-delete-model-cred-refs-when-cred-is-force-removed)

anastasiamac:
$$merge$$

PR #10594, opened 2019-09-03
Fix race in TestWatchAPIHostPorts.

howbazaar:
Yet another case where we are creating a watcher before necessary creation events have finished flowing through the low level event subsystem.

Since it is a mixin test class, we need to pass it a method to sync on since it can't access the JujuConnSuite methods directly.


howbazaar:
Example test failure:
https://jenkins.juju.canonical.com/view/Unit%20tests/job/RunUnittests-race-amd64/1787/testReport/junit/github/com_juju_juju_api_provisioner/TestAll/

howbazaar:
!!build!! failed to resolve archive.ubuntu.com

howbazaar:
$$merge$$

PR #10596, opened 2019-09-04
Cleaner test runs (for /tmp)

howbazaar:
When we run 'make test', the machine ends up with a bunch of files and directories left in /tmp.
If these aren't regularly pruned, then tests will start failing - like the centos test machine.

This change introduces an explicit temp dir for the duration of the tests, and it deletes it at the end of a successful run.

## QA steps

* run 'make test' and observe /tmp


howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #10598, opened 2019-09-04
Fix TestWatchCloudService by moving a badly placed WaitForModelWatchersIdle call

wallyworld:
## Description of change

Fix TestWatchCloudService by moving a badly placed WaitForModelWatchersIdle call


wallyworld:
$$merge$$

PR #10601, opened 2019-09-04
Expose GetChangesMapArgs for clients to use

SimonRichardson:
## Description of change

The following exposes GetChangesMapArgs on the Bundle api server,
including adding the methods to the Bundle client. The changes in
the cmd/.../bundle.go requires a LOT of changes, which requires
more time than I've currently got. So instead of doing it in one
stage, I've updated it to make it better than before and left a
very large TODO. Either I'll get time to finish it off at some
point, or someone else will...

In an ideal world, most of the overlay bundle changes would also
be put on the server, again this is a multi-day effort to get it
working and tested, so that again is left as an exercise of the
reader.

What do work at the moment is getting a map of arguments from the
api server, so clients like pylib can work without jumping through
hoops and crashing everytime someone updates bundlechanges.

Note: bundlechanges is a minefield to clients :o

## QA steps

TBA: Will need pylibjuju to get it working for QA steps.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -455,11 +455,11 @@
   revision = "b99631de12cf04a906c1d4e4ec54fb86eae5863d"
 
 [[projects]]
-  digest = "1:3bf45368929d01a44755247cee64f93057c8a0f012ade9a227e4a60a6d2b85ff"
+  digest = "1:ebbd30a06de5bc3933e34d426cfaa69a0cefa10f575d520e6597acbf37e4d132"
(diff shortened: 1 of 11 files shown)


SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

Uniter test failures.

SimonRichardson:
$$merge$$

    failed to start mongo: mongod failed to listen on port 41364

SimonRichardson:
$$merge$$

     Snap failure

PR #10603, opened 2019-09-05
Ensure pod config values with special characters are quoted

wallyworld:
## Description of change

pod spec yaml can contain strings with special characters; these need to be quoted, eg

```
spec:
  containers:
  - env:
    - name: BAR
      value: h'e'llo@world
    - name: FOO
      value: '["hello", "world"]'
```

## QA steps

Deploy a k8s charm with yaml like the above and use kubectl to check pod definition afterwards.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1842691

### DIFF
--- a/caas/kubernetes/provider/k8s_test.go
+++ b/caas/kubernetes/provider/k8s_test.go
@@ -256,6 +256,7 @@ func getBasicPodspec() *caas.PodSpec {
 				"restricted": "'yes'",
 				"bar":        true,
 				"switch":     "on",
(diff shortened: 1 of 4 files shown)


wallyworld:
[Array in YAML causes Kubernetes charm failed deployment](https://trello.com/c/9TgAMzFt/907-array-in-yaml-causes-kubernetes-charm-failed-deployment)

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10612, opened 2019-09-09
Fix some action test races

wallyworld:
## Description of change

Two fixes here: protect the bytes buffer used to record a copy of stdout with a mutex, and ensure hook loggers are stopped prior to reading the stdout buffers.

## QA steps

go test --race

### DIFF
--- a/worker/uniter/runner/runner.go
+++ b/worker/uniter/runner/runner.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 	"unicode/utf8"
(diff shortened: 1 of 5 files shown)


worker/uniter/runner/runner.go
```diff
@@ -369,6 +381,7 @@ func (runner *runner) runCharmHookOnRemote(hookName string, env []string, charmL
 	// When running an action, We capture stdout and stderr
 	// separately to pass back.
 	var actionErr = actionOut
+	var hookErrLogger *charmrunner.HookLogger
```
ycliuhw: all the two loggers are already stopped here in defer, no need to pass into `ExecParams` to call stop inside `executor`.

ycliuhw: as discussed, we can't combine them. so all good.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10613, opened 2019-09-09
Fix action id matching

hpidcock:
## Description of change

Fixes `Exception: Action id not found in output: Action queued with id: network-health-xenial-1`

Action ID's don't appear to be UUIDs anymore. This allows broader matching of the action ID, probably should look at reworking `action_do` to not rely on juju 1.25 legacy output from `cmd/juju/action/runaction.go`

### DIFF
--- a/acceptancetests/jujupy/client.py
+++ b/acceptancetests/jujupy/client.py
@@ -1877,7 +1877,7 @@ class ModelClient:
 
         output = self.get_juju_output("run-action", *args)
         action_id_pattern = re.compile(
-            'Action queued with id: ([a-f0-9\-]{36})')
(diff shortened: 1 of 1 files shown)


acceptancetests/jujupy/client.py
```diff
@@ -1877,7 +1877,7 @@ def action_do(self, unit, action, *args):
 
         output = self.get_juju_output("run-action", *args)
         action_id_pattern = re.compile(
-            'Action queued with id: ([a-f0-9\-]{36})')
+            'Action queued with id: ([a-z0-9\-]+)')
```
wallyworld: Worth doing an OR to account for old uuids and new <applicationname>-N ids?
The regexp for <appname>-N would be

`((?:[a-z][a-z0-9]*(?:-[a-z0-9]*[a-z][a-z0-9]*)*)-)(?:0|[1-9][0-9]*)`

hpidcock:
$$merge$$

PR #10617, opened 2019-09-10
ask-or-tell: 'autoload-credentials'. 

anastasiamac:
## Description of change

If Juju detects current controller, instead of using it by default, it will prompt the user to confirm whether detected credentials should be uploaded to it or not. If the upload is always desirable, but the prompt is not, the user can use ''-no-prompt" option.

As a drive-by, this PR updates command help doc.

### DIFF
--- a/cmd/juju/cloud/detectcredentials.go
+++ b/cmd/juju/cloud/detectcredentials.go
@@ -412,7 +412,7 @@ func (c *detectCredentialsCommand) interactiveCredentialsUpdate(ctxt *cmd.Contex
 			continue
 		}
 		if cloud.Type != cred.cloudType {
-			ctxt.Warningf("%v", errors.Errorf("chosen credentials not compatible with %q cloud", cloud.Type))
+			ctxt.Warningf("%v", errors.Errorf("chosen credential not compatible with %q cloud", cloud.Type))
 			continue
 		}
 
--- a/cmd/juju/cloud/detectcredentials.go
+++ b/cmd/juju/cloud/detectcredentials.go
@@ -29,7 +29,6 @@ import (
 
 type detectCredentialsCommand struct {
 	modelcmd.OptionalControllerCommand
-	cmd.CommandBase
 	out cmd.Output
 
 	cloudType string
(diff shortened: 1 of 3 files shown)


anastasiamac:
[ask-or-tell: autoload-credentials](https://trello.com/c/vtcnRMnt/908-ask-or-tell-autoload-credentials)

cmd/juju/cloud/detectcredentials.go
```diff
@@ -182,7 +178,8 @@ func (c *detectCredentialsCommand) allClouds() (map[string]jujucloud.Cloud, erro
 	for k, v := range personalClouds {
 		clouds[k] = v
 	}
-	if !c.Local {
+	if !c.Local && c.ControllerName != "" {
+		ctxt.Infof("\nLooking for cloud information on controller %q...", c.ControllerName)
 		// If there is a cloud definition for the same cloud both
 		// on the controller and on the client and they conflict,
 		// we want definition from the controller to take precedence.
```
wallyworld: I wonder if this warrants a warning?

cmd/juju/cloud/detectcredentials.go
```diff
@@ -402,11 +408,11 @@ func (c *detectCredentialsCommand) interactiveCredentialsUpdate(ctxt *cmd.Contex
 		}
 		cloud, err := common.CloudOrProvider(cloudName, c.cloudByNameFunc)
 		if err != nil {
-			fmt.Fprintln(ctxt.Stderr, err.Error())
+			ctxt.Warningf("%v", err)
 			continue
 		}
 		if cloud.Type != cred.cloudType {
-			fmt.Fprintln(ctxt.Stderr, errors.Errorf("chosen credentials not compatible with a %s cloud", cloud.Type))
+			ctxt.Warningf("%v", errors.Errorf("chosen credentials not compatible with %q cloud", cloud.Type))
```
wallyworld: chosen credential

cmd/juju/cloud/detectcredentials.go
```diff
@@ -490,17 +497,15 @@ func (c *detectCredentialsCommand) addRemoteCredentials(ctxt *cmd.Context, cloud
 			result, err := client.UpdateCloudsCredentials(verified)
 			if err != nil {
 				logger.Errorf("%v", err)
-				ctxt.Warningf("Could not add credentials remotely, on controller %q", c.ControllerName)
+				ctxt.Warningf("Could not upload credentials to controller %q", c.ControllerName)
```
wallyworld: "credential"
this deals with one at a time right?

anastasiamac: No, this deals with per-cloud credentials.. could be more than one.

cmd/juju/cloud/detectcredentials.go
anastasiamac: Yes  indeed :)

cmd/juju/cloud/detectcredentials.go
anastasiamac: I do not think it's a warning... I think it's information, as in 'we've looked locally, now we are looking on the controller..." Kind of like a progress of processing or a commentary :)

anastasiamac:
$$merge$$

PR #10618, opened 2019-09-10
state: Ensure units leave scope if the relation is force-destroyed

babbageclunk:
## Description of change

We've seen some cases with cross-model relations where the relation can't be removed because there are remaining relation scope docs (either because of a network problem or a bug in CMR) that means the the relation has a nonzero unit count. This change adds a fallback cleanup when the relation is force-destroyed that leaves scope on any that are left behind. Because the unit might not exist we construct the RelationUnit directly rather than trying to get a unit first.

## QA steps

* Bootstrap a controller
* Deploy two applications and relate them
* Stop one of the unit agents
* Remove the relation with force - eventually it will be removed, even though the unit agent wasn't running to leave.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/state/cleanup.go
+++ b/state/cleanup.go
@@ -4,6 +4,7 @@
 package state
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/juju/errors"
@@ -31,6 +32,7 @@ var (
 const (
 	// SCHEMACHANGE: the names are expressive, the values not so much.
 	cleanupRelationSettings              cleanupKind = "settings"
+	cleanupForceDestroyedRelation        cleanupKind = "forceDestroyRelation"
 	cleanupUnitsForDyingApplication      cleanupKind = "units"
 	cleanupCharm                         cleanupKind = "charm"
 	cleanupDyingUnit                     cleanupKind = "dyingUnit"
@@ -154,6 +156,8 @@ func (st *State) Cleanup() (err error) {
 		switch doc.Kind {
 		case cleanupRelationSettings:
 			err = st.cleanupRelationSettings(doc.Prefix)
+		case cleanupForceDestroyedRelation:
+			err = st.cleanupForceDestroyedRelation(doc.Prefix)
 		case cleanupCharm:
 			err = st.cleanupCharm(doc.Prefix)
 		case cleanupApplication:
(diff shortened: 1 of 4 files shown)


state/cleanup.go
```diff
@@ -237,6 +240,66 @@ func (st *State) cleanupRelationSettings(prefix string) error {
...
 
+func (st *State) cleanupForceDestroyedRelation(prefix string) (err error) {
+	relation, err := st.KeyRelation(prefix)
+	if err != nil {
+		return errors.Annotatef(err, "getting relation %q", prefix)
+	}
+	scopes, closer := st.db().GetCollection(relationScopesC)
+	defer closer()
+
+	sel := bson.D{{"_id", bson.D{{"$regex", "^" + st.docID(relation.globalScope())}}}}
```
wallyworld: To be safe should we add a "#" (I think that's the separator) to the end of the regexp expression?

state/cleanup.go
```diff
@@ -237,6 +240,66 @@ func (st *State) cleanupRelationSettings(prefix string) error {
...
+		// isPrincipal and isLocalUnit, and we're only using the RU to
+		// call LeaveScope on it.
+		ru := RelationUnit{
+			st:          st,
+			relation:    relation,
+			unitName:    unitName,
+			endpoint:    matchingEp,
+			scope:       scope,
+			isPrincipal: true,
+			isLocalUnit: true,
```
wallyworld: isLocalUnit is only used on EnterScope() isn't it? So do we even need to set it here?
Same with isPrincipal

babbageclunk: Well, I thought they might also be used in some other RelationUnit methods - but definitely not in leave scope so no, we don't need to set it here.

state/cleanup.go
babbageclunk: Yeah, good call.

babbageclunk:
Good call about the unit deletion - I've added a case like that to one of the tests.

$$merge$$

babbageclunk:
Weird failure in cmd/juju/status tests `StatusSuite.TestStatusAllFormats` test 11 step 43:

```
10:40:07 ... mismatch at ["applications"]["mysql"]["units"]["mysql/0"]["subordinates"]["logging/1"]: length mismatch, 4 vs 5; obtained map[interface {}]interface {}{"leader":true, "public-address":"10.0.2.1", "workload-status":map[interface {}]interface {}{"current":"error", "message":"somehow lost in all those logs", "since":"01 Apr 15 01:23+10:00"}, "juju-status":map[interface {}]interface {}{"current":"idle", "since":"01 Apr 15 01:23+10:00"}}; expected map[interface {}]interface {}{"workload-status":map[interface {}]interface {}{"since":"01 Apr 15 01:23+10:00", "current":"error", "message":"somehow lost in all those logs"}, "branch":"apple", "juju-status":map[interface {}]interface {}{"current":"idle", "since":"01 Apr 15 01:23+10:00"}, "leader":true, "public-address":"10.0.2.1"}
```

Reading it carefully the actual logging subordinate unit's map is missing the branch key. Model cache being slow? I haven't been able to reproduce it locally.

$$merge$$

PR #10619, opened 2019-09-10
1843219: All watcher missing info status

SimonRichardson:
## Description of change

The following PR ensures that we set the info.status before
performing a store update if the old info is not nil. In other
words, we need to reuse the data from the old info and set it on
the new info which we'll be saving.

The reason for the tests changes, is now that we have the old
data we don't need to save any changes if the data is the same.

This has been tested with pylibjuju and the status changes are
now populated from the start, allowing you to monitor the changes
of a charm.

## QA steps

Unfortunately there is some work to test this out and requires 
pylibjuju for the quickest turn around...

Checkout 2.6 branch and bootstrap.
```sh
git checkout 2.6
juju bootstrap lxd test --no-gui --build-agent
watch --color -n 1 juju status --color
```

Setup pylibjuju and apply the following changes to pylibjuju 
example/deploy_bundle.py: [print-diff.txt](https://github.com/juju/juju/files/3596102/print-diff.txt)
It requires quite a large bundle to trigger this, so for me a
very basic openstack did this.
```sh
tox -e example -- examples/deploy_bundle.py
```

Notice how some status fields in the print output will be empty 
and has no "since" field:
```json
"status": {"version": "", "current": "", "message": ""}
```
----
To verify the patch, check this branch out and bootstrap:
```sh
juju bootstrap lxd tester --no-gui --build-agent
watch --color -n 1 juju status --color
```

Then run the pylibjuju example again, with the diff applied:
```sh
tox -e example -- examples/deploy_bundle.py
```
The output should be:
```json
"status": {"since": "2019-09-10T13:23:44.99699161Z", "version": "", "current": "waiting", "message": "waiting for machine"}
```

### DIFF
--- a/state/allwatcher.go
+++ b/state/allwatcher.go
@@ -554,19 +554,19 @@ func (app *backingApplication) updated(st *State, store *multiwatcherStore, id s
 		}
 	} else {
 		// The entry already exists, so preserve the current status.
-		oldInfo := oldInfo.(*multiwatcher.ApplicationInfo)
-		info.Constraints = oldInfo.Constraints
-		info.WorkloadVersion = oldInfo.WorkloadVersion
-		if info.CharmURL == oldInfo.CharmURL {
+		appInfo := oldInfo.(*multiwatcher.ApplicationInfo)
+		info.Constraints = appInfo.Constraints
+		info.WorkloadVersion = appInfo.WorkloadVersion
+		if info.CharmURL == appInfo.CharmURL {
 			// The charm URL remains the same - we can continue to
 			// use the same config settings.
-			info.Config = oldInfo.Config
+			info.Config = appInfo.Config
 		} else {
 			// The charm URL has changed - we need to fetch the
 			// settings from the new charm's settings doc.
 			needConfig = true
 		}
-		info.Status = oldInfo.Status
+		info.Status = appInfo.Status
 	}
 	if needConfig {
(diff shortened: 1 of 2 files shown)


SimonRichardson:
https://trello.com/c/aeQmXJSa/1802-1843219-application-status-not-always-correct-in-allwatcher-investigate-adding-tests-to-pylib-and-juju

state/allwatcher.go
```diff
@@ -554,18 +554,19 @@ func (app *backingApplication) updated(st *State, store *multiwatcherStore, id s
 		}
 	} else {
 		// The entry already exists, so preserve the current status.
-		oldInfo := oldInfo.(*multiwatcher.ApplicationInfo)
-		info.Constraints = oldInfo.Constraints
-		info.WorkloadVersion = oldInfo.WorkloadVersion
-		if info.CharmURL == oldInfo.CharmURL {
+		appInfo := oldInfo.(*multiwatcher.ApplicationInfo)
```
jameinel: I definitely prefer not reusing an outer and inner definition of a variable.

SimonRichardson:
$$merge$$

SimonRichardson:
Fixes: https://github.com/juju/python-libjuju/issues/352

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

I've run the failing test locally and can't reproduce it, so I'll give another try.

PR #10622, opened 2019-09-11
During Uniter teardown, don't wait for BlockUntilLeadershipReleased

jameinel:
## Description of change

The APIServer isn't tearing down, and the cancel channel doesn't work on
API calls. Instead, we just let the goroutine die on its own.

This only triggers when we have an application with multiple units, and
the Uniter agent for a unit that is not the leader wants to restart.

## QA steps

I'm still working on adding a 'wrench' that will let us kill the uniter on demand.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1843376

### DIFF
--- a/worker/leadership/tracker.go
+++ b/worker/leadership/tracker.go
@@ -65,11 +65,6 @@ func NewTracker(tag names.UnitTag, claimer leadership.Claimer, clock clock.Clock
 			for _, ticketCh := range t.waitingMinion {
 				close(ticketCh)
 			}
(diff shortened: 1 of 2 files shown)


howbazaar:
$$merge$$

howbazaar:
@manadart, @jameinel and I discussed the more extensive fixes of providing cancellation through the API client interface. We agreed that this change is a bit too invasive for a 2.6 fix, but plan to get this into 2.7.

PR #10624, opened 2019-09-12
Add fingerprint to docker resources

wallyworld:
## Description of change

Add a fingerprint to k8s docker resources. This is needed to trigger a charm change notification when a new resource is juju attached.
The client deploy CLL was sending the docker data as a json string, but this breaks the fingerprint checks since the resource is subsequently stored as YAML. So the CLI now sends as YAML with a backwards compatibility check.

## QA steps

bootstrap
deploy k8s charm
juju attach a new docker image resource
verify that upgrade-charm hook runs
do the same with a 2.6 client

## Bug reference

https://bugs.launchpad.net/bugs/1843470


wallyworld:
$$merge$$

PR #10626, opened 2019-09-12
2.6 uniter wrench on idle

jameinel:
## Description of change

When the Uniter goes into Idle state, we allow a '/var/lib/juju/wrench/UNIT' file to indicate that the Uniter should go into an error state if 'error-on-idle' is set. This allows us to test the restart behavior of the Uniter.

Also, this cleans up a bunch of log messages from Wrench. They used to report:
```
  wrench for ubuntu-0/error-on-idle is active
```
However, that looks like a directory path (/var/lib/juju/wrench/ubuntu-0/error-on-idle). However, it is actually not the file that needs to be created but the content of the file. So I switched to ':' as I thought it would be clearer.

## QA steps

```
$ juju bootstrap lxd lxd
$ juju model-config logging-config="<root>=DEBUG;juju.wrench=TRACE"
$ juju deploy cs:~jameinel/ubuntu-lite -n2
$ juju ssh 1
$$ sudo su -
$$ mkdir -p /var/lib/juju/wrench
$$ echo "error-on-idle" > /var/lib/juju/wrench/unit-ubuntu-lite-1
$$ tail -f /var/log/juju/unit-ubuntu-lite-1.log
```

You may need to restart the unit agent to have it go *back* into idle. But when it does, you should end up seeing a message in the error log that 'wrench thrown for ubuntu-lite/1', and `juju status` should report a 'failed' state for that unit.

## Documentation changes

None.

## Bug reference

None.

### DIFF
--- a/worker/uniter/uniter.go
+++ b/worker/uniter/uniter.go
@@ -43,6 +43,7 @@ import (
 	"github.com/juju/juju/worker/uniter/runner/jujuc"
 	"github.com/juju/juju/worker/uniter/storage"
 	"github.com/juju/juju/worker/uniter/upgradeseries"
+	"github.com/juju/juju/wrench"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


worker/uniter/uniter.go
```diff
@@ -314,6 +315,9 @@ func (u *Uniter) loop(unitTag names.UnitTag) (err error) {
 			// error state.
 			return nil
 		}
+		if wrench.IsActive(u.unit.Tag().String(), "error-on-idle") {
+			return errors.Errorf("wrench thrown for %s", u.unit.Name())
```
hpidcock: Does this need a test?

PR #10631, opened 2019-09-13
Cmr consuming types

SimonRichardson:
## Description of change

The following PR introduces the relation network to the
importing and exporting nomenclature of juju.

Interestingly I had to change a few things to get this to work,

 1. Introduction of RelationNetworkDirection type, which helps
    with knowing if we should use ingress or egress for the
    relation network.
 2. Parsing of the relation network type docID. This is required
    because we don't know if we should be admin or not, it's
    only ever stored in the docID (it seems), so we need to
    grab it out of the docID.

Add import and export functions to get the models into state:

 - [ ] ExternalController
 - [x] RelationNetworks
 - [ ] FirewallRule
 - [ ] RemoteEntity

## QA steps

TBA

## Documentation changes

TBA


Gopkg.toml
```diff
@@ -62,8 +62,9 @@
   name = "github.com/juju/bundlechanges"
 
 [[constraint]]
-  revision = "7eee51ddb1f5033244c6501ec83b399af9d23861"
+  revision = "2fb1f4b5169c5cb20fbf341b87b7392c99b0d55e"
   name = "github.com/juju/description"
+  source = "github.com/SimonRichardson/description"
```
SimonRichardson: This is while I'm developing, will remove once PR is finalised.

PR #10634, opened 2019-09-16
Protect sends with a check of tomb.Dying()

jameinel:
## Description of change

We were doing bare sends on the out channel if we saw a Changes(), and
if we raced with a tomb.Dying the goroutine could block and never exit.

This was just by inspection, it is unlikely to be visible in the wild,
since you only leak a goroutine if you get a controller change at the
same time you're shutting down.

This is just a drive by cleanup.

## QA steps

None, isn't likely to be an obvservable change.

## Documentation changes

None.

## Bug reference

None.


SimonRichardson:
$$merge$$

PR #10636, opened 2019-09-16
Show and list cloud commands with ask-or-tell.

anastasiamac:
## Description of change

This PR changes:
* 'juju clouds';
* 'juju regions';
* 'juju credentials';
* 'juju show-clouds'; 
* 'juju show-credential';

to prompt user if the detected current controller needs to be used when listing/showing information.

As a drive-by, ensure that 'juju regions' command works on remote clouds, bug https://bugs.launchpad.net/juju/+bug/1843455

### DIFF
--- a/cmd/juju/cloud/list.go
+++ b/cmd/juju/cloud/list.go
@@ -4,6 +4,7 @@
 package cloud
 
 import (
+	"fmt"
 	"io"
 	"sort"
 
@@ -20,7 +21,6 @@ import (
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/cmd/output"
-	"github.com/juju/juju/feature"
 	"github.com/juju/juju/jujuclient"
 )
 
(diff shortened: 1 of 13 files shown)


anastasiamac:
[ask-or-tell: list and show cloud pkg commands](https://trello.com/c/xzZZgQyg/913-ask-or-tell-list-and-show-cloud-pkg-commands)

cmd/juju/cloud/list_test.go
```diff
@@ -164,43 +173,35 @@ func (s *listSuite) assertListTabular(c *gc.C, expectedOutput string) {
...
+	c.Assert(out, jc.Contains, expectedOutput)
 }
 
 func (s *listSuite) TestListTabular(c *gc.C) {
 	s.assertListTabular(c, `
-Cloud    Regions  Default  Type       Description
-antnest        1  default  openstack  
-beehive        1  default  k8s        
-
+Cloud           Regions  Default          Type        Description
```
wallyworld: So save repeating the description, we can print "from controller foo" at the top of the table. There will only ever be one controller, especially if we do  not include local results unless asked for. If we do go with allowing both local and controller if specified by the user, we can print them in separate tables; the controller and local clouds do serve different purposes and it makes sense to look at them separately.

cmd/juju/cloud/listcredentials.go
```diff
@@ -246,11 +254,23 @@ func (c *listCredentialsCommand) Run(ctxt *cmd.Context) error {
...
+		if c.cloudName != "" {
+			cloudMsg = fmt.Sprintf("for cloud %q ", c.cloudName)
+		}
+		c.ControllerName, err = c.MaybePromptCurrentController(ctxt, fmt.Sprintf("list credentials %vfrom", cloudMsg))
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+	}
+	if c.ControllerName == "" {
+		return nil, errors.New("Not listing credentials from a controller: no controller specified.")
```
wallyworld: can we make the messages and behaviour consistent with list clouds?

cmd/juju/cloud/listcredentials.go
```diff
@@ -400,8 +422,8 @@ func formatCredentialsTabular(writer io.Writer, value interface{}) error {
 			w.Println()
 		}
 	}
-	printGroup(credentials.Remote, nil)
-	printGroup(credentials.Local, ansiterm.Foreground(ansiterm.BrightBlue))
+	printGroup(credentials.Remote, ansiterm.Foreground(ansiterm.BrightBlue))
```
wallyworld: i think local creds should stay blue?

cmd/juju/cloud/listcredentials.go
anastasiamac: Hmm, it kind of is ... In 'list-clouds' we specify 
```
No controllers were specified.
You can bootstrap a new controller using one of these clouds:...
```
which is specific to 'clouds' command. So each command has a consistent message to same similar thing - 'no controller specified. not listing/showing blah'

cmd/juju/cloud/list.go
```diff
@@ -318,10 +321,14 @@ func formatCloudsTabular(writer io.Writer, value interface{}) error {
 			w.Println(len(info.Regions), defaultRegion, displayCloudType(info.CloudType), description)
 		}
 	}
+	personalColor := ansiterm.Foreground(ansiterm.BrightBlue)
```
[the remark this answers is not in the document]
anastasiamac: Sounds good! 

What do you think we should pick?

Here is the list of all possible colors, ignoring the Black and the Default:

```
	Red
	Green
	Yellow
	Blue
	Magenta
	Cyan
	Gray
	DarkGray
	BrightRed
	BrightGreen
	BrightYellow
	BrightBlue
	BrightMagenta
	BrightCyan
	White
```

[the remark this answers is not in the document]
anastasiamac: I like the idea of Green or BrightGreen :D

anastasiamac:
This has been alive too long and the surface difference as well as specification for this PR keeps changing. I'll close it and will re-propose in different, digestible chunks.

PR #10639, opened 2019-09-17
vsphere: Don't acquire the mutex when bootstrapping

babbageclunk:
## Description of change

The mutex in vsphereclient is designed to prevent multiple models' provisioners from clobbering each other when creating template VMs - it doesn't need to be done on the local machine when creating the initial bootstrap instance.

## QA steps

* Bootstrap on vsphere.
* Ensure that no lock file named `juju-juju-vsphere-bionic` or `juju-vsphere-bionic` is created under `/tmp`.

## Documentation changes
None

## Bug reference
None

### DIFF
--- a/provider/vsphere/internal/vsphereclient/createvm.go
+++ b/provider/vsphere/internal/vsphereclient/createvm.go
@@ -111,7 +111,15 @@ type CreateVirtualMachineParams struct {
 	// EnableDiskUUID controls whether the VMware disk should expose a
 	// consistent UUID to the guest OS.
 	EnableDiskUUID bool
+
+	// IsBootstrap indicates whether the requested instance will be a
+	// newly bootstrapped controller.
+	IsBootstrap bool
 }
+
+// MutexAcquirer wraps mutex.Acquire so we can test the mutex
+// handling.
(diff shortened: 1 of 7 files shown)


provider/vsphere/internal/vsphereclient/createvm.go
```diff
@@ -170,22 +189,22 @@ func (c *Client) ensureTemplateVM(
...
-	if err != nil {
-		return nil, errors.Annotate(err, "acquiring lock")
+	if !args.IsBootstrap {
+		// Each controller maintains its own image cache. All compute
+		// provisioners (i.e. each model's) run on the same controller
+		// machine, so taking a machine lock ensures that only one
+		// process is updating VMDKs at the same time. We lock around
+		// access to the series directory.
+		release, err := acquireMutex(mutex.Spec{
+			Name:  "juju-vsphere-" + args.Series,
```
jameinel: we want to remove the 'juju-' prefix because it is automatically added by the mutex code.

babbageclunk:
$$merge$$

babbageclunk:
That was lucky - the merge run failed because it dep complained about names v2 vs v3, since I was targeting the wrong branch! Fixed now.

$$merge$$

PR #10640, opened 2019-09-18
vsphere: Don't expose acquireMutex outside vsphereclient

babbageclunk:
## Description of change

Rather than passing `vsphereclient.AcquireMutex` into CreateVirtualMachine (requiring the provider code to know about it), store it in the client struct, set it to the default acquire function in vsphereclient.Dial and override it in the tests.

This is much cleaner - I'm blaming jetlag for doing it the other way first. Doesn't change the behaviour so it doesn't need to be included in the release - just tidies up the code.

## QA steps

* Bootstrap on vsphere
* Doesn't create a lock file /tmp/juju-vsphere-bionic on the local machine
* Delete the template vm in vsphere
* Deploy and application
* Does create the lock file on the controller machine

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/provider/vsphere/client.go
+++ b/provider/vsphere/client.go
@@ -23,7 +23,7 @@ type Client interface {
 	Close(context.Context) error
 	ComputeResources(context.Context) ([]*mo.ComputeResource, error)
 	ResourcePools(context.Context, string) ([]*object.ResourcePool, error)
-	CreateVirtualMachine(context.Context, vsphereclient.MutexAcquirer, vsphereclient.CreateVirtualMachineParams) (*mo.VirtualMachine, error)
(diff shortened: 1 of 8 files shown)


jameinel:
https://jenkins.juju.canonical.com/job/github-juju-check-jobs/773/console

I don't see why that is failing.

jameinel:
!!build!!

jameinel:
I missed the actual test failure which was:
FAIL: machine_charms_test.go:53: MachineWithCharmsSuite.TestManageModelRunsCharmRevisionUpdater
which failed because of:
[LOG] 0:10.078 DEBUG juju.worker.dependency "http-server" manifold worker stopped: listen tcp :41292: bind: address already in use

babbageclunk:
Thanks @jameinel!

$$merge$$

PR #10641, opened 2019-09-18
Merge 2.6 20190918

wallyworld:
## Description of change

Merge 2.6 branch pulling in these commits:

#10639 don't acquire mutex when bootstrapping
#10640 don't acquire mutex outside vsphere client


wallyworld:
$$merge$$

PR #10643, opened 2019-09-19
Fix Block Device Matching and Device Path regex

hpidcock:
# Description of change

LP1790728: Match on just block device serial when bus id is missing.
LP1830326: Extend device path regex to match against partitions.

## QA steps

Bootstrap OpenStack and Deploy postgresql charm with storage. Application should have storage attached successfully.

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1790728
https://bugs.launchpad.net/juju/+bug/1830326

### DIFF
--- a/apiserver/common/storagecommon/blockdevices.go
+++ b/apiserver/common/storagecommon/blockdevices.go
@@ -4,6 +4,8 @@
 package storagecommon
 
 import (
+	"strings"
+
 	"github.com/juju/loggo"
 
 	"github.com/juju/juju/state"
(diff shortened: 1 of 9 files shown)


worker/diskmanager/lsblk_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"os"
 	"strings"
+	"fmt"
```
wallyworld: this looks like it's in the wrong spot

apiserver/common/storagecommon/blockdevices.go
```diff
@@ -76,6 +79,14 @@ func MatchingBlockDevice(
 			logger.Tracef("no match for block device hardware id: %v", dev.HardwareId)
 			continue
 		}
+		if volumeInfo.VolumeId != "" && dev.SerialId != "" {
```
wallyworld: needs a test

worker/diskmanager/lsblk_test.go
hpidcock: 😱

hpidcock:
$$merge$$

hpidcock:
$$merge$$

PR #10647, opened 2019-09-20
state: Clean up apps stuck in dying state if the model is force-destroyed

babbageclunk:
## Description of change

If a model is force-destroyed while an application is stuck in a dying state (for example if there's a bug in CMR that means that a remote unit is still in scope for a relation), ensure that we correctly propagate the destruction down to the dying application in the cleanup.

## QA steps
I haven't come up with a simple way to reproduce this situation, other than commenting out the leave scope code in the remote relations worker facade and in the remote application terminate operation method.
* Create a cross-model relation.
* Remove the consuming application - because of the broken leave-scope logic it won't go away.
* Destroy the model with --force - it will successfully remove it.

## Documentation changes
None

## Bug reference
None


wallyworld:
$$merge$$

PR #10648, opened 2019-09-20
'juju help model-defaults' had wrong information.

jameinel:
## Description of change

It never supported a model, only a controller, so update the help text
to show that in the examples.

## QA steps

```
$ juju help model-defaults
```

## Documentation changes

None

## Bug reference

None


jameinel:
$$merge$$

PR #10652, opened 2019-09-24
add sorting for instanceTypes by Name (aws, openstack, google, azure)

nammn:
## Description of change

When a nonexistent constraint is given juju-cli returns possible constraint.
Those are unsorted and provider dependent.
This PR addresses this by sorting the output, which should make parsing/human reading easier.

The new sort, type sort `ByType` sorts by the first delimiter, if they are the same it sorts by the cost. Should be seen as best effort. 
Openstack uses its own `instancetypes` kind of struct, which has no cost associated. That`s why a normal lexical sort is used. To have a more fine graned sorting, a own sort has to be implemented which should parse the delimiter and then sort by cpu, ram and finally disk. 

## QA steps
old: `juju bootstrap aws local --constraints "instance-type=foo"` returns unsorted output.
new: `juju bootstrap aws local --constraints "instance-type=foo"` returns sorted output. 

e.g. output:
aws
```
 [a1.4xlarge a1.2xlarge a1.xlarge a1.large a1.medium c1.xlarge c1.medium c3.8xlarge c3.4xlarge c3.2xlarge c3.xlarge c3.large ...]
```
openstack
```
 [cpu1-ram1-disk10 cpu1-ram1-disk20 cpu1-ram1-disk50 cpu1-ram12-disk10 cpu1-ram12-disk20 cpu1-ram12-disk50 cpu1-ram16-disk10 cpu1-ram16-disk20 cpu1-ram16-disk50 cpu1-ram2-disk10 cpu1-ram2-disk20 cpu1-ram2-disk50 cpu1-ram4-disk10 cpu1-ram4-disk20...]
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1764963

### DIFF
--- a/environs/instances/instancetype.go
+++ b/environs/instances/instancetype.go
@@ -212,10 +212,10 @@ func (s byMemory) Less(i, j int) bool {
 	return inst0.Cost > inst1.Cost
 }
 
-//by Type is used to sort a slice by name by best effort. As we have different separators for different providers
-//we do it by best effort using "-" and "."
-//A possible sort could be:
-//a1.2xlarge, a1.large, lexical sort would put a1.2xlarge before a1.large -- aws
+// ByName is used to sort a slice by name by best effort. As we have different separators for different providers
+// A possible sort could be:
+// We sort by using a lexical sort for the type, which is before the delimiter,
+// and if they are the same, we sort by using the cost
 
 type ByName []InstanceType
 
--- a/environs/instances/instancetype.go
+++ b/environs/instances/instancetype.go
@@ -6,6 +6,7 @@ package instances
 import (
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/juju/juju/core/constraints"
 )
@@ -209,4 +210,30 @@ func (s byMemory) Less(i, j int) bool {
 	// Memory is equal, so use cost as a tie breaker.
 	// Result is in descending order of cost so instance with lowest cost is used.
 	return inst0.Cost > inst1.Cost
+}
+
+//by Type is used to sort a slice by name by best effort. As we have different separators for different providers
+//we do it by best effort using "-" and "."
+//A possible sort could be:
+//a1.2xlarge, a1.large, lexical sort would put a1.2xlarge before a1.large -- aws
+
+type ByName []InstanceType
+
+func (bt ByName) Len() int      { return len(bt) }
+func (bt ByName) Swap(i, j int) { bt[i], bt[j] = bt[j], bt[i] }
+func (bt ByName) Less(i, j int) bool {
+	inst0, inst1 := &bt[i], &bt[j]
+	baseInst0 := strings.FieldsFunc(inst0.Name, splitDelimiters)
+	baseInst1 := strings.FieldsFunc(inst1.Name, splitDelimiters)
+	if baseInst0[0] != baseInst1[0] {
+		return baseInst0[0] < baseInst1[0]
+	}
+	// Name is equal, so use cost as a tie breaker.
+	// Result is in ascending order of cost so instance with lowest cost is first.
+	return inst0.Cost > inst1.Cost
+
+}
+
+func splitDelimiters(r rune) bool {
+	return r == ',' || r == '-' || r == '.'
 }
--- a/environs/instances/instancetype.go
+++ b/environs/instances/instancetype.go
@@ -230,7 +230,7 @@ func (bt ByName) Less(i, j int) bool {
 	}
 	// Name is equal, so use cost as a tie breaker.
 	// Result is in ascending order of cost so instance with lowest cost is first.
-	return inst0.Cost > inst1.Cost
+	return inst0.Cost < inst1.Cost
 
 }
 
--- a/provider/azure/environ.go
+++ b/provider/azure/environ.go
@@ -371,12 +371,11 @@ func (env *azureEnviron) ConstraintsValidator(ctx context.ProviderCallContext) (
 	if err != nil {
 		return nil, err
 	}
-	instTypes := convertInstanceTypeMapToSlice(instanceTypes)
-	sort.Sort(instances.ByName(instTypes))
-	instTypeNames := make([]string, len(instTypes))
-	for i, itype := range instTypes {
-		instTypeNames[i] = itype.Name
+	instTypeNames := make([]string, 0, len(instanceTypes))
+	for instTypeName := range instanceTypes {
+		instTypeNames = append(instTypeNames, instTypeName)
 	}
+	sort.Strings(instTypeNames)
 
 	validator := constraints.NewValidator()
 	validator.RegisterUnsupported([]string{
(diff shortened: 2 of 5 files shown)


nammn:
I should probably add a test for the custom sort, what do you guys think?

environs/instances/instancetype.go
```diff
@@ -210,3 +211,29 @@ func (s byMemory) Less(i, j int) bool {
...
+		return baseInst0[0] < baseInst1[0]
+	}
+	// Name is equal, so use cost as a tie breaker.
+	// Result is in descending order of cost so instance with lowest cost is first.
+	return inst0.Cost > inst1.Cost
+
+}
+
+func splitDelimiters(r rune) bool {
+	return r == ',' || r == '-' || r == '.'
```
nammn: I choose multiple delimiters for multiple providers. Initially this was for openstack as the single output contains delimiters by "-" and ".".

environs/instances/instancetype.go
```diff
@@ -210,3 +211,29 @@ func (s byMemory) Less(i, j int) bool {
...
+
+type ByName []InstanceType
+
+func (bt ByName) Len() int      { return len(bt) }
+func (bt ByName) Swap(i, j int) { bt[i], bt[j] = bt[j], bt[i] }
+func (bt ByName) Less(i, j int) bool {
+	inst0, inst1 := &bt[i], &bt[j]
+	baseInst0 := strings.FieldsFunc(inst0.Name, splitDelimiters)
+	baseInst1 := strings.FieldsFunc(inst1.Name, splitDelimiters)
+	if baseInst0[0] != baseInst1[0] {
```
SimonRichardson: Can you be sure that `baseInst0` and `baseInst0` are greater than 0, if the `.Name` of either is empty, then this will panic for slice bound issue. You should add a check there.

provider/ec2/environ.go
```diff
@@ -180,10 +181,12 @@ func (e *environ) ConstraintsValidator(ctx context.ProviderCallContext) (constra
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
+	sort.Sort(instances.ByName(instanceTypes))
```
SimonRichardson: A test for this might be worth while.

environs/instances/instancetype.go
```diff
@@ -210,3 +211,29 @@ func (s byMemory) Less(i, j int) bool {
...
+		return baseInst0[0] < baseInst1[0]
+	}
+	// Name is equal, so use cost as a tie breaker.
+	// Result is in descending order of cost so instance with lowest cost is first.
```
manadart: Lowest first would be _ascending_, yes?

environs/instances/instancetype.go
```diff
@@ -210,3 +211,29 @@ func (s byMemory) Less(i, j int) bool {
 	// Result is in descending order of cost so instance with lowest cost is used.
 	return inst0.Cost > inst1.Cost
 }
+
+//by Type is used to sort a slice by name by best effort. As we have different separators for different providers
+//we do it by best effort using "-" and "."
+//A possible sort could be:
+//a1.2xlarge, a1.large, lexical sort would put a1.2xlarge before a1.large -- aws
+
+type ByName []InstanceType
```
manadart: Need some tests for this. Would be nice to get a few different provider samples.

provider/openstack/provider.go
```diff
@@ -562,6 +563,7 @@ func (e *Environ) ConstraintsValidator(ctx context.ProviderCallContext) (constra
 	for i, flavor := range flavors {
 		instTypeNames[i] = flavor.Name
 	}
+	sort.Strings(instTypeNames)
```
manadart: I would just use the same sort as for ec2. Without delimiters it should be the same anyway.
Might as well add it to the other providers too.

nammn: As they are not using the "instanceTypes" types I would need to add this to the underlying one of "flavor". Is that what you expect?

environs/instances/instancetype.go
nammn: go strings package: `If s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s.`
Tested if locally, even if nothing is found the return is always a len(slice) == 1.

provider/ec2/environ.go
nammn: :+1:

environs/instances/instancetype.go
nammn: :+1:

environs/instances/instancetype.go
nammn: Oh, updating it :+1:

nammn:
!!build!!

environs/instances/instancetype.go
nammn: Added some test cases.
I did add changes to:
azure, aws, openstack and google.

provider/azure/environ.go
```diff
@@ -371,11 +371,12 @@ func (env *azureEnviron) ConstraintsValidator(ctx context.ProviderCallContext) (
 	if err != nil {
 		return nil, err
 	}
-	instTypeNames := make([]string, 0, len(instanceTypes))
-	for instTypeName := range instanceTypes {
-		instTypeNames = append(instTypeNames, instTypeName)
+	instTypes := convertInstanceTypeMapToSlice(instanceTypes)
```
nammn: @manadart did you mean it like this? So I exchanged the existing azure implementation to use the newly added sorting. 
For this, to work I had to convert the map[string]InstanceType to a slice[]instanceType

provider/ec2/environ.go
nammn: @SimonRichardson  added a test for the instance-type place itself or did you mean in the provider package itself?

SimonRichardson: I meant testing the output from that, along with the sort.

provider/openstack/provider.go
manadart: Missed that. 
We *could* convert the flavours to `InstanceTypes`, but I think it's fine to leave it as it is.

environs/instances/instancetype.go
```diff
@@ -210,3 +211,29 @@ func (s byMemory) Less(i, j int) bool {
 	// Result is in descending order of cost so instance with lowest cost is used.
 	return inst0.Cost > inst1.Cost
 }
+
+//by Type is used to sort a slice by name by best effort. As we have different separators for different providers
```
manadart: "ByName" is used...

Nit: can we get a space between the slashes and the text?

nammn: Oh thanks, I missed that!

nammn:
!!build!!

provider/azure/environ.go
```diff
@@ -402,6 +405,25 @@ func (env *azureEnviron) ConstraintsValidator(ctx context.ProviderCallContext) (
 	return validator, nil
 }
 
+func getSliceOfAliasAndInstanceTypes(instanceTypes map[string]instances.InstanceType) ([]instances.InstanceType, []string) {
```
nammn: @manadart Azure has constraint aliases. So that `standard_D2` can also be called `D2`.
So I had two choices:
- revert the change and just let it be
- split the map into its components (instances Slice, alias slice) and add them together
  - advantage: uses the common sorting logic
  - disadvantage: more code and logic added 

What step would you prefer?

manadart: OK, let's leave this one as it was. We're getting fiddly without gaining much.

nammn:
$$merge$$

nammn:
[invalid instance type should sort the valid instance types](https://trello.com/c/ULA617sa/1826-invalid-instance-type-should-sort-the-valid-instance-types)

PR #10660, opened 2019-09-27
Add logging context for workers.

howbazaar:
In the past workers that ran on the controller had all the logging go into the log collection of the controller model. That information was also written into the machine-x.log file.

This PR adds a new logging context for model workers started by the controller on behalf of a model.

This logging context has two log writers:
 * a file based writer that goes to /var/log/juju/models/<uuid>.log
 * a db log writer that writes into the model's log collection

## QA steps

 * Bootstrap a controller
 * deploy something into the default model, I used ubuntu-lite
 * add a few models and deploy things into those models too

SSH into the controller machine and check the /var/log/juju directory. Look at the logs in the appropriate model's log files.

Use the debug-log command on the client to see that the log entries ended up in the right model's collection.

## Documentation changes

No real documentation changes needed.

## Bug reference

I'm sure there is one somewhere.

### DIFF
--- a/cmd/jujud/agent/model/manifolds.go
+++ b/cmd/jujud/agent/model/manifolds.go
@@ -46,6 +46,7 @@ import (
 	"github.com/juju/juju/worker/lifeflag"
 	"github.com/juju/juju/worker/logforwarder"
 	"github.com/juju/juju/worker/logforwarder/sinks"
+	"github.com/juju/juju/worker/logger"
 	"github.com/juju/juju/worker/machineundertaker"
 	"github.com/juju/juju/worker/metricworker"
 	"github.com/juju/juju/worker/migrationflag"
@@ -80,6 +81,11 @@ type ManifoldsConfig struct {
 	// Clock supplies timing services to any manifolds that need them.
 	// Only a few workers have been converted to use them fo far.
 	Clock clock.Clock
+
+	// LoggingContext holds the model writers so that the loggers
+	// for the workers running on behalf of other models get their logs
+	// written into the model's logging collection rather than the controller's.
+	LoggingContext *loggo.Context
 
 	// InstPollerAggregationDelay is the delay before sending a batch of
 	// requests in the instancepoller.Worker's aggregate loop.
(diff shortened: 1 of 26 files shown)


worker/modelworkermanager/shim.go
```diff
@@ -4,19 +4,36 @@ package modelworkermanager
...
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	defer ps.Release()
+	return state.NewDbLogger(ps), nil
+}
+
+// Config is part of the Controller interface.
+func (g StatePoolController) Config() (controller.Config, error) {
+	return g.StatePool.SystemState().ControllerConfig()
```
hpidcock: SystemState looks like it can return nil.

cmd/jujud/agent/model/manifolds.go
```diff
@@ -140,6 +146,15 @@ func commonManifolds(config ManifoldsConfig) dependency.Manifolds {
 			Filter:        apiConnectFilter,
 		}),
 
+		// The logging config updater listents for logging config updates
```
hpidcock: listents -> listens

controller/config.go
```diff
@@ -600,6 +618,19 @@ func (c Config) AllowModelAccess() bool {
...
+// ModelLogfileMaxBackups is the number of old model log files to keep (compressed).
+func (c Config) ModelLogfileMaxBackups() int {
+	return c.intOrDefault(ModelLogfileMaxBackups, DefaultModelLogfileMaxBackups)
+}
+
+// ModelLogfileMaxSizeMB is the maximum size of the log file written out by the
+// controller on behalf of workers running for a model.
+func (c Config) ModelLogfileMaxSizeMB() int {
+	// Value has already been validated.
+	val, _ := utils.ParseSize(c.mustString(ModelLogfileMaxSize))
```
hpidcock: Does this need defaulting?

worker/modelworkermanager/shim.go
howbazaar: Fair enough. Updated method.

controller/config.go
howbazaar: No, because the configuration validation process has pushed though the schema checker, which adds a default value if one isn't set.

howbazaar:
$$merge$$

PR #10665, opened 2019-09-27
configcommand: change key checking logic

nammn:
## Description of change
Rewrite error-handling message, as the error message can be too cryptic.
As well as refactoring the methods.

Old check flow:
- is file?
- it is a key of the model -> weird error if neither

Current check flow:
- is file?
- is key of the model?
- is file like?

Wanted check flow:
- is key of model?
- is file like?
- is file?

Problem is that with the current code the order actually matters. To achieve the wanted check flow more changes has to be done.
## QA steps
- added an additional unit test
old output:
```
[14:08:16] nam:juju git:(ede251dd0a) $ juju model-config bundlesk8s-dsds       
ERROR key "bundlesk8s-dsds" not found in {<nil> ""} model.
[14:08:28] nam:juju git:(ede251dd0a) $ juju model-config bundlesk8s-dsds.json
ERROR key "bundlesk8s-dsds.json" not found in {<nil> ""} model.
[14:08:33] nam:juju git:(ede251dd0a) $ juju model-config bundlesk8s-dsds.yaml  
ERROR key "bundlesk8s-dsds.yaml" not found in {<nil> ""} model.

```

new output:
```
[14:06:53] nam:juju git:(model-config-msg) $ juju model-config bundlesk8s-dsds       
ERROR value: "bundlesk8s-dsds" seems to be neither a file nor a key of the currently targeted model: {<nil> ""}
[14:06:56] nam:juju git:(model-config-msg) $ juju model-config bundlesk8s-dsds.yaml
ERROR value: "bundlesk8s-dsds.yaml" seems to be a file but not found
[14:07:03] nam:juju git:(model-config-msg) $ juju model-config bundlesk8s-dsds.json
ERROR value: "bundlesk8s-dsds.json" seems to be a file but not found
```
## Documentation changes

nothing should change
## Bug reference
https://bugs.launchpad.net/juju/+bug/1843456

### DIFF
--- a/cmd/juju/model/configcommand.go
+++ b/cmd/juju/model/configcommand.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -304,7 +305,7 @@ func (c *configCommand) getConfig(client configCommandAPI, ctx *cmd.Context) err
 		_, _ = ctx.Stdout.Write(certBytes)
 		return nil
 	}
-	attrs, err := c.getSkimmedModel(client)
+	attrs, err := c.getFilteredModel(client)
 	if err != nil {
 		return err
 	}
@@ -315,15 +316,15 @@ func (c *configCommand) getConfig(client configCommandAPI, ctx *cmd.Context) err
 		return c.out.Write(ctx, attrs)
 	} else if len(c.keys) > 0 && !finished {
 		if isFileLike(c.keys[0]) {
-			return errors.Errorf("value: %q seems to be a file but not found", c.keys[0])
+			return errors.Errorf("%q seems to be a file but not found", c.keys[0])
 		} else {
-			return errors.Errorf("value: %q seems to be neither a file nor a key of the currently targeted model: %q", c.keys[0], attrs["name"])
+			return errors.Errorf("%q seems to be neither a file nor a key of the currently targeted model: %q", c.keys[0], attrs["name"])
 		}
 	}
 	return nil
 }
 
-func (c *configCommand) getSkimmedModel(client configCommandAPI) (config.ConfigValues, error) {
+func (c *configCommand) getFilteredModel(client configCommandAPI) (config.ConfigValues, error) {
 	attrs, err := client.ModelGetWithMetadata()
 	if err != nil {
 		return nil, err
@@ -338,10 +339,15 @@ func (c *configCommand) getSkimmedModel(client configCommandAPI) (config.ConfigV
 }
 
 func isFileLike(fileLike string) bool {
-	if strings.HasSuffix(fileLike, ".yaml") || strings.HasSuffix(fileLike, ".json") {
-		return true
+	r, _ := regexp.Compile("\\.[a-zA-Z]{0,4}$")
+	match := r.FindString(fileLike)
+	if strings.HasSuffix(match, ".") {
+		return false
 	}
-	return false
+	if match == "" || match == " " {
+		return false
+	}
+	return true
 }
 
 func (c *configCommand) handleIsKeyOfModel(attrs config.ConfigValues, ctx *cmd.Context) (config.ConfigValues, error, bool) {
--- a/cmd/juju/model/configcommand.go
+++ b/cmd/juju/model/configcommand.go
@@ -301,22 +301,50 @@ func (c *configCommand) setConfig(client configCommandAPI, ctx *cmd.Context) err
 // get writes the value of a single key or the full output for the model to the cmd.Context.
 func (c *configCommand) getConfig(client configCommandAPI, ctx *cmd.Context) error {
 	if len(c.keys) == 1 && certBytes != nil {
-		ctx.Stdout.Write(certBytes)
+		_, _ = ctx.Stdout.Write(certBytes)
 		return nil
 	}
-	attrs, err := client.ModelGetWithMetadata()
+	attrs, err := c.getSkimmedModel(client)
 	if err != nil {
 		return err
 	}
+	attrs, err, finished := c.handleIsKeyOfModel(attrs, ctx)
+	if err != nil {
+		return err
+	} else if attrs != nil && finished {
+		return c.out.Write(ctx, attrs)
+	} else if len(c.keys) > 0 && !finished {
+		if isFileLike(c.keys[0]) {
+			return errors.Errorf("value: %q seems to be a file but not found", c.keys[0])
+		} else {
+			return errors.Errorf("value: %q seems to be neither a file nor a key of the currently targeted model: %q", c.keys[0], attrs["name"])
+		}
+	}
+	return nil
+}
 
+func (c *configCommand) getSkimmedModel(client configCommandAPI) (config.ConfigValues, error) {
+	attrs, err := client.ModelGetWithMetadata()
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


nammn:
[model-config prints a cryptic error message if a file was not found](https://trello.com/c/noVDUWF0/1822-model-config-prints-a-cryptic-error-message-if-a-file-was-not-found)

cmd/juju/model/configcommand.go
```diff
@@ -301,43 +301,63 @@ func (c *configCommand) setConfig(client configCommandAPI, ctx *cmd.Context) err
...
 	if err != nil {
 		return err
 	}
+	attrs, err, finished := c.handleIsKeyOfModel(attrs, ctx)
+	if err != nil {
+		return err
+	} else if attrs != nil && finished {
+		return c.out.Write(ctx, attrs)
+	} else if len(c.keys) > 0 && !finished {
+		if isFileLike(c.keys[0]) {
```
nammn: @mitechie so I was not 100% sure if there can be cases where the "keyname" could actually be looking like a yml file. That's why I did the check here and not above in https://github.com/juju/juju/pull/10665/files#diff-fc38c887834ec43ba3bdea51be58fd06L157-R160

I do think that, if the keyname could not be fileLike, I would move it from here to above. So ordering makes a little more sense in this flow.

cmd/juju/model/configcommand.go
```diff
@@ -301,43 +301,63 @@ func (c *configCommand) setConfig(client configCommandAPI, ctx *cmd.Context) err
...
+			return errors.Errorf("value: %q seems to be a file but not found", c.keys[0])
+		} else {
+			return errors.Errorf("value: %q seems to be neither a file nor a key of the currently targeted model: %q", c.keys[0], attrs["name"])
```
nammn: so the main idea of this PR was to refactor the main method into smaller ones.

cmd/juju/model/configcommand.go
```diff
@@ -301,43 +301,63 @@ func (c *configCommand) setConfig(client configCommandAPI, ctx *cmd.Context) err
...
 		return err
 	}
+	attrs, err, finished := c.handleIsKeyOfModel(attrs, ctx)
+	if err != nil {
+		return err
+	} else if attrs != nil && finished {
+		return c.out.Write(ctx, attrs)
+	} else if len(c.keys) > 0 && !finished {
+		if isFileLike(c.keys[0]) {
+			return errors.Errorf("value: %q seems to be a file but not found", c.keys[0])
```
mitechie: can we remove the "value: " part of the message please.

cmd/juju/model/configcommand.go
```diff
@@ -301,43 +301,63 @@ func (c *configCommand) setConfig(client configCommandAPI, ctx *cmd.Context) err
...
+		} else {
+			return errors.Errorf("value: %q seems to be neither a file nor a key of the currently targeted model: %q", c.keys[0], attrs["name"])
+		}
+	}
+	return nil
+}
 
+func (c *configCommand) getSkimmedModel(client configCommandAPI) (config.ConfigValues, error) {
```
mitechie: what's a "skimmed" model?

cmd/juju/model/configcommand.go
```diff
@@ -301,43 +301,63 @@ func (c *configCommand) setConfig(client configCommandAPI, ctx *cmd.Context) err
...
-		// via show-model.
+		// We don't want model attributes included, these are available via show-model.
 		if c.isModelAttribute(attrName) {
 			delete(attrs, attrName)
 		}
 	}
+	return attrs, nil
+}
+
+func isFileLike(fileLike string) bool {
```
mitechie: can we check what we do around this for things like juju config and controller-config as well. In theory all config, model-config, and controller-config should be similar.

cmd/juju/model/configcommand.go
```diff
@@ -301,43 +301,63 @@ func (c *configCommand) setConfig(client configCommandAPI, ctx *cmd.Context) err
...
+	if strings.HasSuffix(fileLike, ".yaml") || strings.HasSuffix(fileLike, ".json") {
```
mitechie: given that this is keys in mongodb I don't know that we support things with a period in them at all. It might be ok just to check for a 3 or 4 char extension as something "file like"

cmd/juju/model/configcommand_test.go
```diff
@@ -106,7 +106,15 @@ func (s *ConfigCommandSuite) TestSingleValueOutputFile(c *gc.C) {
...
 func (s *ConfigCommandSuite) TestGetUnknownValue(c *gc.C) {
 	context, err := s.run(c, "unknown")
-	c.Assert(err, gc.ErrorMatches, `key "unknown" not found in {<nil> ""} model.`)
+	c.Assert(err, gc.ErrorMatches, `value: "unknown" seems to be neither a file nor a key of the currently targeted model: {<nil> \"\"}`)
+
+	output := cmdtesting.Stdout(context)
+	c.Assert(output, gc.Equals, "")
+}
+
+func (s *ConfigCommandSuite) TestGetProperErrorMessageOnPaths(c *gc.C) {
```
mitechie: if we go the route of the isFileLike should that have a series of unit tests for things we expect to be file-like vs not?

nammn: hmm, yes makes sense! Iam taking a further look into this

cmd/juju/model/configcommand.go
nammn: You mean something along the line of `strings.HasSuffix(fileLike,".{[a-Z]+}")` instead of checking for yaml and JSON?

cmd/juju/model/configcommand.go
nammn: :+1:

cmd/juju/model/configcommand.go
nammn: :+1:

cmd/juju/model/configcommand.go
nammn: In my mind, I wanted to refactor and extra the part where we get the `modelMetadata` and later remove some model-related things which we don't want to show the user in the output.
That's why I named it skimmed. Do you have a better name in mind? @mitechie

mitechie: maybe just filtered vs skimmed?

cmd/juju/model/configcommand.go
mitechie: Yea, maybe even capping that filter at 0-4 chars?

cmd/juju/model/configcommand_test.go
mitechie: ty

cmd/juju/model/configcommand.go
nammn: So I tried to regex through and could not find anything special. 
It seems to be there is no "file-like" fuzzy thing. The other config do it by checking whether the file exists. There is seems to be no fuzziness around.
But maybe I overlooked something?

cmd/juju/model/configcommand_test.go
nammn: Added 2 additional tests. Others are already testing partly the functionality

cmd/juju/model/configcommand.go
nammn: done

nammn:
$$merge$$

nammn:
thanks for the review!

PR #10667, opened 2019-09-27
GitHub actions - Client side testing with mongo

SimonRichardson:
## Description of change

The following commits add ubuntu and macOS client tests to the github
actions setup. Unfortunately windows isn't as simple as I first thought, so
some serious thinking about how to do it efficiently is going to be needed.

I've split the tests up so we have pre-checking aka "static analysis", this
should be fine to run concurrently and gives you need feedback.

The other file `client-tests.yml` will check if the clients have been broken
at any point. We could and probably should prevent concurrent running
of these and allow github to cancel these if there is a new PR. I'll look into
that next week.

### DIFF
--- a/.github/workflows/client-tests.yml
+++ b/.github/workflows/client-tests.yml
@@ -8,7 +8,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macOS-latest]
+        # macOS-latest is disabled because there is an issue of the tests timing
+        # out. No effort has been done to work out why they currently timeout
+        # on macOS, but we should investigate that to improve our coverage on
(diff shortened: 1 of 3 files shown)


SimonRichardson:
Turns out I need to install mongo directly, I'll speak @jameinel on Monday about this.

SimonRichardson:
!!build!!

SimonRichardson:
!!build!!

SimonRichardson:
!!build!!

SimonRichardson:
!!build!!

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

PR #10675, opened 2019-10-02
Example test suite

SimonRichardson:
## Description of change

The following is an example test suite to help others create
their own test suites. It should now be a case of paint by
numbers to create their own, if that's not the case I should
look into why and help fill in the gaps.

## QA steps

```
cd tests
./main.sh examples
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ vendor/
 juju-backup-*.tar.gz
 .editorconfig
 scripts/mgo-run-txn/mgo-run-txn
+tests/tmp.*
--- a/tests/suites/examples/README.md
+++ b/tests/suites/examples/README.md
@@ -2,6 +2,12 @@
 
 The following is a very quick overview of what each file does and why.
 
+```console
+ - task.sh           -> suite registration, bootstraps
+ - example.sh        -> test runner and tests
+ - other.sh          -> test runner and other tests
+```
+
 ## task.sh
 
 `task.sh` file is the entry point to the suite. You can bootstrap a juju
--- a/tests/suites/examples/README.md
+++ b/tests/suites/examples/README.md
@@ -1,0 +1,20 @@
+# Example Test Suite
+
+The following is a very quick overview of what each file does and why.
+
+## task.sh
+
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -24,3 +24,4 @@ vendor/
 juju-backup-*.tar.gz
 .editorconfig
 scripts/mgo-run-txn/mgo-run-txn
+tests/tmp.*
```
SimonRichardson: If a test fails, then we leave these around for investigation, but can be added to juju repo by accident.

SimonRichardson:
https://trello.com/c/JnLGZAR4/1868-add-example-test-suite-shell-tests

tests/suites/examples/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+Registration of tests are also done in this file.
+
+## example.sh
+
+`example.sh` describes a test in the test suite. Generally the grouping of tests
+inside of a file should be of the same ilk.
+
+## other.sh
+
```
nammn: Would be great if we could just add a sample tree layout as kind of tl'dr:
Something like a general strucutre.
```
- test
-- task.sh -> entrypoint, creates controller
-- example.sh -> test itself, each test can be used with ensure to create a model
-- something -> ?
```

And use the tree for explaining.
What do you think?

nammn: Do we want to add document some common functions the lib provided?
E.g.
- ensure
- bootstrap
- check_contains

tests/suites/examples/example.sh
```diff
@@ -0,0 +1,40 @@
+run_example1() {
+    # Echo out to ensure nice output to the test suite.
+    echo
+
+    # The following ensures that a bootstrap juju exists
+    file="${TEST_DIR}/test-example1.txt"
+    ensure "example1" "${file}"
```
nammn: from outside (I do know because you told me) it can be unclear what e.g. ensure actually does.
From context, one could guess creating a model. It would be cooler to be more explicit in the example.

SimonRichardson:
$$merge$$

PR #10676, opened 2019-10-02
Introspect controller on destroy

SimonRichardson:
## Description of change

The following attempts to collate as much information about a controller
when destroying it, so that we can better understand how the controller
behaves when under testing.

To run you need to have specified an artifact otherwise the artifact is
cleaned up. You can do this with the following setup:

     ./main.sh -a output.tar.gz smoke

All files will be namespaced per controller.

## QA steps

```sh
./main.sh -a output.tar.gz smoke
```

### DIFF
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -152,6 +152,14 @@ destroy_controller() {
         return
     fi
 
+    echo "====> Introspection gathering"
+
+    set +e
+    introspect_controller "${name}" || true
+    set_verbosity
+
+    echo "====> Introspection gathered"
+
     output="${TEST_DIR}/${name}-destroy-controller.txt"
 
     echo "====> Destroying juju ($(green "${name}"))"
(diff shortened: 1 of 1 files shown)


tests/includes/juju.sh
```diff
@@ -189,3 +197,12 @@ wait_for() {
         sleep 5
     done
 }
+
+introspect_controller() {
+    local name
+
+    name=${1}
+
+    juju ssh -m controller 0 "source /etc/profile.d/juju-introspection.sh && juju_engine_report" > "${TEST_DIR}/${name}-juju_engine_report.txt"
```
SimonRichardson: We have to force the inclusion of the script here, because it's not loaded unless you do a full login i.e. `juju ssh -m controller 0` vs `juju ssh -m controller 0 bash`. The later doesn't load anything in `profile.d`...

SimonRichardson:
https://trello.com/c/oTxFJhBi/1867-introspect-controller-on-destroy

SimonRichardson:
$$merge$$

tests/includes/juju.sh
```diff
@@ -189,3 +197,12 @@ wait_for() {
         sleep 5
     done
 }
+
+introspect_controller() {
+    local name
+
+    name=${1}
+
+    juju ssh -m controller 0 bash -lic "juju_engine_report" > "${TEST_DIR}/${name}-juju_engine_report.txt"
```
howbazaar: This seems a bit flakey given that it is possible that the controller doesn't have a machine 0, or may be in HA mode.

SimonRichardson: What options do we have, quite happy to change... Also if it fails it doesn't cause the destroy to fail.

PR #10677, opened 2019-10-02
Fix the recording of api host ports in controller collection.

howbazaar:
For any provider that understands spaces, and records spaces with the network addresses, the recording of the controller api host ports became broken in 2.6.2 with the addition of the space provider id to the addresses stored with the machines. Since the space provider id wasn't stored with the api host ports, the sorting method that is used for equality never had items match as it uses the stringified network host port, which has ip@space if no id is present and id@(space:id) if it is.

This means that every minute this gets updated by the peer grouper (because of a separate bug). The peergrouper should only be telling the DB when it changes, but the peergrouper is iterating over a dict in random order every minute, so for a single node, nothing changes, but in HA, the machines are likely to have a different order every minute causing the call to the state package to record them. Since the arrays never match, they are recorded every minute.

When this document is updated it triggers updates for the ProxyUpdater to update as the controller host addresses are added to the noproxy values.

On a large deployment this causes every agent to call back every minute, and this puts a lot of load on the controllers, which may also be why we are seeing mongo elections mid openstack deploy on maas. 

## QA steps

* bootstrap into guimaas with 2.6.8
* enable-ha
* juju model-config -m controller logging-config="juju=info;juju.state.txn=trace"

Observe that the apiHostPorts is being updated every minute

* upgrade the controller with the built agent

Observe the host ports are no longer being updated every minute.

### DIFF
--- a/state/address.go
+++ b/state/address.go
@@ -395,22 +395,24 @@ func networkAddresses(addrs []address) []network.Address {
 // stuff at some point. We want to use juju-specific network names
 // that point to existing documents in the networks collection.
 type hostPort struct {
-	Value       string `bson:"value"`
-	AddressType string `bson:"addresstype"`
-	Scope       string `bson:"networkscope,omitempty"`
-	Port        int    `bson:"port"`
-	SpaceName   string `bson:"spacename,omitempty"`
+	Value           string `bson:"value"`
+	AddressType     string `bson:"addresstype"`
+	Scope           string `bson:"networkscope,omitempty"`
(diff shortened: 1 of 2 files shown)


howbazaar:
$$merge$$

howbazaar:
$$merge$$ intermittent watcher failure

jameinel:
For posterity, I filed:
https://bugs.launchpad.net/juju/+bug/1847016
so we have a track of what we fixed here.

PR #10679, opened 2019-10-03
Forward port 2.6 test

howbazaar:
Bring the new api host port test from the 2.6 branch. This shows that the problem isn't in the 2.7 branch any more as the network changes have already fixed the problem here.


howbazaar:
$$merge$$

PR #10686, opened 2019-10-07
Fix intermittent failures in httpserver package.

howbazaar:
Address two intermittent failures in the httpserver tests.

The first is the wait for the quick request to be processed. Under a machine with heavy load, this sometimes takes longer than 50ms. See https://jenkins.juju.canonical.com/view/unit%20tests/job/RunUnittests-amd64-bionic/1573/testReport/junit/github/com_juju_juju_worker_httpserver/TestPackage/

The second race is where the quick endpoint is hit too soon after the kill that the event is processed successfully, but doesn't return before the 50ms timeout which causes the test to assume that it is in fact blocked.

While there is no way to provide definite synchronisation between the worker and the test without other call back functions, the fix here is twofold: first, add a small sleep before the request check, and secondly, we don't actually care what the error is that was returned.

### DIFF
--- a/worker/httpserver/worker_test.go
+++ b/worker/httpserver/worker_test.go
@@ -285,27 +285,29 @@ func (s *WorkerSuite) TestHeldListener(c *gc.C) {
 	// Stop the server.
 	s.worker.Kill()
 
-	// Eventually quick requests get blocked by the held listener.
-	var quickBlocked bool
 	// A very small sleep will allow the kill to be more likely to be processed
 	// by the running loop.
 	time.Sleep(10 * time.Millisecond)
-attempts:
+
 	// We actually try the quick request more than once, on the off chance
 	// that the worker hasn't finished processing the kill signal. Since we have
 	// no other way to check, we just try a quick request, and decide that if
 	// it doesn't respond quickly, the main loop is waithing for the clients to
 	// be done.
-	for i := 0; i < 5; i++ {
+	quickBlocked := false
+	timeout := time.After(coretesting.LongWait)
+	for !quickBlocked {
+		c.Log("try to hit the quick endpoint")
 		go request()
 		select {
 		case <-quickErr:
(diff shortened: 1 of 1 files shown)


worker/httpserver/worker_test.go
```diff
@@ -308,9 +316,13 @@ attempts:
...
-	case err := <-quickErr:
+	case <-quickErr:
+		// There is a race in the queueing of the request. It is possible that
+		// the timer will fire a short wait before an unheld request gets to the
+		// phase where it would return nil. However this is only under significant
+		// load, and it isn't easy to synchronise. This is why we don't actually
+		// check the error.
 		// It doesn't really matter what the error is.
-		c.Assert(err, gc.NotNil)
 	case <-time.After(coretesting.ShortWait):
```
babbageclunk: Should this be a LongWait as well, just in case? It's definitely in the not-expected path.

worker/httpserver/worker_test.go
```diff
@@ -287,8 +287,16 @@ func (s *WorkerSuite) TestHeldListener(c *gc.C) {
...
+	// by the running loop.
+	time.Sleep(10 * time.Millisecond)
 attempts:
-	for a := coretesting.LongAttempt.Start(); a.Next(); {
+	// We actually try the quick request more than once, on the off chance
+	// that the worker hasn't finished processing the kill signal. Since we have
+	// no other way to check, we just try a quick request, and decide that if
+	// it doesn't respond quickly, the main loop is waithing for the clients to
+	// be done.
+	for i := 0; i < 5; i++ {
```
babbageclunk: I think it would be better to still have this try for up to LongWait (even if you don't use LongAttempt).

babbageclunk: Oops - disregard obvs.

howbazaar:
$$merge$$

PR #10687, opened 2019-10-07
Pass logger into storage provisioner worker

howbazaar:
Update the storage provisioner worker to get the clock and logger from the manifold config. This means we have the logger from the right logging context when operating in the controller.


howbazaar:
$$merge$$

PR #10688, opened 2019-10-07
Tests can expect a numeric action id

wallyworld:
## Description of change

The network health CI tests were failing because they were checking for an action id that is a UUID. Juju now produces numeric action ids.

## QA steps

Run the network health test on AWS and LXD.


wallyworld:
$$merge$$

PR #10690, opened 2019-10-07
Systemd uses 'infinity' where ulimit uses 'unlimited'

jameinel:
## Description of change

It is unclear why these use a different syntax, and we could just put
'infinity' in our Service definitions. However, this feels like a
'ulimit' vs 'systemd' thing, so lets encode the Systemd knowledge inside
the Systemd service renderer.

To make testing easier, I also sort the limit variables, otherwise we're iterating a map and they appear in random order.

## QA steps

```$ juju bootstrap
$ juju ssh -m controller 0
$$ sudo vim /lib/systemd/system/juju-db/juju-db.service
```
You should see things like:
```
LimitFS=infinity
```
Rather than
```
LimitFS=unlimited
```

You should also be able to verify this by using something like
```
$ ps -C mongod
# get the PID
$ cat /proc/$PID/limits
```

To see what the ulimits for that process is.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1845350

### DIFF
--- a/service/systemd/conf.go
+++ b/service/systemd/conf.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


jameinel:
closing in favor of just landing in 2.7 here:
https://github.com/juju/juju/pull/10698

PR #10692, opened 2019-10-08
Fix model-defaults functional test

babbageclunk:
## Description of change

This was failing in the region-specific defaults test for a couple of reasons: the region wasn't being correctly passed into get_new_model_config so the model wasn't being created in the specific region, and the new model was always being created with the bootstrap config which overrode the setting that we wanted to inherit.

An example of the failure: https://jenkins.juju.canonical.com/job/nw-model-defaults/1567/console

## QA steps

* Rerun the test with `aws --series xenial --region ca-central-1` and see that it passes.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/acceptancetests/assess_model_defaults.py
+++ b/acceptancetests/assess_model_defaults.py
@@ -23,16 +23,16 @@ __metaclass__ = type
 log = logging.getLogger('assess_model_defaults')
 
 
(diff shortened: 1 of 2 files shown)


babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #10693, opened 2019-10-08
acceptance tests: Tweak assess runner to run under python 3

babbageclunk:
## Description of change

Pass `universal_newlines=True` to subprocess calls so the returned outputs are strings rather than bytes. Use `safe_load_all` to placate a yaml warning. Also bubble the return code of the test up.

## QA steps

* Ran the model defaults functional test in a Python 3.7 virtualenv.

## Documentation changes

None

## Bug reference

None


babbageclunk:
!!build!!

babbageclunk:
$$merge$$

PR #10694, opened 2019-10-08
Thread the logger through the apicaller.

howbazaar:
Pass a logger into the apicaller manifold config, and use that logger internally.

### DIFF
--- a/cmd/jujud/agent/caasoperator/manifolds.go
+++ b/cmd/jujud/agent/caasoperator/manifolds.go
@@ -130,6 +130,7 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 			APIOpen:              api.Open,
 			APIConfigWatcherName: apiConfigWatcherName,
 			NewConnection:        apicaller.OnlyConnect,
+			Logger:               loggo.GetLogger("juju.worker.apicaller"),
 		}),
 
 		clockName: clockManifold(config.Clock),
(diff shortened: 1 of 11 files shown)


cmd/jujud/agent/caasoperator/manifolds.go
```diff
@@ -130,6 +130,7 @@ func Manifolds(config ManifoldsConfig) dependency.Manifolds {
 			APIOpen:              api.Open,
 			APIConfigWatcherName: apiConfigWatcherName,
 			NewConnection:        apicaller.OnlyConnect,
+			Logger:               loggo.GetLogger("juju.worker.apicaller"),
```
babbageclunk: So this one will still be a package level one? Oh, I guess this worker isn't model specific?

cmd/jujud/agent/model/manifolds.go
```diff
@@ -145,6 +145,7 @@ func commonManifolds(config ManifoldsConfig) dependency.Manifolds {
 			APIOpen:       api.Open,
 			NewConnection: apicaller.OnlyConnect,
 			Filter:        apiConnectFilter,
+			Logger:        config.LoggingContext.GetLogger("juju.worker.apicaller"),
```
babbageclunk: Right, and I guess this is the key part that gets the model-specific logger for this case.

howbazaar: Yup

howbazaar:
$$merge$$

PR #10699, opened 2019-10-08
change permissions: let group members read logs as well

nammn:
## Description of change

Adds read rights to group users. This should be seen as a complement PR to #10696, which adds the logs to the `adm` group.

## QA steps
- unit test updated
- should PR #10696  be merged first I can add an additional acceptance test
- manual:
  - bootstrap new controller
  - deploy something (ubuntu)
```
[17:35:40] nam:juju git:(add_read_rights) $ juju ssh 0 -- ls -la /var/log/juju                                                     
total 172
drwxr-xr-x  2 syslog adm     4096 Oct  8 15:35 .
drwxrwxr-x 10 root   syslog  4096 Oct  8 15:35 ..
-rw-r-----  1 syslog syslog 27701 Oct  8 15:35 machine-0.log
-rw-r--r--  1 root   root     861 Oct  8 15:36 machine-lock.log
-rw-r-----  1 syslog syslog 73491 Oct  8 15:36 unit-ntp-0.log
-rw-r-----  1 syslog syslog 60788 Oct  8 15:35 unit-ubuntu-0.log
```
## Bug reference

https://bugs.launchpad.net/juju/+bug/1679479

### DIFF
--- a/cloudconfig/userdatacfg_test.go
+++ b/cloudconfig/userdatacfg_test.go
@@ -353,7 +353,7 @@ echo 'Installing Juju machine agent'.*
 /var/lib/juju/tools/1\.2\.3-precise-amd64/jujud bootstrap-state --timeout 10m0s --data-dir '/var/lib/juju' --debug '/var/lib/juju/bootstrap-params'
 ln -s 1\.2\.3-precise-amd64 '/var/lib/juju/tools/machine-0'
 echo 'Starting Juju machine agent \(service jujud-machine-0\)'.*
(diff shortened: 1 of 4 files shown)


nammn:
Hmmm, I am thinking about extracting this centrally somewhere.. Do we have a good place for that?

mitechie:
Can you check in with an upgrade step to perform this on a unit upgrade so
that this is setup for already running controllers please?

howbazaar:
This doesn't actually fix the bug. It does set group read permissions, but the group is `syslog` when the bug asks for the group to be `adm`

nammn:
@howbazaar I actually separated this into 2 PR's. One for setting the permissions and one for changing the groups. 
https://github.com/juju/juju/pull/10696
Is probably confusing :confused: I can put them into one though.

nammn:
@mitechie you are right, right now only new machines have the correct steps. I will add upgrade steps to the pr :+1:

nammn:
Closing this one in favour of #10696, to put those pr's into one

PR #10704, opened 2019-10-09
vsphere: Change extend disk polling to investigate bug 1847245

babbageclunk:
## Description of change

We're getting reports that provisioning machines on vsphere fails if the root disk is extended.

I haven't been able to reproduce this locally, so making some changes in the hope that we can get the users hitting the problem to try it. 

* Bump up the timeout to 10m in case 30s isn't sufficient - we'll check much more frequently initially anyway. 
* Log of the current size on each failed check so we can see whether there's a problem with the way we're getting the disk size.

## QA steps

* Deploy a machine with a bigger root disk than the default 10G, make sure it still extends and check logging.

## Documentation changes

None

## Bug reference

### DIFF
--- a/provider/vsphere/internal/vsphereclient/client.go
+++ b/provider/vsphere/internal/vsphereclient/client.go
@@ -29,7 +29,7 @@ import (
 const (
 	// maxExtendWait is the longest we'll wait for a disk to be
 	// extended.
-	maxExtendWait = 30 * time.Second
+	maxExtendWait = 10 * time.Minute
 
 	// extendInitialDelay is the shortest time we'll wait for a disk
 	// extend to succeed.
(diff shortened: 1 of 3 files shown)


provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -29,15 +29,15 @@ import (
...
+	maxExtendWait = 10 * time.Minute
 
 	// extendInitialDelay is the shortest time we'll wait for a disk
 	// extend to succeed.
 	extendInitialDelay = 10 * time.Millisecond
 
 	// extendMaxDelay is the longest we'll wait before checking
 	// again whether the disk extend has succeeded.
-	extendMaxDelay = 5 * time.Second
+	extendMaxDelay = 30 * time.Second
```
babbageclunk: I increased this as well to avoid spamming the log every 5 seconds.

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #10705, opened 2019-10-09
Fix block device matching priority

hpidcock:
## Description of change

Fix block device matching priority. Previously priority was inconsistently applied/would skip devices that should have matched.

## QA steps

```
juju bootstrap --no-gui aws aws-ceph
juju deploy -n 3 ceph-mon
juju deploy -n 3 ceph-osd --storage osd-devices=2G,2 --storage osd-journals=3G,1
juju relate ceph-mon ceph-osd
```

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1847128

### DIFF
--- a/apiserver/common/storagecommon/blockdevices.go
+++ b/apiserver/common/storagecommon/blockdevices.go
@@ -41,77 +41,99 @@ func MatchingBlockDevice(
 	attachmentInfo state.VolumeAttachmentInfo,
 	planBlockInfo state.BlockDeviceInfo,
 ) (*state.BlockDeviceInfo, bool) {
-	logger.Tracef("looking for block device for volume %#v", volumeInfo)
-	for _, dev := range blockDevices {
-		if planBlockInfo.HardwareId != "" {
+	logger.Tracef("looking for block device to match one of planBlockInfo %#v volumeInfo %#v attachmentInfo %#v",
+		planBlockInfo, volumeInfo, attachmentInfo)
+
+	if planBlockInfo.HardwareId != "" {
+		for _, dev := range blockDevices {
 			if planBlockInfo.HardwareId == dev.HardwareId {
-				logger.Tracef("plan hwid match on %v", volumeInfo.HardwareId)
+				logger.Tracef("plan hwid match on %v", planBlockInfo.HardwareId)
 				return &dev, true
 			}
 		}
-		if planBlockInfo.WWN != "" {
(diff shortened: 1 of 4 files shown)


apiserver/common/storagecommon/blockdevices.go
```diff
@@ -41,77 +41,99 @@ func MatchingBlockDevice(
...
 			if attachmentInfo.BusAddress == dev.BusAddress {
 				logger.Tracef("bus address match on %v", attachmentInfo.BusAddress)
 				return &dev, true
 			}
-			logger.Tracef("no match for block device bus address: %v", dev.BusAddress)
-			continue
 		}
-		// Only match on block device link if the block device is published
-		// with device link information.
-		if attachmentInfo.DeviceLink != "" && len(dev.DeviceLinks) > 0 {
```
wallyworld: I think removing
len(dev.DeviceLinks) > 0
may lead to a bug. It was added to avoid "continue" when a check on device links failed because the dev didn't have any

apiserver/common/storagecommon/blockdevices.go
```diff
@@ -41,77 +41,99 @@ func MatchingBlockDevice(
...
+	}
+
+	if planBlockInfo.WWN != "" {
+		for _, dev := range blockDevices {
 			if planBlockInfo.WWN == dev.WWN {
-				logger.Tracef("plan wwn match on %v", volumeInfo.WWN)
+				logger.Tracef("plan wwn match on %v", planBlockInfo.WWN)
 				return &dev, true
 			}
-			continue
```
wallyworld: It seems the crux of the change in this PR is to remove a couple of places where we continue the loop without the chance performing other matches. I am not sure about the other refactoring to introduce several places where the block devices are iterated. Won't we fix the bug simply by omitting the continues and keeping the loop logic the same?

hpidcock: Yes but removing the continues affects the priority of the matching. By breaking out each match into their own loops, priority is maintained while also not skipping the lower priority match.

apiserver/common/storagecommon/blockdevices.go
hpidcock: Yep, but with the altered loop logic, this isn't required any more.

hpidcock:
$$merge$$

PR #10714, opened 2019-10-10
vsphere: Specify the target datastore when cloning the template

babbageclunk:
## Description of change

Cloning a template VM wasn't respecting the root-disk-source constraint. That meant the VM would be created in the same datastore as the template, which could be different from the requested datastore if the template already existed.

Pull the datastore selection up so that the selected datastore can be used when cloning the template.

## QA steps

* Bootstrap a vsphere controller with datastore1.
* Deploy an app with `--constraints "root-disk-source=datastore2" - the deployed VM should be on datastore 2.
* Deploy another one to datastore 3 and ensure that the new VM is in the right datastore.

## Documentation changes

None

## Bug reference

Fixes https://bugs.launchpad.net/juju/+bug/1847278

### DIFF
--- a/provider/vsphere/internal/vsphereclient/client.go
+++ b/provider/vsphere/internal/vsphereclient/client.go
@@ -459,6 +459,7 @@ func (c *Client) cloneVM(
 	args CreateVirtualMachineParams,
 	srcVM *object.VirtualMachine,
 	vmFolder *object.Folder,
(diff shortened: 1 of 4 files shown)


babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #10715, opened 2019-10-10
Make the controller config more robust.

howbazaar:
Controller config was somewhat brittle around missing string values. This change makes the code more robust, even though the values should be there.


howbazaar:
$$merge$$

howbazaar:
@nammn This branch needs to land before any upgrades will work. The upgrade step to add the missing values doesn't get to run before we try to read the values. This work makes the controller config more robust to missing values.

howbazaar:
$$merge$$

PR #10720, opened 2019-10-10
Set up TERM variable in hook environments

dshcherb:
## Description of change

In order for a terminal to work properly in a debug-hooks environment, TERM environment variable needs to be set. This change addresses that according to the tmux documentation https://github.com/tmux/tmux/wiki/FAQ in a cross-distribution way.

* `TERM=tmux-256color` is used for distributions that have "ncurses 5.9 - patch 20150502" (https://github.com/mirror/ncurses/commit/be512fa073c00c2d52567c973d16b121414870da).
* `TERM=screen-256color` is used for older distributions.

`tmux-256color` which uses `tmux` (which in turn uses `screen`)
https://github.com/mirror/ncurses/blob/be512fa073c00c2d52567c973d16b121414870da/misc/terminfo.src#L6124-L6129

`screen-256color` which uses `screen`.
https://github.com/mirror/ncurses/blob/be512fa073c00c2d52567c973d16b121414870da/misc/terminfo.src#L5962-L5963

http://man7.org/linux/man-pages/man5/terminfo.5.html (see "Similar Terminals" on that covers the `use` keyword)

## QA steps

```
juju deploy ubuntu
juju debug-hooks ubuntu/0

# once in the hook environment you should see the TERM variable set appropriately to "tmux-256color" or "screen-256color"
printenv TERM

# try to use `less`, it must NOT show this:

less config.yaml
WARNING: terminal is not fully functional
config.yaml  (press RETURN)
```

NOTE: just doing something like `juju run --machine 0 -m controller 'printenv TERM'` will not work because those environment variables are injected by Juju only for hook/action execution environments.

## Documentation changes

None

## Bug reference

https://bugs.launchpad.net/juju/+bug/1838772

### DIFF
--- a/worker/uniter/runner/context/env.go
+++ b/worker/uniter/runner/context/env.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 
 	jujuos "github.com/juju/os"
+	"github.com/juju/os/series"
 )
 
 // OSDependentEnvVars returns the OS-dependent environment variables that
@@ -22,6 +23,8 @@ func OSDependentEnvVars(paths Paths) []string {
 		return centosEnv(paths)
 	case jujuos.OpenSUSE:
 		return opensuseEnv(paths)
+	case jujuos.GenericLinux:
+		return genericLinuxEnv(paths)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


mitechie:
!!build!!

howbazaar:
@dshcherb code looks fine but the test needs to be updated

dshcherb:
@howbazaar 

Thanks for the review.

I modified the tests and added new ones as well for missing series. Given that there is now a version dependency I have to validate that the environment is generated properly for all supported series.

dshcherb:
@howbazaar ping.

howbazaar:
$$merge$$

PR #10722, opened 2019-10-11
For jujuv3, actions are called functions

wallyworld:
## Description of change

Behind the juju-v3 feature flag, adjust the actions CLI to refer instead to functions.
This includes command names and output.

## QA steps

Run juju help 'cmdname'
Run the commands and checkout output

### DIFF
--- a/cmd/juju/action/call.go
+++ b/cmd/juju/action/call.go
@@ -44,7 +44,7 @@ type callCommand struct {
 	api           APIClient
 	unitReceivers []string
 	leaders       map[string]string
-	actionName    string
+	functionName  string
 	paramsYAML    cmd.FileVar
 	parseStrings  bool
 	background    bool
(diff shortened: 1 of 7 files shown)


cmd/juju/action/call.go
```diff
@@ -54,28 +54,27 @@ type callCommand struct {
 }
 
 const callDoc = `
-Run an Action for execution on a given unit, with a given set of params.
-The Action ID is returned for use with 'juju show-action-output <ID>' or
-'juju show-action-status <ID>'.
+Run a function for execution on a given unit, with a given set of params.
```
howbazaar: Run a charm function?

cmd/juju/action/export_test.go
```diff
@@ -40,7 +40,7 @@ func (c *CallCommand) UnitNames() []string {
 }
 
 func (c *CallCommand) ActionName() string {
```
howbazaar: FunctionName?

cmd/juju/action/show.go
```diff
@@ -71,7 +74,11 @@ func (c *showCommand) Info() *cmd.Info {
 		Doc:     showActionDoc,
 	})
 	if featureflag.Enabled(feature.JujuV3) {
+		info.Name = "show-function"
```
howbazaar: Ick...

wallyworld: Yeah, but it's the same command, just with a different name. Achieves the result with very little code. So smaller amount of churn when flag is removed.

cmd/juju/action/call.go
wallyworld: yup

cmd/juju/action/export_test.go
wallyworld: yup

wallyworld:
$$merge$$

PR #10723, opened 2019-10-11
Construct correct target tag type in pubsub worker

wallyworld:
## Description of change

When controller nodes were introduced, a new controller agent tag was used to represent them. The pubsub worker had a hard coded machine tag which was used to see if a message needed to be forwarded (filtering not to send to self). We now need to account for the different tag types or else messages get sent back to the source in a tight loop.

## QA steps

deploy a k8s controller and a charm
check top
check logs

## Bug reference

https://bugs.launchpad.net/juju/+bug/1847145


wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10726, opened 2019-10-13
Moved extendDisk method to use vm.disk.change

phvalguima:

----

## Description of change

This change is a partial fix for LP #1847245.
It searches for the root disk of the VM and runs vm.Reconfigure method from govmomi rather than methods.ExtendVirtualDisk_Task to extend it.

This is a partial fix because GRUB fails to find root device after extend-disk process.

## QA steps

Deploy node with root-disk constraint set to a different value than VM template.
Run govc to recover VMs' disk size and compare with requested value.

## Documentation changes

None

## Bug reference

### DIFF
--- a/provider/vsphere/internal/vsphereclient/client.go
+++ b/provider/vsphere/internal/vsphereclient/client.go
@@ -19,11 +19,9 @@ import (
 	"github.com/vmware/govmomi/list"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/property"
-	"github.com/vmware/govmomi/vim25/methods"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/soap"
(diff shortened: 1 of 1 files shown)


phvalguima:
WIP: improve logs and resolve VMs not finding root disk

timClicks:
@phvalguima Thanks for helping us with this issue! It's really great to receive contributions from other teams.

mitechie:
!!build!!

babbageclunk:
This is doing essentially the same thing as using ExtendVirtualDisk but in a fiddlier way (since the Reconfigure method is more flexible). Now that the extend code can wait for the task rather than polling the disk size I don't think we need this change, hopefully you agree @phvalguima? (The grub error turned out to be a red herring - the VMs were being created with only 4Mb ram rather than the intended 4Gb, which was preventing them from booting.)

PR #10727, opened 2019-10-14
vsphere: Stop polling for disk extension

babbageclunk:
## Description of change

Users reported that specifying a non-default root-disk size caused provisioning the VM to fail - it would timeout waiting for the extend disk task (even though the task could be seen to be finished in the UI).

The bug that caused govmomi to hang when waiting for the task extension has been fixed - it will now correctly propagate errors up to the caller. Stop polling the disk size, just wait for the extend task.

The hang was caused by a permission error when trying to get the state of the extend task - this can be avoided by granting the model user the Read-only role on the vcenter instance (with propagate to children turned on). This seems like overkill, and I'm still trying to determine if we can do it with a more specific permission, but I wanted to push this up so @phvalguima can try it out.

Once we determine the right permission I'll add a pre-check for it so we can provide a better error.

## QA steps

* Deploy an application with `--constraints "root-disk=20G"` - it should provision successfully.

## Documentation changes

* We'll need to call out the correct permission in the vsphere provider documentation.

## Bug reference

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1145,7 +1145,7 @@
   version = "v1.0.2"
 
 [[projects]]
-  digest = "1:e5396b66ff8c90a824413a27fd41ed4413161c949fa681869bd2caf714a155f9"
+  digest = "1:363c7e47cd3449872c15a27a9785900c2af3652a55e17487b1401031cd211202"
   name = "github.com/vmware/govmomi"
   packages = [
     ".",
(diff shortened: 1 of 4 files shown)


wallyworld:
$$merge$$

wallyworld:
$$merge$$

babbageclunk:
Gah, thanks @wallyworld - I had entered another merge command but not submitted it.

PR #10729, opened 2019-10-14
Handle BOOTSTRAP_REUSE_LOCAL

SimonRichardson:
## Description of change

The following allows us to completely reuse a local bootstrap,
so much so, that you can now test the integrations on your local
work. This also doesn't destroy your local controller, only the
models that where created whilst running the tests.

Along with these changes, we also get the debug logs out of the
bootstrapping controller, by passing `-V`. This has been an
outstanding issue of mine, so that it becomes easier to debug when
a controller doesn't work correctly.

As a driveby I've made sure that the CLI tests pass again.

## QA steps

```sh
juju deploy lxd test
BOOTSTRAP_REUSE_LOCAL=test ./main.sh cli
```

### DIFF
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -69,8 +69,14 @@ bootstrap() {
     fi
     bootstrapped_name=$(grep "." "${TEST_DIR}/jujus" | tail -n 1)
     if [ -z "${bootstrapped_name}" ]; then
-        # No bootstrapped juju found, unset the the variable.
-        unset BOOTSTRAP_REUSE
+        if [ -n "${BOOTSTRAP_REUSE_LOCAL}" ]; then
+            bootstrapped_name="${BOOTSTRAP_REUSE_LOCAL}"
+            export BOOTSTRAP_REUSE="true"
+        else
+            # No bootstrapped juju found, unset the the variable.
+            echo "====> Unable to reuse bootstrapped juju"
+            unset BOOTSTRAP_REUSE
+        fi
     fi
 
     version=$(juju_version)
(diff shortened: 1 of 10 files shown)


tests/includes/juju.sh
```diff
@@ -79,14 +85,14 @@ bootstrap() {
     if [ -n "${BOOTSTRAP_REUSE}" ]; then
         echo "====> Reusing bootstrapped juju ($(green "${version}"))"
 
-        OUT=$(juju models --format=json 2>/dev/null | jq '.models[] | .["short-name"]' | grep "${model}" || true)
+        OUT=$(juju models -c "${bootstrapped_name}" --format=json 2>/dev/null | jq '.models[] | .["short-name"]' | grep "${model}" || true)
         if [ -n "${OUT}" ]; then
             echo "${model} already exists. Use the following to clean up the environment:"
             echo "    juju destroy-model --force -y ${model}"
```
SimonRichardson: Interestingly `destroy-model` doesn't take a `-c`, which means this message could be wrong from time to time. Probably should change it to include `juju switch ${bootstrapped_name}`

SimonRichardson:
https://trello.com/c/QSQOtbqV/1898-handle-bootstrap-reuse-local

SimonRichardson:
$$merge$$

nammn:
@SimonRichardson I am a little late to the party, but would have loved to add your qa example into the README.md I can do that, or did I overlook something?
```
juju deploy lxd test
BOOTSTRAP_REUSE_LOCAL=test ./main.sh cli
```

tests/main.sh
```diff
@@ -3,6 +3,9 @@
 
 # Always ignore SC2230 ('which' is non-standard. Use builtin 'command -v' instead.)
 export SHELLCHECK_OPTS="-e SC2230 -e SC2039 -e SC2028 -e SC2002 -e SC2005"
+export BOOTSTRAP_REUSE_LOCAL=
```
nammn: Maybe there is a difference in our shell usage, but in my case running it will always lead to false and thus not reusing the old one.
```
~/golang/src/github.com/juju/juju/tests version_acceptancetest*
❯ juju controllers
Use --refresh option with this command to see the latest information.

Controller       Model                         User   Access     Cloud/Region        Models  Nodes    HA  Version
older*           local-charm-deploy-wrong-git  admin  superuser  localhost/default        4      1  none  2.7-beta1.1  
```

```
~/golang/src/github.com/juju/juju/tests version_acceptancetest*
❯ BOOTSTRAP_REUSE_LOCAL=older ./main.sh cli

==> Checking for dependencies
==> TEST BEGIN: test_cli
==> Checking for dependencies
====> Unable to reuse bootstrapped juju
```

Removing that line works for me

SimonRichardson:
I'll take a look tomorrow

PR #10730, opened 2019-10-14
Introducing client-only and controller-only options for cloud commands.

anastasiamac:
## Description of change

Newly introduced --client option if not enough for cloud based commands - CRUD for clouds, credentials and CAAS add and remove. We need the ability to allow operators to manage both client and controller copies of information as well as the ability to manage either one or the other without affecting the copy that was not explicitly stated.

This means that by default the commands will operate on both copies. However, when --client-only or --controller-only options are specified, the command will only deal with desired copy.

Since --client-copy is a simple rename of an existing option, it's fully functional. As --controller-only is a new option it's only introduced here, with full functionality taking affect as each command that requires it is modified to support it.

### DIFF
--- a/cmd/juju/caas/remove.go
+++ b/cmd/juju/caas/remove.go
@@ -55,7 +55,7 @@ type RemoveCAASCommand struct {
 // NewRemoveCAASCommand returns a command to add caas information.
 func NewRemoveCAASCommand(cloudMetadataStore CloudMetadataStore) cmd.Command {
 	store := jujuclient.NewFileClientStore()
-	cmd := &RemoveCAASCommand{
+	command := &RemoveCAASCommand{
 		OptionalControllerCommand: modelcmd.OptionalControllerCommand{
 			Store:       store,
 			EnabledFlag: feature.MultiCloud,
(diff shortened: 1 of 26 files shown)


anastasiamac:
[introduce --client-only and --controller-only flags](https://trello.com/c/8UbezpcP/933-introduce-client-only-and-controller-only-flags)

cmd/juju/caas/remove.go
```diff
@@ -98,7 +101,7 @@ func (c *RemoveCAASCommand) Init(args []string) (err error) {
 
 // Run is defined on the Command interface.
 func (c *RemoveCAASCommand) Run(ctxt *cmd.Context) error {
-	if c.ControllerName == "" && !c.Local {
+	if c.ControllerName == "" && !c.ClientOnly {
 		return errors.Errorf(
 			"There are no controllers running.\nTo remove cloud %q from the current client, use the --client option.", c.cloudName)
```
wallyworld: --client-only
(here, maybe elsewhere was well)

cmd/juju/cloud/listcredentials.go
```diff
@@ -391,6 +394,10 @@ func formatCredentialsTabular(writer io.Writer, value interface{}) error {
 					credentialNames = append(credentialNames, credentialName)
 				}
 			}
+			if len(credentialNames) == 0 {
+				w.Println(fmt.Sprintf("No credentials for cloud %v to display", cloudName))
```
wallyworld: Maybe reads better as
No credentials to display for cloud foo

cmd/juju/caas/remove.go
anastasiamac: True \o/ Changed. Thank you for the pick up :D

cmd/juju/cloud/listcredentials.go
anastasiamac: Changed. Thank you.

anastasiamac:
$$merge$$

PR #10731, opened 2019-10-14
Add proxy update worker to caas controllers

wallyworld:
## Description of change

Move the proxy update worker manifold from IAAS only to the common set for machine agents.
When run on a k8s controller, the worker ignores legacy non-Juju proxy values and also only updates in-memory use; it doesn't write out any systemd files etc.

## QA steps

bootstrap a k8s controller with a test value for the https proxy
$ juju bootstrap microk8s --config juju-https-proxy=10.0.0.1:8888 --debug

Try and add a model and it fails because it tries to use the proxy
Look at logs files to see that proxy is being set up

Also bootstrap an IAAS controller to check that proxy worker is fine there as well.

### DIFF
--- a/cmd/jujud/agent/machine/manifolds.go
+++ b/cmd/jujud/agent/machine/manifolds.go
@@ -256,6 +256,9 @@ type ManifoldsConfig struct {
 
 	// NewBrokerFunc is a function opens a instance broker (LXD/KVM)
 	NewBrokerFunc containerbroker.NewBrokerFunc
+
+	// IsCaasConfig is true if this config is for a caas agent.
+	IsCaasConfig bool
 }
 
 // commonManifolds returns a set of co-configured manifolds covering the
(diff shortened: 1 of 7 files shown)


wallyworld:
[proxy update worker for k8s controller](https://trello.com/c/PG3TUg6k/929-proxy-update-worker-for-k8s-controller)

cmd/jujud/agent/machine/manifolds.go
```diff
@@ -289,6 +292,11 @@ func commonManifolds(config ManifoldsConfig) dependency.Manifolds {
 		return crosscontroller.NewClient(conn), nil
 	}
 
+	var externalUpdateProxyFunc func(proxy.Settings) error
+	if runtime.GOOS == "linux" && !config.IsCaasConfig {
```
howbazaar: Is runtime.GOOS going to cause issues in tests? Normally we use os.HostOS()

wallyworld: There's a few other places where we use this instead. Using os.HostOS() seems to be for those times we want to check for == Windows. Here, we want linux (of any variant) and our os.HostOS() has many different linux variants, so we'd need a switch of something and then it gets fragile.

wallyworld:
$$merge$$

PR #10736, opened 2019-10-15
ASK-or_TELL: list clouds changes.

anastasiamac:
## Description of change

Several changes need to take place with 'juju clouds' command.

* In multi-cloud environment, users are also interested to see what clouds are available to them on the controller. These clouds are now listed first.

* For clouds that are available on the client side, users want to see clouds that are of use, for example, can be used to bootstrap a controller. By definition, these clouds  already have registered Juju credentials. So by default, the command now only lists clouds that have some credentials associated with it (these credentials number is also listed). If there are other clouds available, users can list them too with '--all' option. There is a message that will appear and guide the users.

* Ask-or-tell component of the change is applicable when a user did not specify a controller nor explicitly asked for --client-only clouds but the presence of a current controller was detected. In that instance, users will be prompted to confirm if the current controller is to be used.  For automated environments, --no-prompt option will automatically use a current controller when it's detected.

* Infrequently, users may not want to see both the clouds that are available on the client and the clouds that are available on the controller. In these instances, '--client-only' or '--controller-only' options can be used to filter one or the other part of the list.

* Previous implementation of the output had a pretty coloring for some clouds but the legend of what that color meant was hidden from the users. Coloring has been removed as it does not add clarity to the output.

## Documentation changes

All of the above

### DIFF
--- a/cmd/juju/cloud/list.go
+++ b/cmd/juju/cloud/list.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"sort"
 
-	"github.com/juju/ansiterm"
 	"github.com/juju/cmd"
 	"github.com/juju/errors"
 	"github.com/juju/gnuflag"
@@ -20,7 +19,6 @@ import (
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/cmd/output"
-	"github.com/juju/juju/feature"
 	"github.com/juju/juju/jujuclient"
 )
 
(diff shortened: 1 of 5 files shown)


anastasiamac:
[ask-or-tell: list clouds](https://trello.com/c/iHbWuzFC/937-ask-or-tell-list-clouds)

cmd/juju/cloud/list.go
```diff
@@ -242,10 +278,34 @@ func (c *cloudList) all() map[string]*CloudDetails {
...
+		return false
+	}
+	if len(c.personal) == 0 && len(c.builtin) == 0 && len(c.public) == 0 && len(c.remote) == 0 {
+		return false
+	}
+
+	result := false
+	examine := func(someClouds map[string]*CloudDetails) {
+		for name, cloud := range someClouds {
+			if cloud.CredentialCount == 0 {
```
wallyworld: Will this break the display of LXD and microk8s where there's no credential up front but we create one on the fly the first time.
This is also a breaking CLI change. Is this what we want?

cmd/juju/cloud/list.go
```diff
@@ -302,7 +360,7 @@ func formatCloudsTabular(writer io.Writer, value interface{}) error {
 		return cloudNames
 	}
 
-	printClouds := func(someClouds map[string]*CloudDetails, color *ansiterm.Context) {
```
wallyworld: The colour legend is not shown but for me it still adds significant value as it breaks up the wall of text. And it's not that difficult to discern looking at the output what the breakdown is. YMMV. DId we get agreement to remove the colour?

mitechie:
+1 I think the "don't show without credentials" might only fits for public clouds. Especially as to add a maas cloud, for instance, you first have to run add-cloud, then add-credential. Doing an add-cloud and then not seeing it in the list would be troubling.

cmd/juju/cloud/list.go
anastasiamac: Yes, this has been insisted on by both @jameinel and @mitechie as the most useful information that this command can produce. We have all agreed.

cmd/juju/cloud/list.go
anastasiamac: Color had no meaning for users beforehand (only devs that saw the code knew that there is a difference between public, built-in and personal clouds). Now with clearer division between client and controller clouds, it makes no sense.

anastasiamac:
@mitechie, sounds good. I'll filter out only "public" clouds based on the credentials.

cmd/juju/cloud/list.go
anastasiamac: The other bits was that there were 3 blocks displayed, each sorted alphabetically, but again that grouping made no sense. All client clouds are now displayed together and sorted together.

cmd/juju/cloud/list.go
wallyworld: You are right. I totally forgot that had been discussed.

cmd/juju/cloud/list.go
wallyworld: In which case my comment is totally irrelevant,

anastasiamac:
$$merge$$

PR #10737, opened 2019-10-16
vsphere: Check for System.Read privilege if root-disk constraint is set

babbageclunk:
## Description of change

If a machine has a non-default root-disk constraint, we'll need the global System.Read privilege to wait for the extend task to complete. Check that we have it in PrecheckInstance so that we can give early feedback to the user, rather than failing and retrying in the provisioner.

## QA steps

* Ensure the juju user doesn't have System.Read for the vsphere root (i.e. it hasn't been granted Administrator or ReadOnly)
* Deploy an application without a root-disk constraint - it works.
* Deploy one with root-disk - it's rejected with an error explaining what privilege should be granted.
* Grant ReadOnly to the juju user.
* Try the root-disk deploy again - it's accepted and the unit deploys successfully with the expanded disk.

## Documentation changes

* We should add a note to the vsphere provider docs explaining the privilege needed.

## Bug reference

Rest of the fix for: https://bugs.launchpad.net/juju/+bug/1847245

### DIFF
--- a/provider/vsphere/environ_policy.go
+++ b/provider/vsphere/environ_policy.go
@@ -30,7 +30,10 @@ func (env *sessionEnviron) PrecheckInstance(ctx context.ProviderCallContext, arg
 	if err := env.checkZones(ctx, args.Constraints.Zones); err != nil {
 		return errors.Trace(err)
 	}
-	err := env.checkDatastore(ctx, args.Constraints.RootDiskSource)
+	if err := env.checkDatastore(ctx, args.Constraints.RootDiskSource); err != nil {
+		return errors.Trace(err)
+	}
+	err := env.checkExtendPermissions(ctx, args.Constraints.RootDisk)
 	return errors.Trace(err)
 }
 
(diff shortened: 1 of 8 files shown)


provider/vsphere/environ_policy.go
```diff
@@ -71,6 +74,24 @@ func (env *sessionEnviron) checkDatastore(ctx context.ProviderCallContext, datas
...
+	// If we're going to need to resize the root disk we need to have
+	// the System.Read privilege on the root level folder - this seems
+	// to be because the extend disk task doesn't get created with a
+	// target, so the root-level permissions are applied.
+	ok, err := env.client.UserHasRootLevelPrivilege(env.ctx, "System.Read")
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if !ok {
+		return errors.Errorf("the System.Read privilege is required at the root level to extend disks - please grant the ReadOnly role to this model's user")
```
wallyworld: can we include the user name in the message?

babbageclunk: Yeah, good call - I'll see if that's accessible here.

babbageclunk: It was!

babbageclunk:
$$merge$$

wallyworld:
$$merge$$

babbageclunk:
Thanks @wallyworld - the check failures look like network dropouts while cloning the repo, the merge one was a failure in `util_unix_test.go:105: UniterSuite.TestRunCommand`

PR #10745, opened 2019-10-16
Thread logger through logforwarder and firewaller.

howbazaar:
More threading of model logging context through workers.


howbazaar:
$$merge$$

howbazaar:
$$merge$$ try again?

PR #10746, opened 2019-10-17
'show-cloud' ask-or-tell and exclusivity.

anastasiamac:
## Description of change

Several changes need to take place with 'juju show-cloud' command.

* In multi-cloud environment, users are interested to see details of the clouds from the controller as well as the client. The details of the clouds from the controller are now listed first.

* Ask-or-tell component of the change is applicable when a user did not specify a controller nor explicitly asked for --client-only cloud details but the presence of a current controller was detected. In that instance, users will be prompted to confirm if the current controller is to be used. For automated environments, --no-prompt option will automatically use a current controller when it's detected.

* Occasionally, users may not want to see both the cloud details from the client and the cloud details from the controller. In these instances, '--client-only' or '--controller-only' options can be used to filter one or the other.

## Documentation changes

All of the above.

### DIFF
--- a/cmd/juju/cloud/show.go
+++ b/cmd/juju/cloud/show.go
@@ -50,7 +50,7 @@ Use --controller option to show a cloud from a different controller.
 
 Use --client-only option to only show a cloud known locally on this client.
 
-Use --controller-only option to only show a cloud known remotely on the controller.
+Use --controller-only option to only show a cloud known on the controller.
 
 Examples:
 
@@ -172,7 +172,7 @@ func (c *showCloudCommand) Run(ctxt *cmd.Context) error {
 				ctxt.Warningf("%v", remoteErr)
 				displayErr = cmd.ErrSilent
 			} else {
-				ctxt.Infof("No cloud %q exists remotely on the controller.", c.CloudName)
+				ctxt.Infof("No cloud %q exists on the controller.", c.CloudName)
 			}
 		}
 	}
--- a/cmd/juju/cloud/show.go
+++ b/cmd/juju/cloud/show.go
@@ -17,7 +17,6 @@ import (
 	jujucmd "github.com/juju/juju/cmd"
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
-	"github.com/juju/juju/feature"
 	"github.com/juju/juju/jujuclient"
 )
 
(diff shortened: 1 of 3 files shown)


anastasiamac:
[ask-or-tell: show cloud](https://trello.com/c/ca7ZnaCG/939-ask-or-tell-show-cloud)

cmd/juju/cloud/show.go
```diff
@@ -103,56 +110,131 @@ func (c *showCloudCommand) Init(args []string) error {
...
+			if err := c.displayCloud(ctxt, remoteCloud, fmt.Sprintf("Cloud %q from controller %q:\n", c.CloudName, c.ControllerName), showRemoteConfig, remoteErr); err != nil {
+				ctxt.Warningf("%v", err)
+				displayErr = cmd.ErrSilent
+			}
+		} else {
+			if remoteErr != nil {
+				ctxt.Warningf("%v", remoteErr)
+				displayErr = cmd.ErrSilent
+			} else {
+				ctxt.Infof("No cloud %q exists remotely on the controller.", c.CloudName)
```
wallyworld: do we need the word "remotely" here. I think "exists on the controller" is sufficient and clearer

cmd/juju/cloud/show.go
```diff
@@ -103,56 +110,131 @@ func (c *showCloudCommand) Init(args []string) error {
...
-	displayCloud.CloudType = displayCloudType(displayCloud.CloudType)
-	if err := c.out.Write(ctxt, displayCloud); err != nil {
+	var displayErr error
+	showRemoteConfig := c.includeConfig
+	if localCloud != nil && remoteCloud != nil {
+		// It's possible that a local cloud named A is different to
+		// a remote cloud named A. If their types are different and we
+		// need to display config, we'd need to list each cloud type config.
+		// If the clouds' types are the same, we only need to list
+		// config once after the local cloud information.
```
wallyworld: Is that true? The types for client and controller cloud could both be "openstack" and config for each could be totally different. Or "maas" etc.
This also applies to public clouds like "aws" as I could have a local version with bespoke config which overrides the public version.

I may have missed it - it there a test for this scenario?

anastasiamac: It is true that for individual instances of a particular cloud type the configuration values are different. However, if you look at the code of this command we do not show the values but the metadata for the configuration based on the cloud type.

For example, for 'aws' which is of type 'ec2':
```
The available config options specific to ec2 clouds are:
vpc-id:
  type: string
  description: Use a specific AWS VPC ID (optional). When not specified, Juju requires
    a default VPC or EC2-Classic features to be available for the account/region.
vpc-id-force:
  type: bool
  description: Force Juju to use the AWS VPC ID specified with vpc-id, when it fails
    the minimum validation criteria. Not accepted without vpc-id
```

wallyworld: great, that makes sense \o/

anastasiamac:
$$merge$$

PR #10747, opened 2019-10-17
Add logger to unit assigner and instance poller.

howbazaar:
Not a lot else to say.


howbazaar:
I didn't spend any time cleaning up the instance poller because @achilleasa is working in that area.

howbazaar:
$$merge$$

PR #10748, opened 2019-10-17
'show-credential' ask-or-tell and exclusivity.

anastasiamac:
## Description of change

Several changes need to take place with 'juju show-credential' command.

* In multi-cloud environment, users are interested to see details of the credentials from the controller as well as the client. The details of the credential from the controller are now listed first.

* Ask-or-tell component of the change is applicable when a user did not specify a controller nor explicitly asked for --client-only credential details but the presence of a current controller was detected. In that instance, users will be prompted to confirm if the current controller is to be used. For automated environments, --no-prompt option will automatically use a current controller when it's detected.

* Occasionally, users may not want to see both the credential details from the client and the credential details from the controller. In these instances, '--client-only' or '--controller-only' options can be used to filter one or the other.

## Documentation changes

All of the above.

### DIFF
--- a/cmd/juju/cloud/export_test.go
+++ b/cmd/juju/cloud/export_test.go
@@ -176,7 +176,7 @@ func NewUpdateCredentialCommandForTest(testStore jujuclient.ClientStore, api Cre
 
 func NewShowCredentialCommandForTest(testStore jujuclient.ClientStore, api CredentialContentAPI) cmd.Command {
 	command := &showCredentialCommand{
-		store: testStore,
+		OptionalControllerCommand: modelcmd.OptionalControllerCommand{Store: testStore},
 		newAPIFunc: func() (CredentialContentAPI, error) {
 			return api, nil
 		},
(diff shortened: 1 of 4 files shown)


anastasiamac:
[ask-or-tell: show credential](https://trello.com/c/xzZZgQyg/913-ask-or-tell-show-credential)

anastasiamac:
$$merge$$

anastasiamac:
error: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.
16:47:59 fatal: The remote end hung up unexpectedly
16:47:59 fatal: early EOF
16:47:59 fatal: index-pack failed
16:47:59 : command failed: [git clone --recursive -v --progress https://go.googlesource.com/text /home/ubuntu/go/pkg/dep/sources/https---go.googlesource.com-text]: exit status 128
16:47:59 make: *** [dep] Error 1
16:47:59 Makefile:74: recipe for target 'dep' failed

PR #10754, opened 2019-10-18
List tasks command

wallyworld:
## Description of change

Add a new lists-tasks command. This command will display any tasks resulting from calling a charm function. Filtering can be done on target (application, unit), function name, and status.

Commit 1 is the api server facade implementation
Commit 2 is the action facade client
Commit 3 is the list-tasks CLI

## QA steps

deploy a charm with an action and add some units
set the juju-v3 feature flag
call the charm action a few times
use juju tasks with plain and yaml format with combinations of filter params

### DIFF
--- a/api/action/client.go
+++ b/api/action/client.go
@@ -35,6 +35,16 @@ func (c *Client) Actions(arg params.Entities) (params.ActionResults, error) {
 	return results, err
 }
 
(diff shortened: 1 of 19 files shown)


wallyworld:
$$merge$$

PR #10757, opened 2019-10-18
Add --watch support to show-task

wallyworld:
## Description of change

Add a --watch option to show-task. This is an alias for "--wait 0".
When a task is watched, an progress log messages are printed and when the task finishes, the result is printed. The current behaviour is just to wait for a result without printing any progress logs.

The watcher handling and test code is based on what is done in the call command.

As a drive by, fix an issue in the state action log watcher which causes messages to be spat out multiple times.
Also tweak the watcher handling in the call command to ensure it is always explicitly stopped.

## QA steps

deploy a charm with a long running action which uses action-log
call the action with --background
show-task <id> --watch
observe the log messages and the output when the task is done


wallyworld:
[show-task --watch](https://trello.com/c/IKYNCRPi/945-show-task-watch)

cmd/juju/action/showoutput.go
```diff
@@ -130,7 +160,39 @@ func (c *showOutputCommand) Run(ctx *cmd.Context) error {
...
+	var logsWatcher watcher.StringsWatcher
+	haveLogs := false
+
+	shouldWatch := waitDur.Nanoseconds() >= 0
+	if shouldWatch {
+		result, err := fetchResult(api, c.requestedId)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		shouldWatch = result.Status != params.ActionCompleted
```
hpidcock: Should this only watch on Status == ActionPending?

wallyworld: yes, yes it should

wallyworld:
$$merge$$

PR #10760, opened 2019-10-21
'add-k8s' ask-or-tell and exclusivity.

anastasiamac:
## Description of change

Several changes need to take place with 'juju add-k8s' command.

* Ask-or-tell component of the change is applicable when a user did not specify a controller nor explicitly asked for --client-only operation but the presence of a current controller was detected. In that instance, users will be prompted to confirm if the current controller is to be used. For automated environments, --no-prompt option will automatically use a current controller when it's detected.

* Occasionally, users may not want to add-k8s to both the client and a controller. In these instances, '--client-only' or '--controller-only' options can be used to filter one or the other.

Drive-by:
Removed multi-cloud feature flag.

## Documentation changes

all of the above.

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -27,7 +27,6 @@ import (
 	jujucmdcloud "github.com/juju/juju/cmd/juju/cloud"
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
-	"github.com/juju/juju/feature"
 	"github.com/juju/juju/jujuclient"
 )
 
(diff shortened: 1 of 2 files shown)


anastasiamac:
[add-k8s ask-or-tell and exlusivity](https://trello.com/c/xKOZwzy3/947-add-k8s-ask-or-tell-and-exlusivity)

cmd/juju/caas/add.go
```diff
@@ -715,6 +732,9 @@ func (c *AddCAASCommand) addCredentialToLocal(cloudName string, newCredential ju
 }
 
 func (c *AddCAASCommand) addCredentialToController(apiClient AddCloudAPI, newCredential jujucloud.Credential, credentialName string) error {
+	if c.ControllerName == "" {
+		return errors.New("Not adding k8s to the controller: no controller was specified.")
```
wallyworld: k8s cluster

anastasiamac:
$$merge$$

PR #10763, opened 2019-10-21
Use model owner and name for model logfile name on controller

howbazaar:
Instead of just having the model's log in the controller's log directory just using the model UUID, use the owner and name of the model, and a short part of the UUID just in case the operator is looking for the model-uuid.


howbazaar:
$$merge$$ optimistically

howbazaar:

howbazaar:
$$merge$$ fixed intermittent failure

PR #10769, opened 2019-10-21
CAAS unit init container for hook ctx and juju-run

hpidcock:
## CAAS unit init container for hook ctx and juju-run

This work moves the prepare logic outside of the command runner
for k8s into an init container that initializes the pod's /var/lib/juju for the
hook context and juju-run.

The init container shares /var/lib/juju and /usr/bin/juju-run via an
emptyDir volume with the workload containers. When the init container
starts it waits on an abstract domain socket for context to be delivered
via the caas-unit-init worker.

The caas-unit-init worker watches on the k8s api for the juju-init-pod to
enter the running state. It then uses the exec framework to copy over various
required files such as the charm and config then finally invokes the
command to complete the init.

## QA steps

- Bootstrap microk8s
- Deploy a workload with a test action
- Enable the developer-mode feature flag
- Test the action runs
- kubectl into the pod and test juju-run

## Documentation changes

N/A

## Bug reference

N/A

### DIFF
--- a/apiserver/facades/agent/caasoperator/operator.go
+++ b/apiserver/facades/agent/caasoperator/operator.go
@@ -10,8 +10,11 @@ import (
 	"github.com/juju/juju/apiserver/common"
 	"github.com/juju/juju/apiserver/facade"
 	"github.com/juju/juju/apiserver/params"
+	"github.com/juju/juju/caas"
 	k8sspecs "github.com/juju/juju/caas/kubernetes/provider/specs"
 	"github.com/juju/juju/core/status"
+	corewatcher "github.com/juju/juju/core/watcher"
+	"github.com/juju/juju/state/stateenvirons"
 	"github.com/juju/juju/state/watcher"
 )
 
@@ -19,6 +22,7 @@ type Facade struct {
 	auth      facade.Authorizer
 	resources facade.Resources
 	state     CAASOperatorState
+	broker    CAASBrokerInterface
 	*common.LifeGetter
 	*common.AgentEntityWatcher
 	*common.Remover
@@ -28,11 +32,19 @@ type Facade struct {
 	model Model
 }
 
+type CAASBrokerInterface interface {
+	WatchUnitStart(appName string) (corewatcher.StringsWatcher, error)
+}
+
 // NewStateFacade provides the signature required for facade registration.
 func NewStateFacade(ctx facade.Context) (*Facade, error) {
 	authorizer := ctx.Auth()
 	resources := ctx.Resources()
-	return NewFacade(resources, authorizer, stateShim{ctx.State()})
+	caasBroker, err := stateenvirons.GetNewCAASBrokerFunc(caas.New)(ctx.State())
+	if err != nil {
+		return nil, errors.Annotate(err, "getting caas client")
+	}
+	return NewFacade(resources, authorizer, stateShim{ctx.State()}, caasBroker)
 }
 
 // NewFacade returns a new CAASOperator facade.
@@ -40,6 +52,7 @@ func NewFacade(
 	resources facade.Resources,
 	authorizer facade.Authorizer,
 	st CAASOperatorState,
+	broker CAASBrokerInterface,
 ) (*Facade, error) {
 	if !authorizer.AuthApplicationAgent() {
 		return nil, common.ErrPerm
@@ -87,6 +100,7 @@ func NewFacade(
 		resources:          resources,
 		state:              st,
 		model:              model,
+		broker:             broker,
 	}, nil
 }
 
@@ -229,4 +243,41 @@ func (f *Facade) watchUnits(tagString string) (string, []string, error) {
 		return f.resources.Register(w), changes, nil
 	}
 	return "", nil, watcher.EnsureErr(w)
+}
+
+// WatchUnitStart starts a StringWatcher to watch for Unit start events
+// on the CAAS api.
+func (f *Facade) WatchUnitStart(args params.Entities) (params.StringsWatchResults, error) {
+	results := params.StringsWatchResults{
+		Results: make([]params.StringsWatchResult, len(args.Entities)),
+	}
+	for i, arg := range args.Entities {
+		id, changes, err := f.watchUnitStart(arg.Tag)
+		if err != nil {
+			results.Results[i].Error = common.ServerError(err)
+			continue
+		}
+		results.Results[i].StringsWatcherId = id
+		results.Results[i].Changes = changes
+	}
+	return results, nil
+}
+
+func (f *Facade) watchUnitStart(tagString string) (string, []string, error) {
+	tag, err := names.ParseApplicationTag(tagString)
+	if err != nil {
+		return "", nil, errors.Trace(err)
+	}
+	w, err := f.broker.WatchUnitStart(tag.Name)
+	if err != nil {
+		return "", nil, errors.Trace(err)
+	}
+	uw, err := newUnitIDWatcher(f.model, w)
+	if err != nil {
+		return "", nil, errors.Trace(err)
+	}
+	if changes, ok := <-uw.Changes(); ok {
+		return f.resources.Register(uw), changes, nil
+	}
+	return "", nil, watcher.EnsureErr(uw)
 }
--- a/apiserver/facades/agent/caasoperator/watcher.go
+++ b/apiserver/facades/agent/caasoperator/watcher.go
@@ -1,0 +1,105 @@
+// Copyright 2019 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package caasoperator
+
+import (
+	"github.com/juju/errors"
+	"github.com/juju/utils/set"
+	"gopkg.in/juju/worker.v1"
+	"gopkg.in/juju/worker.v1/catacomb"
+
+	corewatcher "github.com/juju/juju/core/watcher"
+	"github.com/juju/juju/state"
+)
+
+type unitIDWatcher struct {
+	catacomb catacomb.Catacomb
+	out      chan []string
+	src      corewatcher.StringsWatcher
+	model    watcherModelInterface
+}
+
+type watcherModelInterface interface {
+	Containers(providerIds ...string) ([]state.CloudContainer, error)
+}
+
+// newUnitIDWatcher watches a StringsWatcher and converts ProviderIDs into UnitIDs and re-emits them.
+func newUnitIDWatcher(model watcherModelInterface, src corewatcher.StringsWatcher) (*unitIDWatcher, error) {
+	w := &unitIDWatcher{
+		out:   make(chan []string),
+		src:   src,
+		model: model,
+	}
+	err := catacomb.Invoke(catacomb.Plan{
+		Site: &w.catacomb,
+		Work: w.loop,
+		Init: []worker.Worker{src},
+	})
+	return w, err
+}
+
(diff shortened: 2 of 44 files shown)


apiserver/facades/agent/caasoperator/operator.go
```diff
@@ -230,3 +244,38 @@ func (f *Facade) watchUnits(tagString string) (string, []string, error) {
 	}
 	return "", nil, watcher.EnsureErr(w)
 }
+
+func (f *Facade) WatchUnitStart(args params.Entities) (params.StringsWatchResults, error) {
```
wallyworld: add comment

apiserver/facades/agent/caasoperator/watcher.go
```diff
@@ -0,0 +1,108 @@
...
+	out      chan []string
+	src      corewatcher.StringsWatcher
+	model    watcherModelInterface
+}
+
+type watcherModelInterface interface {
+	Containers(providerIds ...string) ([]state.CloudContainer, error)
+}
+
+func newUnitIDWatcher(model watcherModelInterface, src corewatcher.StringsWatcher) (*unitIDWatcher, error) {
```
wallyworld: even though not exported, warrants a comment to explain what this watcher does, what it emits etc

apiserver/facades/agent/caasoperator/watcher.go
```diff
@@ -0,0 +1,108 @@
...
+			for _, containerInfo := range containers {
+				providerIDToUnitName[containerInfo.ProviderId()] = containerInfo.Unit()
+			}
+			var result []string
+			for _, providerID := range providerIDs {
+				if unitName, ok := providerIDToUnitName[providerID]; ok {
+					result = append(result, unitName)
+				}
+			}
+			w.out <- result
```
wallyworld: as discussed live, refactor to remove the naked send

apiserver/facades/agent/caasoperator/watcher.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+	}
+}
+
+func unique(in []string) []string {
```
wallyworld: We have a set.NewStrings() in juju/collections/set that can replace this to dedup the ids

apiserver/facades/agent/caasoperator/watcher_test.go
```diff
@@ -0,0 +1,51 @@
...
+	m.containers = []state.CloudContainer{
+		&mockCloudContainer{unit: "A", providerID: "a"},
+		&mockCloudContainer{unit: "C", providerID: "c"},
+	}
+	wc := make(chan []string, 4)
+	wc <- []string{"a"}
+	wc <- []string{"b"}
+	wc <- []string{"c"}
+	wc <- []string{"d"}
+	srcWatcher := watchertest.NewMockStringsWatcher(wc)
```
wallyworld: wrap this in
testing.NewStringsWatcherC

then you have various assert methods for no change, expected changes etc

apiserver/facades/controller/caasunitprovisioner/provisioner.go
```diff
@@ -269,6 +271,7 @@ func (f *Facade) provisioningInfo(model Model, tagString string) (*params.Kubern
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
+	operatorImagePath := podcfg.GetJujuOCIImagePath(controllerCfg, version.Current)
```
wallyworld: This should be consistent with caasoperator provisioner?

```
	vers, ok := modelConfig.AgentVersion()
	if !ok {
		return params.OperatorProvisioningInfo{}, errors.NewNotValid(nil,
			fmt.Sprintf("agent version is missing in model config %q", modelConfig.Name()),
		)
	}
	vers.Build = 0
	imagePath := podcfg.GetJujuOCIImagePath(cfg, vers)
```

caas/kubernetes/provider/k8s.go
```diff
@@ -1740,6 +1804,82 @@ func (k *kubernetesClient) WatchUnits(appName string) (watcher.NotifyWatcher, er
 	return k.newWatcher(w, appName, k.clock)
 }
 
+func (k *kubernetesClient) WatchUnitStart(appName string) (watcher.StringsWatcher, error) {
```
wallyworld: add comment

caas/kubernetes/provider/k8s.go
```diff
@@ -2245,37 +2392,35 @@ func defaultSecurityContext() *core.SecurityContext {
...
-		}
-		if spec.ReadinessProbe != nil {
-			podContainers[i].ReadinessProbe = spec.ReadinessProbe
-		}
-		if spec.SecurityContext != nil {
-			podContainers[i].SecurityContext = spec.SecurityContext
-		} else {
-			podContainers[i].SecurityContext = defaultSecurityContext()
+		pc.SecurityContext = defaultSecurityContext()
+		if c.ProviderContainer != nil {
```
wallyworld: using continue here avoids the extra indentation

caas/kubernetes/provider/k8swatcher.go
```diff
@@ -101,3 +101,92 @@ func (w *kubernetesWatcher) Kill() {
...
+			logger.Tracef("received k8s event: %+v", evt.Type)
+			if emittedEvent, ok := w.filterFunc(evt); ok {
+				pendingEvents = append(pendingEvents, emittedEvent)
+				if delayCh == nil {
+					delayCh = w.clock.After(sendDelay)
+				}
+			}
+		case <-delayCh:
+			if pendingEvents != nil {
+				w.out <- pendingEvents
```
wallyworld: this also need refactoring.
make it look like the notify one

cmd/jujud/agent/caasunitinit.go
```diff
@@ -0,0 +1,201 @@
...
+	f.BoolVar(&c.Wait, "wait", false, "wait for args init via socket")
+	f.BoolVar(&c.Send, "send", false, "send args for init via socket")
+	f.StringVar(&c.Args.Unit, "unit", "", "unit name")
+	f.StringVar(&c.Args.OperatorFile, "operator-file", "", "operator client info file")
+	f.StringVar(&c.Args.OperatorCACertFile, "operator-ca-cert-file", "", "ca cert for operator")
+	f.StringVar(&c.Args.CharmDir, "charm-dir", "", "directory containing the charm")
+}
+
+func (c *CAASUnitInitCommand) Init(args []string) error {
+	if c.Wait && c.Send {
```
wallyworld: what happens if neither are set.
we need at least one right?

cmd/jujud/agent/caasunitinit.go
```diff
@@ -0,0 +1,201 @@
...
+		err = json.NewDecoder(conn).Decode(&c.Args)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		w := loggo.NewSimpleWriter(conn, loggo.DefaultFormatter)
+		err = loggo.RegisterWriter("socket", w)
+		if err != nil {
+			return errors.Trace(err)
+		}
+	} else if c.Send {
```
wallyworld: can we move this block first since it returns
then the wait stuff can go beneath it with the rest of the logic below

worker/caasoperator/caasoperator.go
```diff
@@ -210,6 +223,19 @@ func NewWorker(config Config) (worker.Worker, error) {
 		return nil, errors.Annotatef(err, "cannot create deployer")
 	}
 
+	podInit, err := newPodInitWorker(podInitConfig{
```
wallyworld: I do think this needs to be a separate worker in the caasoperator manifold
And the NewWorker func is part of the manifold config so it can be tested etc

ie make the pod init worker and associated stuff look like the caasoperator worker
the can both run independently and as sibling workers of the top level runner
is there a core need to make this worker a child of the caas operator worker?

hpidcock: @wallyworld can this be a part of a follow-up PR?

apiserver/facades/controller/caasunitprovisioner/provisioner.go
hpidcock: @wallyworld for some reason I remember AgentVersion always returning ok = false here. I'll have a look again.

worker/caasoperator/caasoperator.go
hpidcock: never mind, I'm adding this

worker/caasunitinit/initializer.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// WriteFile is used to write files to the local state.
+	WriteFile func(string, []byte, os.FileMode) error
+
+	// TempDir is used for creating a temporary directory.
+	TempDir func(string, string) (string, error)
+}
+
+// NewUnitInitializer returns a new worker start function to initilize the Unit.
+func NewUnitInitializer(config UnitInitializerConfig, unitTag names.UnitTag) func() (worker.Worker, error) {
```
wallyworld: This doesn't need to be a worker. Just have it as a method which takes a cancel channel. That channel is set up in the other worker and closed as needed by the other worker.

apiserver/facades/agent/caasoperator/watcher.go
```diff
@@ -0,0 +1,109 @@
...
+
+	for {
+		select {
+		case <-w.catacomb.Dying():
+			return w.catacomb.ErrDying()
+		case providerIDs, ok := <-w.src.Changes():
+			if !ok {
+				return errors.Errorf("event watcher closed")
+			}
+			if len(providerIDs) == 0 {
```
wallyworld: we don't need this check, just the sendInitial one below

apiserver/facades/agent/caasoperator/watcher.go
```diff
@@ -0,0 +1,109 @@
...
+				if sendInitial {
+					out = w.out
+					sendInitial = false
```
wallyworld: don't set to false here (and a second time below) - set to false in the out <- results bit

caas/kubernetes/provider/k8s.go
```diff
@@ -1740,6 +1804,82 @@ func (k *kubernetesClient) WatchUnits(appName string) (watcher.NotifyWatcher, er
...
+
+	podsList, err := pods.List(v1.ListOptions{
+		LabelSelector:        applicationSelector(appName),
+		IncludeUninitialized: true,
+	})
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	providerID := func(pod *core.Pod) string {
```
wallyworld: This functionality to get the provider id is used here but also in Units(). Can we have a helper function used in both places or else the logic is in danger of diverging.

hpidcock:
!!build!!

cmd/jujud/agent/caasunitinit.go
wallyworld: if resolving without doing, could we add a comment to explain why

cmd/jujud/agent/caasunitinit.go
wallyworld: an answer if resolving?

hpidcock: Ahh sorry I added a comment inside the Run method.
// The waiting process waits for arguments to be sent by the
// sending process. If neither wait or send is specified,
// continue with the arguments already specified.

hpidcock:
$$merge$$

PR #10774, opened 2019-10-22
update-cloud ask-or-tell and exclusivity.

anastasiamac:
## Description of change

Several changes need to take place with 'juju update-cloud' command.

* By default, the command operates on both client and controller copies.

* Ask-or-tell component of the change is applicable when a user did not specify a controller nor explicitly asked for --client-only operation but the presence of a current controller was detected. In that instance, users will be prompted to confirm if the current controller is to be used. For automated environments, --no-prompt option will automatically use a current controller when it's detected.

* Occasionally, users may not want to add-k8s to both the client and a controller. In these instances, '--client-only' or '--controller-only' options can be used to filter one or the other.

## Documentation changes

all of the above

### DIFF
--- a/cmd/juju/cloud/export_test.go
+++ b/cmd/juju/cloud/export_test.go
@@ -80,13 +80,12 @@ func NewUpdatePublicCloudsCommandForTest(publicCloudURL string) *updatePublicClo
 func NewUpdateCloudCommandForTest(
 	cloudMetadataStore CloudMetadataStore,
 	store jujuclient.ClientStore,
-	cloudAPI func(string) (UpdateCloudAPI, error),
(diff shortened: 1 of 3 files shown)


anastasiamac:
[update-cloud  ask-or-tell and exlusivity](https://trello.com/c/ayaMmi9H/949-update-cloud-ask-or-tell-and-exlusivity)

anastasiamac:
unrelated failure in github.com/juju/juju/worker/caasoperator.TestPackage

anastasiamac:
win tests failed because of deps? am getting a few of "cannot find package "github.com/juju/cmd" in any of:...."

anastasiamac:
$$merge$$

PR #10785, opened 2019-10-23
Add action logs to model export

wallyworld:
## Description of change

Add action logs to model export.
Also, as a driveby, limit the log count just in case an action wants to misbehave.

## QA steps

run a charm with an action which logs messages
juju dump-model and look at the yaml to see the logs included

### DIFF
--- a/state/action.go
+++ b/state/action.go
@@ -112,8 +112,18 @@ type actionDoc struct {
 
 // ActionMessage represents a progress message logged by an action.
 type ActionMessage struct {
-	Message   string    `bson:"message"`
-	Timestamp time.Time `bson:"timestamp"`
+	MessageValue   string    `bson:"message"`
+	TimestampValue time.Time `bson:"timestamp"`
+}
+
+// Timestamp returns the message timestamp.
+func (m ActionMessage) Timestamp() time.Time {
+	return m.TimestampValue
+}
+
+// Message returns the message string.
+func (m ActionMessage) Message() string {
+	return m.MessageValue
 }
 
 // action represents an instruction to do some "action" and is expected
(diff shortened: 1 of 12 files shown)


wallyworld:
[action progress logs in model export/migrations](https://trello.com/c/OFppnzVi/909-action-progress-logs-in-model-export-migrations)

state/action.go
```diff
@@ -112,8 +112,18 @@ type actionDoc struct {
 
 // ActionMessage represents a progress message logged by an action.
 type ActionMessage struct {
-	Message   string    `bson:"message"`
-	Timestamp time.Time `bson:"timestamp"`
+	Message_   string    `bson:"message"`
```
hpidcock: This just seems a bit clumsy. Possibly MessageValue and TimestampValue

state/migration_export.go
```diff
@@ -1422,7 +1422,13 @@ func (e *exporter) actions() error {
 			Results:    results,
 			Message:    message,
 			Id:         action.Id(),
-		})
+		}
+		messages := action.Messages()
+		arg.Messages = make([]description.ActionMessage, len(messages))
```
hpidcock: possibly:
```
Messages: append([]description.ActionMessage(nil), action.Messages()...),
```

wallyworld: different types so need to copy

state/action.go
wallyworld: We use this elsewhere when we need to have stuff available for serialisation but it's really internal (like in the juju/description package). I think this is the first time in juju/juju so maybe best to avoid it here. I can change to Value

wallyworld:
$$merge$$

PR #10786, opened 2019-10-23
add-cloud exclusivity.

anastasiamac:
## Description of change

Occasionally, users may not want to add-k8s to both the client and a controller. In these instances, '--client-only' or '--controller-only' options can be used to filter one or the other.

This PR adds the support for this exclusivity to 'add-cloud'.

As a drive-by, this PR removes the need to flag.Parse() from featuretest package. The way it was implemented there is no longer supported in go 1.13 due to the change of precedence and order of package level and testing init. In addition, the flag that the original code was catering for - 'featuretest' - has never been used.

### DIFF
--- a/cmd/juju/cloud/add.go
+++ b/cmd/juju/cloud/add.go
@@ -69,20 +69,23 @@ positional argument:
     juju add-cloud mycloud -f ~/mycloud.yaml
 
 When <cloud definition file> is provided with <cloud name>,
-Juju stores that definition in the current controller (after
-validating the contents), or the specified controller if
---controller is used. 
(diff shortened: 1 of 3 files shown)


anastasiamac:
[add-cloud exclusivity](https://trello.com/c/rSNvG5kK/952-add-cloud-exclusivity)

anastasiamac:
$$merge$$

anastasiamac:
another lint timeout exceeded during 'check' ... 17:00:31 checking: golinters ...
17:01:49 level=error msg="Timeout exceeded: try increase it by passing --timeout option"

anastasiamac:
$$merge$$

PR #10788, opened 2019-10-23
Make "juju bind" return an error if no bindings are specified

achilleasa:
## Description of change

Having `juju bind X` just print out the current bindings is slightly confusing for a command that is supposed to only be mutating them. Besides, we have other commands that display bindings anyway. 

This PR makes `juju bind` fail if no bindings are specified.

## QA steps

This is a client-side change only. Just compile and run the following:

```console
$ juju bind $a-deployed-app
ERROR no set of bindings specified
```

## Documentation changes

@timClicks this is a small change in behavior from #10734.


jameinel:
'juju bind app' displaying bindings would match 'juju config app'.
We can have a separate command (juju show-bindings app), just mentioning
there is some precedent for a single command to do both reading and writing.

cmd/juju/application/bind.go
```diff
@@ -130,6 +127,10 @@ func (c *bindCommand) Run(ctx *cmd.Context) error {
 		return err
 	}
 
+	if len(c.Bindings) == 0 {
+		return errors.New("no set of bindings specified")
```
hmlanigan: Suggest "no bindings specified"

achilleasa:
$$merge$$

PR #10792, opened 2019-10-24
Fix caasoperator id watcher test race condition

hpidcock:
## Description of change

Fix caasoperator id watcher test race condition.

## QA steps

Run the tests.

## Documentation changes

N/A

## Bug reference

N/A


hpidcock:
$$merge$$

PR #10794, opened 2019-10-24
Add appdata-source and appdata-sink test charms

babbageclunk:
# Description of change

These are cloned from the dummy-source and -sink charms but instead use the config token set to populate application relation data so we can see it flow across a relation.

## QA steps

* Deploy appdata-source and appdata-sink and relate them.
```
juju deploy ./acceptancetests/repository/charms/appdata-source -n 3
juju deploy ./acceptancetests/repository/charms/appdata-sink -n 3
juju relate appdata-source appdata-sink
```
* Set the token on the source application.
```
juju config token="some value"
```
* The leader source unit should take the token and store it as application data in the relation (see the log). 
* The sink units should all receive the application token and set it as status, as well as storing it in `/var/run/appdata-sink/token`.

## Documentation changes

None

## Bug reference

None


acceptancetests/repository/charms/appdata-sink/actions.yaml
```diff
@@ -0,0 +1,6 @@
+echo:
+  description: Echo takes a string value and echo's it out
+  params:
+    value:
+      type: string
+      description: The value that should be echo'd
```
jameinel: trailing newline?

acceptancetests/repository/charms/appdata-sink/actions/echo
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -eux
+
```
jameinel: Is it worth copying in the actions?

acceptancetests/repository/charms/appdata-sink/config.yaml
```diff
@@ -0,0 +1,5 @@
+options:
+  token:
+    type: string
+    default: ''
```
jameinel: I thought only the source had a token configured, and then that was sent via relation data to the sink

acceptancetests/repository/charms/appdata-sink/hooks/source-relation-changed
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -eux
+
+remote_app=${JUJU_REMOTE_APP:-}
+remote_unit=${JUJU_REMOTE_UNIT:-}
+if [[ -n $remote_app ]]; then
```
jameinel: So if I did it correctly JUJU_REMOTE_APP is *always* set, but JUJU_REMOTE_UNIT is set sometimes.
So I think we want
if [[ -n $remote_unit ]]; then

as the check, so that if there is a remote unit, then we treat it as a unit context, not "if there isn't an app", we treat it as a unit context. (Obviously if we change the if, we need to swap the content bodies)

acceptancetests/repository/charms/appdata-source/hooks/leader-elected
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -eux
+
+is_leader="$(is-leader)"
+if [ "$is_leader" = "False" ]; then
```
jameinel: I'm curious if we'd want something like per-unit tokens so that we can see that we take only the app token, but maybe not.

Is it worth something here that does some sort of "relation-get/set" as though it was the leader and sees that it gets an error (eg, trap the error, and return 0 from the hook, but make sure that we *get* the error)

jameinel:
I also just realized something else, "dummy-sink" just calls "relation-get". So with luck, you should be able to relate appdata-source and dummy-sink, and even though appdata-source is setting it as the --app, 'relation-get' in the hook that fires because the app-data changes, should still cause dummy-sink to see the right content. Care to try it?

acceptancetests/repository/charms/appdata-sink/actions/echo
babbageclunk: I'll ditch them from appdata-sink and update the -source ones to have the unit try to set the token unconditionally.

acceptancetests/repository/charms/appdata-sink/config.yaml
babbageclunk: I'll remove this.

acceptancetests/repository/charms/appdata-source/hooks/leader-elected
babbageclunk: I like your idea of an action that tries to set the token - I don't think we need to be so careful about returning 0 like we would in a hook.

acceptancetests/repository/charms/appdata-sink/hooks/source-relation-changed
babbageclunk: Ah right - I'll update this.

babbageclunk:
> I also just realized something else, "dummy-sink" just calls "relation-get". So with luck, you should be able to relate appdata-source and dummy-sink, and even though appdata-source is setting it as the --app, 'relation-get' in the hook that fires because the app-data changes, should still cause dummy-sink to see the right content. Care to try it?

Ooh, cunning - because dummy-sink calls relation-get without a source it'll use the env vars, so it gets the app-level stuff?
Yeah, I'll try that too.

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #10798, opened 2019-10-24
Add test for upgrade step and handle nested directories.

howbazaar:
There was an idempotency issue with the upgrade step to set the ownership and permissions of the log file.

In 2.7 we now have a nested "models" directory which holds the logs for the workers run on the controller on behalf of the models. The walking of the files in the /var/log/juju directory wasn't taking this into account when creating the path to use. Luckily this is actually one of the parameters to the walk function.


howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #10800, opened 2019-10-25
Check for incorrect endpoint in remove offer url

wallyworld:
## Description of change

Users got confused when running juju-remove offer and included the endpoint in the name. This caused juju to try and remove an offer called "offer:ep" instead of "offer".
Add a check to remove-offer to ensure no endpoint is included.

## QA steps

remove-remove offer mariadb:db mysql
ERROR These offers contain endpoints. Only specify the offer name itself:
 -admin/default.mariadb:db


wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10801, opened 2019-10-25
Fixes lp#1848104: updating existing credential with invalid value invalidates other credentials

anastasiamac:
## Description of change

On the controller facades, creating cloud call context in the facade construction makes no sense since call context is model specific. 

This PR changes the creation of the context to be in the actual facade call methods. All context creation has been moved to the calls that use context rather than forcing it to happen too early and for every call. 

## QA steps

steps in bug report do not result in the controller model cloud credential to be invalidated. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1848104

### DIFF
--- a/apiserver/common/networkingcommon/networkconfigapi.go
+++ b/apiserver/common/networkingcommon/networkconfigapi.go
@@ -212,7 +212,7 @@ func (api *NetworkConfigAPI) getOneMachineProviderNetworkConfig(m *state.Machine
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	interfaceInfos, err := netEnviron.NetworkInterfaces(state.CallContext(model.State()), instId)
+	interfaceInfos, err := netEnviron.NetworkInterfaces(state.CallContext(api.st), instId)
 	if errors.IsNotSupported(err) {
 		// It's possible to have a networking environ, but not support
 		// NetworkInterfaces(). In leiu of adding SupportsNetworkInterfaces():
--- a/apiserver/common/networkingcommon/networkconfigapi.go
+++ b/apiserver/common/networkingcommon/networkconfigapi.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/juju/juju/apiserver/common"
 	"github.com/juju/juju/apiserver/params"
-	"github.com/juju/juju/environs/context"
 	"github.com/juju/juju/state"
 	"github.com/juju/juju/state/stateenvirons"
 )
@@ -25,15 +24,12 @@ var logger = loggo.GetLogger("juju.apiserver.common.networkingcommon")
 type NetworkConfigAPI struct {
 	st           *state.State
 	getCanModify common.GetAuthFunc
-
-	callContext context.ProviderCallContext
 }
 
-func NewNetworkConfigAPI(st *state.State, callCtx context.ProviderCallContext, getCanModify common.GetAuthFunc) *NetworkConfigAPI {
+func NewNetworkConfigAPI(st *state.State, getCanModify common.GetAuthFunc) *NetworkConfigAPI {
 	return &NetworkConfigAPI{
 		st:           st,
 		getCanModify: getCanModify,
-		callContext:  callCtx,
 	}
 }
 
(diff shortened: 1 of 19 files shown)


anastasiamac:
[api call context need to be relevant to model](https://trello.com/c/9yZxLIP1/960-api-call-context-need-to-be-relevant-to-model)

apiserver/facades/client/applicationoffers/state.go
```diff
@@ -106,6 +110,14 @@ func (s *stateShim) Model() (Model, error) {
 	return &modelShim{m}, err
 }
 
+func (s *stateShim) GetModelCallContext() (context.ProviderCallContext, error) {
+	m, err := s.st.Model()
```
wallyworld: I don't get this. m.State() is the same as s.st so getting the model is redundant.

apiserver/facades/client/cloud/backend.go
```diff
@@ -97,32 +98,18 @@ func NewModelPoolBackend(st *state.StatePool) ModelPoolBackend {
...
+	modelState, err := s.StatePool.Get(modelUUID)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer modelState.Release()
+	m, err := modelState.Model()
+	if err != nil {
+		return nil, nil, err
+	}
+	return credentialcommon.NewPersistentBackend(m.State()), state.CallContext(m.State()), err
```
wallyworld: why not just use modelState from above?
It will be the same as m.State()?

apiserver/facades/client/applicationoffers/state.go
anastasiamac: True, changed to s.st

apiserver/facades/client/cloud/backend.go
anastasiamac: modelState is actually a PooledState, so I have changed to use modelState.State rather than going through model for simplicity.

apiserver/common/networkingcommon/networkconfigapi.go
```diff
@@ -216,8 +212,7 @@ func (api *NetworkConfigAPI) getOneMachineProviderNetworkConfig(m *state.Machine
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-
-	interfaceInfos, err := netEnviron.NetworkInterfaces(api.callContext, instId)
+	interfaceInfos, err := netEnviron.NetworkInterfaces(state.CallContext(model.State()), instId)
```
wallyworld: api.st
instead of model.State()

apiserver/facades/client/cloud/instance_information.go
```diff
@@ -49,6 +49,14 @@ func instanceTypes(api *CloudAPI,
 	if err != nil {
 		return params.InstanceTypesResults{}, errors.Trace(err)
 	}
+	modelCfg, err := api.ctlrBackend.ModelConfig()
```
wallyworld: To save an api call, we can use m.UUID()
We already get the model above

apiserver/facades/controller/migrationtarget/migrationtarget.go
```diff
@@ -251,7 +247,7 @@ func (api *API) AdoptResources(args params.AdoptResourcesArgs) error {
 		return errors.Trace(err)
 	}
 
-	return errors.Trace(ra.AdoptResources(api.callContext, st.ControllerUUID(), args.SourceControllerVersion))
+	return errors.Trace(ra.AdoptResources(state.CallContext(m.State()), st.ControllerUUID(), args.SourceControllerVersion))
```
wallyworld: st

anastasiamac: No, st here is a PooledState. I needed State.

anastasiamac:
$$merge$$

anastasiamac:
fatal: The remote end hung up unexpectedly
17:13:38 fatal: early EOF
17:13:38 fatal: index-pack failed
17:13:38 : command failed: [git clone --recursive -v --progress https://code.googlesource.com/google-api-go-client /home/ubuntu/go/pkg/dep/sources/https---code.googlesource.com-google--api--go--client]: exit status 128
17:13:38 Makefile:74: recipe for target 'dep' failed
17:13:38 make: *** [dep] Error 1

anastasiamac:
unrelated failure in github.com/juju/juju/worker/raft/raftflag.TestPackage

anastasiamac:
$$merge$$

PR #10803, opened 2019-10-25
Clean up branches on model destroy

SimonRichardson:
## Description of change

The following cleans up any branches associated with a model, when
the model is destroyed. Ensuring two things:

 1. We remove any orphaned branch models from the db
 2. We ensure that there won't be any conflicts from leaving any
    branches around.

The code attempts to use the latest Apply change strategy, which
applies a visitor pattern to a change operation to clean up the
branches.

It would be nice for these strategies to be moved out from state
and fully mocked, but probably not today!

## QA steps

TBA

### DIFF
--- a/tests/suites/branches/branches.sh
+++ b/tests/suites/branches/branches.sh
@@ -1,0 +1,44 @@
+run_branch() {
+    # Echo out to ensure nice output to the test suite.
+    echo
+
+    # The following ensures that a bootstrap juju exists
+    file="${TEST_DIR}/test-branches.txt"
+    ensure "branches" "${file}"
+
+    juju branch | grep -q 'Active branch is "master"'
+
+    juju deploy redis
+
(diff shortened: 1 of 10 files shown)


tests/suites/branches/branches.sh
```diff
@@ -0,0 +1,44 @@
...
+
+    # The following ensures that a bootstrap juju exists
+    file="${TEST_DIR}/test-branches.txt"
+    ensure "branches" "${file}"
+
+    juju branch | grep -q 'Active branch is "master"'
+
+    juju deploy redis
+
+    wait_for "redis" ".applications | select(.redis | .units | .[\"redis/0\"] | .[\"juju-status\"] | .current == \"idle\") | keys[0]"
```
SimonRichardson: We need to ensure that redis is deployed, before continuing on.

SimonRichardson:
https://trello.com/c/1Rgc4XMz/979-remove-generation-docs-when-a-model-is-destroyed

SimonRichardson:
$$merge$$

PR #10807, opened 2019-10-28
CAAS unit actions wait for container start.

hpidcock:
## CAAS unit actions wait for container start.

- CAAS broker can now watch on specific container start events.
- CAAS unit provisioner annotates pods with the unit name, this is used
to filter container start events until pod name/uid is associated with a
unit. Fixes a pod init race condition for pods created by k8s.
- Uniter resolver triggered by CAAS container start events.
- Action resolver defers running actions until container starts

## QA steps

- Bootstrap k8s
- `juju deploy cs:~juju/redis-k8s-1 && juju exec -u redis-k8s/0 hostname`
- Hostname should print out.

## Documentation changes

N/A

## Bug reference

N/A


api/uniter/uniter.go
```diff
@@ -252,6 +252,30 @@ func (st *State) ActionBegin(tag names.ActionTag) error {
 	return nil
 }
 
+// ActionReset marks a running action as pending.
+func (st *State) ActionReset(tag names.ActionTag) error {
```
wallyworld: ResetAction to match ResetActions?

apiserver/facades/agent/caasoperator/operator.go
```diff
@@ -245,14 +245,14 @@ func (f *Facade) watchUnits(tagString string) (string, []string, error) {
 	return "", nil, watcher.EnsureErr(w)
 }
 
-// WatchUnitStart starts a StringWatcher to watch for Unit start events
-// on the CAAS api.
-func (f *Facade) WatchUnitStart(args params.Entities) (params.StringsWatchResults, error) {
```
wallyworld: Might be worth retaining this method as deprecated and delete in 2.7.1
(since existing caas operators from 2.7-beta1 could call it).
Although it's only the unit init worker and any pods would almost certainly be running when the upgrade is done to this version. So probs is ok

apiserver/params/params.go
```diff
@@ -1233,3 +1233,15 @@ type CharmProfilingInfoResult struct {
 	CurrentProfiles []string            `json:"current-profiles"`
 	Error           *Error              `json:"error"`
 }
+
+// WatchContainerStartArg contains arguments for watching for container start
+// events on a CAAS application.
+type WatchContainerStartArg struct {
+	Entity    Entity `json:"entity"`
+	Container string `json:"container"`
```
wallyworld: omitempty

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -1170,6 +1170,22 @@ func (u *UniterAPI) BeginActions(args params.Entities) (params.ErrorResults, err
 	return common.BeginActions(args, actionFn), nil
 }
 
+// ResetActions marks the actions represented by the passed in Tags as running.
```
wallyworld: as pending

apiserver/common/action.go
```diff
@@ -103,6 +103,29 @@ func BeginActions(args params.Entities, actionFn func(string) (state.Action, err
 	return results
 }
 
+// ResetActions calls reset on every action passed in through args.
+// It's a helper function currently used by the uniter
+// It needs an actionFn that can fetch an action from state using it's id, that's usually created by AuthAndActionFromTagFn
+func ResetActions(args params.Entities, actionFn func(string) (state.Action, error)) params.ErrorResults {
```
wallyworld: Should we call this "RequeueActions"?
It perhaps conveys more meaning as "Reset" is maybe too generic?

state/action.go
```diff
@@ -220,6 +220,29 @@ func (a *action) Begin() (Action, error) {
 	return m.Action(a.Id())
 }
 
+// Reset a running action to pending.
+// It asserts that the action is currently running.
+func (a *action) Reset() (Action, error) {
```
wallyworld: Requeue() ?

worker/caasoperator/action.go
```diff
@@ -10,70 +10,84 @@ import (
...
+	unitPaths uniter.Paths,
+	params runner.ExecParams) (*utilexec.ExecResponse, error) {
+	if err := providerIDGetter.Refresh(); err != nil {
+		return nil, errors.Trace(err)
+	}
+	podNameOrID := providerIDGetter.ProviderID()
+	unitName := providerIDGetter.Name()
+	logger.Debugf("exec on pod %q for unit %q, cmd %v", podNameOrID, unitName, params.Commands)
+	if podNameOrID == "" {
+		//return nil, errors.NotFoundf("pod for %q", unitName)
```
wallyworld: delete me?

worker/caasoperator/caasoperator.go
```diff
@@ -413,13 +417,22 @@ func (op *caasOperator) loop() (err error) {
...
+	}
+	if err := op.catacomb.Add(containerStartWatcher); err != nil {
+		return errors.Trace(err)
+	}
+
 	if err := op.setStatus(status.Active, ""); err != nil {
 		return errors.Trace(err)
 	}
 
 	// Keep a record of the alive units an a channel used to notify
```
wallyworld: update comment

worker/caasoperator/caasoperator.go
```diff
@@ -452,45 +465,63 @@ func (op *caasOperator) loop() (err error) {
 					return errors.Trace(err)
 				}
 				// Notify all uniters of the change so they run the upgrade-charm hook.
-				for unitId, changedChan := range aliveUnits {
-					logger.Debugf("trigger upgrade charm for caas unit %v", unitId)
+				for unitID, changedChan := range unitAppChannels {
```
wallyworld: prefer original name of aliveUnits as it has clearer meaning

worker/caasoperator/caasoperator.go
```diff
@@ -452,45 +465,63 @@ func (op *caasOperator) loop() (err error) {
...
+			if !ok {
+				return errors.New("container start watcher closed channel")
+			}
+			for _, unitID := range units {
+				if status, ok := unitStatusChannels[unitID]; ok {
+					logger.Debugf("trigger status change for caas unit %v", unitID)
+					select {
+					case <-op.catacomb.Dying():
+						return op.catacomb.ErrDying()
+					case status <- struct{}{}:
```
wallyworld: statusChan?
to be consistent

worker/uniter/runner/context/context.go
```diff
@@ -777,6 +777,17 @@ func (ctx *HookContext) Prepare() error {
 	return nil
 }
 
+// Reset implements the Context interface.
+func (ctx *HookContext) Reset() error {
```
wallyworld: Context has generic methods like Prepare() and Flush().
Plus task specific ones like ResetExecutionSetUnitStatus()

I think here RequeueAction() is better. We already have an ActionData() method. ie action related methods have "Action" in the name

worker/uniter/operation/runaction.go
```diff
@@ -73,6 +74,18 @@ func (ra *runAction) Execute(state State) (*State, error) {
...
+	cause := errors.Cause(err)
+	switch {
+	case cause == context.ErrResetAndSleep:
+		err = ra.runner.Context().Reset()
+		if err != nil {
+			return nil, errors.Annotatef(err, "resetting action %q", ra.name)
+		}
+		return stateChange{
+			Kind: Continue,
+			Step: Pending,
```
wallyworld: It seems with this approach, any event that wake sup the resolver will cause the action to be reset, rather than a pod status change. That's probably ok to start with.

apiserver/facades/agent/caasoperator/operator.go
wallyworld: Discussed live, this was only introduced after beta1 so ok

worker/caasoperator/caasoperator.go
```diff
@@ -452,45 +467,63 @@ func (op *caasOperator) loop() (err error) {
...
+			if !ok {
+				return errors.New("container start watcher closed channel")
+			}
+			for _, unitID := range units {
+				if runningChan, ok := unitRunningChannels[unitID]; ok {
+					logger.Debugf("trigger running status for caas unit %v", unitID)
+					select {
+					case <-op.catacomb.Dying():
+						return op.catacomb.ErrDying()
+					case runningChan <- struct{}{}:
```
wallyworld: We should consider the case where the unit goes to dead (below). We should abort any attempt to send on the runningChan. So we should create a unitDead chan or something and close it below and select on it here

hpidcock: We have the same problem with the aliveUnits chans, I don't think it's a terrible problem since the uniter worker restarts and uses the same channels.

Proper way to deal with this is have the resolver create watchers for both application changes and running status changes, that way the watcher is tied to the lifecycle of the resolver/uniter.
Then we can use the watcher's dying channel to safely notify.

hpidcock:
!!build!!

hpidcock:
!!build!!

worker/caasoperator/caasoperator.go
```diff
@@ -413,13 +420,25 @@ func (op *caasOperator) loop() (err error) {
...
 	}
 
-	// Keep a record of the alive units an a channel used to notify
+	// Keep a record of the alive units and a channel used to notify
 	// their uniter workers when the charm version has changed.
 	aliveUnits := make(map[string]chan struct{})
+	// Channels used to notify uniter worker that the workload container
+	// is running.
+	unitRunningChannels := make(map[string]chan struct{})
+	unitRunning := make(map[string]struct{})
```
wallyworld: this can be a set

hpidcock:
$$merge$$

hpidcock:
> Can we also ensure that's it's been tested with a model upgraded from 2.6.9 to ensure all the bits line up for preexisting pods/units etc

Just checked, appears to work with 2.6.10

PR #10808, opened 2019-10-28
k8s CLI usability fixes

wallyworld:
## Description of change

There's several small usability fixes for the k8s CLI:
- remove-unit has a better error message when a unit name is supplied
- use "k8s" consistently instead of "kubernetes" or "Kubernetes" in CLI text
- add-k8s reports existing or new storage class correctly
- CLI help text improved
- add-k8s --cloud and --region options improved (dumb cases no longer allowed)

The last change means that:
-- cloud only accepts a cloud name
-- you can't specify just a region with --region unless for gke or aks

There's a slight incompatibility with 2.6 but the logical way to use the options is the what's supported; only the pathological corner cases have been removed. It's now less confusing. Plus using microk8s or GKE/AWS or CDK on a supported cloud it's not an issue anyway; it's only for cases where the undercloud cannot be detected.

## QA steps

juju remove-unit <unitname> and check the error message
add-k8s on a GKE cluster
add-k8s on microk8s
add-k8s on CDK without node labels with and without an existing storage class
(testing the combinations of --cloud and --region)

## Bug reference

https://bugs.launchpad.net/juju/+bug/1847149

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"strings"
 	"time"
 
 	jujuclock "github.com/juju/clock"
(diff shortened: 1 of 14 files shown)


wallyworld:
[better remove-unit feedback when a unit name is supplied](https://trello.com/c/tjX0WsWr/962-better-remove-unit-feedback-when-a-unit-name-is-supplied)

wallyworld:
!!build!!

cmd/juju/caas/add.go
```diff
@@ -77,8 +78,19 @@ It's also possible to select a context by name using --context-name.
...
 detected to enable storage to be correctly configured. If the cloud/region cannot
-be detected automatically, use --region <cloudType|cloudName>/<someregion> to specify the host
-cloud type and region.
+be detected automatically, use either
+  --cloud <cloudType|cloudName> to specify the host cloud
+or
+  --region <cloudType|cloudName>/<someregion> to specify the host
+  cloud type and region.
+
+Region is strictly necessary only when adding a k8s cluster to a JAAS controller.
```
anastasiamac: Nice explanation!

cmd/juju/caas/add_test.go
```diff
@@ -508,6 +513,24 @@ func (s *addCAASSuite) TestCloudAndRegionFlag(c *gc.C) {
 			cloud:          "brokenteststack",
 			expectedErrStr: `validating cloud region "azure": cloud region "azure" not valid`,
 		},
+		{
+			title:          "specify cloud with gke",
+			cloud:          "aws",
+			gke:            true,
+			expectedErrStr: `do not specify --cloud when adding a GKE or AKS cluster`,
+		},
```
anastasiamac: Brackets could be on one line :)
```
}, {
```

wallyworld:
$$merge$$

PR #10812, opened 2019-10-28
Deploy integration tests

SimonRichardson:
## Description of change

The following changes rework the deploy test cases, so that we
can test juju deploy in a more consistent manor.

The following deploys a series of lxd profiles in both principal and
subordinate charms.

There is a massive issue around destroying storage on a lxd machine.
Some time should be spent trying to workout why that fails. It's not
like it's a rare occurrence, every other time at least when running

    ./main.sh -v deploy

## QA steps

```sh
cd tests
./main.sh -v deploy
```


SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #10816, opened 2019-10-29
state/watcher: Add a timeout to hubwatcher sending

babbageclunk:
## Description of change

If a state-level watcher doesn't unwatch all of the items it has watched when it stops, all of the other watchers for that model can be blocked behind it. The flush method will sit there not sending any changes out, which has the effect that no watchers will fire. This can be time consuming to track down.

Since the watchers that use the hubwatcher should all be in tight loops receiving changes, add a 10s timeout to the send in flush() so that we can raise an error in the log. Capture the stack for each added watch and report that in the logged error to indicate the misbehaving watcher.

## QA steps

* No behaviour change in the normal case. Bootstrap and check that juju.state.watcher logging is as expected.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/state/watcher/hubwatcher.go
+++ b/state/watcher/hubwatcher.go
@@ -6,6 +6,7 @@ package watcher
 import (
 	"fmt"
 	"reflect"
+	"runtime/debug"
 	"time"
 
 	"github.com/juju/errors"
@@ -22,6 +23,10 @@ var (
 	// having notified no watchers to be considered idle.
 	HubWatcherIdleTime = 50 * time.Millisecond
 )
+
+// syncEventSendTimeout is the amount of time we'll wait to send an
+// event on a watch's channel before complaining about it.
+const syncEventSendTimeout = 10 * time.Second
 
 // HubSource represents the listening aspects of the pubsub hub.
 type HubSource interface {
@@ -253,6 +258,7 @@ func (w *HubWatcher) WatchMulti(collection string, ids []interface{}, ch chan<- 
 		ids:         ids,
 		completedCh: make(chan error),
 		watchCh:     ch,
+		source:      debug.Stack(),
 	}
 	return errors.Trace(w.sendAndWaitReq(req))
 }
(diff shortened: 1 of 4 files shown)


state/watcher/hubwatcher.go
```diff
@@ -464,7 +485,7 @@ func (w *HubWatcher) flush() bool {
 // handle deals with requests delivered by the public API
 // onto the background watcher goroutine.
 func (w *HubWatcher) handle(req interface{}) {
-	w.logger.Tracef("%p got request: %#v", w, req)
+	w.logger.Tracef("%p got request: %s", w, req)
```
babbageclunk: I changed these to avoid spamming the trace logging with big []uint8 junk.

state/watcher/hubwatcher.go
```diff
@@ -436,6 +458,9 @@ func (w *HubWatcher) flush() bool {
 			case e.ch <- outChange:
 				w.logger.Tracef("%p e.ch=%v has been notified %v", w, e.ch, outChange)
 				watchersNotified = true
+			case <-deadline:
+				w.logger.Errorf("%p e.ch=%v did not accept %v - missing Unwatch?\nwatch source:\n%s", w, e.ch, outChange, e.watchSource)
```
howbazaar: There is a bit of a weird race here, but not one that I think we care about. It is possible that we may iterate through the inner loop more than once and have the underlying sync event change due to the processing of `w.handle` request. I don't think this will every actually be a problem due to the size of the deadline, and the number of requests. All this really means is that we may inaccurately give a particular item 9.999s rather than 10s, so, no big deal.

Critical maybe? In similar areas elsewhere, I've prefixed the message with "programming error" as it isn't an expected error.

babbageclunk: Yeah, I'll change it to critical and indicate that it's a programming error.

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #10817, opened 2019-10-29
New ask-or-tell.

anastasiamac:
## Description of change

This PR changes cloud level commands (CRUD for clouds, credentials and k8s). 
Since these commands can operate on both client and controller copies of the entities, the ambiguity of what user intends needs to be clarified. Thus, when user supplies:
* ***default, i.e. no options supplied***: User will be prompted to confirm whether the operation is to take place on the client and or controller (and if on controller, then which one). Once the answers are obtained, the operation will proceed accordingly;
* ***--client***: the operation will take place on the client;
* ***-c/--controller <controller name>***: the operation will take place on the controller.

This approach also allows the combination of the options, i.e. when both --client and -c options are correctly provided, the operation will take place on both the client and the controller.

For this reason, --client-only option has been renamed to --client.

## Bug reference

https://bugs.launchpad.net/bugs/1849601

### DIFF
--- a/cmd/juju/caas/remove.go
+++ b/cmd/juju/caas/remove.go
@@ -104,13 +104,14 @@ func (c *RemoveCAASCommand) Run(ctxt *cmd.Context) error {
 		return errors.Trace(err)
 	}
 
-	if c.ControllerName == "" && !c.Client {
+	if c.ControllerName == "" {
 		ctxt.Infof("There are no controllers running.\nTo remove cloud %q from the current client, use the --client option.", c.cloudName)
 	}
 
-	// TODO(caas): only do RBAC cleanup for removing from both client and controller to less complexity.
-	if err := cleanUpCredentialRBACResources(c.cloudName, c.cloudMetadataStore, c.credentialStoreAPI); err != nil {
-		return errors.Trace(err)
+	if c.ControllerName != "" && c.Client { // TODO(caas): only do RBAC cleanup for removing from both client and controller to less complexity.
+		if err := cleanUpCredentialRBACResources(c.cloudName, c.cloudMetadataStore, c.credentialStoreAPI); err != nil {
+			return errors.Trace(err)
+		}
 	}
 
 	if c.Client {
--- a/cmd/juju/caas/remove.go
+++ b/cmd/juju/caas/remove.go
@@ -25,21 +25,11 @@ Removes the specified k8s cloud from this client.
 If --controller is used, also removes the cloud 
 from the specified controller (if it is not in use).
 
-If --controller option was not used and the current controller can be detected, 
-a user will be prompted to confirm if the k8s cloud needs to be removed from it. 
-If the prompt is not needed and the k8s cloud is always to be removed from
-the current controller if that controller is detected, use --no-prompt option.
-
-If you just want to update your controller and not
-your current client, use the --controller-only option.
-
-If you just want to update your current client and not
-a running controller, use the --client-only option.
+Use --client option to update your current client.
 
 Examples:
     juju remove-k8s myk8scloud
-    juju remove-k8s myk8scloud --client-only
-    juju remove-k8s myk8scloud --controller-only --no-prompt
+    juju remove-k8s myk8scloud --client
     juju remove-k8s --controller mycontroller myk8scloud
     
 See also:
@@ -110,30 +100,20 @@ func (c *RemoveCAASCommand) Init(args []string) (err error) {
 
 // Run is defined on the Command interface.
 func (c *RemoveCAASCommand) Run(ctxt *cmd.Context) error {
-	if c.BothClientAndController || c.ControllerOnly {
-		if c.ControllerName == "" {
-			// The user may have specified the controller via a --controller option.
-			// If not, let's see if there is a current controller that can be detected.
-			var err error
-			c.ControllerName, err = c.MaybePromptCurrentController(ctxt, fmt.Sprintf("remove k8s cloud %v from ", c.cloudName))
-			if err != nil {
-				return errors.Trace(err)
-			}
-		}
+	if err := c.MaybePrompt(ctxt, fmt.Sprintf("remove k8s cloud %v from ", c.cloudName)); err != nil {
+		return errors.Trace(err)
 	}
 
-	if c.ControllerName == "" && !c.ClientOnly {
-		ctxt.Infof("There are no controllers running.\nTo remove cloud %q from the current client, use the --client-only option.", c.cloudName)
+	if c.ControllerName == "" && !c.Client {
+		ctxt.Infof("There are no controllers running.\nTo remove cloud %q from the current client, use the --client option.", c.cloudName)
 	}
 
-	if c.BothClientAndController {
-		// TODO(caas): only do RBAC cleanup for removing from both client and controller to less complexity.
-		if err := cleanUpCredentialRBACResources(c.cloudName, c.cloudMetadataStore, c.credentialStoreAPI); err != nil {
-			return errors.Trace(err)
-		}
+	// TODO(caas): only do RBAC cleanup for removing from both client and controller to less complexity.
+	if err := cleanUpCredentialRBACResources(c.cloudName, c.cloudMetadataStore, c.credentialStoreAPI); err != nil {
+		return errors.Trace(err)
 	}
 
-	if c.BothClientAndController || c.ClientOnly {
+	if c.Client {
 		if err := removeCloudFromLocal(c.cloudName, c.cloudMetadataStore); err != nil {
 			return errors.Annotatef(err, "cannot remove cloud from current client")
 		}
@@ -141,11 +121,8 @@ func (c *RemoveCAASCommand) Run(ctxt *cmd.Context) error {
 		if err := c.Store.UpdateCredential(c.cloudName, cloud.CloudCredential{}); err != nil {
 			return errors.Annotatef(err, "cannot remove credential from current client")
 		}
-		if c.ClientOnly {
-			return nil
-		}
 	}
-	if c.BothClientAndController || c.ControllerOnly {
+	if c.ControllerName != "" {
 		if err := jujuclient.ValidateControllerName(c.ControllerName); err != nil {
 			return errors.Trace(err)
 		}
@@ -173,7 +150,7 @@ func cleanUpCredentialRBACResources(
 	if personalClouds == nil {
 		return nil
 	}
-	cloud, ok := personalClouds[cloudName]
+	pCloud, ok := personalClouds[cloudName]
 	if !ok {
 		return nil
 	}
@@ -186,7 +163,7 @@ func cleanUpCredentialRBACResources(
 		return nil
 	}
 	for _, credential := range cloudCredentials.AuthCredentials {
-		if err := cleanUpCredentialRBAC(cloud, credential); err != nil {
+		if err := cleanUpCredentialRBAC(pCloud, credential); err != nil {
 			logger.Warningf("unable to remove RBAC resources for credential %q", credential.Label)
 		}
 	}
--- a/cmd/juju/caas/remove_test.go
+++ b/cmd/juju/caas/remove_test.go
@@ -130,8 +130,8 @@ func (s *removeCAASSuite) TestRemoveControllerOnly(c *gc.C) {
 	s.fakeCloudAPI.CheckCall(c, 0, "RemoveCloud", "myk8s")
 
 	// client side operations
-	s.cloudMetadataStore.CheckCallNames(c, "PersonalCloudMetadata")
-	s.store.CheckCallNames(c, "CredentialForCloud")
+	s.cloudMetadataStore.CheckNoCalls(c)
+	s.store.CheckNoCalls(c)
 }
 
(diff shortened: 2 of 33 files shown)


anastasiamac:
[new ask-or-tell](https://trello.com/c/ETLjraHK/963-new-ask-or-tell)

cmd/juju/cloud/listcredentials.go
```diff
@@ -243,15 +235,22 @@ func (c *listCredentialsCommand) Run(ctxt *cmd.Context) error {
 		ctxt.Infof("secrets are not shown in tabular format")
 		c.showSecrets = false
 	}
-	credentials := credentialsMap{ClientOnly: c.ClientOnly}
+	credentials := credentialsMap{ClientOnly: c.Client}
+	cloudMsg := ""
+	if c.cloudName != "" {
+		cloudMsg = fmt.Sprintf("for cloud %q ", c.cloudName)
+	}
+	if err := c.MaybePrompt(ctxt, fmt.Sprintf("list credentials %vfrom", cloudMsg)); err != nil {
```
wallyworld: this prompt will be wrong if c.cloudName == "" since cloudMsg will be ""

cmd/juju/cloud/listcredentials.go
```diff
@@ -383,11 +366,12 @@ func formatCredentialsTabular(writer io.Writer, value interface{}) error {
 		return errors.Errorf("expected value of type %T, got %T", credentials, value)
 	}
 
-	if len(credentials.Client) == 0 {
-		fmt.Fprintln(writer, "No credentials from this client to display.")
-	}
-	if len(credentials.Controller) == 0 {
-		if !credentials.ClientOnly {
+	if credentials.ClientOnly {
```
wallyworld: I'm unclear why the change from credentials.Client to credentials.ClientOnly

cmd/juju/cloud/remove.go
```diff
@@ -102,33 +93,23 @@ func (c *removeCloudCommand) Init(args []string) (err error) {
...
-			if err != nil {
-				return errors.Trace(err)
-			}
-		}
+	if err := c.MaybePrompt(ctxt, fmt.Sprintf("remove cloud %v from", c.Cloud)); err != nil {
+		return errors.Trace(err)
 	}
-	if c.ControllerName == "" && !c.ClientOnly {
-		ctxt.Infof("To remove cloud %q from this client, use the --client-only option.", c.Cloud)
+	if c.ControllerName == "" && !c.Client {
```
wallyworld: Here is a case where I think the logic is correct now but was wrong originally?

cmd/juju/cloud/remove.go
```diff
@@ -102,33 +93,23 @@ func (c *removeCloudCommand) Init(args []string) (err error) {
...
+		ctxt.Infof("To remove cloud %q from this client, use the --client option.", c.Cloud)
```
wallyworld: We can return here right?
If not controller and not client then nothing to do.
Should this logic be used elsewhere? It seems this is the only place where we do this check and can exit early? Maybe I'm wrong, lots of code to look at

cmd/juju/cloud/remove_test.go
```diff
@@ -144,8 +140,9 @@ func (s *removeSuite) TestRemoveCloudController(c *gc.C) {
 
 func (s *removeSuite) TestCannotRemovePublicCloud(c *gc.C) {
 	s.createTestCloudData(c)
-	ctx, err := cmdtesting.RunCommand(c, cloud.NewRemoveCloudCommand(), "prodstack", "--client-only")
-	c.Assert(err, jc.ErrorIsNil)
+	ctx, _ := cmdtesting.RunCommand(c, cloud.NewRemoveCloudCommand(), "prodstack", "--client")
+	//ctx, err := cmdtesting.RunCommand(c, cloud.NewRemoveCloudCommand(), "prodstack", "--client")
```
wallyworld: remove me

cmd/juju/cloud/removecredential.go
```diff
@@ -127,22 +116,14 @@ func (c *removeCredentialCommand) SetFlags(f *gnuflag.FlagSet) {
...
+		return errors.Trace(err)
 	}
-	if c.ControllerName == "" && !c.ClientOnly {
-		ctxt.Infof("To remove credential %q for cloud %q from this client, use the --client-only option.", c.credential, c.cloud)
+	if c.ControllerName == "" && !c.Client {
+		ctxt.Infof("To remove credential %q for cloud %q from this client, use the --client option.", c.credential, c.cloud)
 	}
 	var client RemoveCredentialAPI
-	if !c.ClientOnly {
+	if c.ControllerName != "" {
```
wallyworld: This logic can be moved down to the other if controllerName != "" block

cmd/juju/cloud/show.go
```diff
@@ -176,7 +160,7 @@ func (c *showCloudCommand) Run(ctxt *cmd.Context) error {
 			}
 		}
 	}
-	if c.BothClientAndController || c.ClientOnly {
+	if c.Client {
```
wallyworld: if c.Client && localCloud != nil
Actually, just if localCloud != nil is sufficient right?
And same with remoteCloud != nil above

cmd/juju/cloud/updatecloud.go
```diff
@@ -145,20 +138,11 @@ func (c *updateCloudCommand) Run(ctxt *cmd.Context) error {
...
-				return errors.Trace(err)
-			}
-		}
+	if err := c.MaybePrompt(ctxt, fmt.Sprintf("update cloud %q on", c.Cloud)); err != nil {
+		return errors.Trace(err)
 	}
-	if c.ControllerName == "" && !c.ClientOnly {
-		ctxt.Infof("To update cloud %q on this client, use the --client-only option.", c.Cloud)
+	if c.ControllerName == "" && !c.Client {
+		ctxt.Infof("To update cloud %q on this client, use the --client option.", c.Cloud)
```
wallyworld: we can return early here etc

cmd/juju/cloud/updatecredential.go
```diff
@@ -35,16 +35,9 @@ a model was created with to the new and valid details on controller.
...
-If the prompt is not needed and the cloud is always to be updated on
-the current controller if that controller is detected, use --no-prompt option.
+Use --controller option to update a cloud on a controller. 
 
-Use --controller option to update a cloud on a different controller. 
-
-Use --controller-only option to only update controller copy of the cloud.
-
-If --client-only is used, Juju updates credential only on this client.
+If --client is used, Juju updates credential on this client.
```
wallyworld: This wording differs from elsewhere

cmd/modelcmd/controller.go
```diff
@@ -418,8 +407,7 @@ type OptionalControllerCommand struct {
 // SetFlags initializes the flags supported by the command.
 func (c *OptionalControllerCommand) SetFlags(f *gnuflag.FlagSet) {
 	c.CommandBase.SetFlags(f)
-	f.BoolVar(&c.ClientOnly, "client-only", false, "Client operation only; controller not affected")
-	f.BoolVar(&c.ControllerOnly, "controller-only", false, "Controller operation only; client not affected")
+	f.BoolVar(&c.Client, "client", false, "Command to operate on or with a copy from this client")
```
wallyworld: i don't parse the description

cmd/modelcmd/controller.go
```diff
@@ -429,74 +417,81 @@ func (c *OptionalControllerCommand) SetFlags(f *gnuflag.FlagSet) {
 
 // Init populates the command with the args from the command line.
 func (c *OptionalControllerCommand) Init(args []string) (err error) {
-	if c.Local && !c.ClientOnly {
-		c.ClientOnly = c.Local
+	if c.Local && !c.Client {
```
wallyworld: c.Client = c.Client || c.Local

cmd/modelcmd/controller.go
```diff
@@ -429,74 +417,81 @@ func (c *OptionalControllerCommand) SetFlags(f *gnuflag.FlagSet) {
...
-	use, err := pollster.YN(fmt.Sprintf("Do you want to %v current controller %q", action, controllerName), true)
-	if err != nil {
-		return "", errors.Trace(err)
+	if defaultController == "" {
+		defaultController = controllers[0]
 	}
-	if use {
-		return controllerName, nil
+	cName, err := pollster.SelectVerify(interact.List{
+		Singular: "controller name",
```
wallyworld: Just "controller" and "controllers"

cmd/juju/cloud/listcredentials.go
anastasiamac: Yes, it is intentional and reads fine.
If cloud name is given it'll read 
```
list credentials for cloud AAAA from [client/controller]
````
If not, it'll read
```
list credentials from [cloud/controller]
```

I've experimented with this quiet a bit to get the wording and spacing correct.

File: cmd/juju/caas/remove.go [hunk 13]
```diff
@@ -25,21 +25,11 @@ Removes the specified k8s cloud from this client.
...
-a user will be prompted to confirm if the k8s cloud needs to be removed from it. 
-If the prompt is not needed and the k8s cloud is always to be removed from
-the current controller if that controller is detected, use --no-prompt option.
-
-If you just want to update your controller and not
-your current client, use the --controller-only option.
-
-If you just want to update your current client and not
-a running controller, use the --client-only option.
+Use --client option to update your current client.
```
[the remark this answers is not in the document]
anastasiamac: In all our commands that I have just looked at, we do not add an article in this cases.

I have done the search on "Use -" and we apply phrase it this way everywhere - not just cloud and caas package but applications, backups, action, etc.

cmd/juju/caas/remove.go
```diff
@@ -25,21 +25,11 @@ Removes the specified k8s cloud from this client.
...
 
 Examples:
     juju remove-k8s myk8scloud
-    juju remove-k8s myk8scloud --client-only
-    juju remove-k8s myk8scloud --controller-only --no-prompt
+    juju remove-k8s myk8scloud --client
```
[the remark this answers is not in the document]
anastasiamac: I do not think so... We do not always have all possible combinations and permutations of options.

cmd/juju/caas/remove_test.go
```diff
@@ -124,63 +122,31 @@ func (s *removeCAASSuite) TestRemove(c *gc.C) {
...
 
 	// client side operations
-	s.cloudMetadataStore.CheckNoCalls(c)
-	s.store.CheckNoCalls(c)
+	s.cloudMetadataStore.CheckCallNames(c, "PersonalCloudMetadata")
+	s.store.CheckCallNames(c, "CredentialForCloud")
 }
 
-func (s *removeCAASSuite) TestRemoveLocalOnly(c *gc.C) {
+func (s *removeCAASSuite) TestRemoveClientOnly(c *gc.C) {
```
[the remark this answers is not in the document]
anastasiamac: This test checks that the change is only on the client so the naming still makes sense.

cmd/juju/cloud/listcredentials.go
anastasiamac: Here .Client has the list of credentials whereas .ClientOnly holds whether the client list was requested or not. The difference is in principal and is needed in the formatter when we get an empty list - is it that there are no credentials on the client or is it that we were not even looking on the client for credentials.

anastasiamac: I've added comments to clarify.

cmd/juju/cloud/remove.go
anastasiamac: Actually, I do not think that we need it anymore...

It looks like long ago, the original intent of this code was to detect that the controller and client operation was requested but no controller was specified/detected. In this case, in the next run of this command, the user could have just said "--client"... This is what this message is about - the information/education of the user. 

I do not think it's needed anymore as we no longer detect controllers and users need to specify them explicitly.

cmd/juju/cloud/removecredential.go
anastasiamac: No, it can't since the line right after it uses this client. We are checking remote clouds and needed to open that connection.

cmd/juju/cloud/show.go
anastasiamac: Well, no. I've simplified the section a little.

But basically we need to know whether user has requested a client cloud or not, especially in case where we just did not find it.

This is needed because we are displaying config after the cloud details. If both controller and client clouds were requested and since we display controller clouds first, we need to know if we are showing the config of remote cloud or not.

cmd/juju/cloud/updatecloud.go
anastasiamac: No, this was information message. And anyway, it's gone now.

cmd/juju/cloud/updatecredential.go
anastasiamac: yes, some original wording makes command help unique. The same is on k8s commands. I've updated what I needed here.

cmd/modelcmd/controller.go
anastasiamac: Changed to "Client operation" echoing deprecated "Local operation only,..." and "Controller to operate in"

anastasiamac:
$$merge$$

PR #10818, opened 2019-10-29
Ensure all k8s resources have correct annotations

wallyworld:
## Description of change

When creating k8s resources, ensure that the annotations contain the controller uuid.
Also, any resource tags set in model config should be included in the annotations. This was done for some things but not all.

## QA steps

bootstrap k8s and deploy workload
kubectl describe the pods, services, service accounts etc to ensure juju.io/controller is set


wallyworld:
$$merge$$

PR #10819, opened 2019-10-29
Fix init container issue upgrading from 2.6

wallyworld:
## Description of change

When upgrading a k8s controller from 2.6, for non controller models we were attempting to run a jujud command in an init container which is not implemented until 2.7, since those models are still using a 2.6 jujud. 

Tweak the init container script to account for this.

## QA steps

juju bootstrap a k8s 2.6 controller
Add a model and deploy a workload
upgrade-controller to 2.7
see that the 2.6 workload pods are ok
upgrade the model to 2.7


wallyworld:
[2.6.9 -> 2.7 pod init issue](https://trello.com/c/uIZhWEhb/964-269-27-pod-init-issue)

wallyworld:
$$merge$$

PR #10824, opened 2019-10-29
Attempt to use github mirror.

SimonRichardson:
## Description of change

The following uses github as a mirror for the flaky golang.org
service. This should attempt to get the source from the same
locations as the other deps and make our dependency woes less
sucky.

More investigation should be done around this as it doesn't look
like the lock file picks up the changes, but essentially this should
work according to the docs.

## QA steps

```sh
make rebuild-dependencies
make dep
```


SimonRichardson:
closing in favour of https://github.com/juju/juju/pull/10832

PR #10827, opened 2019-10-30
Fix intermittent failures in RelationUnit watch tests

babbageclunk:
## Description of change

They would occasionally fail with watchers being triggered unexpectedly - adding WaitForModelWatchersIdle fixes these under stress-race.

## QA steps

* Run under stress-race; it doesn't fail.

## Documentation changes

None

## Bug reference

None


babbageclunk:
$$merge$$

PR #10829, opened 2019-10-30
ask-or-tell: one question

anastasiamac:
## Description of change

Ask-or-tell PR to cater for different cases when neither --client nor --controller is specified on a cloud level command:

1. There is no current controller and no controllers registered on a client: confirm (Y/n) whether the operation is to be run on the client but also inform the user that a controller can be registered or bootstrapped;

2. There is no current controller but other controllers are registered: confirm (Y/n) whether the operation is to be run on the client and also inform the user that a controller can be specified;

3. There is a current controller:  multi-choice prompt to operate (1) on a client; (2) on a controller, (3) on both. This option caters for invalid input as well as for a quit for prompt.

### DIFF
--- a/cmd/modelcmd/controller.go
+++ b/cmd/modelcmd/controller.go
@@ -402,18 +402,19 @@ type OptionalControllerCommand struct {
 
 	ControllerName string
 
-	SkipCurrentControllerPrompt bool
+	// ReadOnly read only commands do not require to prompt the user for clarification
+	// on whether the client or current controller is to be used.
+	ReadOnly bool
 }
 
 // SetFlags initializes the flags supported by the command.
 func (c *OptionalControllerCommand) SetFlags(f *gnuflag.FlagSet) {
 	c.CommandBase.SetFlags(f)
 	f.BoolVar(&c.Client, "client", false, "Client operation")
-	// TODO (juju3) remove me
-	f.BoolVar(&c.Local, "local", false, "DEPRECATED (use --client-only): Local operation only; controller not affected")
 	f.StringVar(&c.ControllerName, "c", "", "Controller to operate in")
 	f.StringVar(&c.ControllerName, "controller", "", "")
-	f.BoolVar(&c.SkipCurrentControllerPrompt, "no-prompt", false, "Skip prompting for confirmation to use current controller, always use it when detected")
+	// TODO (juju3) remove me
+	f.BoolVar(&c.Local, "local", false, "DEPRECATED (use --client-only): Local operation only; controller not affected")
 }
 
 // Init populates the command with the args from the command line.
@@ -437,13 +438,19 @@ func (c *OptionalControllerCommand) MaybePrompt(ctxt *cmd.Context, action string
 		return nil
 	}
 
-	ctxt.Infof("This operation can be applied to both a copy on this client and to the one on a controller.")
-
 	currentController, err := DetermineCurrentController(c.Store)
 	if err != nil && !errors.IsNotFound(err) {
 		return errors.Trace(err)
 	}
 
+	if c.ReadOnly {
+		// No need to prompt the user, just assume that both client and a current controller are needed.
+		c.Client = true
+		c.ControllerName = currentController
+		return nil
+	}
+
+	ctxt.Infof("This operation can be applied to both a copy on this client and to the one on a controller.")
 	if currentController == "" {
 		msg := "No current controller was detected"
 		all, err := c.Store.AllControllers()
--- a/cmd/modelcmd/controller.go
+++ b/cmd/modelcmd/controller.go
@@ -462,9 +462,9 @@ func (c *OptionalControllerCommand) MaybePrompt(ctxt *cmd.Context, action string
 			return errors.Trace(err)
 		}
 		if !useClient {
-			return errors.New("Neither client nor controller specified - nothing to do.")
+			ctxt.Infof("Neither client nor controller specified - nothing to do.")
 		}
-		c.Client = true
+		c.Client = useClient
 		return nil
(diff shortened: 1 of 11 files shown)


anastasiamac:
[ask-or-tell 1 question](https://trello.com/c/eQzUPYqr/967-ask-or-tell-1-question)

cmd/modelcmd/controller.go
```diff
@@ -425,92 +425,80 @@ func (c *OptionalControllerCommand) Init(args []string) (err error) {
 // MaybePrompt checks if the command was give a --client or --controller options.
 // If not, it will prompt user to clarify whether the operation is to take place on
 // a client copy, a controller copy or both.
+// When neither client nor controller is specified on the command,
+// several scenarios need to be catered for when prompting:
+// 1. there is a current controller to prompt user with;
+// 2. there is no current controller but there are other controllers, so inform users how to use -c;
+// 3. there is no current controller and there are no registered controllers, so inform users to bootstrap or register.
+// When there is no current controller, the prompt becomes a Y/N client question instead of a multi-choice
```
wallyworld: The other option for this last scenario is when the command is read only, ie list, show, just do it on the client and include in the output what was done. Adds/removes/updates warrant the prompt. Just a thought, could be done as an improvement next

cmd/modelcmd/controller_test.go
```diff
@@ -171,156 +170,176 @@ func (s *OptionalControllerCommandSuite) TestControllerCommandLocal(c *gc.C) {
...
+	expectedPrompt          string
+	expectedInfo            string
+	expectedErr             string
+	expectedControllerName  string
+	expectedClientOperation bool
+	args                    []string
+}
+
+func (s *OptionalControllerCommandSuite) assertPrompted(c *gc.C, store jujuclient.ClientStore, t testData) {
+	//ctx, command, _ := s.assertPrompt(c, store, t.action, t.userAnswer, t.args...)
```
wallyworld: delete me

cmd/modelcmd/controller_test.go
```diff
@@ -171,156 +170,176 @@ func (s *OptionalControllerCommandSuite) TestControllerCommandLocal(c *gc.C) {
...
+	s.assertPrompted(c, setupTestStore(), testData{
+		action:       "build a snowman on",
+		expectedInfo: "This operation can be applied to both a copy on this client and to the one on a controller.\n",
+		expectedPrompt: `
+Do you want to build a snowman on:
+    1. client only (--client)
+    2. controller "fred" only (--controller fred)
+    3. both (--client --controller fred)
+Enter your choice, or type Q|q to quit: `[1:],
+		userAnswer:              "q\n",
```
wallyworld: Can we test with 'Q' as well?

cmd/modelcmd/controller_test.go
```diff
@@ -171,156 +170,176 @@ func (s *OptionalControllerCommandSuite) TestControllerCommandLocal(c *gc.C) {
...
-`[1:]
-
-var testShortOutput = `
-This operation can be applied to both a copy on this client and a controller of your choice.
-Do you want to test on this client? (Y/n): 
-Do you want to test on a controller? (Y/n): 
-`[1:]
+func (s *OptionalControllerCommandSuite) TestPromptInvalidChoice(c *gc.C) {
+	s.assertPrompted(c, setupTestStore(), testData{
+		action: "build a snowman on",
```
wallyworld: lol

wallyworld:
ERROR Neither client nor controller specified - nothing to do.

Not sure it warrants an error. Info at best, ie just display as straight text on the console

wallyworld:
```
$ juju remove-cloud foo
This operation can be applied to both a copy on this client and to the one on a controller.
No current controller was detected and there are no registered controllers on this client: either bootstrap one or register one.
Do you ONLY want to add k8s cloud foo to this client? (Y/n):
```

Can we wrap the 2 info lines above - they are very long

wallyworld:
Testing notes for other reviewers:

1. there are controller(s) bit none is current/default
```
$ juju add-k8s foo
This operation can be applied to both a copy on this client and to the one on a controller.
No current controller was detected and there are no registered controllers on this client: either bootstrap one or register one.
Do you ONLY want to add k8s cloud foo to this client? (Y/n):
```

2. there is a current controller (out of one or more registered on the client)
```
$ juju add-k8s foo
This operation can be applied to both a copy on this client and to the one on a controller.
Do you want to add k8s cloud foo to:
    1. client only (--client)
    2. controller "test" only (--controller test)
    3. both (--client --controller test)
Enter your choice, or type Q|q to quit:
```

3. there are no controllers at all registered
```
$ juju add-k8s foo
This operation can be applied to both a copy on this client and to the one on a controller.
No current controller was detected and there are no registered controllers on this client: either bootstrap one or register one.
Do you ONLY want to add k8s cloud foo to this client? (Y/n):
```

cmd/modelcmd/controller.go
anastasiamac: As an improvement next... But then again, feels like a potentially nice improvement for all the commands... When there is only one choice, why not just do it and inform the user? All our operations are reversable if user makes a mistake.

anastasiamac:
@wallyworld 
>> Can we wrap the 2 info lines above - they are very long

Sure \o/ Suggestions? :D

howbazaar:
I do like the change in the question when we ask.

It does seem weird that we ask for the read-only commands rather than just listing both. We should definitely be asking around any modification commands, but if we are just listing or showing, then I think it would be better to just show both.

anastasiamac:
$$merge$$

PR #10832, opened 2019-10-30
Retry all snapcraft run commands

SimonRichardson:
## Description of change

The following attempts to make installing dependencies for building
snaps more reliable. Essentially we're going to try and retry all
run commands so at least we've not bailed out on the first try.

This should hopefully improve the current ongoing failures we have
when building snaps.


SimonRichardson:
@jameinel @mitechie I'm going to merge this after the rc1 release.

SimonRichardson:
$$merge$$

PR #10835, opened 2019-10-31
Sort instance IDs to fix intermittent test failure in ec2 provider

achilleasa:
## Description of change

Fix for an intermittent failure for test introduced to the ec2 provider in #10805.

### DIFF
--- a/provider/ec2/local_test.go
+++ b/provider/ec2/local_test.go
@@ -1622,6 +1622,9 @@ func (t *localServerSuite) TestNetworkInterfacesForMultipleInstances(c *gc.C) {
 		ids[i] = inst.Id()
 	}
 
+	// Sort instance list so we always get consistent results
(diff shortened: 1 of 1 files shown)


provider/ec2/local_test.go
```diff
@@ -1622,6 +1622,9 @@ func (t *localServerSuite) TestNetworkInterfacesForMultipleInstances(c *gc.C) {
 		ids[i] = inst.Id()
 	}
 
+	// Sort instance list so we always get consistent results
+	sort.Slice(ids, func(l, r int) bool { return ids[l] < ids[r] })
```
jameinel: instance.Id is:
type Id string

so you can just use
sort.Strings([]string(ids))

achilleasa: This will cause a compiler error because the slice type is `[]instance.Id` which while equivalent to `[]string` confuses the Go compiler.

achilleasa:
$$merge$$

PR #10839, opened 2019-10-31
Dev fix incorrect handling of endpoint bindings when migrating

achilleasa:
## Description of change

When migrating from a 2.6 -> 2.7 controller, we must be careful to properly re-map the endpoint bindings from the 2.6 application description so the 2.7 code can properly parse them into a Bindings object. 

If a 2.6 application had **all** its bindings assigned to the default space (represented as `""` in 2.6 controllers), the NewBindings c-tor would get confused and treat the map as if it contained space IDs instead of space names and thus preventing the operator from adding new units (the endpoint bindings document would contain `""` values instead of the default space ID)

Note: when the DefaultAlphaSpaceName changes land, we should update the code in import_migration to re-map the 2.6 default space (`""`) to the alpha space name.

## QA steps

```console
# bootstrap 2.6 controller on lxd
$ juju add-model test
# wait for the unit machine to boot...

# bootstrap 2.7 controller on lxd
$ juju switch 2.6
$ juju migrate test 2.7
$ juju switch 2.7
$ juju add-unit ubuntu-lite
```

If you peek into the DB, the endpointbindings document for the application should map all endpoints to "0" **and include a key "" with value "0"** (2.6 only sets the key for the default space if different than network.DefaultSpaceName whereas 2.7 always sets it)

## Bug reference

Tentative fix for https://bugs.launchpad.net/juju/+bug/1850589

### DIFF
--- a/state/endpoint_bindings.go
+++ b/state/endpoint_bindings.go
@@ -464,13 +464,20 @@ func NewBindings(st EndpointBinding, givenMap map[string]string) (*Bindings, err
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	namesErr := allOfOne(spacesNamesToIDs, givenMap)
+
+	// If givenMap contains space names empty values are allowed (e.g. they
+	// may be present when migrating a model from a 2.6.x controller).
+	namesErr := allOfOne(spacesNamesToIDs, givenMap, true)
 
 	spacesIDsToNames, err := st.SpaceNamesByID()
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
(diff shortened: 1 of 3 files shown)


achilleasa:
!!build!!

hmlanigan:
QA procedures above pass.  Per 1850589: I also deployed the kubernetes-core bundle on 2.6.10 and migrated to a new model for test.  No no-op log messages found in the new controller's debug logs, nor " ERROR juju.worker.dependency "undertaker" manifold "

hmlanigan:
$$merge$$

PR #10841, opened 2019-11-01
Check the model version to decide the action ID type;

ycliuhw:
## Description of change

Use `UUID` for the actionID if the model agent version < 2.7;

## CaaS

```console
# controller version 2.7
$ juju controllers
Controller  Model  User   Access     Cloud/Region        Models  Nodes  HA  Version
k1*         t1     admin  superuser  microk8s/localhost       2      1   -  2.7.0


# model version 2.6
$ juju status --relations --color --storage -m k1:t1                                                                        
Model  Controller  Cloud/Region        Version  SLA          Timestamp
t1     k1          microk8s/localhost  2.6.11   unsupported  13:47:03+11:00

# run action got errors, that's ok because 2.6 doesnot support k8s actions yet.
$ juju run --unit mariadb-k8s/0 --timeout=30s "network-get server" --operator=false
ERROR timed out waiting for result from: unit mariadb-k8s/0

# check actionID in DB - it's `UUID`
juju:PRIMARY> db.actions.find().pretty()
{
	"_id" : "146d3193-4756-4492-872c-db9ba112be6f:b27e5bcc-9b7e-4817-8b14-632ba7715a94",
	"model-uuid" : "146d3193-4756-4492-872c-db9ba112be6f",
	"receiver" : "mariadb-k8s/0",
	"name" : "juju-run",
	"parameters" : {
		"workload-context" : true,
		"command" : "network-get server",
		"timeout" : NumberLong("30000000000")
	},
	"enqueued" : ISODate("2019-11-01T02:43:35Z"),
	"started" : ISODate("0001-01-01T00:00:00Z"),
	"completed" : ISODate("0001-01-01T00:00:00Z"),
	"status" : "pending",
	"message" : "",
	"results" : {

	},
	"messages" : [ ],
	"txn-revno" : NumberLong(2),
	"txn-queue" : [
		"5dbb9bd6fe5c050008b25e02_d6a87d84"
	]
}

# upgrade model;
$ juju upgrade-juju --agent-version 2.7.0 -m t1
best version:
    2.7.0
started upgrade to 2.7.0

# run action again
$ juju run --unit mariadb-k8s/0 --timeout=30s "network-get server" --operator=false

# check actionID in DB - it's `number`
juju:PRIMARY> db.actions.find().pretty()
{
	"_id" : "146d3193-4756-4492-872c-db9ba112be6f:4",
	"model-uuid" : "146d3193-4756-4492-872c-db9ba112be6f",
	"receiver" : "mariadb-k8s/0",
	"name" : "juju-run",
	"parameters" : {
		"command" : "network-get server",
		"timeout" : NumberLong("30000000000"),
		"workload-context" : true
	},
	"enqueued" : ISODate("2019-11-01T03:29:54Z"),
	"started" : ISODate("0001-01-01T00:00:00Z"),
	"completed" : ISODate("0001-01-01T00:00:00Z"),
	"status" : "pending",
	"message" : "",
	"results" : {

	},
	"messages" : [ ],
	"txn-revno" : NumberLong(2),
	"txn-queue" : [
		"5dbba6b1fe5c0500085bb340_86016a6f"
	]
}

```

## IaaS

```console
# 2.7 controller
$ juju controllers

Controller  Model    User   Access     Cloud/Region         Models  Nodes    HA  Version
k1*         default  admin  superuser  localhost/localhost       2      1  none  2.7.0.1

# 2.6 model
$ juju status --relations --color  --storage -m k1:default
Model    Controller  Cloud/Region         Version   SLA          Timestamp
default  k1          localhost/localhost  2.6.11.1  unsupported  16:58:48+11:00

# run action - `UUID` format of action ID;
$ juju run-action  postgresql/1 wal-e-list-backups
Action queued with id: 8ee87083-6a85-4477-8483-515f741218ba

# upgrade model to 2.7
$ juju upgrade-juju -m default
best version:
    2.7.0.1
started upgrade to 2.7.0.1

# run action - `number` format of action ID;
$ juju run-action  postgresql/1 wal-e-list-backups
Action queued with id: "1"
```

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/state/action.go
+++ b/state/action.go
@@ -332,10 +332,13 @@ func newAction(st *State, adoc actionDoc) Action {
 	}
 }
 
-var newActionIDFirstSupportedVersion = version.MustParse("2.7.0")
+// MinVersionSupportNewActionID should be un-exposed after 2.7 released.
+// TODO(action): un-expose MinVersionSupportNewActionID and IsNewActionIDSupported and remove those helper functions using these two vars in tests from 2.7.0.
+var MinVersionSupportNewActionID = version.MustParse("2.7.0")
 
-func newActionIDSupported(ver version.Number) bool {
-	return ver.Compare(newActionIDFirstSupportedVersion) >= 0
+// IsNewActionIDSupported checks if new action ID is supported for the specified version.
+func IsNewActionIDSupported(ver version.Number) bool {
+	return ver.Compare(MinVersionSupportNewActionID) >= 0
 }
 
 // newActionDoc builds the actionDoc with the given name and parameters.
@@ -347,7 +350,7 @@ func newActionDoc(mb modelBackend, receiverTag names.Tag, actionName string, par
 
 	// we support machine actions anymore.
 	var actionId string
-	if receiverTag.Kind() == names.UnitTagKind && newActionIDSupported(modelAgentVersion) {
+	if receiverTag.Kind() == names.UnitTagKind && IsNewActionIDSupported(modelAgentVersion) {
 		id, err := sequence(mb, "task")
 		if err != nil {
 			return actionDoc{}, actionNotificationDoc{}, err
--- a/state/action.go
+++ b/state/action.go
@@ -10,6 +10,7 @@ import (
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	"github.com/juju/utils"
+	"github.com/juju/version"
 	"gopkg.in/juju/names.v3"
 	"gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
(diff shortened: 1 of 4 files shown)


ycliuhw:
$$merge$$

state/action.go
```diff
@@ -331,15 +332,25 @@ func newAction(st *State, adoc actionDoc) Action {
 	}
 }
 
+// MinVersionSupportNewActionID should be un-exposed after 2.7 released.
+// TODO(action): un-expose MinVersionSupportNewActionID and IsNewActionIDSupported and remove those helper functions using these two vars in tests from 2.7.0.
+var MinVersionSupportNewActionID = version.MustParse("2.7.0")
```
jameinel: Why are we using a MinVersion check here? This is the reason we added Facade versioning. It should use the fact that the Uniter API is at an older version to maintain consistency, not a modelAgentVersion. If we need to pass a boolean from the Uniter API to indicate that it needs to do an old method of generating action identifiers, then it can do so.

But 

a) We should never remove the compatibility check until we actually drop compatibility with the *old* version of the agent. (eg, if we remove this, then when someone in the future upgrades a controller from 2.5 => 2.8 all of the model agents that aren't upgraded yet will break.)

b) We already have Versioned API facades to do this sort of check. Since we expect to do it at that level, that is more robust than doing it as a model agent version check.


Is there a reason this was done at the State layer instead?

jameinel: Ah, maybe the issue is that the client requesting the Action is 2.7, and the Controller is 2.7, but the agents that are watching for actions are 2.6 (thus a different API that is creating this vs the one that is consuming it).
That makes a bit more sense. But I'd like us to update the comment to indicate that it isn't about when 2.7 is released or not, but that it is compatibility with older unit agents.

ycliuhw: What I meant was to change MinVersionSupportNewActionID to `minVersionSupportNewActionID` after 2.7 released. 
It's exposed to the outside of the state package for passing some tests(The `currentVersion` is smaller than this min version which failed some new action ID related tests.).
Sorry for the confusion.

PR #10842, opened 2019-11-01
Allow add-k8s to work with no undercloud specified

wallyworld:
## Description of change

There's a need to use add-k8s without specify any underlying cloud.
This PR supports that case, and records the cloud type as "other".
The cloud type is used internally by Juju when creating the controller service type. The service type for "other" is now "ClusterIP" instead of "LoadBalancer" as we can't always be sure a load balancer service is supported. To allow bootstrapping in more cases, two new boostrap config params are added:
controller-service-type (cluster, external, loadbalancer)
controller-external-ips

This fixes another bug preventing bootstrap to a remote microk8s, plus allows the user to control what type of controller service is created by Juju to cover more scenarios on bespoke k8s clusters.
Known clusters like GKE etc still work out of the box.

Note - add-k8s will use a default storage class if one is found if no undercloud is specified. So most times just "juju add-k8s" is sufficient.

## QA steps

add-k8s on a cluster with no undercloud
bootstrap to that cluster using --config controller-service-type=blah

## Bug reference

https://bugs.launchpad.net/bugs/1850877
https://bugs.launchpad.net/juju/+bug/1841960

### DIFF
--- a/caas/kubernetes/provider/bootstrap.go
+++ b/caas/kubernetes/provider/bootstrap.go
@@ -52,11 +52,13 @@ const (
 type controllerServiceSpec struct {
 	// ServiceType is required.
 	ServiceType core.ServiceType
+	// ExternalIPs is optional.
+	ExternalIPs []string
 	// Annotations is optional.
 	Annotations k8sannotations.Annotation
 }
 
-var controllerServiceSpecs = map[string]*controllerServiceSpec{
+var controllerServiceSpecs = map[string]controllerServiceSpec{
 	caas.K8sCloudAzure: {
 		ServiceType: core.ServiceTypeLoadBalancer,
 	},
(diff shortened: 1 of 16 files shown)


caas/kubernetes/provider/bootstrap.go
```diff
@@ -343,24 +345,30 @@ func (c *controllerStack) Deploy() (err error) {
...
-func (c *controllerStack) getControllerSvcSpec(cloudType string) (*controllerServiceSpec, error) {
+func (c *controllerStack) getControllerSvcSpec(cloudType string, cfg *podcfg.BootstrapConfig) (*controllerServiceSpec, error) {
 	spec, ok := controllerServiceSpecs[cloudType]
 	if !ok {
 		logger.Debugf("fallback to default svc spec for %q", cloudType)
 		spec = controllerServiceSpecs[caas.K8sCloudOther]
 	}
-	if spec == nil || spec.ServiceType == "" {
+	if cfg != nil && cfg.ControllerServiceType != "" {
+		spec.ServiceType = core.ServiceType(cfg.ControllerServiceType)
```
hpidcock: This is wrong. See https://github.com/juju/juju/blob/57deffaf61559ec3ff6ceeb31f2d3ef1e22e3999/caas/kubernetes/provider/k8s.go#L1650, it needs to be converted from a juju service type into a k8s service type.

wallyworld: yeah, good catch, fixed

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #10846, opened 2019-11-03
Move state multiwatcher package

howbazaar:
As a prelude to making the all watchers combined and their own worker, it made sense to move the actual multiwatcher delta types.

Due to name clashes in the params package, two of the delta types have been renamed to use Update suffix rather than Info.

There is only one actual difference in this change, and it is in serialization param for the address type. The new Address uses "space-id" for the provider ID. The old structure used something else. I'm almost 100% confident that this was never used by the GUI nor anyone else, and I didn't feel that it necessitated having its own type just for something that was unused.


howbazaar:
Oh, part of this was also to move the Job and Block constants into core/model rather than the params package.

These constants were used throughout the code, and had no place in the params package.

apiserver/params/applications.go
```diff
@@ -422,8 +422,9 @@ type ScaleApplicationInfo struct {
 	Scale int `json:"num-units"`
 }
 
-// ApplicationInfo holds an application info.
-type ApplicationInfo struct {
+// ApplicationResult holds an application info.
+// NOTE: we should look to combine ApplicationResult and ApplicationInfo.
+type ApplicationResult struct {
```
babbageclunk: Seems weird to have a type named *Result without any error field.

howbazaar:
$$merge$$

PR #10847, opened 2019-11-03
Fix race+test with actions blocked and snapshot.

hpidcock:
## Description of change

Fix non-critical race condition with actions blocked and Snapshot.
If Snapshot is called too quickly it wouldn't return the expected state.

## QA steps

Run tests for github.com/juju/juju/worker/uniter/remotestate

## Documentation changes

N/A

## Bug reference

N/A


hpidcock:
$$merge$$

PR #10848, opened 2019-11-04
Fix k8s controller config validation

wallyworld:
## Description of change

When validation k8s controller config, "loadbalancer" services are also allowed to have external ips, not just "external" services. Also, "loadbalancer" services can just specify 1 external IP.

## QA steps

$ juju bootstrap microk8s test
  --config controller-service-type=loadbalancer
  --config controller-external-ips=[10.0.0.1]


wallyworld:
$$merge$$

PR #10852, opened 2019-11-04
update  nw-export-bundle to use bionic bundle

hmlanigan:
## Description of change

Allow nw-export-bundle test to succeed, use bionic bundle.

Adding new bundle which uses bionic as media wiki hasn't been
updated past trusty.  Remove the old bundle.

## QA steps

./assess_bundle_export.py lxd /snap/bin/juju /tmp/artifacts


hmlanigan:
$$merge$$

hmlanigan:
> Nice, we should move this to the new integration tests.

Agreed!

PR #10856, opened 2019-11-05
CI test for application-level relation data

babbageclunk:
## Description of change

This is a basic smoke test that the application-level data can get from the source application to the sink units. More detailed test cases will be added in the future, but I wanted to get feedback about it general setup first.

## QA steps

* Run the test - it passes.

## Documentation changes

None

## Bug reference

None


tests/includes/wait-for.sh
```diff
@@ -12,8 +12,8 @@ wait_for() {
     name=${1}
     query=${2}
 
-    # shellcheck disable=SC2046,SC2143
-    until [ $(juju status --format=json 2> /dev/null | jq "${query}" | grep "${name}") ]; do
+    # shellcheck disable=SC2143
+    until [ "$(juju status --format=json 2> /dev/null | jq "${query}" | grep "${name}")" ]; do
```
babbageclunk: I needed to add these quotes because the query I was passing to `wait_for` returned a value with spaces in it, which caused `[` to freak out. (Took me ages to work out - I'm not super-experienced with debugging shell.)

SimonRichardson: Ah, yes, good shout.

tests/suites/appdata/appdata.sh
```diff
@@ -0,0 +1,54 @@
...
+
+run_appdata_basic() {
+    echo
+
+    file="${TEST_DIR}/appdata-basic.txt"
+
+    ensure "appdata-basic" "${file}"
+
+    juju deploy ./acceptancetests/repository/charms/appdata-source
+    juju deploy -n 2 ./acceptancetests/repository/charms/appdata-sink
```
SimonRichardson: Question, should we link to the acceptance tests here, previously I've just re-included them next to the test. Then each test manages what they deploy, rather than a global repo of charms/bundles. The advantage is that it's easier to see whom owns what etc?

tests/suites/appdata/appdata.sh
```diff
@@ -0,0 +1,54 @@
...
+    output=$(juju ssh appdata-sink/0 cat /var/run/appdata-sink/token)
+    check_contains "$output" "appdata-source/0 test-value"
+
+    output=$(juju ssh appdata-sink/1 cat /var/run/appdata-sink/token)
+    check_contains "$output" "appdata-source/0 test-value"
+
+    destroy_model "appdata-basic"
+}
+
+unit_status() {
```
SimonRichardson: Can we move this into the https://github.com/juju/juju/blob/develop/tests/includes/wait-for.sh

tests/suites/appdata/task.sh
```diff
@@ -0,0 +1,21 @@
...
+    echo "==> Checking for dependencies"
+    check_dependencies juju
+
+    file="${TEST_DIR}/test-appdata.txt"
+
+    bootstrap "test-appdata" "${file}"
+
+    test_appdata_int
+
+    destroy_model "test-appdata"
```
SimonRichardson: Officially you shouldn't need this here.

babbageclunk: I added it because when running it with `BOOTSTRAP_REUSE_LOCAL` it would leave the test-appdata model around so the next run it would fail. That model isn't actually needed since `test_appdata_int` creates its own - is there a way to avoid creating test-appdata? Maybe I should add one?

tests/suites/appdata/appdata.sh
babbageclunk: Yeah, that's a good call - in that case there's no old test that needs them, I'll just move them here.

tests/suites/appdata/appdata.sh
babbageclunk: Yup yup

babbageclunk:
> Looks great, although I think the learning curve is somewhat different and maybe steeper at the beginning - it clearly shows your intent on what you're aiming to test. It makes it very easy for me to understand what you're testing.

Yeah, the initial setup was a bit fiddly (although I don't think worse than for the old CI tests) but the actual test feels *much* clearer.

babbageclunk:
I had a bit of trouble with `idle_condition` - can you explain how that query is supposed to work? Debugging it I found that the results coming back from 
```.applications | select(.["appdata-source"] | .units | .["appdata-source/1"] | .["juju-status"] | .current == "idle")```
...included both appdata-sink and appdata-source in arbitrary order, so I couldn't know which of 0 or 1 to pass in as app_index in the test. I don't really get why it was including both apps - does the application name filtering need to be moved up out of the select?

babbageclunk:
$$merge$$

SimonRichardson:
@babbageclunk so the issue is that jq doesn't sort the object keys by default (performance reasons), but they do provide a flag for it. https://github.com/juju/juju/pull/10862/commits/791653e9afc706df7f65190529ea85f585fe7274

PR #10862, opened 2019-11-05
Run subtests

SimonRichardson:
## Description of change

Every person seems to hit the same issue, in that they attempt to run
a subtest directly. This wasn't originally the main idea, instead they
where meant to run the test that runs the subtests and then see if they
broke anything whilst doing that.

I've been persuaded otherwise, a subtest should be able be run as you
want to iterate hard on that and instead of blocking people from doing
fast iterations we should enable it.

## QA steps

```sh
./main.sh deploy run_deploy_bundle
```

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -47,14 +47,14 @@ echo "failed" | grep -q "passes"   # fails
 To get started, it's best to quickly look at the help command from the runner.
 
 ```sh
-./main.sh -h
+cd tests && ./main.sh -h
 ```
 
 Running a full sweep of the integration tests (which will take a long time), can
 be done by running:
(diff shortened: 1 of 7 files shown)


tests/README.md
```diff
@@ -2,6 +2,46 @@
...
+run_deploy_bundles() {             # Subtest
+
+}
+
+test_deploy_bundles() {            # Test
+    run "run_deploy_bundles"       # Run subtest
+}
+```
+
+## Exit codes / Success
```
hmlanigan: 👍

SimonRichardson:
$$merge$$

PR #10876, opened 2019-11-07
Fixes lp#1844440: update-cloud supports public clouds

anastasiamac:
## Description of change

'update-public-clouds' command is being deprecated in this PR in favor of 'update-cloud' command. 'update-cloud' now supports updating public clouds too where for the client, we'd query single point of truth (in simplestreams) and for the controller, we'd still use local copy of a  public copy.

The advantages of this approach is that we are now collapsing 2-steps required to update public cloud on a controller into 1 -  'update-cloud aws --client --controller mycontroller' 

As a drive-by, the functionality shared between 'update-public-clouds' and now 'update-cloud' has been moved to ensure that the file with update-public-cloud command can be easily removed when the time comes. 

## QA steps

1. Public cloud definition has changed:
```
$ juju update-cloud aws --client
Fetching latest public cloud list...
Updated public cloud "aws" on this client, 5 cloud regions added:

    added cloud region:
        - aws/ap-northeast-1
        - aws/ap-northeast-2
        - aws/ap-northeast-3
        - aws/me-south-1
        - aws/sa-east-1
```
2. Public cloud definition is the same (no-op):
```
$ juju update-cloud aws --client
Fetching latest public cloud list...
No new details for cloud "aws" have been published.
```

## Documentation changes

'update-public-clouds' command is deprecated in favor of 'update-cloud'

## Bug reference

https://bugs.launchpad.net/juju/+bug/1844440

### DIFF
--- a/cmd/juju/cloud/cloudsdiff.go
+++ b/cmd/juju/cloud/cloudsdiff.go
@@ -1,0 +1,180 @@
+// Copyright 2016 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package cloud
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/juju/collections/set"
+	jujucloud "github.com/juju/juju/cloud"
+)
+
+func diffCloudDetails(cloudName string, new, old jujucloud.Cloud, diff *changes) {
+	sameAuthTypes := func() bool {
+		if len(old.AuthTypes) != len(new.AuthTypes) {
+			return false
+		}
+		newAuthTypes := set.NewStrings()
+		for _, one := range new.AuthTypes {
+			newAuthTypes.Add(string(one))
+		}
+
+		for _, anOldOne := range old.AuthTypes {
+			if !newAuthTypes.Contains(string(anOldOne)) {
+				return false
+			}
+		}
+		return true
+	}
+
+	endpointChanged := new.Endpoint != old.Endpoint
+	identityEndpointChanged := new.IdentityEndpoint != old.IdentityEndpoint
+	storageEndpointChanged := new.StorageEndpoint != old.StorageEndpoint
+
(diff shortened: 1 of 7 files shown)


anastasiamac:
[deprecate update-public-clouds in favor of update-cloud](https://trello.com/c/xCdbyruU/969-deprecate-update-public-clouds-in-favor-of-update-cloud)

mitechie:
Thanks, it's good to see things simplify into a fewer number of steps for folks. 

This seems very verbose on the user's end currently. The need to specify --client or --controller (or both?) and the need to specify the clouds involved seems a bit much for the user to get right. 

some proposals that might be more purpose built

    juju update-cloud --sync-public-clouds
    juju update-cloud --all
    juju update-cloud --all-public

I also wonder if it's worth requiring the arguments for client/controller. In the spirit of ask/tell would it be worth supporting it without any client/controller but just noting what would be updated and asking if that's ok?

    juju update-cloud --all
    New client aws regions:
        - aws/ap-northeast-1
        - aws/ap-northeast-2
        - aws/ap-northeast-3
        - aws/me-south-1
        - aws/sa-east-1
    New controller aws regions:
        - aws/ap-northeast-3
        - aws/me-south-1
        - aws/sa-east-1
    Update (Y/n)

with a --yes or the like for the scripted updates? What do you think?

anastasiamac:
@wallyworld and I had another discussion around the area. We have actually decided to close this PR and have a different approach:

* 'update-public-clouds' command is currently aliased to 'update-clouds'. This alias is confusing as it it very close to a completely different 'update-cloud' command. Remove it;
* add ask-or-tell functionality to 'update-public-clouds' command to enable update on both client (current behavior) and a controller. This will still removes the need for the existing, 2-step update process on a controller;
* do not change 'update-cloud' command to support fetch from simplestreams. This should be a sole responsibility of 'update-public-clouds' command...

This approach is subject to @howbazaar 's stamp. Your opinion, @mitechie is of interest as well :D

PR #10884, opened 2019-11-08
Remove the params.Life type and use life.Value from core/life package.

howbazaar:
Some clean up that needed to be done.

We don't need an additional Life type and constants in the apiserver/params package.

Tests pass, and there is no user impact at all.

### DIFF
--- a/apiserver/params/multiwatcher.go
+++ b/apiserver/params/multiwatcher.go
@@ -14,18 +14,9 @@ import (
 
 	"github.com/juju/juju/core/constraints"
 	"github.com/juju/juju/core/instance"
+	"github.com/juju/juju/core/life"
 	"github.com/juju/juju/core/model"
 	"github.com/juju/juju/core/status"
-)
-
-// Life describes the lifecycle state of an entity ("alive", "dying"
-// or "dead").
-type Life string
-
-const (
-	Alive Life = "alive"
-	Dying Life = "dying"
-	Dead  Life = "dead"
 )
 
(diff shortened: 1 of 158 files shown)


apiserver/params/multiwatcher.go
```diff
@@ -14,20 +14,11 @@ import (
 
 	"github.com/juju/juju/core/constraints"
 	"github.com/juju/juju/core/instance"
+	"github.com/juju/juju/core/life"
 	"github.com/juju/juju/core/model"
 	"github.com/juju/juju/core/status"
 )
 
-// Life describes the lifecycle state of an entity ("alive", "dying"
```
wallyworld: strictly speaking, each layer should define its own consts / data structures, and as info flows between layers, the data is transformed to the relevant representation for that layer.
+1 to the change to use core.Life in the api layer and its callers
- 1/2 to this bit since the purely correct approach is to map state.Life <-> params.Life in the apiserver layer and params.Life <-> core.Life in the api layer

howbazaar: As discussed, we argree on having params types for structures, but not simple scalar types.

howbazaar:
@SimonRichardson schema updated.

SimonRichardson:
@howbazaar done :dancer:

howbazaar:
$$merge$$

howbazaar:
$$merge$$ fatal: unable to access 'https://go.googlesource.com/oauth2/'

PR #10888, opened 2019-11-08
make test less error prone on log changes

nammn:
## Description of change

making sure that small log output changes don't destroy the test itself.
Those logs are caught by the unit test anyway. Checking for the correct sha is less error-prone and should lead to the same result.

## QA steps

```
 ./main.sh -l <local_controller> cli
```


nammn:
$$merge$$

PR #10892, opened 2019-11-11
Fixes to lp#1851852 and lp#1851849 - usability: pipes with prompts.

anastasiamac:
## Description of change

add-k8s command is frequently piped. However, the behavior of the command has changed to prompt for user input when neither --client nor --controller option was specified to support an ASK-OR-TELL brief.

Since the pipe and commands that require user input don't work together, if the pipe was used, the command to fail immediately with a descriptive message. 

As a drive-by, rename remaining reference to an ephemeral --client-only option.

## QA steps

1. $ microk8s.config | juju add-k8s k8s
```
This operation can be applied to both a copy on this client and to the one on a controller.
ERROR The command is piped and Juju cannot prompt to clarify whether the --client or a --controller is to be used.
Please clarify by re-running the command with the desired option(s).
```
2. $ juju add-credential aws (to ensure everything else works)
```
This operation can be applied to both a copy on this client and to the one on a controller.
No current controller was detected and there are no registered controllers on this client: either bootstrap one or register one.
Do you ONLY want to add a credential to this client? (Y/n):
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1851849
https://bugs.launchpad.net/juju/+bug/1851852

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -449,11 +449,6 @@ func (c *AddCAASCommand) Run(ctx *cmd.Context) (err error) {
 	// We need to do this later then other commands since
 	// piping is done regularly with add-k8s.
 	if err := c.MaybePrompt(ctx, fmt.Sprintf("add k8s cloud %v to", c.caasName)); err != nil {
-		if errors.Cause(err) == io.EOF {
-			ctx.Infof("The command is piped and Juju cannot prompt to clarify whether the --client or a --controller is to be used.\n" +
-				"Please clarify by re-running the command with the desired option(s).")
-			return cmd.ErrSilent
-		}
 		return errors.Trace(err)
 	}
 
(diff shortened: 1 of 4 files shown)


wallyworld:
Can we target to 2.7 to include in rc3

anastasiamac:
@wallyworld,

Yes, we can but I am going to change this to detect if pipe has been used to make it more applicable globally too.

anastasiamac:
@wallyworld, 
I'll backport to 2.7 once this lands... You cannot simple change the dop-down box to target 2.7 since there are considerable differences between develop and 2.7 :) and I have used develop as a base since 2.7 is meant to be frozen for code changes and is only open to Critical fixes or tests.

anastasiamac:
$$merge$$

anastasiamac:
[deal with --client-only references](https://trello.com/c/31D1l5G0/976-deal-with-client-only-references)

anastasiamac:
[detect when pipe is used and err instead of prompt](https://trello.com/c/JpiJyvw2/977-detect-when-pipe-is-used-and-err-instead-of-prompt)

PR #10894, opened 2019-11-11
fix asses_cloud_display

nammn:
## Description of change

didnt work for some time because of multiple reasons:
- we added some yaml output, which should not be parsed but still got parsed
https://github.com/juju/juju/blob/0e8fea3415fcef38e8cb2fe8b97a6d581ef26bd8/cmd/juju/cloud/show.go#L159

- sometimes regions cannot be parsed

- we have new "cloud" types

## Bug reference

https://jenkins.juju.canonical.com/view/acceptance-tests/job/nw-clouds-display/1962/console

### DIFF
--- a/acceptancetests/assess_cloud_display.py
+++ b/acceptancetests/assess_cloud_display.py
@@ -2,12 +2,13 @@
 
 from __future__ import print_function
 
+import os
 from argparse import ArgumentParser
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


nammn:
$$merge$$

nammn:
$$merge$$

nammn:
$$merge$$

PR #10895, opened 2019-11-11
Replace bash double-bracket syntax with single brackets in space tests

achilleasa:
## Description of change

The test-runner assumes a generic sh binary; this PR replaces double brackets in if blocks with single brackets to avoid errors when running on CI


achilleasa:
$$merge$$

PR #10898, opened 2019-11-12
Feature/podpec validation

ycliuhw:
## Description of change

This PR introduces top-level validation for both legacy and v2 pod spec.

## QA steps

- deploy a CaaS charm with `unknown` fields in pod spec - now Juju will complain about the field;

## Documentation changes

No

## Bug reference

https://bugs.launchpad.net/juju/+bug/1851571

### DIFF
--- a/caas/kubernetes/provider/specs/decode.go
+++ b/caas/kubernetes/provider/specs/decode.go
@@ -1,0 +1,89 @@
+// Copyright 2019 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package specs
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"strings"
+
+	"github.com/juju/errors"
+	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
+)
+
+// YAMLOrJSONDecoder attempts to decode a stream of JSON documents or
+// YAML documents by sniffing for a leading { character.
+type YAMLOrJSONDecoder struct {
+	bufferSize int
+	r          io.Reader
+	rawData    []byte
+
+	strict bool
+}
+
+func newStrictYAMLOrJSONDecoder(r io.Reader, bufferSize int) *YAMLOrJSONDecoder {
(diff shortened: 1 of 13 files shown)


caas/specs/legacy.go
```diff
@@ -11,24 +11,12 @@ import (
 // a pod on the CAAS substrate.
 type PodSpecLegacy struct {
 	podSpecBase `yaml:",inline"`
-
-	// legacy version has containers/initContainers two blocks.
-	InitContainers []ContainerSpec `json:"initContainers" yaml:"initContainers"`
```
ycliuhw: `InitContainers` and `Containers` are moved to the provider level. 
They are combined and filled into `Containers` with `Init bool` properly set.

ycliuhw:
[pod spec yaml validation](https://trello.com/c/tlOmmfp1/975-pod-spec-yaml-validation)

ycliuhw:
!!build!!
failed to fetch source for https://go.googlesource.com/oauth2: unable to get repository

caas/kubernetes/provider/specs/decode.go
```diff
@@ -0,0 +1,89 @@
...
+	"io/ioutil"
+	"strings"
+
+	"github.com/juju/errors"
+	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
+)
+
+// YAMLOrJSONDecoder attempts to decode a stream of JSON documents or
+// YAML documents by sniffing for a leading { character.
+type YAMLOrJSONDecoder struct {
```
wallyworld: can't we use the one from the k8s sdk?

ycliuhw: The reason why we have this new decoder is that they don't expose a way to enable strict mode for validation.
You can see it's still using the converter from the SDK to convert YAML to JSON then use golang JSON decoder to unmarshal the JSON string.
One thing we do here is to set `DisallowUnknownFields` before decoding.

ycliuhw:
!!build!!
failed to fetch source for https://go.googlesource.com/oauth2: unable to get repository

ycliuhw:
$$merge$$

ycliuhw:
failed to fetch source for https://go.googlesource.com/oauth2: unable to get repository
$$merge$$

PR #10906, opened 2019-11-14
Retry all run commands

SimonRichardson:
This is a backport of #10832

## Description of change

The following attempts to make installing dependencies for building
snaps more reliable. Essentially we're going to try and retry all
run commands so at least we've not bailed out on the first try.

This should hopefully improve the current ongoing failures we have
when building snaps.


SimonRichardson:
$$merge$$

PR #10908, opened 2019-11-14
actions v2 behind feature flag again; fix yaml output with backwards compatibility

wallyworld:
## Description of change

New action/function commands back behind feature flag.

When serving the action result data map, the new v5 facade will now use lower case yaml attributes. The v4 and earlier facade continues to use the CamelCase versions. To make this change, several pieces were needed:
- server side conversion of data map from state to new format
- on client side, when older action commands are being used, convert back to old format

Fix the rendering of timestamps in task yaml and list tasks. For list tasks, move time column to end. In yaml, ensure the timezone offset is always included.

## QA steps

start a 2.6 controller
add a model
run an action and use juju run
ensure show-action-output works with 2.6 and 2.7 client
upgrade the controller to 2.7
use list tasks to see the ids from the previous runs
use 2.6.9 client to run show-action-output
use 2.7 client to run show task and show-action-output for all task ids

### DIFF
--- a/cmd/juju/commands/exec_test.go
+++ b/cmd/juju/commands/exec_test.go
@@ -39,7 +39,7 @@ func (s *ExecSuite) SetUpTest(c *gc.C) {
 }
 
 func newTestExecCommand(clock clock.Clock, modelType model.ModelType) cmd.Command {
-	return newExecCommand(minimalStore(modelType), clock.After)
+	return newExecCommand(minimalStore(modelType), clock.After, false)
 }
 
 func minimalStore(modelType model.ModelType) *jujuclient.MemStore {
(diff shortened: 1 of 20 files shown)


cmd/juju/commands/exec_test.go
```diff
@@ -298,15 +297,15 @@ func (s *ExecSuite) TestConvertRunResults(c *gc.C) {
...
 		expected: map[string]interface{}{
-			"UnitId":     "unit/0",
-			"Stdout":     "stdout",
-			"Stderr":     "stderr",
-			"Message":    "msg",
-			"ReturnCode": 42,
+			"unit":        "unit/0",
+			"stdout":      "stdout",
+			"stderr":      "stderr",
+			"Message":     "msg",
```
howbazaar: "Message" should be "message" as a key.

wallyworld:
$$merge$$

PR #10909, opened 2019-11-14
Tell update-public-clouds to update for client.

hmlanigan:
## Description of change

Tell update-public-clouds to update for client.

Resolve issue with test failing because question in prompt could not
be answered when command used with a pipe.


hmlanigan:
$$merge$$

PR #10911, opened 2019-11-15
Rename tasks to operations

wallyworld:
## Description of change

The result of running a function is now an operation rather than a task. 

## QA steps

juju call somefunction
juju list-tasks
juju show-task


wallyworld:
$$merge$$

PR #10913, opened 2019-11-15
Ensure juju can bootstrap microk8s as root user

wallyworld:
## Description of change

Adjust the initial check to ensure that juju can bootstrap microk8s as the root user.

## QA steps

```sh
su -i root
juju bootstrap microk8s
```


wallyworld:
$$merge$$

PR #10915, opened 2019-11-15
2.7 logging read sink 1852815

jameinel:
## Description of change

Bring in bugfix from 2.6.

It is possible to get an error that wasn't a close error, and wasn't an
unexpected close, and we weren't logging those errors.
Instead, log expected close messages at Trace, but all other errors get
logged at debug.

## QA steps

We are only seeing these errors in production, and don't really have a way to trigger them directly yet.

## Documentation changes

None

## Bug reference

https://bugs.launchpad.net/juju/2.6/+bug/1852815


jameinel:
$$merge$$

PR #10918, opened 2019-11-18
acceptancetest fix:  cloud show

nammn:
### Checklist

----

## Description of change
accceptancetest interactive: add newline for better formatting
acceptancetest many: fix show-cloud by using "" instead of None

## QA steps

```sh
/home/nam/golang/src/github.com/juju/juju/acceptancetests/assess_cloud_display.py ~/example-clouds.yaml /home/nam/golang/bin/juju
assess_clouds (no_clouds): PASS
assess_clouds: PASS
assess_show_cloud: PASS
```

### DIFF
--- a/acceptancetests/assess_cloud_display.py
+++ b/acceptancetests/assess_cloud_display.py
@@ -37,8 +37,9 @@ def remove_display_attributes(cloud):
     # bug #1645783.
     defined = cloud.pop('defined')
     assert_equal(defined, 'local')
-    description = cloud.pop('description', None)
-    # Pop Nones, which are "errors from parsing the yaml"
+    description = cloud.pop('description', "")
+
+    # Delete None values, which are "errors" from parsing the yaml
     # E.g. output can show values which we show to the customers but should actually not parsed and compared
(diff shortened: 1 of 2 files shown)


acceptancetests/assess_cloud_display.py
```diff
@@ -37,7 +37,7 @@ def remove_display_attributes(cloud):
     # bug #1645783.
     defined = cloud.pop('defined')
     assert_equal(defined, 'local')
-    description = cloud.pop('description', None)
+    description = cloud.pop('description', "")
```
SimonRichardson: `# Pop Nones, which are "errors from parsing the yaml"`

From the comment below, i think the change is wrong OR we change the comment.

nammn: The comment is related to the code below not above. We pop them because somehow the code accesses them to check them later. 
As they are using direct access this will eventually lead to an `exception` thus we pop them below.
But pop is the wrong wording. It should be delete :\

nammn:
$$merge$$

PR #10922, opened 2019-11-19
Fix test failure due to RSA PSS signatures.

hpidcock:
## Description of change

Testing RSA private keys are too small to encode a PSS signature.
Only affects test runs on newer versions of Go.

## QA steps

Run tests.

## Documentation changes

N/A

## Bug reference

N/A


hpidcock:
$$merge$$

PR #10926, opened 2019-11-19
Retry the mongo config

SimonRichardson:
### Checklist

 - [ ] ~~Checked if it requires a [pylibjuju](https://github.com/juju/python-libjuju) change?~~

----

## Description of change

The following attempts to get the mongo memory profile to check that
it has been set in the config. It would seem that sometimes the
mongo service is slow and so it requires that to change, so that it
can then be correctly reflected in the config.

## QA steps

```sh
cd tests && ./main.sh controller
```


hmlanigan:
$$merge$$

PR #10931, opened 2019-11-21
All watcher worker - not yet used

howbazaar:
Create a real worker for managing the multiwatchers. Future branches will remove the workers that are running in the state objects.

The all watcher code that the multiwatcher worker uses is based on the one that watches all the models. Watching an individual model is now a simple filter on the event stream for the entire controller.

The multiwatcher worker is added to the machine manifolds, but neither the apiserver nor the modelcache worker are yet updated to use it. It has been added to confirm that it is running and the metrics are being exported.
 
## QA steps

```sh
juju bootstrap lxd multiwatcher
juju ssh -m controller 0

juju_engine_report | grep -A 8 multiwatcher
Querying @jujud-machine-0 introspection socket: /depengine
  multiwatcher:
    inputs:
    - state
    - is-controller-flag
    - upgrade-database-flag
    report:
      num-watchers: 0
      restart-count: 0
      store-size: 3

juju_metrics | grep multiwatcher
Querying @jujud-machine-0 introspection socket: /metrics/
# HELP juju_multiwatcher_restart_count The number of times the all watcher has been restarted.
# TYPE juju_multiwatcher_restart_count gauge
juju_multiwatcher_restart_count 0
# HELP juju_multiwatcher_store_size The number of entities in the store.
# TYPE juju_multiwatcher_store_size gauge
juju_multiwatcher_store_size 3
# HELP juju_multiwatcher_watcher_count The number of multiwatcher type watchers there are.
# TYPE juju_multiwatcher_watcher_count gauge
juju_multiwatcher_watcher_count 0
```

## Documentation changes

This is an internal only change.

### DIFF
--- a/worker/multiwatcher/watcher.go
+++ b/worker/multiwatcher/watcher.go
@@ -1,0 +1,131 @@
+// Copyright 2019 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package multiwatcher
+
+import (
+	"github.com/juju/errors"
+
+	"github.com/juju/juju/core/multiwatcher"
+)
+
+type control interface {
+	Dying() <-chan struct{}
+	Err() error
+}
+
+// Watcher watches any changes to the state.
+type Watcher struct {
+	request chan *request
+	control control
+	logger  Logger
+	err     chan error
+
+	filter func([]multiwatcher.Delta) []multiwatcher.Delta
+
+	// The following fields are maintained by the Worker goroutine.
+	revno   int64
+	stopped bool
+
+	// used indicates that the watcher was used (i.e. Next() called).
+	used bool
+}
+
+// Stop stops the watcher.
+func (w *Watcher) Stop() error {
+	select {
+	case w.request <- &request{watcher: w}:
+		// We asked to be stopped, and this is processed.
+	case _, stillOpen := <-w.err:
+		// We have been stopped already.
+		if stillOpen {
+			close(w.err)
+		}
+		// Since we are asking to stop, we don't care about the underlying
+		// error from the read.
+	case <-w.control.Dying():
+		// The worker is dying and we will be cleaned up.
+	}
+	// In all of the above cases, we're fine, and don't need to return an
+	// error that would be logged.
+	return nil
+}
+
+// Next retrieves all changes that have happened since the last
+// time it was called, blocking until there are some changes available.
+//
+// The result from the initial call to Next() is different from
+// subsequent calls. The latter will reflect changes that have happened
+// since the last Next() call. In contrast, the initial Next() call will
+// return the deltas that represent the model's complete state at that
+// moment, even when the model is empty. In that empty model case an
+// empty set of deltas is returned.
+func (w *Watcher) Next() ([]multiwatcher.Delta, error) {
+	// In order to be able to apply the filter, yet not signal the caller when
+	// all deltas were filtered out, we need an outer loop.
+	var changes []multiwatcher.Delta
+	for len(changes) == 0 {
+		req := &request{
+			watcher: w,
+			reply:   make(chan bool),
+		}
+		if !w.used {
+			req.noChanges = make(chan struct{})
+			w.used = true
+		}
+
+		select {
+		case err, open := <-w.err:
+			if open {
+				w.logger.Tracef("hit an error: %v", err)
+				close(w.err)
+				return nil, errors.Trace(err)
+			}
+			w.logger.Tracef("err channel closed")
+			return nil, errors.Trace(multiwatcher.NewErrStopped())
+		case <-w.control.Dying():
+			w.logger.Tracef("worker dying")
+			err := w.control.Err()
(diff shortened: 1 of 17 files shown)


worker/multiwatcher/watcher.go
```diff
@@ -0,0 +1,131 @@
...
+func (w *Watcher) Next() ([]multiwatcher.Delta, error) {
+	// In order to be able to apply the filter, yet not signal the caller when
+	// all deltas were filtered out, we need an outer loop.
+	var changes []multiwatcher.Delta
+	for len(changes) == 0 {
+		req := &request{
+			watcher: w,
+			reply:   make(chan bool),
+		}
+		if !w.used {
```
hpidcock: Probably a silly question, but does Next() get called from multiple goroutines?

worker/multiwatcher/worker.go
```diff
@@ -0,0 +1,352 @@
...
+	// nil, the Multiwatcher will be stopped.  If the reply is true,
+	// the request has been processed; if false, the Multiwatcher
+	// has been stopped,
+	reply chan bool
+
+	// noChanges receives a message when the manager checks for changes
+	// and there are none.
+	noChanges chan struct{}
+
+	// On reply, changes will hold changes that have occurred since
```
hpidcock: reform comment to start with "changes"

worker/multiwatcher/worker.go
```diff
@@ -0,0 +1,352 @@
...
+	for {
+		err := w.inner()
+		select {
+		case <-w.tomb.Dying():
+			return nil
+		default:
+			w.mu.Lock()
+			w.restartCount++
+			w.errors = append(w.errors, err)
+			if len(w.errors) > 5 {
```
hpidcock: Magic number 5?

worker/multiwatcher/worker.go
```diff
@@ -0,0 +1,352 @@
...
+			return
+		}
+
+		w.removeWaitingReq(watcher, req)
+		w.store.AddReference(revno)
+	}
+}
+
+func (w *Worker) removeWaitingReq(watcher *Watcher, req *request) {
+	if req := req.next; req == nil {
```
hpidcock: can we not shadow req please, this looks like a singly-linked list, so better naming would be helpful.

worker/multiwatcher/worker_internal_test.go
```diff
@@ -0,0 +1,361 @@
...
+	assertReplied(c, true, req0)
+	assertReplied(c, true, req1)
+	assertWaitingRequests(c, sm, nil)
+
+	deltas := []multiwatcher.Delta{{Entity: &multiwatcher.MachineInfo{ID: "1"}}}
+	c.Assert(req0.changes, gc.DeepEquals, deltas)
+	c.Assert(req1.changes, gc.DeepEquals, deltas)
+}
+
+var respondTestChanges = [...]func(store multiwatcher.Store){
```
hpidcock: Would be good to test multi models. There is filtering logic for models, so unless it's tested elsewhere and I'm blind, would be good to test it  somewhere here.

worker/multiwatcher/watcher.go
howbazaar: The Next() calls come from the apiserver generated goroutines. So very independent from the goroutines of the worker responsible for the underlying changes.

worker/multiwatcher/worker.go
howbazaar: I had figured since it was a internal member, and lint didn't complain it wasn't a big deal. Will update though.

worker/multiwatcher/worker.go
howbazaar: Yeah. I didn't want to remember too many, but it wanted more than just the last one. Five seemed like a reasonable number. This bit is new. We could define a constant, but it is only used in one place. Perhaps a comment would be enough to explain?

worker/multiwatcher/worker.go
howbazaar: Done.

worker/multiwatcher/worker_internal_test.go
howbazaar: Will look and add if missing.

howbazaar:
Thanks @hpidcock for the review.

I should have mentioned that all this code was lifted from the state/multiwatcher.go file and turned into a real worker. In order to minimise the risk of the changes, no attempts were made to restructure or optimise the code. The linked-list structure was very much in use in the original code. We may want to look at it later, but since this bit isn't broken, I didn't look to fix it.

worker/multiwatcher/worker_internal_test.go
howbazaar: watcher_test.go has tests for the multiple uuids.

worker/multiwatcher/worker.go
hpidcock: Quick comment would be fine I think.

howbazaar:
$$merge$$

PR #10933, opened 2019-11-21
Allow the httpserver dualport to close early.

jameinel:
## Description of change

This is a small cleanup for the http dualPort code. If the http server was shutting down, the old code wouldn't notice until timeout was reached. This didn't actually block anything (as far as I can tell) because the dualListener was opened from a Hub message, which didn't block the other worker from shutting down. So all this really does is cleanup a goroutine a bit earlier.

Still, since I noticed it, I figured it was worth landing. The test doesn't actually fail if you don't exit early, because we can't wait for the openAPIPort to return, because it is triggered from the Hub subscription. Though I think the setup of the test is still relevant (you shouldn't fail if requested to shut down before the timeout occurs.)

## QA steps

New Unit test. You can see how it is triggered if you do 'go test -check.vv' where you can see that it notices right away that shutdown was requested.

## Documentation changes

None.

## Bug reference

None.

### DIFF
--- a/worker/httpserver/worker.go
+++ b/worker/httpserver/worker.go
@@ -484,7 +484,14 @@ func (d *dualListener) openAPIPort(topic string, conn apiserver.APIConnection, e
 		d.status = "waiting prior to opening agent port"
 		d.mu.Unlock()
 		logger.Infof("waiting for %s before allowing api connections", d.delay)
-		<-d.clock.After(d.delay)
+		select {
+		case <-d.done:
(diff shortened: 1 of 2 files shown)


jameinel:
I'm targeting 2.7, but I want to wait to land this until 2.7.0 is final.

howbazaar:
$$merge$$

jameinel:
$$merge$$

jameinel:
$$merge$$

PR #10955, opened 2019-11-26
Improve sorting for juju/cmd#68 changes

SimonRichardson:
### Checklist

----

## Description of change

The following brings in the changes from juju/cmd#68 PR changes.

------

The following removes the math.MaxInt64, as we have failure to build on
32-bit architectures. Instead the sorting algorithm now performs one
sort over the indexed collections, first by checking the distance
between two strings and then if they're the same distance, falling back
onto the name.

This should improve the performance as you're only looping over once
instead of twice as previous algorithm did, which is a nice bonus!

## QA steps

```sh
➜ juju controll
ERROR juju: "controll" is not a juju command. See "juju --help".

Did you mean:
	controllers
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -472,14 +472,14 @@
   revision = "9c5c9712527c7986f012361e7d13756b4d99543d"
 
 [[projects]]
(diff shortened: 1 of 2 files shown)


SimonRichardson:
$$merge$$

PR #10957, opened 2019-11-27
Relax credential validity check for update.

anastasiamac:
### Checklist

_No API change._

 - [ ] ~~Added [integration tests](https://github.com/juju/juju/tree/develop/tests) for the PR?~~
 - [ ] ~~Added or updated [doc.go](https://discourse.jujucharms.com/t/readme-in-packages/451) related to packages changed?~~

----

## Description of change

Cloud credential check is performed when managing credentials as well as when migrating a model. Juju checks that the cloud instances for the machines are accessible as well as that all cloud instances for a model have associated machines.

For model migration, the second check is a sanity check that verifies that machines are migrated and cloud instances are re-tagged completely and correctly. However, for credential management, this check is not necessary (and can cause problems) since Juju only needs to make sure that machines' cloud instances are accessible.
 
This PR simplifies validity check for credential management by by-passing 2nd check and ensuring that migration checks remain the same.

## QA steps

1. bootstrap
2. deploy application with some units
3. remove a unit such that machine's cloud instance remains (either using harvest mode or db surgery)
4. invalidate cloud credential (first in the cloud console, then try deploying another application or add a unit or a machine)
5. re-validate credential ('juju update-credential...')

(or follow the MAAS locked machine example in linked bug)

## Bug reference

Part 1 for https://bugs.launchpad.net/juju/+bug/1852412

### DIFF
--- a/apiserver/common/credentialcommon/modelcredential.go
+++ b/apiserver/common/credentialcommon/modelcredential.go
@@ -18,7 +18,7 @@ import (
 )
 
 // ValidateExistingModelCredential checks if the cloud credential that a given model uses is valid for it.
-func ValidateExistingModelCredential(backend PersistentBackend, callCtx context.ProviderCallContext) (params.ErrorResults, error) {
+func ValidateExistingModelCredential(backend PersistentBackend, callCtx context.ProviderCallContext, migrating bool) (params.ErrorResults, error) {
 	model, err := backend.Model()
 	if err != nil {
 		return params.ErrorResults{}, errors.Trace(err)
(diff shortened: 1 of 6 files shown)


anastasiamac:
[simplify credential validity check for cred mgmt](https://trello.com/c/fi93ninA/984-simplify-credential-validity-check-for-cred-mgmt)

anastasiamac:
$$merge$$

babbageclunk:
(I realise this is a bit after the horse has bolted though.)

anastasiamac:
@babbageclunk,
I did originally want to split the checks. However, that check is performed by kind of "in passing", while collecting the list of instance ids for the first check. It was easier and was a smaller difference surface to do as I did.

I am not really fussed about parameter name. Calling it "migrating" makes sense from a caller perspective.. I came from the api call and it was easier to distinguish "am I *migrating* Or not" at that level. It made sense when sending the parameters from the outside of the internal logic of the method.. However, am happy to change it in the future.

Thank you for the review :D Since I hit merger before receiving your comments, I might change the follow up...

PR #10966, opened 2019-11-28
Don't suppress errors in Start.

jameinel:
### Checklist

 - [No] Checked if it requires a [pylibjuju](https://github.com/juju/python-libjuju) change?
 - [N/A] Added [integration tests](https://github.com/juju/juju/tree/develop/tests) for the PR?
 - [N/A ] Do comments answer the question of why design decisions were made?

## Description of change

Don't suppress errors in Start.

When running a charm hook, we would notice if the hook exited nonzero,
but if we failed to Start the hook, that error ended up being
suppressed. This patch just puts the error in the same place as existing
exit error.

## QA steps

You can write a charm that has an invalid 'install' hook. Something like:
```sh
$ cd acceptancetests/repository/charms
$ echo "exit 2" > dummy-sink/hooks/install
$ juju deploy ./dummy-sink
```

Without this patch, it ignores the install hook failure, with this patch you get something like:
```
$ juju status
...
Unit           Workload  Agent  Machine  Public address  Ports  Message
dummy-sink/2*  error     idle   2        10.210.24.221          hook failed: "install"
```

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1854338

### DIFF
--- a/worker/uniter/runner/runner.go
+++ b/worker/uniter/runner/runner.go
@@ -515,6 +515,8 @@ func (runner *runner) runCharmHookOnLocal(hookName string, env []string, charmLo
 		runner.context.SetProcess(hookProcess{ps.Process})
 		// Block until execution finishes
 		exitErr = ps.Wait()
+	} else {
+		exitErr = err
 	}
 	// Ensure hook loggers are stopped before reading stdout/stderr
 	// so all the output is captured.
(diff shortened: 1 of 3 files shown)


jameinel:
$$merge$$

PR #10967, opened 2019-11-28
cmr saas migration: add initial step for finding the remote model (only works for model migration of same controller)

nammn:
### Checklist

----

### Edge Cases to work out
- [x] `consumer` and `saas` one same controller a but different models;' saas moved from a to b (same controller, different models)
- [ ]  `consumer` and `saas` one different controller (a, b) and different models; saas moved from b to c  (need to be another pr)
  - because the collection (migrated models) we access only saves the `migration` information from controller a and b as external controller. So we would need to watch b for changed in the migration, which this patch does not check for.

## Description of change

- small refactorings regarding the phases of a model migration for test usages.

- the actual feature: 

cross model relation does not support migration cases.
E.g. QA steps below are not working if we move the offering `SAAS` model from the consuming model.
After the migration, the model is still looking for the original model.
```
controller-0: 15:22:01 ERROR juju.worker.remoterelations error in remote application worker for mysql: external controller with model eaa39728-1adf-451c-85ec-a45686d74c43 not found
```

Luckily we already have a lot of things in place. Especially the collection which keeps track whether something was migrated or not. 
This PR uses them.

## QA steps

```sh
juju bootstrap dst
juju bootstrap src

juju add-model db
juju deploy mysql
juju offer mysql:db

juju add-model blog
juju deploy wikimedia
juju add-relation wikimedia:db mysql:db 
```
another terminal look at the debug-log

```
juju debug-log <- of mediawiki
```
```
juju migrate mysql dst
```

#### from (without this patch):
```
controller-0: 15:21:01 ERROR juju.worker.remoterelations error in remote application worker for mysql: external controller with model eaa39728-1adf-451c-85ec-a45686d74c43 not found
controller-0: 15:22:01 ERROR juju.worker.remoterelations error in remote application worker for mysql: external controller with model eaa39728-1adf-451c-85ec-a45686d74c43 not found
controller-0: 15:23:01 ERROR juju.worker.remoterelations error in remote application worker for mysql: external controller with model eaa39728-1adf-451c-85ec-a45686d74c43 not found
```

####  to the following outcome:
```
controller-0: 17:26:28 DEBUG juju.worker.remoterelations processing remote application changes for: [mysql]
controller-0: 17:26:28 DEBUG juju.worker.remoterelations starting watcher for remote application "mysql"
controller-0: 17:26:28 DEBUG juju.worker.remoterelations remote controller api addresses: [10.75.147.162:17070]
controller-0: 17:26:28 DEBUG juju.worker.remoterelations remote offer for mysql has been removed
```

## Depends on
https://github.com/juju/juju/pull/10960 to fully work

### DIFF
--- a/apiserver/common/controllerconfig.go
+++ b/apiserver/common/controllerconfig.go
@@ -8,6 +8,7 @@ import (
 	"gopkg.in/juju/names.v3"
 
 	"github.com/juju/juju/apiserver/params"
+	"github.com/juju/juju/core/crossmodel"
 	"github.com/juju/juju/state"
 )
 
@@ -76,9 +77,38 @@ func (s *controllerStateShim) ControllerInfo(modelUUID string) (addrs []string, 
 		return StateControllerInfo(s.State)
 	}
 
-	// Now check any external controllers.
 	ec := state.NewExternalControllers(s.State)
 	info, err := ec.ControllerForModel(modelUUID)
+
+	// the model on this controller may have been migrated, if that is the case we try to get the information.
+	// If that is true we try to save the migration controller as an external controller.
+	// This does not take cmr across controller into consideration controller a (consumer) --> b (offer).
+	// If a cross controller model has been migrated b (offer moving) --> c (new offer).
+	// There will be no migration information on this collection on controller a (consumer)
+	// about the migrated offer from b to c
+	if errors.IsNotFound(err) {
+		migration, err := s.State.CompletedMigration(modelUUID)
+		if err != nil {
+			return nil, "", errors.Trace(err)
+		}
+		target, err := migration.TargetInfo()
+		if err != nil {
+			return nil, "", errors.Trace(err)
+
+		}
+		logger.Debugf("found migrated model on another controller, saving the information")
+		_, err = ec.Save(crossmodel.ControllerInfo{
+			ControllerTag: target.ControllerTag,
+			Alias:         target.ControllerAlias,
+			Addrs:         target.Addrs,
+			CACert:        target.CACert,
+		}, modelUUID)
+		if err != nil {
+			return nil, "", errors.Trace(err)
+		}
+		return target.Addrs, target.CACert, nil
+	}
+
 	if err != nil {
 		return nil, "", errors.Trace(err)
 	}
--- a/core/migration/phase.go
+++ b/core/migration/phase.go
@@ -39,6 +39,19 @@ var phaseNames = []string{
 	"ABORTDONE",
 }
 
+// Those phases are only used to get a complete successful round for testing purposes.
+func SuccessfulMigrationPhases() []Phase {
+	return []Phase{
+		IMPORT,
+		PROCESSRELATIONS,
+		VALIDATION,
+		SUCCESS,
+		LOGTRANSFER,
+		REAP,
+		DONE,
+	}
+}
+
 // String returns the name of an model migration phase constant.
 func (p Phase) String() string {
 	i := int(p)
--- a/state/modelmigration.go
+++ b/state/modelmigration.go
@@ -814,10 +814,10 @@ func checkTargetController(st *State, targetControllerTag names.ControllerTag) e
 // LatestMigration returns the most recent ModelMigration (if any) for a model
 // that has not been removed from the state. Callers interested in
 // ModelMigrations for models that have been removed after a successful
-// migration to another controller should use LatestRemovedModelMigration
+// migration to another controller should use CompletedMigrationForModel
 // instead.
 func (st *State) LatestMigration() (ModelMigration, error) {
-	mig, phase, err := st.latestMigration()
+	mig, phase, err := st.latestMigration(st.ModelUUID())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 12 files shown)


apiserver/common/controllerconfig.go
```diff
@@ -5,6 +5,7 @@ package common
 
 import (
 	"github.com/juju/errors"
+	"github.com/juju/juju/core/crossmodel"
```
nammn: dont forgett to clean that up

apiserver/common/controllerconfig.go
```diff
@@ -76,15 +77,44 @@ func (s *controllerStateShim) ControllerInfo(modelUUID string) (addrs []string,
 		return StateControllerInfo(s.State)
 	}
 
-	// Now check any external controllers.
 	ec := state.NewExternalControllers(s.State)
 	info, err := ec.ControllerForModel(modelUUID)
+
+	// the model may have been migrated, if that is the case we try to get the information.
+	// If that is true we try to save the migration controller as an external controller.
+	if errors.IsNotFound(err) {
```
manadart: This doesn't look like it will work.

External controllers are stored by models consuming remote offers so that once they add the relation they can notify that model of consumption and watch the remote offer for changes.

If you are already trying to locate an _external controller_ for a model, it is unlikely the model was previously _local to this controller_.

nammn: I am not following 100%. 
Please correct me if my thoughts are wrong here:
- after the model migration this code cannot find the `modelUUID` in the current controller collection anymore  -> returns error.notFound
- Thats why I look into the `migrations` collection to search for the model to see whether it has been migrated.
- If that is the case save this controller into the `externalcontroller` collection

At least thats how I understood it and coded it into the corresponding unit test.

manadart: Aha. 

I needed to expand above this in the Github UI. I didn't see that looking for external controllers was already a fall-back for checking the local model.

We still have the problem of locating a migrated model _from_ an external controller, but we will have discussions about this next week.

nammn:
Ah,
Yes that makes a lot of sense, need to take those change into account as
well.
Let's discuss that next week!

apiserver/common/controllerconfig.go
```diff
@@ -76,15 +77,44 @@ func (s *controllerStateShim) ControllerInfo(modelUUID string) (addrs []string,
...
+		}, modelUUID)
+		if err != nil {
+			return handleControllerInfoErr(err)
+		}
+		return target.Addrs, target.CACert, nil
+	}
+
 	if err != nil {
-		return nil, "", errors.Trace(err)
+		return handleControllerInfoErr(err)
```
SimonRichardson: I don't know what this gives you?

nammn: for me, it is mostly usability with the ide. So if I need to return an err again I can just start the functionname and the IDE auto completed for me, instead of me typing those words. But I have no hard feelings about that. I can change it.

apiserver/common/controllerconfig.go
```diff
@@ -76,15 +77,44 @@ func (s *controllerStateShim) ControllerInfo(modelUUID string) (addrs []string,
...
-	// Now check any external controllers.
 	ec := state.NewExternalControllers(s.State)
 	info, err := ec.ControllerForModel(modelUUID)
+
+	// the model may have been migrated, if that is the case we try to get the information.
+	// If that is true we try to save the migration controller as an external controller.
+	if errors.IsNotFound(err) {
+		migration, err := s.State.LatestRemovedModelMigrationByUUID(modelUUID)
+		if err != nil {
+			return handleControllerInfoErr(err)
```
nammn: we might want to add information here. Something along the line of:
`The model might be migrated from an external controller to begin with, therefore we cannot know whether this model has been migrated or not in general...`

apiserver/common/controllerconfig.go
SimonRichardson: It means another function call I have to lookup, that just wraps the error. It inhibits readability for little gain imo.

nammn: Fair point. For me, either version works for me. I can change that with the other review points (when given)

SimonRichardson:
Kewl

state/modelmigration.go
```diff
@@ -856,12 +856,33 @@ func (st *State) LatestRemovedModelMigration() (ModelMigration, error) {
 	return mig, nil
 }
 
+// LatestRemovedModelMigrationByUUID returns the most recent ModelMigration (if any)
+// for a model that has been removed from the state after a successful
+// migration to another controller given by the modelUUID
+func (st *State) LatestRemovedModelMigrationByUUID(modelUUID string) (ModelMigration, error) {
```
manadart: I would rename the method above this to `CompleteMigration` to cut down on work-sauce. It should also call this one.

Then:
```go
// CompletedMigrationForModel returns the most recent migration for the 
// input model UUID if it reached the DONE phase and caused the model 
// to be relocated.
func (st *State) CompletedMigrationForModel(modelUUID string) (ModelMigration, error) {
```

core/migration/phase.go
```diff
@@ -39,6 +39,19 @@ var phaseNames = []string{
 	"ABORTDONE",
 }
 
+// Those phases are only used to get a complete successful round for testing purposes.
+func GetCompletePhaseMigrationRound() []Phase {
```
manadart: I don't like adding an exported function that is only used in tests, but I can't think of a suitable place to put it at this point.

In any case:
```go
// SuccessfulMigrationPhases returns the phases that a successful migration
// transitions through in correct order. It is used for testing.
func SuccessfulMigrationPhases() []Phase {
```

nammn: I agree with this! I did not have any other place in mind though, for me, refactoring was still a win.

nammn:
$$merge$$

PR #10969, opened 2019-11-29
Add JUJU_HOOK_NAME to hook environments

dshcherb:
There are currently indirect mechanisms for a charm to identify the hook
name (argv[0] or JUJU_CONTEXT_ID which does not have a publicly
documented format).

This change adds it to the hook context and environment variables.

JUJU_HOOK_NAME will be set to a function name for functions.

See LP: #1854505


mitechie:
Before we land this I'd like a chance to get the process change hook event
spec together and make sure this is in line with that work.

dshcherb:
@mitechie @manadart sounds good. I would be interested to collaborate on it.

mitechie:
Definitely, we'll look for sign off from folks working on the charming
efforts so that we make sure it properly meets your expectations and needs.
The team's started it this week and hopefully we'll have it your way soon.

dshcherb:
Something to consider:

```
JUJU_HOOK_NAME = install
JUJU_FUNCTION_NAME = install
```
This will require charms with one entry point to look at `JUJU_FUNCTION_NAME` first to find out if it is a function or not as names might collide.

As an idea, we could expose something like: `JUJU_EVENT=<type>:name` where <type> can be `hook` or `function` so that charms can look at a single variable to determine the event type.

mitechie:
I want to explore not using env vars but more direct like a hook tool,
something more direct like a context call that would spit out the name
along with any associated event data that's related. I wonder how this
falls into line with our "ask about the state of things" discussions.

timClicks:
A hook tool would be much more effective.

p.s. can we please port this conversation over to the development category
of discourse

hmlanigan:
$$merge$$

PR #10975, opened 2019-12-02
Get juju output without calling model command

SimonRichardson:
### Checklist

## Description of change

The following prevents an issue where the pylib code requests a model
before asking to switch. This can cause issues where it expects the
model to be there, even though we just unregistered it.

So the code now calls into the "global" juju commands, as we don't have
any other controllers (it checks it prior to calling this code) and so
it's safe to call `juju switch` and see the output correctly.

The code is actually doing what we expect here, rather than calling
`juju models` first.

### DIFF
--- a/acceptancetests/assess_unregister.py
+++ b/acceptancetests/assess_unregister.py
@@ -65,7 +65,7 @@ def assess_unregister(client):
 
 def assert_switch_raises_error(client):
     try:
-        client.get_juju_output('switch', include_e=False)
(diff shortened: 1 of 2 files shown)


SimonRichardson:
$$merge$$

SimonRichardson:
There are a lot of uniter flakiness currently.

$$merge$$

SimonRichardson:
$$merge$$

PR #10981, opened 2019-12-04
Adds a core/multiwatcher package.

howbazaar:
This is step one of the multiwatcher refactoring.

Introduce a new core package for multiwatcher types. Have the state multiwatchers use this core type. This necessitates changes in the modelcache worker and in the apiserver.

The apiserver now does the translation between the core/multiwatcher types and the types returned over the API using apiserver/params types. This should almost be the last of the apiserver/params dependencies in the state package.

## QA steps

Unit tests keep passing. Any bundle deployment uses the multiwatchers.
Bootstrap a new controller, deploy stuff into some models and look at the controller engine report to see that the model cache is being populated properly.

## Documentation changes

Everything is internal, no user facing changes.


howbazaar:
$$merge$$

PR #10982, opened 2019-12-04
Rework the cmd/juju/application tests to avoid using the real charmstore

wallyworld:
## Description of change

The tests in the cmd/juju/application package no longer use the real charmstore. The various test suites in use were already a mix of old and style go mocks, some used a real database, there were several variations of making a test command etc.

This PR doesn't change the world - it does what's needed to remove the need to import charmstore.v5. There's not a large code change - it's vast majority of changes are in the tests. The main aspects include:
- breaking up larger interfaces to make stubbing easier (eg deployAPI)
- with the above, introduce a smaller charm store interface
- inject the charm store implementation into the commands (use stubs for tests)
- place code to upload charms etc to the real store with functions to set up the stubs with expected api calls to mimic the desired store behaviour
- remove some tests which only tested real store behaviour (eg auth errors, registration of terms)

## QA steps

bootstrap a controller
deploy various local/store charms (with resources) and bundles
redeploy the bundle
upgrade the charms

### DIFF
--- a/apiserver/testing/fakecharmstore.go
+++ b/apiserver/testing/fakecharmstore.go
@@ -95,8 +95,3 @@ func (s *CharmStoreSuite) UploadCharmMultiSeries(c *gc.C, url, name string) (*ch
 	client := &charmstoreClientShim{s.Client}
 	return testcharms.UploadCharmMultiSeries(c, client, url, name)
 }
-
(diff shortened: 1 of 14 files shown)


wallyworld:
[remove charmstore.v5 from cmd/juju/application](https://trello.com/c/dmfUyrdf/987-remove-charmstorev5-from-cmd-juju-application)

wallyworld:
$$merge$$

PR #10990, opened 2019-12-05
Self issue

SimonRichardson:
### Checklist

----

## Description of change

Fix the issue where I typo'd the first argument AFTER I tested, but
before I pushed :(

Anyway, this fixes the issue around that.

## QA steps

```sh
cd acceptancetests
./assess_unregister.py lxd $GOPATH/bin/juju /tmp/artifacts
```

### DIFF
--- a/acceptancetests/jujupy/client.py
+++ b/acceptancetests/jujupy/client.py
@@ -1142,10 +1142,15 @@ class ModelClient:
         """
         model = self._cmd_model(kwargs.get('include_e', True),
                                 kwargs.get('controller', False))
+        # Get the model here first, before using the get_raw_juju_output, so
+        # we can ensure that the model exists on the controller.
         return self.get_raw_juju_output(command, model, *args, **kwargs)
 
     def get_raw_juju_output(self, command, model, *args, **kwargs):
         """Call a juju command without calling a model for it's values first.
+        Passing in the model, ensures that we target the juju command with the
+        right model. For global commands that aren't model specific, then you
+        can pass None.
 
         Sub process will be called as 'juju <command> <args> <kwargs>'. Note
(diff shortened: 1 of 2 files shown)


acceptancetests/jujupy/client.py
```diff
@@ -1142,7 +1142,7 @@ def get_juju_output(self, command, *args, **kwargs):
         """
         model = self._cmd_model(kwargs.get('include_e', True),
                                 kwargs.get('controller', False))
-        return self.get_raw_juju_output(self, command, model, *args, **kwargs)
```
SimonRichardson: I typo'd after I tested :(

acceptancetests/jujupy/client.py
```diff
@@ -1142,7 +1142,7 @@ def get_juju_output(self, command, *args, **kwargs):
...
+        return self.get_raw_juju_output(command, model, *args, **kwargs)
 
     def get_raw_juju_output(self, command, model, *args, **kwargs):
```
hmlanigan: should we put a default value for model in the def of get_raw_juju_output?

acceptancetests/jujupy/client.py
```diff
@@ -1142,7 +1142,7 @@ def get_juju_output(self, command, *args, **kwargs):
...
         """Call a juju command without calling a model for it's values first.
```
hmlanigan: An explanation of why a model is necessary give this line would be helpful.

acceptancetests/jujupy/client.py
SimonRichardson: I don't think we need to here, but we can re-evaluate at a later time.

SimonRichardson:
$$merge$$

PR #10991, opened 2019-12-05
WIP: cmd stderr handling: remove inconsistent output update dep

nammn:
### Checklist

----

## Description of change

Forces most of the command, which support formatting directives to use those on error cases.
The output will be: "ErrorMsg": "<msg>"

Right now, on error case, the formatting directive is ignored and we get `tabular` format, which sucks for automating things around juju cli.

## QA steps
run some commands using formatting directives

```sh
❯ juju controllers --format=json
{"ErrorMsg":"No controllers registered.\n\nPlease either create a new controller using \"juju bootstrap\" or connect to\nanother controller that you have been given access to using \"juju register\".\n"}
```

related pr: https://github.com/juju/cmd/pull/69

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -284,8 +284,7 @@
 
 [[constraint]]
   name = "github.com/juju/cmd"
-  revision = "5b90825cf5340f8272af5e2561ff1bafcaa3b8e6"
-  source = "github.com/nammn/cmd"
+  revision = "74d589ad88264276d1a38c24be18f5d068c201bc"
 
 [[constraint]]
   name = "github.com/juju/collections"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -284,7 +284,8 @@
 
 [[constraint]]
   name = "github.com/juju/cmd"
-  revision = "74922f23f6436a661dacd8c5fdf9cb15641e7a2e"
+  revision = "5b90825cf5340f8272af5e2561ff1bafcaa3b8e6"
+  source = "github.com/nammn/cmd"
 
 [[constraint]]
   name = "github.com/juju/collections"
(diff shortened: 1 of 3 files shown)


Gopkg.toml
```diff
@@ -284,7 +284,8 @@
 
 [[constraint]]
   name = "github.com/juju/cmd"
-  revision = "74922f23f6436a661dacd8c5fdf9cb15641e7a2e"
+  revision = "b890f76cf043622bd2ce4b4cab1713db13aff4a1"
+  source = "github.com/nammn/cmd"
```
nammn: will be updated later

cmd/juju/controller/listcontrollers.go
```diff
@@ -83,7 +83,7 @@ func (c *listControllersCommand) Run(ctx *cmd.Context) error {
 	if err != nil {
 		return errors.Annotate(err, "failed to list controllers")
 	}
-	if len(controllers) == 0 && c.out.Name() == "tabular" {
```
nammn: needed to be done to be consistent with other cli behaviours

cmd/juju/controller/listmodels.go
```diff
@@ -85,12 +85,10 @@ func (c *modelsCommand) SetFlags(f *gnuflag.FlagSet) {
 func (c *modelsCommand) Run(ctx *cmd.Context) error {
 	controllerName, err := c.ControllerName()
 	if err != nil {
-		ctx.Infof(err.Error())
```
nammn: same as above..

nammn:
https://github.com/juju/cmd/pull/69

nammn:
> I think it would be good to see how you handle multiple writes to stdout. i.e. we should prevent someone writing to stdout twice which would break yaml/json parsers. Is that even possible, it might be worth looking at?

I agree I did not take that into consideration. Did we ever do that?

nammn:
old pr was only for `run` we need this for `init` as well:
https://github.com/juju/cmd/pull/71

nammn:
Closing this for now as too many changes are needed for the cmdline package

PR #10992, opened 2019-12-05
Fix some intermittent tests racing with internal events.

howbazaar:
With some of the testing infrastructure coming under increasing load, it changes some of the timing for the tests. This is leading to some intermittent test failures where we didn't see them before.

When tests that are full stack tests, like JujuConnSuite, and they create entities, we need to ensure that the internal events have been processed before we continue after creation, or we get spurious test failures like:
```
09:13:48 unitupgrader_test.go:132:
09:13:48     // Initial event
09:13:48     wc.AssertOneChange()
09:13:48 /home/ubuntu/go/src/github.com/juju/juju/core/watcher/watchertest/notify.go:95:
09:13:48     c.Fatalf("watcher sent unexpected change: (_, %v)", ok)
09:13:48 ... Error: watcher sent unexpected change: (_, true)
```


howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #10996, opened 2019-12-09
Fix action tag lookup with new ids

wallyworld:
## Description of change

If juju is upgraded from 2.6 and there are old action uuids, then running new actions creates new ids such that we have a mix of old and new. Using show-action-output or using --wait would cause the new numeric id to match with an older uuid as a prefix match was done.
This PR changes to logic so that if a new ids are expected, an exact match is done. A legacy prefix match is still possible against any older uuids by including at least up to the "-" in the match string.

## QA steps

bootstrap 2.6 and run some actions until a uuid starting with 1 is used
upgrade to 2.7
run an action --wait and ensure the correct behaviour

## Bug reference

https://bugs.launchpad.net/juju/+bug/1855084
Also driveby fix for
https://bugs.launchpad.net/juju/+bug/1855093

### DIFF
--- a/apiserver/facades/client/action/action.go
+++ b/apiserver/facades/client/action/action.go
@@ -200,7 +200,10 @@ func (a *ActionAPI) FindActionTagsByPrefix(arg params.FindTags) (params.FindTags
 		if err != nil {
 			return params.FindTagsResults{}, errors.Trace(err)
 		}
-		found := m.FindActionTagsById(prefix)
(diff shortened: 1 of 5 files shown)


wallyworld:
[action --wait error](https://trello.com/c/VRlkMRKY/992-action-wait-error)

wallyworld:
$$merge$$

PR #11000, opened 2019-12-09
[dev] use cidr based ip notation when merging lld addrs

achilleasa:
## Description of change

To create a new link layer device address, we are required to populate an object which among other things includes the IP address in CIDR notation. The code in the state package breaks this down into an IP and CIDR value which is then persisted to the DB.

When we retrieve a link layer device address from disk, we can only access the IP and subnet CIDR values.

The changes introduced by #10979 ([this](https://github.com/juju/juju/pull/10979/commits/2393a6d3d1ce273564080359a62d46090b952ceb) commit in particular) included additional logic to backfill the provider-specific IDs to existing link layer devices and their addresses. The introduced merge code would create a new set of link layer device addresses by re-using the original values (reported by the machiner) plus the provider-specific IDs. 

Unfortunately, that code contained a bug: instead of converting the existing address into CIDR format, it pulled the data from the `SubnetCIDR` method of the link layer device address. By passing that value back into the state, we ended up with invalid IP addresses for the machines (following x.y.z.0 and x.y.0.0 patterns due to the way that the state code works).

This PR contributes a small fix where we basically recombine the existing IP and CIDR into an IP-in-CIDR-format value which we can then pass to the state code and get the correct address values persisted to the DB.

## QA steps

Run `nw-network-health-aws` with this PR (this is how we caught the bug in the first place)

### DIFF
--- a/apiserver/facades/controller/instancepoller/instancepoller.go
+++ b/apiserver/facades/controller/instancepoller/instancepoller.go
@@ -237,13 +237,18 @@ func backfillProviderIDs(m StateMachine, ifaces []network.InterfaceInfo) error {
 				ParentName:  existingDev.ParentName(),
 			})
 
+			addrInCIDRNotation, err := network.IPToCIDRNotation(addr.Value(), addr.SubnetCIDR())
+			if err != nil {
+				return err
+			}
+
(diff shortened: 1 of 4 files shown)


achilleasa:
$$merge$$

PR #11014, opened 2019-12-11
Fix intermittent failure in ActionSuite.TestFindActionTagsByLegacyId

babbageclunk:
## Description of change

It repeatedly enqueues actions until it gets a uuid that starts with a digit, then switches to integer ids and tries to ensure a collision with the first digit of the uuid (to make sure the querying avoids them). But if the first digit of the uuid was a 0 it would fail because the numeric ids start from 1 so no actions were generated.

## QA steps

Running the test under stress-race used to fail in about 10-20 runs, now it runs way more than 50.
```sh
cd ~/juju/state; stress-race -check.f ActionSuite.TestFindActionTagsByLegacyId -check.v
```

## Documentation changes
None

## Bug reference
None

### DIFF
--- a/state/action_test.go
+++ b/state/action_test.go
@@ -601,12 +601,13 @@ func (s *ActionSuite) TestFindActionTagsByLegacyId(c *gc.C) {
 	for {
 		a, err := s.model.EnqueueAction(s.unit.Tag(), "action-1", nil)
 		c.Assert(err, jc.ErrorIsNil)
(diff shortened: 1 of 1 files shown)


babbageclunk:
!!build!!

babbageclunk:
$$merge$$

babbageclunk:
> Thanks for fixing my mess up

Thankfully *I've* never committed an intermittently failing test!

babbageclunk:
$$merge$$

PR #11018, opened 2019-12-11
Fixes lp#1844976: re-authenticate cloud client when model credential changes.

anastasiamac:
### Checklist

 - [ ] ~~Checked if it requires a [pylibjuju](https://github.com/juju/python-libjuju) change?~~
 - [ ] ~~Added [integration tests](https://github.com/juju/juju/tree/develop/tests) for the PR?~~
 - [ ] ~~Added or updated [doc.go](https://discourse.jujucharms.com/t/readme-in-packages/451) related to packages changed?~~

## Description of change

Each Juju model has a cloud client that allows communication with the cloud provider. When cloud information is updated on the model, CloudSpec watcher fires and causes EnvironTracker to re-authenticate this client. CloudSpec watcher was meant to have catered for the credential changes but didn't.

This PR ensures that, in addition to cloud changes  like regions, etc, CloudSpec watcher also reacts to changes to model credential reference (for example, when model credential is replaced), as well as to the changes to cloud credential content/attributes (for example, a password change). 

## QA steps

1. bootstrap
2. add model with a credential that will be manipulated, say password update 
3. deploy an application and wait for all units to stabilize
4. manipulate password both on the cloud and on Juju side
5. deploy another unit/application [units will stabilize]

## Bug reference

https://bugs.launchpad.net/juju/+bug/1844976

### DIFF
--- a/apiserver/common/cloudspec/cloudspec.go
+++ b/apiserver/common/cloudspec/cloudspec.go
@@ -31,9 +31,11 @@ type CloudSpecAPI interface {
 type cloudSpecAPI struct {
 	resources facade.Resources
 
-	getCloudSpec   func(names.ModelTag) (environs.CloudSpec, error)
-	watchCloudSpec func(tag names.ModelTag) (state.NotifyWatcher, error)
-	getAuthFunc    common.GetAuthFunc
+	getCloudSpec                           func(names.ModelTag) (environs.CloudSpec, error)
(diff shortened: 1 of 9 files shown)


anastasiamac:
[update credentials does not work for existing models](https://trello.com/c/ELax5dHF/1003-update-credentials-does-not-work-for-existing-models)

anastasiamac:
$$merge$$

PR #11029, opened 2019-12-13
Make help

SimonRichardson:
### Checklist

----

## Description of change

Understanding what is inside make can be somewhat annoying as we have a
LOT of if statements. The following commit aims to solve this, by adding
a very rudimental help command that will print out what's available.

I've already found it quite useful since I added this locally, so I'm
pushing this up so others might also find this beneficial.

It's really easy to add a new item to the help, just start a comment
with double hashes (##) and it will be picked up.

   `## <cmd name>: <description of cmd>`

The output for the make file currently is the following:

```sh
➜ make help
Usage:

  build                        Create Juju binaries
  check-deps                   Check dependencies are correct versions
  check-k8s-model              Check if k8s model is present in show-model
  check                        Verify Juju code using static analysis and unit tests
  clean                        Clean the cache and test caches
  dep                          Installs go dependencies
  format                       Format the go source code
  go-build                     Build Juju binaries without updating dependencies
  go-install                   Install Juju binaries without updating dependencies
  install-dependencies         Install all the dependencies
  install-etc                  Install auto-completion
  install                      Install Juju binaries
  install-mongo-dependencies   Install Mongo and its dependencies
  install-snap-dependencies    Install the supported snap dependencies
  local-operator-update        Update local operator image
  microk8s-operator-update     Save and update the operator image for use with microk8s
  operator-imaage              Build the operator image
  pre-check                    Verify go code via static analysis
  push-operator-image          Push up the build operator image via docker
  rebuild-dependencies         Update the dependencies
  rebuild-schema               Rebuild the schema for clients with the latest facades
  release-build                Construct Juju binaries, without building schema
  release-install              Install Juju binaries
  run-tests                    Run the unit tests
  setup-lxd                    Auto configure LXD
  simplify                     Format and simplify the go source code
  statick-analysis             Check the go code using static-analysis
  test                         Verify Juju code using unit tests
```

## QA steps

```sh
make help
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -234,11 +234,11 @@ operator-image:
 	rm -rf ${JUJUD_STAGING_DIR}
 
 push-operator-image: operator-image
-## push-operator-image: Push up the build operator image via docker
+## push-operator-image: Push up the new built operator image via docker
 	docker push ${OPERATOR_IMAGE_PATH}
 
 microk8s-operator-update: operator-image
-## microk8s-operator-update: Save and update the operator image for use with microk8s
+## microk8s-operator-update: Push up the new built operator image for use with microk8s
 	docker save ${OPERATOR_IMAGE_PATH} | microk8s.ctr --namespace k8s.io image import -
 
 check-k8s-model:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -206,16 +234,20 @@ operator-image:
 	rm -rf ${JUJUD_STAGING_DIR}
 
 push-operator-image: operator-image
+## push-operator-image: Push up the build operator image via docker
```
nammn: * build and push up ..., Else people might think that they need to run `make operator-image` beforehand IMO

Makefile
```diff
@@ -206,16 +234,20 @@ operator-image:
...
 	docker push ${OPERATOR_IMAGE_PATH}
 
 microk8s-operator-update: operator-image
+## microk8s-operator-update: Save and update the operator image for use with microk8s
```
nammn: build, save and update...

may want to add that `... Save and update the operator image in registry`

Makefile
```diff
@@ -224,6 +256,7 @@ local-operator-update: check-k8s-model operator-image
 STATIC_ANALYSIS_JOB ?= 
 
 static-analysis:
+## statick-analysis: Check the go code using static-analysis
```
nammn: typo

Makefile
nammn: Or we could just name the step which is used 
`## push-operator-image: run operator-image and push up .... `

SimonRichardson:
$$merge$$

PR #11031, opened 2019-12-13
Comment and compiler warning updates to uniter context

hmlanigan:
### Checklist

----

## Description of change

During uniter investigation, updated out of date comments and added new in context.go.  Fixed a few compiler warnings as well.

## QA steps

```sh
juju bootstrap localhost testme --build-agent
juju deploy ubuntu
```

### DIFF
--- a/worker/uniter/runner/context/context.go
+++ b/worker/uniter/runner/context/context.go
@@ -75,7 +75,6 @@ type Clock interface {
 }
 
 var logger = loggo.GetLogger("juju.worker.uniter.context")
-var mutex = sync.Mutex{}
 var ErrIsNotLeader = errors.Errorf("this unit is not the leader")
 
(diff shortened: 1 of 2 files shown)


worker/uniter/runner/context/context.go
```diff
@@ -75,7 +75,6 @@ type Clock interface {
 }
 
 var logger = loggo.GetLogger("juju.worker.uniter.context")
-var mutex = sync.Mutex{}
```
achilleasa: I wonder if the intent of this being global is to block concurrent execution of hooks on the same machine. If that's the case then we shouldn't be moving it into the HookContext struct.

@jameinel are you aware of any reasons why this has to be global?

hmlanigan:
$$merge$$

PR #11035, opened 2019-12-16
Speeds up the loading of the units by caching the opened ports.

howbazaar:
The opened ports are kept on documents associated with the machines and the subnet.
All the current code around dealing with opened ports in the allwatcher assume
an empty subnet ID. This will need to be updated as we start work on the limited
exposing of ports to subnets.

As the model is processed it sets the model type in the context.

## QA steps

Unit tests continue to pass.

### DIFF
--- a/state/allwatcher.go
+++ b/state/allwatcher.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/juju/juju/core/constraints"
 	"github.com/juju/juju/core/network"
+	corenetwork "github.com/juju/juju/core/network"
 	"github.com/juju/juju/core/status"
 	"github.com/juju/juju/environs/config"
 	"github.com/juju/juju/state/multiwatcher"
@@ -30,6 +31,7 @@ var allWatcherLogger = loggo.GetLogger("juju.state.allwatcher")
 type allWatcherStateBacking struct {
 	st               *State
 	watcher          watcher.BaseWatcher
+	collections      []string
 	collectionByName map[string]allWatcherStateCollection
 }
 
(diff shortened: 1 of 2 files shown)


state/allwatcher.go
```diff
@@ -1336,30 +1328,34 @@ type backingEntityDoc interface {
 
 func newAllWatcherStateBacking(st *State, params WatchParams) Backing {
 	collectionNames := []string{
+		// The ordering here matters. We want to load machines, then
```
hpidcock: Is there a test you could write (or one that already exists) to confirm this ordering?

state/allwatcher.go
```diff
@@ -1826,6 +1861,65 @@ func (ctx *allWatcherContext) getInstanceData(id string) (instanceData, error) {
 	return getInstanceData(ctx.state, id)
 }
 
+func (ctx *allWatcherContext) assignedMachineID(u *Unit) (string, error) {
+	// c.f. unit.AssignedMachineId.
```
hpidcock: Remove comment?

state/allwatcher.go
howbazaar: The existing tests would have started failing due to the change in processing within a single unit if the ordering wasn't working.

The new code to determine the opened ports for a unit require the principal to have been processed first, otherwise it would raise errors.

state/allwatcher.go
howbazaar: I'll expand on what I want this comment to mean.

howbazaar:
$$merge$$

PR #11036, opened 2019-12-16
More the multiwatcher store into the multiwatcher core package.

howbazaar:
The multiwatcherStore type is completely independent of the state package.
It is only concerned with storing entities that the multiwatcher iterates
over and provides a way to reference count the entities. This type is
now moved to the core/multiwatcher package.

This is entirely an internal move of implementation. There are no user facing changes.


howbazaar:
$$merge$$

PR #11043, opened 2019-12-16
Juju version sourced from correct jujud bin.

hpidcock:
### Checklist

 - [x] Checked if it requires a [pylibjuju](https://github.com/juju/python-libjuju) change?
 - [x] Added [integration tests](https://github.com/juju/juju/tree/develop/tests) for the PR?
 - [x] Added or updated [doc.go](https://discourse.jujucharms.com/t/readme-in-packages/451) related to packages changed?
 - [x] Do comments answer the question of why design decisions were made?

----

## Description of change

Build infra uses a jujud not in PATH, version wasn't being
picked up correctly.

## QA steps

Assuming you only have one jujud in your PATH
```
TMP_DIR=`mktemp -d`
mv ${GOPATH}/bin/jujud ${TMP_DIR}/
make JUJUD_BIN_DIR=${TMP_DIR} operator-image
``` 

## Documentation changes

N/A

## Bug reference

N/A

PR #11038, opened 2019-12-16
Fix charm-storage test.

hpidcock:
## Description of change

Fix charm-storage test, storage constraints need to be passed as multiple arguments.

## QA steps

N/A

## Documentation changes

N/A

## Bug reference

N/A

PR #11046, opened 2019-12-17
Forward port the fixes from 2.7 branch.

howbazaar:
Brings in changes from:
 #11027, #11012, #11017, #11035, #11038, #11040, #11041, and #11043


howbazaar:
$$merge$$

PR #11055, opened 2019-12-19
cmr-api: Fix intermittent failure in TestWatchRelationChangesV1Fallback

babbageclunk:
## Description of change
Occasionally (after about 50 runs under stress-race) the API calls from the client + watcher were coming in in a different order. Sometimes `commonLoop` manages to call Next before the main loop of the watcher has called RelationUnitSettings, so instead of seeing `[WatchRelationUnits, RelationUnitSettings, Next, Next, Stop]` we see `[WatchRelationUnits, Next, RelationUnitSettings, Next, Stop]`.
 Just check that the calls all happen without worrying about the order.

## QA steps
Ran under stress-race for a long time without failure.

## Documentation changes
None

## Bug reference
None

### DIFF
--- a/api/crossmodelrelations/crossmodelrelations_test.go
+++ b/api/crossmodelrelations/crossmodelrelations_test.go
@@ -462,15 +462,25 @@ func (s *CrossModelRelationsSuite) TestWatchRelationChangesV1Fallback(c *gc.C) {
 
 	err = workertest.CheckKilled(c, w)
 	c.Check(err, gc.ErrorMatches, "UHOH")
-	calls.CheckCallNames(c,
+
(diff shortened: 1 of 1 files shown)


howbazaar:
Wouldn't it be better to know why they come out of order?

babbageclunk:
I mean, sure - but it's because goroutines. The out of order part is that sometimes `commonLoop` manages to call Next before the main loop of the watcher has called RelationUnitSettings, so instead of seeing [WatchRelationUnits, RelationUnitSettings, Next, Next, Stop] we see [WatchRelationUnits, Next, RelationUnitSettings, Next, Stop]

babbageclunk:
$$merge$$

PR #11056, opened 2019-12-19
Use the multiwatcher worker in the apiserver and model cache.

howbazaar:
This requires updates to the manifolds. For the model cache it replaces
the need to get the state pool. For the apiserver, it needs the
multiwatcher worker in addition to its other pieces.

The multiwatcher worker is now started in the apiserver tests and
in the dummy provider to make the multiwatchers available.

## QA steps

Changes are all internal, no user facing changes. As long as the unit tests all continue to pass and we can deploy a bundle, we should be all good.

## Documentation changes

Everything is internal, no user facing change.

Except that now if you call Watch on a model, you also get a model update. But I don't think this is documented anywhere anyway.


howbazaar:
!!build!!

worker/modelcache/manifold_test.go
```diff
@@ -141,46 +137,12 @@ func (s *ManifoldSuite) TestNewWorkerArgs(c *gc.C) {
...
-}
-
-// Return an invalid but non-zero state pool pointer.
-// Is only ever used for comparison.
-func (f *fakeStateTracker) Use() (*state.StatePool, error) {
-	return f.pool(), nil
-}
-
-func (f *fakeStateTracker) pool() *state.StatePool {
-	return (*state.StatePool)(unsafe.Pointer(f))
```
hpidcock: ohmy

howbazaar: heh, yeah, I just wanted a non-zero pointer

howbazaar:
$$merge$$

hpidcock:
Kube core bundle deployed on lxd nicely.

howbazaar:
$$merge$$

PR #11062, opened 2019-12-19
Build less version checking

SimonRichardson:
### Checklist

----

## Description of change

The following brings in build less version checking. When looking for
streams or updates, using the build numbers to signify a change isn't
required. The uniqueness of a build number offers to much variation for
juju to check, instead Major.Minor.Patch.Tag is enough as that's
currently what we store for streams.

These changes should fix the issues around not being able to bootstrap
without the remote source available if you're not on ubuntu.

## QA steps

This is quite involved.

Build the binary for windows, then scp the build to a windows box.
Attempt to bootstrap juju on aws.

```sh
juju.exe bootstrap aws aws --debug --config agent-metadata-url=https://ci-run-streams.s3.amazonaws.com/ --config agent-stream=build-52959e1
```

Note the build agent config stream doesn't actually matter here, as it will fail before that can be used!

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -779,11 +779,11 @@
   revision = "bf9cc5bdd62dabc40b7f634b39a5e2dc44d44c45"
 
 [[projects]]
-  digest = "1:eab278f98004677adf1d70d6cdaba1c01a446ab10dddb1cf7d83cfd1475c22cc"
+  digest = "1:ac49627092b42eca5480db7fdae2814f60066dc922907b6c49bfddd30db5824e"
   name = "github.com/juju/version"
   packages = ["."]
   pruneopts = ""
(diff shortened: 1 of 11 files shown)


hmlanigan:
@SimonRichardson is this the expected result?  https://pastebin.canonical.com/p/JJjbTNnrrp/

SimonRichardson:
More work around streams is required... Will do it tomorrow.

hmlanigan:
Tried a newer agent-stream value and got to https://pastebin.canonical.com/p/vfQ3gTpznZ/.  Which  is past the fail in ci.

hmlanigan:
Also ran qa on lxd, to do upgrades etc.  No issues there.

hpidcock:
I believe #11065 fixes issues @hmlanigan was having.

PR #11063, opened 2019-12-19
Remove the now unneeded multiwatcher aspects in the state package.

howbazaar:
Now that we have the multiwatcher worker hooked up, and used by the model cache and api server, the deprecated methods in state can now be removed.

There is now only one type of all watcher backing, and that is the one that looks at all the models. The collections that were previously only in the single model watcher are added to the multi-model backing.

The end to end tests are not needed in state, and existing end to end tests can be found in the apiserver facade tests.

## QA steps

Running the unit tests. This branch just deletes stuff.

## Documentation changes

Internal only, no doc changes needed.


howbazaar:
$$merge$$

PR #11064, opened 2019-12-20
Address issues found with scale testing.

howbazaar:
During unit teardown, the all watcher code tries to update the application
status as it is updating the unit. This can cause a timing issue where
the unit is removed in the middle of the attempt to derive an application
status. A bug has been filed to remove the deriviation of the status.
    
Also the ec2 provider mentioned *a lot* that it was opening the model.
This has been reduced from INFO to DEBUG.


howbazaar:
$$merge$$

PR #11065, opened 2019-12-20
Fix tools dir when using build number

hpidcock:
## Description of change

This fixes another issue with the build number usage.
The versioned tools directory was including the build number as it
in some cases was derived from jujuversion.Current.

Contains changes from #11062

## QA steps

Run unit tests + bootstrap IAAS with local jujud.

```
make JUJU_BUILD_NUMBER=10 install
juju bootstrap localhost
juju deploy cs:trusty/mariadb-7
```

Should be also tested with simplestreams.

## Documentation changes

N/A

## Bug reference

N/A

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -779,11 +779,11 @@
   revision = "bf9cc5bdd62dabc40b7f634b39a5e2dc44d44c45"
 
 [[projects]]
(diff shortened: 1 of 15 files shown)


wallyworld:
$$merge$$

PR #11071, opened 2020-01-06
Record official build number as separate variable

wallyworld:
## Description of change

We are adding the Jenkins build number to version.Current for beta/rc builds. This conflicts with the original intent of the Build attribute, which was to track non-official juju builds. And we were incorrectly zeroing out the Build value when making the SharedToolsDir (where the agents are unpacked to), which screws upgrades and causes other unitended issues.

This PR reverts the version.Build semantics to what it was originally, and introduces a new OfficialBuild variable to the juju/version package - this is what is set by the Jenkins build process and what is used to compute the jujud oci image path, rather than trying to overload the use of Build.

## QA steps

Build a version of Juju with an "official" build number added in:
JUJU_BUILD_NUMBER=666; make install
As a developer, bootstrap a controller and ensure upgrades with --build-agent work.
Hack the official version yaml file to perform the test with what juju thinks is an official version.
Acceptance tests should also start passing.

### DIFF
--- a/agent/tools/tools_test.go
+++ b/agent/tools/tools_test.go
@@ -208,90 +208,53 @@ func (t *ToolsSuite) TestReadGUIArchiveSuccess(c *gc.C) {
 }
 
 func (t *ToolsSuite) TestChangeAgentTools(c *gc.C) {
-	tests := []struct {
-		version1 version.Binary
-		version2 version.Binary
-		dir1     string
-		dir2     string
-	}{
-		{
-			version.MustParseBinary("1.2.3-quantal-amd64"),
-			version.MustParseBinary("1.2.4-quantal-amd64"),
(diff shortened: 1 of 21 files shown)


agent/tools/tools_test.go
```diff
@@ -208,90 +208,53 @@ func (t *ToolsSuite) TestReadGUIArchiveSuccess(c *gc.C) {
 }
 
 func (t *ToolsSuite) TestChangeAgentTools(c *gc.C) {
-	tests := []struct {
```
wallyworld: All this is a revert back to old 2.7 behaviour

wallyworld:
[fix juju version build number issue](https://trello.com/c/7fSGuJlq/1016-fix-juju-version-build-number-issue)

wallyworld:
$$merge$$

PR #11074, opened 2020-01-07
Adds controller name to controller config.

tlm:
### Checklist

----

## Description of change

Allows a user to now set a canonical name for a controller that will be passed through at bootstrap time.

## QA steps

```sh
juju bootstrap localhost batman # exec to the local lxd container and check /var/lib/juju/bootstrap-params, also check juju controller-config for controller-name set
juju bootstrap localhost  --config controller-name=batman1 # Will result in error saying controller name cannot be set via config
juju bootstrap localhost batman-arg --config controller-name=batman-config # should fail with an error
juju bootstrap localhost # controller name in juju controller-config should be defaulted to something like localhost-localhost
juju controller-config controller-name=batman2 # confirm changed with juju controller-config
```


tlm:
[controller name in model watcher](https://trello.com/c/bBiwYOW7/1009-controller-name-in-model-watcher)

cmd/juju/commands/bootstrap.go
```diff
@@ -631,9 +631,6 @@ to create a new model to deploy k8s workloads.
 	if err != nil {
 		return errors.Trace(err)
 	}
-	if c.controllerName == "" {
```
wallyworld: we want to retain this, see below

cmd/juju/commands/bootstrap.go
```diff
@@ -653,6 +651,15 @@ to create a new model to deploy k8s workloads.
 		}
 	}
 
+	cfgControllerName, _ := bootstrapCfg.controller[controller.ControllerName].(string)
+	if c.controllerName != "" && cfgControllerName != "" {
+		return errors.NotValidf("controller name and controller config name both set, only one allowed")
```
wallyworld: this will print 
"controller name and controller config name both set, only one allowed not valid"

you need to use the errors.NewNotValid() form

also there's no unit tests for this code (would have picked up the dodgy error message)

cmd/juju/commands/bootstrap.go
```diff
@@ -653,6 +651,15 @@ to create a new model to deploy k8s workloads.
 		}
 	}
 
+	cfgControllerName, _ := bootstrapCfg.controller[controller.ControllerName].(string)
```
wallyworld: For now, let's not allow the controller name to be set via --config at all - we want to avoid 2 ways of doing the same thing.
So just check if bootstrapCfg.controller[controller.ControllerName] is not empty and complain if it is set

controller/config.go
```diff
@@ -43,6 +43,9 @@ const (
 	// properly.
 	ControllerAPIPort = "controller-api-port"
 
+	// Canonical name for the controller
```
wallyworld: It won't be used in this PR, but we need a ControllerName() getter and a unit test

cmd/juju/commands/bootstrap.go
tlm: Done & Done

cmd/juju/commands/bootstrap.go
tlm: Done & Done

controller/config.go
tlm: Done & Done

tlm:
$$merge$$

tlm:
$$merge$$

tlm:
$$merge$$

PR #11076, opened 2020-01-07
Fix intermittent test failure in TestClientWatchAllReadPermission

howbazaar:
Ensure that the internal events for the new machine have flowed through the system before starting the watcher.

The internal multiwatcher store needs to have been updated with the machine before we start the watcher to ensure that the initial event has both the model and the machine.

## QA steps

```sh
cd apiserver/facades/client/client
stress -check.f TestClientWatchAllReadPermission
```

## Documentation changes

Internal test fix only


howbazaar:
$$merge$$

PR #11093, opened 2020-01-10
Get CharmProfilingInfo from state

SimonRichardson:
### Checklist

----

## Description of change

Bug 1856832 has shown that when applying lxd profiles to existing containers, it was possible that the actual profile was empty.  This could only happen when there was more than 1 unit of the given charm and was not guaranteed to happen.  This was not seen on newer, faster machines.  

The lxd profile contents was retrieved from the model cache.  There is an outstanding issue to find of why the profile was applied with empty contents, there are checks to prevent that.  

This fix changes CharmProfileInfo to get data from state rather than the model cache.

## QA steps

Note: these two charms will have missing relations and that is okay for the purposes of this test, as the issue is seen or not when the subordinate (neutron-openvswitch) is added.
```sh
juju bootstrap lxd
juju deploy nova-compute -n 8 --to lxd,lxd,lxd,lxd,lxd,lxd,lxd,lxd
juju deploy cs:neutron-openvswitch-271
juju add-relation nova-compute neutron-openvswitch
```
Check that the lxd profile 'juju-default-neutron-openvswitch-271' applied to each container is not empty.
```sh
juju ssh 0 -- sudo  lxc profile show juju-default-neutron-openvswitch-271
```
 
## Bug reference

https://bugs.launchpad.net/juju/+bug/1856832


howbazaar:
While doing some work on my model summary watcher, I noticed that the unit update event in the cache model looks like it is emitting a unit added event every time a subordinate updates its status. Probably worth checking.

hmlanigan:
This change will fix https://bugs.launchpad.net/juju/+bug/1856832.  You predicted it @SimonRichardson!

hmlanigan:
@SimonRichardson, tested with fdba050 and could no longer reproduce the issue in 1856832.  Priviously with 6 units I was 50% failures.  With our changes, at 7 units and 0% failure.  I started with this bundle:  https://paste.ubuntu.com/p/xRYT2m6kbg/ and kept doing `juju add-unit --to lxd nova-compute`

apiserver/facades/agent/instancemutater/shim.go
```diff
@@ -4,44 +4,57 @@
...
-type modelCacheShim struct {
-	*cache.Model
-}
-
-func (s *modelCacheShim) WatchMachines() (cache.StringsWatcher, error) {
-	return s.Model.WatchMachines()
+func (s *instanceMutaterStateShim) Application(appName string) (Application, error) {
+	app, err := s.State.Application(appName)
+	if err != nil {
+		return nil, err
```
SimonRichardson: nit:

```suggestion
		return nil, errors.WithStack(err)
```

hmlanigan: updated to error.Trace.

SimonRichardson:
@hmlanigan I can't approve my own PR, but I approve of all the changes :)

hmlanigan:
$$merge$$

PR #11095, opened 2020-01-13
Add permissions to the allwatcher.

howbazaar:
Adds user permissions for the model to the allwatcher.

This PR also introduces a change in the way that model updates are handled. Instead of regetting the model config, status info, and constraints, it just copies those values to the new model structure using the same pattern that is used for machines, applications, and others. This change made one of the tests invalid, so it was removed.

## QA steps

Just the unit tests at this stage. A followup branch will introduce the permissions to the model cache.

## Documentation changes

Internal change only.

## Bug reference

No bug, but this work is needed for the model summary watcher for the new dashboard.


howbazaar:
$$merge$$

PR #11096, opened 2020-01-13
Add a health endpoint to the apiserver.

howbazaar:
This is an unauthenticated endpoint that can be used by external services to detect if
the server is in a state where it will handle user connections.

## QA steps

```sh
juju bootstrap lxd test
juju ssh -m controller 0
curl --insecure https://localhost:17070/health
```

## Documentation changes

The purpose of this endpoint is to be used by the proxy sitting in front of an HA controller set. Perhaps if we had a section on managing named deployments with HA proxy, this is something we'd add to that section.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1797848

### DIFF
--- a/apiserver/apiserver.go
+++ b/apiserver/apiserver.go
@@ -5,6 +5,7 @@ package apiserver
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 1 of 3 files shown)


howbazaar:
$$merge$$

apiserver/apiserver.go
```diff
@@ -880,6 +903,14 @@ func (srv *Server) trackRequests(handler http.Handler) http.Handler {
 	})
 }
 
+func (srv *Server) healthHandler(w http.ResponseWriter, req *http.Request) {
+	srv.mu.Lock()
+	status := srv.healthStatus
+	srv.mu.Unlock()
+
+	fmt.Fprintf(w, "%s\n", status)
```
jameinel: doesn't this mean we return 200 under every case? Shouldn't we return a different HTTP response if the server isn't ready?

PR #11108, opened 2020-01-13
acceptancetests: Make destroy-model test more reliable on slow machines

babbageclunk:
## Description of change

The destroy-model acceptance test would reliably fail on grumpig because the error was "model has been removed" rather than "model not found" - this is returned when the model is still in the state pool. Change the test to wait (up to 100s) for the removed message to change to not found.

https://jenkins.juju.canonical.com/job/nw-destroy-model-lxd/2188/console

## QA steps

Ran the test locally, it passes. Tried to hack the local juju version to display the "model has been removed" message for longer, but couldn't for reasons I don't understand - pushing the test change up anyway.

## Documentation changes

None

## Bug reference

None

### DIFF
--- a/acceptancetests/assess_destroy_model.py
+++ b/acceptancetests/assess_destroy_model.py
@@ -7,6 +7,7 @@ import argparse
 import logging
 import sys
 import subprocess
+import time
 
 from deploy_stack import (
     BootstrapManager,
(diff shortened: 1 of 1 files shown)


babbageclunk:
Adding this after @manadart's fix because I think we still want to ensure that the error does eventually resolve to "model not found".

babbageclunk:
$$merge$$

PR #11109, opened 2020-01-13
raft-forwarder: Track metrics for request times

babbageclunk:
### Checklist

----

## Description of change

We're seeing raft lease operation times longer than we'd expect, particularly given the metrics from the raft internals. Adding metrics to the forwarder should give us more information about where the time is going. The forwarder will report both the full time for each operation from request to response, as well as the time for raft to apply each one and give a response.

## QA steps

Running `juju_metrics` on the raft-leader controller machine includes the new `juju_raftforwarder` metrics, broken down into "full" and "apply" sections.

```
# HELP juju_raftforwarder_request Request times for raft forwarder operations in ms
# TYPE juju_raftforwarder_request summary
juju_raftforwarder_request{section="apply",quantile="0.5"} 3.252875
juju_raftforwarder_request{section="apply",quantile="0.9"} 3.655554
juju_raftforwarder_request{section="apply",quantile="0.99"} 5.74508
juju_raftforwarder_request_sum{section="apply"} 601.1322479999994
juju_raftforwarder_request_count{section="apply"} 190
juju_raftforwarder_request{section="full",quantile="0.5"} 3.487774
juju_raftforwarder_request{section="full",quantile="0.9"} 3.8945
juju_raftforwarder_request{section="full",quantile="0.99"} 5.960011
juju_raftforwarder_request_sum{section="full"} 638.9324020000003
juju_raftforwarder_request_count{section="full"} 190
```

## Documentation changes
None

## Bug reference
None

### DIFF
--- a/worker/raft/raftforwarder/metrics.go
+++ b/worker/raft/raftforwarder/metrics.go
@@ -1,0 +1,58 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package raftforwarder
+
+import (
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	metricsNamespace = "juju_raftforwarder"
+)
+
+// metricsCollector is a prometheus.Collector that collects metrics
+// about how long it's taking to forward requests to raft and get
+// responses.
+type metricsCollector struct {
+	requests *prometheus.SummaryVec
+}
+
+func newMetricsCollector() *metricsCollector {
+	return &metricsCollector{
+		requests: prometheus.NewSummaryVec(prometheus.SummaryOpts{
+			Namespace: metricsNamespace,
+			Name:      "request",
+			Help:      "Request times for raft forwarder operations in ms",
+			Objectives: map[float64]float64{
+				0.5:  0.05,
+				0.9:  0.01,
+				0.99: 0.001,
+			},
+		}, []string{
+			// section can be "apply" (just the time to apply req and
+			// get a response from raft) or "full" (total time from
+			// request received to response sent)
+			"section",
+		}),
+	}
+}
+
+func (m *metricsCollector) record(start time.Time, section string) {
+	elapsedMS := float64(time.Now().Sub(start)) / float64(time.Millisecond)
+	m.requests.With(prometheus.Labels{
+		"section": section,
+	}).Observe(elapsedMS)
+}
+
+// Describe is part of prometheus.Collector.
+func (c *metricsCollector) Describe(ch chan<- *prometheus.Desc) {
+	c.requests.Describe(ch)
+}
+
+// Collect is part of prometheus.Collector.
+func (c *metricsCollector) Collect(ch chan<- prometheus.Metric) {
+	c.requests.Collect(ch)
+}
--- a/worker/raft/raftforwarder/worker.go
+++ b/worker/raft/raftforwarder/worker.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/raft"
 	"github.com/juju/errors"
 	"github.com/juju/pubsub"
+	"github.com/prometheus/client_golang/prometheus"
 	"gopkg.in/juju/worker.v1"
 	"gopkg.in/juju/worker.v1/catacomb"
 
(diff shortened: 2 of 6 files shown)


worker/raft/raftforwarder/metrics.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func newMetricsCollector() *metricsCollector {
+	return &metricsCollector{
+		requests: prometheus.NewSummaryVec(prometheus.SummaryOpts{
+			Namespace: metricsNamespace,
+			Name:      "request",
+			Help:      "Request times for raft forwarder operations in ms",
+			Objectives: map[float64]float64{
+				0.5:  0.05,
```
howbazaar: I don't understand this map.

Also, what is the difference between a SummaryVec and a GuageVec?

worker/raft/raftforwarder/worker.go
```diff
@@ -104,11 +111,16 @@ func (w *forwarder) Wait() error {
 
 func (w *forwarder) loop() error {
 	defer w.unsubscribe()
+	if w.config.PrometheusRegisterer != nil {
```
howbazaar: We don't need the nil check as a nil registerer is invalid config right?

worker/raft/raftforwarder/worker.go
```diff
@@ -129,15 +141,18 @@ func (w *forwarder) handleRequest(_ string, req raftlease.ForwardRequest, err er
...
+		w.metrics.record(start, "full")
 	}()
 }
 
 func (w *forwarder) processRequest(command string) (raftlease.ForwardResponse, error) {
 	var empty raftlease.ForwardResponse
+	start := time.Now()
 	future := w.config.Raft.Apply([]byte(command), applyTimeout)
 	if err := future.Error(); err != nil {
 		return empty, errors.Trace(err)
```
howbazaar: Wouldn't it make sense to have in a record before the return as well?

howbazaar: I guess there is a potential here that future.Error is returning an error sometimes

worker/raft/raftforwarder/metrics.go
```diff
@@ -0,0 +1,58 @@
...
+			// section can be "apply" (just the time to apply req and
+			// get a response from raft) or "full" (total time from
+			// request received to response sent)
+			"section",
+		}),
+	}
+}
+
+func (m *metricsCollector) record(start time.Time, section string) {
+	elapsedMS := float64(time.Now().Sub(start)) / float64(time.Millisecond)
```
howbazaar: Why not just use `time.Now().Sub(start).Milliseconds()`, it returns an int64

worker/raft/raftforwarder/metrics.go
babbageclunk: Summary is what produces the quantiles we use in the raft charts. Gauges are more limited - you can't calculate a rate for them, for example. We use them to track number of models, the docs use them for things like temperature - quantities that go up and down but don't represent specific events.
https://prometheus.io/docs/concepts/metric_types/

The map sets what quantiles the summary collects and the estimation error for each quantile, as far as I understand it. I copied those values from the docs - they're the defaults but they say not to rely on the defaults since they'll be changing at some point in the future. 
https://godoc.org/github.com/prometheus/client_golang/prometheus#Summary
https://godoc.org/github.com/prometheus/client_golang/prometheus#SummaryOpts

worker/raft/raftforwarder/worker.go
babbageclunk: Good point, removing.

worker/raft/raftforwarder/worker.go
babbageclunk: I wasn't going to include errors in the stats - do you think I should, as another axis?

worker/raft/raftforwarder/metrics.go
babbageclunk: Not sure - I cargo-culted this from the raftlease store metrics - just looking to see if there's a reason.

babbageclunk: Duration.Milliseconds() is Go 1.13. I think we're still Go 1.10 on the snap builders, maybe the others?

worker/raft/raftforwarder/worker.go
howbazaar: I think it would be interesting to see an error rate for the future. It should be really low right?

worker/raft/raftforwarder/metrics.go
howbazaar: Ah, fair enough. Perhaps leave a note to clean up with Go 1.13 :)

worker/raft/raftforwarder/metrics.go
howbazaar: OK

babbageclunk:
$$merge$$

PR #11124, opened 2020-01-17
Add relations to the cache, and extra model details.

howbazaar:
In order to provide the information to the upcoming model summary watcher (which will be implemented in the cache), we need to get some extra information into the cache.

There is a drive by that renames the ModelUpdate type from the core/multiwatcher package back to ModelInfo. This was renamed previously when the type was moved into the apiserver/params package and we had to deal with a naming collision. Renamed it back for this type for consistency with the rest of the package.

Adds three pieces of information from the state.Model in through the allwatcher: Cloud, Region, and Credential. Each of these is just a string value in the underlying document, so just adding these to the multiwatcher info type is sufficient to pass them along. Also needed to add these to the cache change type, and this makes them available to the internals of the cache.

The cache needed to know about relations, so pass the relation changes from the modelcache worker though to the controller, and add methods to add and remove the relations to the cached model.

## QA steps

Nothing is exposed to the outside world yet, that is coming.

## Documentation changes

Everything is internal, no user facing change.

### DIFF
--- a/core/cache/cachetest/state.go
+++ b/core/cache/cachetest/state.go
@@ -4,6 +4,8 @@
 package cachetest
 
 import (
+	"strings"
+
 	"github.com/juju/errors"
 	jc "github.com/juju/testing/checkers"
 	gc "gopkg.in/check.v1"
@@ -12,6 +14,7 @@ import (
 	"github.com/juju/juju/core/life"
 	"github.com/juju/juju/core/lxdprofile"
 	"github.com/juju/juju/core/model"
+	"github.com/juju/juju/core/permission"
 	"github.com/juju/juju/network"
 	"github.com/juju/juju/state"
 )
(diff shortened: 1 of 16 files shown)


core/cache/cachetest/state.go
```diff
@@ -32,13 +35,22 @@ func ModelChange(c *gc.C, model *state.Model) cache.ModelChange {
 	status, err := model.Status()
 	c.Assert(err, jc.ErrorIsNil)
 
+	users, err := model.Users()
+	permissions := make(map[string]permission.Access)
+	for _, user := range users {
+		// Cache permission map is always lower case.
+		permissions[strings.ToLower(user.UserName)] = user.Access
```
wallyworld: This works for external users also? ie fred@external.
I can't recall if UserName includes the domain suffix

howbazaar: It does. UserName is the fred@external bit.

howbazaar:
$$merge$$

howbazaar:

howbazaar:
$$merge$$ different intermittent failure

PR #11128, opened 2020-01-20
Make non-running status be '503 Service Unavailable'

howbazaar:
The health endpoint in the apiserver now returns a '503 Service Unavailable' instead of '200 OK' if the apiserver health isn't 'running'.

## QA steps

```sh
juju bootstrap lxd test
juju ssh -m controller 0
curl --insecure https://localhost:17070/health
```

## Documentation changes

When we add docs around this as outlined in #11096 we'd specify the status code return values.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1797848


howbazaar:
$$merge$$

PR #11131, opened 2020-01-21
Surface HA primary controller machine to the user.

anastasiamac:
## Description of change

In HA setups, there is no way to find out which machine is "primary". Primary controller machine is the one has a local mongo that is considered to be primary (or master) in a replicaset.

If the machine is a primary HA. then in HA setup, this will be reflected in the output of:
* 'juju status --format=yaml' - https://pastebin.ubuntu.com/p/XnkmKB4BD9/ 
* 'juju machines --format=yaml' - https://pastebin.ubuntu.com/p/2B4FC3tNWC/
* 'juju show-machine' - https://pastebin.ubuntu.com/p/KFZF72vwvg/
* 'juju show-controller' - https://pastebin.ubuntu.com/p/bvbgdBV5wR/

## QA steps

1. bootstrap
2. enable ha
3. run any of the command above
4. change replicaset master (I've logged in into primary and ```rs.stepDown(90,30)``` until a new primary was selected
5. run any of the modified command to see new primary

## Bug reference

https://bugs.launchpad.net/juju/+bug/1860291

### DIFF
--- a/apiserver/common/machine_test.go
+++ b/apiserver/common/machine_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/juju/naturalsort"
 	jc "github.com/juju/testing/checkers"
 	gc "gopkg.in/check.v1"
+	"gopkg.in/juju/names.v3"
 
 	"github.com/juju/juju/apiserver/common"
 	"github.com/juju/juju/apiserver/params"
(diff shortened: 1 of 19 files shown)


mitechie:
In status what do you think of having it declare that the non leaders have the key but a false value vs omission?

mitechie:
This is great. Thanks for getting into all the places.

anastasiamac:
@mitechie,

I was thinking about providing 'false' for all non-primary.. But it'll be a lot of noise... no? 

And if we are going to show false for non-primary, it'll be in all outputs for consistency :) not just in 'status'.

anastasiamac:
10:06:54 FAIL EXPECTED: contextfactory_test.go:518: ContextFactorySuite.TestNewHookContextRelationChangedUpdatesRelationContextAndCaches (application settings should be properly cached)	0.032s

!!build!!

apiserver/common/machine_test.go
```diff
@@ -239,10 +240,53 @@ func (s *machineSuite) TestMachineInstanceInfoWithSetDisplayName(c *gc.C) {
...
+	})
+}
+
 type mockState struct {
 	common.ModelManagerBackend
-	machines        map[string]*mockMachine
-	controllerNodes map[string]*mockControllerNode
+	machines         map[string]*mockMachine
+	controllerNodes  map[string]*mockControllerNode
+	haPimaryMachineF func() (names.MachineTag, error)
```
babbageclunk: typo: haPimaryMachineF

anastasiamac: Oooh ! eye of an eagle :) Thank you for spotting.

anastasiamac:
$$merge$$

mitechie:
Yes agree it'd be more than status. Since the yaml is meant to be machine
read it's often easier to just read a value vs check for its existence as
well. Also since it's meant to be more machine readable in figured the
noise wasn't much of an impact.

PR #11136, opened 2020-01-21
Cache unit state

hmlanigan:
## Description of change

As part of server-side state storage for charms in uniter++ work:
- Create API endpoints (in uniter facade) for reading/writing per-unit charm state.
- Add code in the uniter's hook context which will use the above APIs.  Nothing uses it yet.

There are Trello cards for pylibjuju updates and integration tests related to this project: 

- https://trello.com/c/SITS4ELq/2247-pyibjuju-updates 
- https://trello.com/c/VU01V4vG/2246-integration-tests

## QA steps

There is currently no direct way to test this code, it's part of the ongoing uniter++ project. There should be no impact to current juju operation.


worker/uniter/runner/context/context.go
```diff
@@ -892,6 +981,17 @@ func (ctx *HookContext) Flush(process string, ctxErr error) (err error) {
 		}
 	}
 
+	// TODO: We need a uniter API to persist the following in a single txn:
+	// - the state
+	// - relation data
+	// - storage changes
+	// - podspec
+	if ctx.cacheValues != nil && writeChanges {
```
hmlanigan: There is no way to call code which will initialize cacheValues, so we're safe with this for now.

worker/uniter/runner/jujuc/context.go
```diff
@@ -95,6 +96,19 @@ type actionHookContext interface {
 	LogActionMessage(string) error
 }
 
+// unitCacheContext is cache for charm state to be held in the context.
+type unitCacheContext interface {
```
achilleasa: If we want to break from the everything-is-some-kind-of-context pattern which is everywhere in the uniter we could rename this to something like `unitCacheReadWriter`. Guess since all other interfaces names are following the `xyzContext` pattern we are probably fine.

worker/uniter/runner/context/context.go
achilleasa: I really like this approach as it allows us to implement the server-side bits in a different PR.

hmlanigan:
!!build!!

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #11142, opened 2020-01-22
[dev] Setup mongo collection for server side charm state

achilleasa:
## Description of change

This PR is part of the work for supporting server-side stage storage for charms using the new operator framework. 

The per-unit state is stored within a new collection called `unitstates`. Each document in the new collection tracks the state for a particular unit which can be set via the `SetState` method and retrieved via the `State` method on a `Unit` instance. The state itself is stored as a map with string keys and values. Keys are allowed to contain special characters (e.g. dots and dollar signs) and such characters will automatically be escaped/unescaped when saving/loading state data.

Note that the absence of a document is treated as _no state available_ in which case a nil map will be returned by calls to `State`. This behavior also ensures that no migration steps are required from older controller versions.

It is also important to note that the reasoning behind reading/writing maps instead of individual KV pairs is that the uniter will buffer state changes in memory and persist them as a single batch when the uniter's hook context is flushed.

NOTE: 
This PR does not tackle unit state migrations as this requires changes to the juju/description package. This will be handled via a separate PR.

### DIFF
--- a/state/allcollections.go
+++ b/state/allcollections.go
@@ -288,7 +288,8 @@ func allCollections() CollectionSchema {
 				Key: []string{"model-uuid", "machineid"},
 			}},
 		},
-		minUnitsC: {},
+		unitStatesC: {},
+		minUnitsC:   {},
 
 		// This collection holds documents that indicate units which are queued
 		// to be assigned to machines. It is used exclusively by the
@@ -623,6 +624,7 @@ const (
 	txnLogC                    = "txns.log"
 	txnsC                      = "txns"
 	unitsC                     = "units"
+	unitStatesC                = "unitstates"
 	upgradeInfoC               = "upgradeInfo"
 	userLastLoginC             = "userLastLogin"
 	usermodelnameC             = "usermodelname"
--- a/state/unit.go
+++ b/state/unit.go
@@ -321,6 +321,93 @@ func (u *Unit) PasswordValid(password string) bool {
 	return false
 }
 
+// unitStateDoc records the state persisted by the charm executing in the unit.
+type unitStateDoc struct {
+	// DocID is always the same as a unit's global key.
+	DocID string `bson:"_id"`
+
+	// State encodes the unit's persisted state as a list of key-value pairs.
+	State map[string]string `bson:"state,omitempty"`
+
+	TxnRevno int64 `bson:"txn-revno"`
+}
+
+// SetState persisted the state for a unit.
+func (u *Unit) SetState(unitState map[string]string) error {
+	unitGlobalKey := u.globalKey()
+	txnDoc := struct {
+		TxnRevno int64 `bson:"txn-revno"`
+	}{}
+
+	buildTxn := func(attempt int) ([]txn.Op, error) {
+		coll, closer := u.st.db().GetCollection(unitStatesC)
+		defer closer()
+
+		if err := coll.FindId(unitGlobalKey).One(&txnDoc); err != nil {
+			if err != mgo.ErrNotFound {
+				return nil, errors.Trace(err)
+			}
+
+			escapedState := make(map[string]string, len(unitState))
+			for k, v := range unitState {
+				escapedState[mgoutils.EscapeKey(k)] = v
+			}
+
+			return []txn.Op{{
+				C:      unitStatesC,
+				Id:     unitGlobalKey,
+				Assert: txn.DocMissing,
+				Insert: unitStateDoc{
+					DocID: unitGlobalKey,
+					State: escapedState,
+				},
+			}}, nil
+		}
+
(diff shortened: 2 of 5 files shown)


state/unit.go
```diff
@@ -321,6 +321,103 @@ func (u *Unit) PasswordValid(password string) bool {
...
+	buildTxn := func(attempt int) ([]txn.Op, error) {
+		coll, closer := u.st.db().GetCollection(unitStatesC)
+		defer closer()
+
+		if err := coll.FindId(unitGlobalKey).One(&txnDoc); err != nil {
+			if err != mgo.ErrNotFound {
+				return nil, errors.Trace(err)
+			}
+
+			escapedState := make(map[string]string, len(unitState))
```
hmlanigan: Looks the same as line 378, suggest doing once before the if at 356.

state/unit.go
```diff
@@ -321,6 +321,103 @@ func (u *Unit) PasswordValid(password string) bool {
...
+// document associated with the given globalKey.
+func removeUnitStateOp(mb modelBackend, globalKey string) txn.Op {
+	return txn.Op{
+		C:      unitStatesC,
+		Id:     mb.docID(globalKey),
+		Remove: true,
+	}
+}
+
+// SetState persisted the state for a unit.
```
hmlanigan: Suggest adding that this is a blanket replacement of the state, no attempt is made to merge old with new.

state/unit.go
```diff
@@ -321,6 +321,103 @@ func (u *Unit) PasswordValid(password string) bool {
...
+func (u *Unit) SetState(unitState map[string]string) error {
+	unitGlobalKey := u.globalKey()
+	txnDoc := struct {
+		TxnRevno int64 `bson:"txn-revno"`
+	}{}
+
+	buildTxn := func(attempt int) ([]txn.Op, error) {
+		coll, closer := u.st.db().GetCollection(unitStatesC)
+		defer closer()
+
```
hmlanigan: We should ensure the unit is alive as part of the transaction in general.

state/allcollections.go
```diff
@@ -623,6 +624,7 @@ const (
 	txnLogC                    = "txns.log"
 	txnsC                      = "txns"
 	unitsC                     = "units"
+	unitStatesC                = "unitstates"
```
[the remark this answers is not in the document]
achilleasa: That would be preferred but I followed the all-lowercase convention used by the majority(?) of the other collection name (e.g. ipaddresses, usermodelname, storageattachments etc.)

state/unit.go
achilleasa: That's a valid point. I will add an assertion and a test for this case

state/unit.go
achilleasa: Actually, the types are different so I opted to dup the assignment code to group it with the initialization bits. 

This one is`map[string]string` the other is a `bson.M` (map[string]interface{})

state/unit_test.go
```diff
@@ -67,6 +67,36 @@ func (s *UnitSuite) TestApplication(c *gc.C) {
...
+func (s *UnitSuite) TestUnitStateMutation(c *gc.C) {
+	// Try fetching the state without a state doc present
+	ust, err := s.unit.State()
+	c.Assert(err, gc.IsNil)
+	c.Assert(ust, gc.IsNil, gc.Commentf("expected to receive a nil map when no state doc is present"))
+
+	// Set initial state; this should create a new unitstate doc
+	initialState := map[string]string{
+		"foo":          "bar",
+		"key.with.dot": "must work",
```
hmlanigan: Nice!

state/unit.go
```diff
@@ -321,6 +321,103 @@ func (u *Unit) PasswordValid(password string) bool {
...
+					State: escapedState,
+				},
+			}}, nil
+		}
+
+		// State keys may contain dots or dollar chars which need to be escaped.
+		escapedState := make(bson.M, len(unitState))
+		for k, v := range unitState {
+			escapedState[mgoutils.EscapeKey(k)] = v
+		}
```
hmlanigan: Do we want to protect against the change already being made?  Or no change necessary here?

achilleasa: Already made? You mean to diff the existing settings and the new and make the set a no-op?

achilleasa: (if nothing needs to change)

state/allcollections.go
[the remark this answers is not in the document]
hmlanigan: I've done it that way, but looks like we are not consistent at all in the code.

state/unit.go
hmlanigan: yes.

achilleasa:
$$merge$$

PR #11160, opened 2020-01-29
cmr: Handle applications being offered with a different name

babbageclunk:
### Checklist

----

## Description of change
The appdata changes to the crossmodelrelations and remoterelations facades incorrectly assumed that the application token always translates to the application tag, when in fact in offering models it contains the offer name (as an application tag). This meant that if an application was offered under a different name the crossmodel code would return a NotFound error when the worker in the consuming model tried to watch the new relation, so the worker would mark the SAAS terminated in its model and give up.

Split GetRelationTokens into a consuming version (which gets the application name from the relation) and an offering one that gets it from the offer.

## QA steps

```sh
juju deploy -m offering-model percona-cluster
juju offer offering-model.percona-cluster.db offered-db

juju deploy -m consuming-model mediawiki
juju relate -m consuming-model mediawiki:db offering-model.offered-db

# mediawiki should become unblocked because it has a db.
# The offered-db saas in consuming-model shouldn't be in "terminated" state.
```

## Documentation changes

None

## Bug reference

https://bugs.launchpad.net/juju/+bug/1860992

### DIFF
--- a/apiserver/facades/controller/crossmodelrelations/crossmodelrelations.go
+++ b/apiserver/facades/controller/crossmodelrelations/crossmodelrelations.go
@@ -378,7 +378,7 @@ func (api *CrossModelRelationsAPI) WatchRelationChanges(remoteRelationArgs param
 		if !ok {
 			return nil, empty, common.ErrPerm
 		}
-		relationToken, appToken, err := commoncrossmodel.GetRelationTokens(api.st, relationTag)
+		relationToken, appToken, err := commoncrossmodel.GetOfferingRelationTokens(api.st, relationTag)
 		if err != nil {
 			return nil, empty, errors.Trace(err)
 		}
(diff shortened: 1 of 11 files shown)


apiserver/facades/controller/crossmodelrelations/crossmodelrelations.go
```diff
@@ -406,6 +406,7 @@ func (api *CrossModelRelationsAPI) WatchRelationChanges(remoteRelationArgs param
 	for i, arg := range remoteRelationArgs.Args {
 		w, changes, err := watchOne(arg)
 		if err != nil {
+			logger.Tracef("not found watching relation %s: %s", arg.Token, errors.ErrorStack(err))
```
babbageclunk: This error stack and the one in the worker were really useful in working out exactly what wasn't found - so I figured I'd leave them in at trace level.

wallyworld:
$$merge$$

apiserver/facades/controller/crossmodelrelations/crossmodelrelations.go
SimonRichardson: Very much a fan of these!

PR #11158, opened 2020-01-29
[dev] Implement server-side bits for the State/SetState uniter API endpoints

achilleasa:
## Description of change

This PR is part of the work required for supporting server-side storage of charm state. 

It builds on the changes introduced by #11136 and adds the server-side logic for handling the new SetState/State calls for the uniter API

## QA steps

QA steps not available as the client-side bits are not currently wired to call out to this facade.

PR #11164, opened 2020-01-30
[dev] Persist changes to cached charm state when flushing uniter's context

achilleasa:
## Description of change

This PR is part of the work for persisting per-unit charm state on the controller. 

The changes from this PR bridge together the work from #11136 and #11158. The PR introduces a dirty flag to keep track of cache mutations (additions, modifications, deletions of existing keys) and skips redundant calls to `SetState` when flushing the context. 

The dirty flag is automatically cleared upon a successful `SetState` call which sets us up for the upcoming bits of work that involve flush requests that are explicitly triggered by charms.

## QA steps

QA steps not available as the hook tools for manipulating the charm state has not yet landed.

### DIFF
--- a/worker/uniter/runner/context/context.go
+++ b/worker/uniter/runner/context/context.go
@@ -285,6 +285,9 @@ type HookContext struct {
 	// the context is flushed.
 	cacheValues map[string]string
 
+	// A flag that keeps track of whether the unit's state has been mutated.
+	cacheDirty bool
+
 	mu sync.Mutex
 }
 
@@ -313,7 +316,13 @@ func (ctx *HookContext) SetCacheValue(key, value string) error {
 		return err
 	}
 
+	curValue, exists := ctx.cacheValues[key]
+	if exists && curValue == value {
+		return nil // no-op
+	}
+
 	ctx.cacheValues[key] = value
+	ctx.cacheDirty = true
 	return nil
 }
 
@@ -327,7 +336,12 @@ func (ctx *HookContext) DeleteCacheValue(key string) error {
 		return err
 	}
 
+	if _, exists := ctx.cacheValues[key]; !exists {
+		return nil // no-op
+	}
+
 	delete(ctx.cacheValues, key)
+	ctx.cacheDirty = true
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


worker/uniter/runner/context/context.go
```diff
@@ -313,7 +316,13 @@ func (ctx *HookContext) SetCacheValue(key, value string) error {
 		return err
 	}
 
+	curValue, exists := ctx.cacheValues[key]
+	if exists && curValue == value {
+		return nil // no-op
+	}
+
 	ctx.cacheValues[key] = value
+	ctx.cacheDirty = true
```
hmlanigan: It looks like there is a small gap in understanding if a real change has been made to the charm's saved state: a value is changed and returned to its original value before the data is flushed.   This approach appears to be best if we don't pull the data from state to verify a change before flush, or letting state determine that.

worker/uniter/runner/context/context_test.go
```diff
@@ -644,6 +644,48 @@ func (s *mockHookContextSuite) TestSetCacheStateErr(c *gc.C) {
 	c.Assert(err, gc.ErrorMatches, "loading unit state from database: testing an error")
 }
 
+func (s *mockHookContextSuite) TestFlushWithNonDirtyCache(c *gc.C) {
+	defer s.setupMocks(c).Finish()
+	hookContext := context.NewMockUnitHookContext(s.mockUnit)
+
+	s.mockUnit.EXPECT().State().Return(map[string]string{
```
hmlanigan: you could use s.expectStateValues() here as a replacement.

worker/uniter/runner/context/context_test.go
```diff
@@ -644,6 +644,48 @@ func (s *mockHookContextSuite) TestSetCacheStateErr(c *gc.C) {
...
+	err := hookContext.Flush("success", nil)
+	c.Assert(err, jc.ErrorIsNil)
+}
+
+func (s *mockHookContextSuite) TestSequentialFlushOfCacheValues(c *gc.C) {
+	defer s.setupMocks(c).Finish()
+	hookContext := context.NewMockUnitHookContext(s.mockUnit)
+
+	// We expect a single call for the following API endpoints
+	s.mockUnit.EXPECT().State().Return(map[string]string{
```
hmlanigan: you could use s.expectStateValues() here as a replacement.

worker/uniter/runner/context/context.go
achilleasa: If this scenario does happen (I believe it's rather unlikely but could be wrong) we will make an extra call to the server to persist the state and this will be a no-op server-side as we already check for state changes inside the update txn.

However, I don't anticipate this to be an issue; once we switch to a single transactional API call we will be making a call to the controller anyway when flushing.

worker/uniter/runner/context/context_test.go
achilleasa: Good idea. Let me add that in!

achilleasa:
$$merge$$

PR #11169, opened 2020-01-31
Finish move from utils/set to collections/set.

hmlanigan:
## Description of change

Complete the move from using utils/set to collections/set in preparation for removing set from utils.

## QA steps

There should be no change to current juju behavior.


hmlanigan:
had to run `make rebuild-dependencies` to update Gopkg.lock to fix the test.

hmlanigan:
$$merge$$

PR #11172, opened 2020-02-03
Fix race in worker/modelcache

wallyworld:
## Description of change

Fix a race in the modelcache worker.
When we were asking the cache for changes, a reference was being take, not a copy. This allowed new changes to arrive and update the reference from underneath the consumer.
Add a clone method to the EntityInfo types and use that when gathering the changes.

A generic json Marshall/Unmarshall clone was not suitable due to float vs int issues.

## QA steps

go test --race

### DIFF
--- a/core/multiwatcher/store.go
+++ b/core/multiwatcher/store.go
@@ -242,9 +242,10 @@ func (a *store) ChangesSince(revno int64) ([]Delta, int64) {
 			// and removed since the revno.
 			continue
 		}
+		// Use clone to make a copy to avoid races.
 		changes = append(changes, Delta{
 			Removed: entry.removed,
-			Entity:  entry.info,
+			Entity:  entry.info.Clone(),
 		})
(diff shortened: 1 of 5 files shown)


core/multiwatcher/store.go
```diff
@@ -216,7 +217,7 @@ func (a *store) Get(id EntityID) EntityInfo {
 
 // ChangesSince returns any changes that have occurred since
 // the given revno, oldest first.
-func (a *store) ChangesSince(revno int64) ([]Delta, int64) {
+func (a *store) ChangesSince(revno int64) ([]Delta, int64, error) {
```
howbazaar: Why would clone fail?

wallyworld: ah, it used to when a reflect based approach was tried but not now. i'll fix

wallyworld:
$$merge$$

PR #11184, opened 2020-02-05
Operations persistence

wallyworld:
## Description of change

Add "operations" to the juju persistent data model.
Operations group together actions run against multiple units - each action invocation on a unit is a task. There 2 commits:
1. the core operations persistence code
2. upgrade steps

The main work items are:
- add operations to mongo state
- update parent operation status / completion time etc whenever a child task/action is updated
- adjust pruning so that operations are pruned, not actions

The actions code to Begin()/Finish() an action wasn't using the buildTxn pattern so that all needed updating to properly implement checks for race conditions when updating an operation.

Pruning was adjusted so that for prune by size, the actions collection is what is used to measure the desired size (actions can be large), but the pruning is done on operations. A ratio of operation count to action count is used to estimate how many operations need to be pruned to reduce the actions collection size. For large counts, this will suffice.

## QA steps

Operations are not exposed yet. Test by juju running an action across multiple units and using the mongo client to look at the operations collection to see the mongo doc for the operation.

Next PR will start to wire the operations up to the CLI.


wallyworld:
[operations state model](https://trello.com/c/RhSD98ea/1037-operations-state-model)

api/uniter/action_test.go
```diff
@@ -46,7 +46,10 @@ func (s *actionSuite) TestAction(c *gc.C) {
 
 	for i, actionTest := range actionTests {
 		c.Logf("test %d: %s", i, actionTest.description)
+		operationId, err := s.Model.EnqueueOperation("a test")
```
hpidcock: go idiomatic gripe of mine, should be operationID..

apiserver/facades/client/action/action.go
```diff
@@ -270,6 +271,25 @@ func (a *ActionAPI) Enqueue(arg params.Actions) (params.ActionResults, error) {
...
+		}
+		if operationName == "" {
+			operationName = a.Name
+			continue
+		}
+		if operationName != a.Name {
+			operationName = "multiple actions"
+		}
+	}
+	operationId, err := a.model.EnqueueOperation(fmt.Sprintf("%v run on %v", operationName, strings.Join(receivers, ",")))
```
hpidcock: would prefer the summary to be put into a variable and passed in.

state/action_test.go
```diff
@@ -24,6 +23,7 @@ import (
 	"github.com/juju/juju/testing"
 	coretesting "github.com/juju/juju/testing"
 	"github.com/juju/juju/testing/factory"
+	jc "github.com/juju/testing/checkers"
```
hpidcock: This shouldn't have moved.

state/operation.go
```diff
@@ -0,0 +1,266 @@
...
+		ModelUUID: modelUUID,
+		Enqueued:  mb.nowToTheSecond(),
+		Status:    ActionPending,
+		Summary:   summary,
+	}, operationId, nil
+}
+
+// EnqueueOperation records the start of an operation.
+func (m *Model) EnqueueOperation(summary string) (string, error) {
+	doc, id, err := newOperationDoc(m.st, summary)
```
hpidcock: Shouldn't this be in a transaction builder due to the monotonic key ids?

state/prune.go
```diff
@@ -18,21 +18,45 @@ import (
...
-func pruneCollection(mb modelBackend, maxHistoryTime time.Duration, maxHistoryMB int, collectionName string, ageField string, timeUnit TimeUnit) error {
+func pruneCollection(
+	mb modelBackend, maxHistoryTime time.Duration, maxHistoryMB int,
+	collectionName string, ageField string, filter bson.D,
+	timeUnit TimeUnit,
+) error {
+	return pruneCollectionAndChildren(mb, maxHistoryTime, maxHistoryMB, collectionName, ageField, "", "", filter, 1, timeUnit)
+}
 
+// pruneCollection removes collection entries until
```
hpidcock: pruneCollectionAndChildren

state/prune.go
```diff
@@ -117,6 +155,33 @@ func (p *collectionPruner) pruneByAge() error {
...
+		return 0, nil
+	}
+	count, err := coll.Count()
+	if err == mgo.ErrNotFound || count <= 0 {
+		return 0, nil
+	}
+	if err != nil {
+		return 0, errors.Annotatef(err, "counting %s records", coll.Name)
+	}
+	// We are making the assumption that item sizes can be averaged for
```
hpidcock: Consider revising:
// For large numbers of items we are making an assumption that the size of items can be averaged to give a reasonable number of items to drop to reach the goal size.

state/prune.go
```diff
@@ -182,6 +238,8 @@ func (p *collectionPruner) pruneBySize() error {
 
 func deleteInBatches(
 	coll *mgo.Collection,
+	childColl *mgo.Collection,
+	childfield string,
```
hpidcock: childField

state/state_test.go
```diff
@@ -3217,6 +3221,9 @@ var findEntityTests = []findEntityTest{{
 }, {
 	tag: names.NewActionTag("fedcba98-7654-4321-ba98-76543210beef"),
 	err: `action "fedcba98-7654-4321-ba98-76543210beef" not found`,
+}, {
+	tag: names.NewOperationTag("666"),
+	err: `operation "666" not found`,
```
hpidcock: :smiling_imp:

state/upgrades.go
```diff
@@ -2836,3 +2836,21 @@ func RemoveControllerConfigMaxLogAgeAndSize(pool *StatePool) error {
 	}
 	return nil
 }
+
+// IncrementTasksSequence adds 1 to the "tasks" sequence.
+// Previously, numbering started at 0, now it starts at 1
+// so we need to ensure that upgraded controllers do not
+// get a conflicting task id.
+func IncrementTasksSequence(pool *StatePool) error {
```
hpidcock: From memory actions IDs started at 1. Now I didn't like that but is this inconsistent? or do other things start at 0 and actions was the odd one out.

upgrades/steps_28.go
```diff
@@ -0,0 +1,17 @@
...
+package upgrades
+
+// stateStepsFor272 returns upgrade steps for Juju 2.8.0.
+func stateStepsFor28() []Step {
+	return []Step{
+		&upgradeStep{
+			description: "increment tasks sequence by 1",
+			targets:     []Target{DatabaseMaster},
+			run: func(context Context) error {
+				return context.State().IncrementTasksSequence()
```
hpidcock: I feel like we should take the action seq add one and make it the operation id sequence. As in, all current actions should be put under operation entities with the same name. Like action-5 gets a operation-5 with action-5 being the child task.

api/uniter/action_test.go
wallyworld: righto

apiserver/facades/client/action/action.go
wallyworld: ok

state/action_test.go
wallyworld: fixed

state/prune.go
wallyworld: fixed

state/prune.go
wallyworld: ok

state/prune.go
wallyworld: doh

state/state_test.go
wallyworld: :-)

state/upgrades.go
wallyworld: they started at 1 because we used sequence returning 0 for the first call and then we added 1 outside that call. Now we use sequenceWithMin and that returns a number starting at 1. So we don't add 1 anymore so that can lead to a conflict the first time after upgrade.

upgrades/steps_28.go
wallyworld: operations and tasks share the same sequence generator
The operation IDs and task IDs must be unique to avoid ambiguity. So we never want operation-5 and action-5. It must always be operation-5 and action-6 (or 7 or 8 or whatever).

state/operation.go
wallyworld: yes it should, good catch

wallyworld:
$$merge$$

PR #11189, opened 2020-02-06
Enhance run action CLI to deal with new operations

wallyworld:
### Checklist

libjuju uses the existing Enqueue API - at some point as the need arises it can be updated to work with operations/tasks.

----

## Description of change

Introduce a new EnqueueV2 API for the actions facade which returns the overall operation id for the queued action(s). Enhance the run action CLI to use the operation id. Subsequent work will flesh out the new operations/tasks CLI.

## QA steps

deploy a charm with actions
enable the juju-v3 feature flag
run an action across multiple units, with/without background
run an action across a single unit, with/without background
check the text output of the CLI

### DIFF
--- a/api/action/client.go
+++ b/api/action/client.go
@@ -66,15 +66,15 @@ func (c *Client) Enqueue(arg params.Actions) (params.ActionResults, error) {
 	return results, err
 }
 
-// EnqueueV2 takes a list of Actions and queues them up to be executed by
-// the designated ActionReceiver, returning the ids of the overall operation
-// and each individual task.
-func (c *Client) EnqueueV2(arg params.Actions) (params.EnqueuedActions, error) {
+// EnqueueOperation takes a list of Actions and queues them up to be executed as
+// an operation, each action running as a task on the the designated ActionReceiver.
+// We return the ID of the overall operation and each individual task.
+func (c *Client) EnqueueOperation(arg params.Actions) (params.EnqueuedActions, error) {
 	results := params.EnqueuedActions{}
 	if v := c.BestAPIVersion(); v < 6 {
-		return results, errors.Errorf("EnqueueV2 not supported by this version (%d) of Juju", v)
+		return results, errors.Errorf("EnqueueOperation not supported by this version (%d) of Juju", v)
 	}
-	err := c.facade.FacadeCall("EnqueueV2", arg, &results)
+	err := c.facade.FacadeCall("EnqueueOperation", arg, &results)
 	return results, err
 }
 
--- a/api/action/client.go
+++ b/api/action/client.go
@@ -66,6 +66,18 @@ func (c *Client) Enqueue(arg params.Actions) (params.ActionResults, error) {
 	return results, err
 }
 
+// EnqueueV2 takes a list of Actions and queues them up to be executed by
+// the designated ActionReceiver, returning the ids of the overall operation
+// and each individual task.
+func (c *Client) EnqueueV2(arg params.Actions) (params.EnqueuedActions, error) {
+	results := params.EnqueuedActions{}
+	if v := c.BestAPIVersion(); v < 6 {
+		return results, errors.Errorf("EnqueueV2 not supported by this version (%d) of Juju", v)
+	}
+	err := c.facade.FacadeCall("EnqueueV2", arg, &results)
+	return results, err
+}
+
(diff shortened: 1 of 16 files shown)


api/action/client.go
```diff
@@ -66,6 +66,18 @@ func (c *Client) Enqueue(arg params.Actions) (params.ActionResults, error) {
 	return results, err
```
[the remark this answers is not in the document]
babbageclunk: After talking on IRC I get why this one needs to be kept around.

api/action/client.go
```diff
@@ -66,6 +66,18 @@ func (c *Client) Enqueue(arg params.Actions) (params.ActionResults, error) {
...
 }
 
+// EnqueueV2 takes a list of Actions and queues them up to be executed by
+// the designated ActionReceiver, returning the ids of the overall operation
+// and each individual task.
+func (c *Client) EnqueueV2(arg params.Actions) (params.EnqueuedActions, error) {
```
babbageclunk: This will be the method that sticks around after the old legacy one is removed (when the feature flag goes away) - it'll be a pity to have V2 on the end forever. Change it to EnqueueOperation?

apiserver/facades/client/action/action.go
```diff
@@ -247,13 +261,42 @@ func (a *ActionAPI) FindActionsByNames(arg params.FindActionsByNames) (params.Ac
...
+// EnqueueV2 takes a list of Actions and queues them up to be executed by
+// the designated ActionReceiver, returning the ids of the overall operation
+// and each individual task.
+func (a *ActionAPI) EnqueueV2(arg params.Actions) (params.EnqueuedActions, error) {
+	operationId, actionResults, err := a.enqueue(arg)
+	if err != nil {
+		return params.EnqueuedActions{}, err
+	}
+	results := params.EnqueuedActions{
+		OperationTag: names.NewOperationTag(operationId).String(),
```
babbageclunk: Oh, we can make a new tag type for operations but not for offers? ;)

cmd/juju/action/run.go
```diff
@@ -61,9 +61,15 @@ type runCommand struct {
 }
 
 const runDoc = `
-Run a charm action for execution on a given unit, with a given set of params.
+Run a charm action for execution on the given unit(s), with a given set of params.
 An ID is returned for use with 'juju show-operation <ID>'.
 
+A action executed on a given unit becomes a task with an ID that can be
+used with 'juju show-task <ID>'.
```
babbageclunk: Maybe "becomes a task with an ID. You can check the task's status with 'juju show-task <ID>'."? I found the current wording a bit confusing.

babbageclunk: So the command behaves differently running an action on one unit than running it on multiple units (no operation created)? Is that going to be confusing? Is it only if you specifically target it at one unit? Or does the same thing happen if you specify a whole application but it only has one unit?

Oh - reading ahead I think I've gone off half-cocked - sounds like it always makes an operation and gives you the op-id, but if it only contains one task it also tells you that id for convenience. Fine then!

cmd/juju/action/run.go
```diff
@@ -189,43 +197,56 @@ func (c *runCommand) Run(ctx *cmd.Context) error {
...
+		info[result.receiver] = map[string]string{
 			"id": actionTag.Id(),
 		}
 	}
+	ctx.Infof("")
 	if c.background {
-		if len(results.Results) == 1 {
-			ctx.Infof("Scheduled Operation %s", actionTag.Id())
-			ctx.Infof("Check status with 'juju show-operation %s'", actionTag.Id())
+		if numTasks == 1 {
```
babbageclunk: This Run method has become very long - maybe split it up a bit?

cmd/juju/action/run.go
wallyworld: correct :-)

api/action/client.go
wallyworld: ok

cmd/juju/action/run.go
wallyworld: ok

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
[task grouping CLI](https://trello.com/c/YJqzZLkH/1041-task-grouping-cli)

PR #11191, opened 2020-02-07
Move unused actions facade APIs out of the way and delete unused juju client code

wallyworld:
### Checklist

----

## Description of change

There's apis on the actions facade that are unused by any juju client api:
- ListAll
- ListPending
- ListRunning
- ListComplete

Move these out of the way so the there's less clutter when adding the new operations apis.
The client code in juju is deleted, and the api facade code is moved to a legacy file.
libjuju still uses these old apis, so they are retained till juju3.

## QA steps

bootstrap a controller and run an action


api/action/client.go
```diff
@@ -86,33 +86,6 @@ func (c *Client) FindActionsByNames(arg params.FindActionsByNames) (params.Actio
 	return results, err
 }
 
-// ListAll takes a list of Entities representing ActionReceivers and returns
```
wallyworld: all this code is moved to legacy.go

apiserver/facades/client/action/action.go
```diff
@@ -266,72 +266,6 @@ func (a *ActionAPI) Enqueue(arg params.Actions) (params.ActionResults, error) {
 	return results, err
 }
 
-// ListAll takes a list of Entities representing ActionReceivers and
```
wallyworld: all this code is moved to legacy.go

apiserver/facades/client/action/action_test.go
```diff
@@ -338,297 +338,6 @@ var testCases = []testCase{{
 	},
 }}
 
-func (s *actionSuite) TestListAll(c *gc.C) {
```
wallyworld: all this code is moved to legacy_test.go

wallyworld:
$$merge$$

PR #11198, opened 2020-02-10
internal status test: adding check for race condition

nammn:
### Checklist

----

## Description of change
After talking to @SimonRichardson about possible ways to fix this problem we came to the following conclusion: 

- This patch does not actually fix the test
- The `controllerstore` `file` access indicates the race condition, but to be sure adding the below patch should help a future fix by isolating the problem.
   - My gut feeling says that the problem may lie around the `controllerStore` access. As it is a filestore implementation on the `juju con suite` site. A fix may or may not be a change to the Memstore.
But our locks in place should not let that race condition happen.
- Hence, by splitting up the check to `runStatus` and the `file/modelaccess` we can make sure where the test fails.


nammn:
$$merge$$

PR #11199, opened 2020-02-10
[2.7] skip client-side validation for channels when resolving charm URLs

achilleasa:
This change removes the client-side validation for the channel argument to address the attached issue. 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1862091.


achilleasa:
$$merge$$

PR #11209, opened 2020-02-12
Update the schema-gen dependency and rebuild the scheam.

howbazaar:
Bring in the fix to the jsonschema-gen package to have the schema work again.

The introduction of the core/life package to the apiserver/params package brought the problem to light. The schema-gen package would create a constraints.Value type internally, but just call it Value. Then when it came across the life.Value type it incorrectly mapped this to the created constraints type.

The schema is also updated, and you'll see the life values are now strings again (as they are in 2.7).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -554,11 +554,11 @@
   revision = "a0ef8b74ebcffeeff9fc374854deb4af388f037e"
 
 [[projects]]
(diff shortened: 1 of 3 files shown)


howbazaar:
$$merge$$

howbazaar:

PR #11212, opened 2020-02-12
testing: move subnets stubs to mocks 

nammn:
## Checklist

----

## Description of change

this patch adds mocks to subnets as a preparation for further adding features to subnets.
Additionally, we plan to move from stubs to mocks. This patch should prepare most of the things for it.

The other file touched is the status_internal_test.go where I needed to skip some flaky test because of the possible slow modelcache. We should look into this further at the sprint

## QA steps
No feature/refactoring outside testing is done.
A successful test suite should be enough QA.


apiserver/facades/client/subnets/subnets_test.go
```diff
@@ -114,7 +174,7 @@ func (s *SubnetsSuite) TestNewAPIWithBacking(c *gc.C) {
 	agentAuthorizer := s.authorizer
 	agentAuthorizer.Tag = names.NewMachineTag("42")
 	facade, err = subnets.NewAPIWithBacking(
-		apiservertesting.BackingInstance,
+		&stubBacking{apiservertesting.BackingInstance},
```
nammn: this is done in preparation that we don't need to update the apiservertesting stubs, but can update them here and slowly move away from the stubs.

cmd/juju/status/status_internal_test.go
```diff
@@ -6099,6 +6099,7 @@ func (s *StatusSuite) PrepareBranchesOutput(c *gc.C) *context {
 }
 
 func (s *StatusSuite) TestBranchesOutputTabular(c *gc.C) {
+	c.Skip("bug: #1862376 - can't always read our own writes. Model-cache may be too slow")
```
nammn: this should be further discussed in a sprint. -> How do we want to handle test which rely on the state, hence modelcache.
If the modelcache is sometimes to slow. Do we have some kind of wait?

nammn:
$$merge$$

PR #11214, opened 2020-02-13
Facade/state work for action abort

hpidcock:
### Checklist

----

## Facade/state work for action abort

This PR encapsulates state and facade changes to deal with `juju cancel-action` aborting running actions.

Actions can enter an Aborting state when `juju cancel-action`, this will refire the action notifications watcher, allowing the uniter to handle the abort. The uniter will need to fetch the current action status hence the introduction of the ActionStatus facade call.

Second PR will follow to handle aborting action status.
Third PR will follow to handle aborting machine actions.

## QA steps

Create an action that sleeps for a few minutes.

juju run-action myapp/0 mysleepyaction
juju show-action-status action-1

Action should be running.

juju cancel-action action-1

Action should enter the aborting state, but not do anything.
Action should complete successfully after sleep has elapsed.

## Documentation changes

N/A

## Bug reference

N/A

### DIFF
--- a/apiserver/common/action.go
+++ b/apiserver/common/action.go
@@ -30,6 +30,10 @@ func ParamsActionExecutionResultsToStateActionResults(arg params.ActionExecution
 		status = state.ActionFailed
 	case params.ActionPending:
 		status = state.ActionPending
+	case params.ActionAborting:
+		status = state.ActionAborting
+	case params.ActionAborted:
+		status = state.ActionAborted
 	default:
 		return state.ActionResults{}, errors.Errorf("unrecognized action status '%s'", arg.Status)
 	}
@@ -162,7 +166,8 @@ func Actions(args params.Entities, actionFn func(string) (state.Action, error)) 
 	return results
 }
 
-// WatchOneActionReceiverNotifications to create a watcher for one receiver.
+// WatchOneActionReceiverNotifications returns a function for creating a
+// watcher on all action notifications (action adds + changes) for one receiver.
 // It needs a tagToActionReceiver function and a registerFunc to register
 // resources.
 // It's a helper function currently used by the uniter and by machineactions
@@ -174,6 +179,30 @@ func WatchOneActionReceiverNotifications(tagToActionReceiver func(tag string) (s
 			return nothing, err
 		}
 		watch := receiver.WatchActionNotifications()
+
+		if changes, ok := <-watch.Changes(); ok {
+			return params.StringsWatchResult{
+				StringsWatcherId: registerFunc(watch),
+				Changes:          changes,
+			}, nil
+		}
+		return nothing, watcher.EnsureErr(watch)
+	}
+}
+
+// WatchPendingActionsForReceiver returns a function for creating a
+// watcher on new pending Actions for one receiver.
+// It needs a tagToActionReceiver function and a registerFunc to register
+// resources.
+// It's a helper function currently used by the uniter and by machineactions
+func WatchPendingActionsForReceiver(tagToActionReceiver func(tag string) (state.ActionReceiver, error), registerFunc func(r facade.Resource) string) func(names.Tag) (params.StringsWatchResult, error) {
+	return func(tag names.Tag) (params.StringsWatchResult, error) {
+		nothing := params.StringsWatchResult{}
+		receiver, err := tagToActionReceiver(tag.String())
+		if err != nil {
+			return nothing, err
+		}
+		watch := receiver.WatchPendingActionNotifications()
 
 		if changes, ok := <-watch.Changes(); ok {
 			return params.StringsWatchResult{
--- a/state/action.go
+++ b/state/action.go
@@ -11,6 +11,7 @@ import (
 	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
+	jujutxn "github.com/juju/txn"
 	"github.com/juju/utils"
 	"github.com/juju/version"
 	"gopkg.in/juju/names.v3"
@@ -49,6 +50,13 @@ const (
 
 	// ActionRunning indicates that the Action is currently running.
 	ActionRunning ActionStatus = "running"
+
+	// ActionAborting indicates that the Action is running but should be
+	// aborted.
+	ActionAborting ActionStatus = "aborting"
+
+	// ActionAborted indicates the Action was aborted.
+	ActionAborted ActionStatus = "aborted"
 )
 
 type actionNotificationDoc struct {
@@ -68,6 +76,11 @@ type actionNotificationDoc struct {
 	// ActionID is the unique identifier for the Action this notification
 	// represents.
 	ActionID string `bson:"actionid"`
+
+	// Changed represents the time when the corrosponding Action had a change
+	// worthy of notifying on.
+	// NOTE: changed should not be set on pending actions, see actionNotificationWatcher.
+	Changed time.Time `bson:"changed,omitempty"`
 }
 
 type actionDoc struct {
@@ -137,6 +150,23 @@ func (m ActionMessage) Message() string {
 type action struct {
 	st  *State
 	doc actionDoc
+}
+
+// Refresh the contents of the action.
+func (a *action) Refresh() error {
+	actions, closer := a.st.db().GetCollection(actionsC)
+	defer closer()
+	id := a.Id()
+	doc := actionDoc{}
+	err := actions.FindId(id).One(&doc)
+	if err == mgo.ErrNotFound {
+		return errors.NotFoundf("action %q", id)
+	}
+	if err != nil {
+		return errors.Annotatef(err, "cannot get action %q", id)
+	}
+	a.doc = doc
+	return nil
 }
 
 // Id returns the local id of the Action.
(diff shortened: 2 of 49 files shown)


apiserver/common/action.go
```diff
@@ -166,14 +170,14 @@ func Actions(args params.Entities, actionFn func(string) (state.Action, error))
 // It needs a tagToActionReceiver function and a registerFunc to register
 // resources.
 // It's a helper function currently used by the uniter and by machineactions
-func WatchOneActionReceiverNotifications(tagToActionReceiver func(tag string) (state.ActionReceiver, error), registerFunc func(r facade.Resource) string) func(names.Tag) (params.StringsWatchResult, error) {
+func WatchOneActionReceiverNotifications(watchChanges bool, tagToActionReceiver func(tag string) (state.ActionReceiver, error), registerFunc func(r facade.Resource) string) func(names.Tag) (params.StringsWatchResult, error) {
```
wallyworld: This is confusing at first read cause it seems we are creating a watcher but optionally not watching for changes.
I think this may be better split up as 2 api calls:
- WatchPendingActionsForReceiver
- WatchOneActionReceiverNotifications

These would also be on the receiver below etc

state/action.go
```diff
@@ -68,6 +76,10 @@ type actionNotificationDoc struct {
 	// ActionID is the unique identifier for the Action this notification
 	// represents.
 	ActionID string `bson:"actionid"`
+
+	// Changed represents the time when the corrosponding Action had a change
+	// worthy ofnotifying on.
```
wallyworld: typo

state/action.go
```diff
@@ -672,15 +764,25 @@ func (st *State) matchingActionsByReceiverId(id string) ([]Action, error) {
...
+	running := bson.D{{"$or", []bson.D{
+		{{"status", ActionRunning}},
+		{{"status", ActionAborting}},
+	}}}
+	return st.matchingActionsByReceiverAndStatus(ar.Tag(), running)
+}
+
+// matchingActionsAborting finds actions that match ActionReceiver and
+// that are aborting.
+func (st *State) matchingActionsAborting(ar ActionReceiver) ([]Action, error) {
```
wallyworld: This doesn't appear to be called anywhere?

state/interface.go
```diff
@@ -131,7 +131,7 @@ type InstanceIdGetter interface {
 // queued up for itself
 type ActionsWatcher interface {
 	Entity
-	WatchActionNotifications() StringsWatcher
+	WatchActionNotifications(watchChanges bool) StringsWatcher
```
wallyworld: As stated above, let's add a second api instead

worker/uniter/remotestate/watcher.go
```diff
@@ -169,6 +170,10 @@ func (w *RemoteStateWatcher) Snapshot() Snapshot {
 	copy(snapshot.Actions, w.current.Actions)
```
wallyworld: maybe rename to PendingActions

worker/uniter/remotestate/snapshot.go
```diff
@@ -78,6 +78,10 @@ type Snapshot struct {
 	// be performed by this unit.
 	Actions []string
 
+	// ActionChanged contains a channel for each action used to
+	// signal when remote state of an action has changed.
+	ActionChanged map[string]chan struct{}
```
wallyworld: The snapshot doesn't typically contain channels - the channels are in the watcher implementation. The snapshop just represents the modelled state for the unit agent to reconcile.

state/watcher.go
```diff
@@ -3221,12 +3221,133 @@ func ensureSuffixFn(marker string) func(string) string {
...
+		return newCollectionWatcher(st, colWCfg{
+			col:    actionNotificationsC,
+			filter: makeIdFilter(st, actionMarker, receivers...),
+			idconv: actionNotificationIdToActionId,
+		})
+	}
+	return newLegacyActionNotificationWatcher(st, receivers...)
+}
+
+// collectionWatcher is a StringsWatcher that watches for changes on the
```
wallyworld: fix me

state/watcher.go
```diff
@@ -3221,12 +3221,133 @@ func ensureSuffixFn(marker string) func(string) string {
...
+type legacyActionNotificationWatcher struct {
+	commonWatcher
+	source chan watcher.Change
+	sink   chan []string
+	filter func(interface{}) bool
+}
+
+// newLegacyActionNotificationWatcher starts and returns a new StringsWatcher configured
+// with the given collection and filter function
+func newLegacyActionNotificationWatcher(backend modelBackend, receivers ...ActionReceiver) StringsWatcher {
```
wallyworld: I wonder if we can't unify the legacy and new watchers - the legacy watcher would only trigger if the status is pending; the new watcher would trigger if the status is pending or aborting. We'd have a single watcher with a filter used to expose the relevant changes.

worker/uniter/remotestate/snapshot.go
hpidcock: Yeah I agree I don't like it, but I'm going to need to refactor the resolver a bit to allow async operations to get fed changes in new snapshots. The alternative is adding an Action watcher that the uniter can invoke while its executing the specific action. Then we can leave the action notifications watcher alone.

wallyworld: It would be good to keep the solution "clean", so I'm not adverse to some refactoring if needed

worker/uniter/operation/executor.go
```diff
@@ -70,9 +71,22 @@ func (x *executor) Run(op Operation) error {
 	switch err := x.do(op, stepPrepare); errors.Cause(err) {
 	case ErrSkipExecute:
 	case nil:
+		done := make(chan struct{})
+		go func() {
+			for {
+				select {
+				case rs := <-remoteStateChange:
```
wallyworld: Need a check for channel closing?
```
rs, ok := <-remoteStateChange:
if !ok {
  return
}
```

worker/uniter/operation/executor.go
```diff
@@ -70,9 +71,22 @@ func (x *executor) Run(op Operation) error {
...
+					op.RemoteStateChanged(rs)
+				case <-done:
+					return
+				}
+			}
+		}()
 		if err := x.do(op, stepExecute); err != nil {
+			close(done)
```
wallyworld: defer close(done) above?

hpidcock: No because I only wanted the remote state change to trigger during the execute phase of the operation.

worker/uniter/operation/executor.go
hpidcock: the channel never closes but I can do that in case the semantics change

apiserver/common/action.go
```diff
@@ -162,7 +166,7 @@ func Actions(args params.Entities, actionFn func(string) (state.Action, error))
 	return results
 }
 
-// WatchOneActionReceiverNotifications to create a watcher for one receiver.
+// WatchOneActionReceiverNotifications to create a watcher actions for one receiver.
```
wallyworld: eparse?

apiserver/common/action.go
```diff
@@ -185,6 +189,29 @@ func WatchOneActionReceiverNotifications(tagToActionReceiver func(tag string) (s
 	}
 }
 
+// WatchPendingActionsForReceiver to create a watcher on pending actions for one receiver.
```
wallyworld: eparse?

state/action.go
```diff
@@ -49,6 +50,13 @@ const (
 
 	// ActionRunning indicates that the Action is currently running.
 	ActionRunning ActionStatus = "running"
+
+	// ActionAborting indicates that the Action is running but should be
+	// aborted.
+	ActionAborting ActionStatus = "aborting"
```
wallyworld: can we add these extra status values to the CLI list-operations command checks

worker/uniter/operation/executor.go
```diff
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/juju/errors"
+	"github.com/juju/juju/worker/uniter/remotestate"
```
wallyworld: add blank line

worker/uniter/resolver/loop_test.go
```diff
@@ -176,7 +176,65 @@ func (s *LoopSuite) TestLoop(c *gc.C) {
...
+			close(s.abort)
+			break
+		}
+		return nil, resolver.ErrNoOperation
+	})
+
+	s.executor.run = func(op operation.Operation, rs <-chan remotestate.Snapshot) error {
+		select {
+		case _, ok := <-rs:
+			c.Assert(ok, jc.IsTrue)
```
wallyworld: can we include a counter to check that we get the expected number of remote state watcher triggers?

apiserver/common/action.go
hpidcock: It's a copy paste. But I'll refactor the comments.

worker/uniter/resolver/loop.go
```diff
@@ -60,16 +60,44 @@ func Loop(cfg LoopConfig, localState *LocalState) error {
...
+			remoteStateChanged := make(chan remotestate.Snapshot)
+			done := make(chan struct{})
+			go func() {
+				var rs chan remotestate.Snapshot
+				for {
+					select {
+					case <-cfg.Watcher.RemoteStateChanged():
+						// We consumed a remote state change event
+						// so we need a way to trigger the select below
+						// in case it was a new operation.
```
wallyworld: Do we have a unit test for this?

state/action.go
hpidcock: yes but so you know @wallyworld missed ActionCancelled from list-operations :grinning:

worker/uniter/resolver/loop.go
hpidcock: LoopSuite.TestLoopWithChange

hpidcock: Tests both the refire logic and the remote state snapshot channel.

hpidcock:
!!build!! weird network issue

state/watcher.go
```diff
@@ -3218,14 +3218,152 @@ func ensureSuffixFn(marker string) func(string) string {
...
+}
+
+// actionNotificationWatcher is a StringsWatcher that watches for changes on the
+// action notification collection, but only triggers events once per action.
+type actionNotificationWatcher struct {
+	commonWatcher
+	source   chan watcher.Change
+	sink     chan []string
+	filter   func(interface{}) bool
+	fireOnce bool
```
wallyworld: to save the reader groking the code, the you add a doc line to explain what this is for

state/watcher.go
```diff
@@ -3218,14 +3218,152 @@ func ensureSuffixFn(marker string) func(string) string {
...
+		}
+	}
+}
+
+func (w *actionNotificationWatcher) initial() ([]string, error) {
+	var ids []string
+	var doc actionNotificationDoc
+	coll, closer := w.db.GetCollection(actionNotificationsC)
+	defer closer()
+	iter := coll.Find(nil).Iter()
```
wallyworld: this could get really messy if there's a large number of unpruned completed actions.
can we filter on pending or running...?

state/watcher.go
```diff
@@ -3218,14 +3218,152 @@ func ensureSuffixFn(marker string) func(string) string {
...
+	for iter.Next(&doc) {
+		if w.filter(doc.DocId) {
+			ids = append(ids, actionNotificationIdToActionId(doc.DocId))
+		}
+	}
+	return ids, iter.Close()
+}
+
+func (w *actionNotificationWatcher) filterAndMergeIds(changes *[]string, updates map[interface{}]bool, active map[string]struct{}) error {
```
wallyworld: let's add a comment to explain how this differs from mergeIds()

state/watcher.go
hpidcock: to preserve backwards compatibility, we need to fire for pending and running, at least once. If the uniter crashes it should get the running actions it had started so it can error them.
number of pending/running actions is unbounded, but I believe this being for a legacy facades only we can accept that risk.

A) We could fire all initial events, then subsequently only fire when the notify document is missing a change field, but that requires a read for each notification. This also changes the semantics of the fire once for each action. 

B) A way around this is two collections, one for legacy watchers (fires only on pending) and one the new facade. This seems just as heavy as the current solution.

I'll implement option A and change all the semantics of fireOnce to notifyNewPending. But this solution isn't as robust as the current map or option B, future changes to the notify collection will need to keep in mind legacy behaviour.

This would have been substantially easier if the mongo collection watchers send the new doc state in the Change value (or a way to read the document in the past with revno)

wallyworld: Pending and running yes. I was suggesting that'd we'd filter out anything else - completed, failed, aborted etc. I would not expect the number of pending/running to be too bad.

hpidcock:
$$merge$$

state/watcher.go
hpidcock: There is no notification doc for completed states AFAIK.

hpidcock:
$$merge$$

hpidcock:
$$merge$$ removed test logging

PR #11217, opened 2020-02-13
Improve MachineLXDProfileWatcher to reduce unnecessary notification.

hmlanigan:
### Checklist

----

## Description of change

If an application had been see, we added the unit as known, without checking that we'd see it before.  Notifying even when seen previously caused unnecessary notifications to the instance-mutater.

Added comments regarding publishing changes for subordinate unit changes.  Improvement can be done there as well.  This changes should help address the high number of calls to CharmProfileInfo from the instance-mutater seen in prod stack recently.

## QA steps

```sh
juju bootstrap localhost --config logging-config="<root>=DEBUG;unit=DEBUG;juju.worker.instancemutater=TRACE"
juju deploy nova-compute
juju deploy neutron-openvswitch
juju add-relation nova-compute neutron-openvswitch 
juju debug-log  --include machine-0 --include-module juju.worker.instancemutater
```
Check the logs, excessive output of "no changes necessary to machine-0" should not longer happen after the neutron-openvswitch unit is added.

### DIFF
--- a/core/cache/lxdprofilewatcher.go
+++ b/core/cache/lxdprofilewatcher.go
@@ -233,7 +233,7 @@ func (w *MachineLXDProfileWatcher) addUnit(_ string, value interface{}) {
 			return
 		}
 		if w.machineId != principal.MachineId() {
-			logger.Tracef("watching unit changes on machine-%s not machine-%s", w.machineId, unitMachineId)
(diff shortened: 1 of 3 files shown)


hmlanigan:
$$merge$$

PR #11220, opened 2020-02-14
Add support for juju list-operations

wallyworld:
### Checklist

New API but list-operation sis behind a jujuv3 feature flag.

----

## Description of change

Add support for a revamped list-operations CLI.
We now return operations results, which may each have matching tasks (actions). Previously we were returning a list of tasks (actions) directly. A bunch of old action specific query code in the facade is removed and replaced by a direct state call to fetch the operations.
For future use, add offset/limit params so the CLI can batch possibly large results.

## QA steps

Run some actions.
Run list-operations (with/without --format yaml) to see that the selected operations and any child tasks are returned.

### DIFF
--- a/cmd/juju/action/listoperations.go
+++ b/cmd/juju/action/listoperations.go
@@ -6,6 +6,7 @@ package action
 import (
 	"fmt"
 	"io"
+	"reflect"
 	"sort"
 	"strings"
 	"time"
@@ -13,12 +14,12 @@ import (
 	"github.com/juju/cmd"
 	"github.com/juju/errors"
 	"github.com/juju/gnuflag"
-	"github.com/juju/juju/apiserver/params"
-	"github.com/juju/juju/cmd/output"
 	"gopkg.in/juju/names.v3"
 
+	"github.com/juju/juju/apiserver/params"
 	jujucmd "github.com/juju/juju/cmd"
 	"github.com/juju/juju/cmd/modelcmd"
+	"github.com/juju/juju/cmd/output"
 )
 
 func NewListOperationsCommand() cmd.Command {
@@ -32,29 +33,27 @@ type listOperationsCommand struct {
 	utc              bool
 	applicationNames []string
 	unitNames        []string
-	functionNames    []string
+	actionNames      []string
 	statusValues     []string
 }
 
 const listOperationsDoc = `
 List the operations with the specified query criteria.
-With no query arguments, any completed operations will be listed.
-A completed operation is one that has run successfully, been cancelled, or failed.
-
 When an application is specified, all units from that application are relevant.
 
 Examples:
     juju operations
     juju operations --format yaml
-    juju operations --functions backup,restore
+    juju operations --actions backup,restore
     juju operations --apps mysql,mediawiki
     juju operations --units mysql/0,mediawiki/1
     juju operations --status pending,completed
-    juju operations --apps mysql --units mediawiki/0 --status running --functions backup
+    juju operations --apps mysql --units mediawiki/0 --status running --actions backup
 
 See also:
-    call
+    run
     show-operation
+    show-task
 `
 
 // Set up the output.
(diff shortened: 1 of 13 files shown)


wallyworld:
[list operations CLI](https://trello.com/c/XuJUhV2g/1043-list-operations-cli)

state/operation.go
```diff
@@ -269,3 +269,107 @@ func (m *Model) AllOperations() ([]Operation, error) {
...
+// ListOperations returns operations that match the specified criteria.
+func (m *Model) ListOperations(
+	actionNames []string, actionReceivers []names.Tag, operationStatus []ActionStatus,
+	offset, limit int,
+) ([]OperationSummary, bool, error) {
+	// First gather the matching actions and record the parent operation ids we need.
+	actionsCollection, closer := m.st.db().GetCollection(actionsC)
+	defer closer()
+
+	var receiverIds []string
```
hpidcock: IDs

state/operation.go
```diff
@@ -269,3 +269,107 @@ func (m *Model) AllOperations() ([]Operation, error) {
...
+	if len(actions) == 0 {
+		return nil, false, nil
+	}
+
+	// We have the operation ids which are parent to any actions matching the criteria.
+	// Combine these with additional operation filtering criteria to do the final query.
+	operationIds := make([]string, len(actions))
+	operationActions := make(map[string][]actionDoc)
+	for i, action := range actions {
+		operationIds[i] = m.st.docID(action.Operation)
```
hpidcock: or maybe this should be docId :laughing:

cmd/juju/action/listoperations.go
```diff
@@ -197,58 +200,142 @@ func operationDisplayTime(r params.ActionResult) time.Time {
...
+			singleAction = actionSummary{
+				Name:       task.Action.Name,
+				Parameters: make(map[string]interface{}),
+			}
+			for k, v := range task.Action.Parameters {
+				singleAction.Parameters[k] = v
+			}
+			continue
+		}
+		if haveSingleAction && (task.Action.Name != singleAction.Name || !reflect.DeepEqual(task.Action.Parameters, singleAction.Parameters)) {
```
hpidcock: break this condition out into a variable that has a name that explains this a little better.

cmd/juju/action/listoperations.go
```diff
@@ -46,15 +47,16 @@ When an application is specified, all units from that application are relevant.
...
     juju operations --units mysql/0,mediawiki/1
     juju operations --status pending,completed
-    juju operations --apps mysql --units mediawiki/0 --status running --functions backup
+    juju operations --apps mysql --units mediawiki/0 --status running --actions backup
 
 See also:
-    call
+    run
     show-operation
+    show-task
```
hpidcock: Does show-task exist yet?

cmd/juju/action/listoperations.go
```diff
@@ -177,16 +177,19 @@ func (c *listOperationsCommand) formatTabular(writer io.Writer, value interface{
...
-			w.Println(formatTimestamp(line.timestamp, false, c.utc, true))
+			w.Print(line.id, line.status)
+			w.Print(formatTimestamp(line.started, false, c.utc, true))
+			w.Print(formatTimestamp(line.finished, false, c.utc, true))
+			w.Print(strings.Join(line.tasks, ","))
+			w.Println(line.operation)
 		}
 	}
-	w.Println("Id", "Operation", "Status", "Unit", "Time")
+	w.Println("Id", "Status", "Started", "Finished", "Tasks", "Summary")
```
hpidcock: A better name for Tasks would be good. Task IDs perhaps is a better label. When I initially ran a single task, I got the number 3 which I thought was the number of tasks :thinking:

cmd/juju/action/listoperations.go
wallyworld: not yet, but it will

state/operation.go
```diff
@@ -269,3 +269,107 @@ func (m *Model) AllOperations() ([]Operation, error) {
...
+	// Don't let the user shoot themselves in the foot.
+	if limit == 0 {
+		limit = defaultMaxOperationsLimit
+	}
+	query = query.Limit(limit)
+	err = query.Sort("_id").All(&docs)
+	if err != nil {
+		return nil, false, errors.Trace(err)
+	}
+	truncated := nominalCount > len(docs)
```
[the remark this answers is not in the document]
wallyworld: Yeah, our find queries are done straight on the collection, no txn

cmd/juju/action/listoperations.go
wallyworld: added code comment as well

wallyworld:
$$merge$$

PR #11223, opened 2020-02-17
Protect store entity list with a mutex

wallyworld:
## Description of change

race tests were failing due to concurrent access to the store's entity list from an update event invoking Update() and generating an engine report invoking Report().
Use a mutex to protect the entity list.

## QA steps

go test --race


howbazaar:
The store itself as by design single threaded. We should fix this at the report stage, not internally.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #11228, opened 2020-02-18
backport pr 10969 JUJU_HOOK_NAME

hmlanigan:
Backport of #10969 

There are currently indirect mechanisms for a charm to identify the hook
name (argv[0] or JUJU_CONTEXT_ID which does not have a publicly
documented format).

This change adds it to the hook context and environment variables.

JUJU_HOOK_NAME will be set to a function name for functions.

See LP: #1854505


hmlanigan:
$$merge$$

PR #11237, opened 2020-02-20
[dev] introduce remove hook

achilleasa:
## Description of change

This PR introduces a new hook type called `remove` which runs immediately after the `stop` hook.

## QA steps

```console
$ juju bootstrap lxd test
$ juju deploy ./testcharms/charm-repo/bionic/dummy

# In another terminal
$ juju debug-log

# Once the dummy charm is alive run:
$ juju remove-application dummy

# Check the output log for:
unit-dummy-0: 16:35:00 DEBUG unit.dummy/0.remove Removing service files
unit-dummy-0: 16:35:00 INFO juju.worker.uniter.operation ran "remove" hook
```

## Documentation changes

We need to add a reference to the new hook type to our docs.

### DIFF
--- a/cmd/juju/commands/debughooks_test.go
+++ b/cmd/juju/commands/debughooks_test.go
@@ -121,7 +121,7 @@ var debugHooksTests = []struct {
 }, {
 	info:  `invalid hook`,
 	args:  []string{"mysql/0", "invalid-hook"},
-	error: `unit "mysql/0" contains neither hook nor action "invalid-hook", valid actions are [anotherfakeaction fakeaction] and valid hooks are [collect-metrics config-changed install juju-info-relation-broken juju-info-relation-changed juju-info-relation-departed juju-info-relation-joined leader-deposed leader-elected leader-settings-changed meter-status-changed metrics-client-relation-broken metrics-client-relation-changed metrics-client-relation-departed metrics-client-relation-joined post-series-upgrade pre-series-upgrade server-admin-relation-broken server-admin-relation-changed server-admin-relation-departed server-admin-relation-joined server-relation-broken server-relation-changed server-relation-departed server-relation-joined start stop update-status upgrade-charm]`,
+	error: `unit "mysql/0" contains neither hook nor action "invalid-hook", valid actions are [anotherfakeaction fakeaction] and valid hooks are [collect-metrics config-changed install juju-info-relation-broken juju-info-relation-changed juju-info-relation-departed juju-info-relation-joined leader-deposed leader-elected leader-settings-changed meter-status-changed metrics-client-relation-broken metrics-client-relation-changed metrics-client-relation-departed metrics-client-relation-joined post-series-upgrade pre-series-upgrade remove server-admin-relation-broken server-admin-relation-changed server-admin-relation-departed server-admin-relation-joined server-relation-broken server-relation-changed server-relation-departed server-relation-joined start stop update-status upgrade-charm]`,
 }, {
 	info:  `no args at all`,
 	args:  nil,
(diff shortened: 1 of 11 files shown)


cmd/juju/commands/debughooks_test.go
```diff
@@ -121,7 +121,7 @@ var debugHooksTests = []struct {
...
+	error: `unit "mysql/0" contains neither hook nor action "invalid-hook", valid actions are [anotherfakeaction fakeaction] and valid hooks are [collect-metrics config-changed install juju-info-relation-broken juju-info-relation-changed juju-info-relation-departed juju-info-relation-joined leader-deposed leader-elected leader-settings-changed meter-status-changed metrics-client-relation-broken metrics-client-relation-changed metrics-client-relation-departed metrics-client-relation-joined post-series-upgrade pre-series-upgrade remove server-admin-relation-broken server-admin-relation-changed server-admin-relation-departed server-admin-relation-joined server-relation-broken server-relation-changed server-relation-departed server-relation-joined start stop update-status upgrade-charm]`,
```
SimonRichardson: Nice error message :scream: 

We should clean this up at some point, as it's only going to get larger with time.

worker/uniter/resolver.go
```diff
@@ -61,7 +61,7 @@ func (s *uniterResolver) NextOp(
 	remoteState remotestate.Snapshot,
 	opFactory operation.Factory,
 ) (operation.Operation, error) {
-	if remoteState.Life == life.Dead || localState.Stopped {
+	if remoteState.Life == life.Dead || localState.Removed {
```
SimonRichardson: Is this correct, I don't know enough about the consequences to this...?

cmd/juju/commands/debughooks_test.go
achilleasa: I will fix this as a drive-by when I add the relation-created hook in an upcoming PR.

worker/uniter/resolver.go
achilleasa: Before this change, `stop` was the last hook that got executed before the uniter was signaled to die. So basically, once the unit is dying you would first hit [this](https://github.com/juju/juju/pull/11237/files/907b02253fbddceb29953517186e2da532ff73cd#diff-2e393918016123f29cf986dc718755d8R279) line which triggers the `stop` hook. Then you would enter `NextOp` because of [this](https://github.com/juju/juju/blob/907b02253fbddceb29953517186e2da532ff73cd/worker/uniter/operation/runhook.go#L257-L260) bit of logic (TLDR: asks uniter to figure out the next op) and exit with `ErrTerminate` to kill the uniter.

Now, the last hook in the chain is `remove` hence the need for the above change; otherwise the state machine would stop at `stop`.

testcharms/charm-repo/bionic/dummy/hooks/remove
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
hmlanigan: Let's add a stop hook as well, to ensure that stop gets called before remove.  I did so for my testing.

achilleasa:
@hmlanigan I believe this behavior is already indirectly checked by the existing tests (see my changes in the before/after tests). 

Note that the logic added [here](https://github.com/juju/juju/pull/11237/files/907b02253fbddceb29953517186e2da532ff73cd#diff-2e393918016123f29cf986dc718755d8R279-R283) ensures that you can only run `remove` after a `close` and the change at the top of that method (see Simon's comment above) ensures that you don't return `ErrTerminate` until the `remove` completes. Therefore, any test that expects a clean teardown of the uniter would fail unless the hooks where invoked in the expected order (`close`, then `remove`). 

If you think that this is not sufficient, let me know and I can add another test tomorrow.

testcharms/charm-repo/bionic/dummy/hooks/remove
achilleasa: Added

achilleasa:
$$merge$$

PR #11244, opened 2020-02-24
Add annotations to core multiwatcher types.

howbazaar:
This is preparation for including annotations in the model summary watcher.

Annotations in the allwatcher code right now were incorrectly done, but we need to keep API compatibility. This branch adds annotations to the core/multiwatcher types for Model, Machine, Unit, and Application.

The model annotations are going to be used by the dashboard to store model oriented annotations, and these are going to be needed to be returned through the summary structure. This will be updated to include the annotations, which will need another branch to add the annotations to the model cache.

## QA steps

Just the tests for now.


howbazaar:
$$merge$$

PR #11248, opened 2020-02-25
Add annotations to the cache for objects that have annotations.

howbazaar:
PR #11244 added annotations to the multiwatcher types. This PR threads those annotations through to the cache.

Additionally we add model annotations to the model summaries so model annotations can be used to mute models for errors in the dashboard.

Facade schema is updated, and a current PR on pylibjuju will be updated to use it once this is landed.

## QA steps

```sh
cd core/cache
go test -check.v
cd ../../workers/modelcache
go test -check.v
```


core/cache/modelsummarywatcher_test.go
```diff
@@ -221,6 +221,44 @@ func (s *modelSummaryWatcherSuite) TestRemoveModelRemovesModel(c *gc.C) {
...
+		Life:      life.Alive,
+		Owner:     "bob",
+		UserPermissions: map[string]permission.Access{
+			"bob":  permission.AdminAccess,
+			"mary": permission.ReadAccess,
+		},
+		Annotations: map[string]string{
+			"muted": "true",
+		},
+	}, s.events)
```
jameinel: This test doesn't clearly indicate that only the Annotations are different between before and after.

howbazaar: Made more clear.

howbazaar:
$$merge$$

PR #11259, opened 2020-02-28
Fix/caas model teardown

ycliuhw:
### Checklist

 - [x] ~Checked if it requires a [pylibjuju](https://github.com/juju/python-libjuju) change?~

----

## Description of change

Destroying CAAS model now ensures cluster scope resources get deleted as well;

Driveby fix: 
- remove the `Optional` field from `volumeConfig` items;
- In k8s provider baseSuite, wacher slices wasn't properly cleaned up for each test;

## QA steps

- deploy CAAS app with CRDs, MutatingWebhookConfiguration etc cluster scope resources;
- destroy the model;
- kubectl checks those resources are gone;

## Documentation changes

No

## Bug reference

https://bugs.launchpad.net/juju/+bug/1862426

### DIFF
--- a/caas/kubernetes/provider/admissionregistration.go
+++ b/caas/kubernetes/provider/admissionregistration.go
@@ -123,16 +123,20 @@ func (k *kubernetesClient) listMutatingWebhookConfigurations(labels map[string]s
 	return cfgList.Items, nil
 }
 
-func (k *kubernetesClient) deleteMutatingWebhookConfigurations(appName string) error {
+func (k *kubernetesClient) deleteMutatingWebhookConfigurations(labels map[string]string) error {
 	err := k.client().AdmissionregistrationV1beta1().MutatingWebhookConfigurations().DeleteCollection(&metav1.DeleteOptions{
 		PropagationPolicy: &defaultPropagationPolicy,
 	}, metav1.ListOptions{
-		LabelSelector: labelsToSelector(k.getAdmissionControllerLabels(appName)),
+		LabelSelector: labelsToSelector(labels),
 	})
 	if k8serrors.IsNotFound(err) {
 		return nil
 	}
 	return errors.Trace(err)
+}
+
+func (k *kubernetesClient) deleteMutatingWebhookConfigurationsForApp(appName string) error {
+	return errors.Trace(k.deleteMutatingWebhookConfigurations(k.getAdmissionControllerLabels(appName)))
 }
 
 func (k *kubernetesClient) ensureValidatingWebhookConfigurations(
@@ -236,14 +240,18 @@ func (k *kubernetesClient) listValidatingWebhookConfigurations(labels map[string
 	return cfgList.Items, nil
 }
 
-func (k *kubernetesClient) deleteValidatingWebhookConfigurations(appName string) error {
+func (k *kubernetesClient) deleteValidatingWebhookConfigurations(labels map[string]string) error {
 	err := k.client().AdmissionregistrationV1beta1().ValidatingWebhookConfigurations().DeleteCollection(&metav1.DeleteOptions{
 		PropagationPolicy: &defaultPropagationPolicy,
 	}, metav1.ListOptions{
-		LabelSelector: labelsToSelector(k.getAdmissionControllerLabels(appName)),
+		LabelSelector: labelsToSelector(labels),
 	})
 	if k8serrors.IsNotFound(err) {
 		return nil
 	}
 	return errors.Trace(err)
+}
+
+func (k *kubernetesClient) deleteValidatingWebhookConfigurationsForApp(appName string) error {
+	return errors.Trace(k.deleteValidatingWebhookConfigurations(k.getAdmissionControllerLabels(appName)))
 }
--- a/caas/kubernetes/provider/customresourcedefinitions.go
+++ b/caas/kubernetes/provider/customresourcedefinitions.go
@@ -154,7 +154,7 @@ func (k *kubernetesClient) deleteCustomResources(labelsGetter func(apiextensions
 	}
 	for _, crd := range crds.Items {
 		labels := labelsGetter(crd)
-		if labels == nil {
+		if len(labels) == 0 {
 			continue
 		}
 		for _, version := range crd.Spec.Versions {
@@ -173,6 +173,38 @@ func (k *kubernetesClient) deleteCustomResources(labelsGetter func(apiextensions
 		}
 	}
 	return nil
+}
+
+func (k *kubernetesClient) listCustomResources(labelsGetter func(apiextensionsv1beta1.CustomResourceDefinition) map[string]string) (out []unstructured.Unstructured, err error) {
+	crds, err := k.extendedCient().ApiextensionsV1beta1().CustomResourceDefinitions().List(v1.ListOptions{
+		// CRDs might be provisioned by another application/charm from a different model.
+	})
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	for _, crd := range crds.Items {
+		labels := labelsGetter(crd)
+		if labels == nil {
+			continue
+		}
+		for _, version := range crd.Spec.Versions {
+			crdClient, err := k.getCustomResourceDefinitionClient(&crd, version.Name)
+			if err != nil {
+				return nil, errors.Trace(err)
+			}
+			list, err := crdClient.List(v1.ListOptions{
+				LabelSelector: labelsToSelector(labels),
+			})
+			if err != nil && !k8serrors.IsNotFound(err) {
+				return nil, errors.Trace(err)
+			}
+			out = append(out, list.Items...)
+		}
+	}
+	if len(out) == 0 {
+		return nil, errors.NewNotFound(nil, "no customer resource found")
+	}
+	return out, nil
 }
 
 type apiVersionGetter interface {
--- a/caas/kubernetes/provider/customresourcedefinitions.go
+++ b/caas/kubernetes/provider/customresourcedefinitions.go
@@ -135,42 +135,29 @@ func (k *kubernetesClient) deleteCustomResourceDefinitions(labels map[string]str
 }
 
 func (k *kubernetesClient) deleteCustomResourcesForApp(appName string) error {
-	crds, err := k.extendedCient().ApiextensionsV1beta1().CustomResourceDefinitions().List(v1.ListOptions{})
-	if err != nil {
-		return errors.Trace(err)
-	}
-	for _, crd := range crds.Items {
-		for _, version := range crd.Spec.Versions {
-			crdClient, err := k.getCustomResourceDefinitionClient(&crd, version.Name)
-			if err != nil {
-				return errors.Trace(err)
-			}
-			labels := k.getCRDLabelsGlobal(appName)
-			if isCRDNameSpaced(crd.Spec.Scope) {
-				labels = k.getCRDLabelsNamespaced(appName)
-			}
-			err = crdClient.DeleteCollection(&v1.DeleteOptions{
-				PropagationPolicy: &defaultPropagationPolicy,
-			}, v1.ListOptions{
-				LabelSelector: labelsToSelector(labels),
-			})
-			if err != nil && !k8serrors.IsNotFound(err) {
-				return errors.Trace(err)
-			}
+	labelsGetter := func(crd apiextensionsv1beta1.CustomResourceDefinition) map[string]string {
+		labels := k.getCRDLabelsGlobal(appName)
+		if isCRDNameSpaced(crd.Spec.Scope) {
+			labels = k.getCRDLabelsNamespaced(appName)
 		}
+		logger.Criticalf("labels %+v", labels)
+		return labels
 	}
-	return nil
+	return k.deleteCustomResources(labelsGetter)
 }
 
-func (k *kubernetesClient) deleteGlobalCustomResources(labels map[string]string) error {
-	selector := labelsToSelector(labels)
+func (k *kubernetesClient) deleteCustomResources(labelsGetter func(apiextensionsv1beta1.CustomResourceDefinition) map[string]string) error {
 	crds, err := k.extendedCient().ApiextensionsV1beta1().CustomResourceDefinitions().List(v1.ListOptions{
-		LabelSelector: selector,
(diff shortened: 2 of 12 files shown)


ycliuhw:
[Ensure k8s global resource(crds etc) removed when model is removed](https://trello.com/c/3gkVDVI2/1044-ensure-k8s-global-resourcecrds-etc-removed-when-model-is-removed)

caas/kubernetes/provider/k8s.go
```diff
@@ -789,18 +782,18 @@ func (k *kubernetesClient) DeleteService(appName string) (err error) {
 	if err := k.deleteAllServiceAccountResources(appName); err != nil {
 		return errors.Trace(err)
 	}
-	// Order matters: delete custom resources first then custom resource definitions.
-	if err := k.deleteCustomResources(appName); err != nil {
-		return errors.Trace(err)
-	}
-	if err := k.deleteCustomResourceDefinitions(appName); err != nil {
+	// // Order matters: delete custom resources first then custom resource definitions.
```
ycliuhw: TODO: needs double check

caas/kubernetes/provider/teardown_test.go
```diff
@@ -0,0 +1,593 @@
...
+	storagev1 "k8s.io/api/storage/v1"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/juju/juju/testing"
+)
+
+func (s *K8sBrokerSuite) TestdeleteClusterScropeResourcesModelTeardown(c *gc.C) {
```
wallyworld: TestDelete...

caas/kubernetes/provider/teardown_test.go
```diff
@@ -0,0 +1,593 @@
...
+	err := s.clock.WaitAdvance(time.Second, testing.ShortWait, 7)
+	c.Assert(err, jc.ErrorIsNil)
+	select {
+	case <-done:
+	case <-time.After(testing.LongWait):
+		c.Fatalf("timed out waiting for DeleteClusterScropeResourcesModelTeardown return")
+	}
+}
+
+func (s *K8sBrokerSuite) TestdeleteClusterScropeResourcesModelTeardownTimeout(c *gc.C) {
```
wallyworld: TestDelete...

caas/kubernetes/provider/customresourcedefinitions.go
```diff
@@ -26,14 +26,14 @@ import (
...
-func (k *kubernetesClient) getCRDLabels(appName string) map[string]string {
+func (k *kubernetesClient) getCRDLabelsGlobal(appName string) map[string]string {
 	return map[string]string{
 		labelApplication: appName,
 		labelModel:       k.namespace,
 	}
 }
 
-func (k *kubernetesClient) getCRLabels(appName string) map[string]string {
+func (k *kubernetesClient) getCRDLabelsNamespaced(appName string) map[string]string {
```
wallyworld: These seem to be the wrong way around?
ie getCRDLabelsNamespaced() should contain the model/namespace?

Ah, I think I understand the use case now. But the method names are confusing.
Especially since we use getCRDLabelsGlobal() for CRDs regardless of scope.
We only use these 2 methods for CRs.

Would it be easier simply to always add the app name and mode name labels to all CRs regardless of scope?

caas/kubernetes/provider/customresourcedefinitions.go
```diff
@@ -104,24 +104,59 @@ func (k *kubernetesClient) getCustomResourceDefinition(name string) (*apiextensi
...
+			labels = k.getCRDLabelsNamespaced(appName)
+		}
+		return labels
+	}
+	return k.deleteCustomResources(labelsGetter)
+}
+
+func (k *kubernetesClient) deleteCustomResources(labelsGetter func(apiextensionsv1beta1.CustomResourceDefinition) map[string]string) error {
+	crds, err := k.extendedCient().ApiextensionsV1beta1().CustomResourceDefinitions().List(v1.ListOptions{
+		// CRDs might be provisioned by another application/charm from a different model.
```
wallyworld: can we filter here on anything?
seems not optimal to load everything?

caas/kubernetes/provider/teardown.go
```diff
@@ -0,0 +1,263 @@
...
+	}()
+
+	if err = deleter(labels); err != nil {
+		if errors.IsNotFound(err) {
+			err = nil
+		}
+		return
+	}
+
+	interval := 1 * time.Second
```
wallyworld: not sure if this is a const somewhere(ie we use a similar polling pattern elsewhere and we can use a const to define the polling interval)

caas/kubernetes/provider/teardown.go
```diff
@@ -0,0 +1,263 @@
...
+				err = nil
+				return
+			}
+			if err != nil {
+				err = errors.Trace(err)
+				return
+			}
+		}
+		// Keep checking.
+		ticker.Reset(interval)
```
wallyworld: I guess we don't use a timeout elsewhere in destroy-model
Not sure if we need one. Maybe useful to timeout after 10 or 15 minutes. The user can always be prompted to retry.
Not sure....

caas/kubernetes/provider/teardown.go
```diff
@@ -0,0 +1,263 @@
...
+			if errors.IsNotFound(err) {
+				// Namespace has been deleted.
+				err = nil
+				return
+			}
+			if err != nil {
+				err = errors.Trace(err)
+				return
+			}
+			logger.Debugf("namespace %q is still been terminating", k.namespace)
```
wallyworld: typo

caas/kubernetes/provider/export_test.go
```diff
@@ -134,6 +137,14 @@ func (k *kubernetesClient) ConfigurePodFiles(
 	return k.configurePodFiles(appName, annotations, workloadSpec, containers, cfgMapName)
 }
 
+func (k *kubernetesClient) DeleteClusterScropeResourcesModelTeardown(ctx context.Context, wg *sync.WaitGroup, errChan chan<- error) {
```
wallyworld: typo

caas/kubernetes/provider/customresourcedefinitions.go
```diff
@@ -50,7 +50,7 @@ func (k *kubernetesClient) ensureCustomResourceDefinitions(
 			ObjectMeta: v1.ObjectMeta{
 				Name:        name,
 				Namespace:   k.namespace,
-				Labels:      k.getCRDLabels(appName),
+				Labels:      k.getCRDLabelsGlobal(appName),
```
wallyworld: Why do we always use getCRDLabelsGlobal and ignore the scope?

caas/kubernetes/provider/customresourcedefinitions.go
```diff
@@ -104,24 +104,59 @@ func (k *kubernetesClient) getCustomResourceDefinition(name string) (*apiextensi
...
 	return errors.Trace(err)
 }
 
-func (k *kubernetesClient) deleteCustomResources(appName string) error {
-	crds, err := k.extendedCient().ApiextensionsV1beta1().CustomResourceDefinitions().List(v1.ListOptions{})
+func (k *kubernetesClient) deleteCustomResourcesForApp(appName string) error {
+	labelsGetter := func(crd apiextensionsv1beta1.CustomResourceDefinition) map[string]string {
+		labels := k.getCRDLabelsGlobal(appName)
+		if isCRDNameSpaced(crd.Spec.Scope) {
+			labels = k.getCRDLabelsNamespaced(appName)
```
wallyworld: labels in this case will just contain the app name
won't that mean that we also select and delete all the global custom resources as well?

caas/kubernetes/provider/customresourcedefinitions.go
```diff
@@ -170,7 +237,11 @@ func (k *kubernetesClient) ensureCustomResources(
 			if err != nil {
 				return cleanUps, errors.Trace(err)
 			}
-			crSpec.SetLabels(k.getCRLabels(appName))
+			labels := k.getCRDLabelsGlobal(appName)
+			if isCRDNameSpaced(crd.Spec.Scope) {
+				labels = k.getCRDLabelsNamespaced(appName)
```
wallyworld: Confusing that we use getCRDLabels for a CR

caas/kubernetes/provider/teardown.go
ycliuhw: Undertaker worker could pass timeout via dying chan then Destroy() will cancel the context.
Finally all these goroutines get aborted.

caas/kubernetes/provider/customresourcedefinitions.go
ycliuhw: Here we just keep the labels are consistent for all global resources.
Like cluster scope RBAC resources, we have appName and model-uuid in labels,
But all the namespaced resources don't have model-uuid in labels.

caas/kubernetes/provider/customresourcedefinitions.go
ycliuhw: As we discussed yesterday, CRDs might be created by another charm even from a different model.
So here we have to consider all CRDs.

caas/kubernetes/provider/teardown.go
tlm: Also 1 * time.Second == time.Second

caas/kubernetes/provider/teardown.go
tlm: Would it be better to use a timed context here over the ticker?

caas/kubernetes/provider/customresourcedefinitions.go
ycliuhw: probabaly the naming of the func is confusing.
I have renamed it to `getAPIExtensionLabelsNamespaced` because it's used for CRD and CR(CR could be cluster/namespaced scope)

caas/kubernetes/provider/customresourcedefinitions.go
```diff
@@ -26,14 +26,14 @@ import (
 
 //go:generate mockgen -package mocks -destination mocks/crd_getter_mock.go github.com/juju/juju/caas/kubernetes/provider CRDGetterInterface
 
-func (k *kubernetesClient) getCRDLabels(appName string) map[string]string {
+func (k *kubernetesClient) getCRDLabelsGlobal(appName string) map[string]string {
```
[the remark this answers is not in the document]
ycliuhw: We have been put lots of effects to make the label clean enough rather than put anything possible for any resources.
For example, for namespaces resources, they are already in a specific namespace. 
So there is no need to put model-uuid/model-name stuff into the labels.
This is the general rule for labeling on any already supported resources by Juju.

caas/kubernetes/provider/teardown.go
```diff
@@ -0,0 +1,263 @@
+// Copyright 2020 Canonical Ltd.
```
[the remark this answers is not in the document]
ycliuhw: I would prefer the current way.
The teardown.go isn't too big now. And all the funcs here are just splitted from `Destroy()` but not something like helper funcs could be reused for other places.

ycliuhw:
I think the main concern here is the labeling stuff and there might be some misunderstandings between CRDs and CRs.
It would be easier to discuss in HO further rather than raising the same question back and forth here.

caas/kubernetes/provider/customresourcedefinitions.go
tlm: @ycliuhw I think this comment may still need to be addressed with trying to isolate this logic to one place? Could be wrong.

caas/kubernetes/provider/admissionregistration.go
```diff
@@ -123,18 +123,22 @@ func (k *kubernetesClient) listMutatingWebhookConfigurations(labels map[string]s
 	return cfgList.Items, nil
 }
 
-func (k *kubernetesClient) deleteMutatingWebhookConfigurations(appName string) error {
+func (k *kubernetesClient) deleteMutatingWebhookConfigurations(labels map[string]string) error {
```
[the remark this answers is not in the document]
tlm: Still leaving this as a suggestion as I think it needs to be done but won't hold the PR because of it.

caas/kubernetes/provider/customresourcedefinitions.go
```diff
@@ -362,7 +438,7 @@ func (k *kubernetesClient) getCustomResourceDefinitionClient(crd *apiextensionsv
 			Resource: crd.Spec.Names.Plural,
 		},
 	)
-	if crd.Spec.Scope != apiextensionsv1beta1.NamespaceScoped {
+	if !isCRDNameSpaced(crd.Spec.Scope) {
```
tlm: s/isCRDNameSpaced/isCRDNamespaced/g

caas/kubernetes/provider/admissionregistration.go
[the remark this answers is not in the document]
ycliuhw: As discussed, this is the current generic pattern in many places, we will change them all if we need to in the future.

caas/kubernetes/provider/teardown.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/juju/errors"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+
+	"github.com/juju/juju/core/watcher"
+)
+
+func (k *kubernetesClient) deleteClusterScropeResourcesModelTeardown(ctx context.Context, wg *sync.WaitGroup, errChan chan<- error) {
+	defer wg.Done()
+
+	labels := map[string]string{
```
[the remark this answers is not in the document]
ycliuhw: don't see any benefit to having a separate function for this since nowhere else use it.

caas/kubernetes/provider/admissionregistration.go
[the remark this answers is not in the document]
ycliuhw: As discussed, this is the current generic pattern in many places, we will change them all if we need to in the future.

ycliuhw:
$$merge$$

PR #11262, opened 2020-02-29
Fix FullStatus extra queries on spaces on link layer devices

howbazaar:
During status the code that gets the space name for the link layer device was causing extra queries to the database.

In order to test, John and I introduce a query tracker on the state object to cound the number of database read queries.

### DIFF
--- a/apiserver/facades/client/client/status_test.go
+++ b/apiserver/facades/client/client/status_test.go
@@ -146,7 +146,9 @@ func (s *statusSuite) TestFullStatusInterfaceScaling(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 
 	// The number of queries should be the same.
-	c.Check(tracker.ReadCount(), gc.Equals, queryCount)
+	c.Check(tracker.ReadCount(), gc.Equals, queryCount,
+		gc.Commentf("if the query count is not the same, there has been a regression "+
+			"in the way the addresses are processed"))
 }
 
 func (s *statusSuite) createSpaceAndSubnetWithProviderID(c *gc.C, spaceName, CIDR, providerSubnetID string) {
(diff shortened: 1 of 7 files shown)


apiserver/facades/client/client/status_test.go
```diff
@@ -101,6 +101,86 @@ func (s *statusSuite) TestFullStatusUnitLeadership(c *gc.C) {
...
+	c.Assert(err, jc.ErrorIsNil)
+
+	s.WaitForModelWatchersIdle(c, s.State.ModelUUID())
+	tracker.Reset()
+
+	_, err = client.Status(nil)
+	c.Assert(err, jc.ErrorIsNil)
+
+	// The number of queries should be the same.
+	c.Check(tracker.ReadCount(), gc.Equals, queryCount)
```
jameinel: Would it be worth using gc.Commentf to give context about what queries were done in the case that queryCount doesn't match? Ideally you would have a delta of the previous queries vs the current queries. I guess you can do that when you actually need to debug the failure.

state/querytracker.go
```diff
@@ -0,0 +1,82 @@
...
+
+import "runtime/debug"
+
+// QueryTracker provides a way for tests to determine how many
+// database queries have been made, and who made them.
+type QueryTracker interface {
+	Reset()
+	ListQueries() []QueryDetails
+	ReadCount() int
+	WriteCount() int
```
jameinel: we don't currently make use of WriteCount, is it worth postponing it?

state/querytracker.go
```diff
@@ -0,0 +1,82 @@
...
+			count++
+		}
+	}
+	return count
+}
+
+// TrackRead records the read query against the collection specified
+// and where the call came from.
+func (q *queryTracker) TrackRead(collectionName string, query interface{}) {
+	q.queries = append(q.queries, QueryDetails{
```
jameinel: We are missing direct unit tests of QueryTracker itself, but given it is testing infrastructure, I could be convinced of that.

howbazaar:
$$merge$$

PR #11265, opened 2020-03-02
Fixes lp#1860665: when no controllers are registered on the client, assume client-only operation.

anastasiamac:
## Description of change

Commands in cmd/juju/cloud package can operate on both client and controller copies of clouds/credentials/regions/etc. When there is an ambiguity where to operate, a command will prompt the operator for confirmation. However, when there are no controllers registered on the client, the operation will only operate using client copy. 

## QA steps

Ensure there are no controllers registered on your client, run any of the commands with *-cloud or *-credential

## Bug reference

https://bugs.launchpad.net/juju/+bug/1860665

### DIFF
--- a/cmd/juju/cloud/list_test.go
+++ b/cmd/juju/cloud/list_test.go
@@ -110,6 +110,43 @@ beehive:
 `[1:])
 }
 
+func (s *listSuite) TestListClientAndController(c *gc.C) {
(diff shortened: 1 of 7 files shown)


anastasiamac:
$$merge$$

anastasiamac:
$$merge$$

PR #11266, opened 2020-03-02
Fixes regression in 'clouds'.

anastasiamac:
## Description of change

Revert code that introduced a regression where the user is prompted to clarify whether client or controller copies of the clouds are to be listed.

As a read-only command, 'clouds' will always try to list both client and controller copies, if current controller is found.

## QA steps

1. run 'juju clouds' [should not be a prompt]
2. bootstrap a controller
3. run 'juju clouds' [should not be a prompt] 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1865527


anastasiamac:
@SimonRichardson Yes! Let's work on CI tomorrow :)

anastasiamac:
$$merge$$

PR #11273, opened 2020-03-03
Synchronise while collecting cache metrics.

howbazaar:
It has been observed in the wild that the metrics that were being
collected from the model cache was not consistent. The collector
resets the guages and then interatively increments the counters
based on the life and status of the values. If we don't protect
against concurrent collect calls, overlapping calls will cause
under counting in one return and over counting in the other.

## QA steps

```sh
cd core/cache
go test
```

### DIFF
--- a/core/cache/metrics_test.go
+++ b/core/cache/metrics_test.go
@@ -4,9 +4,12 @@ package cache_test
 
 import (
 	"bytes"
+	"fmt"
+	"sync"
 
 	"github.com/juju/loggo"
 	jc "github.com/juju/testing/checkers"
+	"github.com/juju/utils"
 	"github.com/prometheus/client_golang/prometheus/testutil"
 	gc "gopkg.in/check.v1"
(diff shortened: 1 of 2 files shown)


core/cache/metrics_test.go
```diff
@@ -57,3 +60,43 @@ juju_cache_units{agent_status="active",life="alive",workload_status="active"} 1
 
 	workertest.CleanKill(c, controller)
 }
+
+func (s *ControllerSuite) TestCollectIsolation(c *gc.C) {
+	controller, events := s.new(c)
+
+	// Populate the cache with 100 models so the collect takes
+	// more time.
+	for i := 0; i < 10; i++ {
```
wallyworld: 10 or 100?

howbazaar:
$$merge$$

core/cache/metrics_test.go
howbazaar: updated

PR #11281, opened 2020-03-04
Main Juju command stderr changes.

anastasiamac:
## Description of change

Information messages should be on stderr and they should react to -v and -q.

This change ensures that main Juju commands output is on the right writer and is responsive to -v and -1 flag by using centrally provided functionality in ctx.Infof(). 

## QA steps

Run 'juju -h' on a new machine.


anastasiamac:
I have a better way to fix this issue. I'll work on it today.

PR #11284, opened 2020-03-04
Standardise action commands.

anastasiamac:
## Description of change

Change all direct writes to ctx.Stderr to go through ctx.Infof() and put data on ctx.Stdout.


anastasiamac:
unit test on win did not run

anastasiamac:
$$merge$$

PR #11290, opened 2020-03-05
Change introduction messages to react to --quiet.

anastasiamac:
## Description of change

On a new install of Juju on a machine or a new install of Juju 2 on a machine that has Juju 1, we are helpful printing guidance and hints as well as report updating public clouds.

However, under some circumstances, the operator might want not to see this messages. This PR ensures that with "-q" or "--quiet" option, these messages are not displayed.
 
## QA steps

On a new machine, run 
```
juju version
```
or 
```
juju version -q
```
to see the messages display on or off.

(I've used different directories as juju home to exhibit this behavior)

### DIFF
--- a/cmd/juju/cloud/updatepublicclouds.go
+++ b/cmd/juju/cloud/updatepublicclouds.go
@@ -92,7 +92,10 @@ type PublicCloudsAccessDetails struct {
 // PublicCloudsAccess contains information about
 // where to find published public clouds details.
 func PublicCloudsAccess() PublicCloudsAccessDetails {
-	return PublicCloudsAccessDetails{keys.JujuPublicKey, "https://streams.canonical.com/juju/public-clouds.syaml"}
+	return PublicCloudsAccessDetails{
+		publicSigningKey: keys.JujuPublicKey,
+		publicCloudURL:   "https://streams.canonical.com/juju/public-clouds.syaml",
+	}
 }
 
 // Init populates the command with the args from the command line.
@@ -139,9 +142,17 @@ func (c *updatePublicCloudsCommand) Run(ctxt *cmd.Context) error {
 	fmt.Fprint(ctxt.Stderr, "Fetching latest public cloud list...\n")
 	var returnedErr error
 	publishedClouds, msg, err := FetchAndMaybeUpdatePublicClouds(
-		PublicCloudsAccessDetails{c.publicSigningKey, c.publicCloudURL},
+		PublicCloudsAccessDetails{
+			publicSigningKey: c.publicSigningKey,
+			publicCloudURL:   c.publicCloudURL,
+		},
 		c.Client)
 	if err != nil {
+		// Since FetchAndMaybeUpdatePublicClouds retrieves public clouds
+		// as well as updates client copy of the clouds, it is
+		// possible that the returned error is related to clouds retrieval.
+		// If there are no public clouds returned, we can assume that the
+		// retrieval itself was unsuccessful and abort further processing.
 		if len(publishedClouds) == 0 {
 			return errors.Trace(err)
 		}
@@ -164,8 +175,11 @@ func (c *updatePublicCloudsCommand) Run(ctxt *cmd.Context) error {
 // and updates client copy of public clouds if desired.
 // This call returns discovered public clouds and a user-facing message
 // whether they are different with what was known prior to the call.
+// Since this call can also update a client copy of clouds, it is possible that the public
+// clouds have been retrieved but the client update fail. In this case, we still
+// return public clouds as well as the client error.
 var FetchAndMaybeUpdatePublicClouds = func(access PublicCloudsAccessDetails, updateClient bool) (map[string]jujucloud.Cloud, string, error) {
-	msg := ""
+	var msg string
 	publishedClouds, err := PublishedPublicClouds(access.publicCloudURL, access.publicSigningKey)
 	if err != nil {
 		return nil, msg, errors.Trace(err)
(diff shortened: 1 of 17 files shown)


cmd/juju/cloud/updatepublicclouds.go
```diff
@@ -83,6 +84,17 @@ func (c *updatePublicCloudsCommand) Info() *cmd.Info {
...
 
+type PublicCloudsAccessDetails struct {
+	publicSigningKey string
+	publicCloudURL   string
+}
+
+// PublicCloudsAccess contains information about
+// where to find published public clouds details.
+func PublicCloudsAccess() PublicCloudsAccessDetails {
+	return PublicCloudsAccessDetails{keys.JujuPublicKey, "https://streams.canonical.com/juju/public-clouds.syaml"}
```
SimonRichardson: nit: we should use named fields.
```suggestion
	return PublicCloudsAccessDetails{
		publicSigningKey: keys.JujuPublicKey, 
		publicCloudURL: "https://streams.canonical.com/juju/public-clouds.syaml",
	}
```

cmd/juju/cloud/updatepublicclouds.go
```diff
@@ -125,16 +137,19 @@ func (c *updatePublicCloudsCommand) Run(ctxt *cmd.Context) error {
...
-	if err != nil {
-		return errors.Trace(err)
-	}
 	var returnedErr error
-	if c.Client {
-		if err := c.updateClientCopy(ctxt, publishedClouds); err != nil {
-			ctxt.Infof("ERROR %v", err)
-			returnedErr = cmd.ErrSilent
+	publishedClouds, msg, err := FetchAndMaybeUpdatePublicClouds(
+		PublicCloudsAccessDetails{c.publicSigningKey, c.publicCloudURL},
```
SimonRichardson:
```suggestion
		PublicCloudsAccessDetails{
				publicSigningKey: c.publicSigningKey, 
				publicCloudURL: c.publicCloudURL,
		},
```

cmd/juju/cloud/updatepublicclouds.go
```diff
@@ -125,16 +137,19 @@ func (c *updatePublicCloudsCommand) Run(ctxt *cmd.Context) error {
...
+		c.Client)
+	if err != nil {
+		if len(publishedClouds) == 0 {
```
SimonRichardson: I'm unsure about this, as normally we always `return nil` for the published clouds, although in this instance I'm actually unsure the path forward. Maybe add a comment stating the fact?

cmd/juju/cloud/updatepublicclouds.go
```diff
@@ -145,26 +160,42 @@ func (c *updatePublicCloudsCommand) Run(ctxt *cmd.Context) error {
 	return returnedErr
 }
 
-func (c *updatePublicCloudsCommand) updateClientCopy(ctxt *cmd.Context, publishedClouds map[string]jujucloud.Cloud) error {
+// FetchAndMaybeUpdatePublicClouds gets published public clouds information
+// and updates client copy of public clouds if desired.
+// This call returns discovered public clouds and a user-facing message
+// whether they are different with what was known prior to the call.
+var FetchAndMaybeUpdatePublicClouds = func(access PublicCloudsAccessDetails, updateClient bool) (map[string]jujucloud.Cloud, string, error) {
+	msg := ""
```
SimonRichardson: nit: 
```suggestion
	var msg string
```

cmd/juju/commands/main.go
```diff
@@ -155,15 +157,17 @@ func installProxy() error {
 	return nil
 }
 
-func (m main) maybeWarnJuju1x() (newInstall bool) {
+func (m main) maybeWarnJuju1x() (newInstall bool, jujuMsg string) {
+	jujuMsg = ""
```
SimonRichardson: nit:
```suggestion
	var jujuMsg string
```

cmd/juju/commands/main_test.go
```diff
@@ -246,7 +247,7 @@ Welcome to Juju %s.
 If you want to use Juju 1.25.0, run 'juju' commands as 'juju-1'. For example, 'juju-1 bootstrap'.
    See https://jujucharms.com/docs/stable/juju-coexist for installation details. 
 
-Since Juju 2 is being run for the first time, downloading latest cloud information.`[1:]+"\n", jujuversion.Current))
+Since Juju 2 is being run for the first time, downloaded latest public cloud information.`[1:]+"\n", jujuversion.Current))
```
SimonRichardson:
```suggestion
Since Juju 2 is being run for the first time, downloaded the latest public cloud information.`[1:]+"\n", jujuversion.Current))
```

cmd/juju/cloud/updatepublicclouds.go
anastasiamac: I am happy to update it but this code styling is a bit heavy. For example, GoLand implies the keys anyway.

cmd/juju/commands/main.go
anastasiamac: Actually jujuMsg is already declared as a return var, so there is no need to re-declare it.

cmd/juju/cloud/updatepublicclouds.go
SimonRichardson: I believe we should, there are downsides to not actually knowing what fields we're using whilst reading the code.

It's one of our aims to add these at some point.

cmd/juju/cloud/updatepublicclouds.go
anastasiamac: Added comment. Essentially since the Fetch may also update client copy, the error may be either with the retrieval of the public clouds or with the client copy update. If we got an error and no clouds, we know that the error is with the retrieval which means that there is no point trying to update a controller.

anastasiamac:
$$merge$$

PR #11296, opened 2020-03-07
Fix 1860083 suitable availability zone for machine not found

hpidcock:
## Description of change

Minor refactor to cleanup branch logic for distribution groups
for AZs.

The actual fix was in dropping the FilterZones method as it was
relying on improper slice expectations. The append mutates the
array, the correct way would have been a nil array.
Go only returns a new underlying array for an append operation
when the array capacity is increased.
See https://play.golang.org/p/gUuDY0y1sqa for an example.

## QA steps

Follow the steps in the bug.

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1860083

### DIFF
--- a/worker/provisioner/provisioner_task.go
+++ b/worker/provisioner/provisioner_task.go
@@ -753,6 +753,20 @@ type AvailabilityZoneMachine struct {
 	ExcludedMachineIds set.Strings // Don't use these machines in the zone.
 }
 
+// MatchesConstraints against an AZ. If the constraints specifies Zones, make sure
+// this AZ matches a listed ZoneName.
+func (az *AvailabilityZoneMachine) MatchesConstraints(cons constraints.Value) bool {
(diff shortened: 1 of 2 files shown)


worker/provisioner/provisioner_task.go
```diff
@@ -753,6 +753,19 @@ type AvailabilityZoneMachine struct {
 	ExcludedMachineIds set.Strings // Don't use these machines in the zone.
 }
 
+// MatchesConstraints against an AZ.
```
wallyworld: can we expand a bit?

phvalguima:
Hi, I gave this PR a quick trial on our environment and it seemed to work fine!
Can we get it merged so I can pull it directly from edge channel and test it end-to-end?

Thanks for this!

hpidcock:
$$merge$$

PR #11301, opened 2020-03-10
Recreate exec k8s client on failure

hpidcock:
## Recreate exec k8s client on failure

In the case the caas init worker fails, recreate the exec client
on restart. This is due to possible k8s service account being refreshed
and the older credentials still loaded into memory.

## QA steps

deploy kubeflow on microk8s

## Documentation changes

N/A

## Bug reference

N/A

### DIFF
--- a/worker/caasunitinit/manifold.go
+++ b/worker/caasunitinit/manifold.go
@@ -112,11 +112,6 @@ func Manifold(config ManifoldConfig) dependency.Manifold {
 				return unit.ProviderID(), nil
 			}
 
-			execClient, err := config.NewExecClient(os.Getenv(provider.OperatorNamespaceEnvName))
-			if err != nil {
-				return nil, errors.Trace(err)
-			}
-
 			cfg := Config{
 				Logger:                config.Logger,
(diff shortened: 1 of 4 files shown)


worker/caasunitinit/manifold.go
```diff
@@ -112,19 +112,15 @@ func Manifold(config ManifoldConfig) dependency.Manifold {
...
 			cfg := Config{
 				Logger:                config.Logger,
 				Application:           applicationTag.Id(),
 				Clock:                 clock,
 				DataDir:               agentConfig.DataDir(),
 				ContainerStartWatcher: client,
 				UnitProviderIDFunc:    unitProviderID,
-				ExecClient:            execClient,
+				NewExecClient:         config.NewExecClient,
+				Namespace:             os.Getenv(provider.OperatorNamespaceEnvName),
```
wallyworld: We don't need to pass in namespace if we tweak how we define the NewExecClient param - pass a lamda which calls config.NewExecClient
I'd prefer that to limit the coupling and reduce the number of config params

hpidcock:
$$merge$$

PR #11302, opened 2020-03-10
Fix 1866658 jujud called in Makefile before built

hpidcock:
## Fix 1866658 jujud called in Makefile before built

Test for existence of jujud before calling it. Due to each make target
evaluating the Makefile we can just ignore the existence of jujud.

## QA steps

Make sure no jujud is in PATH
make operator-image

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1866658


wallyworld:
$$merge$$

PR #11303, opened 2020-03-10
Increment juju to 2.7.4

jameinel:
## Description of change

Actually cherry-pick the bump to 2.7.4 to do so correctly, instead of just version/version.go

## QA steps

Look at it. (The bot normally does this, I should have just use the bot's commit as I did here.)

## Documentation changes

None.

## Bug reference

None.


mitechie:
$$merge$$

PR #11304, opened 2020-03-10
Protect against nil profile from the charm.

howbazaar:
The 2.7.3 release changed where the instance mutator got its lxd profile from, but did not guard against a nil struct return which old charms will have.

### DIFF
--- a/apiserver/facades/agent/instancemutater/shim.go
+++ b/apiserver/facades/agent/instancemutater/shim.go
@@ -90,6 +90,9 @@ type charmShim struct {
 
 func (s *charmShim) LXDProfile() lxdprofile.Profile {
 	profile := s.Charm.LXDProfile()
+	if profile == nil {
+		return lxdprofile.Profile{}
+	}
 	return lxdprofile.Profile{
 		Config:      profile.Config,
 		Description: profile.Description,


apiserver/facades/agent/instancemutater/shim.go
```diff
@@ -90,6 +90,9 @@ type charmShim struct {
 
 func (s *charmShim) LXDProfile() lxdprofile.Profile {
 	profile := s.Charm.LXDProfile()
+	if profile == nil {
```
babbageclunk: I think this needs to initialise the maps in Config and Devices, otherwise it's likely to just push the problem further down the code.

apiserver/facades/agent/instancemutater/shim.go
```diff
@@ -90,6 +90,9 @@ type charmShim struct {
...
+		return lxdprofile.Profile{}
```
timClicks:
```suggestion
		return lxdprofile.Profile{
			Config:      make(map[string]string),
			Description: "",
			Devices:     make(map[string]map[string]string),
		}
```

apiserver/facades/agent/instancemutater/shim.go
howbazaar: No, we don't need this.

howbazaar:
$$merge$$

howbazaar:
Verified interactively that this fixes the panic.

hmlanigan:
Fixes: https://bugs.launchpad.net/juju/+bug/1866916

PR #11306, opened 2020-03-11
Ensure k8s juju upgrades select correct agent version.

wallyworld:
## Description of change

Running upgrade-controller would select the most recent version, even if it were a devel release.
It queries the docker images, but there's no metadata describing the version. So we query simplestreams agent metadata and use that to guide the version selection. This leads to common code with iaas so a bunch of refactoring was done. It includes deleting a bunch of k8s specific code and tweaking logic for version selection to better organise the workflow. Also, the streams metadata is not mandatory for k8s (there is a fallback) so to allow for installs where streams data is not available, the fetch tools method is used with a timeout.

Tests are unchanged except for better (more generic error messages) and additional tests for the k8s version selection.

## QA steps

bootstrap a k8s 2.7.2 controller
run a juju upgrade-controller --dry-run to see that 2.7.3 is offered but 2.8-beta1 is not
run an upgrade-controller and ensure upgrade works
do the same for a k8s model

rinse and repeat with a lxd controller

## Bug reference

https://bugs.launchpad.net/juju/+bug/1865416

### DIFF
--- a/cmd/juju/commands/upgradecontroller.go
+++ b/cmd/juju/commands/upgradecontroller.go
@@ -5,14 +5,18 @@ package commands
 
 import (
 	"fmt"
+	"strings"
+	"time"
 
 	"github.com/juju/cmd"
 	"github.com/juju/errors"
 	"github.com/juju/gnuflag"
+	"github.com/juju/utils/set"
 	"github.com/juju/version"
 	"gopkg.in/juju/names.v3"
 
 	"github.com/juju/juju/api/modelconfig"
+	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/cloudconfig/podcfg"
 	jujucmd "github.com/juju/juju/cmd"
 	"github.com/juju/juju/cmd/modelcmd"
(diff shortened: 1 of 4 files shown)


wallyworld:
[upgrading a controller goes to beta version by default](https://trello.com/c/Iv1ydDw2/1068-upgrading-a-controller-goes-to-beta-version-by-default)

cmd/juju/commands/upgradecontroller.go
```diff
@@ -133,6 +138,47 @@ func (c *upgradeControllerCommand) Run(ctx *cmd.Context) (err error) {
...
+			}
+		}
+		if err != nil {
+			errChan <- err
+		} else {
+			result <- findResult.List
+		}
+	}()
+
+	var resultList tools.List
```
hpidcock: not sure why this is declared here, we don't do anything with it except return

wallyworld: was left over from older version of the code, fixed

cmd/juju/commands/upgrademodel.go
```diff
@@ -802,7 +750,7 @@ func (context *upgradeContext) maybeChoosePackagedAgent() (err error) {
 		// If not completely specified already, pick a single tools version.
 		filter := coretools.Filter{Number: context.chosen}
 		if context.packagedAgents, err = context.packagedAgents.Match(filter); err != nil {
-			return err
```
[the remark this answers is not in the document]
wallyworld: We want to use Wrap() since we need to hide the underlying tools match error which talks about binaries

wallyworld:
$$merge$$

PR #11311, opened 2020-03-12
minimal jujuc 1867021

jameinel:
### Checklist

## Description of change

This goes back to an old way, and splits out the 'hook-tool' command into an executable called 'jujuc' (which ends up around 8MB when stripped), separated from 'jujud' (116MB stripped). It also updates the symlink build so that if it finds a 'jujuc' it will create the links to it, and updates the tarball to include 'jujuc', etc.

The big win for this, is that 'jujud' takes 60ms to start the process before it realizes that it just wants to send commands to the juju agent, while 'jujuc' takes approx 1ms to get to the same point.

It isn't without cost. There is a price for having the extra 8MB binary:
```
24M Mar 12 10:52 with.tar.xz
21M Mar 12 10:52 without.tar.xz

35M Mar 12 10:52 with.tgz
31M Mar 12 10:52 without.tgz

58M Mar 12 10:50 with-nostrip.tgz
52M Mar 12 10:49 without-nostrip.tgz
```

So it costs around 11-15% to have the extra file in there, regardless of whether we strip or whether we use .xz instead of .gz. However, note that with.tar.xz is still a lot smaller than without.tgz. We could pay back the cost of the extra size if we wanted to go to .xz compression.

I don't actually think this should be in a 2.7 release, but because 2.7 hasn't merged to develop and I prototyped this on 2.7, I wanted to get feedback before rebasing, etc.

This likely makes a huge difference for any charms that call a lot of hook tools (like juju-log). I would really like to see a run of 'juju deploy openstack' with and without this change.

## QA steps

```sh
$ juju bootstrap lxd lxd
$ juju switch controller
$ juju deploy cs:~jameinel/ubuntu-lite --to 0
$ juju ssh 0
$$ ls -l /var/lib/juju/tools/machine-*/
total 53058
-rw-r--r-- 1 ubuntu ubuntu         7 Mar 12 07:35 FORCE-VERSION
lrwxrwxrwx 1 root   root          46 Mar 12 07:38 action-fail -> /var/lib/juju/tools/2.7.5.1-bionic-amd64/jujuc
lrwxrwxrwx 1 root   root          46 Mar 12 07:38 action-get -> /var/lib/juju/tools/2.7.5.1-bionic-amd64/jujuc
lrwxrwxrwx 1 root   root          46 Mar 12 07:38 action-log -> /var/lib/juju/tools/2.7.5.1-bionic-amd64/jujuc
lrwxrwxrwx 1 root   root          46 Mar 12 07:38 action-set -> /var/lib/juju/tools/2.7.5.1-bionic-amd64/jujuc
lrwxrwxrwx 1 root   root          46 Mar 12 07:38 add-metric -> /var/lib/juju/tools/2.7.5.1-bionic-amd64/jujuc
lrwxrwxrwx 1 root   root          46 Mar 12 07:38 application-version-set -> /var/lib/juju/tools/2.7.5.1-bionic-amd64/jujuc
lrwxrwxrwx 1 root   root          46 Mar 12 07:38 close-port -> /var/lib/juju/tools/2.7.5.1-bionic-amd64/jujuc
lrwxrwxrwx 1 root   root          46 Mar 12 07:38 config-get -> /var/lib/juju/tools/2.7.5.1-bionic-amd64/jujuc
lrwxrwxrwx 1 root   root          46 Mar 12 07:38 credential-get -> /var/lib/juju/tools/2.7.5.1-bionic-amd64/jujuc
-rw-r--r-- 1 root   root         209 Mar 12 07:36 downloaded-tools.txt
...
-rwxr-xr-x 1 ubuntu ubuntu   8962048 Mar 12 07:35 jujuc
-rwxr-xr-x 1 ubuntu ubuntu 115477864 Mar 12 07:35 jujud
...
```

You can see that all of those target 'jujuc' instead of 'jujud', and that both 'jujuc' and 'jujud' are installed.

I also added an action to my local copy of ubuntu-lite, which is just:
```
#!/usr/bin/env python3

import subprocess
import time


def count():
    output = subprocess.check_output(['action-get', 'count'])
    return int(output.strip())

def output(val, *extra):
    subprocess.check_call(['action-set', 'output=' + str(val)] + list(extra))


def log(msg):
    #subprocess.check_output(['echo', str(msg), '>/tmp/out.txt'], shell=True)
    subprocess.check_output(['juju-log', str(msg)])


def main():
    c = count()
    start = time.clock_gettime(time.CLOCK_MONOTONIC)
    last = start
    for i in range(c):
        now = time.clock_gettime(time.CLOCK_MONOTONIC)
        log('%8.3fs %8.3fs count: %d' % (now - start, now-last, i))
        # output(last-now)
        # config = subprocess.check_output(['config-get'])
        last = now
    now = time.clock_gettime(time.CLOCK_MONOTONIC)
    total = now - start
    # output(str(total / c), 'config='+str(config))
    output(str(total / c))


if __name__ == '__main__':
    main()
```

Without this patch, on lxd on my machine, I see it taking ~60ms per call to 'juju-log'. With the patch, it drops down to around 6ms.

Note that `juju bootstrap` with prebuilt binaries and `juju upgrade-controller --build-agent` use slightly different codepaths, so it can be worthwhile to test both. I also haven't vetted the full release process, to make sure that snaps will have all the binaries we need, nor that the process that generates streams.canonical.com would do the right thing. 

with the patch:
```
$ juju run-action --wait ubuntu-lite/0 logging-test
unit-ubuntu-lite-0:
  UnitId: ubuntu-lite/0
  id: "1"
  results:
    output: "0.006217624710407108"
```
without the patch:
```
$ juju run-action --wait ubuntu-lite/0 logging-test
unit-ubuntu-lite-0:
  UnitId: ubuntu-lite/0
  id: "1"
  results:
    output: "0.07029640914988704"
```

## Documentation changes

Nothing changes for the user workflow, though it should make install and various hooks quite a bit faster. Especially reactive charms that spend a lot of time in 'juju-log'.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1867021


hpidcock:
I do prefer this to one binary rules them all approach. This seems more of a 2.8 chunk of work, so as you mentioned I'd rather this off develop.

Bunch of k8s work would also need to be done, we move the jujud binary around in a few places such as in the dockerfile, init scripts etc. Symlink creation also would need to be updated in k8s. 

One thing that I'm kinda curious about is why jujud startup time is slow. I don't think it would have much to do about the size of the binary (there is some level of caching in the kernel), rather than all the init code over the whole project. Would be good to do some pre-main perf analysis to see where we spend the most time.

jameinel:
As far as the size of the codebase, import worker/uniter/runner/jujuc and components.RegisterForServer were both necessary to go from 80MB down to 9MB.
I'm guessing all the server code that registers what API/etc are available is taking at least some of the time.
It's pretty hard to profile what happens *before* you hit main().
Supposedly you can use 'perf' but I didn't find a great way to figure it out.

achilleasa:
I wonder if the jujuc start time could be reduced further by mounting a ramdisk and copying the stripped jujuc binary there.

Also, just curious: as the jujuc server already talks to the unit agent over unix sockets, did we ever investigate daemonizing jujuc and making the hook tools dumb shell scripts that talk to jujuc over another unix socket (passing the relevant envvars as peer creds)?

hpidcock:
I don't think a ramdisk is necessary. I do like your suggestion of using shell scripts for the jujuc instead, but I think it will bring issues with windows agents, we will need to duplicate scripts with powershell. 

Additionally jujuc communicates over TCP+TLS for k8s. I would actually possibly consider we use http/https for jujuc, curl could then be used. Not sure if curl supports unix socks but worth looking into.

jameinel:
"jujuc server already talks to the unit agent over unix sockets"
The jujuc *server* runs inside the unit agent. The jujuc client that we invoke from hook scripts connects to that over the unix socket.
So it isn't worth running an intermediary, AFAICT.

I did look at updating charms to speak directly to the agent socket, as the actual structs being passed back and forth are small. Unfortunately it uses 'gob' as the encoding for the structs, which means it is non-trivial to bring into Python. I did find [pygob](https://github.com/mgeisler/pygob). I don't really want to add it as a dependency, but I can prototype with it, I guess.

jameinel:
!!build!!

environs/tools/build.go
```diff
@@ -204,6 +204,19 @@ func copyExistingJujud(dir string) error {
 	if err != nil {
 		return errors.Trace(err)
 	}
+	jujucLocation := filepath.Join(jujuDir, names.Jujuc)
+	jujucTarget := filepath.Join(dir, names.Jujuc)
+	if os.Stat(jujucLocation); os.IsNotExist(err) {
+		logger.Infof("jujuc not found at %s not including", jujucLocation)
```
wallyworld: , not including

cmd/jujuc/main.go
```diff
@@ -0,0 +1,205 @@
...
+		return "", err
+	}
+	abs, err := filepath.Abs(dir)
+	if err != nil {
+		return "", err
+	}
+	return abs, nil
+}
+
+func getSocket() (sockets.Socket, error) {
```
wallyworld: perhaps this function can be shared between jujuc and jujud as it is non trivial and having a common implementation would be good, maybe a helper in github.com/juju/juju/juju/sockets

agent/tools/symlinks_test.go
```diff
@@ -57,15 +57,19 @@ func (s *SymlinksSuite) TestEnsureSymlinksSymlinkedDir(c *gc.C) {
 }
 
 func (s *SymlinksSuite) testEnsureSymlinks(c *gc.C, dir string) {
+	// If we have both 'jujuc' and 'jujud' prefer 'jujuc'
```
wallyworld: Is there a test with just jujud where we check the links?

jameinel: not currently, we probably should add it.

cmd/jujuc/main.go
jameinel: indeed. we'll want to be a little careful about dependencies, but somewhere like 'sockets' seems a reasonable place (and we already depend on it).

It also feels like something I'd really like to decouple from env vars for testability, and then the values passed in could come from env vars.

environs/tools/build.go
jameinel: done.

jameinel:
Another point that I ran into, while playing with this. 'juju-with-jujuc upgrade-controller --build-agent' will fail if your current GOPATH juju/juju directory doesn't have jujuc (because it is looking to build jujuc and put it in the tarball). conversely 'juju-without upgrade-controller' where your source tree *does* have it, won't complain and will happily upgrade you to a tarball that doesn't have it.
I don't know a concrete way around it, unless we change '--build-agent' to not directly build anything but to call "make install" in the source directory, and have *that* object build the tarball.
(I do still think we should have a clearer 'make tool-tarball' step, rather than multiple places with that sort of logic.)

jameinel:
!!build!!

jameinel:
!!build!!

jameinel:
$$merge$$

the failures all appear spurious, both are CAAS related 'did not reach desired state' errors.

wallyworld:
$$merge$$

jameinel:
$$merge$$
util_test.go:812:
    c.Fatalf("never reached desired status")
... Error: never reached desired status

jameinel:
$$merge$$

jameinel:
util_test.go:812:
    c.Fatalf("never reached desired status")
... Error: never reached desired status

PR #11313, opened 2020-03-13
Fix a watcher race that was in the ControllerAddressesSuite.

howbazaar:
It looks like the two WatchAPIHostPorts tests were moved from state_test.go
around nine months ago, but somehow we ended up with both. I have removed
the copy that was in the state_test file.

The StateSuite had the WaitForModelWatchersIdle at the end of the setup,
but the ControllerAddressesSuite didn't. This allowed for initial creation
events to still be propagating through the system when the tests started.

The StateSuite was also reading the model from the database, but this was
already there through the statetesting.StateSuite that the ConnSuite included.
So a drive by to delete that and use the model that was already there.


howbazaar:
$$merge$$

PR #11321, opened 2020-03-16
Add api client calls for the watching of model summaries.

howbazaar:
Even though there is no current Juju API client for this, it is expected
that the Juju Intelligent Model Manager (JIMM) component of JAAS will
be calling this.

## QA steps

Just the tests.

## Documentation changes

We don't really have any API client documentation. Perhaps we should look into this at some stage.

### DIFF
--- a/api/controller/controller.go
+++ b/api/controller/controller.go
@@ -189,6 +189,27 @@ func (c *Client) WatchAllModels() (*api.AllWatcher, error) {
 	return api.NewAllModelWatcher(c.facade.RawAPICaller(), &info.AllWatcherId), nil
 }
 
+// WatchModelSummaries returns a SummaryWatcher, from which you can request
+// the Next set of ModelAbstracts for all models the user can see.
+func (c *Client) WatchModelSummaries() (*SummaryWatcher, error) {
+	var info params.SummaryWatcherID
+	if err := c.facade.FacadeCall("WatchModelSummaries", nil, &info); err != nil {
+		return nil, err
+	}
+	return NewSummaryWatcher(c.facade.RawAPICaller(), &info.WatcherID), nil
+}
+
+// WatchAllModelSummaries returns a SummaryWatcher, from which you can request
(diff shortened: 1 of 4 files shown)


api/controller/controller.go
```diff
@@ -189,6 +189,27 @@ func (c *Client) WatchAllModels() (*api.AllWatcher, error) {
 	return api.NewAllModelWatcher(c.facade.RawAPICaller(), &info.AllWatcherId), nil
 }
 
+// WatchModelSummaries returns a SummaryWatcher, from which you can request
+// the Next set of ModelAbstracts for all models the user can see.
```
mhilton: As an aside, why are there two methods for this? From a client point of view, if you are a superuser then the set of models you can see is all the models on the controller. Therefore you would get the same results for either method?

api/controller/summarywatcher.go
```diff
@@ -0,0 +1,58 @@
...
+		watcher.caller.BestFacadeVersion(watcher.objType),
+		*watcher.id,
+		"Next",
+		nil, &info,
+	)
+	return info.Models, err
+}
+
+// Stop shutdowns down a summary watcher.
+func (watcher *SummaryWatcher) Stop() error {
```
alesstimec: a bit worried about this.. JIMM will have a long-running watcher for all model summaries.. and normally won't stop the watcher unless something bad happens or a restart is required.. what happens if we lose watcher IDs and fail to stop watchers?

api/controller/controller.go
howbazaar: I specifically split this in two because even though I may be a superuser in a controller, perhaps I only want to see "my" models. Without this there would be no way to support this mechanism.

It also simplifies the authentication and internals to be explicit.

api/controller/summarywatcher.go
howbazaar: If the api connection goes away, all the resources which are associated with the connection are stopped automatically.

howbazaar:
$$merge$$

PR #11323, opened 2020-03-16
Add new k8s-spec-set and k8s-spec-get commands

wallyworld:
## Description of change

The pod-spec-set/get commands are deprecated in favour of k8s-spec-set/get.
Also, the previous work to rename actions back to functions left some cruft so this was fixed.

## QA steps

deploy a k8s charm
exec into the operator pod and ensure that k8s-spec-get/set commands and the pod-spec versions are present
juju-run both pod-spec-get and k8s-spec-get
edit a local charm to use k8s-spec-set instead of pod-spec-set and deploy


worker/uniter/runner/jujuc/k8s-spec-get.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2019 Canonical Ltd.
```
wallyworld: This is renamed from pod-spec-get.go but somehow git thinks it was deleted and created so looks like a new file

wallyworld:
[rename pod-spec-set to k8s-spec-set](https://trello.com/c/3Rh4kAvb/1018-rename-pod-spec-set-to-k8s-spec-set)

cmd/juju/commands/helptool_test.go
```diff
@@ -58,13 +58,15 @@ Currently available charm hook tools are:
...
+    k8s-spec-get             get k8s spec information
+    k8s-spec-set             set k8s spec information
     leader-get               print application leadership settings
     leader-set               write application leadership settings
     network-get              get network config
     open-port                register a port or range to open
     opened-ports             lists all ports or ranges opened by the unit
-    pod-spec-get             get pod spec information
-    pod-spec-set             set pod spec information
+    pod-spec-get             get k8s spec information
```
hmlanigan: suggest adding (deprecated) on the end?

wallyworld: ok

wallyworld:
$$merge$$

PR #11325, opened 2020-03-16
Fix BuildJuju  on s390x

hmlanigan:
## Description of change

Update the sys pkg to all for the update crypto pkg to not fail make rebuild-schema on s390x.

## QA steps

Testing by building juju on the s390x-slave.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1299,7 +1299,7 @@
   revision = "42b317875d0fa942474b76e1b46a6060d720ae6e"
 
 [[projects]]
-  digest = "1:2e1aadff5bda52e267975408bf6b5e90135fc643829c9e4e964ff82c3826a1dd"
+  digest = "1:aa645e6a58c6180d625e0824c75fe72a5a7deb88e5d87b850ca8a639ddb579f4"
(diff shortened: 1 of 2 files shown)


hmlanigan:
I took the tip of master SHA.  Perhaps tip of release-branch.go1.14-std or release-branch.go1.14-cmd would be better. However the difference is unclear.

hmlanigan:
This also brings up the problem of using go 1.11 and 1.12 for builds and releases still and having some go libraries at 1.14.  Are we introducing any new problems?

hpidcock:
As discussed with @hmlanigan I think the x/ packages are fine to be bumped to higher versions. If we started vendoring std pkgs then I'd be concerned.

I tested building on s390x, arm64, ppc64le, amd64, windows/amd64 and all built.

hpidcock:
Also with regards to the difference of 1.14-std and 1.14-cmd. The cmd branch is used by the go tooling, 1.14-std is the pinned version by the std library.

hmlanigan:
Updated sha to the tip of 1.14-std.

hmlanigan:
$$merge$$

PR #11331, opened 2020-03-17
Fix goimports

hmlanigan:
## Description of change

Fix some import stanzas which go missed when go imports was not installed for the GitHub-static-analysis-go job.


hmlanigan:
$$merge$$

PR #11334, opened 2020-03-19
Add new update-k8s command

wallyworld:
## Description of change

Add a new update-k8s command.
The command can either accept a built-in cloud name like "microk8s", or else
behaves the same as update-cloud where a cloud file is used to pass in the updated info.

The CloudFileReader from cmd/juju/cloud/add.go was reused so was exported. The CloudMetadataStore used in the cmd/juju/cloud had a ParseCloudMetadataFile() method but this caused duplicate test data to be needed in tests/mocks and didn't work well with the new cass/update-k8s tests. So it was reworked to have a ReadCloudData() method instead.

Also remove from unused code from update-cloud.
And a driveby fix for https://bugs.launchpad.net/juju/+bug/1865894

## QA steps

Run the command for microk8s and also a bespoke clouds.yaml file
$ juju update-k8s microk8s

$ juju update-k8s mytestk8s -f clouds.yaml

Do it for both --client and a k8s added to a lxd controller
Check cloud is updated with show-cloud

## Bug reference

https://bugs.launchpad.net/juju/+bug/1864259

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -35,7 +35,7 @@ import (
 var logger = loggo.GetLogger("juju.cmd.juju.k8s")
 
 type CloudMetadataStore interface {
-	ParseCloudMetadataFile(path string) (map[string]jujucloud.Cloud, error)
+	ReadCloudData(path string) ([]byte, error)
 	ParseOneCloud(data []byte) (jujucloud.Cloud, error)
 	PublicCloudMetadata(searchPaths ...string) (result map[string]jujucloud.Cloud, fallbackUsed bool, _ error)
 	PersonalCloudMetadata() (map[string]jujucloud.Cloud, error)
--- a/cmd/juju/caas/update.go
+++ b/cmd/juju/caas/update.go
@@ -1,0 +1,367 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package caas
+
+import (
+	"fmt"
+
+	"github.com/juju/cmd"
+	"github.com/juju/errors"
+	"github.com/juju/gnuflag"
+	"gopkg.in/juju/names.v3"
+
+	cloudapi "github.com/juju/juju/api/cloud"
+	"github.com/juju/juju/apiserver/params"
+	"github.com/juju/juju/caas"
+	"github.com/juju/juju/caas/kubernetes/provider"
+	jujucloud "github.com/juju/juju/cloud"
+	jujucmd "github.com/juju/juju/cmd"
+	"github.com/juju/juju/cmd/juju/cloud"
+	"github.com/juju/juju/cmd/juju/common"
+	"github.com/juju/juju/cmd/modelcmd"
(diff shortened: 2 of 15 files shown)


cmd/juju/caas/baseaddupdate.go
```diff
@@ -0,0 +1,471 @@
+// Copyright 2020 Canonical Ltd.
```
wallyworld: This is all extracted/moved from add.go

cmd/juju/caas/add.go
```diff
@@ -4,20 +4,13 @@
 package caas
 
 import (
-	"bytes"
```
wallyworld: Al this deleted code is moved to baseaddupdate.go

wallyworld:
[juju bootstrap --clouds ends with k8s workloads message](https://trello.com/c/vqJ1IUMu/1076-juju-bootstrap-clouds-ends-with-k8s-workloads-message)

wallyworld:
[cannot update-cloud microk8s](https://trello.com/c/5qhQhONM/1069-cannot-update-cloud-microk8s)

cmd/juju/caas/update.go
```diff
@@ -0,0 +1,406 @@
...
+		}
+	}
+
+	if clusterName == "" && cloudToUpdate.HostCloudRegion != caas.K8sCloudOther {
+		clusterName = cloudToUpdate.HostCloudRegion
+	}
+	if clusterName != "" {
+		clusterName = fmt.Sprintf("%q ", clusterName)
+	}
+	successMsg := fmt.Sprintf("k8s substrate %sknown to Juju as cloud %q updated.", clusterName, c.caasName)
```
ycliuhw: `%s known`

cmd/juju/caas/update.go
```diff
@@ -0,0 +1,406 @@
+// Copyright 2020 Canonical Ltd.
```
ycliuhw: There are still lots of code here are similar to `add.go`. 
But it seems that it's not very to refactor further.

cmd/juju/caas/update.go
wallyworld: The substituted value for clusterName has a " " at the end (it may be empty).

cmd/juju/caas/update.go
wallyworld: Yeah, there's common-ish code but the add command does querying of the k8s substrate to get storage info so it's different enough that it's messy to try and pull apart.

wallyworld:
```yaml
    regions:
      localhost:
        endpoint: https://172.16.104.128:16443
```

That endpoint is meant to be empty (for microks8). I'll need to tweak the logic. microk8s doesn't define any regions so we add "localhost" as a special case (for add-k8s too)

wallyworld:
I'll reworked things so that this command behaves more like the update-cloud command.

wallyworld:
The endpoint under region is not there anymore - it was a result of the initial implementation which has been reworked.

wallyworld:
$$merge$$

PR #11338, opened 2020-03-19
small refactor of uniter's context FactoryConfig.

hmlanigan:
## Description of change

Replace UnitTag with Unit in context.FactoryConfig.  The caller of context.NewContextFactory already has a unit from state. No need to get a second copy of it, just pass as a config value.

## QA steps

There should be no visible change to juju operation.
```console
juju bootstrap localhost
juju deploy ubuntu

# once the unit is deployed, check the logs.
juju debug-log
```


hmlanigan:
$$merge$$

PR #11339, opened 2020-03-19
move uniter state server side

hmlanigan:
## Description of change

Move the uniter internal state currently saved in a local file, to the controller db.  This will allow juju to reduce the number persistent volumes used in k8s clouds.

operations.StateFile was renamed to operations.StateOps.  The functionality of Read and Write to a local file was replaced with calls to the uniter api for State and SetState.  

Migration and other uniter internal state will be handled in a future pr.  Upgrade is handled here.  

## QA steps

```console 
juju bootstrap localhost
juju deploy ubuntu

# once the unit is deployed, check the juju db
juju-db.bash
juju:PRIMARY> db.unitstates.find()

# run integration tests of hook-tools
(cd tests ; ./main.sh hook_tools)
```

Bootstrap and deploy a unit to microk8s.  No change should be noted.


worker/uniter/operation/state.go
```diff
@@ -195,35 +212,50 @@ func (change stateChange) apply(state State) *State {
 	return &state
 }
 
-// StateFile holds the disk state for a uniter.
-type StateFile struct {
+// StateOps reads and writes uniter state to/from the controller.
+type StateOps struct {
 	path string
```
hmlanigan: I'll remove this once the upgrade step is done.

state/unit.go
```diff
@@ -359,23 +362,32 @@ func removeUnitStateOp(mb modelBackend, globalKey string) txn.Op {
...
+// Change the arg to a struct as storage and relation state is added.
+func (u *Unit) SetStateOperation(unitState map[string]string, uniterState string) ModelOperation {
+	return &unitSetStateOperation{u: u, newState: unitState, newUniterState: uniterState}
 }
 
 // State returns the persisted state for a unit.
-func (u *Unit) State() (map[string]string, error) {
+//
+// TODO (hml) 03-18-2020
+// Return a structure instead as storage and relation state is added.
```
achilleasa: How about extracting all the Get/SetState bits into a standalone PR so this TODO can be addressed? I would prefer we minimize the amount of TODOS given that 2.8 is near. Doing so would make the changes in this PR easier to review.

upgrades/steps_28.go
```diff
@@ -64,3 +80,83 @@ func prepopulateRebootHandledFlagsForDeployedUnits(ctx Context) error {
 	}
 	return nil
 }
+
+func moveUniterStateToController(ctx Context) error {
```
achilleasa: I was under the impression that you meant to do this as a fallback in the code that reads the uniter state. No objection to doing it as an upgrade step (guess it's cleaner) but I am curious as to why you decided to go with this approach.

worker/uniter/operation/executor.go
```diff
@@ -27,27 +27,52 @@ var (
...
+// exists, the executor's starting state will be the supplied InitialState.
+func NewExecutor(cfg ExecutorConfig) (Executor, error) {
+	if err := cfg.validate(); err != nil {
+		return nil, err
+	}
+	// TODO: hml 03-18-2020
+	// handle fallback order better in relation to ugprades
+	// of a unit from state stateOps to state on controller
+	// or implement upgrade steps.
+	stateOps := NewStateOps(cfg.StateFilePath, cfg.StateOps)
```
achilleasa: Since `cfg.StateOps` implements the same interface as the `StateOps` type, I think it would be much cleaner if the caller instantiated it directly (it is aware of the file path) and passed it as part of the configuration.

It would also make it easier to test the executor by mocking it.

state/migration_export.go
```diff
@@ -918,10 +918,11 @@ func (e *exporter) addApplication(ctx addApplicationContext) error {
 		if cloudContainer, found := ctx.cloudContainers[unit.globalKey()]; found {
 			args.CloudContainer = e.cloudContainer(cloudContainer)
 		}
-		if args.State, err = unit.State(); err != nil {
+		// TODO: hml 18-mar-2020
+		// add the rest of unit.State() to model migration
+		if args.State, _, err = unit.State(); err != nil {
```
achilleasa: Same comment about minimizing TODOs. I would recommend updating juju/description with the required fields and then updating this PR to use them.

apiserver/params/internal.go
```diff
@@ -270,8 +281,9 @@ type SetUnitStateArgs struct {
 // SetUnitStateArg holds a unit tag and a map with KV-pairs that represent a local
 // charm state to be persisted by the controller
 type SetUnitStateArg struct {
-	Tag   string            `json:"tag"`
-	State map[string]string `json:"state"`
+	Tag         string            `json:"tag"`
+	State       map[string]string `json:"state"`
```
achilleasa: Since we are changing the payload format, can we add an `omitempty` for this field as well?

state/unit.go
achilleasa: Keep in mind that we also still need to support storing the uniter relation state data so I think having a wrapper struct here that also includes an extra `map[int]interface{}`or `map[int]map[string]interface{}` field for this would be beneficial before we proceed with the actual uniter changes.

upgrades/steps_28.go
hmlanigan: I had thought to do as a fall back but changed my mind after talking to @mitechie.  It's cleaner to do the upgrade step and lets us remove the old code as well.  Thought I'd give it a day to get the upgrade step working.  Looks like that's a go.

state/unit.go
hmlanigan: I'll pull the state piece out into a separate PR.  I was planning to leave all of the uniter state pieces, uniter, relation, storage etc, as a string of the marshaled yaml of the struct if possible.  As done for the uniter internal state.

worker/uniter/operation/executor.go
hmlanigan: This will be rejiggered now that I can remove the StateFilePath part.

hmlanigan:
@achilleasa, most of the todos are reminders to myself as I was working on different sections of the code.  The only piece I intend to leave for another PR is the model migration piece, all at once with the other state moving server side since it will require a change to juju/description.

state/unit.go
achilleasa: SGTM; however I don't think it will be as easy for the relation/storage bits as it is a yaml per relation (not sure what's the equivalent for storage) so maybe a `map[int]string` (int = relation id) then for the individual serialized payloads?

hmlanigan:
The original work was broken into pieces.  The api/apiserver/state pieces were in PR #11342

hmlanigan:
rebased to fix a merge conflict

worker/uniter/operation/state.go
```diff
@@ -4,11 +4,10 @@
 package operation
 
 import (
-	"os"
-
 	"github.com/juju/errors"
-	"github.com/juju/utils"
+	"github.com/juju/juju/apiserver/params"
```
hmlanigan: I need to move this down a stanza

apiserver/facades/agent/upgradesteps/upgradesteps.go
```diff
@@ -97,6 +126,45 @@ func (api *UpgradeStepsAPI) ResetKVMMachineModificationStatusIdle(arg params.Ent
 	return result, nil
 }
 
+// WriteUniterState did not exist prior to v2.
+func (*UpgradeStepsAPIV1) WriteUniterState(_, _ struct{}) {}
+
+// WriteUniterState write the uniter state for the set of units provided.
+// Related to the move of the uniter state from a local file to the controller.
+func (api *UpgradeStepsAPI) WriteUniterState(args params.SetUnitStateArgs) (params.ErrorResults, error) {
```
achilleasa: Is it possible to delegate this operation to the uniter facade or would that be too much effort? The implementation looks the same

apiserver/facades/agent/upgradesteps/upgradesteps.go
```diff
@@ -114,3 +182,23 @@ func (api *UpgradeStepsAPI) getMachine(canAccess common.AuthFunc, tag names.Mach
 	}
 	return machine, nil
 }
+
+func (api *UpgradeStepsAPI) getUnit(canAccess common.AuthFunc, tag names.UnitTag) (Unit, error) {
+	if !canAccess(tag) {
```
achilleasa: Just curious: is it typical for upgrade steps to log at critical?

apiserver/facades/agent/upgradesteps/upgradesteps.go
```diff
@@ -114,3 +182,23 @@ func (api *UpgradeStepsAPI) getMachine(canAccess common.AuthFunc, tag names.Mach
...
+		return nil, err
+	}
+	// The authorization function guarantees that the tag represents a
+	// unit.
+	var unit Unit
+	var ok bool
+	if unit, ok = entity.(Unit); !ok {
+		return nil, errors.NotValidf("unit entity")
+	}
+	return unit, nil
```
achilleasa: Maybe a bit cleaner:

```suggestion
	if unit, ok = entity.(Unit); ok {
	        return unit, nil
	}
	return nil, errors.NotValidf("unit entity")
```

apiserver/facades/agent/upgradesteps/upgradesteps_test.go
```diff
@@ -23,30 +23,33 @@ import (
...
+type machineUpgradeStepsSuite struct {
+	upgradeStepsSuite
+
+	tag     names.Tag
+	arg     params.Entity
+	machine *mocks.MockMachine
+}
 
-func (s *upgradeStepsSuite) SetUpTest(c *gc.C) {
+//var _ = gc.Suite(&upgradeStepsSuite{})
```
achilleasa: Should this be commented out?

apiserver/facades/agent/upgradesteps/upgradesteps.go
hmlanigan: nope, removed

apiserver/facades/agent/upgradesteps/upgradesteps_test.go
hmlanigan: yes, it's in another spot, forgot to remove.  done now.

apiserver/facades/agent/upgradesteps/upgradesteps.go
hmlanigan: Worker's should only share facades if they are in apiserver/.../common. Then we'd have the uniter state operations split between common and the uniter, as upgrade steps won't use the State() piece.  It'd also be a bit weird as the uniter and upgradsteps would look very different on the api/common side.  Would we only implement the apiserver common piece? We'd also have to guard against mucking up the upgrade steps for 2.8 if this changes in the future on the uniter side.

upgrades/steps_28.go
```diff
@@ -72,3 +88,103 @@ func prepopulateRebootHandledFlagsForDeployedUnits(ctx Context) error {
...
+		// No units, nothing to do.
+		return nil
+	}
+
+	var fileNames []string
+	if ctx.AgentConfig().Value(agent.ProviderType) == k8sprovider.CAASProviderType {
+		if fileNames, err = caasOperatorLocalState(ctx, unitNames); err != nil {
+			return err
+		}
+	} else {
```
achilleasa: nested if can be hoisted to the parent else

upgrades/steps_28.go
```diff
@@ -72,3 +88,103 @@ func prepopulateRebootHandledFlagsForDeployedUnits(ctx Context) error {
...
+		var st operation.State
+		if err = utils.ReadYaml(uniterStateFile, &st); err != nil {
+			// No file available, move on.
+			if os.IsNotExist(err) {
+				continue
+			}
+			return nil, errors.Annotatef(err, "failed to read uniter state from %q", uniterStateFile)
+		}
+		fileNames = append(fileNames, uniterStateFile)
+		if err = st.Validate(); err != nil {
```
achilleasa: Unrelated note: this might be a good place to to consider landing a workaround for https://bugs.launchpad.net/juju/+bug/1869275

upgrades/steps_28.go
```diff
@@ -72,3 +88,103 @@ func prepopulateRebootHandledFlagsForDeployedUnits(ctx Context) error {
...
+			return nil, errors.Annotatef(err, "failed to validate uniter state from %q", uniterStateFile)
+		}
+
+		data, err := yaml.Marshal(st)
```
achilleasa: Sucks that we cannot pass a buffered reader to utils.ReadYaml so we could read the serialized data once instead of unmarshaling and marshaling...

worker/uniter/operation/state.go
```diff
@@ -195,35 +200,51 @@ func (change stateChange) apply(state State) *State {
 	return &state
 }
 
-// StateFile holds the disk state for a uniter.
-type StateFile struct {
-	path string
+// StateOps reads and writes uniter state to/from the controller.
+type StateOps struct {
+	UniterStateOps
```
achilleasa: How about renaming `UniterStateOps` to `UniterStateReadWriter`?
Embedding `UniterStateOps` into `StateOps` took me a while to parse...

Also, why embed instead of assigning to a variable? Are we calling any of the `UniterStateOps` methods on a `StateOps` instance?

worker/uniter/uniter.go
```diff
@@ -166,7 +166,7 @@ type UniterParams struct {
 	RebootQuerier RebootQuerier
 }
 
-type NewOperationExecutorFunc func(string, operation.State, func(string) (func(), error)) (operation.Executor, error)
+type NewOperationExecutorFunc func(operation.ExecutorConfig) (operation.Executor, error)
```
achilleasa: Please add comment since this is an exported type.

upgrades/steps_28.go
hmlanigan: agreed, however in this case will make the code harder to read.

upgrades/steps_28.go
hmlanigan: Do we have a card for this?

worker/uniter/uniter.go
hmlanigan: done

worker/uniter/operation/state.go
hmlanigan: Renaming and using a variable, the interface should not be used directly.

hmlanigan:
Force due to rebase to current develop

apiserver/facades/agent/upgradesteps/upgradesteps.go
achilleasa: Makes sense. Thanks for the clarification.

upgrades/steps_28.go
achilleasa: It's punted for later so no need to worry for now.

worker/uniter/operation/executor.go
```diff
@@ -27,27 +27,42 @@ var (
...
-// NewExecutor returns an Executor which takes its starting state from the
-// supplied path, and records state changes there. If no state file exists,
-// the executor's starting state will include a queued Install hook, for
-// the charm identified by the supplied func.
-func NewExecutor(stateFilePath string, initialState State, acquireLock func(string) (func(), error)) (Executor, error) {
-	file := NewStateFile(stateFilePath)
-	state, err := file.Read()
-	if err == ErrNoStateFile {
-		state = &initialState
+type ExecutorConfig struct {
```
achilleasa: Comment?

worker/uniter/operation/state.go
```diff
@@ -195,35 +200,53 @@ func (change stateChange) apply(state State) *State {
 	return &state
 }
 
-// StateFile holds the disk state for a uniter.
-type StateFile struct {
-	path string
+// StateReadWriter reads and writes uniter state to/from the controller.
```
achilleasa: s/StateReadWriter/StateOps

achilleasa: Shouldn't that be from/to?

worker/uniter/operation/state.go
```diff
@@ -195,35 +200,53 @@ func (change stateChange) apply(state State) *State {
...
+//go:generate mockgen -package mocks -destination mocks/uniterstaterw_mock.go github.com/juju/juju/worker/uniter/operation UnitStateReadWriter
+type UnitStateReadWriter interface {
+	State() (params.UnitStateResult, error)
+	SetState(unitState params.SetUnitStateArg) error
 }
 
-// Read reads a State from the file. If the file does not exist it returns
-// ErrNoStateFile.
-func (f *StateFile) Read() (*State, error) {
+// Read reads a State from the controller. If the saved state does not exist
```
achilleasa: s/Read reads/Read ?

worker/uniter/operation/errors.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 var (
-	ErrNoStateFile            = errors.New("uniter state file does not exist")
+	ErrNoSavedState           = errors.New("saved uniter state does not exist")
```
achilleasa: We should probably add some comments here next time we are around this part of the code-base.

hmlanigan: done

worker/uniter/operation/state.go
hmlanigan: done

worker/uniter/operation/state.go
hmlanigan: done

worker/uniter/operation/executor.go
hmlanigan: done.

hmlanigan:
squashed some commits and made final pr changes

hmlanigan:
$$merge$$

PR #11342, opened 2020-03-20
Enhance UnitState and add SetState.

hmlanigan:
### Checklist

----

## Description of change

Add UniterState, RelationState, and StorageState to unit state saved on
the controller in prepartion to move the uniter internal state from a
unit local files to the controller.  Made the relevant changes to the uniter
api and apiserver pkgs.

Implement SetState to save the data outside of CommitHookChanges.

A unit package was added to core to avoid creating a UnitState structure
in both the API and state.  There may be a better location for this.

The migration step changes needed to enhance what data is moved will be
handled in a near future pr.

## QA steps

```console
# bootstrap and deploy, should work as normal
juju bootstrap localhost 
juju deploy ubuntu

# run the integration test
(cd tests ; ./main.sh hook_tools)
```


core/unit/unit.go
```diff
@@ -0,0 +1,38 @@
...
+	RelationState map[int]string
+
+	// StorageState is a serialized yaml string containing storage internal
+	// state for this unit from the uniter.
+	StorageState string
+}
+
+// BSONRelationData makes the RelationState data BSON friendly by
+// translating the int map key to a string key.
+func (u UnitState) BSONRelationData() map[string]string {
```
achilleasa: I think this should be an internal implementation detail of the serialization logic in the state package similar to how we escape map keys to make things mongo-friendly.

state/unit_ops.go
```diff
@@ -57,20 +58,37 @@ func (op *unitSetStateOperation) Build(attempt int) ([]txn.Op, error) {
...
 		escapedState[mgoutils.EscapeKey(k)] = v
 	}
 
 	// Check if we need to update
-	if stDoc.stateMatches(escapedState) {
+	setFields := bson.D{}
+	if !stDoc.stateMatches(escapedState) {
+		setFields = append(setFields, bson.DocElem{"state", escapedState})
+	}
+	if op.newState.UniterState != "" {
```
achilleasa: You should also compare with current state (e.g. `if op.newState.UniterState != "" && op.newState.UniterState != stDoc.UniterState`)

We should avoid bumping the revno if we are pushing exactly the same state data.

state/unit_ops.go
```diff
@@ -57,20 +58,37 @@ func (op *unitSetStateOperation) Build(attempt int) ([]txn.Op, error) {
...
+		setFields = append(setFields, bson.DocElem{"uniter-state", op.newState.UniterState})
+	}
+	if len(op.newState.RelationState) > 0 {
```
achilleasa: Likewise, if the lengths match we should compare the contents for changes.

state/unit_ops.go
```diff
@@ -57,20 +58,37 @@ func (op *unitSetStateOperation) Build(attempt int) ([]txn.Op, error) {
...
+		setFields = append(setFields, bson.DocElem{"relation-state", op.newState.RelationState})
+	}
+	if op.newState.StorageState != "" {
```
achilleasa: Likewise, we should compare the contents for changes.

state/unit_test.go
```diff
@@ -71,32 +72,38 @@ func (s *UnitSuite) TestApplication(c *gc.C) {
...
+	uState, err = s.unit.State()
 	c.Assert(err, gc.IsNil)
-	c.Assert(ust, gc.DeepEquals, newState)
+	c.Assert(uState.State, gc.DeepEquals, newState)
+	// UniterState should never be removed.  Since SetState had an empty string,
+	// ensure the initialUniterState is still there.
+	c.Assert(uState.UniterState, gc.Equals, initialUniterState)
 }
 
 func (s *UnitSuite) TestUnitStateNopMutation(c *gc.C) {
```
achilleasa: Can you expand this test (or duplicate if you prefer) and make sure that calling SetState with the same data (uniter state, relations, storage) results in ErrNoOperations.

state/unit_ops.go
hmlanigan: Sure, pays to be paranoid.  I'm relying on the uniter to only send changes when they were necessary.

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -3322,15 +3324,64 @@ func (u *UniterAPI) State(args params.Entities) (params.UnitStateResults, error)
 			res[i].Error = common.ServerError(err)
 			continue
 		}
-
-		if res[i].State, err = unit.State(); err != nil {
+		unitState, err := unit.State()
+		if err != nil {
 			res[i].Error = common.ServerError(err)
```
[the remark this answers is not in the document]
hmlanigan: darn it. glad you caught this.

state/unit_ops.go
achilleasa: Never trust the client ;-)

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -3322,15 +3324,64 @@ func (u *UniterAPI) State(args params.Entities) (params.UnitStateResults, error)
...
+		unit, err := u.getUnit(unitTag)
+		if err != nil {
+			res[i].Error = common.ServerError(err)
+			continue
+		}
+
+		unitState := coreunit.UnitState{
+			State:         arg.State,
+			UniterState:   arg.UniterState,
+			RelationState: nil,
```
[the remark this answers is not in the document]
hmlanigan: done

core/unit/unit.go
hmlanigan: moved

state/unit_ops.go
hmlanigan: updated

state/unit_ops.go
hmlanigan: updated

state/unit_test.go
hmlanigan: done

apiserver/params/internal.go
```diff
@@ -270,8 +277,11 @@ type SetUnitStateArgs struct {
 // SetUnitStateArg holds a unit tag and a map with KV-pairs that represent a local
```
achilleasa: For future reference, can you add a comment that these fields are optional and deletions can be specified by the caller by setting the pointer to an empty value?

state/unitstate.go
```diff
@@ -0,0 +1,243 @@
...
+	storageStateSet bool
+}
+
+// NewUnitState returns a new UnitState struct.
+func NewUnitState() *UnitState {
+	return &UnitState{}
+}
+
+// NoChanges returns true if no values of the struct have been set.
+func (u *UnitState) NoChanges() bool {
```
achilleasa: Perhaps `Modified` with an inverted check would be a better name for this method.

state/unitstate.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// UniterState returns the uniter state and bool indicating
+// whether the data was set.
+func (u *UnitState) UniterState() (string, bool) {
+	return u.uniterState, u.uniterStateSet
+}
+
+// SetRelationState sets the relation state value.
+func (u *UnitState) SetRelationState(state map[int]string) {
```
achilleasa: Do we care if the caller passes a nil map here? Guess not; we will still treat that as a modification, right?

hmlanigan: Correct, if you add a nil map, it will delete all the data.  If you don't call it, nothing will be done to relation data.

state/unitstate.go
hmlanigan: Is there a better name without an inverted check?

apiserver/params/internal.go
hmlanigan: added

state/unitstate.go
achilleasa: Personally I prefer `Modified` or `Mutated` because I like the way the check reads `if !modified { return }`

hmlanigan: done

hmlanigan:
Changes squashed to one after renaming NoChanges.

hmlanigan:
$$merge$$

hmlanigan:
fatal: unable to access 'https://github.com/golang/dep/': Could not resolve host: github.com

$$merge$$

PR #11344, opened 2020-03-23
Protect query tracker from concurrent access

wallyworld:
## Description of change

As seen here:

https://jenkins.juju.canonical.com/job/RunUnittests-race-amd64/2328/console

The allwatcher worker can access the query tracker queries slice at the same time as code which performs db operations. So add a mutex to protect the queries slice.

## QA steps

go test --race

### DIFF
--- a/state/querytracker.go
+++ b/state/querytracker.go
@@ -3,7 +3,10 @@
 
 package state
 
-import "runtime/debug"
+import (
+	"runtime/debug"
+	"sync"
+)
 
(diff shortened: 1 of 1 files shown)


howbazaar:
Can we retarget to 2.7?

wallyworld:
@babbageclunk IMO not worth the effort since go test --race finds the issue for us

wallyworld:
$$merge$$

PR #11350, opened 2020-03-24
Store machineID on subordinate unit document.

howbazaar:
Subordinate units are created when the principal unit enters the relation scope. At this time, the machine ID is known and never changes. Rather than have the subordinate unit always going throught the principal unit document to work out the assigned machine ID, have the subordinate document created with the machine ID from the principal.

There was quite a bit of test fallout with this change as many of the tests did not follow the real world and didn't bother assigning the principal unit to a machine before creating the subordinate. These tests have been updated.

One of the primary reasons for doing this piece of work is to allow the allwatcher to be able to process the initial units of a model without additional queries per unit. Before this change a principal unit would have read the machine document twice, once for the public address, and once for the private address. A subordinate would have read the principal document twice in addition to the machine document. The allwatcher change now stores the public and private address on the core/multiwatcher type, and the unit handling now doesn't need additional queries to get the machine ID. CAAS units still get the addresses directly from the unit itself as the logic there is more complicated and deals with entities that aren't yet stored in the multiwatcher store.

## QA steps

```sh
juju deploy ~jameinel/ubuntu-lite
juju deploy telegraf
juju relate ubuntu-lite telegraf
```
Observe that the telegraf unit gets an address in juju status.

## Documentation changes

Internal change only.

### DIFF
--- a/state/relationunit.go
+++ b/state/relationunit.go
@@ -244,17 +244,31 @@ func (ru *RelationUnit) subordinateOps() ([]txn.Op, string, error) {
 		return nil, "", err
 	}
 	if len(related) != 1 {
-		return nil, "", fmt.Errorf("expected single related endpoint, got %v", related)
+		return nil, "", errors.Errorf("expected single related endpoint, got %v", related)
 	}
+	// Find the machine ID that the principal unit is deployed on, and use
+	// that for the subordinate. It is worthwhile noting that if the unit is
+	// in a CAAS model, there are no machines.
+	principal, err := ru.st.Unit(ru.unitName)
+	if err != nil {
+		return nil, "", errors.Annotate(err, "unable to load principal unit")
+	}
+	// We dont care just now if the machine isn't assigned, as CAAS models
+	// will return that error. For IAAS models, the machine *should* always
+	// be assigned before it is able to enter scope.
+	principalMachineID, _ := principal.AssignedMachineId()
 	applicationname, unitName := related[0].ApplicationName, ru.unitName
 	selSubordinate := bson.D{{"application", applicationname}, {"principal", unitName}}
 	var lDoc lifeDoc
 	if err := units.Find(selSubordinate).One(&lDoc); err == mgo.ErrNotFound {
+		logger.Criticalf("ru.unitName: %s, principleMachineID: %q", ru.unitName, principalMachineID)
 		application, err := ru.st.Application(applicationname)
 		if err != nil {
 			return nil, "", err
 		}
-		_, ops, err := application.addUnitOps(unitName, AddUnitParams{}, nil)
+		_, ops, err := application.addUnitOps(unitName, AddUnitParams{
+			machineID: principalMachineID,
+		}, nil)
 		return ops, "", err
 	} else if err != nil {
 		return nil, "", err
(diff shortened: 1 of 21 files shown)


howbazaar:
Manual testing: 

Deployed a model with subordinate on 2.7.4, and migrated to 2.8. Used juju-db to check the database to ensure that the machine ID was set for the subordinate.

Deployed a model with subordinate on 2.7.4, upgraded controller, and used juju-db to check the database to ensure upgrade steps added machine ID.

state/relationunit.go
```diff
@@ -244,17 +244,31 @@ func (ru *RelationUnit) subordinateOps() ([]txn.Op, string, error) {
...
+	// that for the subordinate. It is worthwhile noting that if the unit is
+	// in a CAAS model, there are no machines.
+	principal, err := ru.st.Unit(ru.unitName)
+	if err != nil {
+		return nil, "", errors.Annotate(err, "unable to load principal unit")
+	}
+	// We dont care just now if the machine isn't assigned, as CAAS models
+	// will return that error. For IAAS models, the machine *should* always
+	// be assigned before it is able to enter scope.
+	principalMachineID, _ := principal.AssignedMachineId()
```
wallyworld: It's dangerous to ignore all errors as the implementation might change and introduce a bug.
It is safer to use the ShouldBeAssigned() helper. This would also help document all cases where we make a caas vs iaas decision to help the refactoring we'd ideally like to do later.

state/unit.go
```diff
@@ -1178,6 +1178,7 @@ func (u *Unit) relations(predicate relationPredicate) ([]*Relation, error) {
 
 // DeployerTag returns the tag of the agent responsible for deploying
 // the unit. If no such entity can be determined, false is returned.
+// XXX: this method is only used in tests, we should probably kill it.
```
wallyworld: // TODO
instead of XXX

state/unit.go
```diff
@@ -2162,6 +2143,9 @@ func assignContextf(err *error, unitName string, target string) {
 // AssignToMachine assigns this unit to a given machine.
 func (u *Unit) AssignToMachine(m *Machine) (err error) {
 	defer assignContextf(&err, u.Name(), fmt.Sprintf("machine %s", m))
+	if u.doc.Principal != "" {
+		return fmt.Errorf("unit is a subordinate")
```
wallyworld: errors.NotSupportedf()

state/unit.go
```diff
@@ -2768,7 +2752,18 @@ func (u *Unit) findCleanMachineQuery(requireEmpty bool, cons *constraints.Value)
...
+func (u *Unit) assignToCleanMaybeEmptyMachine(requireEmpty bool) (_ *Machine, err error) {
+	context := "clean"
+	if requireEmpty {
+		context += ", empty"
+	}
+	context += " machine"
+	defer assignContextf(&err, u.Name(), context)
+
+	if u.doc.Principal != "" {
+		err = fmt.Errorf("unit is a subordinate")
```
wallyworld: ditto

state/relationunit.go
howbazaar: Added a ShouldBeAssigned check, but still ignore the error because of the amount of test fallout.

state/unit.go
howbazaar: Removed the method instead.

state/unit.go
howbazaar: This is out of scope for the current work as it changes the structure of the error message. We should make a separate pass through the unit code to update errors. There are many non-current things in here.

howbazaar:
$$merge$$

PR #11351, opened 2020-03-24
Optional k8s charm pv

wallyworld:
### Checklist

There's a change to api parameters but it's for a controller facade.

----

## Description of change

A k8s charm can now signal that it does not require a PV to be created for charm state storage. It does this by setting min-juju-version=2.8.0. Existing charms which do require PV storage are still catered for, and if a charm is upgraded, any PV is retained.

The first commit factors out the min version check to a common function. It has been tweaked to allow a charm specifying a major.minor.0 version to work with a beta/rc (to allow testing).

## QA steps

Deploy a k8s charm
Edit the charm to add min-juju-version=2.8.0
upgrade the charm and check the PV and PVC are still used
deploy the edited charm again as a different application
check that no operator PV is created

### DIFF
--- a/caas/broker.go
+++ b/caas/broker.go
@@ -371,9 +371,9 @@ type OperatorConfig struct {
 	// Version is the Juju version of the operator image.
 	Version version.Number
 
-	// CharmStorage defines parameters used to create storage
-	// for operators to use for charm state.
-	CharmStorage CharmStorageParams
+	// CharmStorage defines parameters used to optionally
+	//create storage for operators to use for charm state.
+	CharmStorage *CharmStorageParams
 
 	// AgentConf is the contents of the agent.conf file.
 	AgentConf []byte
--- a/caas/broker.go
+++ b/caas/broker.go
@@ -372,7 +372,7 @@ type OperatorConfig struct {
 	Version version.Number
 
 	// CharmStorage defines parameters used to optionally
-	//create storage for operators to use for charm state.
+	// create storage for operators to use for charm state.
 	CharmStorage *CharmStorageParams
 
 	// AgentConf is the contents of the agent.conf file.
--- a/caas/kubernetes/provider/operator.go
+++ b/caas/kubernetes/provider/operator.go
@@ -215,12 +215,47 @@ func (k *kubernetesClient) EnsureOperator(appName, agentPath string, config *caa
 	podWithoutStorage := pod
 
 	numPods := int32(1)
-	statefulset := &apps.StatefulSet{
+	operatorPvc, err := k.operatorVolumeClaim(appName, operatorName, config.CharmStorage)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if operatorPvc != nil {
+		logger.Debugf("using persistent volume claim for operator %s: %+v", appName, operatorPvc)
+		statefulset := &apps.StatefulSet{
+			ObjectMeta: v1.ObjectMeta{
+				Name:        operatorName,
+				Labels:      labels,
+				Annotations: annotations.ToMap()},
+			Spec: apps.StatefulSetSpec{
+				Replicas: &numPods,
+				Selector: &v1.LabelSelector{
+					MatchLabels: labels,
+				},
+				Template: core.PodTemplateSpec{
+					ObjectMeta: v1.ObjectMeta{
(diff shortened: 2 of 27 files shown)


caas/kubernetes/provider/operator.go
```diff
@@ -197,44 +197,7 @@ func (k *kubernetesClient) EnsureOperator(appName, agentPath string, config *caa
 		}
 	}
 
-	// Set up the parameters for creating charm storage.
```
wallyworld: This code is moved to a new separate operatorVolumeClaim() function below

wallyworld:
[k8s charms do not require storage](https://trello.com/c/vhhzwBXc/1079-k8s-charms-do-not-require-storage)

caas/broker.go
```diff
@@ -371,9 +371,9 @@ type OperatorConfig struct {
 	// Version is the Juju version of the operator image.
 	Version version.Number
 
-	// CharmStorage defines parameters used to create storage
-	// for operators to use for charm state.
-	CharmStorage CharmStorageParams
+	// CharmStorage defines parameters used to optionally
+	//create storage for operators to use for charm state.
```
hpidcock: //(space)create storage

caas/kubernetes/provider/providerconfig.go
```diff
@@ -7,16 +7,34 @@ import (
...
 	"github.com/juju/schema"
+	"github.com/juju/version"
 	"gopkg.in/juju/environschema.v1"
 
 	"github.com/juju/juju/environs/config"
 )
 
 const (
+	// WorkloadStorageKey is the model config attribute used to specify
+	//the storage class for provisioning workload storage.
```
hpidcock: I need some space here

caas/kubernetes/provider/providerconfig.go
```diff
@@ -7,16 +7,34 @@ import (
...
 	WorkloadStorageKey = "workload-storage"
+
+	// OperatorStorageKey is the model config attribute used to specify
+	//the storage class for provisioning operator storage.
```
hpidcock: and here

caas/kubernetes/provider/operator.go
```diff
@@ -269,20 +231,80 @@ func (k *kubernetesClient) EnsureOperator(appName, agentPath string, config *caa
...
+func (k *kubernetesClient) operatorVolumeClaim(appName, operatorName string, storageParams *caas.CharmStorageParams) (*core.PersistentVolumeClaim, error) {
+	// We may no longer need storage for charms, but if the charm has previously been deployed
+	// with storage, we need to retain that.
+	operatorVolumeClaim := "charm"
+	if isLegacyName(operatorName) {
+		operatorVolumeClaim = fmt.Sprintf("%v-operator-volume", appName)
+	}
+	if storageParams == nil {
+		existingClaim, err := k.getPVC(operatorVolumeClaim)
+		if err != nil && !errors.IsNotFound(err) {
```
hpidcock: Ugh I'd prefer this
```
if errors.IsNotFound(err) {
    logger.Debugf("found existing volume claim for operator %s", operatorName)
    return nil, nil
} else if err != nil {
    return nil, errors.Annotatef(err, "getting operator volume claim")
}
return existingClaim, nil
```

caas/kubernetes/provider/operator_test.go
```diff
@@ -621,6 +621,138 @@ func (s *K8sBrokerSuite) TestEnsureOperatorUpdate(c *gc.C) {
 	}
 }
 
+func (s *K8sBrokerSuite) TestEnsureOperatorNoStorage(c *gc.C) {
```
hpidcock: This test is TestEnsureOperatorNoStorageKeepsPVC

Need another to ensure it handles no existing PVC and doesn't create one.

wallyworld: yeah, just added a test for that. without a pvc, gets deployed as a deployment

wallyworld:
$$merge$$

hpidcock:
I don't think we should be using deployments for the operator. Statefulsets have guarantees with regards to the number of pods running. I don't believe the operator/uniter logic will play well at all with two pods running the operator.

PR #11352, opened 2020-03-24
Fix 1867783 Charm deployment hangs with 2.8

hpidcock:
## Fix 1867783 Charm deployment hangs with 2.8

- Fixes stateful set creation to no longer immediately
update the newly created statefulset.
- Fixes OperatorExists method to check for all created resources.
- Fixes config map RAW/WAW race condition.
- Adds tests for OperatorExists which was missing before.
- Fixes tag passed to IssueOperatorCertificate facade

## QA steps

before:
1. bootstrap k8s
2. deploy an app
3. remove the app
4. repeat from 2 until deployment hangs

see bug for more info.

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1867783

### DIFF
--- a/caas/kubernetes/provider/operator.go
+++ b/caas/kubernetes/provider/operator.go
@@ -295,19 +295,159 @@ func (k *kubernetesClient) validateOperatorStorage() (string, error) {
 // OperatorExists indicates if the operator for the specified
 // application exists, and whether the operator is terminating.
 func (k *kubernetesClient) OperatorExists(appName string) (caas.OperatorState, error) {
-	var result caas.OperatorState
 	operatorName := k.operatorName(appName)
+	exists, terminating, err := k.operatorStatefulSetExists(appName, operatorName)
+	if err != nil {
+		return caas.OperatorState{}, errors.Trace(err)
+	}
+	if exists || terminating {
+		if terminating {
+			logger.Tracef("operator %q exists and is terminating")
+		} else {
+			logger.Tracef("operator %q exists")
+		}
+		return caas.OperatorState{Exists: exists, Terminating: terminating}, nil
+	}
+	checks := []struct {
+		label string
+		check func(appName string, operatorName string) (bool, error)
+	}{
+		{"rbac", k.operatorRBACResourcesRemaining},
+		{"config map", k.operatorConfigMapExists},
+		{"configurations config map", k.operatorConfigurationsConfigMapExists},
+		{"service", k.operatorServiceExists},
+		{"secret", k.operatorSecretExists},
+		{"deployment", k.operatorDeploymentExists},
+		{"pods", k.operatorPodExists},
+	}
+	for _, c := range checks {
+		exists, err := c.check(appName, operatorName)
+		if err != nil {
+			return caas.OperatorState{}, errors.Annotatef(err, "%s resource check", c.label)
+		}
+		if exists {
+			logger.Debugf("operator %q exists and is terminating due to dangling %s resource(s)", c.label)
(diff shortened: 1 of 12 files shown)


hpidcock:
[Charm deployment hangs with 2.8](https://trello.com/c/RDjMMyG0/1078-charm-deployment-hangs-with-28)

caas/kubernetes/provider/operator.go
```diff
@@ -295,19 +295,158 @@ func (k *kubernetesClient) validateOperatorStorage() (string, error) {
...
+		{"pods", k.operatorPodExists},
+	}
+	for _, c := range checks {
+		exists, err := c.check(appName, operatorName)
+		if err != nil {
+			return caas.OperatorState{}, errors.Annotatef(err, "%s resource check", c.label)
+		}
+		if exists {
+			logger.Debugf("operator %q exists and is terminating due to dangling %s resource(s)", c.label)
+			return caas.OperatorState{true, true}, nil
```
wallyworld: caas.OperatorState{Exists: true, Terminating: true},

for consistency

caas/kubernetes/provider/operator.go
```diff
@@ -295,19 +295,158 @@ func (k *kubernetesClient) validateOperatorStorage() (string, error) {
...
+		}
+	}
+	return caas.OperatorState{}, nil
+}
+
+func (k *kubernetesClient) operatorStatefulSetExists(appName string, operatorName string) (bool, bool, error) {
```
wallyworld: either a comment or name the return params

(bool, bool) doesn't indicate what they represent

caas/kubernetes/provider/operator.go
```diff
@@ -295,19 +295,158 @@ func (k *kubernetesClient) validateOperatorStorage() (string, error) {
...
+	}
+	exists := true
+	terminating := operator.DeletionTimestamp != nil
+	return exists, terminating, nil
+}
+
+func (k *kubernetesClient) operatorRBACResourcesRemaining(appName string, operatorName string) (bool, error) {
+	_, err := k.getServiceAccount(operatorName)
+	if errors.IsNotFound(err) {
+	} else if err != nil {
```
wallyworld: this looks "incomplete" even though correct
perhaps a comment in the body of the if

caas/kubernetes/provider/operator.go
```diff
@@ -295,19 +295,158 @@ func (k *kubernetesClient) validateOperatorStorage() (string, error) {
...
+		return false, errors.Trace(err)
+	} else {
+		return true, nil
+	}
+	_, err = k.getRole(operatorName)
+	if errors.IsNotFound(err) {
+	} else if err != nil {
```
wallyworld: etc

caas/kubernetes/provider/operator.go
```diff
@@ -590,6 +730,7 @@ func operatorConfigMap(appName, name string, labels, annotations map[string]stri
 			Name:        name,
 			Labels:      labels,
 			Annotations: annotations,
+			Generation:  config.ConfigMapGeneration,
```
wallyworld: what happens when there's a Generation mismatch?
do we get an error from k8s that causes us to need to try again?
I can't see where we do that

hpidcock: Handled like any other error. K8s would return a Conflict error, because the generation doesn't match what was read.

hpidcock: The caasoperatorprovisioner retries already.

hpidcock:
$$merge$$

PR #11353, opened 2020-03-24
Remove obsolete space/subnet commands

SimonRichardson:
## Description of change

The update-space command was originally implemented for the MVP, but was
never implemented at the api client or api facade level. This means if
you actually tried to use it, it would panic for method not found as no
client ever implemented the UpdateSpace method.

Removing this also helps when we want to implement MoveToSpace as there
seemed to be some overlap between the ideas of the two. Instead we will
opt for a more pure MoveCommand.

I believe there could be more clean up as I go, but this is the first
stab at that.

## QA steps

Nothing, it never worked before. So removing dead code :dagger: :smile: 

## Documentation changes

We should check that this isn't in the documentation @timClicks


SimonRichardson:
https://trello.com/c/L6MgiYg0/2435-juju-juju-remove-update-spaces-commands-deadcode

SimonRichardson:
$$merge$$

PR #11356, opened 2020-03-24
[dev] Relation departed should note who goes away

achilleasa:
## Description of change

This PR enables charms to better react to `relation-departed` hooks by including additional information about the unit that is going away. The resolver figures this out by querying the life for the local unit and checking whether it is `Alive` or not prior to emitting a relation-departed hook. Depending on the outcome of the comparison, it populates the `DepartedUnit` field in the emitted `hook.Info` value.

When the `relation-departed` hook is executed, the information about the departing unit is exposed via the `JUJU_DEPARTING_UNIT` envvar. Interested charms can compare its value to `JUJU_REMOTE_UNIT` to figure out who is going away. An example of how to do this can be found in the integration test that accompanies this PR. 

### For peer relations

```console
$ juju deploy ./testcharms/charm-repo/quantal/all-hooks -n 3

# Wait for all units to come up and then remove all-hooks/2
$ juju remove-unit all-hooks/2

# Check logs
$  juju debug-log --replay --include unit-all-hooks-0 | grep 'departing the relation'
unit-all-hooks-0: 09:10:17 DEBUG unit.all-hooks/0.self-relation-departed Remote unit all-hooks/2 is departing the relation

$ juju debug-log --replay --include unit-all-hooks-1 | grep 'departing the relation'
unit-all-hooks-1: 09:10:17 DEBUG unit.all-hooks/1.self-relation-departed Remote unit all-hooks/2 is departing the relation

# Note: all-hooks/2 gets 2 relation-departed hooks for the other 2 units but in both cases it is 
# the one departing hence the following output
$  juju debug-log --replay --include unit-all-hooks-2 | grep 'departing the relation'
unit-all-hooks-2: 09:10:17 DEBUG unit.all-hooks/2.self-relation-departed Local unit is departing the relation
unit-all-hooks-2: 09:10:17 DEBUG unit.all-hooks/2.self-relation-departed Local unit is departing the relation


```

### For non-peer relations

```console
$ juju model-config logging-config='<root>=ERROR;juju.worker.uniter.operation=DEBUG'
$ juju deploy wordpress -n2 && juju deploy mysql && juju relate wordpress mysql

# Wait for all units to come up and this time remove the mysql *application*
$ juju remove-application mysql

# Check logs
$ juju debug-log --replay --include unit-wordpress-0 | grep 'committing' | grep 'departee'
unit-wordpress-0: 09:18:00 DEBUG juju.worker.uniter.operation committing operation "run relation-departed (3; unit: mysql/0, departee: mysql/0) hook"

$ juju debug-log --replay --include unit-wordpress-1 | grep 'committing' | grep 'departee'                                                                                            [NORMAL]
unit-wordpress-1: 09:18:12 DEBUG juju.worker.uniter.operation committing operation "run relation-departed (3; unit: mysql/0, departee: mysql/0) hook"

# mysql gets relation-departed for both wordpress units but in both cases it's the one departing
$ juju debug-log --replay --include unit-mysql-0 | grep 'committing' | grep 'departee'
unit-mysql-0: 09:52:52 DEBUG juju.worker.uniter.operation committing operation "run relation-departed (2; unit: wordpress/0, departee: mysql/0) hook"
unit-mysql-0: 09:52:52 DEBUG juju.worker.uniter.operation committing operation "run relation-departed (2; unit: wordpress/1, departee: mysql/0) hook"
```

## Documentation changes

We need to update our docs to highlight the presence of the new envvar.

## Bug reference
https://bugs.launchpad.net/juju-core/+bug/1417874


achilleasa:
!!build!! (lookup streams.canonical.com: Temporary failure in name resolution)

achilleasa:
!!build!! (more dns issues)

achilleasa:
!!build!! (even more dns issues)

worker/uniter/hook/hook.go
```diff
@@ -41,6 +41,10 @@ type Info struct {
 
 	// StorageId is the ID of the storage instance relevant to the hook.
 	StorageId string `yaml:"storage-id,omitempty"`
+
+	// DepartingUnit is the name of the unit that goes away. It is only set
+	// when Kind indicates a relation-departed hook.
+	DepartingUnit string `yaml:"departee,omitempty"`
```
jameinel: I wonder if we also want to optionally set DepartingApplication if we know that all units of the application will be going away. (a departing unit needs to remove themselves from the application config, and rebalance the cassandra cluster, etc. but a departing application just shuts it all down.)

worker/uniter/relation/resolver.go
```diff
@@ -176,12 +176,28 @@ func (r *relationsResolver) nextHookForRelation(localStateDir *StateDir, remote
...
+			// away. Note that if the app is removed, the unit will
+			// still be alive but its parent app will by dying.
+			localUnitLife, localAppLife, err := r.stateTracker.LocalUnitAndApplicationLife()
+			if err != nil {
+				return hook.Info{}, errors.Trace(err)
+			}
+
+			var departee = unitName
+			if localUnitLife != life.Alive || localAppLife != life.Alive {
+				departee = r.stateTracker.LocalUnitName()
```
jameinel: so I see that you set 'departee' to the local unit, but I don't see where you set it to the remote unit.
Am I missing an else clause somewhere?

worker/uniter/relation/resolver_test.go
```diff
@@ -514,7 +539,7 @@ func (s *relationResolverSuite) TestHookRelationChangedSuspended(c *gc.C) {
 	relationsResolver := relation.NewRelationResolver(r, nil)
 	op, err := relationsResolver.NextOp(localState, remoteState, &mockOperations{})
 	c.Assert(err, jc.ErrorIsNil)
-	assertNumCalls(c, &numCalls, numCallsBefore)
+	assertNumCalls(c, &numCalls, numCallsBefore+2) // Refresh/Life calls made by the resolver prior to emitting a RelationDeparted hook
```
jameinel: Would this be clearer called out as a separate line:
// relationsResolver.NextOp will notice it wants to fire relation-departed, and makes 2 calls
// to the controller to check if it is the unit that is departing.
expectedNumCalls := numCallsBefore + 2

achilleasa:
$$merge$$

worker/uniter/relation/statetracker.go
```diff
@@ -481,3 +488,23 @@ func (r *relationStateTracker) Name(id int) (string, error) {
...
+
+// LocalUnitName returns the name for the local unit.
+func (r *relationStateTracker) LocalUnitName() string {
+	return r.unit.Name()
+}
+
+// LocalUnitAndApplicationLife returns the life values for the local unit and
+// application.
+func (r *relationStateTracker) LocalUnitAndApplicationLife() (life.Value, life.Value, error) {
+	if err := r.unit.Refresh(); err != nil {
```
jameinel: It seems odd to me that we need to call Refresh here. Is the unit actually changing since we started the loop? It feels a bit like we're sneaking a peek at the future, when things like the Life of the unit changing should already be part of the Hook Context. (it was already looking for state-based changes on all the objects it is responsible for.)

worker/uniter/runner/context/env_test.go
```diff
@@ -110,24 +110,25 @@ func (s *EnvSuite) getContext(newProxyOnly bool) (ctx *context.HookContext, expe
...
+		UnitName:            "this-unit/123",
+		MeterCode:           "PURPLE",
+		MeterInfo:           "proceed with care",
+		SLALevel:            "essential",
+		AvailZone:           "some-zone",
+		APIAddresses:        []string{"he.re:12345", "the.re:23456"},
+		LegacyProxySettings: legacyProxy,
+		JujuProxySettings:   jujuProxy,
+		MachineTag:          names.NewMachineTag("42"),
+	}), expected
```
jameinel: This is just putting the names here, right? (i find it *way* easier to understand with the fields)' but I can't tell if you added anything to it.

worker/uniter/relation/resolver.go
achilleasa: L189 inits the var with the _remote_ unit's name. Given (AFAICT from the code) that we do not have permissions to query the life for the remote unit/app, we instead query the life for the local unit/app and override the departee to the local unit on L191.

worker/uniter/relation/resolver_test.go
achilleasa: I am really keen on spending some time on the next cycle (if we have enough capacity) to rewrite all these tests to use mocks. Then we won't have to resort to this hacky approach where adding a new call anywhere in our logic means that we need to bump the expected call count in a bunch of unrelated tests...

worker/uniter/relation/statetracker.go
achilleasa: I bumped into this while doing my initial tests and this is my workaround (IIRC I briefly described this in the commit log).

So basically, removing the unit:
- causes the relation to be removed from the remote
- the above code detects the removal but the local unit has not been marked as dying yet even though it is marked so in the db.

Forcing a refresh works around this issue. It can be verified by removing the refresh and running the first QA scenario where it will mistakingly report that the remote unit is going away.

worker/uniter/runner/context/env_test.go
achilleasa: This is a purely cosmetic drive-by fix.  All it does is to replace an insanely long arg list expected by `NewModelHookContext` with a struct just in case we want to add more fields in the future.

worker/uniter/hook/hook.go
achilleasa: IIRC this was raised in a standup and we punted it for now. While we can easily get this information for the local app we cannot get it for the remote one due to lack of permissions.

worker/uniter/relation/statetracker.go
achilleasa: The local unit instance eventually syncs up but that happens after this code runs.

worker/uniter/runner/context/env_test.go
jameinel: great, that's what it looked like, and I really appreciate the change.

worker/uniter/relation/resolver.go
jameinel: yeah, I just misread 189. Thanks for clarifying.

PR #11358, opened 2020-03-25
[2.7] fix infinite loop bug in bundlechanges

achilleasa:
## Description of change

If an overlay attempts to create an offer for an application that has already been deployed, `bundlechanges.FromData` would enter an infinite loop looking for a dependency requirement (the deployment) that doesn't exist.

To replicate:

```console
$ cat bundle.yaml
series: trusty
applications:
  apache2:
    charm: cs:apache2-26
    num_units: 1

$ cat overlay.yaml
applications:
  apache2:
    offers:
      my-offer:
        endpoints:
        - apache-website
        acl:
          admin: admin

$ juju deploy  ./bundle.yaml

# This command gets stuck forever with existing juju clients
$ juju deploy --debug ./bundle.yaml --overlay overlay.yaml
```

To avoid the infinite loop, the offer-handling code in this package will obtain a list of applications already deployed to the controller (and present in the bundle being deployed) and check whether the application is already deployed; if so, the add-offer change does not have any dependencies and can be executed at any time.

# QA steps

Running the above commands with a juju client from snap/HEAD should block; using a juju client from this PR should work as expected.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -484,11 +484,11 @@
   revision = "b99631de12cf04a906c1d4e4ec54fb86eae5863d"
 
 [[projects]]
-  digest = "1:ef49e9c114183e3d64d2ed7602dfb175213eb1ba5edcf822376104c4ba9e6e42"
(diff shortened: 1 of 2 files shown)


hmlanigan:
$$merge$$

PR #11360, opened 2020-03-26
Do not deploy operator as a deployment even with no storage

wallyworld:
## Description of change

Even when a new charm is deployed and doesn't need a PV, still deploy the operator as a statefulset.
Also a driveby fix for WatchOperator() selector.

## QA steps

bootstrap microk8s
deploy a charm with min-juju-version=2.8.0
check cluster


wallyworld:
$$merge$$

PR #11370, opened 2020-03-27
Update dependency for lumberjack to remove goroutine leak.

howbazaar:
Use juju's copy of the lumberjack repo, and bring in the change that ensures the goroutine is closed.

## QA steps

* deploy some units
* wait some time
* go and look at the goroutines for the unit and check lumberjack goroutines

There will be at most one, and none until it has done a rotation.

## Documentation changes

Internal only.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1865901

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1695,11 +1695,12 @@
   source = "github.com/juju/mgo"
 
 [[projects]]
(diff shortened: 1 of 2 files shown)


howbazaar:
$$merge$$

howbazaar:
$$merge$$ killed somehow in the middle of running tests

wallyworld:
$$merge$$

PR #11371, opened 2020-03-27
Move CallContext out of State

SimonRichardson:
## Description of change

The CallContext has no need to be in the state, it used to depend on 
state, but in reality, it doesn't. Supplying an interface that satisfies the
dependency requirements is enough to get what we want.

By removing this, we remove the dependency on state from within the
API server. We adhere to better vertical boundaries and use dependency
isolation to give what we only need. This is a different mindset, but
it's a worthy one, it just requires effort to get there.

This is a long road to do that and biting small chunks off as we go is
a worthwhile endeavour.

## QA steps

The tests pass.


SimonRichardson:
https://trello.com/c/gmrzRb9X/2447-move-state-callcontext-to-the-right-abstraction-level

SimonRichardson:
$$merge$$

PR #11373, opened 2020-03-30
Update presence recorder to handle race during server connection.

howbazaar:
During the start sequence of an apiserver, there is a potential race
between the agent connection events, and the event handling in the
presence worker. The pubsub forwarder worker is involved in this race
as well.

When pubsub message forwarding is established between two HA nodes
and event is published. The presence worker responds to this event
by asking the other side to let them know their understanding of
presence.

If an agent connection event is published on the controller after
a request for presence state has turned up, but before the presence
worker has responded, the event sequence on the other controller
will get the agent connection before the presence response. However
the presence worker will create the response before it has processed
the new agent connection. If this occurs, one side thinks that the
agent is there while the other doesn't know because it replaced
all that it knew about the other server with the update.

The fix for this is to have the presence recorder only remove
entries from the recorder that aren't alive. When pubsub forwarding
is disconnected, the worker will mark the other server as down
in its own recorder, which updates their state to missing. If
an agent connection event is received before the presence response
message, then we want to keep the alive connection information and
augment with the response from the other side.

## QA steps

Due to the racy nature of this, it is very hard to reproduce locally, but unit tests were updated
to match behaviour of the system.

## Documentation changes

This is internal behaviour only, and doesn't need any doc updates.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1866120

### DIFF
--- a/core/presence/presence.go
+++ b/core/presence/presence.go
@@ -281,10 +281,17 @@ func (r *recorder) UpdateServer(server string, connections []Value) error {
 		return errors.New("recorder not enabled")
 	}
 
+	// Alive is a map of connection IDs for connections that are alive
+	// to an index into the entries slice.
+	alive := make(map[uint64]int)
 	entries := make([]Value, 0, len(r.entries))
 	for _, value := range r.entries {
 		if value.Server != server {
 			entries = append(entries, value)
+		} else if value.Status == Alive {
+			pos := len(entries)
+			entries = append(entries, value)
+			alive[value.ConnectionID] = pos
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


howbazaar:
Targeting 2.7 just in case we do a 2.7.6 release.

core/presence/presence.go
```diff
@@ -281,20 +281,33 @@ func (r *recorder) UpdateServer(server string, connections []Value) error {
...
 		if value.Server != server {
 			return errors.Errorf("connection server mismatch, got %q expected %q", value.Server, server)
 		}
-		value.Status = Alive
-		value.LastSeen = r.clock.Now()
-		entries = append(entries, value)
+		// If the connection has already been recorded as alive,
+		// just update the timestamp, otherwise add it in.
+		if i, found := alive[value.ConnectionID]; found {
+			entries[i].LastSeen = r.clock.Now()
```
babbageclunk: Nice!

howbazaar:
$$merge$$

PR #11374, opened 2020-03-30
Add support for running k8s operator actions on the charm operator pod.

wallyworld:
### Checklist

This is an agent only facade.

## Description of change

The newly added k8s operator charms don't have a workload pod - their actions need to execute in the charm operator. To do this, the "workload-context" parameter for k8s operator actions is set to false, so that the uniter runs the action locally. Also, the Charm() api call is extended to return the deployment mode so that the uniter can avoid the mechanics of blocking actions for a workload pod startup which will never happen.

## QA steps

deploy both a "normal" charm and an operator charm on k8s and ensure actions run as expected.

### DIFF
--- a/worker/caasoperator/caasoperator_test.go
+++ b/worker/caasoperator/caasoperator_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/juju/clock/testclock"
 	"github.com/juju/errors"
+	"github.com/juju/juju/caas"
 	"github.com/juju/os/series"
 	"github.com/juju/testing"
 	jc "github.com/juju/testing/checkers"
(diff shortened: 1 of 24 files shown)


wallyworld:
[k8s operator actions](https://trello.com/c/UN4yP8mg/1083-k8s-operator-actions)

worker/caasoperator/caasoperator_test.go
```diff
@@ -13,6 +13,7 @@ import (
 
 	"github.com/juju/clock/testclock"
 	"github.com/juju/errors"
+	"github.com/juju/juju/caas"
```
hpidcock: wrong block

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #11377, opened 2020-03-31
Remove legacy presence implementation.

howbazaar:
The pubsub based presence was introduced in Juju 2.5, and became the default in 2.6.0 with a feature flag to turn the legacy implementation back on. While we did see the legacy-leases flag used, I'm not aware of the legacy-presence feature flag being used at all.

This PR removes the old presence implemenation that was based on agent pingers writing into the presence database.


howbazaar:
Only thing to add is a test for the upgrade step.

howbazaar:
Added a test for the upgrade step.

howbazaar:
$$merge$$

PR #11380, opened 2020-03-31
Forward port 2.7 fixes.

howbazaar:
Bring the fixes for the presence and unit agent memory leak into develop.


howbazaar:
$$merge$$

PR #11383, opened 2020-03-31
move State and SetState from uniter facade to common.UnitState

hmlanigan:
## Description of change
Move State and SetState from uniter facade to common.UnitState.  Done so that the same calls can be used by the metrics workers and the caas operator worker facades in upcoming changes.  Part of work to move uniter internal state from local files to the controller.

## QA steps
Same QA steps as #11339


hmlanigan:
make-check-juju unrelated, but in code I recently merged.  need to fix intermittent fail in upgradeStepsSuite.TestWriteUniterState.

hmlanigan:
The integration test is fixed with #11391 
The intermittent unit failure with #11392 

$$merge$$

PR #11384, opened 2020-04-01
upgrader: Remove checks in the upgrader to prevent downgrading

babbageclunk:
## Description of change

Previously the upgrader rejected a version change if the target minor or major version was less than the current version. This check has been removed to support rolling back the Juju version by restoring a backup taken before an upgrade.

The same change has been made in the caasupgrader worker - the upgrade process is different in k8s land but this check was essentially the same.

With these changes restoring a previous version of a controller automatically triggers the agents to download the required version of the agent, symlink it and update the upgradedToVersion in the agent.conf.

There was a mention of a bug in upgrading from 1.16 to 1.18: https://bugs.launchpad.net/juju-core/+bug/1299802
My understanding is that this behaviour isn't a concern any more because the unit agent no longer considers the machine agent's version; everything is driven by the model agent version instead.

## QA steps

* Bootstrap with Juju 2.7.
* Deploy some units.
* Take a backup of the controller.
```sh
juju create-backup -m controller
```
* Upgrade to 2.8 with this change, and upgrade the model as well.
```sh
juju upgrade-model
```
* Run juju-restore on the controller machine. At the moment this requires a version of juju-restore hacked to remove the check requiring the same version between controller and backup.
```sh
juju scp -m controller $(which juju-restore) 0:
juju scp -m controller juju-backup-whenever-it-is.tar.gz
juju ssh -m controller 0
# On machine
sudo grep statepassword /var/lib/juju/agents/machine-0/agent.conf
./juju-restore --username machine-0 --password thepassword --verbose juju-backup-whenever.tar.gz
```
* After restore the controller agent will start complaining about downgrading state from 2.8 to 2.7.4 - this is unnecessary because the db is already at 2.7.4, it's just because the agent.conf is wrong. Stop the agent and edit agent.conf to set upgradedToVersion back to 2.7.4:
```sh
$ sudo systemctl stop jujud-machine-0.service
$ sudo vi /var/lib/juju/agents/machine-0/agent.conf
$ sudo systemctl start jujud-machine-0.service
```
(Once the version downgrade changes are made to juju-restore this will happen automatically.)

* The unit and machine agents in the model will update their symlinks to point to the 2.7.4 binary and their agent.confs with the correct upgradedToVersion.
* Their log files will show that the right version of the binary is running.

## Documentation changes

No change to documentation.

## Bug reference

None

### DIFF
--- a/worker/caasupgrader/upgrader_test.go
+++ b/worker/caasupgrader/upgrader_test.go
@@ -121,6 +121,23 @@ func (s *UpgraderSuite) TestUpgraderDowngradePatch(c *gc.C) {
 	s.operatorUpgrader.CheckCall(c, 0, "Upgrade", "machine-0", s.upgraderClient.desired)
 }
 
+func (s *UpgraderSuite) TestUpgraderDowngradeMinor(c *gc.C) {
+	// We'll allow this for the case of restoring a backup from a
+	// previous juju version.
+	vers := version.MustParse("6.6.7")
+	s.PatchValue(&jujuversion.Current, vers)
+	s.upgraderClient.desired = version.MustParse("6.5.10")
+
+	u := s.makeUpgrader(c, names.NewMachineTag("0"))
+	workertest.CleanKill(c, u)
+
+	s.expectInitialUpgradeCheckNotDone(c)
+	c.Assert(s.upgraderClient.actual.Number, gc.DeepEquals, vers)
+	s.upgraderClient.CheckCallNames(c, "SetVersion", "DesiredVersion")
+	s.operatorUpgrader.CheckCallNames(c, "Upgrade")
+	s.operatorUpgrader.CheckCall(c, 0, "Upgrade", "machine-0", s.upgraderClient.desired)
+}
+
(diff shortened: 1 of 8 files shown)


babbageclunk:
### Checklist

worker/caasupgrader/upgrader_test.go
```diff
@@ -121,6 +121,23 @@ func (s *UpgraderSuite) TestUpgraderDowngradePatch(c *gc.C) {
 	s.operatorUpgrader.CheckCall(c, 0, "Upgrade", "machine-0", s.upgraderClient.desired)
 }
 
+func (s *UpgraderSuite) TestUpgraderDowngradeMinor(c *gc.C) {
```
wallyworld, line 124: So we have tests for downgrade patch and minor.
What about major?
We should allow that but only if >=2.

So we may need to retain a form of the code that was deleted?

babbageclunk: Yeah, that makes sense - I'll update it.

babbageclunk:
$$merge$$

babbageclunk:
$$merge$$

PR #11385, opened 2020-04-01
Remove the old apiserver metrics.

howbazaar:
When we first had metrics for the apiserver they were incorrectly using
the "api" subsystem. We added new ones that used "apiserver" as the subsystem
but left the old ones around while we transitioned.

The old ones are now removed. Originally we had a summary vec and a counter
not realising that the summary vec also gave a count. This is why we are removing
two and leaving one metric.

## QA steps

Bootstrap a controller.

```sh
juju bootstrap lxd test
juju exec -m controller juju_metrics
```

Observe that the old metrics starting "juju_api_" are no longer there, but the "juju_apiserver_" ones continue to count the api calls.

## Documentation changes

I don't think we actually talk about the metrics Juju exposes anywhere. We probably should.


wallyworld:
$$merge$$

howbazaar:
$$merge$$

PR #11392, opened 2020-04-01
Fix intermittent failure in upgradeStepsSuite.TestWriteUniterState

hmlanigan:
## Description of change

Fix intermittent failure in TestWriteUniterState.

DeepEquals doesn't work on []params.SetUnitStateArg due to potential
for different order.  SameContents does not work with pointers.  Use
DeepEquals on one SetUnitStateArg after verifying both have the same
tag.  Ensure the two slices have the same length first.

## QA steps

```sh
$ cd api/upgradesteps/
# run the juju stress test for a count of 100 or more
```


hmlanigan:
$$merge$$

PR #11396, opened 2020-04-02
Remove params.StateServingInfo and state.StateServingInfo

wallyworld:
## Description of change

The last remaining dependency of the state package on apiserver/params was the use of params.StateServingInfo. Plus, the state package itself declared its own copy. The type is generic and associated with a controller so a new copy of the struct was defined in the existing top level controller package.

There was also an upgrade error in state that just used a string from params that was also fixed.

A lot of the churn here is regeneration of mocks.

## QA steps

bootstrap a controller
enable ha
deploy a charm
check for errors

### DIFF
--- a/api/agent/state.go
+++ b/api/agent/state.go
@@ -13,6 +13,7 @@ import (
 	"github.com/juju/juju/api/common"
 	"github.com/juju/juju/api/common/cloudspec"
 	"github.com/juju/juju/apiserver/params"
+	"github.com/juju/juju/controller"
 	"github.com/juju/juju/core/instance"
 	"github.com/juju/juju/core/life"
 	"github.com/juju/juju/core/model"
@@ -60,10 +61,22 @@ func (st *State) getEntity(tag names.Tag) (*params.AgentGetEntitiesResult, error
 	return &results.Entities[0], nil
 }
 
-func (st *State) StateServingInfo() (params.StateServingInfo, error) {
+func (st *State) StateServingInfo() (controller.StateServingInfo, error) {
 	var results params.StateServingInfo
 	err := st.facade.FacadeCall("StateServingInfo", nil, &results)
-	return results, err
+	if err != nil {
+		return controller.StateServingInfo{}, errors.Trace(err)
+	}
+	return controller.StateServingInfo{
+		APIPort:           results.APIPort,
+		ControllerAPIPort: results.ControllerAPIPort,
+		StatePort:         results.StatePort,
+		Cert:              results.Cert,
+		PrivateKey:        results.PrivateKey,
+		CAPrivateKey:      results.CAPrivateKey,
+		SharedSecret:      results.SharedSecret,
+		SystemIdentity:    results.SystemIdentity,
+	}, nil
 }
 
 // IsMaster reports whether the connected machine
--- a/cmd/jujud/util/util.go
+++ b/cmd/jujud/util/util.go
@@ -12,10 +12,8 @@ import (
 	"github.com/juju/os/series"
 
 	"github.com/juju/juju/agent"
-	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/core/paths"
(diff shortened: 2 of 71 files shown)


api/agent/state.go
```diff
@@ -60,10 +61,22 @@ func (st *State) getEntity(tag names.Tag) (*params.AgentGetEntitiesResult, error
 	return &results.Entities[0], nil
 }
 
-func (st *State) StateServingInfo() (params.StateServingInfo, error) {
+func (st *State) StateServingInfo() (controller.StateServingInfo, error) {
```
howbazaar, line 64: This is much nicer.

cmd/jujud/util/util.go
```diff
@@ -231,17 +229,3 @@ func NewEnsureServerParams(agentConfig agent.Config) (mongo.EnsureServerParams,
 	}
 	return params, nil
 }
-
```
howbazaar, line 234: Yay.

controller/api.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package controller
+
+// StateServingInfo holds network/auth information needed by a controller.
```
howbazaar, line 6: Was wondering about the filename, but couldn't suggest a better one.

juju/testing/conn.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/juju/juju/cloud"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/controller"
+	jujucontroller "github.com/juju/juju/controller"
```
howbazaar, line 37: You don't need this import. The line above imports as "controller".

state/controller.go
```diff
@@ -246,23 +246,23 @@ func readRawControllerInfo(session *mgo.Session) (*ControllerInfo, error) {
 const stateServingInfoKey = "stateServingInfo"
 
 // StateServingInfo returns information for running a controller machine
-func (st *State) StateServingInfo() (StateServingInfo, error) {
+func (st *State) StateServingInfo() (jujucontroller.StateServingInfo, error) {
 	controllers, closer := st.db().GetCollection(controllersC)
 	defer closer()
 
-	var info StateServingInfo
+	var info jujucontroller.StateServingInfo
```
howbazaar, line 253: I think we should have a different struct defined in the state package that has the bson serialization tags.

Use that to get things in and out of the database, but return the controller.StateServingInfo as we are now.

state/state.go
```diff
@@ -103,26 +102,6 @@ type State struct {
 	CloudImageMetadataStorage cloudimagemetadata.Storage
 }
 
-// StateServingInfo holds information needed by a controller.
```
howbazaar, line 106: I feel that we should have the document type that is stored in the database defined here in state. But expose to the world the controller.StateServingInfo

controller/api.go
wallyworld: i thought about state.go and ran away quickly :-)

state/state.go
wallyworld: yup

wallyworld:
$$merge$$

state/initialize.go
```diff
@@ -247,7 +247,7 @@ func Initialize(args InitializeParams) (_ *Controller, err error) {
 			C:      controllersC,
 			Id:     stateServingInfoKey,
 			Assert: txn.DocMissing,
-			Insert: &StateServingInfo{},
+			Insert: &controller.StateServingInfo{},
```
howbazaar, line 250: You probably want to change this one too.

wallyworld:
$$merge$$

wallyworld:
[remove state dependency on params](https://trello.com/c/HhnFiXrl/1089-remove-state-dependency-on-params)

wallyworld:
$$merge$$

PR #11399, opened 2020-04-03
Ensure specified region is always used for add-model

wallyworld:
## Description of change

When using
juju add-model somecloud/someregion --credential mine

id the credential did not exist locally and was just in the controller, we would set the cloud region to "" instead of using the one specifed.

## QA steps

bootstrap an aws controller
delete the local credential
add-model aws/us-west-2 --credential mine

ensure the model is created in us-west-2 and not the default region (see bug)

## Bug reference

https://bugs.launchpad.net/juju/+bug/1849916


wallyworld:
[jaas add-model doesn't respect region](https://trello.com/c/ani7QdF7/1090-jaas-add-model-doesnt-respect-region)

wallyworld:
$$merge$$

PR #11401, opened 2020-04-03
[dev] Rename max-uniter-state-size to max-agent-state-size

achilleasa:
This PR renames the controller config option for setting a limit on the internal uniter state (originally introduced by #11390 ) to make the configuration setting a bit more intuitive and congruent with future work to unify the various agents.


achilleasa:
$$merge$$

achilleasa:
@wallyworld: renamed option as suggested.

PR #11403, opened 2020-04-03
unit state ops must merge relation data instead of replace.

hmlanigan:
## Description of change

Uniter relation state data is updated on a per relation basis, not as an entire chunk.  This change will be utilized by a future PR to move the uniter's relation internal state from local files to the controller.

## QA steps

There should be no change to normal functionality.

### DIFF
--- a/state/unit_ops.go
+++ b/state/unit_ops.go
@@ -66,7 +66,6 @@ func (op *unitSetStateOperation) buildTxn(attempt int) ([]txn.Op, error) {
 		if err != nil {
 			return nil, errors.Trace(err)
 		}
-
 		return []txn.Op{unitNotDeadOp, {
 			C:      unitStatesC,
 			Id:     unitGlobalKey,
@@ -82,6 +81,7 @@ func (op *unitSetStateOperation) buildTxn(attempt int) ([]txn.Op, error) {
 	} else if len(setFields) <= 0 && len(unsetFields) <= 0 {
 		return nil, jujutxn.ErrNoOperations
 	}
+
 	updateFields := bson.D{}
 	if len(setFields) > 0 {
 		updateFields = append(updateFields, bson.DocElem{"$set", setFields})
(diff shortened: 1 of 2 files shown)


hmlanigan:
just added new arg to a few test needed after rebased to develop picking up quota changes.

state/unit_ops.go
```diff
@@ -189,7 +189,7 @@ func (op *unitSetStateOperation) fields(currentDoc unitStateDoc) (bson.D, bson.D
 		if len(rState) == 0 {
 			unsetFields = append(unsetFields, bson.DocElem{Name: "relation-state"})
 		} else if matches := currentDoc.relationStateMatches(rState); !matches {
-			setFields = append(setFields, bson.DocElem{"relation-state", rState})
+			setFields = append(setFields, bson.DocElem{"relation-state", mergeRelations(currentDoc.RelationState, rState)})
```
achilleasa, line 192: You need to pull mergeRelations result into a var and pass that to the Check call below instead of `rState`

state/unit_ops.go
```diff
@@ -230,5 +230,25 @@ func (op *unitSetStateOperation) getUniterStateQuotaChecker() quota.Checker {
 	)
 }
 
+func mergeRelations(currentRelations map[string]string, newRelations map[string]string) map[string]string {
+	// For keys in newRelations:
+	// if key not in currentDoc, add key/value
+	// if key in currentDoc, if value == "", remove
+	// if key in currentDoc, if value != "", replace
+	for id, value := range newRelations {
```
achilleasa, line 238: Can you please rename `value` to `newValue`; it makes the logic below easier to follow

state/unit_ops.go
hmlanigan: Why did we include the uniter internal state data in the quota? Isn't there a potential to stop the unit from working?

state/unit_ops.go
hmlanigan: done

state/unit_ops.go
hmlanigan: make the change

achilleasa: As per #11390 we have two quotas, one for the charm state and one for the uniter state (uniter, relations, storage, meter status).  The code in this method uses separate quota checkers for applying the quota limits. Because `SetState` is not called at the same time for *both* charm and uniter state, charm quota checks can independently fail without affecting the storage of the uniter's state.

hmlanigan:
$$merge$$

PR #11410, opened 2020-04-06
Expect deployment to be on unit 0

SimonRichardson:
## Description of change

The following changes test when running in HA mode that the deployment
lands where we expect it to do so.

## QA steps

```sh
cd tests && ./main.sh manual
```

### DIFF
--- a/tests/suites/manual/deploy_manual.sh
+++ b/tests/suites/manual/deploy_manual.sh
@@ -163,7 +163,7 @@ EOF
 
     juju deploy percona-cluster
 
-    wait_for "percona-cluster" "$(idle_condition "percona-cluster")"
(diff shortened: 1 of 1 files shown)


tests/suites/manual/deploy_manual.sh
```diff
@@ -163,7 +163,7 @@ EOF
 
     juju deploy percona-cluster
 
-    wait_for "percona-cluster" "$(idle_condition "percona-cluster")"
+    wait_for "percona-cluster" "$(idle_condition "percona-cluster" 0 0)"
```
SimonRichardson, line 166: Ensure that we land in unit 0 when deploying.

SimonRichardson:
$$merge$$

PR #11405, opened 2020-04-03
[dev] move meter status state to the controller

achilleasa:
## Description of change

This PR is part of the ongoing work for migrating the unit-agent state to the controller. Here, we move the state of the `meter-status` worker to the controller.

The meterstatus worker is special in that it actually consists of two workers: a connected worker that can talk to the controller (e.g. to watch things or persist its state) and an isolated worker that runs when no connection to the controller is available.

As our goal is to (eventually) persist the metric status state to the controller without losing any data, the isolated worker still uses a disk-backed state store mechanism. When starting, the connected worker will first check whether there is any state present on disk (either because the isolated worker was briefly running OR we upgraded from an older juju version which always stored the state on disk). If that's the case, then the worker will persist the found state to the controller and then delete it from disk.

This step also serves as an auto-magic upgrade step for the worker so we don't need a dedicated update step for this.

## QA steps

*Please replace with how we can verify that the change works?*

## Automatic state migration when upgrading

```console
# Bootstap 2.7.x controller from snap
$ /snap/bin/juju bootstrap lxd test --no-gui

$ juju model-config logging-config='<root>=DEBUG'
$ juju deploy ./testcharms/charm-repo/bionic/metered-multi-series --series bionic

# Wait for the unit to start
$ juju debug-log --replay | grep "meter-status"
unit-metered-multi-series-9: 20:29:06 INFO juju.worker.meterstatus skipped "meter-status-changed" hook (missing)

# Then upgrade to this PR HEAD
$ juju upgrade-controller --rebuild-agent
$ juju upgrade-model

# Watch the logs for the meter state migration notice
$  juju debug-log --replay | grep located
unit-metered-multi-series-0: 16:37:15 INFO juju.worker.meterstatus located locally persisted worker state; migrating to the controller

# Verify state in juju db
$ juju:PRIMARY> db.unitstates.find({},{"meter-status-state": 1}).pretty()
{
        "_id" : "d30eb8a4-b0ed-4e0b-839d-2e01b1919a82:u#metered-multi-series/0#charm",
        "meter-status-state" : "status-code: RED\nstatus-info: \"\"\n"
}
```

## Persist state to juju-db for new controllers

```console
$ juju bootstrap lxd test --no-gui

$ juju model-config logging-config='<root>=DEBUG'
$ juju deploy ./testcharms/charm-repo/bionic/metered-multi-series --series bionic

# Wait for the unit to start
$ juju debug-log --replay | grep "meter-status"
[...]

PR #11413, opened 2020-04-06
[dev] Bump meter status facade version

achilleasa:
## Description of change

This PR bumps the MeterStatus facade version after the changes that landed with #11405.

This would not be required under normal circumstances as the agents that are supposed to be calling the V2 methods (provided via a mixin) are guaranteed to be upgraded after the controller. Unfortunately, pylibjuju exposes the agent facades which technically makes it possible for someone to actually try to invoke the new methods on an old controller.

## QA steps

Same as #11405

### DIFF
--- a/api/facadeversions.go
+++ b/api/facadeversions.go
@@ -65,7 +65,7 @@ var facadeVersions = map[string]int{
 	"MachineManager":               6,
 	"MachineUndertaker":            1,
 	"Machiner":                     2,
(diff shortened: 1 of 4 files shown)


achilleasa:
!!build!! unrelated transient error.

achilleasa:
$$merge$$

achilleasa:
$$merge$$ (DNS resolution issues)

PR #11414, opened 2020-04-06
move relations state server side

hmlanigan:
## Description of change

Move the uniter's internal relation state from files on the unit to the controller.

The uniter's internal relation state keeps track of relations made for this unit.  The entire relation.State structure will now be saved, rather than pieces and complicated ways of interpreting files to come with a relation.State.  

Upgrade steps included.  Migration steps for all of the uniter internal state moving server side will be handled together.

## QA steps
```console
$ juju bootstrap
$ juju deploy keystone 
$ juju deploy percona-cluster mysql 
$ juju add-relation keystone:shared-db mysql:shared-db 

# once the units stable, check the juju db
juju-db.bash
juju:PRIMARY> db.unitstates.find({},{"relation-state":1}).pretty()

# add a few units and relate them:
$ juju deploy glance -n 2 ; juju add-relation glance keystone  ; juju add-relation glance mysql

# once the units stable, check the juju db as above

# remove a glance unit
$ juju remove-unit glance/1

# once the units stable, check the juju db as above
```

Run same config upgrading from 2.7.5 to develop

### DIFF
--- a/state/unit_ops.go
+++ b/state/unit_ops.go
@@ -193,9 +193,8 @@ func (op *unitSetStateOperation) fields(currentDoc unitStateDoc) (bson.D, bson.D
 		if len(rState) == 0 {
 			unsetFields = append(unsetFields, bson.DocElem{Name: "relation-state"})
 		} else if matches := currentDoc.relationStateMatches(rState); !matches {
-			newRState := mergeRelations(currentDoc.RelationState, rState)
-			setFields = append(setFields, bson.DocElem{"relation-state", newRState})
-			quotaChecker.Check(newRState)
+			setFields = append(setFields, bson.DocElem{"relation-state", rState})
+			quotaChecker.Check(rState)
 		}
 	} else {
 		quotaChecker.Check(currentDoc.RelationState)
@@ -242,26 +241,6 @@ func (op *unitSetStateOperation) getUniterStateQuotaChecker() quota.Checker {
 	return quota.NewMultiChecker(
 		quota.NewBSONTotalSizeChecker(op.limits.MaxAgentStateSize),
 	)
-}
-
-func mergeRelations(currentRelations map[string]string, newRelations map[string]string) map[string]string {
-	// For keys in newRelations:
-	// if key not in currentDoc, add key/value
-	// if key in currentDoc, if value == "", remove
-	// if key in currentDoc, if value != "", replace
-	for id, newValue := range newRelations {
-		_, found := currentRelations[id]
-		switch {
-		case found && newValue == "":
-			delete(currentRelations, id)
-		default:
-			if currentRelations == nil {
-				currentRelations = make(map[string]string)
-			}
-			currentRelations[id] = newValue
-		}
-	}
-	return currentRelations
 }
 
 // Done implements ModelOperation.
--- a/state/unitstate.go
+++ b/state/unitstate.go
@@ -78,12 +78,14 @@ func (d *unitStateDoc) relationData() (map[int]string, error) {
 }
 
 // relationStateMatches returns true if the RelationState map within the
-// unitStateDoc matches the provided newRS argument.  Assumes that
-// IgnoreRelationsState has been called first, therefore if arg is empty,
-// returns a nil map to be set.
+// unitStateDoc contains all of the provided newRS map.  Assumes that
+// relationStateBSONFriendly has been called first.
 func (d *unitStateDoc) relationStateMatches(newRS map[string]string) bool {
-	for k, v := range d.RelationState {
-		if newRS[k] != v {
+	if len(d.RelationState) != len(newRS) {
+		return false
+	}
+	for k, v := range newRS {
+		if d.RelationState[k] != v {
 			return false
 		}
 	}
--- a/upgrades/steps_28.go
+++ b/upgrades/steps_28.go
@@ -4,9 +4,11 @@
 package upgrades
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/juju/errors"
@@ -128,16 +130,6 @@ func moveUnitAgentStateToController(ctx Context) error {
 
 	// Saving uniter state in the controller succeeded, now clean up
 	// the now unused state files.
-	//
-	// TODO: hml 30-mar-2020
-	// Once the final piece of the upgrade is done, this can be changed to
-	// remove:
-	//     /var/lib/juju/agents/unit-ubuntu-0/state/uniter
-	//     /var/lib/juju/agents/unit-ubuntu-0/state/storage
-	//     /var/lib/juju/agents/unit-ubuntu-0/state/relations
-	//     /var/lib/juju/agents/unit-ubuntu-0/state/metrics
-	// and the caas operator file.  Any not found is fine.
-	// Keep /var/lib/juju/agents/unit-ubuntu-0/state for non persistent needs.
 	for _, name := range fileNames {
 		if err = os.RemoveAll(name); err != nil {
 			// Log the error, but no need to fail the upgrade.  Juju
@@ -188,16 +180,22 @@ func iaasUniterState(ctx Context, unitNames []string) ([]string, error) {
 			uniterStates[i].UniterState = &uniterSt
 		}
 
-		// TODO(achilleasa): migrate relation data.
-
 		storageData, err := readStorageState(uniterStateDir)
-		if err != nil && !os.IsNotExist(err) && !errors.IsNotFound(err) {
+		if err != nil && !errors.IsNotFound(err) {
 			return nil, err
 		} else if err == nil {
 			uniterStates[i].StorageState = &storageData.dataString
 		}
 		if storageData.filename != "" {
 			fileNames.Add(storageData.filename)
+		}
+
+		relationData, err := readRelationState(uniterStateDir)
+		if err != nil && !os.IsNotExist(err) {
+			return nil, err
+		} else if err == nil && len(relationData.data) > 0 {
+			uniterStates[i].RelationState = &relationData.data
+			fileNames.Add(relationData.filename)
 		}
 
 		// NOTE(achilleasa): meter status is transparently migrated to
@@ -249,30 +247,33 @@ func caasOperatorLocalState(ctx Context, unitNames []string) ([]string, error) {
 	return nil, nil
 }
 
-type data struct {
+type storageData struct {
 	dataString string
 	filename   string
 }
 
-func readStorageState(uniterStateDir string) (data, error) {
+func readStorageState(uniterStateDir string) (storageData, error) {
 	storageStateDir := filepath.Join(uniterStateDir, "storage")
 	storage, err := readAllStorageStateFiles(storageStateDir)
 	if err != nil {
-		return data{}, err
+		return storageData{}, err
 	}
 	if len(storage) < 1 {
-		return data{filename: storageStateDir}, errors.NotFoundf("storage state")
+		return storageData{filename: storageStateDir}, errors.NotFoundf("storage state")
 	}
 	dataYaml, err := yaml.Marshal(storage)
 	if err != nil {
-		return data{}, errors.Annotatef(err, "failed to unmarshal storage state from %q", storageStateDir)
+		return storageData{}, errors.Annotatef(err, "failed to unmarshal storage state from %q", storageStateDir)
 	}
-	return data{dataString: string(dataYaml), filename: storageStateDir}, nil
+	return storageData{dataString: string(dataYaml), filename: storageStateDir}, nil
 }
 
 // readAllStorageStateFiles loads and returns every storage stateFile persisted inside
 func readAllStorageStateFiles(dirPath string) (map[string]bool, error) {
 	if _, err := os.Stat(dirPath); err != nil {
+		if os.IsNotExist(err) {
+			return nil, errors.NotFoundf("storage state file")
+		}
 		return nil, err
 	}
 	fis, err := ioutil.ReadDir(dirPath)
@@ -313,6 +314,9 @@ func readStorageStateFile(dirPath string, tag names.StorageTag) (bool, error) {
 
 	var info storageDiskInfo
 	if err := utils.ReadYaml(filePath, &info); err != nil {
+		if os.IsNotExist(err) {
+			return false, errors.NotFoundf("storage state file %q", filePath)
+		}
 		return false, errors.Annotatef(err, "cannot read from storage state file %q", filePath)
 	}
 	if info.Attached == nil {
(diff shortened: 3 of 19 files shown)


hmlanigan:
I need to fix the worker/uniter tests.  But the rest should be good to start review.

hmlanigan:
Hoping the worker/uniter tests are now fixed.

worker/uniter/uniter.go
```diff
@@ -683,6 +679,10 @@ func (u *Uniter) init(unitTag names.UnitTag) (err error) {
 	}
 	u.operationExecutor = operationExecutor
 
+	// Ensure we have an agent directory to to write the socket.
+	if err := os.MkdirAll(u.paths.State.BaseDir, 0755); err != nil {
```
hmlanigan, line 683: We may be able to move this to the uniter tests, as in the real world it appears that it's not needed.  More research necessary.  Will investigate as a refinement.

upgrades/steps_28.go
```diff
@@ -321,6 +331,137 @@ func readStorageStateFile(dirPath string, tag names.StorageTag) (bool, error) {
...
+			return relationData{}, errors.Annotatef(err, "failed to unmarshal relation state from %q", relationStateDir)
+		}
+		retData.data[k] = string(dataYaml)
+	}
+	return retData, nil
+}
+
+// readAllRelationStateDirs loads and returns every relationState persisted directly inside
+// the supplied dirPath. If dirPath does not exist, no error is returned.
+func readAllRelationStateDirs(dirPath string) (dirs map[int]*relationState, err error) {
```
hmlanigan, line 364: this was taken from worker/uniter/relation/state.go

upgrades/steps_28.go
```diff
@@ -321,6 +331,137 @@ func readStorageStateFile(dirPath string, tag names.StorageTag) (bool, error) {
...
+		}
+		dirs[relationId] = rState
+	}
+	return dirs, nil
+}
+
+// readRelationStateDir loads a relationState from the subdirectory of dirPath named
+// for the supplied RelationId. If the directory does not exist, no error
+// is returned,
+func readRelationStateDir(dirPath string, relationId int) (d *relationState, err error) {
```
hmlanigan, line 396: this was taken from worker/uniter/relation/state.go

state/unitstate.go
```diff
@@ -78,12 +78,14 @@ func (d *unitStateDoc) relationData() (map[int]string, error) {
...
 // relationStateMatches returns true if the RelationState map within the
-// unitStateDoc matches the provided newRS argument.  Assumes that
-// IgnoreRelationsState has been called first, therefore if arg is empty,
-// returns a nil map to be set.
+// unitStateDoc contains all of the provided newRS map.  Assumes that
+// relationStateBSONFriendly has been called first.
 func (d *unitStateDoc) relationStateMatches(newRS map[string]string) bool {
-	for k, v := range d.RelationState {
-		if newRS[k] != v {
+	if len(d.RelationState) < 1 {
```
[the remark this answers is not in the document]
achilleasa: Without the proposed change, the for loop below would not detect changes if:
- newRS is a subset of oldRS
- the intersection of the two sets is not the empty set (iow, oldRS contains keys not present in newRS)

upgrades/steps_28.go
```diff
@@ -200,6 +202,14 @@ func iaasUniterState(ctx Context, unitNames []string) ([]string, error) {
 			fileNames.Add(storageData.filename)
 		}
 
+		relationData, err := readRelationState(uniterStateDir)
```
achilleasa, line 205: Please remove TODO placeholder on L193

upgrades/steps_28.go
```diff
@@ -200,6 +202,14 @@ func iaasUniterState(ctx Context, unitNames []string) ([]string, error) {
...
+		if err != nil && !os.IsNotExist(err) && !errors.IsNotFound(err) {
```
achilleasa, line 206: Since we control `readRelationState` why not have it test for `os.IsNotExist` internally and return back errors.IsNotFound? An example of this pattern can be found [here](https://github.com/juju/juju/blob/develop/worker/meterstatus/state.go#L112).

If you do this, can you also please change `readStorageState` above to use the same pattern?

worker/uniter/relation/relationer.go
```diff
@@ -53,15 +53,15 @@ func (r *Relationer) RelationUnit() *apiuniter.RelationUnit {
 
 // Join initializes local state and causes the unit to enter its relation
 // scope, allowing its counterpart units to detect its presence and settings
-// changes. Local state directory is not created until needed.
+// changes.
 func (r *Relationer) Join() error {
 	if r.dying {
 		panic("dying relationer must not join!")
```
achilleasa, line 59: As a drive-by, can you replace this panic with an error?

worker/uniter/relation/relationer.go
```diff
@@ -53,15 +53,15 @@ func (r *Relationer) RelationUnit() *apiuniter.RelationUnit {
...
 	}
-	// We need to make sure the state directory exists before we join the
-	// relation, lest a subsequent ReadAllStateDirs report local state that
+	// We need to make sure the state is persisted inState before we join
+	// the relation, lest a subsequent ReadAllStates report local state that
 	// doesn't include relations recorded in remote state.
-	if err := r.dir.Ensure(); err != nil {
+	if err := r.ops.EnsureState(); err != nil {
```
achilleasa, line 64: Took my a while to figure out what we mean by ensure; since we are refactoring could we rename it to something like `RefreshState` or `ReloadState` that will make sense to us when we revisit the code a few months down the road?

worker/uniter/relation/relationer_test.go
```diff
@@ -304,23 +280,29 @@ func (s *RelationerImplicitSuite) TestImplicitRelationer(c *gc.C) {
...
 	c.Assert(err, jc.ErrorIsNil)
 
-	r := relation.NewRelationer(apiRelUnit, dir)
+	r := relation.NewRelationer(apiRelUnit, ops)
 	c.Assert(r, jc.Satisfies, (*relation.Relationer).IsImplicit)
 
 	// Hooks are not allowed.
-	f := func() { r.PrepareHook(hook.Info{}) }
+	f := func() { _, _ = r.PrepareHook(hook.Info{}) }
 	c.Assert(f, gc.PanicMatches, "implicit relations must not run hooks")
```
achilleasa, line 288: It would be a good opportunity to get rid of panics. If we bubble up the error, only the uniter worker will shutdown (and restart).

worker/uniter/relation/resolver.go
```diff
@@ -153,21 +153,21 @@ func (r *relationsResolver) nextHookForRelation(localStateDir *StateDir, remote
...
+	for unitName := range localState.Members {
 		allUnitNames.Add(unitName)
 	}
 	for unitName := range remote.Members {
 		allUnitNames.Add(unitName)
 	}
 	sortedUnitNames := allUnitNames.SortedValues()
 	if allUnitNames.Contains("") {
-		return hook.Info{}, errors.Errorf("somehow we got the empty unit. local: %v, remote: %v", local.Members, remote.Members)
+		return hook.Info{}, errors.Errorf("somehow we got the empty unit. localState: %v, remote: %v", localState.Members, remote.Members)
```
achilleasa, line 164: small nit: s/remote: %v/remoteState: %v/?

worker/uniter/relation/resolver_test.go
```diff
@@ -142,23 +140,26 @@ func (s *relationResolverSuite) setupRelations(c *gc.C) relation.RelationStateTr
...
-			CharmDir:             s.stateDir,
-			RelationsDir:         s.relationsDir,
+			Unit:                 u,
+			CharmDir:             s.charmDir,
 			NewLeadershipContext: s.leadershipContextFunc,
 			Abort:                abort,
 		})
 	c.Assert(err, jc.ErrorIsNil)
-	assertNumCalls(c, &numCalls, 3)
+	assertNumCalls(c, &numCalls, 4)
```
achilleasa, line 162: I had to resort to the same whack-a-mole game when I did the previous refactoring work here. We should really really try to rewrite these tests at some point to use mocks in the correct way...

worker/uniter/relation/resolver_test.go
```diff
@@ -634,19 +720,13 @@ func (s *relationResolverSuite) TestHookRelationBrokenWhenSuspended(c *gc.C) {
 	relationsResolver := relation.NewRelationResolver(r, nil)
 	op, err := relationsResolver.NextOp(localState, remoteState, &mockOperations{})
 	c.Assert(err, jc.ErrorIsNil)
-	assertNumCalls(c, &numCalls, 11)
+	assertNumCalls(c, &numCalls, 18)
```
achilleasa, line 723: That's a rather significant bump in the number of API calls. Should we be worried?

worker/uniter/relation/resolver_test.go
```diff
@@ -1018,13 +1158,24 @@ func principalWithSubordinateAPICalls() []apiCall {
...
+		}}
+	unitStateResults := params.UnitStateResults{Results: []params.UnitStateResult{{}}}
+
 	return []apiCall{
 		uniterAPICall("Refresh", nrpeUnitEntity, params.UnitRefreshResults{Results: []params.UnitRefreshResult{{Life: life.Alive, Resolved: params.ResolvedNone}}}, nil),
 		uniterAPICall("GetPrincipal", nrpeUnitEntity, params.StringBoolResults{Results: []params.StringBoolResult{{Result: "unit-wordpress-0", Ok: true}}}, nil),
 		uniterAPICall("RelationsStatus", nrpeUnitEntity, relationStatusResults, nil),
 		uniterAPICall("Relation", relationUnits1, relationResults1, nil),
+		uniterAPICall("State", nrpeUnitEntity, unitStateResults, nil),
+		//uniterAPICall("State", nrpeUnitEntity, unitStateResults, nil),
```
achilleasa, line 1175: Can the commented line be removed?

worker/uniter/relation/state.go
```diff
@@ -108,200 +107,196 @@ func (s *State) Validate(hi hook.Info) (err error) {
...
+type UnitStateReadWriter interface {
+	State() (params.UnitStateResult, error)
+	SetState(unitState params.SetUnitStateArg) error
 }
 
-// State returns the current state of the relation.
-func (d *StateDir) State() *State {
-	return d.state.copy()
+// StateOps is a representation of the relation state. Concurrent
+// modifications to the underlying state will have undefined consequences.
```
achilleasa, line 118: Nice to include concurrency limitations in docs for the type!

worker/uniter/relation/state.go
```diff
@@ -108,200 +107,196 @@ func (s *State) Validate(hi hook.Info) (err error) {
...
-func (d *StateDir) Write(hi hook.Info) (err error) {
-	defer errors.DeferredAnnotatef(&err, "failed to write %q hook info for %q on state directory", hi.Kind, hi.RemoteUnit)
+func (s *StateOps) WriteHook(hi hook.Info) (err error) {
+	defer errors.DeferredAnnotatef(&err, "failed to write %q hook info for %q in state", hi.Kind, hi.RemoteUnit)
 	if hi.Kind == hooks.RelationBroken {
-		return d.Remove()
+		return s.Remove()
 	}
-	name := strings.Replace(hi.RemoteUnit, "/", "-", 1)
 	isApp := false
```
achilleasa, line 193: Due to the removal, this block can be condensed to `isApp := hi.RemoteUnit == ""`

worker/uniter/relation/state.go
```diff
@@ -108,200 +107,196 @@ func (s *State) Validate(hi hook.Info) (err error) {
...
 	//  delete "foo-app". Instead, during relation-broken, we cleanup all related applications.
-	for appMember := range d.state.ApplicationMembers {
-		path := filepath.Join(d.path, appMember+"-app")
-		if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
-			return errors.Trace(err)
-		}
+	//
+	// Note(hml): 2020-04-01 We no longer keep persisted state as local files,
+	// instead remove all ApplicationMembers.
+	s.state.ApplicationMembers = nil
```
achilleasa, line 249: I believe this should be at the bottom of the method after L269; we don't want to reset this map before the `SetState` call returns successfully

worker/uniter/relation/state.go
```diff
@@ -108,200 +107,196 @@ func (s *State) Validate(hi hook.Info) (err error) {
...
-	ChangeVersion  *int64 `yaml:"change-version"`
-	ChangedPending bool   `yaml:"changed-pending,omitempty"`
+// State returns the current state of the relation.
+func (s *StateOps) State() *State {
+	return s.state.copy()
+}
+
+// EnsureState writes state via SetState as ReadState returns a
+// new state even if one was not read from State.
+func (s *StateOps) EnsureState() error {
```
achilleasa, line 280: As mentioned above; EnsureState is a bit misleading method name.

worker/uniter/relation/state.go
```diff
@@ -108,200 +107,196 @@ func (s *State) Validate(hi hook.Info) (err error) {
...
+	State() (params.UnitStateResult, error)
+	SetState(unitState params.SetUnitStateArg) error
 }
 
-// State returns the current state of the relation.
-func (d *StateDir) State() *State {
-	return d.state.copy()
+// StateOps is a representation of the relation state. Concurrent
+// modifications to the underlying state will have undefined consequences.
+type StateOps struct {
```
achilleasa, line 119: Is there a particular reason why we don't use a single object to load the complete state data at construction time (just like ReadAllStates) and then provide an API to read/write the data for a particular relation ID? This could be something created by the uniter and passed through to the relation state tracker and from there to the relationers.

This way, your data will always be up to date (and if the uniter restarts it will pull a fresh copy) and you don't need to have calls like `Ensure` in there.

hmlanigan: StateOps have now been replaced with StateManager which handles all of the relation state.

state/unit_ops.go
```diff
@@ -130,6 +130,10 @@ func (op *unitSetStateOperation) newUnitStateDoc(unitGlobalKey string) (unitStat
 		newStDoc.StorageState = storState
 		quotaChecker.Check(storState)
 	}
+	if relState, found := op.newState.relationStateBSONFriendly(); found {
```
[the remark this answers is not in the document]
hmlanigan: didn't see the existing code, read too fast given the order was different.  removing

state/unitstate.go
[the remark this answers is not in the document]
hmlanigan: done

upgrades/steps_28.go
hmlanigan: done, removed another todo as well.

upgrades/steps_28.go
hmlanigan: for the relation piece  IsNotFound will not occur.  it's specifically returned in the readStorageState though.  I can replace the NotExists errors with NotFound for storage.

worker/uniter/relation/state.go
hmlanigan: done

worker/uniter/relation/relationer.go
hmlanigan: I'd rather not do this as a drive-by.  Will double check that the uniter will be restarted on an error here and address in a new PR

upgrades/steps_28.go
```diff
@@ -200,6 +190,14 @@ func iaasUniterState(ctx Context, unitNames []string) ([]string, error) {
 			fileNames.Add(storageData.filename)
 		}
 
+		relationData, err := readRelationState(uniterStateDir)
+		if err != nil && !os.IsNotExist(err) {
```
achilleasa, line 194: Just to double-check: is os.IsNotExist the right error type to check here? We don't seem to wrap it in a not found error, right?

upgrades/steps_28.go
```diff
@@ -321,6 +325,137 @@ func readStorageStateFile(dirPath string, tag names.StorageTag) (bool, error) {
...
+	}
+	return retData, nil
+}
+
+// readAllRelationStateDirs loads and returns every relationState persisted directly inside
+// the supplied dirPath. If dirPath does not exist, no error is returned.
+func readAllRelationStateDirs(dirPath string) (dirs map[int]*relationState, err error) {
+	defer errors.DeferredAnnotatef(&err, "cannot load relations state from %q", dirPath)
+	if _, err := os.Stat(dirPath); os.IsNotExist(err) {
+		return nil, nil
```
achilleasa, line 361: If the dir does not exist, we return nil, and `readRelationState` returns back an empty map. Is this the intended behavior?

upgrades/steps_28.go
achilleasa: It looks like `readRelationState` and the functions it calls never return errors that pass the `os.IsNotExist` predicate.

upgrades/steps_28.go
```diff
@@ -321,6 +325,137 @@ func readStorageStateFile(dirPath string, tag names.StorageTag) (bool, error) {
...
+func readRelationStateDir(dirPath string, relationId int) (d *relationState, err error) {
+	d = &relationState{
+		RelationId:         relationId,
+		Members:            map[string]int64{},
+		ApplicationMembers: map[string]int64{},
+		ChangedPending:     "",
+	}
+	defer errors.DeferredAnnotatef(&err, "cannot load relation state from %q", dirPath)
+	if _, err := os.Stat(dirPath); os.IsNotExist(err) {
+		return d, nil
```
achilleasa, line 399: Same question about suppressing the error and returning back a blank relationState. So, if we find an _empty_ folder for the relation we silently convert it into a blank state? Wouldn't the absense of state data indicate that something is wrong?

worker/uniter/relation/relationer.go
achilleasa: Fair enough; we can definitely do this in a separate PR as you suggest.

worker/uniter/relation/statemanager.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// UnitStateReadWriter encapsulates the methods from a state.Unit
+// required to set and get unit state.
+type UnitStateReadWriter interface {
+	State() (params.UnitStateResult, error)
+	SetState(unitState params.SetUnitStateArg) error
+}
+
+// NewStateManager
```
achilleasa, line 46: nit: Missing comment ;-)

worker/uniter/relation/statemanager.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+type stateManager struct {
+	unitStateRW   UnitStateReadWriter
+	relationState map[int]State
+	mu            sync.Mutex
+}
+
+// RelationState returns a copy of the relation state for
+// the given id. Returns NotFound.
```
achilleasa, line 59: Slight correction;
s/. Returns NotFound/ or returns NotFound if the given id does not belong to a known relation/

worker/uniter/relation/relationer_test.go
```diff
@@ -4,43 +4,42 @@
 package relation_test
 
 import (
-	"strconv"
+	"github.com/juju/juju/apiserver/params"
```
hmlanigan, line 7: I need to fix the stanzas

upgrades/steps_28.go
hmlanigan: It can bubble up from os.Stat in readAllRelationStateDirs.  Although, I wonder if we shouldn't fail an upgrade on IsNotExist because there should always be a peer relation to move. readAllRelationStateDirs was moved from the uniter and dealt with uniter start up scenarios too.

achilleasa:
QA works as expected!

upgrades/steps_28_test.go
```diff
@@ -117,15 +117,16 @@ func (s *mockSteps28Suite) SetUpTest(c *gc.C) {
...
-
 	s.opStorOneYaml, s.opStorOneFileName = writeStorageState(c, unitOneStateDir, s.storTagOne, s.opStorOne)
-
 	s.opRelationYaml, s.opRelationFileName = setupRelationState(c, unitOneStateDir)
 
 	unitTwoStateDir := filepath.Join(agentDir, s.tagTwo.String(), "state")
 	err = os.MkdirAll(unitTwoStateDir, 0755)
 	c.Assert(err, jc.ErrorIsNil)
-	s.opStateTwoYaml, s.opStateTwoFileName = writeUnitStateFile(c, unitTwoStateDir, s.opStateTwo)
+	//s.opStateTwoYaml, s.opStateTwoFileName = writeUnitStateFile(c, unitTwoStateDir, s.opStateTwo)
```
achilleasa, line 126: Don't forget to remove this before merging.

hmlanigan:
rebased

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #11416, opened 2020-04-07
Protect actionData in hook context with a mutex

wallyworld:
## Description of change

The race detector picked up possible concurrent writes to the hook context action data map.
So add a mutex to stop that.

## QA steps

go test --race ./... in the worker uniter package


wallyworld:
$$merge$$

PR #11426, opened 2020-04-07
upgrades: Add a 2.7.6 step to fix hooks in uniter state files

babbageclunk:
### Checklist

----

## Description of change

If an upgrade happens while a hook is queued, the uniter state file will contain details about that hook execution so it can be run once the unit agent is running again. In 2.7 the hook was extended to include `remote-application`, and validation requires it. But if the hook was written by the 2.6 uniter and then the agent is upgraded to 2.7, the uniter won't start because the state file is invalid.

Run an upgrade step when going to 2.7.6 to update any hooks that don't have `remote-application` set to be the application from the remote-unit field. (We're running it in the 2.7.6 upgrade because there could be agents stuck in this state since upgrading from 2.6.x.)

## QA steps

Bootstrap a 2.6 controller, deploy dummy-source and set its token to a value.
```sh
/snap/bin/juju bootstrap lxd n
juju deploy ~/juju/acceptancetests/repository/charms/dummy-source
juju config dummy-source token=hey
```
Modify the dummy-sink so that its `source-relation-changed` hook fails if `token` is set to "block".
```bash
#!/bin/bash
set -eux

mkdir -p /var/run/dummy-sink
token=$(relation-get token)
echo "$token" > /var/run/dummy-sink/token

if [[ "$token" == "block" ]]; then
  juju-log -l WARNING "Blocking: hook won't succeed while token is block"
  status-set blocked "Token is block, change it" || true
  exit 1
fi

if [[ -z "$token" ]]; then
  juju-log -l INFO "Waiting for token"
  status-set waiting "Waiting for token" || true
else
  juju-log -l INFO "Token is $token"
  status-set active "Token is $(echo $token | cut -c 1-20)" || true
fi
```
Deploy that charm and relate it to the source.
```sh
juju deploy ./blocker-sink && juju relate dummy-source blocker-sink
```

Update token to "block" - blocker-sink should get stuck.
```sh
juju config dummy-source token=block
```
Check that the hook is in `/var/lib/juju/agents/unit-blocker-sink-0/state/uniter` without `remote-application`.
Once the charm is blocked, upgrade to 2.7.6.

```sh
juju upgrade-model
```

Watch the blocker-sink machine agent log to see the upgrade step report updating the file.
See that the hook in the state file is updated with the application.
Check that the uniter is started and hook executions are running.
```sh
juju config dummy-source token=fine
juju resolve blocker-sink/0
```

## Documentation changes

None

## Bug reference

https://bugs.launchpad.net/juju/+bug/1869275

### DIFF
--- a/upgrades/steps_276.go
+++ b/upgrades/steps_276.go
@@ -59,30 +59,48 @@ func AddRemoteApplicationToRunningHooks(pattern string) func(Context) error {
 // remote application so that it does. (If it doesn't validate for
 // some other reason we won't change the file.)
 func AddRemoteApplicationToHook(path string) error {
-	var uniterState operation.State
+	var uniterState map[string]interface{}
 	err := utils.ReadYaml(path, &uniterState)
 	if err != nil {
 		return errors.Trace(err)
 	}
 
-	if uniterState.Hook == nil {
-		logger.Debugf("no hook found in %q, unable to fix", path)
+	hookUnconverted, found := uniterState["hook"]
+	if !found {
+		logger.Warningf("no hook found in %q, unable to fix", path)
 		return nil
 	}
 
-	if uniterState.Hook.RemoteApplication != "" {
-		logger.Debugf("remote-application in %q set to %q already", path,
-			uniterState.Hook.RemoteApplication)
+	hook, ok := hookUnconverted.(map[interface{}]interface{})
+	if !ok {
+		logger.Warningf("fixing %q: expected hook to be a map[interface{}]interface{}, got %T", path, hookUnconverted)
 		return nil
 	}
 
-	appName, err := names.UnitApplication(uniterState.Hook.RemoteUnit)
+	if val, found := hook["remote-application"]; found {
+		logger.Debugf("remote-application in %q set to %v already", path, val)
+		return nil
+	}
+
+	unitUnconverted, found := hook["remote-unit"]
+	if !found {
+		logger.Warningf("fixing %q: remote-unit not found")
+		return nil
+	}
+
+	unit, ok := unitUnconverted.(string)
+	if !ok {
+		logger.Warningf("fixing %q: expected remote-unit to be string, got %T", path, unitUnconverted)
+		return nil
+	}
+
+	appName, err := names.UnitApplication(unit)
 	if err != nil {
 		return errors.Trace(err)
 	}
 
 	logger.Debugf("setting remote-application to %q in %q", appName, path)
-	uniterState.Hook.RemoteApplication = appName
-	return errors.Annotatef(operation.NewStateFile(path).Write(&uniterState),
+	hook["remote-application"] = appName
+	return errors.Annotatef(utils.WriteYaml(path, uniterState),
 		"writing updated state to %q", path)
 }
--- a/upgrades/steps_276.go
+++ b/upgrades/steps_276.go
@@ -1,0 +1,88 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package upgrades
+
+import (
+	"path/filepath"
+
+	"github.com/juju/errors"
+	"github.com/juju/utils"
+	"gopkg.in/juju/names.v3"
+
+	"github.com/juju/juju/worker/uniter/operation"
+)
+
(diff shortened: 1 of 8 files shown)


upgrades/steps_276.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+const uniterStateGlob = `/var/lib/juju/agents/unit-*/state/uniter`
+
+// AddRemoteApplicationToRunningHooks finds any uniter state files on
+// the machine with running hooks, and makes sure that they contain a
+// remote-application key.
+func AddRemoteApplicationToRunningHooks(pattern string) func(Context) error {
+	return func(_ Context) error {
+		matches, err := filepath.Glob(pattern)
```
howbazaar, line 34: This is a nicer way to get the files than I was thinking of. Good work.

upgrades/steps_276.go
```diff
@@ -0,0 +1,88 @@
...
+// remote application so that it does. (If it doesn't validate for
+// some other reason we won't change the file.)
+func AddRemoteApplicationToHook(path string) error {
+	var uniterState operation.State
+	err := utils.ReadYaml(path, &uniterState)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	if uniterState.Hook == nil {
```
howbazaar, line 68: The only time we have gotten to here is if the file is not valid. I think we should not just shrug and return nil here.

Perhaps we should write a warning and delete the file? We want to ensure that the uniter is unblocked. Right now if we return from this function without actually fixing something, the uniter will remain blocked. I think it is better to have the uniter effectively "restart" than to just exit and leave the file broken.

Thoughts?

babbageclunk: After discussion this seems like a bad idea - there could be a future change in some other part of the file to make it invalid in another way (with another upgrade step to fix), but if we trash the file here that other step can't do its job.

upgrades/steps_276.go
```diff
@@ -0,0 +1,106 @@
...
+func AddRemoteApplicationToHook(path string) error {
+	var uniterState map[string]interface{}
+	err := utils.ReadYaml(path, &uniterState)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	hookUnconverted, found := uniterState["hook"]
+	if !found {
+		logger.Debugf("no hook found in %q, unable to fix", path)
```
wallyworld, line 70: Warning?

upgrades/steps_276.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil
+	}
+
+	hook, ok := hookUnconverted.(map[interface{}]interface{})
+	if !ok {
+		logger.Debugf("fixing %q: expected hook to be a map[string]interface{}, got %T", path, hookUnconverted)
```
wallyworld, line 76: ditto - i think bad input should be a warning to surface it

upgrades/steps_276_test.go
```diff
@@ -0,0 +1,83 @@
...
+	dir string
+}
+
+var _ = gc.Suite(&steps276Suite{})
+
+func (s *steps276Suite) SetUpTest(c *gc.C) {
+	s.dir = c.MkDir()
+	files, err := filepath.Glob("testdata/uniter-state-*")
+	c.Assert(err, jc.ErrorIsNil)
+	copyCommand := exec.Command("/bin/cp", append(files, s.dir)...)
```
wallyworld, line 35: can we include a file that doesn't match the pattern to make sure it is ignored

babbageclunk:
$$merge$$

upgrades/steps_276.go
```diff
@@ -0,0 +1,106 @@
...
+
+	hookUnconverted, found := uniterState["hook"]
+	if !found {
+		logger.Warningf("no hook found in %q, unable to fix", path)
+		return nil
+	}
+
+	hook, ok := hookUnconverted.(map[interface{}]interface{})
+	if !ok {
+		logger.Warningf("fixing %q: expected hook to be a map[string]interface{}, got %T", path, hookUnconverted)
```
howbazaar, line 76: warning doesn't match cast

babbageclunk: oh duh, thanks

babbageclunk:
$$merge$$

PR #11428, opened 2020-04-08
Fixed go lint check to exclude all mock patterns

tlm:
## Description of change

Our golint check was not excluding all mock paths.

## QA steps

Run golint check.


tlm:
$$merge$$

PR #11431, opened 2020-04-09
Ensure when pod status is reported to juju, the statefu attribute is set

wallyworld:
## Description of change

When reporting pod status, be sure to set the stateful boolean attribute if the pod belongs to a statefulset, or else juju will get confused when pods are restarted as part of a charm upgade.

## QA steps

deploy mariadb-k8s -n 3
edit the pod spec of the charm to change an env var
upgrade-charm
ensure juju status continues to show only 3 units 0, 1, 2 as k8s one by one terminates and restarts each pod

## Bug reference

https://bugs.launchpad.net/juju/+bug/1871756

### DIFF
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -2046,7 +2046,7 @@ func (k *kubernetesClient) Units(appName string) ([]caas.Unit, error) {
 	}
 
 	var units []caas.Unit
-	now := time.Now()
+	now := k.clock.Now()
 	for _, p := range podsList.Items {
 		var ports []string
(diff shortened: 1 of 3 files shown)


achilleasa:
QA works as expected at least for the leader pod. The rest (as mentioned in the bug report) get get into an error state as they try to set the pod spec while not being leaders

mitechie:
Thanks for the reviews everyone. 

$$merge$$

wallyworld:
[upgrade charm duplicate units](https://trello.com/c/kH39jiKE/1094-upgrade-charm-duplicate-units)

PR #11439, opened 2020-04-14
vsphere: Remove code to delete cached VMDKs from all datastores

babbageclunk:
### Checklist

----

## Description of change

This is vestigial code that should have been removed a while ago - we no longer cache VMDKs in the datastore, instead they're used to create a template VM that new VMs are cloned from.

This deletion is causing problems if the user doesn't have access to the datastore - in that case it fails and blocks controller teardown.

## QA steps

Bootstrap a vsphere controller. Tear it down with `--debug` - there shouldn't be any logging from `juju.provider.vmware` about deleting from datastores.

## Documentation changes

None

## Bug reference

https://bugs.launchpad.net/juju/+bug/1872552

### DIFF
--- a/provider/vsphere/environ.go
+++ b/provider/vsphere/environ.go
@@ -4,7 +4,6 @@
 package vsphere
 
 import (
-	"fmt"
 	"path"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


babbageclunk:
$$merge$$

PR #11442, opened 2020-04-14
Fix windows manual provisioning

gabriel-samfira:
This change fixes the WinRM provisioning scripts, which re-enables the manual provider to work with Windows machines.

See: https://paste.ubuntu.com/p/ptZG9zgt4T/

This change depends on:

https://github.com/juju/utils/pull/310

## QA steps

  * Bootstrapped a new controller with manual provider.
  * Added a Windows Server 2019 machine.
  * Deployed the [active directory charm](https://jaas.ai/u/cloudbaseit/active-directory/9)

```sh
juju bootstrap manual --config agent-stream=devel --config agent-metadata-url=http://192.168.100.202/tools
juju add-machine winrm:Administrator@192.168.100.222
juju deploy cs:~cloudbaseit/active-directory --to 0 --config $HOME/charms/config.yaml --series win2019 --force
```

## Documentation changes

No changes to API/CLI. Functionality has been there for a while, just left behind. I will also send documentation changes that will detail the needed steps to use the manual provider with Windows machines.

## Bug reference

N/A

### DIFF
--- a/environs/manual/winrmprovisioner/winrmprovisioner.go
+++ b/environs/manual/winrmprovisioner/winrmprovisioner.go
@@ -64,7 +64,19 @@ function Get-Arch {
 }
 
 function Get-Ram {
-   return (gcim win32_physicalmemory).Capacity
+   # The comma is not a typo. It forces $capacity to be an array.
+   # On machines with multiple slots of memory, the return value
+   # of gcim win32_physicalmemory, may be an array. On a machine
+   # with a single slot of memory, it will be a win32_physicalmemory
+   # object. Forcing an array here, makes a common case out of
+   # both situations and saves us the trouble of testing the
+   # return type.
+   $capacity = ,(gcim win32_physicalmemory).Capacity
+   $ram = 0
+   foreach($i in $capacity){
+      $ram += $i
+   }
+   return $ram
 }
 
 function Get-OSName {
@@ -183,12 +195,6 @@ func enableCertAuth(args *manual.ProvisionMachineArgs) (manual.WinrmClientAPI, e
 		err = args.WinRM.Client.Run(script, args.Stdout, &stderr)
 		if err != nil {
 			return nil, errors.Trace(err)
-		}
-		if stderr.Len() > 0 {
-			return nil, errors.Errorf(
-				"encountered error executing cert auth script: %s",
-				stderr.String(),
-			)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


environs/manual/winrmprovisioner/winrmprovisioner.go
```diff
@@ -64,7 +64,13 @@ function Get-Arch {
 }
 
 function Get-Ram {
-   return (gcim win32_physicalmemory).Capacity
+   # The comma is not a typo. It forces $capacity to be an array
```
howbazaar, line 67: Perhaps it would be worthwhile adding why we want it as an array? In what situations would it return more than one value?

gabriel-samfira: Good point. I will amend the comment, but for posterity will add the details here as well.

The command:

```powershell
gcim win32_physicalmemory
```

may return multiple values if you have multiple memory slots. The sum of the ```Capacity``` property of those values is our total physical memory. If we only have one slot, the above command will return an object of the type ```win32_physicalmemory```, instead of an array. This change simply makes a common case for both single slot machines and multi slot machines. Saves us the trouble of checking the type of the return value from ```gcim```.

Powershell is fun that way.

gabriel-samfira: Comment amended.

gabriel-samfira:
@howbazaar Updated the hash in ```go.mod```. Great to see Juju switched to Go modules! \o/

go.sum
```diff
@@ -393,6 +393,8 @@ github.com/juju/utils v0.0.0-20200116185830-d40c2fe10647 h1:wQpkHVbIIpz1PCcLYku9
 github.com/juju/utils v0.0.0-20200116185830-d40c2fe10647/go.mod h1:6/KLg8Wz/y2KVGWEpkK9vMNGkOnu4k/cqs8Z1fKjTOk=
 github.com/juju/utils v0.0.0-20200423035217-b0a7da72a5fa h1:lv9vemJVV2iGalxqEOqEPw2+bBr0WyZqun7TfguD4fY=
 github.com/juju/utils v0.0.0-20200423035217-b0a7da72a5fa/go.mod h1:6/KLg8Wz/y2KVGWEpkK9vMNGkOnu4k/cqs8Z1fKjTOk=
+github.com/juju/utils v0.0.0-20200424103611-54ececcc5fc7 h1:3+lnQMg3pJ2xF57li+dsLERIgn9qaqnDZ0f4EmIAiUw=
```
howbazaar, line 396: It seems weird to me that this adds the new utils, but keeps the older utils as well. Is this because other repos have go.mod specifying the older revision I wonder?

howbazaar:
$$merge$$

PR #11446, opened 2020-04-15
[dev] Enforce relation limits at deploy time

achilleasa:
## Description of change

The changes from this PR allow juju to enforce the max-relation limit that can be optionally specified as part of the charm's metadata.

Given that these limits were invoked by all previous juju versions, the PR takes a conservative approach to ensure that we don't accidentally break any already deployed models when older controllers get upgraded to 2.8. More specifically, the limits will **only** be applied in the following cases:
- When adding a new relation via `juju relate X Y`
- When attempting to upgrade a charm.

Any already deployed models that violate a max relation limit for a particular relation endpoint, will still work but the operator will not be allowed to create **new** relations for that endpoint.

As far as upgrades are concerned: if the new charm version defines a stricter limit which would not be enforceable given the current deployment configuration, the controller will return back an error. However, the operator can still use `--force` to allow the upgrade to proceed.

Caveats (new bug reports will be opened to keep track of these):
- Limits not enforced when suspending/un-suspending relations
- Limit not enforced early on by the client when deploying bundles; as a result, deploying problematic bundles might fail **midway** when the controller attempts to enforce the limit. Note that fixing the latter would also require changes to pylibjuju.

## QA steps

```console

# This wordpress charm specifies a limit of 1 for the *db* endpoint
$ juju deploy ./testcharms/charm-repo/quantal/wordpress -n2
$ juju deploy mysql
$ juju deploy mariadb

# Attempting to add a second relation for the *db* endpoint causes an error
$ juju add-relation wordpress mysql
$ juju add-relation wordpress mariadb
ERROR cannot add relation "wordpress:db mariadb:db": establishing a new relation for wordpress:db would exceed its maximum relation limit of 1 (quota limit exceeded)

# Patch wordpress and change limit to 2; then trigger an upgrade
$ sed -i 's/limit: 1/limit: 2/' ./testcharms/charm-repo/quantal/wordpress/metadata.yaml
$ juju upgrade-charm --path ./testcharms/charm-repo/quantal/wordpress wordpress

# Now the second relation can be added successfully!
$ juju add-relation wordpress mariadb

# Reset the limit in the metadata and try to upgrade once more
$ git checkout testcharms
$ juju upgrade-charm --path ./testcharms/charm-repo/quantal/wordpress wordpress
Added charm "local:bionic/wordpress-5" to the model.
ERROR cannot upgrade application "wordpress" to charm "local:bionic/wordpress-5": new charm version imposes a maximum relation limit of 1 for wordpress:db which cannot be satisfied by the number of already established relations (2) (quota limit exceeded)

# We can still force it though!
 juju upgrade-charm --path ./testcharms/charm-repo/quantal/wordpress --force wordpress
Added charm "local:bionic/wordpress-6" to the model.
Leaving endpoints in "alpha": admin-api, cache, db, db-client, foo-bar, logging-dir, monitoring-port, url
```

## Documentation changes

We need to update the [docs](https://discourse.juju.is/t/implementing-relations/1051) to indicate that limit is now supported!

## Bug reference
https://bugs.launchpad.net/juju/+bug/1869840


state/application.go
```diff
@@ -1362,6 +1362,19 @@ func (a *Application) SetCharm(cfg SetCharmConfig) (err error) {
 				}}},
 			})
 		} else {
+			// Check if the new charm specifies a relation max limit
+			// that cannot be satisfied by the currently established
+			// relation count.
+			quotaErr := a.preUpgradeRelationLimitCheck(cfg.Charm)
```
SimonRichardson, line 1368: We should think about this long term, if this is the correct place for business logic?
At some point, we probably want to lift this up one level above state.

state/application.go
```diff
@@ -1362,6 +1362,19 @@ func (a *Application) SetCharm(cfg SetCharmConfig) (err error) {
...
+
+			// If the operator specified --force, we still allow
+			// the ugprade to continue with a warning.
+			if errors.IsQuotaLimitExceeded(quotaErr) && cfg.Force {
+				logger.Warningf("%v; allowing upgrade to proceed since the operator specified --force", quotaErr)
```
SimonRichardson, line 1373: Unsure about the wording of this, I'm unsure you need the "since" part.

state/application.go
```diff
@@ -1410,6 +1423,60 @@ func (a *Application) SetCharm(cfg SetCharmConfig) (err error) {
...
+// counts do not violate the max relation limits specified by the charm version
+// we are attempting to upgrade to.
+func (a *Application) preUpgradeRelationLimitCheck(newCharm *Charm) error {
+	var (
+		existingRels []*Relation
+		err          error
+	)
+
+	for relName, relSpec := range newCharm.Meta().CombinedRelations() {
+		if relSpec.Limit == 0 {
```
SimonRichardson, line 1436: `relSpec` has `optional`[1], do we count if this is set to true?

1. https://github.com/juju/charm/blob/v6/meta.go#L884

state/application.go
achilleasa: Totally agree. This (and the other logic in the method) should be handled by a logic layer between the facade and state.

state/application.go
achilleasa: This is  merely a hint that the charm can work even when the relation is not established. The above code only cares for already established relations. For more info see [here](https://discourse.juju.is/t/implementing-relations/1051):

```
optional is ignored by Juju, but if present should only be set to true, on requires relations, to indicate that the charm can still function effectively without having those relations added.
```

achilleasa:
$$merge$$

PR #11456, opened 2020-04-16
Attempt to remove default installed mongo

SimonRichardson:
## Description of change

The github action now automatically adds mongo from multiiverse of all
places and then installs it. There is no way to say, "nope, don't want
that!" - instead we have to get out our hammer and treat everything as a
nail.

This is not great, we can't rely on a clean environment anymore.

## QA steps

Get client tests to pass.


SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #11464, opened 2020-04-17
fix create profile race when multiple controller bootstrapping at same time in new lxd config

hmlanigan:
## Description of change

There is a race creating juju- lxd profiles when multiple controllers are bootstrapping at once, like in CI recently.  Between the time juju checks to see if a profile has been created and creates it, a different controller as already created the juju-controller profile.  Thus the 2nd controller fails bootstrap.  Fix by checking for the profile if create profile fails.  IF the profile exists, we're good.   The profile always has the same contents. 

## QA steps

Not a good way to test the actual race.  Delete the juju-controller profile on one of the CI test boxes and kick off a CI run using this build?

```sh
lxd profile delete juju-controller
juju bootstrap lxd testme
```

Profiles are also named via model.  If you could time it so that multiple models with the same name in different controller are created at the same time successfully?

### DIFF
--- a/provider/lxd/environ_test.go
+++ b/provider/lxd/environ_test.go
@@ -375,6 +375,69 @@ func (s *environSuite) TestInstanceAvailabilityZoneNamesInvalidCredentials(c *gc
 	})
 }
 
+type environCloudProfileSuite struct {
+	lxd.EnvironSuite
+
+	callCtx context.ProviderCallContext
+
+	svr          *lxd.MockServer
+	cloudSpecEnv environs.CloudSpecSetter
+}
+
+var _ = gc.Suite(&environCloudProfileSuite{})
+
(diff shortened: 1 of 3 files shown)


provider/lxd/environ_test.go
```diff
@@ -430,17 +493,17 @@ func (s *environProfileSuite) TestAssignLXDProfiles(c *gc.C) {
 	defer s.setup(c).Finish()
 
 	instId := "testme"
-	old := "old-profile"
-	new := "new-profile"
-	expectedProfiles := []string{"default", "juju-default", new}
-	s.expectAssignLXDProfiles(instId, old, new, []string{}, expectedProfiles, nil)
+	oldP := "old-profile"
```
hmlanigan, line 496: drive-by, new collides with builtin function

hmlanigan:
$$merge$$

PR #11465, opened 2020-04-20
Go mod migration work

hpidcock:
## Description of change

- Move to go mod
- Drop Gopkg.*
- Drop GOPATH requirement
- Lock mockgen
- Fix Azure sdk deps
- Fix packages that needed to be merge (Gopkg.toml was pointing to
unmerged commits)
- Fix snapcraft to use go mod.
- Fix github actions to use go mod and go 1.14
- A bunch of linters were not actually linting (yay). Had to disable deadcode and golint because their fixes would take too long.
- Update CONTRIBUTING.md

## QA steps

Build juju, run unit tests, whatever you can do to test it.

## Documentation changes

N/A

## Bug reference

N/A


.gitignore
```diff
@@ -26,4 +27,4 @@ juju-backup-*.tar.gz
 .editorconfig
 scripts/mgo-run-txn/mgo-run-txn
 tests/tmp.*
-*.out
+*.out
```
howbazaar, line 30: Can you add the new line back in?

api/base/caller.go
```diff
@@ -15,7 +15,7 @@ import (
 	"gopkg.in/macaroon.v2"
 )
 
-//go:generate mockgen -package mocks -destination mocks/caller_mock.go github.com/juju/juju/api/base APICaller,FacadeCaller
+//go:generate go run github.com/golang/mock/mockgen -package mocks -destination mocks/caller_mock.go github.com/juju/juju/api/base APICaller,FacadeCaller
```
howbazaar, line 18: Why do we need this rather than just using mockgen assuming it was built?

api/caasfirewaller/client.go
```diff
@@ -89,7 +89,7 @@ func (c *Client) Life(appName string) (life.Value, error) {
 	if err := results.Results[0].Error; err != nil {
 		return "", maybeNotFound(err)
 	}
-	return life.Value(results.Results[0].Life), nil
+	return results.Results[0].Life, nil
```
howbazaar, line 92: Is this because later Go versions were better at similar type assignments? I'm assuming this is the case, and not a change in either underlying type.

api/base/caller.go
hpidcock: Because mockgen generates code that is checked in, we should all be using the same version. By using go run we force go to use the go.mod locked version.

api/caasfirewaller/client.go
hpidcock: No this is because the unconvert linter was never running properly, so these were supposed to be picked up before.

apiserver/apiserver.go
```diff
@@ -457,21 +457,6 @@ func (srv *Server) getAgentToken() error {
 	return nil
 }
 
-// loggoWrapper is an io.Writer() that forwards the messages to a loggo.Logger.
```
hpidcock, line 460: I need to revert this.

hpidcock:
!!build!!

hpidcock:
!!build!!

hpidcock:
!!build!!

hpidcock:
!!build!!

hpidcock:
!!build!!

hpidcock:
!!build!!

hpidcock:
$$merge$$

PR #11469, opened 2020-04-20
Move go vet github actions tests to own run

SimonRichardson:
## Description of change

The issue is that we're running out of swap space when attempting to run
go vet. This seems like a 2.8 (develop) issue over 2.7. Something has
changed in 2.8 that blows the testing/building of juju out of the
water.

We really should investigate this as it's only going to get worse in the
future.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -33,11 +33,13 @@ jobs:
       shell: bash
       run: |
         set -euxo pipefail
+        export PATH=/snap/bin:$PATH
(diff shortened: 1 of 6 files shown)


hpidcock:
Not a fan of this change. I have a bunch of fixes in my go mod PR that might help with this.

hpidcock:
I haven't had any issues with my go mod branch. I'm thinking we don't need this.

SimonRichardson:
@hpidcock I agree, the issue is that we've run out of credits for github actions.

PR #11473, opened 2020-04-21
When deleting a k8s app, wait for key k8s resources in the cluster

wallyworld:
## Description of change

When deleting a k8s app, wait for key k8s resources in the cluster
to transition from Terminating to removed before removing the app
from the Juju model. This prevents the same app from being deployed
a second time and having the still lingering k8s resources mess up
what Juju thinks about the state of the app.

We add a new application undertaker worker. All k8s apps are created with a "has resources" bool which is set true. The application destroy operation will not remove the app if this is true. Instead, the undertaker will call an API to clear this resource flag and that will finalise the removal of the app.

Much of the LOC in the diff are tests, the logic isn't that much.

Also a driveby fix for some flakey status history tests in state.

## QA steps

deploy a workload k8s charm
juju remove-application
watch kubectl get all and see that the pods stay in terminating for a bit while k8s processes
juju status should show the app still around until the pods disappear

repeat the above with an operator charm

## Bug reference

https://bugs.launchpad.net/juju/+bug/1870457


wallyworld:
[dangling k8s resources](https://trello.com/c/S1eErB03/1103-dangling-k8s-resources)

state/migration_import.go
```diff
@@ -1247,6 +1247,8 @@ func (i *importer) makeApplicationDoc(a description.Application) (*applicationDo
 		MetricCredentials:    a.MetricsCredentials(),
 		DesiredScale:         a.DesiredScale(),
 		Placement:            a.Placement(),
+		// TODO(caas)
```
hpidcock, line 1250: ?

state/migration_internal_test.go
```diff
@@ -398,6 +398,8 @@ func (s *MigrationSuite) TestApplicationDocFields(c *gc.C) {
 		// RelationCount is handled by the number of times the application name
 		// appears in relation endpoints.
 		"RelationCount",
+		// TODO(caas)
```
hpidcock, line 401: ?

state/migration_import.go
wallyworld: needs this to land https://github.com/juju/description/pull/76

state/migration_internal_test.go
wallyworld: ditto

wallyworld:
$$merge$$

PR #11475, opened 2020-04-21
Rename the MongoSnap fields to JujuDBSnap to match snap name.

howbazaar:
In order to avoid future confusion, we want to have the controller config
named 'juju-db-snap-channel' because the channel is for the juju-db snap.

This also means should we change what the underlying db is, we don't have
to change the config name. Although perhaps that is wishful thinking.


agent/format-2.0.go
```diff
@@ -60,7 +60,7 @@ type format_2_0Serialization struct {
 	SystemIdentity     string `yaml:"systemidentity,omitempty"`
 	MongoVersion       string `yaml:"mongoversion,omitempty"`
 	MongoMemoryProfile string `yaml:"mongomemoryprofile,omitempty"`
-	MongoSnapChannel   string `yaml:"mongosnapchannel,omitempty"`
+	JujuDBSnapChannel  string `yaml:"juju-db-snap-channel,omitempty"`
```
wallyworld, line 63: Perhaps can we make this `jujudb-snap-channel`

The other fields don't have hyphens (they should) and reducing the number we use in this new field looks a little tidier IMO

controller/config.go
```diff
@@ -133,9 +133,9 @@ const (
 	// detault
 	MongoMemoryProfile = "mongo-memory-profile"
 
-	// MongoSnapChannel selects the channel to use when installing mongo
+	// JujuDBSnapChannel selects the channel to use when installing mongo
 	// snaps for focal or later. The value is ignored for older releases.
-	MongoSnapChannel = "mongo-snap-channel"
+	JujuDBSnapChannel = "juju-db-snap-channel"
```
wallyworld, line 138: and here wrt other suggestion

worker/agentconfigupdater/manifold_test.go
```diff
@@ -139,7 +139,7 @@ func (s *AgentConfigUpdaterSuite) TestCentralHubMissing(c *gc.C) {
 				*result = params.ControllerConfigResult{
 					Config: map[string]interface{}{
 						"mongo-memory-profile": "default",
-						"mongo-snap-channel":   controller.DefaultMongoSnapChannel,
+						"juju-db-snap-channel": controller.DefaultJujuDBSnapChannel,
```
wallyworld, line 142: ditto

howbazaar:
$$merge$$

achilleasa:
@howbazaar I actually considered renaming this as you did here but decided against it in the end given that the other mongo options (e.g. the memory-profile one) start with `mongo-`. Should we also change any var/const names in the code to align with this change?

PR #11484, opened 2020-04-21
Increase log level so test can find Info log

hmlanigan:
## Description of change

Default root logging level for models other than controller and default is Warning.  Log messages looked for by the run-start-hook-fires-after-reboot, won't be found without a min level of Info.

## QA steps

```sh
(cd tests ;  ./main.sh -V -s test_dispatching_script hooks)
```


hmlanigan:
$$merge$$

PR #11486, opened 2020-04-22
Fixes watcher issuer with controller brokers.

tlm:
### Checklist

----

## Description of change

Change so namespace can not change on the kubernetes broker. It is set once per broker. This allows shared informers not to be broken.

## QA steps

Bootstrap controller into k8s, add models and deploy.

### DIFF
--- a/caas/kubernetes/provider/bootstrap.go
+++ b/caas/kubernetes/provider/bootstrap.go
@@ -28,7 +28,6 @@ import (
 	k8sannotations "github.com/juju/juju/core/annotations"
 	"github.com/juju/juju/core/watcher"
 	"github.com/juju/juju/environs"
-	environsbootstrap "github.com/juju/juju/environs/bootstrap"
 	"github.com/juju/juju/mongo"
 )
 
@@ -121,10 +120,7 @@ type controllerStacker interface {
 	Deploy() error
 }
 
-func controllerCorelation(broker *kubernetesClient) (*kubernetesClient, error) {
-	if broker.Config().Name() != environsbootstrap.ControllerModelName {
-		return broker, nil
-	}
+func controllerCorelation(broker *kubernetesClient) (string, error) {
 	// ensure controller specific annotations.
 	_ = broker.addAnnotations(annotationControllerIsControllerKey, "true")
 
@@ -133,16 +129,12 @@ func controllerCorelation(broker *kubernetesClient) (*kubernetesClient, error) {
 		// No existing controller found on the cluster.
 		// A controller must be bootstrapping now.
 		// It will reply on setControllerNamespace in controller stack to set namespace name.
-		return broker, nil
+		return "", errors.NewNotFound(err, "controller")
 	}
 	if err != nil {
-		return nil, errors.Trace(err)
+		return "", errors.Trace(err)
 	}
-	// This is an existing controller.
-	// Now, link it back.
-	// It should be always one record here based on current annotation design.
-	broker.SetNamespace(ns[0].GetName())
-	return broker, nil
+	return ns[0].GetName(), nil
 }
 
 // DecideControllerNamespace decides the namespace name to use for a new controller.
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -189,6 +189,7 @@ func newK8sBroker(
 	controllerUUID string,
 	k8sRestConfig *rest.Config,
 	cfg *config.Config,
+	namespace string,
 	newClient NewK8sClientFunc,
 	newWatcher NewK8sWatcherFunc,
 	newStringsWatcher NewK8sStringsWatcherFunc,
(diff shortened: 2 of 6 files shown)


caas/kubernetes/provider/k8s.go
```diff
@@ -65,6 +65,8 @@ import (
 var logger = loggo.GetLogger("juju.kubernetes.provider")
 
 const (
+	DefaultNamespace = "default"
```
wallyworld, line 68: Needs comment if exported, but no need to export
Actually, I'd still like a comment to explain what this is used for because unless you know the semantics of the code that uses it (which is in another file), it's not apparent. Maybe even move this to near where it's used.

caas/kubernetes/provider/provider.go
```diff
@@ -116,13 +116,24 @@ func (p kubernetesEnvironProvider) Open(args environs.OpenParams) (caas.Broker,
 		return nil, errors.Trace(err)
 	}
 	broker, err := newK8sBroker(
```
wallyworld, line 118: Can we add a comment to explain that this first broker is to look for the namespace that corresponds with the juju model name

caas/kubernetes/provider/bootstrap.go
```diff
@@ -121,9 +121,9 @@ type controllerStacker interface {
 	Deploy() error
 }
 
-func controllerCorelation(broker *kubernetesClient) (*kubernetesClient, error) {
+func controllerCorelation(broker *kubernetesClient) (string, error) {
 	if broker.Config().Name() != environsbootstrap.ControllerModelName {
```
wallyworld, line 125: I think this check should go into the Open() method.
The Only reason we can to call controllerCorelation() is when we are trying to find the namespace used for the controller model. Otherwise the namespace to use is just the model name. And with this refactor, we should adjust the logic. The logic the caller uses to check ns == "" etc is a bit convoluted and things can be simlified.

caas/kubernetes/provider/k8s.go
tlm: Done & Done

caas/kubernetes/provider/provider.go
tlm: Done & Done

caas/kubernetes/provider/bootstrap.go
tlm: Take a look at that change.

caas/kubernetes/provider/provider.go
```diff
@@ -30,6 +31,13 @@ type kubernetesEnvironProvider struct {
 	brokerGetter       func(environs.OpenParams) (caas.ClusterMetadataChecker, error)
 }
 
+const (
+	// DefaultNamespace is the default namespace to use for brokers that need to
+	// be constructed where the target namespace is not yet known. Think
+	// bootstrapping or making a broker for an unknown controller namespace.
+	DefaultNamespace = "juju"
```
wallyworld, line 38: thanks for comment :-)
let's not export though unless absolutely needed outside package

caas/kubernetes/provider/provider.go
```diff
@@ -115,14 +123,34 @@ func (p kubernetesEnvironProvider) Open(args environs.OpenParams) (caas.Broker,
...
 	}
+
+	// Guinea Pig broker to hunt for the namespace where a controller lives. We
+	// disregard this one in favour of a new one pinned to the correct
+	// controller namespace when we find it.
 	broker, err := newK8sBroker(
-		args.ControllerUUID, k8sRestConfig, args.Config, newK8sClient,
-		newKubernetesNotifyWatcher, newKubernetesStringsWatcher, randomPrefix,
-		jujuclock.WallClock)
+		args.ControllerUUID, k8sRestConfig, args.Config, DefaultNamespace,
```
wallyworld, line 131: Shouldn't we be constructing the broker with broker.Config().Name() here?
We return it on line 140 below with an incorrect namespace if model is not "controller".
In which case, the "Guinea Pig" comment is slightly inaccurate because what we are doing is constructing what we think is the broker to use, and only doing something else if it is a controller model.

I think the flow should be:

1. look at broker.Config().Name()
2. if controller model, construct generic broker to find ns to use
3. construct result broker to return with the correct namespace

caas/kubernetes/provider/bootstrap.go
wallyworld: With the refactoring below, we no longer need
```
if broker.Config().Name() != environsbootstrap.ControllerModelName {
}
```
here

caas/kubernetes/provider/provider.go
tlm: Removed because of comment below

caas/kubernetes/provider/provider.go
tlm: Take a look at that. Thanks for the pickup

caas/kubernetes/provider/bootstrap.go
tlm: Done & Done

tlm:
Try that @wallyworld

tlm:
$$merge$$

PR #11491, opened 2020-04-23
Unit agent worker loggers

howbazaar:
The purpose of this work is to pass loggers into the worker manifolds and remove the package level loggers. This work is particularly targeting the workers for the uniter. This work will mean when we merge the unit and machine agents the logs from the uniter workers will go to the right place.

There is a drive-by fix for the intermittent test failure in the featuretests package for the metrics watcher. The old attempt at fixing the failure with a WaitForModelWatchersIdle was not correct. The new fix has been tested under stress and stress-race loops locally.

Some worker tests were reworked a little so we didn't repeatedly have to add the loggers to many places, and instead makes use of suite methods.

## QA steps

Bootstrap and deploy and agent. If the engine manifolds all start up properly, then everything is wired up. Need to confirm with k8s and lxd to ensure we are covering all aspects of the manifolds.

## Documentation changes

Everything is internal, nothing to document.


worker/caasoperator/caasoperator.go
```diff
@@ -213,10 +213,15 @@ func NewWorker(config Config) (worker.Worker, error) {
 		return nil, errors.Trace(err)
 	}
 	paths := config.getPaths()
+	logger := loggo.GetLogger("juju.worker.uniter.charm")
```
wallyworld, line 216: Maybe pass this 

loggo.GetLogger("juju.worker.uniter")

in as a second logger (eg uniterLogger) on the Config struct, and then use Child("charm") off that second logger and also below...

howbazaar: Passing through the charm logger here keeps the logging at the same place. It is because both of the types here were from the charm package, and hence used that logger.

I do feel like we could do a cleanup pass over everything later to clean up some of these.

howbazaar:
$$merge$$

PR #11499, opened 2020-04-27
Fix 1862026 - inconsistent output from action-get

hpidcock:
## Fix 1862026 - inconsistent output from action-get

Using --format=json and empty parameters results in null not
empty dictionary.

## QA steps

Unit tests should suffice. But you can run an action that returns the result of action-get --format=json
The action must have no params defined.

## Documentation changes

N/A - I think this change is in line with expected use.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1862026


hpidcock:
$$merge$$

PR #11502, opened 2020-04-27
Fix typo in app status params struct

wallyworld:
## Description of change

The json tag for the CharmVersion attribute in the ApplicationStatus params struct had a typo.
To maintain compatibility with existing CLI clients, fix the struct and use custom json marshalling to include the typo.

## QA steps

Run juju status with an old and new CLI client and check charm version.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1871622


wallyworld:
$$merge$$

PR #11506, opened 2020-04-28
Fix failure in TestControllerConfigPubsubChange

howbazaar:
Fixes an intermittent test failure in worker.modelcache TestControllerConfigPubsubChange.

After many many repeated runs, I was able to reproduce locally. In the end it was a race
between the test and the behaviour of the cacheWorker loop function. It was possible that
the initial controller config change was notified by the worker, and before the subscription
to the pubsub topic was done the test advanced and published a message that the worker
should have responded to.

The fix is to move the message subscription before the init method that does the initial
notify of the config changed. In the real world there isn't a problem here as the apiserver
isn't running while the model cache is initializing, so no one would be modifying the
controller config at that time.

I do feel though that there is an issue with the general gate mechanism, as the gates
are only created once, can never be closed again, so the initialization gate seems to
only block things the first time the worker starts. If mongo drops a connection and
causes the state worker to bounce, so will the model cache worker and the apiserver.
The gates however don't get recreated, so there isn't the expected synchronization
the second time though. This is a bigger problem and impacts all the agents using
gates, and will be a different piece of work to fix.

### DIFF
--- a/worker/modelcache/worker.go
+++ b/worker/modelcache/worker.go
@@ -197,16 +197,24 @@ func (c *cacheWorker) loop() error {
 	defer c.config.PrometheusRegisterer.Unregister(allWatcherStarts)
 	defer c.config.PrometheusRegisterer.Unregister(collector)
 
-	if err := c.init(); err != nil {
-		return errors.Trace(err)
-	}
-
+	// Ensure that we are listening for updates before we send the initial
+	// controller config update. In reality, there will be no config changed events
+	// published until the initialize gate is unlocked, as the API server won't
+	// yet be running. However in tests, there is a situation where the test waits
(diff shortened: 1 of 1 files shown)


howbazaar:
Sure

howbazaar:
@SimonRichardson updated.

howbazaar:
$$merge$$

howbazaar:
$$merge$$ intermittent test failure: FAIL: bootstrap_test.go:1323: BootstrapSuite.TestAutoUploadAfterFailedSync

howbazaar:
$$merge$$ intermittent test failure FAIL: machine_test.go:214: MachineLegacyLeasesSuite.TestDyingMachine
 (machine didn't die)

PR #11512, opened 2020-04-29
Fix bogus error for update-credential --force

wallyworld:
## Description of change

The cloud  facade update credentials api was always setting an error value about invalid models even if force is used. As explained in the bug, this confuses the client because with force, the credential update did succeed, but there's no way to tell that as the error is always set. Also, any real error from updating the credential failing overwrites this model error anyway.

The other thing is that the result struct already contains a list of models with errors, so this can be used to see if any models are affected instead.

The PR changes the behaviour of the update credentials api to behave correctly, but also work the old way for older facades. The CLI is also tweaked to improve what's output based on the new behaviour. 

## QA steps

Because the cloud api caller has been updated to mask the old irrelevant error on older controllers, the new CLI will produce the same results on a 2.7 controller or a 2.8 one.

bootstrap a aws controller
edit the credential to contain an invalid secret
juju update-credential aws <credname>

observe the invalid models and a message explaining fix the credential

juju update-credential aws <credname>--force

observe the invalid models and a message explaining that some models need fixing since the cred was updated

update the cred to be good again

juju update-credential aws <credname>

the cred should update without issue

## Bug reference

https://bugs.launchpad.net/bugs/1873273

### DIFF
--- a/apiserver/facades/schema.json
+++ b/apiserver/facades/schema.json
@@ -11521,7 +11521,7 @@
     },
     {
         "Name": "Cloud",
-        "Version": 6,
+        "Version": 7,
         "Schema": {
             "type": "object",
             "properties": {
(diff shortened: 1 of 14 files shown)


apiserver/facades/schema.json
```diff
@@ -11544,17 +11544,6 @@
                         }
                     }
                 },
-                "CheckCredentialsModels": {
```
babbageclunk, line 11547: Don't we need this method anymore?

wallyworld: It's still on V6 facade for compatibility but removed from latest/V7 as it's not used. So old clients can still use it by asking for the V6 facade.

wallyworld:
$$merge$$

PR #11513, opened 2020-04-29
Various fixes to image metadata commands.

wallyworld:
## Description of change

Create the facade always - it was behind a feature flag
Ensure default region is set correctly - when not specified, it needs to be set to the model region.
Do not expire custom image metadata - all metadata added via add-image was being expired but custom metadata needs to be kept.

## QA steps

bootstrap on aws
export JUJU_FEATURES=image-metadata

ensure that juju metadata add-image can be used to add metadata for say centos7, and that if region is left off, the model region is used.
metadata can be viewed with juju metadata list-images
wait 5 minutes ahd list again to ensure it doesn't expire

### DIFF
--- a/apiserver/allfacades.go
+++ b/apiserver/allfacades.go
@@ -7,7 +7,6 @@ import (
 	"reflect"
 
 	"github.com/juju/errors"
-	"github.com/juju/featureflag"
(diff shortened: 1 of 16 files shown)


wallyworld:
$$merge$$

PR #11511, opened 2020-04-29
[2.7] Support spaces in zones and instance type constraints

achilleasa:
## Description of change

This PR back-ports the fix for zones with spaces from #11500 and also applies the same change for values passed to `instance-type` constraints.

## QA steps

`juju bootstrap rackspace test --constraints instance-type=1 GB General Purpose v1`

## Bug reference

- https://bugs.launchpad.net/juju/+bug/1875590
- https://bugs.launchpad.net/juju/+bug/1847259

PR #11500, opened 2020-04-27
Allow to set zones that contains spaces

Pekkari:
Closes-bug: #1847259
Signed-off-by: José Pekkarinen <jose.pekkarinen@canonical.com>

### Checklist

 - [ x ] Checked if it requires a [pylibjuju](https://github.com/juju/python-libjuju) change?
 - [ x ] Added [integration tests](https://github.com/juju/juju/tree/develop/tests) for the PR?
 - [ x ] Added or updated [doc.go](https://discourse.jujucharms.com/t/readme-in-packages/451) related to packages changed?
 - [ x ] Do comments answer the question of why design decisions were made?

----

## Description of change

It's found that in multiple environments exist policies to set availability zones
that contains spaces in the name. With the given patch, it's possible to do
so, and a test to proof it is provided.

## QA steps

go test github.com/juju/juju/core/constraints

## Bug reference

LP: #1847259

*Please add a link to any bugs that this change is related to.*

https://bugs.launchpad.net/juju/+bug/1847259

PR #11519, opened 2020-04-30
[2.7] Handle slash-escaped spaces in constraint strings

achilleasa:
## Description of change

A fix that landed previously (see #11500 and its 2.7 backport #11511) allowed unescaped constraint values with spaces but made the constraint-parsing code more convoluted and relied
on the removal of already-set checks to function properly. This made the implementation flaky as 
the operator would now be allowed to provide a constraint such as `zones=a zones=b` and receive no error from juju (last write wins semantics).

This PR reverts the previous set of changes and introduces a better way for dealing with constraint values with spaces by expecting the operator to slash-escape them as you would normally do in a shell. The fix applies to all constraint values and does not require any modifications to the rest of the validation logic.

To this end, the operator can now enter constraints such as:
`--constraint arch=amd64 zones=zone\ 1,zone2 instance-type=general\ purpose`

## QA steps

```console
$ juju bootstrap rackspace test --constraints instance-type=1 GB General Purpose v1
```

## Documentation changes

We should update the docs to reflect the fact that spaces are now supported if properly escaped.

## Bug reference

Followup fix for:
- https://bugs.launchpad.net/juju/+bug/1875590
- https://bugs.launchpad.net/juju/+bug/1847259

### DIFF
--- a/core/constraints/constraints.go
+++ b/core/constraints/constraints.go
@@ -259,6 +259,12 @@ func (v Value) String() string {
 		s := strings.Join(*v.Zones, ",")
 		strs = append(strs, "zones="+s)
 	}
+
+	// Ensure constraint values with spaces are properly escaped
+	for i := 0; i < len(strs); i++ {
+		strs[i] = strings.Replace(strs[i], " ", `\ `, -1)
+	}
+
 	return strings.Join(strs, " ")
 }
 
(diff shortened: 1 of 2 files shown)


core/constraints/constraints.go
```diff
@@ -330,28 +336,23 @@ func Parse(args ...string) (Value, error) {
 func ParseWithAliases(args ...string) (cons Value, aliases map[string]string, err error) {
```
achilleasa, line 336: For comparison, this was the original code before the previous fix landed: https://github.com/juju/juju/blob/ed1f458896f65f30712624580f72d020cbc8f827/core/constraints/constraints.go#L330-L356

achilleasa:
$$merge$$ 

(CI passed but windows box has mgo/socket issues)

PR #11523, opened 2020-04-30
uniter relationStateTracker  and relationer mock unit tests

hmlanigan:
## Description of change

Add explicit tests for the uniter/relation StateTracker.  Previously implicitly tested as part of the relationResolverSuite.  Part of an effort to remove jujuconnsuite from uniter/relation.

Replace uniter/relation relationer unit tests with mock versions.

## QA steps

There should be no change to currently behavior.  Test with units and relations.


hmlanigan:
More than half of the changes are for the generated mock files.

worker/uniter/relation/relationer.go
```diff
@@ -14,26 +14,26 @@ import (
...
 	stateMgr   StateManager
 	dying      bool
 }
 
-// NewRelationer creates a new Relationer. The unit will not join the
+// NewRelationer creates a new relationer. The unit will not join the
 // relation until explicitly requested.
-func NewRelationer(ru RelationUnit, stateMgr StateManager) *Relationer {
-	return &Relationer{
+func NewRelationer(ru RelationUnit, stateMgr StateManager) Relationer {
```
achilleasa, line 27: In the spirit of "accept interfaces and return concrete types", can you rename and export `relationer` (e.g. RelationManager or something like that) and have `NewRelationer` return a concrete instance that satisfies `Relationer`? You can add a small compile-time check (e.g. `var _ Relationer = (*ConcreteType)(nil)`) to ensure it satisfies the interface.

hmlanigan: I gave it a try.  Unfortunately it breaks down because of: [statetracker.go ln52](https://github.com/juju/juju/blob/8ed98bba62d60df44e5648962bf25bb10060f488/worker/uniter/relation/statetracker.go#L52) which has a func definition as a variable. It is the only place NewRelationer is used outside of test and would require a wrapper to convert *relationer into a Relationer.  Which defeats the purpose of the change in my mind.

This is a handy thing to know for future work though!  TY.

hmlanigan:
$$merge$$

PR #11525, opened 2020-05-01
Add skip-storage option to add-k8s

wallyworld:
## Description of change

A small tweak to add-k8s CLI to allow a --skip-storage option.
This allows k8s clusters without storage to be added to Juju.

## QA steps

Delete any storage classes from a k8s cluster
juju add-k8s --skip-storage
should work as expected to add the cluster to juju.
Deploy a charm with min-juju-version set to 2.8.0 to ensure no storage is used.

### DIFF
--- a/cmd/juju/caas/add.go
+++ b/cmd/juju/caas/add.go
@@ -92,6 +92,9 @@ storage class to provide operator and workload storage. If none is found, use
 of the --storage option is required so that Juju will create a storage class
 with the specified name.
 
+If the cluster does not have a storage provisioning capability, use the
+--skip-storage option to add the cluster without any workload storage configured.
+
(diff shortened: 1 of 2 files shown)


wallyworld:
If you are asking whether we validate that there's storage when bootstrapping, the answer is yes. We will exit early if bootstrapping to a k8s cluster where there's been no operator-storage set up.
Adding a k8s cloud without storage only makes sense to do so to an existing controller.

wallyworld:
$$merge$$

wallyworld:
[remove mandatory storage with add-k8s](https://trello.com/c/HTvb1BGp/1113-remove-mandatory-storage-with-add-k8s)

PR #11233, opened 2020-02-19
[dev] allow leader unit to read the databag for its application

achilleasa:
## Description of change

This PR adds a bit of extra logic to the `relation-get` tool implementation so that leader units can introspect the contents of their application databag.

## QA steps

```sh
$ juju bootstrap lxd --no-gui
$ juju deploy wordpress
$ juju deploy mysql

# In a second terminal  run this command
$ juju debug-hooks mysql/0 db-relation-joined

# Go back to another terminal and run:
$ juju relate wordpress mysql

# Wait for the debug-hook to trigger and in the tmux session run the following commands:
#-----------------------
$ ./hooks/db-relation-joined
$ is-leader
True
$ relation-set --app answer-to-the-ultimate-question=42
$ relation-get --app - mysql        # <- this command would fail before this fix
answer-to-the-ultimate-question: "42"
$ relation-get --app - wordpress
{}
$ env | grep JUJU_RELATION_ID
JUJU_RELATION_ID=db:2          # <- note the relation ID for the next step
$ exit
$ exit  # second exit exits 'juju debug-hooks'
#-----------------------

# Verify that the mutated application data was persisted correctly
# NOTE: replace "db:2" with the relation ID from the above step!
$ juju run --unit mysql/0 'relation-get --app -r db:2 - mysql'
answer-to-the-ultimate-question: "42"
$ juju run --unit mysql/0 'relation-get --app -r db:2 - wordpress'
{}
```

## Bug reference
https://bugs.launchpad.net/juju/+bug/1854348

PR #11526, opened 2020-05-04
[dev] Access local application's data-bag from k8s charms

achilleasa:
## Description of change

The fix from #11233 allows leader charms (non-k8s ones) to read their own application's data bag (the local application). However, that fix does not apply to k8s charms due to the way that the leadership checks are performed by the uniter facade.

More specifically, so as to implement a leadership check prior to reading the application's data bag contents, the controller needs two pieces of information: the application and the unit tag. Prior to this change, in order to fetch the application settings, the code abused the `ReadSettings` API call (which normally expects a relation and a **unit** tag as arguments) and passed an **application** tag instead of a unit tag. The facade implementation examined the type of the tag and used a different code path for applications (which ultimately leads to a leadership check).

For **non-k8s** workloads, each unit runs its own agent which logs in to the controller as the unit. To this end, the facade code pulls the unit tag from the authentication details. If the unit is indeed the leader everything works as expected; otherwise, a permission error is returned.

K8s workloads work differently: for each application, we deploy a single operator instance which executes any required operation on behalf of a particular unit. The operator logs in as the **application** which makes it impossible to figure out a unit tag for the leadership check. As a result, attempts to read the application data bag would always fail with a permission error.

The fix implemented by this PR is to create a new API call (`ReadLocalApplicationSettings`) that receives a `RelationUnit` as an argument and infers the application to be accessed from the unit tag. Note that contrary to other API calls that expect a list of things to operate on, no client will ever be able to read the local application data for another application and therefore, the call expects a **single** argument.

The previous API call has been modified to redirect to the new implementation when reading application data thus preserving the previous behavior when an older (e.g. 2.7.x agent) is talking to a newer controller. Of course, this means that older models will still not be able to read the local application data (for k8s charms) **unless they get upgraded**.

:warning: This fix comes with a security caveat: since there is no way for the controller to verify the validity of the unit tag passed by the client (remember, the operator logs in as the application), the controller assumes that application-level logins should be trusted implicitly and therefore uses the unit-tag, _as-is_ for performing the leadership check. This makes it possible for charm authors that have access to the operator credentials to use pylibjuju and read/write the application data by spoofing the unit-tag to the one that is currently the leader.

### k8s charms

```console
# Initial setup
$ juju deploy cs:~aisrael/bundle/mediawiki-k8s-0
$ juju add-unit mediawiki-k8s

# Leader should be able to write and read the data bag
$ juju run --unit mediawiki-k8s/0 'relation-set -r 0 --app myapp=mediawiki'
$ juju run --unit mediawiki-k8s/0 'relation-get -r 0 --app - mediawiki-k8s'
myapp: mediawiki

# Non-leaders should not be allowed to read or write the *local* data bag
$ juju run --unit mediawiki-k8s/1 'relation-set -r 0 --app myapp=mediawiki'                                                                                                                    
ERROR cannot write relation settings
$ juju run --unit mediawiki-k8s/1 'relation-get -r 0 --app - mediawiki-k8s'
ERROR permission denied

# Note that any remote unit in this relation should be able to read the mediawiki's data bag
$ juju run --unit mariadb-k8s/0 'relation-get -r 0 --app - mediawiki-k8s'
myapp: mediawiki
```

### non-k8s charms

```console
# Initial setup
$ juju deploy cs:bundle/mediawiki-single-9
$ juju add-unit mediawiki

# Leader should be able to write and read the data bag
$ juju run --unit mediawiki/0 'relation-set -r 1 --app myapp=mediawiki'
$ juju run --unit mediawiki/0 'relation-get -r 1 --app - mediawiki'
myapp: mediawiki

# Non-leaders should not be allowed to read or write the *local* data bag
$ juju run --unit mediawiki/1 'relation-set -r 1 --app myapp=mediawiki'                                                                                                                    
ERROR cannot write relation settings
$ juju run --unit mediawiki/1 'relation-get -r 1 --app - mediawiki'
ERROR permission denied

# Note that any remote unit in this relation should be able to read the mediawiki's data bag
$ juju run --unit mysql/0 'relation-get -r 1 --app - mediawiki'
myapp: mediawiki
```

### 2.8 controller with 2.7 model and non-k8s charm still works using the deprectated API call

```console
# Initial setup (install juju 2.7.6 from snap)
$ /snap/bin/juju bootstrap lxd test --no-gui
$ /snap/bin/juju deploy cs:bundle/mediawiki-single-9
$ /snap/bin/juju add-unit mediawiki

# Run the same set of steps as the previous test
$ /snap/bin/juju run --unit mediawiki/0 'relation-set -r 1 --app myapp=mediawiki'
$ /snap/bin/juju run --unit mediawiki/0 'relation-get -r 1 --app - mediawiki'
myapp: mediawiki

# Note this one silently fails; no data is actually written.
$ /snap/bin/juju run --unit mediawiki/1 'relation-set -r 1 --app myapp=mediawiki999'          
$ /snap/bin/juju run --unit mediawiki/1 'relation-get -r 1 --app - mediawiki'                                                                                                                  
ERROR permission denied

# Verify that data did not get mutated!
$ /snap/bin/juju run --unit mediawiki/0 'relation-get -r 1 --app - mediawiki'
myapp: mediawiki

# Upgrade controller (note: model is still 2.7.6)

# TODO: Run the same steps as above and ensure that output remains consistent

# Now, upgrade the model to 2.8
$ juju upgrade-model

# TODO: Run the same steps as above and ensure that output remains consistent
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1876097

### DIFF
--- a/apiserver/facades/agent/uniter/uniter.go
+++ b/apiserver/facades/agent/uniter/uniter.go
@@ -38,8 +38,8 @@ import (
 var logger = loggo.GetLogger("juju.apiserver.uniter")
 
 // UniterAPI implements the latest version (v15) of the Uniter API, which adds
-// the State, CommitHookChanges calls and changes WatchActionNotifications to
-// notify on action changes.
+// the State, CommitHookChanges, ReadLocalApplicationSettings calls and changes
+// WatchActionNotifications to notify on action changes.
 type UniterAPI struct {
 	*common.LifeGetter
 	*StatusAPI
@@ -1610,15 +1610,15 @@ func (u *UniterAPI) LeaveScope(args params.RelationUnits) (params.ErrorResults, 
 
 // ReadSettings returns the local settings of each given set of
 // relation/unit.
+//
+// NOTE(achilleasa): Using this call to read application data is deprecated
+// and will not work for k8s charms (see LP1876097). Instead, clients should
+// use ReadLocalApplicationSettings.
 func (u *UniterAPI) ReadSettings(args params.RelationUnits) (params.SettingsResults, error) {
 	result := params.SettingsResults{
 		Results: make([]params.SettingsResult, len(args.RelationUnits)),
 	}
 	canAccessUnit, err := u.accessUnit()
-	if err != nil {
-		return params.SettingsResults{}, errors.Trace(err)
-	}
-	canAccessApp, err := u.accessApplication()
 	if err != nil {
 		return params.SettingsResults{}, errors.Trace(err)
 	}
@@ -1643,27 +1643,15 @@ func (u *UniterAPI) ReadSettings(args params.RelationUnits) (params.SettingsResu
 			settings = node.Map()
 
 		case names.ApplicationTag:
-			var relation *state.Relation
-			relation, err = u.getRelation(arg.Relation)
-			if err != nil {
-				return nil, errors.Trace(err)
+			// Emulate a ReadLocalApplicationSettings call where
+			// the currently authenticated tag is implicitly
+			// assumed to be the requesting unit.
+			authTag := u.auth.GetAuthTag()
+			if authTag.Kind() != names.UnitTagKind {
+				// See LP1876097
+				return nil, common.ErrPerm
 			}
-			endpoints := relation.Endpoints()
-			isPeerRelation := len(endpoints) == 1 && endpoints[0].Role == charm.RolePeer
-			token := u.leadershipChecker.LeadershipCheck(tag.Id(), u.auth.GetAuthTag().Id())
-			canAccess := func(tag names.Tag) bool {
-				if !canAccessApp(tag) {
-					return false
-				}
-				if isPeerRelation {
-					return true
-				}
-				// For provider-requirer relations only allow the
-				// leader unit to read the application settings.
-				return token.Check(0, nil) == nil
-			}
-			settings, err = u.getRelationAppSettings(canAccess, arg.Relation, tag)
-
+			settings, err = u.readLocalApplicationSettings(arg.Relation, tag, authTag.(names.UnitTag))
 		default:
 			return nil, common.ErrPerm
 		}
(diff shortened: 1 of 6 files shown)


achilleasa:
@howbazaar @wallyworld This PR implements my proposal from the email I sent to the crew list with the security caveat that I mentioned. Can you take a look and +1 before we land this?

hmlanigan:
Got an empty set on the last command of the k8s charm qa steps: https://pastebin.canonical.com/p/b46yRMvvmJ/

achilleasa:
@hmlanigan copying my answer from IRC for future reference.

In the linked pastebin, the `relation-set` command was executed on the **non-leader** unit and failed. Therefore, running `relation-get` on the leader returns `{}` because nothing has been set yet in the _application_ databag. Note that most k8s charms out there do not use the application data bag feature (it's a 2.7+ thing) and instead exchange data via the _unit_ databags. You can verify this by running the `relation-get` commands without `--app` and with the _unit name_ at the end. For example:

```console
$ juju run --unit mediawiki-k8s/0 'relation-get -r 0 - mediawiki-k8s/0'
egress-subnets: 10.152.183.28/32
ingress-address: 10.152.183.28
private-address: 10.152.183.28
```

hmlanigan:
QA is good.

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -1610,6 +1627,10 @@ func (u *UniterAPI) LeaveScope(args params.RelationUnits) (params.ErrorResults,
 
 // ReadSettings returns the local settings of each given set of
 // relation/unit.
+//
+// NOTE(achilleasa): Using this call to read application data is deprecated
+// and will not work for k8s charms (see LP1876097). Instead, clients should
+// use ReadLocalApplicationSettings.
```
hmlanigan, line 1633: can we make this not available in version 16 then?

apiserver/params/internal.go
```diff
@@ -342,6 +342,13 @@ type ControllerAPIInfoResults struct {
 	Results []ControllerAPIInfoResult `json:"results"`
 }
 
+// RelationApplicationUnit holds a relation, an application and a unit tag.
+type RelationApplicationUnit struct {
+	Relation    string `json:"relation"`
+	Application string `json:"application"`
```
hmlanigan, line 348: Is it possible to have a application that is different from the unit?  Do we want this to be possible?  If not, should we derive the application tag from the unit tag?  names.UnitApplication.

apiserver/facades/agent/uniter/uniter.go
achilleasa: No you still need it for things to work when you upgrade your controller and before you upgrade your models (the not-yet-upgraded agents still make use of the call).

apiserver/params/internal.go
achilleasa: Hmmm... I think you are right. I just added it there for future proofing (just in case we ever decide to allow `Application != unitApplication(unit)`) given what happened with the unit-that-is-really-an-app-tag field in `RelationUnit`. If you think it's better to omit it, I can actually drop the new struct and replace it with `RelationUnit`.

apiserver/facades/agent/uniter/uniter.go
hmlanigan: Non-upgraded models will be using version 15, so it'd still be available there.  Just not version 16 and beyond.  We have ways to say, this method is only in version 16 and beyond, why not the opposite?

achilleasa: Note that the method is deprecated for reading the application data-bag _only_. It is still used for reading the unit's data bag (which was its original purpose AFAICT). Therefore, we cannot remove it.

hmlanigan: I wasn't clear on the overloading here.

apiserver/params/internal.go
hmlanigan: Agreed, overloading a tag is bad.  Perhaps we just need to be very clear in expectations between the 2 tags.  This would primarily be an issue if used from pylibjuju I'm thinking.  

If the app from the unit tag doesn't match the app, it errors? Otherwise results from "u.leadershipChecker.LeadershipCheck(appTag.Id(), unitTag.Id())" could be interesting.

wallyworld: I think having the new "ReadLocalApplicationSettings" is the crux of the change needed.
And having that new API means that RelationUnit struct is all that's needed.
The check would be that the unit attribute either matches the auth tag (if a unit is connecting) or is a valid unit of the application (if an application is connecting).

achilleasa: The PR currently includes the auth'd unit matches the unit in the params struct. Is the latter check for the k8s case actually required given that you are only allowed to proceed if you are the leader? A bogus unit name would always fail the leadership check and get back an ErrPerm. Or do we want to be more explicit about the check?

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -1682,6 +1687,87 @@ func (u *UniterAPI) ReadSettings(args params.RelationUnits) (params.SettingsResu
...
+		// provided by the caller.
+		if authTag.String() != unitTag.String() {
+			return res, errors.Trace(common.ErrPerm)
+		}
+	case names.ApplicationTagKind:
+		// In this case (k8s operator), we have no alternative than to
+		// implicitly trust the unit tag argument passed in by the
+		// operator (for more details, see LP1876097). As an early
+		// check, ensure that the application tags match.
+		if authTag.String() != appTag.String() {
```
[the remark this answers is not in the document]
achilleasa: This was added purely for symmetry with the unit check in the other switch case. I will remove it and replace it with a note that this check will be performed at a later stage.

api/facadeversions.go
```diff
@@ -106,7 +106,7 @@ var facadeVersions = map[string]int{
 	"Subnets":                      4,
 	"Undertaker":                   1,
 	"UnitAssigner":                 1,
-	"Uniter":                       15,
+	"Uniter":                       16,
```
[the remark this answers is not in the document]
achilleasa: I bumped it just to be on the safe side because I wasn't sure about pylibjuju (which frustratingly includes the schema for the agent API calls in addition to the client ones). Based on the feedback from @hmlanigan and the above comment I will revert this change and revise the PR later today.

achilleasa:
$$merge$$

PR #11527, opened 2020-05-04
Handle the case when there is no hook to run.

jameinel:
### Checklist

----

## Description of change

Charms may not implement hooks/foo and when running 'juju debug-code' it
shouldn't error out.
After talking with several people the desired behavior is to *not* act like 'debug-hooks' for hooks which aren't implemented. So now, instead of erroring out if the script isn't there, we just log a line about:
```
unit-dummy-source-2: 19:28:26 INFO juju.worker.uniter.runner debug-code active, but hook sink-relation-created not implemented (skipping)
```

Similarly we update the log message when it does exist to:
```
unit-dummy-source-2: 19:28:27 INFO juju.worker.uniter.runner executing sink-relation-joined via debug-code; explicit, bespoke hook script
unit-dummy-source-2: 19:28:27 INFO unit.dummy-source/2.juju-log sink:10: debug running /var/lib/juju/agents/unit-dummy-source-2/charm/hooks/sink-relation-joined for sink-relation-joined
```

## QA steps

```sh
$ juju bootstrap --debug lxd lxd
$ cd acceptancetests/repository/charms
$ juju deploy ./dummy-sink
$ juju deploy ./dummy-source
$ juju debug-code dummy-source/0
```
In another window run
```sh
juju debug-log --level INFO
```
and in a 3rd window run
```sh
juju add-relation dummy-sink dummy-source
```

In debug-log you should see lines about:
```
unit-dummy-source-0: 19:28:26 INFO juju.worker.uniter.runner debug-code active, and hook sink-relation-created not implemented (skipping)
```

Without this change, you would see errors on relation-created and Juju will keep retrying to execute the relation-created hook (via debug-code).

## Documentation changes

As part of documenting 'debug-code' we should note that it differs from 'debug-hooks' when the hook *doesn't* exist. Instead of dropping you into a terminal and letting you know the event happened, we just keep going until you debug actual code from your script.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1876290

### DIFF
--- a/worker/uniter/runner/debug/server.go
+++ b/worker/uniter/runner/debug/server.go
@@ -218,7 +218,11 @@ const debugHooksHookScript = `#!/bin/bash
 echo $$ > $JUJU_DEBUG/hook.pid
 if [ -z "$JUJU_DEBUG_AT" ] ; then
 	exec /bin/bash --noprofile --init-file $JUJU_DEBUG/init.sh
+elif [ ! -x "__JUJU_HOOK_RUNNER__" ] ; then
+    juju-log --log-level INFO "debugging is enabled, but no handler for $JUJU_HOOK_NAME, skipping"
+    echo 0 > $JUJU_DEBUG/hook_exit_status
 else
+    juju-log --log-level INFO "debug running __JUJU_HOOK_RUNNER__ for $JUJU_HOOK_NAME"
 	__JUJU_HOOK_RUNNER__
 	echo $? > $JUJU_DEBUG/hook_exit_status
 fi
--- a/worker/uniter/runner/debug/server.go
+++ b/worker/uniter/runner/debug/server.go
@@ -34,6 +34,11 @@ func (s *ServerSession) MatchHook(hookName string) bool {
 	return s.hooks.IsEmpty() || s.hooks.Contains(hookName)
 }
 
+// DebugAt returns the location for the charm to stop for debugging, if it is set.
+func (s *ServerSession) DebugAt() string {
+	return s.debugAt
+}
+
 // waitClientExit executes flock, waiting for the SSH client to exit.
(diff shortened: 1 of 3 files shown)


worker/uniter/runner/debug/server.go
```diff
@@ -218,7 +223,11 @@ const debugHooksHookScript = `#!/bin/bash
 echo $$ > $JUJU_DEBUG/hook.pid
 if [ -z "$JUJU_DEBUG_AT" ] ; then
 	exec /bin/bash --noprofile --init-file $JUJU_DEBUG/init.sh
+elif [ ! -x "__JUJU_HOOK_RUNNER__" ] ; then
+	juju-log --log-level INFO "debugging is enabled, but no handler for $JUJU_HOOK_NAME, skipping"
```
jameinel, line 227: We could get rid of this logging.
It was needed for me to actually understand what is going on, but it is plausible that it ends up too verbose in actual debug output. Especially since this is a belt-and-suspenders sort of thing, as we shouldn't actually set up hook.sh if we notice there is no script to run. (At one point I was doing it differently.)

worker/uniter/runner/debug/server_test.go
```diff
@@ -236,10 +237,108 @@ func (s *DebugHooksServerSuite) TestRunHook(c *gc.C) {
...
+        # echo "running: ${@: -1}" >&2
+        # cat ${@: -1} >&2
+	    exec "${@: -1}"
+        ;;
+esac
+exit 1`), 0777)
+	c.Assert(err, jc.ErrorIsNil)
+}
+
+// fakeJujuLog installes a script that echos its arguments to stderr,
```
achilleasa, line 337: s/installes/installs (trips linter)

jameinel:
$$merge$$

PR #11528, opened 2020-05-04
vsphere: Enable using datastores in folders

babbageclunk:
### Checklist

----

## Description of change
Update the vsphereclient Datastores method to return datastores in folders under the root folder. Without this we will reject the creation saying there's no such datastore. We don't keep track of the inventory path of the datastore - when the name of the datastore is used for creating a new vm it's looked up on the compute resource, so the fully-qualified inventory path isn't needed.

## QA steps

* Bootstrap a vsphere controller.
* Deploy an application using the name of a datastore that's nested under a folder.
```sh
juju deploy ~jameinel/ubuntu-lite --constraints "root-disk-source=datastore-name"
``` 
The deployment succeeds and the machine is created on the specified datastore.

## Documentation changes

None

## Bug reference

https://bugs.launchpad.net/juju/+bug/1874031


wallyworld:
$$merge$$

PR #11529, opened 2020-05-04
Restore deleted CheckCredentialsModels method

wallyworld:
## Description of change

#11512 removed an unused by Juju method - CheckCredentialsModels - from the V7 Cloud facade. It was retained on V6 facade.
But JAAS needs this API so keep it on the V7 facade also.

As a drive by, delete unused interface definitions from the cloud facade.

## QA steps

Method unused by Juju, just run existng tests.

### DIFF
--- a/apiserver/facades/client/cloud/cloud.go
+++ b/apiserver/facades/client/cloud/cloud.go
@@ -28,6 +28,113 @@ import (
 
 var logger = loggo.GetLogger("juju.apiserver.cloud")
 
+// CloudV7 defines the methods on the cloud API facade, version 7.
+type CloudV7 interface {
+	AddCloud(cloudArgs params.AddCloudArgs) error
+	AddCredentials(args params.TaggedCredentials) (params.ErrorResults, error)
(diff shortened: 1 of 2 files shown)


apiserver/facades/client/cloud/cloud.go
```diff
@@ -28,112 +28,6 @@ import (
 
 var logger = loggo.GetLogger("juju.apiserver.cloud")
 
-// CloudV7 defines the methods on the cloud API facade, version 7.
-type CloudV7 interface {
```
howbazaar, line 32: Personally I actually like these interface definitions. It shows clearly the expectations of the interface. However I know that most places don't follow this pattern. Do you not find them useful?

wallyworld: No, I don't. They are bothersome to maintain, you need to know they are there for them to be of any use, otherwise you just end up adding methods to the facade and the interfaces become stale. And we only use them in this one place; we don't use them outside of this package to define a param type etc. So IMO they don't carry their weight. But I'll leave them in and we can revisit later.

wallyworld:
$$merge$$

PR #11530, opened 2020-05-05
Fix a logic error in the database upgrade worker.

howbazaar:
If one of the HA secondary controllers fails to process the upgrade steps
for whatever reason, like disk space issues, then the upgrade got stuck
unrecoverably.

This is because the worker assumed that the changes to the upgrade info
record in the database would be happening while it was watching. So if the
agent had been bounced, and it wasn't the primary, it would start watching
for changes to the doc, and time out after ten minutes. However the primary
and the other controller had already finished, and the upgrade info doc
had a status of "finishing".

The fix here is two-fold. First, once the watcher has been started,
reread the upgrade info document and check the status before waiting on
the changes from the watcher. If the status is "db-complete" or "finishing"
then the worker records success and is done. Otherwise it waits for the
watcher.

## QA steps

* Bootstrap a 2.7.6 controller using the snap, and enable-ha.
* Choose a machine that isn't the primary, and add a file `/var/lib/juju/wrench/database-upgrade` with a single line with the value `watch-upgrade`. This sets up the wrench.
```sh
```
* `juju status` should show the machine with the wrench as `down`
* Remove the wrench file and restart the machine agent.
* The upgrade should resume and recover. Agent will show as running.
* Could confim the database record `db.upgradeInfo.find().pretty()` should show one completed upgrade. The _id will be an object id and NOT "current", and the `controllersReady` and `controllersDone` should both contain all three machine IDs.

## Documentation changes

Internal implementation only.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1876652

### DIFF
--- a/worker/upgradedatabase/worker.go
+++ b/worker/upgradedatabase/worker.go
@@ -20,6 +20,7 @@ import (
 	"github.com/juju/juju/upgrades"
 	jujuversion "github.com/juju/juju/version"
 	"github.com/juju/juju/worker/gate"
+	"github.com/juju/juju/wrench"
 )
 
 // NewLock creates a gate.Lock to be used to synchronise workers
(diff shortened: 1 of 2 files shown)


howbazaar:
$$merge$$

howbazaar:
$$merge$$  Could not resolve host: github.com after all tests passed.

wallyworld:
!!build!!

howbazaar:
Merged as admin because the merge job passed all tests, but failed due to the schema subtask not having reported in. Couldn't work out how to get github to retry it.

PR #11536, opened 2020-05-05
Merge 2.7 into develop

babbageclunk:
## Description of change

Merging recent changes in 2.7 into the develop branch:
* #11532 - this was accidentally landed against develop in #11528, this is merging the 2.7 change into develop after fixing it so that it doesn't muddy the waters for people doing subsequent merges.

No conflicts.


babbageclunk:
$$merge$$

PR #11538, opened 2020-05-06
Fix edge case in manual provisioning.

howbazaar:
If the machine that you are trying to manually provision has the
2.8/edge juju snap installed, it introduces a snap service called
"snap.juju.fetch-oci". The code was just checking for "juju" in the
returned service list.

We now look for "jujud-machine" as a service name as any machine
that we sould look to provision over SSH would have a machine agent
set up if it was provisioned.

## QA steps

```sh
lxc launch ubuntu:18.04 to-provision
Add you ssh pub key to machine's .ssh/authorized_keys
Also inside the container, 'sudo snap install juju --classic --channel 2.8/edge'
Back outside the container, run:
juju add-machine ssh:<ip-addr>
Where ip-addr is the ip address of the to-provision container
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1876959


howbazaar:
$$merge$$

PR #11541, opened 2020-05-06
[dev] Ensure that reboot requests are queued correctly

achilleasa:
## Description of change

A commit which was part of the 2.8 refactoring work (https://github.com/juju/juju/pull/11204/commits/5fe18b0b3d853545c150ed6f24aed770f874780c) messed up the implementation of the `handleReboot` method. While the method would return the correct error (`ErrReboot` or `ErrRequeAndReboot`), the actual reboot request was never queued due to an early return logic bug.

This PR makes everything work again and includes additional tests to prevent future regressions.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1877065

### DIFF
--- a/worker/uniter/runner/context/context.go
+++ b/worker/uniter/runner/context/context.go
@@ -130,6 +130,7 @@ type HookUnit interface {
 	OpenPorts(protocol string, fromPort, toPort int) error
 	RequestReboot() error
 	SetUnitStatus(unitStatus status.Status, info string, data map[string]interface{}) error
+	SetAgentStatus(agentStatus status.Status, info string, data map[string]interface{}) error
 	State() (params.UnitStateResult, error)
 	Tag() names.UnitTag
 	UnitStatus() (params.StatusResult, error)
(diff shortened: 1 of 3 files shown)


worker/uniter/runner/context/context.go
```diff
@@ -1023,14 +1035,14 @@ func (ctx *HookContext) handleReboot(ctxErr error) error {
...
-		return ErrRequeueAndReboot
+		ctxErr = ErrRequeueAndReboot
 	}
 
-	// Do a best-effort attempt to set the unit status; we don't care if it
-	// fails as we will request a reboot anyway.
-	if err := ctx.unit.SetUnitStatus(status.Rebooting, "", nil); err != nil {
+	// Do a best-effort attempt to set the unit agent status; we don't care
+	// if it fails as we will request a reboot anyway.
+	if err := ctx.unit.SetAgentStatus(status.Rebooting, "", nil); err != nil {
```
achilleasa, line 1045: Original code (before refactoring) always failed (logged but otherwise ignored the error) trying to set the status because `status.Rebooting` is not a valid workload status.

achilleasa:
$$merge$$

PR #11543, opened 2020-05-07
Disable the legacy-leases feature flag

babbageclunk:
### Checklist

----

## Description of change

The raft-based lease system has been in and is now stable enough to remove the legacy lease implementation. The first phase of this is to remove the legacy-leases flag and disable the code-paths that check it. Those code paths have been marked with `if false` for now to minimise structural code changes. A follow-up after rc1 is released will remove them.

## QA steps

```sh
# Bootstrap with the legacy-leases feature flag turned on.
juju bootstrap lxd f --config "features=[legacy-leases]" --build-agent
# Deploy an application with three units.
juju deploy ~jameinel/ubuntu-lite -n 3
# Remove the leader and see that the leadership is claimed by one of the others.
juju remove-unit ubuntu-lite/0
# Check that the engine report shows that the global-clock-updater isn't running (since it's gated by the legacy-leases-flag which isn't set).
juju run -m controller --machine 0 juju_engine_report | grep global-clock-updater -A 5
```
In the database check that the leases collection is empty and the leaseholders collection (filled by the raft lease store) is populated:
```javascript
db.leases.find()
db.leaseholders.find()
```

## Documentation changes
We should remove legacy-leases from the feature flags documentation.

## Bug reference
None

### DIFF
--- a/apiserver/root.go
+++ b/apiserver/root.go
@@ -24,7 +24,6 @@ import (
 	"github.com/juju/juju/core/lease"
 	"github.com/juju/juju/core/multiwatcher"
 	"github.com/juju/juju/core/permission"
-	"github.com/juju/juju/feature"
 	"github.com/juju/juju/rpc"
 	"github.com/juju/juju/state"
 )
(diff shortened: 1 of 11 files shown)


apiserver/root.go
```diff
@@ -499,7 +498,8 @@ func (ctx *facadeContext) ID() string {
 // - only a claimer for the current model can be obtained with legacy
 // leases.
 func (ctx *facadeContext) LeadershipClaimer(modelUUID string) (leadership.Claimer, error) {
-	if ctx.r.shared.featureEnabled(feature.LegacyLeases) {
+	// Never use LegacyLeases
```
wallyworld, line 501: can we make these a TODO(legacyleases)

babbageclunk:
$$merge$$

PR #11558, opened 2020-05-11
Mark 137 error returned from exec API special and treat it as a retri…

ycliuhw:
…able error;

### Checklist

----

## Description of change

## QA steps

```sh
QA steps here
```

## Documentation changes

## Bug reference

### DIFF
--- a/caas/kubernetes/provider/exec/errors.go
+++ b/caas/kubernetes/provider/exec/errors.go
@@ -82,7 +82,9 @@ func IsExecRetryableError(err error) bool {
 
 func handleExecRetryableError(err error) error {
 	if wrench.IsActive("exec", "137") {
-		return newexecRetryableError(errors.New("fake 137"))
+		fakeErr := errors.New("fake 137")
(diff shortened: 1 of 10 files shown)


caas/kubernetes/provider/exec/errors.go
```diff
@@ -54,3 +56,41 @@ func handleContainerNotFoundError(err error) error {
 	}
 	return containerNotRunningError(match[1])
 }
+
+type execRetriableError struct {
```
hpidcock, line 60: Retryable is the more correct spelling. Both are correct but in the context of software Retryable is more correct. We already use "Retryable" in juju not Retriable.

ycliuhw: no worries, change it to `Retryable` lol

ycliuhw:
we decided to retry in the initialization function rather than retry the whole remote init operation.
And here is the final fix https://github.com/juju/juju/pull/11566

So close this one,

PR #11562, opened 2020-05-12
Add support for serving the new dashboard

wallyworld:
## Description of change

Add support for serving the new dashboard GUI.
The old GUI endpoint is retained as well for upgraded controllers.

The juju dashboard CLI will prefer to look for the new /dashboard endpoint but will fall back to using the /gui endpoint for older controllers or older controllers that have been upgraded.

Support is added for the new metadata "min-juju-version" attribute. This allows newer dashboards to be released that might not be compatible with older Juju versions.

Also add a new model config attribute to allow (when bootstrapping) a devel version of the dashboard to be used. Previously Juju was hard coded to only look for release metadata.
The upgrade-gui command has also been changed to support --gui-stream

The new dashboard tarball changes how content is packed. There's no nested folders and there's a separate version file.

## QA steps

Check that a dashboard can be set up manually via JUJU_GUI.

`export JUJU_GUI=jujugui-2.16.0.tar.bz2`

then bootstrap and run `juju dashboard`
It should print out the dashboard URL and credential and going to that in a browser should server the new dashboard.

Also, try using the published beta dashboard

`juju bootstrap --config gui-stream=devel lxd`

To test older GUI, bootstrap a 2.7 controller.
Running `juju dashboard` should print the older GUI URL.

Upgrade the controller to 2.8
Running `juju dashboard` should print the older GUI URL.
Also then check that a browser can use the older GUI from the upgraded controller

Now upgrade to the new dashboard
`juju upgrade-dashboard jujugui-2.16.0.tar.bz2`

Run `juju dashboard` and it will print the new dashboard URL and test in a browser.


cmd/juju/gui/upgradegui.go
```diff
@@ -258,35 +272,7 @@ func findMetadataVersion(allMeta []*gui.Metadata, vers version.Number) (*gui.Met
 			return metadata, nil
 		}
 	}
-	return nil, errors.NotFoundf("Juju GUI release version %s", vers)
-}
-
-// archiveVersion retrieves the GUI version from the juju-gui-* directory
```
wallyworld, line 264: moved to a public method as it is called form a couple of places

environs/gui/dashboard.go
```diff
@@ -0,0 +1,68 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/juju/errors"
+	"github.com/juju/version"
+)
+
+// DashboardArchiveVersion retrieves the Dashboard version
+// from the given tar.bz2 archive reader.
+func DashboardArchiveVersion(r io.Reader) (version.Number, error) {
```
wallyworld, line 21: moved from elsewhere so this code can be shared, it was copied in a couple of places

wallyworld:
[dashboard support](https://trello.com/c/wMlMZFB7/1119-dashboard-support)

hatched:
QA OK!

apiserver/gui.go
```diff
@@ -216,10 +277,17 @@ func uncompressGUI(r io.Reader, sourceDir, targetDir string) error {
 		if err != nil {
 			return errors.Annotate(err, "cannot parse archive")
 		}
-		if hdr.Name != sourceDir && !strings.HasPrefix(hdr.Name, sourceDir+"/") {
+		hName := hdr.Name
+		if sourceDir != "." && strings.HasPrefix(hName, "./") {
+			hName = hName[2:]
+		}
+		logger.Criticalf("HNAME: %v", hName)
```
howbazaar, line 284: This needs to be removed or reduced to Tracef

apiserver/gui.go
```diff
@@ -216,10 +277,17 @@ func uncompressGUI(r io.Reader, sourceDir, targetDir string) error {
...
+		if hName != sourceDir && !strings.HasPrefix(hName, sourceDir+"/") {
+			logger.Criticalf("skipping unknown dashboard file %q", hdr.Name)
```
howbazaar, line 286: This should be reduced in severity too.

apiserver/gui.go
```diff
@@ -216,10 +277,17 @@ func uncompressGUI(r io.Reader, sourceDir, targetDir string) error {
...
 			continue
 		}
 		path := filepath.Join(tempDir, hdr.Name)
+		logger.Criticalf("writing file %q", path)
```
howbazaar, line 290: And another to Tracef

wallyworld:
$$merge$$

PR #11564, opened 2020-05-13
Prefer netplan over ENI when both exist.

howbazaar:
The OpenStack team encountered a problem during the dist-upgrade process. Upgrading from xenial to bionic all worked in general, but has new netplan but also as if up/down installed.

When trying to create a container on a bionic instance that also had if up/down installed, juju made the wrong decision. The intent here is to prefer netplan over the old ENI approach if both netplan and older tools are installed.

## QA steps

TBH not sure how to test right now, tried on AWS but didn't make much progress.

```sh
# ???
```

### DIFF
--- a/container/broker/instance_broker.go
+++ b/container/broker/instance_broker.go
@@ -121,11 +121,13 @@ func prepareHost(config Config) PrepareHostFunc {
 	}
 }
 
+// defaultBridger will prefer to use netplan if there is an /etc/netplan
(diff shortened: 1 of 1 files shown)


manadart:
$$merge$$

manadart:
[Investigate O7k Netplan/ENI Issue](https://trello.com/c/uLbBkOdb/2581-investigate-o7k-netplan-eni-issue)

PR #11565, opened 2020-05-13
Add support for Azure focal images

wallyworld:
## Description of change

The way eoan and focal and later images have been tagged in Azure has changed.
This adds support for finding and using those images.

## QA steps

juju bootstrap --bootstrap-series=focal azure
juju deploy ubuntu --series=eoan --force
juju deploy ubuntu --series=bionic u2

### DIFF
--- a/provider/azure/internal/imageutils/images.go
+++ b/provider/azure/internal/imageutils/images.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
 	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	"github.com/juju/os"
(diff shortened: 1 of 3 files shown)


provider/azure/internal/imageutils/images.go
```diff
@@ -112,26 +112,41 @@ func SeriesImage(
...
+	if err != nil {
+		return "", "", errors.Trace(err)
+	}
+
+	oldSeries := set.NewStrings("trusty", "xenial", "bionic", "cosmic", "disco")
+	if oldSeries.Contains(series) {
+		return ubuntuOffering, seriesVersion, nil
+	}
+	seriesVersion = strings.ReplaceAll(seriesVersion, ".", "_")
+	return fmt.Sprintf("0001-com-ubuntu-server-%s", series), seriesVersion, nil
```
howbazaar, line 126: I got the distinct impression from the email that the 0001 won't always be that value, and it will change as they update things. Does the ListSkus handle partial values?

wallyworld: That 0001 doesn't come from ListSKUs(). The list api returns tags like "20_04-LTS" or "19_10" or "18.04-daily" etc.  That is one parameter used to select an image. The other is the offer string which is what the "0001-com-ubuntu-..." is. We need both the offer string and the SKU.

wallyworld:
[azure focal support](https://trello.com/c/VDpFbM7K/1121-azure-focal-support)

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #11567, opened 2020-05-13
Fix query tracker tests.

howbazaar:
All of the tests that were using the query tracker were subject to races
in other goroutines that may also be hitting the database. Since the
query tracker was loaded globally onto the state object the fix for
this is to initialize the query tracker with the method that we are
interested in tracking.

For the status tests, this is "client.(*Client).FullStatus" as it needs
to match the string that is found in the traceback.

Prior to this fix using stress-ng would always cause the test to fail,
with the fix I'm unable to cause a race failure. Also the wait for model
idle methods are now not needed as that was an attempt to deal with
the above problem of other goroutines hitting the database. Now that we
are filtering them out, we don't need to wait any more.

## QA steps

```sh
# Run the tests with forced verbose logging to ensure we are getting counts:
go test -check.vv -check.f TestFullStatusMachineScaling 
# observe that we have 19 queries initially

# Run the tests with stress-race in one terminal:
stress-race  -check.f TestFullStatusMachineScaling
# Run stress-ng in another
stress-ng --cpu 8 --io 4 --vm 2 --vm-bytes 128M --fork 4 --timeout 30s
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1878220

### DIFF
--- a/state/querytracker.go
+++ b/state/querytracker.go
@@ -5,6 +5,7 @@ package state
 
 import (
 	"runtime/debug"
+	"strings"
 	"sync"
 )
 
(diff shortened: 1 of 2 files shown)


state/querytracker.go
```diff
@@ -93,10 +95,13 @@ func (q *queryTracker) TrackRead(collectionName string, query interface{}) {
...
 	defer q.mu.Unlock()
 
-	q.queries = append(q.queries, QueryDetails{
-		Type:           "read",
-		CollectionName: collectionName,
-		Query:          query,
-		Traceback:      string(debug.Stack()),
-	})
+	traceback := string(debug.Stack())
+	if strings.Contains(traceback, q.method) {
```
wallyworld, line 99: Just say the trackback is for FullStatusFoo, it will match I think.
Can we add a delimiter to the q.method string to tighten the match.

howbazaar:
$$merge$$

howbazaar:

PR #11571, opened 2020-05-13
Prefer netplan over ENI when both exist.

howbazaar:
The OpenStack team encountered a problem during the dist-upgrade process. Upgrading from xenial to bionic all worked in general, but has new netplan but also as if up/down installed.

When trying to create a container on a bionic instance that also had if up/down installed, juju made the wrong decision. The intent here is to prefer netplan over the old ENI approach if both netplan and older tools are installed.

Backport of #11564.


howbazaar:
$$merge$$

PR #11573, opened 2020-05-14
Finish legacy-leases removal

babbageclunk:
### Checklist

----

## Description of change
This PR makes a number of changes to complete the removal of legacy leases that was begun in #11543. It's probably easiest to review in separate commits.

### Fix the skipped machine agent tests

These were in `MachineLegacyLeasesSuite` because they depended on leases working and there were a few prerequisites preventing the raft workers from operating. They were fixed by bootstrapping the (one machine) raft cluster so that it knew the machine had a vote, and publishing a test address on the central hub so that the raft-transport worker had the current address (which was preventing the raft worker from starting). The latter would normally be done by the peergrouper, but that doesn't run properly in the tests because the replicaset isn't configured, and I couldn't work out how to do it in a way that would stick.

The tests have all been moved back to `MachineSuite` and `MachineLegacyLeasesSuite` is gone.

### Remove the `if false` blocks from #11543 

Including removing anything that's not used anymore because of that. In particular, the state leadership and singular lease managers are gone, which only leaves the `txnlog` worker. The `state/lease` package is still kept around because it's used from upgrades (for migrating leases to a new time format).

### Remove the state input from globalclockupdater

Now that we only run one of them in an agent it doesn't need to be able to switch between taking a state or a lease manager as input.

### Remove the featureflag worker

It's not specific to legacy leases, but the legacy-leases-flag was the only use of it in the codebase. Hopefully when it's needed next time someone remembers "oh yeah, we used to have something like that" and resurrects it rather than coding it again (although that's better than having to maintain it just in case). 

## QA steps

Same as for #11543 :
```sh
juju bootstrap lxd f --build-agent
# Deploy an application with three units.
juju deploy ~jameinel/ubuntu-lite -n 3
# Remove the leader and see that the leadership is claimed by one of the others.
juju remove-unit ubuntu-lite/0
```
In the database check that the leases collection is empty and the leaseholders collection (filled by the raft lease store) is populated:
```javascript
db.leases.find()
db.leaseholders.find()
```
## Documentation changes

None

## Bug reference

None


babbageclunk:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #11584, opened 2020-05-15
Provide clock and logger to upgrader worker.

howbazaar:
Some upcoming work was going to need to tweak some upgrader
behaviour. It seemed reasonable to bring the worker more up
to standard before that work was started.

Unit tests cover all changes, nothing user facing at all.


howbazaar:
$$merge$$

PR #11585, opened 2020-05-15
Don't write errors when pubsub websocket disconnects.

howbazaar:
Having the other side of the websocket go away isn't an error.
We now check to see if the error is a close error type, and if
it is, just log at trace. Only log an error if there is some other
error type.

Also had the pubsub handler use a more detailed logger for writing
errors so they can be filtered more easily.

## QA steps

```sh
# bootstrap and enable-ha
juju switch controller
juju model-config logging-config=juju=WARN:juju.apiserver.pubsub=TRACE
# in one window
juju debug-log
# in another window
juju ssh 2
sudo service jujud-machine-2 stop
```
Observe in the log that we get a "websocket closed" log line, not an error.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1759372

### DIFF
--- a/apiserver/pubsub.go
+++ b/apiserver/pubsub.go
@@ -9,6 +9,7 @@ import (
 
 	gorillaws "github.com/gorilla/websocket"
 	"github.com/juju/errors"
+	"github.com/juju/loggo"
 	"github.com/juju/utils/featureflag"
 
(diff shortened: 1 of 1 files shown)


howbazaar:
$$merge$$

PR #11591, opened 2020-05-18
Machine agent tests: increase timeout in TestJobManageModelRunsMinUnitsWorker

babbageclunk:
## Description of change

Now that the agent that gets run in the machine agent tests uses raft leases, the raft workers sometimes take too long to get started and ready, which made `TestJobManageModelRunsMinUnitsWorker` sometimes fail. Use the longer timeout (double the LongWait) that other tests in the suite already use.

## QA steps

Running the test under stress doesn't fail after a lot of runs.

## Documentation changes

None

## Bug reference

None


babbageclunk:
$$merge$$

PR #11592, opened 2020-05-18
Expose the engine worker restart time as a package variable.

howbazaar:
This allows tests to change the restart time so older wall clock
type tests don't have to wait three seconds for workers to restart.

In particular some of the MachineSuite tests are impacted by this as
the api caller worker is restarted when the addresses change. Which is
signalled by another worker.

Without this patch, the TestManageModelRunsCleaner test ran in approx
six seconds, ten seconds, or timed out about one in 12. With this patch
the test is a consistent six seconds, which is still too long, but
at least consistent.

## QA steps

```sh
cd cmd/jujud/agent
stress -check.v -check.f TestManageModelRunsCleaner
```

## Documentation changes

Test change only.

### DIFF
--- a/cmd/jujud/agent/agent.go
+++ b/cmd/jujud/agent/agent.go
@@ -24,6 +24,11 @@ import (
 	"github.com/juju/juju/cmd/jujud/util"
 )
 
+// EngineErrorDelay is the amount of time the dependency engine waits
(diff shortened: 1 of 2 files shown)


howbazaar:
$$merge$$

babbageclunk:
$$merge$$

PR #11609, opened 2020-05-20
Add an acceptance test for CharmRevisionUpdater instead of flakey "unit" test

wallyworld:
## Description of change

The (machine) agent tests which look at the behaviour of certain workers are inherently racey by design. The TestManageModelRunsCharmRevisionUpdater test has been particularly bad of late on merge bot runs. And they are not unit tests. And we have existing tests to check that the worker is wired up in the manifold, and for the specific worker itself.

We want an acceptance test, but the worker only runs every 24 hours (hard coded).
This PR adds an option to thread testing config through to the agent.conf values bucket - until now this needed to be hand edited and the agent restarted to add things like bespoke logging config etc.
The maniold set up for the charm revision updater worker will use the testing value if present, otherwise falls back to the hard coded value.

As a drive by, clean up some unsed consts. And also prefer the use of the LxdBridge rather than the LxcBridge option.

Also add a BUILD_AGENT option to the bash testing framework.

## QA steps

main.sh agents


cloudconfig/podcfg/podcfg_test.go
```diff
@@ -98,3 +98,25 @@ func (*podcfgSuite) TestBootstrapConstraints(c *gc.C) {
...
+		constraints.Value{},
+	)
+	podcfg.FinishControllerPodConfig(
+		podConfig,
+		cfg,
+		map[string]string{"foo": "bar"},
+	)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(podConfig.AgentEnvironment, jc.DeepEquals, map[string]string{
+		"PROVIDER_TYPE": "kubernetes",
```
howbazaar, line 119: I'm not sure we actually use the PROVIDER_TYPE anywhere any more, and it could be removed I think. Worth checking though, but it used to provide custom networking stuff for maas, but shouldn't any more. Not for this branch, but would be a good cleanup.

cmd/jujud/agent/machine.go
```diff
@@ -1106,14 +1106,24 @@ func (a *MachineAgent) startModelWorkers(cfg modelworkermanager.NewModelConfig)
 		logger.Errorf("unable to configure db logging for model: %v", err)
 	}
 
+	charmRevisionUpdateInterval := 24 * time.Hour
```
howbazaar, line 1109: As we get more of these, it'll make the function unwieldy, perhaps extract into a function?

container/broker/lxd-broker.go
```diff
@@ -85,7 +85,10 @@ func (broker *lxdBroker) StartInstance(ctx context.ProviderCallContext, args env
 	// prepareOrGetContainerInterfaceInfo should always return a value. The
 	// test suite currently doesn't think so, and I'm hesitant to munge it too
 	// much.
-	bridgeDevice := broker.agentConfig.Value(agent.LxcBridge)
+	bridgeDevice := broker.agentConfig.Value(agent.LxdBridge)
```
howbazaar, line 88: Is this change going to break anything? Who sets the LxdBridge?

provider/common/bootstrap.go
```diff
@@ -156,7 +157,7 @@ func BootstrapInstance(
 			if icfg.AgentEnvironment == nil {
 				icfg.AgentEnvironment = make(map[string]string)
 			}
-			icfg.AgentEnvironment[agent.LxcBridge] = args.ContainerBridgeName
+			icfg.AgentEnvironment[agent.LxdBridge] = args.ContainerBridgeName
```
howbazaar, line 160: Ah... ok. Is this actually used anywhere?

tests/suites/agents/task.sh
```diff
@@ -0,0 +1,20 @@
...
+    fi
+
+    set_verbosity
+
+    echo "==> Checking for dependencies"
+    check_dependencies juju
+
+    file="${TEST_DIR}/test-agents.txt"
+
+    JUJU_AGENT_TESTING_OPTIONS=CHARM_REVISION_UPDATE_INTERVAL=15s \
```
howbazaar, line 14: How do we deal with the situation of running tests against an existing controller? It isn't going to work here.

wallyworld: No it's not. It was designed to work when bootstrapping a new controller. There's not really a nice way to weave this into an existing controller since it needs an agent restart etc. We could do a big, bespoke ssh script. But the CI tests on Jenkins all use a new controller IIANM.

provider/common/bootstrap.go
wallyworld: Used in lxdBroker StartInstance()

```
	// Something to fallback to if there are no devices given in args.NetworkInfo
	// TODO(jam): 2017-02-07, this feels like something that should never need
	// to be invoked, because either StartInstance or
	// prepareOrGetContainerInterfaceInfo should always return a value. The
	// test suite currently doesn't think so, and I'm hesitant to munge it too
	// much.
	bridgeDevice := broker.agentConfig.Value(agent.LxdBridge)
	if bridgeDevice == "" {
		bridgeDevice = broker.agentConfig.Value(agent.LxcBridge)
	}
	if bridgeDevice == "" {
		bridgeDevice = network.DefaultLXDBridge
	}
```

cloudconfig/podcfg/podcfg_test.go
wallyworld: It's currently used for agent start up to detect if it's a k8s model or not.
The original use is not there anymore.
With a fair bit of refactoring, we can migrate away from it as controller/machine agent separation is done etc.

cmd/jujud/agent/machine.go
wallyworld: yup

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #11613, opened 2020-05-21
Ensure we wait for the message to be sent.

SimonRichardson:
## Description of change

Addresses failure in `MachineWithCharmsSuite.TestManageModelRunsCharmRevisionUpdater`.

The way the pubsub works is to send the message asynchronously and it's
up to the publisher to wait for their message to be accepted and
consumed.

A race between the message being sent, consumed, and reading that message
from the other side was at fault here. Waiting for the message should
prevent this data race (not a code race).

## QA steps

```sh
go test -v ./cmd/jujud/agent/ -check.v -check.f=MachineWithCharmsSuite.TestManageModelRunsCharmRevisionUpdater -count=100
```


SimonRichardson:
$$merge$$

PR #11615, opened 2020-05-21
Remove JujuConnSuite from some api tests

wallyworld:
## Description of change

A small step to removing JujuConnSuite from where it has no right to be.
The api/charmrevisionupdater and api/uniter tests are updated to use the api caller mock pattern instead of a full stack JujuConnSuite set of tests.
In the uniter tests, some were removed entirely as they were really just testing back end behaviour.

## QA steps

Test changes, run the unit tests. They are FAST now.


api/charmrevisionupdater/updater_test.go
```diff
@@ -4,57 +4,32 @@
...
+	"github.com/juju/juju/apiserver/params"
 )
 
-type versionUpdaterSuite struct {
-	jujutesting.JujuConnSuite
-	testing.CharmSuite
-
-	updater *charmrevisionupdater.State
-}
+type versionUpdaterSuite struct{}
```
howbazaar, line 14: it is best practice to use either testing.IsolationSuite or BaseSuite to capture logging

wallyworld: ah i did it for every other one, missed this one sorry

wallyworld:
$$merge$$

PR #11617, opened 2020-05-22
Thread a logging context through the agent initialization.

howbazaar:
This is more prepatory work for the merging of the machine and unit agents.
In the future some of the methods used will have a different logging context.
For now we just pass through the default logging context, which has the
same existing behaviour.

As a drive by there is a Report method added to a StringsWorker. This will
call the Report method of the handler if there is one.

## QA steps

Unit tests should continue to pass.

## Documentation changes

Only internal refactoring.


howbazaar:
optimistic $$merge$$

howbazaar:
$$merge$$

PR #11627, opened 2020-05-26
CMR relation remove

wallyworld:
## Description of change

There's a few issues for CMR relations. The main one is the workers in the consuming model are stopped as soon as the relation dead event arrives, but there could still be a unit scope change event to arrive (different watchers are used). So ensure that the watchers are only stopped when all units have left scope. This is done by adding unit count to the relation-unit change event. Also, when a relation is suspended, only stop the watchers if the relation is alive, otherwise we might miss departing events.

Reduce the cmr worker restart delay to 15 seconds, as 1 minute seems too long (we want to avoid spamming when using cross controller calls hence it's not the standard 3 seconds).

We were also marking the saas status as terminated if the offering side had removed the relation, but this should only be done if the offer (or app) is removed. So add a new watcher to provide that information. Extra unit tests in various places were added also.

There's 3 commits:
- state changes
- apiserver changes
- cmr worker changes

## QA steps

```sh
juju bootstrap lxd
juju add-model database
juju deploy percona-cluster mysql --config min-cluster-size=3 -n 3
juju offer mysql:shared-db
juju switch default
juju deploy vault -n 3
juju consume database.mysql rmysql
juju relate rmysql vault
```

wait for things to settle, then 
juju remove-relation vault rmysql
The relation should be removed both consuming side and offering side.

Remove the offer.
The saas status on the consuming side should go to "terminated".

## Bug reference

https://bugs.launchpad.net/bugs/1879645

### DIFF
--- a/apiserver/common/crossmodel/crossmodel.go
+++ b/apiserver/common/crossmodel/crossmodel.go
@@ -489,10 +489,7 @@ type offerGetter interface {
 // struct for a specified offer name.
 func GetOfferStatusChange(st offerGetter, offerUUID, offerName string) (*params.OfferStatusChange, error) {
 	offer, err := st.ApplicationOfferForUUID(offerUUID)
-	if err != nil && !errors.IsNotFound(err) {
-		return nil, errors.Trace(err)
-	}
-	if err != nil {
+	if errors.IsNotFound(err) {
 		return &params.OfferStatusChange{
 			OfferName: offerName,
 			Status: params.EntityStatus{
@@ -500,13 +497,12 @@ func GetOfferStatusChange(st offerGetter, offerUUID, offerName string) (*params.
 				Info:   "offer has been removed",
 			},
 		}, nil
+	} else if err != nil {
+		return nil, errors.Trace(err)
 	}
 	// TODO(wallyworld) - for now, offer status is just the application status
 	app, err := st.Application(offer.ApplicationName)
-	if err != nil && !errors.IsNotFound(err) {
-		return nil, errors.Trace(err)
-	}
-	if err != nil {
+	if errors.IsNotFound(err) {
 		return &params.OfferStatusChange{
 			OfferName: offerName,
 			Status: params.EntityStatus{
@@ -514,7 +510,10 @@ func GetOfferStatusChange(st offerGetter, offerUUID, offerName string) (*params.
 				Info:   "application has been removed",
 			},
 		}, nil
+	} else if err != nil {
+		return nil, errors.Trace(err)
 	}
+
 	status, err := app.Status()
 	if err != nil {
 		return nil, errors.Trace(err)
--- a/apiserver/common/crossmodel/crossmodel.go
+++ b/apiserver/common/crossmodel/crossmodel.go
@@ -332,12 +332,14 @@ func ExpandChange(
 		})
 	}
 
+	uc := relation.UnitCount()
 	result := params.RemoteRelationChangeEvent{
 		RelationToken:       relationToken,
 		ApplicationToken:    appToken,
 		ApplicationSettings: appSettings,
 		ChangedUnits:        unitChanges,
 		DepartedUnits:       departed,
+		UnitCount:           &uc,
 	}
 
 	return result, nil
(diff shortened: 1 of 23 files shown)


apiserver/common/crossmodel/crossmodel.go
```diff
@@ -485,22 +487,40 @@ type offerGetter interface {
...
+			Status: params.EntityStatus{
+				Status: status.Terminated,
+				Info:   "offer has been removed",
+			},
+		}, nil
+	}
 	// TODO(wallyworld) - for now, offer status is just the application status
 	app, err := st.Application(offer.ApplicationName)
-	if err != nil {
+	if err != nil && !errors.IsNotFound(err) {
```
howbazaar, line 506: Same here.

apiserver/common/crossmodel/crossmodel.go
```diff
@@ -332,12 +332,14 @@ func ExpandChange(
...
 	}
 
+	uc := relation.UnitCount()
 	result := params.RemoteRelationChangeEvent{
 		RelationToken:       relationToken,
 		ApplicationToken:    appToken,
 		ApplicationSettings: appSettings,
 		ChangedUnits:        unitChanges,
 		DepartedUnits:       departed,
+		UnitCount:           &uc,
```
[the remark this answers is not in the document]
howbazaar: Is the 2.8 facade version the same? Would this cause issues with a 2.7.7 model and a 2.8.0 controller? Or would perhaps it be fine because the cleanup is all controller side?

state/applicationoffers_test.go
```diff
@@ -794,3 +794,40 @@ func (s *applicationOffersSuite) TestWatchOfferStatus(c *gc.C) {
...
+	// Ensure that all the creation events have flowed through the system.
+	s.WaitForModelWatchersIdle(c, s.Model.UUID())
+
+	w := s.State.WatchOffer(offer.OfferName)
+
+	defer statetesting.AssertStop(c, w)
+	wc := statetesting.NewNotifyWatcherC(c, s.State, w)
+	// Initial event.
+	wc.AssertOneChange()
+	wc.AssertNoChange()
```
howbazaar, line 823: All the AssertOneChange calls have an implicit AssertNoChange in them as well, so you could remove the extras here.

worker/remoterelations/relationunitsworker.go
```diff
@@ -13,6 +14,13 @@ import (
 	"github.com/juju/juju/apiserver/params"
 )
 
+// RelationUnitChangeEvent encapsulates a remote relation event,
+// adding the tag of the relation which changed.
+type RelationUnitChangeEvent struct {
+	Tag names.RelationTag
+	params.RemoteRelationChangeEvent
```
howbazaar, line 21: It would be better if the api client extracted the params struct and returned what you actually cared about.

worker/remoterelations/remoteapplicationworker.go
```diff
@@ -244,21 +254,24 @@ func (w *remoteApplicationWorker) processRelationDying(key string, r *relation,
...
 	relation, ok := relations[key]
 	if !ok {
 		return nil
 	}
 
-	// For suspended relations on the consuming side
-	// we want to keep the remote lifecycle watcher
-	// so we know when the relation is resumed.
+	// Only stop the watchers for relation unit changes if relation is alive,
+	// as we need to always deal with units leaving scope etc if the relation is dying.
```
howbazaar, line 265: Thank you for the good comments here.

apiserver/common/crossmodel/crossmodel.go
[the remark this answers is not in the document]
wallyworld: This is a controller facade parameter - the api is only called by a worker in the same controller, so it has to be in sync by definition.
It is not called across controllers nor from agents to the controller, hence no need to up the facade version.
If libjuju is affected by this that's a bug we need to fix in libjuju.

worker/remoterelations/relationunitsworker.go
wallyworld: Could do but would prefer not to change too much outside of the current behaviour in a PR that is already quite involved.

apiserver/common/crossmodel/crossmodel.go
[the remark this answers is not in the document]
SimonRichardson: @howbazaar @wallyworld So both of you are right. It does impact libjuju, because any addition should be a facade bump, but @wallyworld is also right that this shouldn't be used by libjuju and should be treated as a bug.

In this circumstance we should be ok I believe to use `omitempty`.

I have another thought though. Juju itself has a package level separation of [Facades via packages](https://github.com/juju/juju/tree/develop/apiserver/facades) (ignoring the common Facade modules - sigh!), but this has never been lifted up into the code. The [`AllFacades`](https://github.com/juju/juju/blob/develop/apiserver/allfacades.go#L109) creates a `Registry`, which is a wrapper around `map[string]versions` (versions being the version of the facade and the factory method to create it`). I propose that we annotate the `Registry` either externally so you have many registries, or to the `Registry` itself. Adding an annotation about what type of Facade it is (`agent`, `controller`, or `client` and we could go more fine-grained here). 

From the Juju side you would still consume the `AllFacades` `Registry` in the same way,, so no difference there. The real difference is that for generating the schema for libjuju we would only generate all the `client` ones for example. This would cut down massively what we expose to the outside world. It would make libjuju smaller and lighter, the added benefit of it making it easier to understand the Juju API because of that factor. We don't expose APIs we don't want libjuju to interact with and it gives us more freedom with PRs like this.

wallyworld:
[cmr relation removal issues](https://trello.com/c/Urruko06/1123-cmr-relation-removal-issues)

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #11630, opened 2020-05-27
Migrating a model from an old controller with incomplete endpoint bin…

wallyworld:
## Description of change

There was a missing upgrade step when processing endpoint bindings coming from 2.6->2.7. The default binding was not added. This then confused the handling endpoint bindings when importing a model = a default space name was added where in some cases it should have been the id, depending on the import source.

This PR fixes the upgrade and also leaves in place extra logic to handle imports from broken models on controllers upgraded to 2.7.6 or earlier.

## QA steps

3 scenarios to check, all migrating a model to a 2.7 controller built off this PR
The model to be migrated has a single charm with at least one endpoint deployed

1. 2.6 controller
2. 2.6 controller upgraded to 2.7.6
3. 2.6 controller upgraded to 2.7 from this PR

In the latter case, you can also show-application prior to migrating to see that the default endpoint binding has been added when upgrading.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1880797

### DIFF
--- a/state/migration_import.go
+++ b/state/migration_import.go
@@ -921,8 +921,10 @@ func (i *importer) application(a description.Application) error {
 // of space names as the map values and can safely be passed to the NewBindings
 // c-tor.
 func (i *importer) parseBindings(bindingsMap map[string]string) (*Bindings, error) {
+	defaultMappingsAreIds := true
 	for epName, spNameOrID := range bindingsMap {
 		if spNameOrID == "" {
+			defaultMappingsAreIds = false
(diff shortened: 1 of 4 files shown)


wallyworld:
[2.6 migration bug](https://trello.com/c/bGH416wf/1127-26-migration-bug)

manadart:
Looks good. 

I ran the QA scenarios on AWS using a space definition and a unit of percona-cluster with "db" bound to the space.

wallyworld:
$$merge$$

PR #11632, opened 2020-05-27
Make vendor content addressable

SimonRichardson:
## Description of change

The following makes the vendor directory content addressable. It uses
the Gopkg.lock file for the sha and then downloads the content from s3
if it's available.

SO in order to update the vendor directory for CI tests, you need to
update your Gopkg.toml, which will cause the Gopkg.lock file to update.
Then run sha256sum on the Gopkg.lock and upload that to the s3 bucket,
ensuring to make it public.

## QA steps

```sh
rm -rf vendor
make dep
```

### DIFF
--- a/make_functions.sh
+++ b/make_functions.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+set -euf
+
+ensure_dep() {
+    if [ ! -d vendor ]; then
+        SHA=$(sha256sum Gopkg.lock | cut -d ' ' -f 1)
+        wget -O vendor.tar.gz "https://juju-dep.s3-eu-west-1.amazonaws.com/${SHA}.tar.gz"
+        tar -xvzf vendor.tar.gz
+        rm -rf vendor.tar.gz
+    else
+        echo "Skipping..."
+    fi
+}
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,7 @@ $(GOPATH)/bin/dep:
 
 # populate vendor/ from Gopkg.lock without updating it first (lock file is the single source of truth for machine).
 dep: $(GOPATH)/bin/dep
+	sh -c '. "${PROJECT_DIR}/make_functions.sh"; ensure_dep "$$@"'
 	$(GOPATH)/bin/dep ensure -vendor-only $(verbose)
 endif
 


SimonRichardson:
### **DON'T MEREGE**

Although this works magic for performance, we want to ensure that the tmpfs isn't also a problem for us when running CI.

SimonRichardson:
!!build!!

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

make_functions.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -euf
+
+ensure_dep() {
+    if [ ! -d vendor ]; then
+        SHA=$(sha256sum Gopkg.lock | cut -d ' ' -f 1)
```
achilleasa, line 6: Why not use s3cmd to check whether the $sha.tar exists and if it doesn't, do a `dep ensure` followed by an tar+upload? Should be fine even in case of a race occurring (last write wins but contents will always be the same) . 

Also, unless I'm mistaken S3 bucket contents have a limited lifespan (unless the admin explicitly overrides this) so unless we auto-generate/upload the files future builds are likely to eventually break needing manual intervention.

achilleasa: Actually, I was able to dig up my old [proposal](https://paste.ubuntu.com/p/KXWYgvxSRD/) for doing something like this as part of the CI setup script. Perhaps you might be able to repurpose some bits from there to make the above steps more resilient.

SimonRichardson: > Why not use s3cmd to check whether the $sha.tar exists

So that means I'd need to introduce `s3cmd` to make and I don't want to do that.

> do a dep ensure followed by an tar+upload?

It fails even attempting to do that, so we can't do that. You have to do it locally and then just download the contents.

> Also, unless I'm mistaken S3 bucket contents have a limited lifespan

It's actually the opposite. You have to explicitly set a lifespan when you create the bucket.

achilleasa: Ah, my bad; was probably thinking of another bucket which already had a policy assigned to it.

In any case, given that this change mostly benefits CI/build bots (what I mean it's fine to do a make dep from time to time when developing locally) you could make this part of the build scripts where s3cmd will be already configured for you...

On the other hand, as we probably wont be bumping deps in 2.7 frequently after 2.8 is out perhaps having someone on the team manually regenerate/upload the tar file when a dep change is made is probably fine...

SimonRichardson:
$$merge$$

achilleasa:
I have cherry-picked the commit from this PR into https://github.com/juju/juju/pull/11660 so we can bypass the dep ensure blockage on CI.

PR #11634, opened 2020-05-28
Ensure relation members in uniter data model are not nil

wallyworld:
## Description of change

The relation state struct maintained by the uniter agent is initialised with empty (not nil) maps for Members and ApplicationMembers. YAML serialisation / deserialisation however results in these becoming nil if they were empty when serialised. There's a subsequent validation check before the next hook invocation which fails as it checks for nil. This wasn't an issue previously when state was stored to disk and not the controller as reading state off disk would ensure the maps were empty not nil.

## QA steps

I tested with a modified build which added a wrench but not the patch.

Use the wrench to set up an error in the leader-elected hook for mariadb and create a relation to mediawiki. Remove mariadb once things appear settled and observe  the logs
unit-mariadb-2: 12:42:34 ERROR juju.worker.dependency "uniter" manifold worker returned unexpected error: preparing operation "run relation-broken (4; app: ) hook": inappropriate "relation-broken" for "": relation is broken and cannot be changed further

upgrade-controller --build-agent with the patch and see that the relation-broken hook is now run and there are no errors.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1881021

### DIFF
--- a/worker/uniter/relation/state.go
+++ b/worker/uniter/relation/state.go
@@ -137,23 +137,33 @@ func (s *State) YamlString() (string, error) {
 	return string(data), nil
 }
 
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
+func (s *State) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type StateCopy State
+	var sc StateCopy
+	err := unmarshal(&sc)
+	if err != nil {
+		return err
+	}
+	*s = State(sc)
+	if s.Members == nil {
+		s.Members = map[string]int64{}
+	}
+	if s.ApplicationMembers == nil {
+		s.ApplicationMembers = map[string]int64{}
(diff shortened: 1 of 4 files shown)


worker/uniter/relation/state.go
```diff
@@ -137,23 +137,33 @@ func (s *State) YamlString() (string, error) {
...
+	}
+	return nil
+}
+
 // copy returns an independent copy of the state.
 func (s *State) copy() *State {
-	stCopy := &State{
-		RelationId:     s.RelationId,
-		ChangedPending: s.ChangedPending,
+	stCopy := NewState(s.RelationId)
```
howbazaar, line 160: Much cleaner and nicer. Thanks.

wallyworld:
$$merge$$

PR #11635, opened 2020-05-28
Clean up more package level loggers.

howbazaar:
Yet more prepatory work for the merging of agents. This time the target was just dealing with worker/uniter/runner package. However this quickly leaked into the worker/uniter/runner/context package, which then impacted worker/meterstatus and worker/metrics/collect.

Using a new method on the loggo.Logger to get access to the Root logger, we can then use the Child method to get a completely unassociated logger but still have the same logging context. This is necessary for the logging of unit hook output and ensuring it goes into the right log file.

As a drive by I updated some of the validation tests to be more how we do things now.

## QA steps

Unit tests, and deploy something to show all things are hooked up.

## Documentation changes

Everything is internal, nothing to see here.


howbazaar:
$$merge$$

wallyworld:
$$merge$$

PR #11644, opened 2020-05-29
Removing the last of the uniter package loggers.

howbazaar:
This should be the last PR for uniter package level loggers.

All are replaced with an empty interface to stop people accidentally adding them back in.

The only exception here is the one for jujuc, which is fine. A comment was added to say it's fine.

Some minor test refactoring to make things easier, but on the whole very mechanical.


howbazaar:
$$merge$$

PR #11645, opened 2020-05-29
[2.8] Improve logging for failing to apply commands using raft

achilleasa:
## Description of change

This PR attempts to improve the quality of the raft-related error messages that juju emits. This is achieved by implementing `fmt.Stringer` on the raft `Command` type used by juju and including it (with additional context) in generated error messages.

The attached bug seems to be caused by lack of disk space. Unfortunately, such an error would be generated when the `raft-forwarder` worker calls `raft.Apply` and while it would be returned back to the caller (e.g. the `core/raftlease` package), it would be logged at TRACE level which is probably the reason why we don't see it in the attached log output! The PR changes the TRACE call to an ERROR call which should let us see something like this in the logs (assuming there is still enough room to write them):

```
 write /var/lib/juju/raft/logs: no space left on device
```

## QA steps

No functional changes to existing behavior.

## Bug reference

These changes should help us collect more information for diagnosing scenarios like: https://bugs.launchpad.net/juju/+bug/1879992

### DIFF
--- a/core/raftlease/fsm.go
+++ b/core/raftlease/fsm.go
@@ -4,6 +4,7 @@
 package raftlease
 
 import (
+	"fmt"
 	"io"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


achilleasa:
!!build!! failed test in api/uniter (but which one is not shown in the logs; also tests pass locally)

achilleasa:
$$merge$$

PR #11646, opened 2020-05-29
[2.8] Add controller option to avoid fsync between writes to the raft log

achilleasa:
## Description of change

Currently, each time juju appends an entry to the raft log (backed by a bolt-db instance), an fsync syscall is performed to ensure that the data gets flushed to disk. However, in some cases it might be beneficial to disable fsyncs if the operator is interested in trading performance for data consistency. Given that our log primarily tracks leadership leases and those are extended frequently, tolerating a potential loss of data (which will be corrected at the next lease renewal) for 
a boost in performance might be an acceptable compromise.

To this end, this PR introduces a new controller config option called `non-synced-writes-to-raft-log` (defaults to `false`) which can be toggled by operators either at bootstrap or while the controller is up and running. In the latter case, the agents will automatically restart once the setting changes.

NOTES:
- To avoid fetching the controller settings from within the worker, the value is passed as an agent config value; this also allows us to hook to the agentconfigupdater and get hot reloading for free.
- The _name_ of the option was intentionally chosen so that its off/false value indicates that we **want** fsyncs by default. I find that this plays nicely with omitempty and prevents juju from accidentally turning fsync off due to a missing value in the agent conf (this is probably not an issue since they get overwritten when upgrading but it's good to be a bit extra paranoid when data integrity is at stake).

## QA steps

```consoel
$ juju bootstrap lxd test
$ juju controller-config non-synced-writes-to-raft-log
True

$ juju controller-config non-synced-writes-to-raft-log=true

# Wait for the controllers to restart and run
$ juju debug-log -m controller --level WARN | grep non-synced
machine-0: 18:02:30 WARNING juju.worker.raft disabling fsync calls between raft log writes as instructed by the "non-synced-writes-to-raft-log option"

# Turn it back off and watch the agents restart again
$ juju controller-config non-synced-writes-to-raft-log=false
```

## Documentation changes

We should document the new option.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -376,15 +376,16 @@ var attributeParams = agent.AgentConfigParams{
 	Paths: agent.Paths{
 		DataDir: "/data/dir",
 	},
-	Tag:               names.NewMachineTag("1"),
-	UpgradedToVersion: jujuversion.Current,
-	Password:          "sekrit",
-	CACert:            "ca cert",
-	APIAddresses:      []string{"localhost:1235"},
-	Nonce:             "a nonce",
-	Controller:        testing.ControllerTag,
-	Model:             testing.ModelTag,
-	JujuDBSnapChannel: "4.0/stable",
+	Tag:                      names.NewMachineTag("1"),
+	UpgradedToVersion:        jujuversion.Current,
+	Password:                 "sekrit",
+	CACert:                   "ca cert",
+	APIAddresses:             []string{"localhost:1235"},
+	Nonce:                    "a nonce",
+	Controller:               testing.ControllerTag,
+	Model:                    testing.ModelTag,
+	JujuDBSnapChannel:        "4.0/stable",
+	NonSyncedWritesToRaftLog: false,
 }
 
 func (*suite) TestAttributes(c *gc.C) {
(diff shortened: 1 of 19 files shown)


agent/agent_test.go
```diff
@@ -376,15 +376,16 @@ var attributeParams = agent.AgentConfigParams{
...
+	Tag:                      names.NewMachineTag("1"),
+	UpgradedToVersion:        jujuversion.Current,
+	Password:                 "sekrit",
+	CACert:                   "ca cert",
+	APIAddresses:             []string{"localhost:1235"},
+	Nonce:                    "a nonce",
+	Controller:               testing.ControllerTag,
+	Model:                    testing.ModelTag,
+	JujuDBSnapChannel:        "4.0/stable",
+	NonSyncedWritesToRaftLog: false,
```
wallyworld, line 388: since this defaults to false, better to test with true

achilleasa: It was set up as such to mimic the expected default value for a bootstrapped controller. Note that the test at https://github.com/juju/juju/pull/11646/commits/32c8614d4522091e94c96afc5f31c326211f98e1#diff-336a19a3907c67c768036d0fae6aa384R668-R678 checks that it works as expected in both states. If you think that it's better to swap this around let me know and I will amend the relevant commit on Monday.

achilleasa:
$$merge$$

achilleasa:
$$merge$$

achilleasa:
$$merge$$

howbazaar:
I wonder if the slightly snappier name "disable-raft-fsync" would be better?

achilleasa:
@howbazaar I am really bad at naming things :D 

I personally prefer the longer and more descriptive variant but don't object to renaming it if you prefer the latter.

achilleasa:
!!build!! 60min timeout

achilleasa:
!!build!! more CI timeouts

achilleasa:
!!build!! more CI timeouts

achilleasa:
!!build!!

achilleasa:
!!build!!

achilleasa:
!!build!!

achilleasa:
$$merge$$ (CI is green but windows tests break due to a misconfiguration; windows tests were green before our changes so I will go ahead and land this)

achilleasa:
$$merge$$ (intermittent test failure)

PR #11656, opened 2020-06-02
Increment juju to 2.8.1

jujubot:


wallyworld:
Not needed cause this is the rc branch.

PR #11665, opened 2020-06-03
Replace a convoluted fake pointer with a simple one.

howbazaar:
The manifold test just needs a non-nil state object. For whatever reason when I did this originally I didn't think to do the obvious thing, but instead used an unsafe pointer conversion. This was then replaced by a more complicated pointer conversion.

The simple thing is just to create a zero'ed version of the struct.

## QA steps

Run the tests.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1881845

### DIFF
--- a/worker/multiwatcher/manifold_test.go
+++ b/worker/multiwatcher/manifold_test.go
@@ -4,8 +4,6 @@
 package multiwatcher_test
 
 import (
-	"unsafe"
-
 	"github.com/juju/errors"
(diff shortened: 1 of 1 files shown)


hpidcock:
Also now we only have "unsafe" imports in windows code!

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #11668, opened 2020-06-04
Acme wtf removal

wallyworld:
## Description of change

The worker/uniter tests were compiling an actual jujud from source and using that in full stack tests. Many of the tests were for functionality tested elsewhere, eg does a hook command reject bad input, or does the deployer unpack a charm correctly. And some tests were for things that should have been covered elsewhere eg there were tests which ran an action and tested results but the test for the results watcher should have been in state package where the watcher is implemented.

This PR creates mocks for 2 key components used by the uniter worker
- the runner which handles execution of hooks, commands, actions
- the deployer which unpacks a charm

The worker config has been changed to allow either the real implementation of a mock to be passed in to the uniter.

In addition, it seems that the tests should have been failing - and they do now without additional changes. New hooks were added - relation-created and remove. The relation related tests were not updated to expect these hooks. With changes in this PR, the tests failed until these expectations were updated.

Finally, the obsoleted GitSuite was removed.

## QA steps

deploy both a k8s and vm charm
use show-status-log to check hook execution
use "juju exec --unit foo/0 -- hostname" to check running scripts
use "juju exec --unit foo/0 -- config-get to check running hook commands
run an action also and check the results


state/action_test.go
```diff
@@ -1292,6 +1292,62 @@ func (s *ActionSuite) TestWatchActionLogs(c *gc.C) {
...
 }
 
+func (s *ActionSuite) TestWatchActionResults(c *gc.C) {
+	w := s.Model.WatchActionResultsFilteredBy(s.unit)
+	defer statetesting.AssertStop(c, w)
+	wc := statetesting.NewStringsWatcherC(c, s.State, w)
+
+	operationID, err := s.Model.EnqueueOperation("a test")
+	c.Assert(err, jc.ErrorIsNil)
+	// queue some actions before starting the watcher
```
howbazaar, line 1302: Comments should be full sentences.

worker/uniter/runner/context/export_test.go
```diff
@@ -102,9 +102,10 @@ func NewMockUnitHookContext(mockUnit *mocks.MockHookUnit) *HookContext {
 
 func NewMockUnitHookContextWithState(mockUnit *mocks.MockHookUnit, state *uniter.State) *HookContext {
 	return &HookContext{
-		unit:   mockUnit,
-		state:  state,
-		logger: loggo.GetLogger("test"),
+		unitName: mockUnit.Tag().Id(),
```
howbazaar, line 105: Perhaps a comment outlining why we need or want a unitName?

worker/uniter/uniter.go
```diff
@@ -444,7 +450,7 @@ func (u *Uniter) loop(unitTag names.UnitTag) (err error) {
 			StartRetryHookTimer: retryHookTimer.Start,
 			StopRetryHookTimer:  retryHookTimer.Reset,
 			Actions: actions.NewResolver(
-				u.logger.Child("actions"),
+				u.logger.Child("expectedActions"),
```
howbazaar, line 453: Why change this? The original had the logger be "juju.worker.uniter.actions" which is what was there before, why change it?

worker/uniter/uniter_test.go
```diff
@@ -50,37 +48,36 @@ var _ = gc.Suite(&UniterSuite{})
...
 	s.dataDir = c.MkDir()
 	toolsDir := tools.ToolsDir(s.dataDir, "unit-u-0")
 	err := os.MkdirAll(toolsDir, 0755)
 	c.Assert(err, jc.ErrorIsNil)
 
-	coretesting.CopyOrBuildJujuBins(c, toolsDir, "jujud", "jujuc")
-
 	s.oldLcAll = os.Getenv("LC_ALL")
-	os.Setenv("LC_ALL", "en_US")
+	err = os.Setenv("LC_ALL", "en_US")
```
howbazaar, line 58: Don't we have a PatchEnvironment? Hmm.. I thought we did, but I can't find it. I guess this is still fine.

worker/uniter/uniter.go
wallyworld: didn't intend to, maybe a search/replace fail

worker/uniter/uniter_test.go
wallyworld: We have it and I changed the test to use it

worker/uniter/runner/context/export_test.go
wallyworld: yup

state/action_test.go
wallyworld: yup

wallyworld:
$$merge$$

PR #11674, opened 2020-06-05
 Add metrics around the queue length and age for the multiwatcher worker.

howbazaar:
This required passing a clock into the worker, which means all the sources
that were using the worker now also need to pass in a clock.

Adds a number of metrics for monitoring the queue length and age, along
with summaries that give us timings for adding items to the pending deque
and how long things are taking to process.

When adding events to the deque, the worker now checks to see if there is already
a pending event, and if there is, we don't need to add another event as when the
event is being processed, it is using the latest information.

## QA steps

```sh
juju bootstrap lxd test
juju ssh -m controller 0
# look at multiwatcher worker report to observe updates
juju_engine_report
# look at the multiwatcher metrics
juju_metrics
```

Internal changes only.

### DIFF
--- a/apiserver/apiserver_test.go
+++ b/apiserver/apiserver_test.go
@@ -83,6 +83,7 @@ func (s *apiserverConfigFixture) SetUpTest(c *gc.C) {
 	s.mux = apiserverhttp.NewMux()
 
 	multiWatcherWorker, err := multiwatcher.NewWorker(multiwatcher.Config{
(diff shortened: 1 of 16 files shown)


hpidcock:
!!build!!

hpidcock:
!!build!!

hpidcock:
!!build!!

howbazaar:
$$merge$$

PR #11686, opened 2020-06-09
Generate the facade schema with docs

SimonRichardson:
## Description of change

The following handles the facade schema with type and method
documentation. It's a copy of what already happens for the old
js-jujulib documentation creation, but instead, we lift this into juju so
other client libs can use it.

## QA steps

```sh
make rebuild-schema
cd scripts/schemadocs
make clean output.html
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1882571
https://bugs.launchpad.net/juju/+bug/1880634


SimonRichardson:
Closing this as https://github.com/juju/juju/pull/11689 supersedes it.

PR #11687, opened 2020-06-10
Show unit command

wallyworld:
### Checklist

It we want to expose the new API, we'll need to update libjuju

----

## Description of change

Add a new show-unit command. There's 3 separate commits:
- apiserver facade / state implementation
- api client
- CLI command itself

This command adds a show option for units, like for machines, applications etc.
A key point is that you can use it to inspect unit relation data. Until now, you would have had to run 3 hook commands using juju exec to get the same result.

## QA steps

Deploy mediawiki related to a 3 unit mariadb
Deploy mediawiki related to a mariadb offer

Then you can try show-unit, eg

```
$ juju show-unit mediawiki/0
mediawiki/0:
  workload-version: 1.19.14
  machine: "0"
  opened-ports:
  - 80/tcp
  public-address: 10.169.54.195
  charm: cs:mediawiki-19
  leader: true
  relation-info:
  - endpoint: db
    cross-model: true
    related-endpoint: db
    application-data: {}
    related-units:
      mariadb/0:
        in-scope: true
        data:
          database: remote-7f82691643ef447f88f2b34d08a51b91
          egress-subnets: 10.169.54.226/32
          host: 10.169.54.226
          ingress-address: 10.169.54.226
          password: aemaevohbeiyohe
          private-address: 10.169.54.226
          slave: "False"
          user: riekoovahfuokah

$ juju show-unit mariadb/0
mariadb/0:
  workload-version: 10.1.40
  machine: "0"
  opened-ports: []
  public-address: 10.169.54.226
  charm: cs:trusty/mariadb-7
  leader: true
  relation-info:
  - endpoint: cluster
    related-endpoint: cluster
    application-data: {}
    local-unit:
      in-scope: true
      data:
        egress-subnets: 10.169.54.226/32
        ingress-address: 10.169.54.226
        private-address: 10.169.54.226
    related-units:
      mariadb/1:
        in-scope: false
        data: {}
      mariadb/2:
        in-scope: false
        data: {}
  - endpoint: db
    cross-model: true
    related-endpoint: db
    application-data: {}
    related-units:
      remote-7f82691643ef447f88f2b34d08a51b91/0:
        in-scope: true
        data:
          egress-subnets: 10.169.54.195/32
          ingress-address: 10.169.54.195
          private-address: 10.169.54.195
```

### DIFF
--- a/api/application/client.go
+++ b/api/application/client.go
@@ -1020,3 +1020,24 @@ func (c *Client) MergeBindings(req params.ApplicationMergeBindingsArgs) error {
 	}
 	return results.OneError()
 }
+
+// UnitsInfo retrieves units information.
+func (c *Client) UnitsInfo(units []names.UnitTag) ([]params.UnitInfoResult, error) {
+	if apiVersion := c.BestAPIVersion(); apiVersion < 12 {
+		return nil, errors.NotSupportedf("UnitsInfo for Application facade v%v", apiVersion)
+	}
+	all := make([]params.Entity, len(units))
+	for i, one := range units {
+		all[i] = params.Entity{Tag: one.String()}
+	}
+	in := params.Entities{Entities: all}
+	var out params.UnitInfoResults
+	err := c.facade.FacadeCall("UnitsInfo", in, &out)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if resultsLen := len(out.Results); resultsLen != len(units) {
+		return nil, errors.Errorf("expected %d results, got %d", len(units), resultsLen)
+	}
+	return out.Results, nil
+}
--- a/api/application/client.go
+++ b/api/application/client.go
@@ -8,6 +8,7 @@
 package application
 
 import (
+	stderrors "errors"
 	"time"
 
 	"github.com/juju/charm/v7"
@@ -1021,8 +1022,41 @@ func (c *Client) MergeBindings(req params.ApplicationMergeBindingsArgs) error {
 	return results.OneError()
 }
 
+// UnitInfo holds information about a unit.
+type UnitInfo struct {
+	Error error
+
+	Tag             string
+	WorkloadVersion string
+	Machine         string
+	OpenedPorts     []string
+	PublicAddress   string
+	Charm           string
+	Leader          bool
+	RelationData    []EndpointRelationData
+
+	// The following are for CAAS models.
+	ProviderId string
+	Address    string
+}
+
+// RelationData holds information about a unit's relation.
+type RelationData struct {
+	InScope  bool
+	UnitData map[string]interface{}
+}
+
+// EndpointRelationData holds information about a relation to a given endpoint.
+type EndpointRelationData struct {
+	Endpoint         string
+	CrossModel       bool
+	RelatedEndpoint  string
+	ApplicationData  map[string]interface{}
+	UnitRelationData map[string]RelationData
+}
+
 // UnitsInfo retrieves units information.
-func (c *Client) UnitsInfo(units []names.UnitTag) ([]params.UnitInfoResult, error) {
+func (c *Client) UnitsInfo(units []names.UnitTag) ([]UnitInfo, error) {
 	if apiVersion := c.BestAPIVersion(); apiVersion < 12 {
 		return nil, errors.NotSupportedf("UnitsInfo for Application facade v%v", apiVersion)
 	}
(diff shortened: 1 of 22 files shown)


api/application/client.go
```diff
@@ -1020,3 +1020,24 @@ func (c *Client) MergeBindings(req params.ApplicationMergeBindingsArgs) error {
 	}
 	return results.OneError()
 }
+
+// UnitsInfo retrieves units information.
+func (c *Client) UnitsInfo(units []names.UnitTag) ([]params.UnitInfoResult, error) {
```
howbazaar, line 1025: I don't like how we return wire protocol structures from the api calls.

Perhaps this would be a good time to make a good example?

apiserver/facades/client/application/application.go
```diff
@@ -230,12 +246,18 @@ func newFacadeBase(ctx facade.Context) (*APIBase, error) {
...
+		return nil, errors.Trace(err)
+	}
+
 	return NewAPIBase(
 		&stateShim{ctx.State()},
 		storageAccess,
 		ctx.Auth(),
 		blockChecker,
-		facadeModel,
+		&modelShim{facadeModel},
```
howbazaar, line 259: It isn't clear here why there is a shim. Comment maybe?

apiserver/facades/client/application/application.go
```diff
@@ -2622,3 +2646,227 @@ func validateAgentVersions(application Application, versioner AgentVersioner) er
...
+		if leader := leaders[unit.ApplicationName()]; leader == unit.Name() {
+			result.Leader = true
+		}
+		if machineId != "" {
+			machine, err := api.backend.Machine(machineId)
+			if err != nil {
+				out[i].Error = common.ServerError(err)
+				continue
+			}
+			publicAddress, err := machine.PublicAddress()
```
howbazaar, line 2699: PublicAddress is old, and we want to remove it. Is there a better way to get an address for the unit?

apiserver/facades/client/application/application.go
```diff
@@ -2622,3 +2646,227 @@ func validateAgentVersions(application Application, versioner AgentVersioner) er
...
+	return params.UnitInfoResults{out}, nil
+}
+
+func (api *APIBase) openPortsOnMachine(unitName, machineID string) ([]string, error) {
+	var result []string
+	allOpenPorts, err := api.model.AllPorts()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	// We are only looking for the open ports on the "" subnet.
```
howbazaar, line 2741: Why are we only looking for open ports on the "" subnet?

cmd/juju/application/showunit.go
```diff
@@ -0,0 +1,283 @@
...
+)
+
+const showUnitDoc = `
+The command takes deployed unit names as an argument.
+
+Optionally, relation data for only a specified endpoint
+or related unit may be shown, or just the application data. 
+
+Examples:
+    $ juju show-unit mysql/0
```
howbazaar, line 28: Is this how we've done other examples? Not sure we use the $ prefix.

cmd/juju/application/showunit.go
```diff
@@ -0,0 +1,283 @@
...
+		plural = ""
+	}
+	return errors.NotValidf(`unit name%v %v`, plural, strings.Join(invalid, `, `))
+}
+
+// SetFlags implements Command.SetFlags.
+func (c *showUnitCommand) SetFlags(f *gnuflag.FlagSet) {
+	c.ModelCommandBase.SetFlags(f)
+	c.out.AddFlags(f, "yaml", cmd.DefaultFormatters.Formatters())
+	f.StringVar(&c.endpoint, "endpoint", "", "Only show relation data for the specified endpoint")
```
howbazaar, line 96: command arg help strings should start lower case.

cmd/juju/application/showunit.go
```diff
@@ -0,0 +1,283 @@
...
+
+// Info implements Command.Run.
+func (c *showUnitCommand) Run(ctx *cmd.Context) error {
+	client, err := c.newAPIFunc()
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	if v := client.BestAPIVersion(); v < 12 {
```
howbazaar, line 124: Why check this here when it is handled in the api client?

cmd/juju/application/showunit.go
```diff
@@ -0,0 +1,283 @@
...
+		tag, info, err := c.createUnitInfo(one)
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+		output[tag.Id()] = info
+	}
+	return output, nil
+}
+
+type UnitRelationData struct {
```
howbazaar, line 185: These types could be in the API package and returned from the API itself.

apiserver/facades/client/application/application.go
wallyworld: We need a shim when we need to use a method that returns an interface as opposed to a state struct.  I'll add a comment

api/application/client.go
wallyworld: I don't like it either. I was sticking to convention for this facade but will change this method

apiserver/facades/client/application/application.go
wallyworld: This is what we display in status so we need to be consistent.
We need to change it for sure but until then show-unit and status should be the same

apiserver/facades/client/application/application.go
wallyworld: That's what status does - the same logic was applied here. Not 100% sure why

cmd/juju/application/showunit.go
wallyworld: In 99% of cases, we omit the $.
The one place I chose to copy from to get started since the logic is similar was show-application and that uses $. I'll fix both.

cmd/juju/application/showunit.go
wallyworld: We tend to do this quite commonly. It does seem redundant.

cmd/juju/application/showunit.go
wallyworld: These have (UX related) format tags and are structured to be fit for purpose for the layer in which they are used, ie the CLI. The API layer should have its own structs, even if similar. Each layer transition should use different structs

wallyworld:
$$merge$$

PR #11702, opened 2020-06-11
Remove corresponding instanceData when machine is removed.

hmlanigan:
## Description of change

Found during test of a different bug, instanceDataDoc for a machine is not removed when the machine is removed.  The MachineSuite.TestRemoves checked that the hardware characteristic data was no longer there after a machine was removed, however the data wasn't set to begin with due to lack of machine.SetProvisioned().

## QA steps

```sh
juju bootstrap
juju add-machine
juju remove-machine 0
juju-db.bash
juju:PRIMARY> db.instanceData.find()
```

Only the instance data for the controller should be seen.

### DIFF
--- a/state/machine.go
+++ b/state/machine.go
@@ -291,6 +291,16 @@ func getInstanceData(st *State, id string) (instanceData, error) {
 	return instData, nil
 }
 
+// removeInstanceDataOp returns the operation needed to remove the
+// instance data document associated with the given globalKey.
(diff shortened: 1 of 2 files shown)


hmlanigan:
Reproduced with aws and the wrench, updating qa steps

hpidcock:
$$merge$$

SimonRichardson:
$$merge$$

> worker_test.go:736: WorkerSuite.TestRemoveWorkloadApplicationWaitsForResources

PR #11703, opened 2020-06-12
The instance mutator worker ignores manually provisioned machines

wallyworld:
## Description of change

When adding manually provisioned machines to juju, the worker which applies any lxd profiles needed by a charm should not mess with such machines - only lxd containers created by juju should have the profile updated. This prevents juju from trying to apply lxd profiles to any manual machine added to an lxd controller - it was incorrectly trying to apply a profile to a manually added kvm machine for example.
It does mean though the user is responsible for ensuring any lxd containers manually added are suitable for running any workloads deployed to those containers.

## QA steps

bootstrap lxd
juju add-machine ssh:somekvmcontainer
juju deploy somecharmwithalxdprofile --to id

check juju status for errors, previously for example
"cannot upgrade machine's lxd profile: 4: not found"

Also deploy a charm with a profile to a lxd controller and check it is applied.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1883178

### DIFF
--- a/state/allwatcher.go
+++ b/state/allwatcher.go
@@ -221,12 +221,22 @@ func (m *backingMachine) updated(ctx *allWatcherContext) error {
 		wantsVote = err == nil && node.WantsVote()
 		hasVote = err == nil && node.HasVote()
 	}
+	modelId, _ := ctx.entityIDForGlobalKey(modelGlobalKey)
+	modelEntity := ctx.store.Get(modelId)
+	var providerType string
+	if modelEntity != nil {
+		modelInfo := modelEntity.(*multiwatcher.ModelInfo)
+		providerType, _ = modelInfo.Config["type"].(string)
+
+	}
+	isManual := isManualMachine(m.Id, m.Nonce, providerType)
 	info := &multiwatcher.MachineInfo{
 		ModelUUID:                m.ModelUUID,
 		Id:                       m.Id,
 		Life:                     multiwatcher.Life(m.Life.String()),
 		Series:                   m.Series,
 		ContainerType:            m.ContainerType,
+		IsManual:                 isManual,
 		Jobs:                     paramsJobsFromJobs(m.Jobs),
 		SupportedContainers:      m.SupportedContainers,
 		SupportedContainersKnown: m.SupportedContainersKnown,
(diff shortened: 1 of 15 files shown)


state/allwatcher.go
```diff
@@ -221,12 +221,18 @@ func (m *backingMachine) updated(ctx *allWatcherContext) error {
 		wantsVote = err == nil && node.WantsVote()
 		hasVote = err == nil && node.HasVote()
 	}
+	modelId, _ := ctx.entityIDForGlobalKey(modelGlobalKey)
```
howbazaar, line 224: Thank you for not making this another db query.

state/machine.go
```diff
@@ -459,29 +459,37 @@ func (m *Machine) IsManager() bool {
...
+	// To avoid unnecessary db lookups, a little of the
+	// logic from isManualMachine() below is duplicated here
+	// so we can exit early if possible.
+	if strings.HasPrefix(m.doc.Nonce, manualMachinePrefix) {
+		return true, nil
+	}
+	if m.doc.Id != "0" {
+		return false, nil
+	}
+	model, err := m.st.Model()
```
howbazaar, line 471: Do we really want to load the model just to get the config?

wallyworld: Outside the state package, no choice, but since we're in state, we can read the setting collection directly, so I made that change

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

worker/instancemutater/mutater.go
```diff
@@ -87,6 +87,15 @@ func (m *mutater) startMachines(tags []names.MachineTag) error {
 				continue
 			}
 
+			profileChangeWatcher, err := api.WatchLXDProfileVerificationNeeded()
```
hmlanigan, line 90: This introduces a 2nd methodology for what to ignore in terms of WatchLXDProfileVerificationNeeded, which I believe will be confusing in the long term.  In the case of KVM containers we don't call WatchLXDProfileVerification, then in the case of manual machines, we fail WatchLXDProfileVerification.

Suggest `if api.IsManual() { continue}` before the call to api.WatchLXDProfileVerificationNeeded() as an alternative.  

This keeps more business logic of the instance mutater in the worker rather than spreading more out to different packages.

PR #11707, opened 2020-06-15
Move the derived status method into the core/status package.

howbazaar:
Add a Status method to the Application type in core/cache, and have
it use the derived status of the units if the application status
value is Unset.

This work is a prelude to a more efficient allwatcher handling of unit status values when no application status has been set.

### DIFF
--- a/core/cache/application_test.go
+++ b/core/cache/application_test.go
@@ -5,7 +5,9 @@ package cache_test
 
 import (
 	"sync"
+	"time"
 
+	jc "github.com/juju/testing/checkers"
 	"github.com/juju/worker/v2/workertest"
 	"github.com/prometheus/client_golang/prometheus/testutil"
 	gc "gopkg.in/check.v1"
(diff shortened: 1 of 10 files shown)


core/cache/application_test.go
```diff
@@ -79,6 +81,147 @@ func (s *ApplicationSuite) TestConfigWatcherChange(c *gc.C) {
...
+		Status: value,
+		Since:  &when,
+	}
+}
+
+func (s *ApplicationSuite) TestStatusWhenSet(c *gc.C) {
+	model := s.NewModel(cache.ModelChange{
+		Name: "test",
+	})
+	appStatus := s.status(status.Active, time.Now())
```
wallyworld, line 95: maybe make this "maintenance" or something other than "active" as that has special logic
or do both?

howbazaar: While "active" has special meaning to us, inside this package it has no special meaning at all.

howbazaar:
$$merge$$

PR #11708, opened 2020-06-15
Feature/new k8s agent

ycliuhw:
### Checklist

----

## Description of change

This PR introduced a new binary `k8sagent` for k8s unit agent and init command.
Currently, the `unit` and `init` commands are not implemented yet.

## QA steps

```console
$ DOCKER_USERNAME=xxx make microk8s-k8sagent-update

$ docker run -it --rm xxx/k8sagent:2.9-beta1 '/opt/k8sagent --help'
$ docker run -it --rm xxx/k8sagent:2.9-beta1 '/opt/k8sagent init --help'
$ docker run -it --rm xxx/k8sagent:2.9-beta1 '/opt/k8sagent unit --help'
$ docker run -it --rm xxx/k8sagent:2.9-beta1 'cp /opt/k8sagent /opt/juju-introspect && /opt/juju-introspect --help'
$ docker run -it --rm xxx/k8sagent:2.9-beta1 'cp /opt/k8sagent /opt/juju-dumplogs && /opt/juju-dumplogs --help'
$ docker run -it --rm xxx/k8sagent:2.9-beta1 'cp /opt/k8sagent /opt/juju-run && /opt/juju-run --help'

```

## Documentation changes

No

## Bug reference

None


Makefile
```diff
@@ -215,29 +216,48 @@ JUJUD_BIN_DIR              ?= ${BIN_DIR}
 OPERATOR_IMAGE_BUILD_SRC   ?= true
 
 # Import shell functions from make_functions.sh
+# For the k8s operator.
 BUILD_OPERATOR_IMAGE=sh -c '. "${PROJECT_DIR}/make_functions.sh"; build_operator_image "$$@"' build_operator_image
 OPERATOR_IMAGE_PATH=sh -c '. "${PROJECT_DIR}/make_functions.sh"; operator_image_path "$$@"' operator_image_path
 OPERATOR_IMAGE_RELEASE_PATH=sh -c '. "${PROJECT_DIR}/make_functions.sh"; operator_image_release_path "$$@"' operator_image_release_path
+# For the new k8s agent.
+BUILD_K8SAGENT_IMAGE=sh -c '. "${PROJECT_DIR}/make_functions.sh"; build_k8sagent_image "$$@"' build_k8sagent_image
```
ycliuhw, line 224: New variables and targets are defined for the new k8sagent stuff instead of refactoring existing ones by adding new toggle variables.
Because we may change a lot or even remove these new stuff.

ycliuhw:
!!build!!

Makefile
howbazaar: I'm wondering whether or not we want to just include the new k8s agent in our existing operator image, so we just have one image to deal with.  Is there a problem with including it in the current image?

cmd/jujud/run/package_test.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2017 Canonical Ltd.
```
howbazaar, line 1: Should probably update the year.

cmd/k8sagent/init/command.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2012-2020 Canonical Ltd.
```
howbazaar, line 1: just 2020 is fine, don't need a range.

cmd/k8sagent/init/command.go
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package init
```
howbazaar, line 4: Oh... while this may work it gives me the heebies. Given that "init" is a reserved word for functions.

Perhaps naming the package "initialize"?

cmd/k8sagent/init/command.go
```diff
@@ -0,0 +1,44 @@
...
+// Info returns a description of the command.
+func (c *initCommand) Info() *cmd.Info {
+	return jujucmd.Info(&cmd.Info{
+		Name:    "init",
+		Purpose: "initialize k8sagent state",
+	})
+}
+
+func (c *initCommand) Run(_ *cmd.Context) error {
+	logger.Infof("starting k8sagent init command")
```
howbazaar, line 42: Could use the Infof method on the cmd.Context to get some output.

cmd/k8sagent/main_nix.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/juju/juju/juju/names"
+	"github.com/juju/juju/juju/osenv"
+	"github.com/juju/juju/utils/proxy"
+	"github.com/juju/juju/worker/logsender"
+)
+
+var logger = loggo.GetLogger("juju.cmd.k8sagent")
+
+func init() {
+	featureflag.SetFlagsFromEnvironment(osenv.JujuFeatureFlagEnvKey)
```
howbazaar, line 41: I think we probably want to remove this. This approach was from way way back, and we are trying to move to controller feature flags now.

cmd/k8sagent/main_nix.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if err := components.RegisterForServer(); err != nil {
```
howbazaar, line 43: These are primarily for payload and resource API server bits I think. By adding this in we are going to drag in all of the state package, and a bunch of other things we don't actually want.

We may need to readdress this at some stage if api client components are needed, but I'd rather just have them moved into the right place. This may be happening as a piece of work by other folks in the 2.9 branch anyway.

cmd/k8sagent/main_nix.go
```diff
@@ -0,0 +1,198 @@
...
+		logger.Criticalf("unable to register server components: %v", err)
+		os.Exit(1)
+	}
+}
+
+func init() {
+	rand.Seed(time.Now().UTC().UnixNano())
```
howbazaar, line 50: We do still want this though :)

cmd/k8sagent/main_nix.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+var k8sAgentDoc = `
+juju provides easy, intelligent service orchestration on top of models
+such as OpenStack, Amazon AWS, or bare metal. k8sagent is a component
+of juju for managing k8s workloads.
+
+https://jujucharms.com/
```
howbazaar, line 58: This should be "https://juju.is" now.

cmd/k8sagent/main_nix.go
```diff
@@ -0,0 +1,198 @@
...
+	// Just show the last element of the module.
+	lastDot := strings.LastIndex(entry.Module, ".")
+	module := entry.Module[lastDot+1:]
+	return fmt.Sprintf("%s %s %s %s", ts, entry.Level, module, entry.Message)
+}
+
+func k8sAgentCommand(ctx *cmd.Context) (cmd.Command, error) {
+	// Assuming an average of 200 bytes per log message, use up to
+	// 200MB for the log buffer.
+	bufferedLogger, err := logsender.InstallBufferedLogWriter(loggo.DefaultContext(), 1048576)
```
howbazaar, line 96: I feel that we should pull the buffered log writer into a core package. It feels like a better place. Not now, but something to think about later.

cmd/k8sagent/main_nix.go
```diff
@@ -0,0 +1,198 @@
...
+		cmd.WriteError(os.Stderr, err)
+		os.Exit(exit_err)
+	}
+	switch filepath.Base(args[0]) {
+	case names.K8sAgent:
+		code = f.k8sAgentCmd(ctx, args)
+	case names.JujuRun:
+		code = f.jujuRun(ctx, args)
+	case names.JujuDumpLogs:
+		code = f.jujuDumpLogs(ctx, args)
```
howbazaar, line 135: I'm trying to remember what this command actually does?

cmd/k8sagent/main_nix.go
```diff
@@ -0,0 +1,198 @@
...
+		os.Exit(exit_err)
+	}
+	return code
+}
+
+// CAAS implementation currently only works for Linux.
+func main() {
+	defer func() {
+		if r := recover(); r != nil {
+			buf := make([]byte, 4096)
```
howbazaar, line 151: I do wonder whether or not 4096 is big enough for a big stack trace.

cmd/k8sagent/main_nix.go
```diff
@@ -0,0 +1,198 @@
...
+				os.Exit(1)
+			}
+			return cmd.Main(cmdToRun, ctx, args[1:])
+		},
+		jujuRun: func(ctx *cmd.Context, args []string) int {
+			lock, err := machinelock.New(machinelock.Config{
+				AgentName:   "juju-run",
+				Clock:       clock.WallClock,
+				Logger:      loggo.GetLogger("juju.machinelock"),
+				LogFilename: filepath.Join(cmdutil.LogDir, machinelock.Filename),
```
howbazaar, line 172: I wonder whether these names will still make sense later.

Perhaps we should create some type of comment to easily search for k8s filenames.

cmd/k8sagent/main_test.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2012-2020 Canonical Ltd.
```
howbazaar, line 1: Year again.

cmd/k8sagent/unit/command.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2012-2020 Canonical Ltd.
```
howbazaar, line 1: And here.

juju/names/names_windows.go
```diff
@@ -8,6 +8,7 @@ const (
 	Juju             = "juju.exe"
 	Jujuc            = "jujuc.exe"
 	Jujud            = "jujud.exe"
+	K8sAgent         = "k8sagent.exe"
```
howbazaar, line 11: Not sure we need this, but I guess it can't hurt. As there may be some tests we run on windows that expect it.

Perhaps change the contstant value to "not-available" or something.

cmd/k8sagent/init/command.go
ycliuhw: right, it's already in ctx

cmd/k8sagent/main_nix.go
ycliuhw: ah, this should be only run on the controller.
Removed!

cmd/k8sagent/main_nix.go
ycliuhw: pod-lock.log ?

cmd/k8sagent/main_nix.go
ycliuhw: Since the sub commands are empty, so removed this for now.
Might continue to resolve here in following PR when we introduced API accessing in those commands.

cmd/k8sagent/main_nix.go
ycliuhw: This is the current jujud config, I assume k8sagent would need the size less than the  controller needs. 
And we set the all=False, so we only have the current gorounte's trace.

Makefile
ycliuhw: As discussed, we are going to keep k8sagent binary in a separate image for now.
We will merge it with the existing image if we want to in the future.

cmd/jujud/main.go
```diff
@@ -24,25 +24,25 @@ import (
...
-
-	// Import the providers.
-	_ "github.com/juju/juju/provider/all"
 	"github.com/juju/juju/upgrades"
 	"github.com/juju/juju/utils/proxy"
 	"github.com/juju/juju/worker/logsender"
 	"github.com/juju/juju/worker/uniter/runner/jujuc"
+
+	// Import the providers.
+	_ "github.com/juju/juju/provider/all"
```
howbazaar, line 45: I don't think we need this. All of the provider specific code should be in the apiserver.

This may bring in a bunch of things we don't want.

cmd/k8sagent/main_nix.go
howbazaar: I think I'd just add a comment for us to think about the names a bit more. Don't have to solve it now.

cmd/k8sagent/main_nix.go
howbazaar: Why is it still here then?

cmd/jujud/main.go
howbazaar: Sorry, I see now that this is jujud main. This is fine.

ycliuhw:
$$merge$$

PR #11712, opened 2020-06-16
Rewrite juju scp help

timClicks:
Expands the details section to be more readable for readers scanning the text in a hurry. It adds more specific advice for the standard invocation, passing args to `scp` and transfers between machines.

Migrates the examples to the style used by the Kubernetes docs.

Adds extra example for using juju scp to extract k8s config from a Charmed Kubernetes cluster.


timClicks:
$$merge$$

hpidcock:
$$merge$$

PR #11713, opened 2020-06-16
Write proxy env files for new proxy settings

dshcherb:
## Description of change

Write /etc/juju-proxy.conf and /etc/juju-proxy-systemd.conf when
juju-* proxy settings are set, not only when legacy proxy settings are
set.

## QA steps

```sh
juju model-config juju-http-proxy=http://192.0.2.1 juju-https-proxy=http://192.0.2.2 juju-no-proxy=192.0.2.0/24
juju deploy cs:ubuntu
juju run --unit ubuntu/0 'grep -i proxy /etc/juju-proxy.conf'
juju run --unit ubuntu/0 'grep -i proxy /etc/juju-proxy-systemd.conf'
# Validate the settings according to the model-config.
```

## Documentation changes

None, maybe worth a release note.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1883656

### DIFF
--- a/cloudconfig/userdatacfg_unix.go
+++ b/cloudconfig/userdatacfg_unix.go
@@ -308,35 +308,37 @@ func (w *unixConfigure) ConfigureJuju() error {
 		return errors.Trace(err)
 	}
 
-	// Write out the normal proxy settings so that the settings are
-	// sourced by bash, and ssh through that.
-	w.conf.AddScripts(
-		// We look to see if the proxy line is there already as
-		// the manual provider may have had it already.
-		// We write this file out whether we are using the legacy proxy
-		// or the juju proxy to deal with runtime changes. The proxy updater worker
-		// only modifies /etc/juju-proxy.conf, so if changes are written to that file
-		// we need to make sure the profile.d file exists to reflect these changes.
-		// If the new juju proxies are used, the legacy proxies will not be set, and the
-		// /etc/juju-proxy.conf file will be empty.
-		`[ -e /etc/profile.d/juju-proxy.sh ] || ` +
-			`printf '\n# Added by juju\n[ -f "/etc/juju-proxy.conf" ] && . "/etc/juju-proxy.conf"\n' >> /etc/profile.d/juju-proxy.sh`)
 	var proxySettings proxy.Settings
 	if w.icfg.JujuProxySettings.HasProxySet() {
 		proxySettings = w.icfg.JujuProxySettings
+		// We look to see if the proxy file is there already as
+		// the manual provider may have had it already and
+		// remove it when new proxy settings are used.
+		w.conf.AddScripts(`rm -f /etc/profile.d/juju-proxy.sh`)
 	} else if w.icfg.LegacyProxySettings.HasProxySet() {
 		proxySettings = w.icfg.LegacyProxySettings
+		w.conf.AddScripts(
+			fmt.Sprintf(
+				// We look to see if the proxy file is there already as
+				// the manual provider may have had it already.
+				`[ -e /etc/profile.d/juju-proxy.sh ] || printf '%%s\n' %s > /etc/profile.d/juju-proxy.sh`,
+				shquote(proxySettings.AsScriptEnvironment())))
 	}
-	exportedProxyEnv := proxySettings.AsScriptEnvironment()
-	w.conf.AddScripts(strings.Split(exportedProxyEnv, "\n")...)
-	w.conf.AddScripts(
-		fmt.Sprintf(
-			`(printf '%%s\n' %s > /etc/juju-proxy.conf && chmod 0644 /etc/juju-proxy.conf)`,
-			shquote(w.icfg.LegacyProxySettings.AsScriptEnvironment())))
 
-	// Write out systemd proxy settings
-	w.conf.AddScripts(fmt.Sprintf(`printf '%%s\n' %[1]s > /etc/juju-proxy-systemd.conf`,
-		shquote(w.icfg.LegacyProxySettings.AsSystemdDefaultEnv())))
+	if proxySettings.HasProxySet() {
+		exportedProxyEnv := proxySettings.AsScriptEnvironment()
+		w.conf.AddScripts(strings.Split(exportedProxyEnv, "\n")...)
+		w.conf.AddScripts(
+			fmt.Sprintf(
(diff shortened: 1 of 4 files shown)


wallyworld:
As per a comment left in the bug, the current behaviour is by design. On the surface, I don't think we want to land this change. We can discuss out of band if needed.

cloudconfig/userdatacfg_unix.go
```diff
@@ -321,18 +321,22 @@ func (w *unixConfigure) ConfigureJuju() error {
...
-		w.conf.AddScripts(fmt.Sprintf(`printf '%%s\n' %[1]s > /etc/juju-proxy-systemd.conf`,
-			shquote(w.icfg.LegacyProxySettings.AsSystemdDefaultEnv())))
+	var proxySettings proxy.Settings
+	if w.icfg.JujuProxySettings.HasProxySet() {
+		proxySettings = w.icfg.JujuProxySettings
+	} else if w.icfg.LegacyProxySettings.HasProxySet() {
+		proxySettings = w.icfg.LegacyProxySettings
 	}
+	exportedProxyEnv := proxySettings.AsScriptEnvironment()
+	w.conf.AddScripts(strings.Split(exportedProxyEnv, "\n")...)
```
wallyworld, line 331: We only want to do this for legacy settings I think, right?

cloudconfig/userdatacfg_unix.go
```diff
@@ -321,18 +321,22 @@ func (w *unixConfigure) ConfigureJuju() error {
...
+	w.conf.AddScripts(
+		fmt.Sprintf(
+			`(printf '%%s\n' %s > /etc/juju-proxy.conf && chmod 0644 /etc/juju-proxy.conf)`,
+			shquote(w.icfg.LegacyProxySettings.AsScriptEnvironment())))
+
+	// Write out systemd proxy settings
```
wallyworld, line 337: Can we fix the comment?

cloudconfig/userdatacfg_unix.go
```diff
@@ -321,18 +321,22 @@ func (w *unixConfigure) ConfigureJuju() error {
...
-				shquote(w.icfg.LegacyProxySettings.AsScriptEnvironment())))
-
-		// Write out systemd proxy settings
-		w.conf.AddScripts(fmt.Sprintf(`printf '%%s\n' %[1]s > /etc/juju-proxy-systemd.conf`,
-			shquote(w.icfg.LegacyProxySettings.AsSystemdDefaultEnv())))
+	var proxySettings proxy.Settings
+	if w.icfg.JujuProxySettings.HasProxySet() {
+		proxySettings = w.icfg.JujuProxySettings
+	} else if w.icfg.LegacyProxySettings.HasProxySet() {
+		proxySettings = w.icfg.LegacyProxySettings
```
wallyworld, line 328: Need to cater for the case where neither juju nor legacy settings are set

dshcherb: If none are set we can just rely on empty strings in the proxy.Settings struct.

https://github.com/juju/proxy/blob/master/proxy.go#L58-L64

I think I need to have a unit test for that though.

cloudconfig/userdatacfg_unix.go
```diff
@@ -321,18 +321,22 @@ func (w *unixConfigure) ConfigureJuju() error {
...
 	}
+	exportedProxyEnv := proxySettings.AsScriptEnvironment()
+	w.conf.AddScripts(strings.Split(exportedProxyEnv, "\n")...)
+	w.conf.AddScripts(
+		fmt.Sprintf(
+			`(printf '%%s\n' %s > /etc/juju-proxy.conf && chmod 0644 /etc/juju-proxy.conf)`,
+			shquote(w.icfg.LegacyProxySettings.AsScriptEnvironment())))
```
dshcherb, line 335: Need to fix the `w.icfg.LegacyProxySettings` usage here. `proxySettings` need to be used instead.

cloudconfig/userdatacfg_unix.go
```diff
@@ -321,18 +321,22 @@ func (w *unixConfigure) ConfigureJuju() error {
...
+
+	// Write out systemd proxy settings
+	w.conf.AddScripts(fmt.Sprintf(`printf '%%s\n' %[1]s > /etc/juju-proxy-systemd.conf`,
+		shquote(w.icfg.LegacyProxySettings.AsSystemdDefaultEnv())))
```
dshcherb, line 339: Same here: need to fix the `w.icfg.LegacyProxySettings` usage here. `proxySettings` need to be used instead.

cloudconfig/userdatacfg_unix.go
dshcherb: If I am not mistaken, this affects the cloud-init shell only (`runcmd`) on first boot.

```
		exportedProxyEnv := proxySettings.AsScriptEnvironment()
		w.conf.AddScripts(strings.Split(exportedProxyEnv, "\n")...)
```

About this one I am not sure since there are potential `preruncmd` and `postruncmd` commands added by a user via model-config.

If we follow the same principle here, I think we could:

* expose JUJU_CHARM_{HTTP, HTTPS, NO}_PROXY in the cloud-init shell environment and document that those should be used explicitly;
* tell users that /etc/juju-proxy.conf can be sourced in any custom `postruncmd`s.
    * wouldn't work on Windows since we don't write a file there.

dshcherb:
I have updated the PR (https://github.com/juju/juju/pull/11713/commits/ca712e53b3cba55bd32e5b745d5be066ad60a0f5), although further updates are required to unit tests for the proxyupdater.

Before adding those, I would like to have a further discussion about:

* whether writing to `/etc/juju-proxy.conf` is acceptable
    * on CAAS models everything is a bit different and having `/etc/juju-proxy.conf` in an operator pod does not have the same effect as one might expect (this may or may not be a problem);
    * The reasons outlined here https://bugs.launchpad.net/charm-glance-simplestreams-sync/+bug/1883656/comments/6 - manual restart of charm-deployed services won't be covered.
* currently writing to a file on Windows isn't present.

dshcherb:
I converted this to a draft and will likely close it given the above comment and the fact that we cat steal proxy settings from `juju-run`:

https://bugs.launchpad.net/charm-glance-simplestreams-sync/+bug/1883656/comments/9

petevg:
Closing to clean up hanging PRs. Feel free to reopen if this is still fresh.

PR #11724, opened 2020-06-18
Tweak ec2 instance parsing to include more instance types

wallyworld:
## Description of change

The way we got instance type metadata for ec2 was just wrong.
A symptom was that we failed to know about instance types in some regions, and falsely claimed instance types were available when they were not (for a given user). We also needed to hand craft and build into juju a large map each time a new instance type was available.

This PR uses the EC2 API to query available instance types for the credential in use, and also looks up the cost info as well. This is cached on the provider and reset when the cloud/cred info changes.

Also, bootstrap --build-agent now takes account of GOARCH so we can test custom agent binaries on different arches.

## QA steps

The test case covers the issue in the bug (arm64 m6g instance in ap-northeast-1)

```
$ export GOARCH=arm64
$ juju bootstrap --constraints 'instance-type=m6g.4xlarge arch=arm64' aws/ap-northeast-1 foo --config vpc-id=vpc-44f2eb26 --build-agent
$ juju exec -m controller --machine 0 -- uname -a
Linux ip-172-31-6-199 5.3.0-1023-aws #25~18.04.1-Ubuntu SMP Fri Jun 5 15:19:18 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1872670


provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
+// Copyright 2016 Canonical Ltd.
```
wallyworld, line 1: file was renamed from instance_information.go and some methods from the old internal/ec2instancetypes package copied across.
supportedInstanceTypes() and instancetypeCosts() are new

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+			instanceTypeRegions[*offering.InstanceType].Add(*offering.Location)
+			instTypeNames = append(instTypeNames, offering.InstanceType)
+		}
+		token = offeringResults.NextToken
+		if token == nil {
+			break
+		}
+	}
+
+	// Populate the costs fo the instance types in use.
```
hpidcock, line 214: fo sure

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+				for _, instArch := range info.ProcessorInfo.SupportedArchitectures {
+					instType.Arches = append(instType.Arches, archName(*instArch))
+				}
+			}
+			if instRegions, ok := instanceTypeRegions[instType.Name]; !ok || instRegions.Size() < len(zoneNames) {
+				instType.Deprecated = true
+			}
+			allInstanceTypes = append(allInstanceTypes, instType)
+		}
+		if len(instTypeNames) <= 100 {
```
hpidcock, line 264: move all this logic to the top of the for loop

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+
+	// Compose the results.
+	var allInstanceTypes []instances.InstanceType
+	for {
+		querySize := len(instTypeNames)
+		if querySize > 100 {
+			querySize = 100
+		}
+		instTypeParams := &ec2.DescribeInstanceTypesInput{
+			InstanceTypes: instTypeNames[0:querySize],
```
hpidcock, line 228: move paging logic to start of for loop like
```
for len(instTypeNames) > 0 {
	querySize := len(instTypeNames)
	if querySize > 100 {
		querySize = 100
	}
	page := instTypeNames[0:querySize]
	instTypeNames = instTypeNames[querySize:]
```

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+		}
+		instTypeResults, err := ec2Session.DescribeInstanceTypes(instTypeParams)
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+		for _, info := range instTypeResults.InstanceTypes {
+			instType := instances.InstanceType{
```
hpidcock, line 235: Breakout into its own function like convertEC2InstanceType

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+				Name:       *info.InstanceType,
```
hpidcock, line 236: nil check needed

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+						uint64(*info.VCpuInfo.DefaultVCpus))
+					instType.CpuPower = &cpupower
+				}
+			}
+			if info.MemoryInfo != nil && info.MemoryInfo.SizeInMiB != nil {
+				instType.Mem = uint64(*info.MemoryInfo.SizeInMiB)
+			}
+			if info.ProcessorInfo != nil {
+				for _, instArch := range info.ProcessorInfo.SupportedArchitectures {
+					instType.Arches = append(instType.Arches, archName(*instArch))
```
hpidcock, line 256: nil check instArch

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+		return nil, errors.Trace(err)
+	}
+	if hasDefaultVPC {
+		return allInstanceTypes, nil
+	}
+
+	// The region has no default VPC, and the user has not specified
+	// one to use. We filter out any instance types that are not
+	// supported in EC2-Classic.
+	supportedInstanceTypes := make([]instances.InstanceType, 0, len(allInstanceTypes))
```
hpidcock, line 284: poor allocation, I'd prefer if we had a numClassic() func or const that can seed the preallocation size here.

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+				price, err := strconv.ParseFloat(*sp.SpotPrice, 32)
+				if err == nil {
+					costs[*sp.InstanceType] = uint64(1000 * price)
+				}
+			}
+		}
+		token = priceResults.NextToken
+		// token should be nil when there's no more records
+		// but it never gets set to nil so there's a bug in the api.
+		if len(priceResults.SpotPriceHistory) < 1000 {
```
hpidcock, line 323: MaxResults and len() here should use a const

provider/ec2/instancetypes.go
hpidcock: 100 should be a const

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	// Query the instance type names for the region and credential in use.
+	var instTypeNames []*string
+	instanceTypeRegions := make(map[string]set.Strings)
+	var token *string
+	for {
+		offeringResults, err := ec2Session.DescribeInstanceTypeOfferings(&ec2.DescribeInstanceTypeOfferingsInput{
+			LocationType: aws.String("availability-zone"),
+			MaxResults:   aws.Int64(100),
```
hpidcock, line 194: 100 should be a const

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+// See:
+//     http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-vpc.html#vpc-only-instance-types
+func supportsClassic(instanceType string) bool {
+	parts := strings.SplitN(instanceType, ".", 2)
+	if len(parts) < 2 {
+		return false
+	}
+	switch strings.ToLower(parts[0]) {
+	case
+		"c1", "c3",
```
hpidcock, line 120: lets just make this a map

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+		spParams.NextToken = token
+		priceResults, err := ec2Session.DescribeSpotPriceHistory(spParams)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		for _, sp := range priceResults.SpotPriceHistory {
+			if _, ok := costs[*sp.InstanceType]; !ok {
+				price, err := strconv.ParseFloat(*sp.SpotPrice, 32)
+				if err == nil {
+					costs[*sp.InstanceType] = uint64(1000 * price)
```
hpidcock, line 316: maybe a const here to describe the conversion factor

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+				logger.Tracef("got instance types:\n%#v", result)
+			}
+			e.instTypes = result
+		}
+	}()
+
+	// Use a cached copy if populated as it's mildly
+	// expensive to fetch each time.
+	if len(e.instTypes) > 0 {
+		return e.instTypes, nil
```
hpidcock, line 167: Cache should expire.

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+		MaxResults:    aws.Int64(1000),
+		StartTime:     aws.Time(time.Now()),
+		Filters:       []*ec2.Filter{{Name: aws.String("availability-zone")}},
+	}
+	for _, zone := range zoneNames {
+		spParams.Filters[0].Values = append(spParams.Filters[0].Values, aws.String(zone))
+	}
+	for {
+		spParams.NextToken = token
+		priceResults, err := ec2Session.DescribeSpotPriceHistory(spParams)
```
hpidcock, line 308: Shouldn't use SpotPriceHistory as this is the spot price, not the actual price. Instead use this API https://docs.aws.amazon.com/sdk-for-go/api/service/pricing/

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+func instanceTypeCosts(ec2Session ec2iface.EC2API, instTypeNames []*string, zoneNames []string, costs map[string]uint64) error {
+	var token *string
+	spParams := &ec2.DescribeSpotPriceHistoryInput{
+		InstanceTypes: instTypeNames,
+		MaxResults:    aws.Int64(1000),
+		StartTime:     aws.Time(time.Now()),
+		Filters:       []*ec2.Filter{{Name: aws.String("availability-zone")}},
+	}
+	for _, zone := range zoneNames {
+		spParams.Filters[0].Values = append(spParams.Filters[0].Values, aws.String(zone))
```
hpidcock, line 304: Lets just make the filter a variable, then move the spParams decl down after we have the filter ready.

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+		if !supportsClassic(instanceType.Name) {
+			continue
+		}
+		supportedInstanceTypes = append(supportedInstanceTypes, instanceType)
+	}
+	return supportedInstanceTypes, nil
+}
+
+// instanceTypeCosts queries the latest spot price for the given instance types.
+func instanceTypeCosts(ec2Session ec2iface.EC2API, instTypeNames []*string, zoneNames []string, costs map[string]uint64) error {
```
hpidcock, line 295: costs should just be a return value

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+		zoneFilter.Values = append(zoneFilter.Values, z.ZoneName)
+	}
+
+	// Query the instance type names for the region and credential in use.
+	var instTypeNames []*string
+	instanceTypeRegions := make(map[string]set.Strings)
+	var token *string
+	for {
+		offeringResults, err := ec2Session.DescribeInstanceTypeOfferings(&ec2.DescribeInstanceTypeOfferingsInput{
+			LocationType: aws.String("availability-zone"),
```
hpidcock, line 193: not sure if this works for "local-zones" but it should

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+			}
+			if info.MemoryInfo != nil && info.MemoryInfo.SizeInMiB != nil {
+				instType.Mem = uint64(*info.MemoryInfo.SizeInMiB)
+			}
+			if info.ProcessorInfo != nil {
+				for _, instArch := range info.ProcessorInfo.SupportedArchitectures {
+					instType.Arches = append(instType.Arches, archName(*instArch))
+				}
+			}
+			if instRegions, ok := instanceTypeRegions[instType.Name]; !ok || instRegions.Size() < len(zoneNames) {
```
hpidcock, line 259: lets change this logic to, if the region has more than 3 zones, then as long as 3 zones support it, allow it, otherwise if the region has less than 3 zones, only support if its supported in all zones.

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,328 @@
...
+			Name:   aws.String("region-name"),
+			Values: []*string{aws.String(e.cloud.Region)},
+		}},
+	})
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	var zoneNames []string
+	zoneFilter := &ec2.Filter{Name: aws.String("location")}
+	for _, z := range zoneResults.AvailabilityZones {
```
hpidcock, line 182: This needs to support local zones like `us-west-2-lax-1a` or filter them out completely. Or if only local zones support an instance type, mark it as deprecated.

provider/ec2/instancetypes.go
wallyworld: cost is just a guide (it's only used to break a tie when name and/or memory and/or cpu are equal)
i looked at that other api and even tried to use it but it doesn't have the filtering needed to allow it to be queried efficiently (that i could see)
spot price (or anything really) is fine for the use case here

provider/ec2/instancetypes.go
wallyworld: yeah, i just copied this and didn't notice.
we're talking small numbers, easier just to append to an initially empty slice

provider/ec2/instancetypes.go
wallyworld: arch is never nil either. if len(SupportedArches) > 0, the it contains values
(the sdk just has everything as a pointer)

provider/ec2/instancetypes.go
wallyworld: InstanceType is never nil

provider/ec2/instancetypes.go
wallyworld: Filters are OR'ed together so it's fine was my assumption.

provider/ec2/instancetypes.go
wallyworld: I am not sure we care about the distinction. The only reason we get zone names is so that we can use them to know when to mark an instance as deprecated (terrible name but it precedes this work). If an instance type is not available across all zones (combination of az plus local), then deprecated = true.

provider/ec2/instancetypes.go
wallyworld: For the purposes of this PR, I think it gets the job done and is much, much better than what we had. The main time we would expect the contents to be stale is when the reion or credential changes and those 2 events trigger a reload.
I can add a TODO to follow up with a proper cache.

provider/ec2/instancetypes.go
wallyworld: made it a set, same thing

provider/ec2/instancetypes.go
hpidcock: Defensive programming would ignore the assumption and just continue the loop if the value is nil.

provider/ec2/instancetypes.go
hpidcock: Defensive programming would ignore the assumption and just continue the loop if the value is nil.

provider/ec2/instancetypes.go
hpidcock: Hmmm yeah lets go with the TODO, in the future we could add a tuneable that defaults to 1 day or something like that.

provider/ec2/instancetypes.go
hpidcock: Considering the spot price is unstable, we need to revisit this.

hpidcock: Possibly even just not using price. Instead actually parsing the instance type name, and creating a better weighting based system. T and M type instances should be chosen above everything else, then Compute and Memory optimized, then all the others after.

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,348 @@
...
+	if info.ProcessorInfo != nil {
+		for _, instArch := range info.ProcessorInfo.SupportedArchitectures {
+			instType.Arches = append(instType.Arches, archName(*instArch))
+		}
+	}
+	instRegions, ok := instanceTypeRegions[instType.Name]
+	if !ok {
+		instType.Deprecated = true
+	} else {
+		instType.Deprecated = instRegions.Size() < 3 && instRegions.Size() < len(zoneNames)
```
hpidcock, line 302: Add a comment

provider/ec2/instancetypes.go
```diff
@@ -0,0 +1,348 @@
...
+	info *ec2.InstanceTypeInfo,
+	instanceTypeRegions map[string]set.Strings,
+	costs map[string]uint64,
+	zoneNames []string,
+) instances.InstanceType {
+	instType := instances.InstanceType{
+		Name:       *info.InstanceType,
+		VirtType:   virtType(info),
+		Deprecated: info.CurrentGeneration == nil || !*info.CurrentGeneration,
+		Cost:       costs[*info.InstanceType],
```
hpidcock, line 278: Is the default cost of 0 treated correctly? We should default unknown costs as max uint64 so they are chosen as a last resort.

wallyworld: Other providers default to 0 but then again they don't have costs for anything. 
It's hard to know because in my testing, only 2 instance types didn't get a cost, one being t1.nano, and that's a low cost instance so we'd want to select that if possible.
Maybe using max int is ok

provider/ec2/instancetypes.go
wallyworld: added check here and in other places eg zone names

wallyworld:
$$merge$$

SimonRichardson:
@wallyworld a discourse post/email would be great about this new (great) change

SimonRichardson:
Also note https://github.com/juju/juju/pull/11725#pullrequestreview-433055598

PR #11731, opened 2020-06-19
Ensure all HA controllers use the same bakery public keys

wallyworld:
## Description of change

Add a new doc to the controllers collection to hold macaroon bakery config.
We store the 4 public keys used in various bakeries. This ensure that each controller uses the same public key, and avoids the issue where a discharge request ends up at a different controller and fails because a different key is in use.

## QA steps

Hard to test the actual issue, but can do a regression test to ensure macaroons work. There's unit tests to check that the same key is served each time it is asked for.

Deploy some HA controllers in 2.7.6
Upgrade to this PR, check controllers collection for bakery config.
Add a user and grant read on a model.
Login as that user and check the cookies jar to confirm macaroons are used.
Run juju status several times; restart the controller agents and repeat.
There will be prompts for password at times but login and status should work each time.
Check a cmr scenario also.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1856071

### DIFF
--- a/apiserver/stateauthenticator/context.go
+++ b/apiserver/stateauthenticator/context.go
@@ -93,9 +93,11 @@ func newAuthContext(
 		//"is-authenticated-user",
 		func(ctx context.Context, cond, arg string) error { return nil },
 	)
+
+	bakeryConfig := st.NewBakeryConfig()
 	location := "juju model " + st.ModelUUID()
 	var err error
-	ctxt.localUserThirdPartyBakeryKey, err = bakery.GenerateKey()
+	ctxt.localUserThirdPartyBakeryKey, err = bakeryConfig.GetLocalUsersThirdPartyKey()
 	if err != nil {
 		return nil, errors.Annotate(err, "generating key for local user third party bakery key")
 	}
@@ -114,7 +116,7 @@ func newAuthContext(
 		return nil, errors.Trace(err)
 	}
 	locator := bakeryutil.BakeryThirdPartyLocator{PublicKey: ctxt.localUserThirdPartyBakeryKey.Public}
-	localUserBakeryKey, err := bakery.GenerateKey()
+	localUserBakeryKey, err := bakeryConfig.GetLocalUsersKey()
 	if err != nil {
 		return nil, errors.Annotate(err, "generating key for local user bakery key")
 	}
@@ -256,7 +258,8 @@ func newExternalMacaroonAuth(st *state.State, clock clock.Clock, identClient ide
 		return nil, errMacaroonAuthNotConfigured
 	}
 	idPK := controllerCfg.IdentityPublicKey()
-	key, err := bakery.GenerateKey()
+	bakeryConfig := st.NewBakeryConfig()
+	key, err := bakeryConfig.GetExternalUsersThirdPartyKey()
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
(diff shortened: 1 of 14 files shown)


apiserver/stateauthenticator/context.go
```diff
@@ -298,8 +301,5 @@ func newExternalMacaroonAuth(st *state.State, clock clock.Clock, identClient ide
 		Location: idURL,
```
howbazaar, line 301: There is a comment from axw just above here about storing the key in mongo. Could remove that now?

state/bakerystorage/config.go
```diff
@@ -0,0 +1,188 @@
...
+	return &keyPair, nil
+}
+
+// GetLocalUsersThirdPartyKey returns the third party key pair used with the local users bakery.
+func (b *bakeryConfig) GetLocalUsersThirdPartyKey() (*bakery.KeyPair, error) {
+	doc, err := b.bakeryConfigDoc()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if doc.LocalUsersKey == "" {
```
howbazaar, line 128: Checking wrong doc element?

state/bakerystorage/config.go
```diff
@@ -0,0 +1,188 @@
...
+	return &keyPair, nil
+}
+
+// GetExternalUsersThirdPartyKey returns the third party key pair used with the external users bakery.
+func (b *bakeryConfig) GetExternalUsersThirdPartyKey() (*bakery.KeyPair, error) {
+	doc, err := b.bakeryConfigDoc()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if doc.LocalUsersKey == "" {
```
howbazaar, line 146: and here

state/bakerystorage/config.go
```diff
@@ -0,0 +1,188 @@
...
+	return &keyPair, nil
+}
+
+// GetOffersThirdPartyKey returns the key pair used with the cross model offers bakery.
+func (b *bakeryConfig) GetOffersThirdPartyKey() (*bakery.KeyPair, error) {
+	doc, err := b.bakeryConfigDoc()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	if doc.LocalUsersKey == "" {
```
howbazaar, line 164: Perhaps there should be a helper method that takes a json serialized string and a name and returns a key pair and error?

wallyworld: good idea

wallyworld:
$$merge$$

wallyworld:
$$merge$$

SimonRichardson:
I'm not sure how this landed, it really shouldn't have because of this https://github.com/juju/juju/pull/11731/files#diff-7f05070aef97e2fdd5573b2bc8fcdc99R151.

I'll try and get the static analysis working again.

state/bakerystorage/config.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (b *bakeryConfig) deserialiseKey(data, label string) (*bakery.KeyPair, error) {
+	if data == "" {
+		return nil, errors.NotValidf("empty " + label)
+	}
+
+	var keyPair bakery.KeyPair
+	err := json.Unmarshal([]byte(data), &keyPair)
+	if err != nil {
+		err = errors.Trace(err)
```
SimonRichardson, line 151: I've made a PR to fix this https://github.com/juju/juju/pull/11733/files#diff-7f05070aef97e2fdd5573b2bc8fcdc99R151

PR #11732, opened 2020-06-19
Move caas status display methods to core/status. 

howbazaar:
Move the unit and application display status functions to the core/status package.

Both the functions and the tests are moved into the core package.
Nothing else is touched here.


core/status/caas.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		// If no unit status set, assume still allocating.
+		return StatusInfo{
+			Status:  Waiting,
+			Message: message,
+			Since:   containerStatus.Since,
+		}
+	}
+	if unitStatus.Status != Active && unitStatus.Status != Waiting && unitStatus.Status != Blocked {
```
hpidcock, line 34: I know this is just relocated but this should be a switch.

hpidcock:
$$merge$$

PR #11737, opened 2020-06-21
Add caas fields to allwatcher and cache.

howbazaar:
In order to be able to correctly derive an application status in the cache for a CAAS model,
additional data needs to be passed in. Both the podSpec and container status for the units
need to be available.

A drive-by in this branch was chaning the loggers used in the allwatcher package to all use
the allWatcherLogger not the standard state package logger. This is really useful when 
debugging the low level watcher tests.

In order to correctly handle the status changes in the allwatcher, the method to determine
the parent had to be refactored to return the suffix as an additional parameter. This is then
used in the statusDoc updated method to update the correct status value.

## QA steps

Just the unit tests.

### DIFF
--- a/core/cache/application.go
+++ b/core/cache/application.go
@@ -70,30 +70,50 @@ func (a *Application) Units() []Unit {
 	return units
 }
 
+// ExpectsWorkload only makes sense for CAAS applications.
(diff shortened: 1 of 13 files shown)


howbazaar:
$$merge$$

PR #11740, opened 2020-06-22
Get application status from the model cache.

howbazaar:
This is the culmination of three branches dealing with all watcher efficiency in larger models.

The primary change is the move of the determination of application status from the database layer to the apiserver and core/cache layer.

Applications are now created with a status of "unset" rather than "waiting", and the apiserver layer now asks the model cache for the application's status. If the application has an unset status, it derives the status from its units. This piece of work has landed already.

One particular change here is around the application status that is used for the cross model offer status. This status is informational only and possibly a little slower in updating now. This will be fixed when the cross model offer watcher is moved from the state layer to the model cache. A bug has been filed for this.

Another change that was not initially expected was the result of the status-get charm hook tool, which called in the uniter facade. This call would poorly iterate all of the units of an application to get the status values. This has also been updated to happen in a single database query now.

## QA steps

Bootstrap and deploy some units. Run juju status.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1479278
https://bugs.launchpad.net/juju/+bug/1857075


state/state.go
```diff
@@ -1208,17 +1216,9 @@ func (st *State) AddApplication(args AddApplicationArgs) (_ *Application, err er
...
-		Updated:    st.clock().Now().UnixNano(),
-		// This exists to preserve questionable unit-aggregation behaviour
-		// while we work out how to switch to an implementation that makes
-		// sense.
-		NeverSet: true,
-	}
-	if model.Type() == ModelTypeCAAS {
-		statusDoc.StatusInfo = status.MessageWaitForContainer
+		ModelUUID: st.ModelUUID(),
+		Status:    status.Unset,
```
ycliuhw, line 1220: So now, we won't see `waiting for machine` and `waiting for container` message anymore?

ycliuhw: not sure if any users have script expects the waiting status

state/upgrades.go
```diff
@@ -2896,3 +2896,52 @@ func AddBakeryConfig(pool *StatePool) error {
...
+			if !ok {
+				// This shouldn't happen, but if it is there,
+				// just ignore this one.
+				continue
+			}
+
+			update := bson.M{"$unset": bson.M{"neverset": nil}}
+			if neverset {
+				update["$set"] = bson.M{
+					"status":     "unset",
```
ycliuhw, line 2932: do we want only change to `waiting` status to `unset`?

apiserver/facades/agent/uniter/status.go
```diff
@@ -75,5 +95,113 @@ func (s *StatusAPI) UnitStatus(args params.Entities) (params.StatusResults, erro
 // ApplicationStatus returns the status of the Applications and its workloads
 // if the given unit is the leader.
 func (s *StatusAPI) ApplicationStatus(args params.Entities) (params.ApplicationStatusResults, error) {
-	return s.applicationGetter.Status(args)
+	result := params.ApplicationStatusResults{
```
ycliuhw, line 98: so this was moved from ApplicationStatusGetter.Status without changes?
not easy to read the diff here

state/leaseupgrades_test.go
```diff
@@ -15,8 +15,6 @@ import (
 	coretesting "github.com/juju/juju/testing"
 )
 
-var _ = gc.Suite(&upgradesSuite{})
```
hpidcock, line 18: Was this registered elsewhere?

state/state.go
howbazaar: If the value is unset, then it will get the information from the units. The unit statuses haven't changed, so they are likely to say "waiting for container". If there are no units, then it will be "unknown"

state/upgrades.go
howbazaar: If anyone sets a value, it changes neverset to false.

apiserver/facades/agent/uniter/status.go
howbazaar: Well, the change was to get the application status from the cache, and the unit status values from the new method that only takes one query.

state/leaseupgrades_test.go
howbazaar: Yes, this was registered in upgrades_test.go.  The fact that it was registered here too meant that we were running all the tests twice.

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #11746, opened 2020-06-24
Rewrite juju status help text

timClicks:
Examples are now more representative, language is simpler and the formatting is better aligned to an 80 character limit.


timClicks:
$$merge$$

PR #11751, opened 2020-06-24
Fix apiserver premature restart.

howbazaar:
The default gate.Manifold() was creating a gate where the lock
was reused for the different start methods. This is not what we
want. If the lock is to be reused, it should use the gate.ManifoldEx
method.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1884709


SimonRichardson:
$$merge$$

PR #11752, opened 2020-06-24
[dev] Export virtual port type when migrating

achilleasa:
## Description of change

Small follow-up PR to the OVS support work to export/import the `VirtualPortType` field for link layer devices when migrating models.
 
## QA steps

Not required.


achilleasa:
$$merge$$

PR #11756, opened 2020-06-25
Handle missing security groups

SimonRichardson:
## Description of change

If a security group is missing or has been previously been removed then
we should output error messages to the operator. Just continue onwards
and remain silent.

The code is actually really simple and just checks for the error before
bubbling it up through the stack.

Fixes LP#1864564

## QA steps

It is expected that you've followed the setting up [microstack with juju](https://microstack.run/docs/juju).

```sh
juju bootstrap microstack test
juju deploy cs:~jameinel/ubuntu-lite-7
juju add-unit ubuntu-lite -n 1
```

Remove the new security group via microstack, openstack CLI.

```sh
juju remove-unit ubuntu-lite/1
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1864564

### DIFF
--- a/provider/openstack/firewaller.go
+++ b/provider/openstack/firewaller.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/juju/clock"
+	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 	"github.com/juju/retry"
 	gooseerrors "gopkg.in/goose.v2/errors"
@@ -33,7 +34,7 @@ const (
 
 var extractControllerRe = regexp.MustCompile(GroupControllerPattern)
 
-//factory for obtaining firawaller object.
+// FirewallerFactory for obtaining firawaller object.
 type FirewallerFactory interface {
 	GetFirewaller(env environs.Environ) Firewaller
 }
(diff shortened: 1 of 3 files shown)


SimonRichardson:
This PR really needs to be finished and then we could have integration tests around this https://github.com/juju/juju/pull/11253

hmlanigan:
@SimonRichardson, agreed we shouldn't fail here.  I'd question how we got to the state though?  What went wrong.

hmlanigan:
@SimonRichardson, looking at neutronFirewaller.ensureGroup(), I wonder if the mutex is missing from another place that is deleting a group or group rule.  As well, perhaps we should also evaluate if deleting group rules at this point is a good idea?

provider/openstack/firewaller.go
```diff
@@ -950,13 +957,16 @@ func (c *neutronFirewaller) closePortsInGroup(ctx context.ProviderCallContext, n
...
 	for _, rule := range rules {
 		for _, p := range group.Rules {
 			if !secGroupMatchesIngressRule(p, rule) {
 				continue
 			}
-			err := neutronClient.DeleteSecurityGroupRuleV2(p.Id)
-			if err != nil {
+			if err := neutronClient.DeleteSecurityGroupRuleV2(p.Id); err != nil {
+				if gooseerrors.IsNotFound(err) {
+					continue
```
howbazaar, line 968: continue or break?

SimonRichardson: `continue`, as we want to continue attempting to clear the rest of the security groups.

SimonRichardson: No, you're right, it should be break.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

> remoteRelationsSuite.TestRemoteRelationsChangedError

PR #11761, opened 2020-06-25
Add output to info.

hmlanigan:
## Description of change

The charmhub api Info call is now available.  Use it to display output.
Data received is currently in alpha form.  There are a few TODOs as a
result.

## QA steps
```console
$ juju bootstrap localhost --build-agent
$ juju info wordpress
WARNING cannot unmarshal charm config: invalid config: empty configuration
name:        wordpress
charm-id:    charmCHARMcharmCHARMcharmCHARM01
summary:     WordPress is a full featured web blogging tool, this charm deploys it.
publisher:   Wordress Charmers
subordinate: false
description: |
  This will install and setup WordPress optimized to run in the cloud. By default it will place
  Ngnix and php-fpm configured to scale horizontally with Nginx's reverse proxy.
channels:
  latest/stable:     1.0.3  2019-12-16  (16)  12MB
  latest/candidate:  1.0.3  2019-12-16  (16)  12MB
  latest/beta:       1.0.3  2019-12-16  (16)  12MB
  latest/edge:       1.0.3  2019-12-16  (16)  12MB

# bundles look slightly different
$ juju info osm
name:        osm
bundle-id:   bundleBUNDLEbundleBUNDLEbundle01
summary:     A bundle by charmed-osm.
publisher:   charmed-osm
description: Single instance OSM bundle.
channels:
  latest/stable:     1.0.3  2019-12-16  (61)  12MB
  latest/candidate:  1.0.3  2019-12-16  (61)  12MB
  latest/beta:       1.0.3  2019-12-16  (61)  12MB
  latest/edge:       1.0.3  2019-12-16  (61)  12MB
```

## Documentation changes

Will be needed for the charm hub integration work for juju 2.9


cmd/juju/charmhub/printinfo.go
```diff
@@ -0,0 +1,535 @@
...
+	publisher, ok := iw.in.Charm.Publisher["display-name"]
+	if !ok {
+		return nil
+	}
+	_, err := fmt.Fprintf(iw.w, "publisher:   %s\n", publisher)
+	return err
+}
+
+func (iw *infoWriter) maybePrintDescription() error {
+	if iw.in.Charm.Description == "nil" {
```
SimonRichardson, line 82: I don't like that :cry:, can we also check for empty as well?

cmd/juju/charmhub/printinfo.go
```diff
@@ -0,0 +1,535 @@
...
+
+func (iw *infoWriter) maybePrintStoreURL() error {
+	// TODO (hml) 2020-06-24
+	// implement once the data is available.
+	return nil
+}
+
+func (iw *infoWriter) maybePrintChannels() error {
+	// TODO (hml) 2020-06-24
+	// Implement up arrow for closed channels
```
SimonRichardson, line 106: Do we have enough information to work that out?

cmd/juju/charmhub/printinfo.go
```diff
@@ -0,0 +1,535 @@
...
+	*infoWriter
+	charmMeta   *charm.Meta
+	charmConfig *charm.Config
+}
+
+func (c *charmInfoWriter) Print() error {
+	c.maybeUnmarshalCharmMetadata()
+	c.maybeUnmarshalCharmConfig()
+	// Order is important here, to keep the output
+	// the same.
```
SimonRichardson, line 188: This should totally be a visitor pattern and that way the order is made explicit by the order of the slice.

```go
printers := []Printer{Name{}, ID{}, Summary{}, Publisher{}}
for _, printer := range printers {
    if err := printer.Print(w); err != nil {
        return errors.Trace(err)
    }
}
```

cmd/juju/charmhub/printinfo.go
```diff
@@ -0,0 +1,535 @@
...
+	if err := c.maybePrintName(); err != nil {
+		return nil
```
SimonRichardson, line 190: Why not return `err` here?

cmd/juju/charmhub/printinfo.go
```diff
@@ -0,0 +1,535 @@
...
+	return nil
+}
+
+func formatRelationPart(rels map[string]charm.Relation) (string, bool) {
+	if len(rels) <= 0 {
+		return "", false
+	}
+	var relations string
+	for k, v := range rels {
+		relations = relations + "  " + k + ": " + v.Name + "\n"
```
SimonRichardson, line 315: `fmt.Sprintf(` would make this a lot easier to read.

cmd/juju/charmhub/printinfo.go
```diff
@@ -0,0 +1,535 @@
...
+
+func (c *charmInfoWriter) maybeUnmarshalCharmMetadata() {
+	if c.in.DefaultRelease.Revision.MetadataYAML == "" {
+		return
+	}
+	m := c.in.DefaultRelease.Revision.MetadataYAML
+	meta, err := charm.ReadMeta(bytes.NewBufferString(m))
+	if err != nil {
+		// Do not fail on unmarshalling metadata, log instead
+		c.warningf(errors.Annotate(err, "cannot unmarshal charm metadata").Error())
```
SimonRichardson, line 328: Why? This should always be valid?

cmd/juju/charmhub/printinfo.go
```diff
@@ -0,0 +1,535 @@
...
+		width = 80
+	}
+
+	width -= 3
+	if width > 100 {
+		// any wider than this and it gets hard to read
+		width = 100
+	}
+
+	return width
```
SimonRichardson, line 502: Not sure why we do this, we're essentially saying you can have between 80 and 100 characters on screen. Even if the user-supplied less we force to 80 or more we say 100, it seems like a lot of effort to say, no.

cmd/juju/charmhub/printinfo.go
hmlanigan: Thanks for catching this.  My IDE changed behavior and I'm not totally used to it, thought I caught most of it.  Select and replace doesn't always do that.

cmd/juju/charmhub/printinfo.go
hmlanigan: This was from snap info output.

cmd/juju/charmhub/printinfo.go
hmlanigan: yes, should always be valid.  however our live test data is not.  :-/

cmd/juju/charmhub/printinfo.go
hmlanigan: no more calculating width.  re-designed to print via yaml.  slices of strings and columns have been converted to string in the yaml struct for readability.

cmd/juju/charmhub/printinfo.go
hmlanigan: We do, but it's a bit ugly and will take some api/apiserver changes to do.  I've made a card: https://trello.com/c/4xjRg73L/2737-add-closed-channels-to-info-output

cmd/juju/charmhub/infowriter.go
```diff
@@ -0,0 +1,275 @@
...
+func newInfoWriter(ctx *cmd.Context, in *charmhub.InfoResponse) InfoWriter {
+	iw := &infoWriter{
+		w:        ctx.Stdout,
+		warningf: ctx.Warningf,
+		in:       in,
+	}
+	if iw.in.Type == "charm" {
+		return &charmInfoWriter{infoWriter: iw}
+	}
+	return &bundleInfoWriter{infoWriter: iw}
```
SimonRichardson, line 43: I bet `infoWriter` and the others don't need the pointers, if that's the case we should remove and rename `newInfoWriter` to `makeInfoWriter`.

hmlanigan: It's not, only the unmarshal methods required a pointer receiver.

hmlanigan:
squashed the commits into one, no need to keep the delta between start and finish.

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #11763, opened 2020-06-26
[2.8] Display channel when upgrading charm

achilleasa:
## Description of change

 The channel from which the charm was fetched is recorded when the charm gets deployed and the operator has no means to query this information when they are trying to upgrade their charm.

This PR adds an info-level message to the juju CLI that includes the configured channel when upgrading to a charm that is hosted remotely.

## QA steps

```console

# The CLI now displays an info message with the configured channel for the upgrade
$ juju upgrade-charm ubuntu-lite --path cs:~jameinel/ubuntu-lite-7
Looking up metadata for charm cs:~jameinel/ubuntu-lite-7 (channel: stable)
ERROR already running specified charm "cs:~jameinel/ubuntu-lite-7"

$ juju upgrade-charm ubuntu-lite --path cs:~jameinel/ubuntu-lite-8 --channel edge
Looking up metadata for charm cs:~jameinel/ubuntu-lite-8 (channel: edge)
ERROR cannot resolve URL "cs:~jameinel/ubuntu-lite-8": charm or bundle not found

# Also, try it out with a *local* charm and ensure that the info message is not displayed as 
# it only makes sense for charms fetched from the store
```

## Documentation changes

@timClicks Do we need to update our `juju upgrade-charm` docs to include the above info message?

## Bug reference

https://bugs.launchpad.net/juju/+bug/1880973

### DIFF
--- a/cmd/juju/application/upgradecharm.go
+++ b/cmd/juju/application/upgradecharm.go
@@ -5,6 +5,7 @@ package application
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 
 	"github.com/juju/charm/v7"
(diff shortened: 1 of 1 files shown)


achilleasa:
$$merge$$

erik78se:
Nice! This is would definetly help a confused user!

PR #11765, opened 2020-06-26
[2.8] Enable relation-list tool to list remote (or peer) application

achilleasa:
## Description of change

Prior to this change, the relation-list tool only produces output when the remote relations has at least one unit present. This makes it hard for charm authors to introspect the remote application name when no units are available (e.g. they are still coming up or they have been explicitly removed without also removing the remote application).

This PR makes the relation-list tool more consistent with our other relation tools by adding a `--app` flag. If the charm specifies this flag as part of the tool invocation, the output will contain the remote application name (or the local application name if this is a peer relation).

## QA steps

```console
$ juju deploy wordpress
$ juju deploy mysql
$ juju relate wordpress mysql

# Wait for both units to come up

# This is the normal output of relation-list
$ juju run --unit mysql/0 "relation-list -r 2"
wordpress/0

# Remove unit and wait for it to go away
$ juju remove-unit wordpress/0

# Relation list should now be empty
$ juju run --unit mysql/0 "relation-list -r 2"

# But we should still be able to grab the remote app name
$ juju run --unit mysql/0 "relation-list -r 2 --app"
wordpress

# This should also work for peer relations
$ juju run --unit mysql/0 "relation-list -r 0 --app"
mysql
```

## Documentation changes

@timClicks we need to update the docs for `relation-list` to mention `--app` (2.8+)

## Bug reference

https://bugs.launchpad.net/juju/+bug/1871480

### DIFF
--- a/tests/main.sh
+++ b/tests/main.sh
@@ -103,7 +103,7 @@ show_help() {
         # shellcheck disable=SC2086
         output="${output}\n    $(green ${test})|Runs the ${name} tests"
     done
-    echo "${output}" | column -t -s "|"
+    echo -e "${output}" | column -t -s "|"
 
     echo ""
     echo "Examples:"
@@ -124,7 +124,7 @@ show_help() {
     exit 1
 }
 
-while getopts "hH?:vVtAsaxrlpS" opt; do
+while getopts "hH?vVtAs:a:x:rl:p:S:" opt; do
     case "${opt}" in
     h|\?)
         show_help
@@ -134,52 +134,42 @@ while getopts "hH?:vVtAsaxrlpS" opt; do
         ;;
     v)
         VERBOSE=2
-        shift
         ;;
     V)
         VERBOSE=3
-        shift
         alias juju="juju --debug"
         ;;
     t)
         TEST_VERBOSE=3
-        shift
         alias juju="juju --debug"
         ;;
     A)
         RUN_ALL="true"
-        shift
         ;;
     s)
-        SKIP_LIST="${2}"
-        shift 2
+        SKIP_LIST="${OPTARG}"
         ;;
     a)
-        ARITFACT_FILE="${2}"
-        shift 2
+        ARITFACT_FILE="${OPTARG}"
         ;;
     x)
-        OUTPUT_FILE="${2}"
-        shift 2
+        OUTPUT_FILE="${OPTARG}"
         ;;
     r)
         export BOOTSTRAP_REUSE="true"
-        shift
         ;;
     l)
-        export BOOTSTRAP_REUSE_LOCAL="${2}"
+        export BOOTSTRAP_REUSE_LOCAL="${OPTARG}"
         export BOOTSTRAP_REUSE="true"
-        CLOUD=$(juju show-controller "${2}" --format=json | jq -r ".[\"${2}\"] | .details | .cloud")
-        export BOOTSTRAP_PROVIDER="${CLOUD}"
-        shift 2
+        CLOUD=$(juju show-controller "${OPTARG}" --format=json | jq -r ".[\"${OPTARG}\"] | .details | .cloud")
+        PROVIDER=$(juju clouds --client 2>/dev/null | grep "${CLOUD}" | awk '{print $4}')
+        export BOOTSTRAP_PROVIDER="${PROVIDER}"
         ;;
     p)
-        export BOOTSTRAP_PROVIDER="${2}"
-        shift 2
+        export BOOTSTRAP_PROVIDER="${OPTARG}"
         ;;
     S)
-        export BOOTSTRAP_SERIES="${2}"
-        shift 2
+        export BOOTSTRAP_SERIES="${OPTARG}"
         ;;
     *)
         echo "Unexpected argument ${opt}" >&2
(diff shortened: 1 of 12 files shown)


achilleasa:
!!build!! (TestStatusAllFormats)

petevg:
I ran the QA steps for this. Everything worked save for the last step, which showed the peer relations if "-r 1" was specified, rather than "-r 0". If that was just a typo in the QA steps, I think that we're good on this one :-)

achilleasa:
@petevg there are basically 3 relation IDs that get generated in this setup (0, 1 and 2) but depending on the order that they get inserted in the DB you may end up with a different ID. You can figure out the correct IDs if you connect to the controller mongo instance  (see [here](https://discourse.juju.is/t/login-into-mongodb/309) for a nice script you can add to your PATH) and run ```db.relations.find().pretty()```.

worker/uniter/runner/context/relation.go
```diff
@@ -175,3 +179,9 @@ func (ctx *ContextRelation) Suspended() bool {
 func (ctx *ContextRelation) SetStatus(status relation.Status) error {
 	return errors.Trace(ctx.ru.Relation().SetStatus(status))
 }
+
+// RemoteApplicationName returns the application on the other end of this
+// relation from the perspective of this unit.
+func (ctx *ContextRelation) RemoteApplicationName() string {
```
hmlanigan, line 185: Remote is making me think of CMR specifically here.  Can RequirerApplicationName be used here?

worker/uniter/runner/jujuc/relation-list.go
```diff
@@ -16,10 +16,11 @@ import (
...
 type RelationListCommand struct {
 	cmd.CommandBase
-	ctx             Context
-	RelationId      int
-	relationIdProxy gnuflag.Value
-	out             cmd.Output
+	ctx                   Context
+	RelationId            int
+	relationIdProxy       gnuflag.Value
+	ListRemoteApplication bool
```
hmlanigan, line 22: inline with the other suggestions: s/Remote/Requirer

worker/uniter/runner/jujuc/relation-list.go
```diff
@@ -49,8 +50,9 @@ func (c *RelationListCommand) Info() *cmd.Info {
 
 func (c *RelationListCommand) SetFlags(f *gnuflag.FlagSet) {
 	c.out.AddFlags(f, "smart", cmd.DefaultFormatters.Formatters())
-	f.Var(c.relationIdProxy, "r", "specify a relation by id")
+	f.Var(c.relationIdProxy, "r", "Specify a relation by id")
 	f.Var(c.relationIdProxy, "relation", "")
+	f.BoolVar(&c.ListRemoteApplication, "app", false, "List remote application instead of participating units")
```
hmlanigan, line 55: s/List remote application/List requiring application/

or participating application.

worker/uniter/runner/context/relation.go
```diff
@@ -32,6 +32,10 @@ type Relation interface {
 
 	// Tag returns the relation tag.
 	Tag() names.RelationTag
+
+	// OtherApplication returns the name of the application on the other
+	// end of the relation (from this unit's perspective).
+	OtherApplication() string
```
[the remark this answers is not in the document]
achilleasa: The type that implements this interface (in api/uniter IIRC) already defined the `OtherApplication` method so I simply exposed it via this interface.

worker/uniter/runner/context/relation.go
achilleasa: My interpretation is that a relation has a local application (the one for which this unit is running a hook for) and the remote application on the other end. It's akin to the local and remote endpoints of a TCP socket.

However, if you think this is confusing, I could simply rename this to `OtherApplicationName` to be consistent with the uniter API method signatures (see above comment)

achilleasa: Note that CMR or not makes no difference. You still have a local and remote application; the only difference is that the remote part can live in another model and/or controller.

worker/uniter/runner/jujuc/relation-list.go
achilleasa: The bug description seems to use the same terminology (remote application) so we should probably ask @petevg to clarify whether external users and/or stakeholders use different terminology and make sure that docs and hook tool usage are consistent.

achilleasa:
@hmlanigan I will be adding some integration tests to this PR.

tests/main.sh
```diff
@@ -103,7 +103,7 @@ show_help() {
         # shellcheck disable=SC2086
         output="${output}\n    $(green ${test})|Runs the ${name} tests"
     done
-    echo "${output}" | column -t -s "|"
+    echo -e "${output}" | column -t -s "|"
```
achilleasa, line 106: The '-e' ensures the line-feeds from L104 get interpreted correctly

tests/main.sh
```diff
@@ -124,7 +124,7 @@ show_help() {
     exit 1
 }
 
-while getopts "hH?:vVtAsaxrlpS" opt; do
+while getopts "hH?vVtAs:a:x:rl:p:S:" opt; do
```
achilleasa, line 127: You can just append a `:` after an option to have getopts put the flag arg into `$OPTARG`. This also allows getopts to handle all shifts for you.

SimonRichardson: Agreed :+1:

achilleasa:
!!build!!

achilleasa:
$$merge$$

PR #11768, opened 2020-06-29
Fix eks lb svc addresses getter;

ycliuhw:
## Description of change

move `svc` ingress address getter from CAAS test to provider level because different provider might have the address in different field.


ycliuhw:
$$merge$$

ycliuhw:
$$merge$$

PR #11769, opened 2020-06-29
str vs unicode for pexcept in acceptancetests

hmlanigan:
## Description of change

Update expected text for nw-grant-revoke and nw-add-cloud-many to be unicode instead of string.  These test are currently failing due to a recent change to except.spawn in client.py to encoding='UTF-8' 

## QA steps

```console
$ cd $GO_PATH/src/github.com/juju/juju/acceptancetests
./assess --clouds-yaml <path to clouds.yaml>  user_grant_revoke
./assess --clouds-yaml <path to clouds.yaml>  add_cloud
```

PR #11770, opened 2020-06-29
one change was missing: str vs unicode for pexcept in acceptancetests

hmlanigan:
## Description of change

Follow on to #11769 , one unicode change got left behind.


hmlanigan:
$$merge$$

PR #11771, opened 2020-06-30
Record the correct device link for Azure disks

wallyworld:
## Description of change

When provisioning Azure storage, we were making up a bus address to store with the block device entity in state. But it was wrong. We instead record the device link. This fixes the deployment of charms with storage.

## QA steps

bootstrap azure
juju deploy postgresql --storage pgdata=5G
juju status --storage

Also test with a bespoke charm with multiple storage defined to check that block devices are matched up correctly.
Also check that upgrading a broken deployment fixes storage.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1884018

### DIFF
--- a/provider/azure/storage.go
+++ b/provider/azure/storage.go
@@ -607,6 +607,8 @@ func (v *azureVolumeSource) AttachVolumes(ctx context.ProviderCallContext, attac
 	return results, nil
 }
 
+const azureDiskDeviceLink = "/dev/disk/azure/scsi1/lun%d"
+
 func (v *azureVolumeSource) attachVolume(
 	vm *compute.VirtualMachine,
 	p storage.VolumeAttachmentParams,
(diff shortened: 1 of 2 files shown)


provider/azure/storage.go
```diff
@@ -607,6 +607,8 @@ func (v *azureVolumeSource) AttachVolumes(ctx context.ProviderCallContext, attac
 	return results, nil
 }
 
+const azureDiskDeviceLink = "/dev/disk/azure/scsi1/lun%d"
```
howbazaar, line 610: is it always scsi1?

wallyworld: From what I can tell from the docs and examples online, yeah

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #11775, opened 2020-06-30
Only create cinder volumes with AZ if it makes sense

wallyworld:
## Description of change

We attempt to provision a cinder volume in the same AZ as the instance it's being attached to.
This isn't always possible - the volumes may have a different set of allowable AZ, or no AZ at all (just the default "nova").
So we first query the supported volume AZ and only if there's a match with the instance AZ do we then use that AZ when creating the volume.

Needs https://github.com/go-goose/goose/pull/87 to land first

## QA steps

on prodstack (via wendigo) which only has volume AZ "nova" and an old API
juju bootstrap
juju deploy postgresql --storage pgdata=cinder,1G
juju status --storage

## Bug reference

https://bugs.launchpad.net/juju/+bug/1885639

### DIFF
--- a/provider/openstack/cinder.go
+++ b/provider/openstack/cinder.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 	"github.com/juju/schema"
 	"github.com/juju/utils"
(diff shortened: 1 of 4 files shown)


hmlanigan:
For QA, from prodstack I got:
$ openstack availability zone list --volume
Availability zones list not supported by Block Storage API

from server stack I got:
$ openstack availability zone list --volume
+-----------+-------------+
| Zone Name | Zone Status |
+-----------+-------------+
| nova      | available   |
+-----------+-------------+

Ran qa steps with server stack.  The downside is that there is only 1 AZ for the instances.

provider/openstack/cinder.go
```diff
@@ -250,16 +251,40 @@ func (s *cinderVolumeSource) createVolume(
 	// This helps to avoid a situation with all storage residing in a single
 	// AZ that upon failure would effectively take down attached instances
 	// whatever zone they were in.
+	// However, we first attempt to query the possible volume availability zones.
+	// If the API is old and does not support explicit volume AZs, or volumes can
+	// only be provisioned in say the default "nova" zone and the instance is in
+	// a different zone, we won't attempt to use the instance zone because that
+	// won't work and we'll get a 400 error back.
 	az := ""
 	if arg.Attachment != nil {
```
hmlanigan, line 260: it'd be nice if we could break this out into a different method.  make createVolume simpler to read.

az, err := s.availabilityZoneForVolume()

wallyworld: good idea

wallyworld:
!!build!!

hpidcock:
!!build!!

hpidcock:
!!build!!

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #11779, opened 2020-06-30
[dev] Switch to core/network type for ProviderInterfaceInfo

achilleasa:
## Description of change

This is a mechanical change that removes `ProviderInterfaceInfo` from the `network` package and changes all references to it to use the same type from the `core/network` package. This allows us to drop a redundant import for types such as `environ.Networking`.

In addition, the `MACAddress` field of this type has been renamed to the more generic `HardwareAddress` to cater for devices that use different address naming conventions (e.g. infiniband uses GUIDs instead of MAC addresses).

## QA steps

Bootstrap a controller on each one of the modified providers.


achilleasa:
$$merge$$

PR #11780, opened 2020-06-30
Export Admin Facade

SimonRichardson:
## Description of change

The following outputs the Admin facade so it makes it possible to login
via another source, other than using the local credentials.

The code is quite simple as we fall back to the facade.Type rather than
asking the registry for the information. We have to do this, as the
Admin facade details are outside of the registry and I'm not sure why
this is...

## QA steps

```sh
make rebuild-schema
```

## Bug reference

Fixes: https://bugs.launchpad.net/juju/+bug/1885324

### DIFF
--- a/apiserver/facades/schema.json
+++ b/apiserver/facades/schema.json
@@ -849,6 +849,267 @@
         }
     },
     {
+        "Name": "Admin",
+        "Description": "admin is the only object that unlogged-in clients can access. It holds any\nmethods that are needed to log in.",
+        "Version": 3,
+        "AvailableTo": [
+            "controller-machine-agent",
+            "machine-agent",
+            "unit-agent",
+            "controller-user",
+            "model-user"
+        ],
+        "Schema": {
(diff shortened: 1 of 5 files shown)


apiserver/facades/schema.json
```diff
@@ -848,6 +848,267 @@
             }
         }
     },
+    {
+        "Name": "Admin",
+        "Description": "admin is the only object that unlogged-in clients can access. It holds any\nmethods that are needed to log in.",
```
SimonRichardson, line 853: I don't believe this is true anymore, we should update the description.

wallyworld: CMR callers can access various facades, but they do need a valid macaroon to get past auth

SimonRichardson:
$$merge$$

SimonRichardson:
https://trello.com/c/qwHY8gdk/2739-bug-1885324-export-the-admin-facade-in-schema

PR #11786, opened 2020-07-01
Add new line to juju config output

SimonRichardson:
## Description of change

The following adds the correct standard Unix newline after the output
of a config value. Originally it was missing, there was some debate about
adding it, as it _might_ breaks existing scripts, but @wallyworld has made
the decision to add fix it (which I think is right).

It should now be possible to correctly use `|` when using `juju config`.

The code is really simple, just force the usage of `fmt.Fprintln`, which
will do exactly what operators expect to happen.

## QA steps

```sh
juju deploy lxd test
juju deploy mediawiki
juju config mediawiki skin
```

Notice the newline is correctly added:
```sh
vector

```

Instead of
```sh
vector%
```

## Documentation changes

We should note this in the release docs @timClicks, that `juju config` correctly
follows Unix conventions.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1879809

### DIFF
--- a/cmd/juju/application/config.go
+++ b/cmd/juju/application/config.go
@@ -396,6 +396,7 @@ func (c *configCommand) getConfig(client applicationAPI, ctx *cmd.Context) error
 	if err != nil {
 		return err
 	}
+
 	if len(c.keys) == 1 {
 		logger.Infof("format %v is ignored", c.out.Name())
 		key := c.keys[0]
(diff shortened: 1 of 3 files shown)


SimonRichardson:
https://trello.com/c/PRD9dOdS/2742-bug-1879809-juju-config-does-not-print-new-line-at-end-of-output

wallyworld:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #11790, opened 2020-07-01
[2.8] Filter non live relations from relation ids output

achilleasa:
## Description of change

This PR provides a stop gap fix for https://bugs.launchpad.net/juju/+bug/1870013 by filtering the output of the `relation-ids` hook tool to exclude relations that are in a `Dead` state.

Note that this PR _does not address the underlying issue_ (correctly populate the hook context for network-get) but rather prevents the manifestation of the linked bug for workflows that iterate the output of `relation-ids` to obtain additional relation information (e.g. via network-get).

## QA steps

```console
$ juju bootstrap lxd test
$ juju deploy mysql
$ juju deploy wordpress
$ juju deploy statusnet
$ juju relate mysql wordpress
$ juju relate mysql statusnet

$ juju run --unit mysql/0 "relation-ids db"
db:2
db:3

# Remove relation and wait a bit for the removal to be ACK'd
$ juju remove-relation mysql wordpress
$ juju run --unit mysql/0 "relation-ids db"
db:3

# The non-patched 2.8 juju includes the deleted relation in the above command output thus causing the bug to manifest
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1870013

### DIFF
--- a/worker/uniter/runner/jujuc/relation-ids.go
+++ b/worker/uniter/runner/jujuc/relation-ids.go
@@ -12,6 +12,7 @@ import (
 	"github.com/juju/gnuflag"
 
 	jujucmd "github.com/juju/juju/cmd"
+	"github.com/juju/juju/core/life"
 )
 
 // RelationIdsCommand implements the relation-ids command.
@@ -73,7 +74,7 @@ func (c *RelationIdsCommand) Run(ctx *cmd.Context) error {
 	}
 	for _, id := range ids {
 		r, err := c.ctx.Relation(id)
-		if err == nil && r.Name() == c.Name {
+		if err == nil && r.Name() == c.Name && r.Life() != life.Dead {
 			result = append(result, r.FakeId())
 		} else if err != nil && !errors.IsNotFound(err) {
 			return errors.Trace(err)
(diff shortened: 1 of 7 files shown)


worker/uniter/runner/jujuc/relation-ids.go
```diff
@@ -73,7 +74,7 @@ func (c *RelationIdsCommand) Run(ctx *cmd.Context) error {
 	}
 	for _, id := range ids {
 		r, err := c.ctx.Relation(id)
-		if err == nil && r.Name() == c.Name {
+		if err == nil && r.Name() == c.Name && r.Life() == life.Alive {
```
wallyworld, line 77: I'm not sure this is exactly what we want. As a stop gap we could include the relation is life != dead (alive or dying ok). We need to allow dying relations to be included still.

The root cause issue is that network-get does not exclusively use cached info on the hook context, and it errors if a relation is not found. It does handle dead relations ok.

So the proper fix is to change how network-get works by including all info it needs on the hook context. But there's a lot of server side logic and a large part of the model involved so it's not trivial.

At best this PR will narrow the window where the issue can occur, so it's worth landing from that point of view (so long as the above check is changed to != dead). NB I think juju does not set a relation to dead until all post remove processing is done so ignoring dead relations I think is ok.

achilleasa: I updated the check to `!=dead` as suggested and edited the commit msg / PR description to indicate that this is a stop gap fix that only applies to workflows that iterate realtion-ids output to query additional relation information. I will copy this to the bug as well later today.

achilleasa:
$$merge$$

PR #11791, opened 2020-07-02
Fix intermittent cmd/status tests

howbazaar:
Since we are using the model cache now for application status, we need to make sure all the events have been processed before calling the status call.

However this wasn't actually the problem even though it is necessary.

The underlying problem was in the handling of the status updates in the all watcher code.

The existing implementation will accidentally overwrite an error status
if a workload status update comes in. Instead of trying to handle both
agent status and workload status updates in one method, made two, one
of which is very trivial.

There was already an existing test for a unit coming out of an error
state, but there was no test for the updating of workload status when
in an error state. That is the test that is added.

## QA steps

Running the tests using stress-race script.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1884918

### DIFF
--- a/cmd/juju/status/status_internal_test.go
+++ b/cmd/juju/status/status_internal_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/juju/charm/v7"
 	"github.com/juju/cmd"
 	"github.com/juju/cmd/cmdtesting"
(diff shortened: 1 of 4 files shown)


howbazaar:
$$merge$$

PR #11792, opened 2020-07-02
Add centos8

wallyworld:
## Description of change

Add support for centos8 workloads.

To aid testing, fixes were done to the agent simplestreams metadata plugin tools and infrastructure.
The `juju metadata generate-tools` command was fixed to run without needing a controller.
At bootstrap, bespoke agent metadata no longer needs to be signed.
The bespoke agent metadata at bootstrap now includes generated metadata for all linux series. 

## QA steps

The real test is only done when the release jobs generate proper agent binaries for centos8.
We can check that the core parts are wired up using fake metadata.
Create some agent tarballs eg, juju-2.8.0-centos8-amd64.tgz, juju-2.8.0-bionic-amd64.tgz
Run juju metadata generate-tools --stream proposed to create the simplestreams metadata
Bootstrap with this metadata, eg juju bootstrap --metadata-source=~/simplestreams --config agent-stream=proposed ...
Add a centos8 machine, eg juju add-machine --series centos8

If the controller is LXD, you'll need to get a centos8 image locally, aliased to juju/centos8/amd64
If the controller is Azure, it will start a centos8 machine
The centos images seem to lack cloud-init or something as the agent doesn't start, but it's the same for centos7.
The key point is that juju handles centos8 as a supported series.

Check also that juju metadata add-image 1234 --series centos8 works

## Bug reference

https://bugs.launchpad.net/bugs/1881451

### DIFF
--- a/cmd/juju/application/setseries.go
+++ b/cmd/juju/application/setseries.go
@@ -8,7 +8,7 @@ import (
 	"github.com/juju/errors"
 	"github.com/juju/gnuflag"
 	"github.com/juju/names/v4"
-	"github.com/juju/utils/series"
+	"github.com/juju/os/series"
 
 	"github.com/juju/juju/api"
 	"github.com/juju/juju/api/application"
--- a/container/lxd/image.go
+++ b/container/lxd/image.go
@@ -8,13 +8,14 @@ import (
 	"path"
 
 	"github.com/juju/errors"
-	"github.com/juju/juju/core/status"
-	"github.com/juju/juju/environs"
+	jujuos "github.com/juju/os"
+	jujuseries "github.com/juju/os/series"
 	jujuarch "github.com/juju/utils/arch"
-	jujuos "github.com/juju/utils/os"
-	jujuseries "github.com/juju/utils/series"
 	lxd "github.com/lxc/lxd/client"
 	"github.com/lxc/lxd/shared/api"
+
+	"github.com/juju/juju/core/status"
+	"github.com/juju/juju/environs"
 )
 
 // SourcedImage is the result of a successful image acquisition.
@@ -186,8 +187,15 @@ func seriesRemoteAliases(series, arch string) ([]string, error) {
 	case jujuos.Ubuntu:
 		return []string{path.Join(series, arch)}, nil
 	case jujuos.CentOS:
-		if series == "centos7" && arch == jujuarch.AMD64 {
-			return []string{"centos/7/amd64"}, nil
+		if arch == jujuarch.AMD64 {
+			switch series {
+			case "centos7":
+				return []string{"centos/7/amd64"}, nil
+			case "centos8":
+				return []string{"centos/8/amd64"}, nil
+			default:
+				return nil, errors.NotSupportedf("series %q", series)
+			}
 		}
 	case jujuos.OpenSUSE:
 		if series == "opensuseleap" && arch == jujuarch.AMD64 {
--- a/container/lxd/server.go
+++ b/container/lxd/server.go
@@ -6,8 +6,8 @@ package lxd
 import (
 	"github.com/juju/clock"
 	"github.com/juju/errors"
+	"github.com/juju/os"
 	"github.com/juju/utils/arch"
-	"github.com/juju/utils/os"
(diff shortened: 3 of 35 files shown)


cmd/juju/application/setseries.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/juju/errors"
 	"github.com/juju/gnuflag"
 	"github.com/juju/names/v4"
-	"github.com/juju/utils/series"
+	"github.com/juju/os/series"
```
howbazaar, line 11: Thank you. We really should remove series from the utils package.

container/lxd/image.go
```diff
@@ -8,13 +8,14 @@ import (
 	"path"
 
 	"github.com/juju/errors"
-	"github.com/juju/juju/core/status"
-	"github.com/juju/juju/environs"
+	jujuos "github.com/juju/os"
```
howbazaar, line 11: I don't suppose you did a grep across the codebase for juju/utils/os ?

container/lxd/server.go
```diff
@@ -6,8 +6,8 @@ package lxd
 import (
 	"github.com/juju/clock"
 	"github.com/juju/errors"
+	"github.com/juju/os"
```
howbazaar, line 9: Looks like you did :)

provider/azure/internal/imageutils/images.go
```diff
@@ -94,7 +94,7 @@ func SeriesImage(
 		publisher = centOSPublisher
 		offering = centOSOffering
 		switch series {
-		case "centos7":
+		case "centos7", "centos8":
 			sku = "7.3"
```
howbazaar, line 98: Is this sku right for centos8?

provider/azure/internal/imageutils/images_test.go
```diff
@@ -86,7 +86,9 @@ func (s *imageutilsSuite) TestSeriesImageWindows(c *gc.C) {
 }
 
 func (s *imageutilsSuite) TestSeriesImageCentOS(c *gc.C) {
-	s.assertImageId(c, "centos7", "released", "OpenLogic:CentOS:7.3:latest")
+	for _, series := range []string{"centos7", "centos8"} {
+		s.assertImageId(c, series, "released", "OpenLogic:CentOS:7.3:latest")
```
howbazaar, line 90: This looks wrong for a centos8 instance.

provider/azure/internal/imageutils/images.go
wallyworld: Appears so, i tested it live.
I couldn't find any definite docs to what what the SKU should be for centos7 vs 8

provider/azure/internal/imageutils/images_test.go
wallyworld: I'm not sure the SKU has any material association with the series. At least I couldn't find any

wallyworld: Actually, I didn't look hard enough
```
You are viewing an offline list of images, use --all to retrieve an up-to-date list
Offer          Publisher               Sku                 Urn                                                             UrnAlias             Version
-------------  ----------------------  ------------------  --------------------------------------------------------------  -------------------  ---------
CentOS         OpenLogic               7.5                 OpenLogic:CentOS:7.5:latest                                     CentOS               latest
CoreOS         CoreOS                  Stable              CoreOS:CoreOS:Stable:latest                                     CoreOS               latest
Debian         credativ                8                   credativ:Debian:8:latest                                        Debian               latest
openSUSE-Leap  SUSE                    42.3                SUSE:openSUSE-Leap:42.3:latest                                  openSUSE-Leap        latest
RHEL           RedHat                  7-RAW               RedHat:RHEL:7-RAW:latest                                        RHEL                 latest
SLES           SUSE                    12-SP2              SUSE:SLES:12-SP2:latest                                         SLES                 latest
UbuntuServer   Canonical               16.04-LTS           Canonical:UbuntuServer:16.04-LTS:latest 
```

So latest for Centos is 7.5. But we should stick with 7.3 unless we are prepared to do a bunch of testing.

wallyworld:
$$merge$$

PR #11793, opened 2020-07-02
Charmhub API find client

SimonRichardson:
## Description of change

This is a very mechanical change for adding the charmhub find client
methods to the existing charmhub client.

The find methods takes a new `params.Query` instead of a `params.Entity` as
you might want to utilize search queries that don't align to our strict
names tagging system.

For example, searching for `"word press"` (with a space) will return
`wordpress`, but would fail our names.ApplicationTag parsing. Additionally
searching for multiple things `"wordpress + mysql"` would hopefully be
clever enough to work that out as well.

Essentially we're allowing free form text in the search. We may want to
restrict the amount you can actually search for, but we don't restrict
input anywhere else and to be honest, we'll probably get it wrong.

## QA steps

```sh
go test -v ./api/charmhub/...
```

### DIFF
--- a/api/charmhub/data.go
+++ b/api/charmhub/data.go
@@ -18,6 +18,14 @@ func convertCharmInfoResult(info params.InfoResponse) InfoResponse {
 	}
 }
 
+func convertCharmFindResults(responses []params.FindResponse) []FindResponse {
+	results := make([]FindResponse, len(responses))
+	for i, resp := range responses {
+		results[i] = convertCharmFindResult(resp)
+	}
+	return results
+}
+
 func convertCharmFindResult(info params.FindResponse) FindResponse {
 	return FindResponse{
 		Type:           info.Type,
--- a/api/charmhub/data.go
+++ b/api/charmhub/data.go
@@ -32,7 +32,6 @@ func convertCharmFindResult(info params.FindResponse) FindResponse {
 		ID:             info.ID,
 		Name:           info.Name,
 		Entity:         convertEntity(info.Entity),
-		ChannelMap:     convertChannelMap(info.ChannelMap),
 		DefaultRelease: convertOneChannelMap(info.DefaultRelease),
 	}
 }
@@ -113,12 +112,11 @@ type InfoResponse struct {
 }
 
 type FindResponse struct {
-	Type           string       `json:"type"`
-	ID             string       `json:"id"`
-	Name           string       `json:"name"`
-	Entity         Entity       `json:"entity"`
-	ChannelMap     []ChannelMap `json:"channel-map"`
-	DefaultRelease ChannelMap   `json:"default-release,omitempty"`
+	Type           string     `json:"type"`
+	ID             string     `json:"id"`
+	Name           string     `json:"name"`
+	Entity         Entity     `json:"entity"`
+	DefaultRelease ChannelMap `json:"default-release,omitempty"`
 }
 
 type ChannelMap struct {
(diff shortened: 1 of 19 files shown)


SimonRichardson:
@hmlanigan the outstanding issue here is and one we really need to think about, is what happens if you change the charmhub URL `model-config`? I'm unsure it's wise to allow somebody to do this on the fly... let's assume that you do allow the operator to change this, then all unique instance ids for charms/bundles won't match up with the charmhub, potentially breaking an upgrade path. Additionally what happens if the other charmhub doesn't even have the charm/bundle? These seem like dangerous edge cases and very hard to diagnose if somebody does change it?

charmhub/transport/find.go
```diff
@@ -12,7 +12,7 @@ type FindResponse struct {
 	Type           string       `json:"type"`
 	ID             string       `json:"id"`
 	Name           string       `json:"name"`
-	Charm          Charm        `json:"charm,omitempty"`
+	Entity         Entity       `json:"charm"`
```
hmlanigan, line 15: Do we want a TODO for changing the json later, when changed with API?

charmhub/info.go
```diff
@@ -37,5 +37,6 @@ func (c *InfoClient) Info(ctx context.Context, name string) (transport.InfoRespo
 	if err := c.client.Get(ctx, path, &resp); err != nil {
 		return resp, errors.Trace(err)
 	}
+
```
hmlanigan, line 40: Normally a nit, I wouldn't comment on, but seem to be the only change in this file.

apiserver/params/charmhub.go
```diff
@@ -3,20 +3,41 @@
...
+	Results []FindResponse `json:"result"`
+	Errors  ErrorResponse  `json:"errors"`
+}
+
+type FindResponse struct {
+	Type           string         `json:"type"`
+	ID             string         `json:"id"`
+	Name           string         `json:"name"`
+	Entity         CharmHubEntity `json:"entity"`
+	ChannelMap     []ChannelMap   `json:"channel-map"`
```
hmlanigan, line 39: Same as for other FindResponse ChannelMaps

charmhub/transport/find.go
```diff
@@ -12,7 +12,7 @@ type FindResponse struct {
...
 	ChannelMap     []ChannelMap `json:"channel-map"`
```
hmlanigan, line 16: ChannelMap is not defined in the Find api docs.

hmlanigan:
@SimonRichardson agreed on the model-config change.  We should write this up.

apiserver/params/charmhub.go
```diff
@@ -3,8 +3,15 @@
 
 package params
 
+// Query holds the query information when attempting to find possible charms or
+// bundles for searching the charmhub.
+type Query struct {
+	Query string `json:"query"`
```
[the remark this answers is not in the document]
SimonRichardson: Yeah, I just can't test it, so haven't done anything yet.

api/charmhub/data.go
```diff
@@ -90,6 +112,15 @@ type InfoResponse struct {
 	DefaultRelease ChannelMap   `json:"default-release,omitempty"`
 }
 
+type FindResponse struct {
+	Type           string       `json:"type"`
+	ID             string       `json:"id"`
+	Name           string       `json:"name"`
+	Charm          Charm        `json:"charm"`
```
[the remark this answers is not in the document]
SimonRichardson: So this is only optional if you pass in the fields flag. I don't think we want to do that, as we always want all the fields all the time. I see fields being useful for displaying things in the web UI, but for us, we always want all the fields, all the time.

SimonRichardson:
$$merge$$

PR #11800, opened 2020-07-03
Allow absolute vmfolder paths

wallyworld:
## Description of change

The vSphere vmfolder was not able to be set as an absolute path - it was always being interpreted as a subfolder of the DC folder. We now allow absolute paths to be handled.

## QA steps

A version of the fix was tested on site.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1884490


ycliuhw:
```console
$ govc ls
/MyFolder/MyDC/vm
/MyFolder/MyDC/network
/MyFolder/MyDC/host
/MyFolder/MyDC/datastore

# but the `dcfolders.VmFolder.InventoryPath` == /MyDC/vm
```

wallyworld:
$$merge$$

PR #11801, opened 2020-07-03
[2.8] Manual provisioning on non-amd64 instances without extra constraints

achilleasa:
## Description of change

When no bootstrap/model constraints are provided when attempting to bootstrap, the current implementation would fallback to using the arch/series from the machine that the juju CLI executes on. This essentially prevents manually bootstrapping to non-amd64 machines without requiring the operator to explicitly provide additional constraints.

However, the manual provider will connect via ssh to the manual machine and detect both its arch and series. To make use of this information **before** juju attempts to locate the appropriate agent binaries for the bootstrap target, the PR defines a new interface (`HardwareCharacteristicsDetector`) in the `environs` package which is currently only implemented by the `manual` provider.

The bootstrap code has been modified to check whether the environment can provide advance (pre-bootstrap) information about the series/arch for the controller about to be bootstrapped and if so, inject bootstrap constraints for the series and arch **if and only if** the operator has not explicitly specified constraints for those fields.

## QA steps

To run these QA steps you need access to an arm64 and an amd64 instance. If you don't happen to have a rPI near you, you can provision them on AWS (use an `a1` instance for the arm machine) using the bionic AMI. However, before you can use the two machines with juju you need to perform the following additional steps:
- Install a suitable ssh key and make sure you can ssh into the boxes.
- Create a shared security group for the two boxes with the following rules:
   - Allow ALL TCP/UDP/ICMP4 where src is the security group
   - Allow TCP to 37017 from 0.0.0.0/0 (yuk!)

The last rule is required as juju will use the _public_ (shadow) IP instead of the internal IP when setting up the replica sets for mongo.

### ARM64 controller + AMD64 machine 

```console
# Bootstrap arm64 and add amd64 machine
$ juju bootstrap manual/ubuntu@arm-machine-IP --debug --agent-version=2.8.0
...
10:26:39 INFO  juju.cmd.juju.commands bootstrap.go:832 combined bootstrap constraints:
10:26:39 DEBUG juju.environs.bootstrap bootstrap.go:308 model "controller" supports application/machine networks: false
10:26:39 DEBUG juju.environs.bootstrap bootstrap.go:310 network management by juju enabled: true
10:26:39 DEBUG juju.environs.bootstrap bootstrap.go:336 auto-selecting bootstrap series "bionic" <--- look for this 
10:26:39 DEBUG juju.environs.bootstrap bootstrap.go:340 auto-selecting bootstrap arch "arm64"  <--- look for this

# Then add the amd machine and ensure that it comes up online as expected
$ juju add-machine ssh:ubuntu@amd-machine-IP --debug
```

### AMD64 controller + ARM64 machine 

Clean up the machines from the previous test by running `/sbin/remove-juju-services`. 
**You probably need to also run `rm -f /etc/systemd/system/juju*` to manually remove the systemd units**.

```console
# Bootstrap amd64 and add arm64 machine
$ juju bootstrap manual/ubuntu@amd-machine-IP --debug --agent-version=2.8.0
...
10:26:39 INFO  juju.cmd.juju.commands bootstrap.go:832 combined bootstrap constraints:
10:26:39 DEBUG juju.environs.bootstrap bootstrap.go:308 model "controller" supports application/machine networks: false
10:26:39 DEBUG juju.environs.bootstrap bootstrap.go:310 network management by juju enabled: true
10:26:39 DEBUG juju.environs.bootstrap bootstrap.go:336 auto-selecting bootstrap series "bionic" <--- look for this 
10:26:39 DEBUG juju.environs.bootstrap bootstrap.go:340 auto-selecting bootstrap arch "amd64"  <--- look for this

# Then add the arm machine and ensure that it comes up online as expected
$ juju add-machine ssh:ubuntu@arm-machine-IP --debug
```

## Bug reference
https://bugs.launchpad.net/juju/+bug/1880422

### DIFF
--- a/environs/bootstrap/bootstrap.go
+++ b/environs/bootstrap/bootstrap.go
@@ -321,6 +321,30 @@ func bootstrapIAAS(
 		}
 	}
 
+	// If the provider allows advance discovery of the series and hw
+	// characteristics of the instance we are about to bootstrap, use this
+	// information to backfill in any missing series and/or arch contstraints.
+	if detector, supported := environ.(environs.HardwareCharacteristicsDetector); supported {
+		detectedSeries, err := detector.DetectSeries()
+		if err != nil {
+			return errors.Trace(err)
+		}
+		detectedHW, err := detector.DetectHardware()
+		if err != nil {
+			return errors.Trace(err)
+		}
+
+		if args.BootstrapSeries == "" && detectedSeries != "" {
+			args.BootstrapSeries = detectedSeries
+			logger.Debugf("auto-selecting bootstrap series %q", args.BootstrapSeries)
+		}
+		if args.BootstrapConstraints.Arch == nil && args.ModelConstraints.Arch == nil && detectedHW.Arch != nil {
+			arch := *detectedHW.Arch
+			args.BootstrapConstraints.Arch = &arch
+			logger.Debugf("auto-selecting bootstrap arch %q", arch)
+		}
+	}
+
 	requestedBootstrapSeries, err := coreseries.ValidateSeries(
 		args.SupportedBootstrapSeries,
 		args.BootstrapSeries,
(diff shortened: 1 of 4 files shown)


achilleasa:
$$merge$$

PR #11802, opened 2020-07-06
Update juju/testing and use AddExpr instead of AddRegex

hpidcock:
## Description of change

- Update juju/testing
- use AddExpr instead of AddRegex

## QA steps

Run state tests.

## Documentation changes

https://discourse.juju.is/t/multichecker-for-unit-tests/3320

## Bug reference

N/A


hpidcock:
> Can we get a discourse post (doesn't have to be that long; "X does this, use it!") about what this gives you in terms of testing... that way people are more likely to use the new methods.

https://discourse.juju.is/t/multichecker-for-unit-tests/3320

hpidcock:
$$merge$$

PR #11806, opened 2020-07-07
Allow dump-model to work with incomplete model data

wallyworld:
## Description of change

juju dump-model errors out if run just after adding a machine or relation, since the machine is still pending and there's no instance data yet, or the relation units are not yet in scope and so relation data is missing.

This PR adds a new IgnoreIncompleteModel bool to export config so that we still get a dumped model output if there's missing data. This is what we want as we only care about complete model data when doing migrations.

## QA steps

juju add-machine && juju dump-model
juju deploy foo && juju deploy bar && juju relate foo bar && juju dump-model

## Bug reference

https://bugs.launchpad.net/bugs/1849746
https://bugs.launchpad.net/bugs/1886369


wallyworld:
$$merge$$

PR #11811, opened 2020-07-09
Use rsync to copy acceptancetests *.py to exclude gitignored local fi…

ycliuhw:
## Description of change

Exclude git ignored local files for *.py listing in acceptancetests folder;


ycliuhw:
$$merge$$

PR #11815, opened 2020-07-09
Charm origin

SimonRichardson:
## Description of change

The following adds charm origin to state. This is independent of the
charmstore charm URL, hence why we want to own the origin for the
charm. 

Additionally, it adds the new changes but doesn't finish wiring them up.
Instead, it joins up the new charm origin, without parsing at every
stage. That will come later when we implement the download, refresh API.

Part of these changes includes the parsing and typing of a charm
channel. It conforms to the snapcraft channels docs[1] and collapses
where possible to be consistent with snap.

Both migration import and export is wired up, but again these are place
holders for when more data might be needed.

Additional work will be required:

 - Adding charm origin to show-application
 - Migrating state charm url and charm channel to charm origin.
   - Charm URL can be quite easily moved to charm origin ID.

1. https://snapcraft.io/docs/channels

## Issues:

 1. CharmURL **must** have a schema, we'll need to fix this in follow up
     PRs, as this is a right pain for upgrades and deployments.
 2. All charms **must** have a charm URL, failure to do so, will cause errors

## QA steps

```sh
juju bootstrap lxd test
juju deploy cs:~jameinel/ubuntu-lite-7
```

Access Mongo to see that the data is correctly stored.


apiserver/facades/client/application/application.go
```diff
@@ -749,6 +797,39 @@ func deployApplication(
...
+	case origin == nil || origin.Source == "" || origin.Source == "unknown":
+		return corecharm.Origin{Source: corecharm.Unknown}, nil
+	case origin.Source == "local":
+		return corecharm.Origin{Source: corecharm.Local}, nil
+	case origin.Source == "charm-store":
+		return corecharm.Origin{Source: corecharm.CharmStore}, nil
+	}
+
+	// Charmhub code can not tell us about the exact charm we want to install.
+	if origin.Revision == nil && origin.Channel == nil {
```
hmlanigan, line 811: A revision can never be nil unless it's unknown.  All 3 types must have a revision.

cmd/juju/application/deploy.go
```diff
@@ -1685,6 +1690,29 @@ func (c *DeployCommand) charmStoreCharm() (deployFn, error) {
...
+	switch url.Schema {
+	case "cs":
+		return application.CharmOrigin{
+			Source: application.OriginCharmStore,
+		}
+	case "local":
+		return application.CharmOrigin{
+			Source: application.OriginLocal,
+		}
+	default:
```
hmlanigan, line 1709: Where does application.OriginCharmHub come into play here?  Or are we waiting for charm.URL to be updated with charmhub?  My understanding was that for deploy, you must specify "cs:" for charm store, a relative path name is a local charm and everything else is a charm hub charm.  Though I know of no specifier for a charm hub charm.

core/charm/channel.go
```diff
@@ -0,0 +1,142 @@
...
+		if isRisk(p[0]) {
+			risk, branch = &p[0], &p[1]
+		} else {
+			track, risk = &p[0], &p[1]
+		}
+	case 3:
+		track, risk, branch = &p[0], &p[1], &p[2]
+	default:
+		return Channel{}, errors.Errorf("channel is malformed and has to many components %q", s)
+	}
```
hmlanigan, line 85: AFAIK, you can't have a track without a risk. I'm wondering if "https://snapcraft.io/docs/channels" is up to date. It seems to indicate that you can have a risk without a track, however I've never seen that in practice. If you specify a risk without a track, it means the track is "latest". I do not know if "latest" is a snap cmd output thing.

core/charm/channel_test.go
```diff
@@ -0,0 +1,120 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package charm_test
+
+import (
+	"github.com/juju/juju/core/charm"
```
hmlanigan, line 7: 2 stanzas

api/application/client.go
```diff
@@ -65,11 +65,45 @@ func (c *Client) ModelUUID() string {
...
+	// OriginLocal represents a local charm.
+	OriginLocal CharmOriginSource = "local"
+	// OriginCharmStore represents a charm from the now old charmstore.
+	OriginCharmStore CharmOriginSource = "charm-store"
+	// OriginCharmhub represents a charm from the new charmhub.
+	OriginCharmhub CharmOriginSource = "charmhub"
+	// OriginUnknown represents that we don't know where this charm came from.
+	// Either the charm was migrated up from an older version of Juju or we
+	//  didn't have enough information when we set the charm.
+	OriginUnknown CharmOriginSource = "unknown"
```
hmlanigan, line 85: Is this possible?  If it doesn't have "cs:" or "local:", then it's charm-hub.

apiserver/facades/client/application/application.go
```diff
@@ -401,6 +416,34 @@ func (api *APIv6) Deploy(args params.ApplicationsDeployV6) (params.ErrorResults,
...
+// using the specified placement directives.
+// V12 deploy did not CharmOrigin, so pass through an unknown source.
+func (api *APIv12) Deploy(args params.ApplicationsDeployV12) (params.ErrorResults, error) {
+	var newArgs params.ApplicationsDeploy
+	for _, value := range args.Applications {
+		newArgs.Applications = append(newArgs.Applications, params.ApplicationDeploy{
+			ApplicationName:  value.ApplicationName,
+			Series:           value.Series,
+			CharmURL:         value.CharmURL,
+			CharmOrigin:      &params.CharmOrigin{Source: "unknown"},
```
hmlanigan, line 429: For older facade versions, the charm origin source will be charm-store or local.  Can we derive the value from the CharmURL?

apiserver/facades/client/application/application.go
```diff
@@ -749,6 +797,39 @@ func deployApplication(
...
 
+func convertCharmOrigin(origin *params.CharmOrigin) (corecharm.Origin, error) {
+	switch {
+	case origin == nil || origin.Source == "" || origin.Source == "unknown":
+		return corecharm.Origin{Source: corecharm.Unknown}, nil
+	case origin.Source == "local":
+		return corecharm.Origin{Source: corecharm.Local}, nil
+	case origin.Source == "charm-store":
+		return corecharm.Origin{Source: corecharm.CharmStore}, nil
+	}
```
hmlanigan, line 808: charm hub not included?

state/charm.go
```diff
@@ -47,6 +47,26 @@ func (m MacaroonCache) Get(u *charm.URL) (macaroon.Slice, error) {
...
+// charm was installed from (charmhub, charmstore, local) and any additional
+// information we can utilise when making modelling decisions for upgrading or
+// changing.
+type CharmOrigin struct {
+	Source   string   `bson:"source"`
+	ID       string   `bson:"id"`
+	Hash     string   `bson:"hash"`
+	Revision *int     `bson:"revision,omitempty"`
+	Channel  *Channel `bson:"channel,omitempty"`
+}
```
hmlanigan, line 68: Did you decide against InstanceID, from the doc?

state/charm.go
```diff
@@ -47,6 +47,26 @@ func (m MacaroonCache) Get(u *charm.URL) (macaroon.Slice, error) {
...
+// Channel identifies and describes completely a store channel.
+// TODO (stickupkid): Add architecture.
+type Channel struct {
+	Track  string `bson:"track,omitempty"`
+	Risk   string `bson:"risk"`
+	Branch string `bson:"branch,omitempty"`
+}
+
+// CharmOrigin holds the original source of a charm. Information about where the
+// charm was installed from (charmhub, charmstore, local) and any additional
```
hmlanigan, line 59: If we're modeling charm origins, for all 3 types, isn't data necessary for the store and local missing?

api/application/client.go
```diff
@@ -65,11 +65,45 @@ func (c *Client) ModelUUID() string {
...
+	return string(c)
+}
+
+const (
+	// OriginLocal represents a local charm.
+	OriginLocal CharmOriginSource = "local"
+	// OriginCharmStore represents a charm from the now old charmstore.
+	OriginCharmStore CharmOriginSource = "charm-store"
+	// OriginCharmhub represents a charm from the new charmhub.
+	OriginCharmhub CharmOriginSource = "charmhub"
```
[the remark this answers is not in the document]
SimonRichardson: Just the way it was written in the spec and in the go code. Charmhub Vs CharmStore. I don't mind either way...

apiserver/facades/client/application/application.go
SimonRichardson: Fair

cmd/juju/application/deploy.go
SimonRichardson: Not sure yet, I had not wired that in yet. This was to ensure that the charm origin was wired up correctly to the db with the migrations.

core/charm/channel.go
SimonRichardson: So you can, but it's normalised to have a track.

api/application/client.go
SimonRichardson: Once we flesh this out for deployment and upgrading, unknown should be only a source of errors.

apiserver/facades/client/application/application.go
SimonRichardson: Unknown should be inferred as an error once we wire up charm url

state/charm.go
SimonRichardson: Yeah no need to store

SimonRichardson:
> I'm wondering if this should go farther and moving charmurl into the charm origin.

Yeah, but that should be done in another PR as it's quite bit of change and would make this harder to grok

core/charm/channel.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+	}
+	return false
+}
+
+// Channel identifies and describes completely a store channel.
+type Channel struct {
+	Track  string
+	Risk   Risk
+	Branch string
```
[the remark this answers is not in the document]
SimonRichardson: We don't lose anything adding this, yet we gain the ability to deploy charms with branches without changing any code.

core/charm/channel_test.go
SimonRichardson: I hate this with a passion!

apiserver/facades/client/application/application.go
SimonRichardson: Yes it is, see further down...

hmlanigan:
QA with develop juju snap client, added charm-origin as anticipated.

hmlanigan:
QA: migration from develop model to qa controller failed: 
sevel  friday      localhost/localhost  2.9-beta1.1  unsupported  12:49:52Z  migrating: aborted, removing model from target controller: model data transfer failed, failed to import model into target controller: runtime error: invalid memory address or nil pointer dereference

hmlanigan:
Migration now working.  Local charms have local charm origin sources

hmlanigan:
Haven't reproduced:  local charm thinking it's a charm-store charm in the db.

hmlanigan:
Migration working dependent on unrelated fix #11883

SimonRichardson:
$$merge$$

PR #11816, opened 2020-07-09
info output for Tags from info response entity []category

hmlanigan:
## Description of change

New information about where to find tags data for juju info output has been provided.  Rather than in the charm metadata, check the entity for category in the response data.  This is a good change as bundles do not have metadata.
## QA steps
```console
$ juju bootstrap localhost --build-agent
$ juju info wordpress | grep tags
tags: blog, applications

# bundles look slightly different
$ juju info osm | grep tags
tags: networking
```


hmlanigan:
Force push to fix the schema.json.

hmlanigan:
$$merge$$

PR #11821, opened 2020-07-10
Use different registered app for azure credentials; allow use of custom resource groups

wallyworld:
## Description of change

There's some issues bootstrapping to Azure.
The first is that the "Juju CLI" app in the canonical.com AD appears to have been setup as single tenant only. This means that depending on your subscription tenant, you could not bootstrap since it was not possible to create a service principal off this app. In my case, I could bootstrap but autoload-credentials failed.

The fix is to create a new "Juju" app in the canonical.com AD and have it be multi-tenant. The Juju change is to update the app id which is used to create the user's service principal. A drive by fix autoload-credentials was done to replace " " in detected credential names with "_".

The second issue is that for some accounts, resource groups cannot be created by policy. So we add a new Azure specific model config "resource-group-name". This will use an existing resource group. To make it work, Juju cannot be allowed to create a "default" model at bootstrap or else the same resource group will be used for both. So a new --no-default-model arg was added to bootstrap. We want to get rid of the default model eventually anyway.

There's a problem though - since we cannot tag these "read only" resource groups, destroy-controller cannot see them, so any models which use a custom group name will not have resources cleaned up. This is not easily fixed. It will be a known issue if this type of deployment is used. The user will need to destroy all models and then the controller.

## QA steps

install the az cli tool (this will be there anyway if azure used previously)
az logout
az login
juju autoload-credentials
(choose interactive method)

There should be a credential created allowing you to juju bootstrap to azure.

To test the resource group, set up a resource group "juju-test" via the azure dashboard, then

juju bootstrap azure -config resource-group-name=juju-test --no-default-model

The juju-test group should have the controller machine etc in it.
Create a new group juju-test2. Add a model

juju add-model test --config resource-group-name=juju-test2
juju add-machine

The juju-test2 group should have machine and some network items.

juju destroy-model test

The juju-test2 group will have everything deleted but the group will remain behind

## Bug reference

https://bugs.launchpad.net/bugs/1885557
https://bugs.launchpad.net/bugs/1869939

### DIFF
--- a/provider/azure/environ.go
+++ b/provider/azure/environ.go
@@ -103,8 +103,8 @@ type azureEnviron struct {
 	// subscription that corresponds to the environment.
 	resourceGroup string
 
-	// envName is the name of the environment.
-	envName string
+	// modelName is the name of the model.
+	modelName string
 
 	// authorizer is the authorizer we use for Azure.
 	authorizer *cloudSpecAuth
@@ -154,9 +154,13 @@ func newEnviron(
 		return nil, errors.Trace(err)
 	}
 
-	modelTag := names.NewModelTag(cfg.UUID())
-	env.resourceGroup = resourceGroupName(modelTag, cfg.Name())
-	env.envName = cfg.Name()
+	env.resourceGroup = env.config.resourceGroupName
+	// If no user specified resource group, make one from the model UUID.
+	if env.resourceGroup == "" {
+		modelTag := names.NewModelTag(cfg.UUID())
+		env.resourceGroup = resourceGroupName(modelTag, cfg.Name())
+	}
+	env.modelName = cfg.Name()
 
 	// We need a deterministic storage account name, so that we can
 	// defer creation of the storage account to the VM deployment,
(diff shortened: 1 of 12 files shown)


wallyworld:
$$merge$$

provider/azure/environ.go
```diff
@@ -1743,6 +1761,41 @@ func (env *azureEnviron) deleteResourceGroup(ctx context.ProviderCallContext, sd
 	return nil
 }
 
+func (env *azureEnviron) deleteResourcesInGroup(ctx context.ProviderCallContext, sdkCtx stdcontext.Context, resourceGroup string) error {
```
ycliuhw, line 1764: should we delete for resources with model tag?

wallyworld: This is the same as the regular behaviour for model destroy, just without the deletion of the resource group itself. All resources for a model are in the resource group.

wallyworld: There's issues with deleting just resources with a given model tag - the Azure query APi is very limited in what you can use as an expression, eg you can't filter on tag name and value and resource group or even resource type. So we can't easily identify all the resources for a given model. e will document that the admin needs to create one resource group per model.

wallyworld:
$$merge$$

PR #11822, opened 2020-07-10
Remove phrase "invalid URL" of datasource error messages.

hmlanigan:
## Description of change

It can be misleading to put a blanket "invalid URL" message on any error from the http client, remove.  Sometimes a not found can be an invalid URL, however other errors can be surface this way such as Forbidden.

Drive by: remove unused function arg and fix code for IDE warning that error from method is ignored.

## QA steps

```console
$ microstack.openstack service create --name product-stream  product-streams
$ microstack.openstack endpoint create --region RegionOne product-streams public http://some.bad.url.com/simplestreams
$ juju bootstrap microstack testme --debug --show-log

# output should include:
DEBUG juju.environs.bootstrap bootstrap.go:896 ignoring image metadata in keystone catalog: "http://some.bad.url.com/simplestreams/streams/v1/index.json" not found
# not:
DEBUG juju.environs.bootstrap bootstrap.go:877 ignoring image metadata in keystone catalog: invalid URL "http://some.bad.url.com/simplestreams/streams/v1/index.json" not found

# when finished, remove the endpoint, or your microstack cloud might not bootstrap again.
```

## Bug

Addresses confusion in: https://bugs.launchpad.net/juju/+bug/1885145 but does not fix it.

### DIFF
--- a/environs/gui/simplestreams_test.go
+++ b/environs/gui/simplestreams_test.go
@@ -162,7 +162,7 @@ var fetchMetadataTests = []struct {
 	about:         "error: stream file not found",
 	stream:        "no-such",
 	jujuVersion:   "2.0.0",
-	expectedError: `error fetching simplestreams metadata: cannot read product data, "test:/streams/v1/com.canonical.streams-no-such-gui.json" not found`,
+	expectedError: `error fetching simplestreams metadata: cannot read product data: "test:/streams/v1/com.canonical.streams-no-such-gui.json" not found`,
 }}
 
 func (s *simplestreamsSuite) TestFetchMetadata(c *gc.C) {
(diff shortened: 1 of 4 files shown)


jhobbs:
I'm not sure removing the error message completely is the right approach. It seems like a better plan would be to surface the real error to the end user via ERROR level logging. We shouldn't have to enable TRACE level logging to find out why a connection to a service required for bootstrapping failed.

hmlanigan:
@jhobbs, the error message is not being removed, part of it is being reworded.  

Surfacing as an Error could be misleading as well.  Juju checks for 6 different files using the product-streams endpoint.  We only need one of those to be successful.   Due to code organization, I'll have to check how easy it is to surface as actual errors if none are found.

hmlanigan:
$$merge$$

hmlanigan:
unrelated fail in uniterSuite.TestWatch
$$merge$$

PR #11823, opened 2020-07-13
Fix UpgradeSeriesAPI.GetMachine() where we ignored an error

wallyworld:
## Description of change

There was a customer outage due to a nil pointer deference. Turns out we were ignoring an error, which would be a "not found" error for a unit. The unit agent would normally restart and then terminate if its unit is not longer there, but instead due to the nil pointer the controller got wedged as the api call never returned and the controller workers got stuck.

## QA steps

just a simple bootstrap and deploy 

## Bug reference

https://bugs.launchpad.net/juju/+bug/1887327


wallyworld:
$$merge$$

wallyworld:
!!build!!

wallyworld:
!!build!!

PR #11831, opened 2020-07-14
[dev] Implement fallback subnet discovery mechanism for older LXD versions

achilleasa:
## Description of change

Unfortunately, older LXD versions (e.g. the ones shipping with xenial
and bionic) do not come with the network_state extension enabled. As a
result, subnet discovery for LXD substrates would fail.

This PR introduces a fallback method for reconstructing the subnet
details by introspecting the devices bridged into the controller
container and discovering the host network that they are bridged to.
This method works as subnet discovery is performed after the controller
has been bootstrapped.

The caveat of this method is that we fail to detect any subnet on the
LXD host that is not bridged into the controller container. This is
probably an acceptable tradeoff given that we are always able to detect
the subnet for the lxd bridged (regardless of whether it's managed by
LXD or not).

Finally, juju will emit a warning when using this fallback and suggest
that the operator upgrades their LXD version and runs `juju
reload-spaces` to get full subnet discovery.

## QA steps

Bootstrap juju on lxd 4 (eoan/focal) and lxd 3 (bionic/xenial). Then the output
of `juju subnets` should look as follows:

```console
# For lxd 3 cloud
$ juju subnets
subnets:
  10.108.67.0/24:
    type: ipv4
    provider-id: subnet-lxdbr0-10.108.67.0/24
    provider-network-id: net-lxdbr0
    status: in-use
    space: alpha
    zones: []

# For lxd 4 cloud (YMMV on this output depending on how the host is configured)
$ juju subnets
subnets:
  10.55.158.0/24:
    type: ipv4
    provider-id: subnet-lxdbr0-10.55.158.0/24
    provider-network-id: net-lxdbr0
    status: in-use
    space: alpha
    zones: []
  172.17.0.0/16:
    type: ipv4
    provider-id: subnet-docker0-172.17.0.0/16
    provider-network-id: net-docker0
    status: in-use
    space: alpha
    zones: []
  192.168.0.0/24:
    type: ipv4
    provider-id: subnet-wlx503eaa910eaa-192.168.0.0/24
    provider-network-id: net-wlx503eaa910eaa
    status: in-use
    space: alpha
    zones: []
```

### DIFF
--- a/provider/lxd/environ_network.go
+++ b/provider/lxd/environ_network.go
@@ -59,6 +59,13 @@ func (e *environ) Subnets(ctx context.ProviderCallContext, inst instance.Id, sub
 	for _, networkName := range networkNames {
 		state, err := srv.GetNetworkState(networkName)
 		if err != nil {
+			// Unfortunately, LXD on bionic and earlier does not
+			// support the network_state extension out of the box
+			// so this call will fail. If that's the case then
+			// use a fallback method for detecting subnets.
+			if isErrMissingAPIExtension(err, "network_state") {
+				return e.subnetDetectionFallback(srv, inst, keepList)
+			}
 			return nil, errors.Annotatef(err, "querying lxd server for state of network %q", networkName)
 		}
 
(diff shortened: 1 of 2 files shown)


achilleasa:
$$merge$$

PR #11834, opened 2020-07-15
Charmhub URL model-config

SimonRichardson:
## Description of change

This wires up the charmhub url to the model config. It currently makes
a new client every time the method is called, which is not ideal. If
performance does become a problem, we can cache the client for a given
charmhub URL. The operator isn't going to be changing that URL very
often, we should be fine having them in a `map[string]Client` type (with
mutex coordination).

The implementation is rather straight forward, with the exception of
adding the charmhub-url config to the immutable slice, prevent further
changes. There is an ongoing discussion about how we correctly model this,
but considering we can ADD mutation at a later date, this PR should be
enough to get us off to the races.

### Question

Should we hide this behind a feature flag?

### Success

```sh
juju bootstrap lxd test
juju find wordpress
```

### Failure

```sh
juju bootstrap lxd test
juju add-model test2 --config charmhub-url=http://google.com
juju find wordpress
```

Should fail when calling find.

### Validation failure

```sh
juju bootstrap lxd test
juju add-model test2 --config charmhub-url="http:// google.com"
```

## Documentation changes

Need to document the charmhub URL API in the model-config and
bootstrapping changes in our official documentation.


hmlanigan:
I tested using https://api.staging.snapcraft.io/ and find still worked... data's the same though.  And got the expected error with url that does not exist.

hmlanigan:
@SimonRichardson I wonder if we can cache the client for bundle deploy?  Might be too much work though.  In debugging other things, I did noticed that sometimes the go http client reuses connections.  When 6 new clients were made in quick succession.

environs/config/config_test.go
```diff
@@ -582,6 +583,19 @@ var configTests = []configTest{
...
+		about:       "Valid charmhub api url",
+		useDefaults: config.UseDefaults,
+		attrs: minimalConfigAttrs.Merge(testing.Attrs{
+			"charmhub-url": "http://test.com",
+		}),
+	}, {
+		about:       "Invalid charmhub api url",
+		useDefaults: config.UseDefaults,
+		attrs: minimalConfigAttrs.Merge(testing.Attrs{
+			"charmhub-url": "http://t est.com",
```
hmlanigan, line 596: suggest adding a test for random words or word strings without the http prefix

hmlanigan:
$ juju add-model test2 --config charmhub-url=http://google.com
Added 'test2' model on localhost/localhost with credential 'localhost' for user 'admin'
$ juju find wordpress
ERROR expected "application/json" contentType from server: text/html; charset=UTF-8
^^^^ This error message needs improving.

hmlanigan:
No error setting model-default value to empty string, though the change is not made which is good.
$ juju model-defaults charmhub-url=""
$ juju model-defaults charmhub-url
Attribute     Default                   Controller
charmhub-url  https://api.snapcraft.io  ""

SimonRichardson:
@hmlanigan I'll do a follow up for `model-defaults` and `migration`

SimonRichardson:
$$merge$$

PR #11835, opened 2020-07-15
Allow the user to specify formats

SimonRichardson:
## Description of change

Juju has a strong pattern of outputting different formats from each
command to allow the user to script against the CLI. The following
updates both find and info to allow this to happen.

The controversial thing about this change is naming the default output,
as for info the output is some sort of hybrid yaml that isn't yaml and
the default format for find is tabular. To be consistent I've named the
human (naming is hard).

The code is easy other than the warning logger to enable the commands to
output a warning if the result from the API isn't parsable. This can
happen if the charm creator fat fingers something and we don't want to
bring down the find/info commands to the user because of this. We'll
ignore the error (log it out) and present what we have to the user.

## QA steps

```sh
juju bootstrap lxd test
juju find wordpress
juju find --format=json wordpress
juju find --format=yaml wordpress
```

### DIFF
--- a/cmd/juju/charmhub/find.go
+++ b/cmd/juju/charmhub/find.go
@@ -5,6 +5,7 @@ package charmhub
 
 import (
 	"fmt"
+	"io"
 
 	"github.com/juju/cmd"
 	"github.com/juju/errors"
@@ -36,6 +37,8 @@ func NewFindCommand() cmd.Command {
 // findCommand supplies the "find" CLI command used to display find information.
 type findCommand struct {
 	modelcmd.ModelCommandBase
+	out        cmd.Output
+	warningLog Log
 
 	api FindCommandAPI
 
@@ -58,6 +61,11 @@ func (c *findCommand) Info() *cmd.Info {
 // It implements part of the cmd.Command interface.
 func (c *findCommand) SetFlags(f *gnuflag.FlagSet) {
 	c.ModelCommandBase.SetFlags(f)
+	c.out.AddFlags(f, "human", map[string]cmd.Formatter{
+		"yaml":  cmd.FormatYaml,
+		"json":  cmd.FormatJson,
+		"human": c.formatter,
+	})
 	// TODO (stickupkid): add the following:
 	// --narrow
 }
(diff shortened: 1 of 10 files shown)


SimonRichardson:
https://trello.com/c/TBqoveMP/2786-charmhub-cli-formatting-options

cmd/juju/charmhub/find.go
```diff
@@ -58,6 +61,11 @@ func (c *findCommand) Info() *cmd.Info {
 // It implements part of the cmd.Command interface.
 func (c *findCommand) SetFlags(f *gnuflag.FlagSet) {
 	c.ModelCommandBase.SetFlags(f)
+	c.out.AddFlags(f, "human", map[string]cmd.Formatter{
+		"yaml":  cmd.FormatYaml,
+		"json":  cmd.FormatJson,
+		"human": c.formatter,
```
hmlanigan, line 67: I like human, but suggest we add "tabular" here to keep in sync with other --format flags like in status

cmd/juju/charmhub/find_test.go
```diff
@@ -35,11 +36,45 @@ func (s *findSuite) TestRun(c *gc.C) {
...
+	c.Assert(cmdtesting.Stdout(ctx), gc.Equals, `
+- type: object
+  id: charmCHARMcharmCHARMcharmCHARM01
+  name: wordpress
+  publisher: Wordress Charmers
+  summary: WordPress is a full featured web blogging tool, this charm deploys it.
+  version: 1.0.3
+  series:
+  - bionic
+  storeurl: https://someurl.com/wordpress
```
hmlanigan, line 74: We should be consistent.  The json and yaml output are different labels right now.  Most structs have it as "store-url"

cmd/juju/charmhub/info.go
```diff
@@ -58,6 +62,11 @@ func (c *infoCommand) Info() *cmd.Info {
 // It implements part of the cmd.Command interface.
 func (c *infoCommand) SetFlags(f *gnuflag.FlagSet) {
 	c.ModelCommandBase.SetFlags(f)
+	c.out.AddFlags(f, "human", map[string]cmd.Formatter{
+		"yaml":  cmd.FormatYaml,
+		"json":  cmd.FormatJson,
+		"human": c.formatter,
```
hmlanigan, line 68: same as in find.

cmd/juju/charmhub/info_test.go
```diff
@@ -34,11 +36,97 @@ func (s *infoSuite) TestRun(c *gc.C) {
...
+    provides:
+      source: dummy-token
+    requires:
+      sink: dummy-token
+  subordinate: false
+  usedby:
+  - wordpress-everlast
+  - wordpress-jorge
+  - wordpress-site
+bundle: null
```
hmlanigan, line 116: can we get an omitempty on bundle and charm?  maybe needs to be a pointer too?

cmd/juju/charmhub/info_test.go
```diff
@@ -34,11 +36,97 @@ func (s *infoSuite) TestRun(c *gc.C) {
...
+type: charm
+id: charmCHARMcharmCHARMcharmCHARM01
+name: wordpress
+description: This will install and setup WordPress optimized to run in the cloud.
+publisher: Wordress Charmers
+summary: WordPress is a full featured web blogging tool, this charm deploys it.
+series:
+- bionic
+- xenial
+storeurl: https://someurl.com/wordpress
```
hmlanigan, line 75: same as with find

cmd/juju/charmhub/info_test.go
```diff
@@ -34,11 +36,97 @@ func (s *infoSuite) TestRun(c *gc.C) {
...
+tags:
+- app
+- seven
+charm:
+  config:
+    options:
```
hmlanigan, line 81: maybe we don't need a --config TODO with this change.

SimonRichardson: Agreed

SimonRichardson:
$$merge$$

PR #11838, opened 2020-07-17
Where possible write fstab entries using the device UUID

wallyworld:
## Description of change

Juju creates an fstab entry for each attached volume mount. It was using device path but this PR changed to use UUID instead.
Depending on the cloud, a volume may be mounted as either a disk or a partition. So we add logic to handle the partition appearing with a UUID after we may have already noticed the disk. The fstab updating will correctly replace any device path entry with the UUID one.
A nofail option is also added to the fstab emtry.

## QA steps

On aws, google, and azure, bootstrap and deploy a charm with storage
(aws using disks and azure and google use partitions so check all 3)
eg for aws
`juju deploy postgresql --storage pgdata=ebs,1G`

ssh into the machine and look at /etc/fstab, there should be a UUID mount, eg
```
# /srv/pgdata was on /dev/sdb1 during installation
UUID=2a3a8eca-b9bd-4c53-942d-085d14a56222 /srv/pgdata ext4 nofail,relatime,rw 0 0
```

reboot the machine and check the mount is recreated after.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1877380


storage/provider/managedfs.go
```diff
@@ -272,31 +273,96 @@ func extractMtabEntry(etcDir string, devicePath, mountPoint string) (string, err
...
 		}
 	}
 	if err := scanner.Err(); err != nil {
 		return errors.Trace(err)
 	}
 
-	// The entry will be written at the end of the fstab file.
-	if _, err = f.WriteString("\n" + entry + "\n"); err != nil {
-		return errors.Annotate(err, "writing /etc/fstab")
+	if !found {
```
ycliuhw, line 349: This is not a real notfound, it could be a dev path entry found or not found.
This `found` var is a bit confusing to read

storage/provider/managedfs.go
```diff
@@ -240,7 +241,7 @@ func mountFilesystem(run runCommandFunc, dirFuncs dirFuncs, devicePath, mountPoi
 	if mtabEntry == "" {
 		return nil
 	}
-	return addFstabEntry(etcDir, devicePath, mountPoint, mtabEntry)
+	return addFstabEntry(etcDir, devicePath, UUID, mountPoint, mtabEntry)
```
ycliuhw, line 244: we probably should only do this `once` to update the dev path entry for the mounted scenario.
Also `ensureFstabEntry` might be a better name now.

wallyworld: renamed found to addNewEntry

storage/provider/managedfs.go
wallyworld: renamed found to addNewEntry

storage/provider/managedfs.go
wallyworld: also, we do only update fstab on first attach and then again a second time only if the UUID changes from "" to "something".

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #11843, opened 2020-07-20
Ensure unavailable instance types are filtered out

wallyworld:
## Description of change

When Azure provider was selecting an instance type to use, the logic to filter out unavailable instance types was broken and it was including those which the user could not use. This caused bootstrap errors depending on region and credential used.

Also, the resource group name was to long - it included the model name and full model UUID when instead we only really need the first chunk of the UUID plus model name. This allows longer model names to be used without hitting the length limit on resource group names.

## QA steps

bootstrap 2.7.7 azure
upgrade to this PR
add a machine
check in the azure portal the resource group names of the models are the full UUID and the machine goes to the correct group
add a model and a machine
check that the short resource group name is now used

## Bug reference

https://bugs.launchpad.net/juju/+bug/1887887


wallyworld:
$$merge$$

hpidcock:
$$merge$$

PR #11844, opened 2020-07-20
Fix the smoke tests

SimonRichardson:
## Description of change

The `make go-build` has changed from 2.7 to 2.8 and the smoke tests
previously checked if there was any output.

The following updates the output from the integration tests to be much
more simplistic. There are now only two modes, the very terse mode,
which is great to just see from a very high-level overview if something
works or broke:

    ./main.sh smoke

Or using the verbose mode that aims to be chatty, but not overly chatty
like with the use of `set -eux`. The `x` part makes it hard to see
what's happening and should be only used for local debugging.

    ./main.sh -v smoke

With this simplistic model, we should be able to see what's failing
within jenkins, as we use `-v` as the default mode for all our testing.

## QA steps

```sh
cd tests && ./make smoke
```

### DIFF
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -115,7 +115,7 @@ bootstrap() {
             exit 1
         fi
 
-        add_model "${model}" "${provider}" "${bootstrapped_name}"
+        add_model "${model}" "${provider}" "${bootstrapped_name}" "${output}"
         name="${bootstrapped_name}"
     else
         echo "====> Bootstrapping juju ($(green "${version}:${provider}"))"
@@ -136,12 +136,13 @@ add_model() {
     model=${1}
     provider=${2}
     controller=${3}
+    output=${4}
 
     OUT=$(juju controllers --format=json | jq '.controllers | .["${bootstrapped_name}"] | .cloud' | grep "${provider}" || true)
     if [ -n "${OUT}" ]; then
-        juju add-model -c "${controller}" "${model}" "${provider}"
+        juju add-model -c "${controller}" "${model}" "${provider}" 2>&1 | OUTPUT "${output}"
     else
-        juju add-model -c "${controller}" "${model}"
+        juju add-model -c "${controller}" "${model}" 2>&1 | OUTPUT "${output}"
     fi
     echo "${model}" >> "${TEST_DIR}/models"
 }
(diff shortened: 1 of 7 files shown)


tests/main.sh
```diff
@@ -2,7 +2,7 @@
 [ -n "${GOPATH:-}" ] && export "PATH=${GOPATH}/bin:${PATH}"
 
 # Always ignore SC2230 ('which' is non-standard. Use builtin 'command -v' instead.)
-export SHELLCHECK_OPTS="-e SC2230 -e SC2039 -e SC2028 -e SC2002 -e SC2005"
+export SHELLCHECK_OPTS="-e SC2230 -e SC2039 -e SC2028 -e SC2002 -e SC2005 -e SC2181"
```
achilleasa, line 5: It would be great to do a quick review of the test suites and remove the comment lines that disable these checks.

tests/suites/smoke/build.sh
```diff
@@ -1,6 +1,6 @@
 run_build() {
     OUT=$(make go-build 2>&1 || true)
```
achilleasa, line 2: Maybe I didn't have enough ☕  yet but wouldn't this command always exit with status 0 if it fails due to the `|| true`?

tests/includes/juju.sh
```diff
@@ -184,8 +184,22 @@ juju_bootstrap() {
...
+            #shellcheck disable=SC2086
+            juju bootstrap ${series} --debug="${debug}" --build-agent=${BUILD_AGENT} "${provider}" "${name}" -d "${model}" "$@" > "${output}" 2>&1
+            ;;
+        2)
+            #shellcheck disable=SC2086
+            juju bootstrap ${series} --debug="${debug}" --build-agent=${BUILD_AGENT} "${provider}" "${name}" -d "${model}" "$@" 2>&1 | tee "${output}"
+            ;;
+        3)
+            #shellcheck disable=SC2086
+            juju bootstrap ${series} --debug="${debug}" --build-agent=${BUILD_AGENT} "${provider}" "${name}" -d "${model}" "$@" > "${output}" 2>&1
```
SimonRichardson, line 199: As we're using `-eux` adding `tee` is a bit sadistic...

tests/suites/smoke/build.sh
```diff
@@ -1,11 +1,5 @@
 run_build() {
-    OUT=$(make go-build 2>&1 || true)
-    if [ -n "${OUT}" ]; then
-        echo ""
-        echo "$(red 'Found some issues:')"
-        echo "\\n${OUT}"
-        exit 1
-    fi
+    make go-build 2>&1
```
SimonRichardson, line 2: Rather than be fancy, just let it fail!

tests/includes/run.sh
```diff
@@ -15,13 +15,16 @@ run() {
...
+  1)
+      $CMD "$@" > "${TEST_DIR}/${TEST_CURRENT}.log" 2>&1
+      ;;
   2)
+      $CMD "$@" 2>&1 | tee "${TEST_DIR}/${TEST_CURRENT}.log"
       cat "${TEST_DIR}/${TEST_CURRENT}.log"
       ;;
   3)
-      cat "${TEST_DIR}/${TEST_CURRENT}.log"
+      $CMD "$@" > "${TEST_DIR}/${TEST_CURRENT}.log" 2>&1
```
achilleasa, line 27: Do you also need a copy of L24 here? Otherwise this looks pretty much the same as case 1

tests/suites/smoke/build.sh
achilleasa: if you still want to be fancy:

```
OUT=$(make go-build 2>&1 || echo "BUILD FAILED")
if [ "$OUT" =~ (BUILD FAILED) ]; then
  ....
fi
```

SimonRichardson: I'll think about it

tests/includes/output.sh
```diff
@@ -0,0 +1,26 @@
...
+
+    output=${1}
+    shift
+
+    if [ -z "${output}" ] || [ "${VERBOSE}" -gt 1 ]; then
+        echo
+    fi
+
+    while read data; do
+        if [ -z "${output}" ]; then
```
achilleasa, line 12: It did take me a while to figure out what the various if blocks do here. It would be great if you could add some comments for others examining this part of the code in the future.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #11846, opened 2020-07-21
Have the apiserver watcher code translate unset application status.

howbazaar:
If an application status hasn't been explicitly set by a unit its status
is derived from the units themselves. This derivation code was moved into
the model cache, but the all watcher translation later itself was overlooked.

This branch uses the cached controller information to determine the application
status if it is Unset, like the status code does.

There were no tests around the all watcher code on the apiserver side,
so added some to test the conversion behaviour.

## QA steps

```sh
juju bootstrap lxd test
juju deploy ~jameinel/ubuntu-lite ubuntu
juju db # using juju-db plugin to get mongo prompt
db.statuses.find().pretty()
# check for _id: <uuid>:a#ubuntu
#        "status" : "unset",
# run pylibjuju example to look at the watcher
tox -e example -- examples/allwatcher.py 
# will see a line like:
['application', 'change', {'model-uuid': '11252a44-d946-4305-8292-a650cfa4b138', 'name': 'ubuntu', 'exposed': False, 'charm-url': 'cs:~jameinel/ubuntu-lite-7', 'owner-tag': '', 'life': 'alive', 'min-units': 0, 'constraints': {}, 'subordinate': False, 'status': {'current': 'waiting', 'message': 'waiting for machine', 'since': '2020-07-21T05:42:48.050629314Z', 'version': ''}, 'workload-version': ''}]
# confirm that status isn't 'unset'
```

### DIFF
--- a/apiserver/watcher.go
+++ b/apiserver/watcher.go
@@ -18,6 +18,7 @@ import (
 	"github.com/juju/juju/core/migration"
 	"github.com/juju/juju/core/multiwatcher"
 	"github.com/juju/juju/core/network"
+	"github.com/juju/juju/core/status"
 	"github.com/juju/juju/state"
 )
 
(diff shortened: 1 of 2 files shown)


howbazaar:
$$merge$$

PR #11848, opened 2020-07-21
Modernise model-defaults command

SimonRichardson:
## Description of change

This follows how model-config works, by allowing users to pass in
what they got, this makes it easier to understand what the command
requires.

The code is very similar to model-config, except model-defaults takes
into consideration region and controller.

Interestingly we don't consolidate when default and controller are the
same value, which seems a shame.

----

The help doc for model-defaults is wrong, the very fact that it states
-m as a possible option is a bit wild. I've removed that as it would
never work and it's confusing to users of juju.

I've also cleaned up if statements to make it easier to read.

----

Additionally whilst testing this I noticed that there is no validation for
model-defaults command and so you can shotgun your own foot when
attempting to add a new model (for instance).

### Piping

Using the new piping aspect (similar to model-config, which we should
use everywhere!)

```sh
juju bootstrap aws test
juju model-defaults --format=yaml test-mode | juju model-defaults eu-west-1 -
```

### Validation

```sh
juju bootstrap lxd test
juju model-defaults test-mode=bar
```

Results in an error!

### Help doc

Removal of `-m` in the help doc, that shouldn't be in there as it makes
no sense.

```sh
juju model-defaults -h
```


SimonRichardson:
https://trello.com/c/a4dtUulW/2791-model-defaults-command

cmd/juju/model/config.go
```diff
@@ -473,23 +483,23 @@ func (c *configCommand) handleIsKeyOfModel(attrs config.ConfigValues, ctx *cmd.C
...
-				attrs["cloudinit-userdata"] = value
-			}
-			return attrs, nil, true
+
+		return attrs, nil, false
+	}
+
+	// In tabular format, don't print "cloudinit-userdata" it can be very long,
+	// instead give instructions on how to print specifically.
+	if value, ok := attrs[config.CloudInitUserDataKey]; ok && c.out.Name() == "tabular" {
```
achilleasa, line 495: That's really neat! Are there any other model-config bits that are also long (e.g. a cert of something like that) and could be abbreviated? If that's the case you could make a "too-long-values" list of keys and apply this logic in a loop.

cmd/juju/model/defaults.go
```diff
@@ -37,24 +38,24 @@ supplied key to the supplied value. This can be repeated for multiple keys.
 You can also specify a yaml file containing key values.
 By default, the model is the current model.
 
-Model default configuration settings are specific to the cloud on which the model runs.
-If the controller host more then one cloud, the cloud (and optionally region) must be specified.
+Model default configuration settings are specific to the cloud on which the
+model runs.
```
achilleasa, line 42: Wondering if `s/model runs/model is deployed` or something similar would be better.

cmd/juju/model/defaults.go
```diff
@@ -151,6 +227,7 @@ func (c *defaultsCommand) SetFlags(f *gnuflag.FlagSet) {
 		"tabular": formatDefaultConfigTabular,
 	})
 	f.Var(cmd.NewAppendStringsValue(&c.reset), "reset", "Reset the provided comma delimited keys")
+	f.BoolVar(&c.skipImmutableErrors, "skip-immutable-errors", false, "Skip immutable errors when passing in the configurations")
```
achilleasa, line 230: I think the help text and perhaps the flag itself might need rephrasing to make it easier for end-users to understand what it actually does (I _bet_ people would be all over SO asking what an immutable error is).  Perhaps something like 'ignore-read-only-fields' "ignore changes to read-only fields instead of reporting an error" would be clearer.

cmd/juju/model/config.go
```diff
@@ -198,7 +203,7 @@ func (c *configCommand) SetFlags(f *gnuflag.FlagSet) {
 	})
 	f.Var(cmd.NewAppendStringsValue(&c.reset), "reset", "Reset the provided comma delimited keys")
 	f.BoolVar(&c.ignoreAgentVersion, "ignore-agent-version", false, "Skip the error when passing in the agent version configuration (deprecated)")
-	f.BoolVar(&c.skipImmutableErrors, "skip-immutable-errors", false, "Skip immutable errors when passing in the configurations")
+	f.BoolVar(&c.ignoreReadOnlyFields, "ignore-read-only-fields", false, "Ignore read only fields that might error during parsing of yaml")
```
achilleasa, line 206: Maybe "might cause errors to be emitted while processing yaml documents"?

SimonRichardson:
$$merge$$

PR #11862, opened 2020-07-28
Charmhub refresh api

SimonRichardson:
## Description of change

This contains the refresh client. It contains one method that you can
configure to do a refresh, an install and a composition of the 
configurations together.

With this abstraction, it becomes really easy to compose a series of
actions together. This could become really handy with things like large
bundles.

Example:

```go
wordpress, _ = InstallOne("wordpress", 16, "latest/stable", "ubuntu", "focal")
mysql, _ = RefreshOne("mysql", 245, "2.4/edge", "ubuntu", "disco")

config = RefreshMany(wordpress, mysql)
```

That way you can define that we want to refresh an existing mysql db and
then install wordpress as well.

#### Issues

 - There are a series of integration tests, but unfortunately, the install one doesn't work, or at least the response back is empty.
 - The `context` key in the request is always required, even if it's empty, which is a shame as we could use `omitempty`. There are some workarounds in the code to ensure that the configuration that we create always has a context and that it's never `nil`. 

#### Useful curl commands for testing the output.

```sh
curl -0 -X POST https://api.snapcraft.io/v2/charms/refresh \
-H 'Accept: application/json' \
-H 'Content-Type: application/json; charset=utf-8' \
--data-binary @- << EOF | jq
{
  "context": [
    {
      "instance-key": "meshuggah",
      "id": "charmCHARMcharmCHARMcharmCHARM01",
      "revision": 15,
      "platform": {"os": "ubuntu", "series": "focal", "architecture": "all"},
      "tracking-channel": "latest/stable",
      "refreshed-date": "2020-01-14T16:22:36.009402127Z"
    }
  ],
  "actions": [
    {
      "action": "refresh",
      "instance-key": "meshuggah",
      "id": "charmCHARMcharmCHARMcharmCHARM01",
      "revision": 15
    }
  ]
}
EOF
```

```sh
curl -0 -X POST https://api.snapcraft.io/v2/charms/refresh \
-H 'Accept: application/json' \
-H 'Content-Type: application/json; charset=utf-8' \
--data-binary @- << EOF | jq
{
  "context": [],
  "actions": [
    {
      "action": "install",
      "instance-key": "meshuggah",
      "id": "charmCHARMcharmCHARMcharmCHARM01",
      "revision": 15
    }
  ]
}
EOF
```

## QA steps

```sh
```


SimonRichardson:
https://trello.com/c/ezoefD6O/2805-add-refresh-client-api

charmhub/client.go
```diff
@@ -37,9 +39,9 @@ const (
 	CharmhubServerEntity  = "charms"
 )
 
-const (
+var (
 	userAgentKey   = "User-Agent"
-	userAgentValue = "Golang_CSClient/4.0"
+	userAgentValue = "Juju/" + version.Current.String()
```
SimonRichardson, line 44: I think the idea is to start sending more fine-grained details about what is asking for the refresh, so this may be shortlived in the end and we might require the user of the client to pass in the agent value.

charmhub/refresh_integration_test.go
```diff
@@ -0,0 +1,65 @@
...
+	gc "gopkg.in/check.v1"
+)
+
+type RefreshClientSuite struct {
+	testing.IsolationSuite
+}
+
+var _ = gc.Suite(&RefreshClientSuite{})
+
+func (s *RefreshClientSuite) TestLiveRefreshRequest(c *gc.C) {
```
hmlanigan, line 23: TestLiveRefreshRequestMany?

charmhub/client.go
SimonRichardson: I believe so.

charmhub/client.go
```diff
@@ -109,14 +123,20 @@ func NewClient(config Config) (*Client, error) {
...
 
+	refreshPath, err := base.Join("refresh")
+	if err != nil {
+		return nil, errors.Annotate(err, "constructing refresh path")
+	}
+
 	httpClient := DefaultHTTPTransport()
 	apiRequester := NewAPIRequester(httpClient)
-	restClient := NewHTTPRESTClient(apiRequester)
+	restClient := NewHTTPRESTClient(apiRequester, config.Headers)
```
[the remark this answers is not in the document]
SimonRichardson: So as we don't use the `http.DefaultClient` in this package and we won't actually pick up the proxy settings. It's only when we don't instantiate a new `&http.Client{}` that we will pick up the proxy settings. We would need to audit library/code to see every time someone created a new HTTP request to ensure that they didn't use the `http.DefaultClient`.

Given the fact that go advertises the default client to everybody in the stdlib was a really bad design flaw. That should have been either a non-pointer version or abstracted away that it forces you to create your own client if you wanted to do something more advanced.

charmhub/refresh.go
```diff
@@ -0,0 +1,267 @@
...
+
+const (
+	// InstallAction defines a install action.
+	InstallAction Action = "install"
+
+	// RefreshAction defines a refresh action.
+	RefreshAction Action = "refresh"
+
+	// RefreshAllAction defines a refresh all action.
+	RefreshAllAction Action = "refresh-all"
```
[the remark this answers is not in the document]
SimonRichardson: Agreed, I'll remove.

charmhub/refresh_integration_test.go
SimonRichardson: Sure.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #11863, opened 2020-07-29
[dev] Use endpoints instead of subnet IDS for representing open machine ports

achilleasa:
## Description of change

With the changes that got landed via #11842 the initial plan for the open-port in-endpoint work was to allow units to specify endpoints when opening ports and then have the controller translate the endpoints into a subnet ID by intersecting the space bindings for the charm application and the subnets available to the machine that a charm is deployed to.

As it later became apparent, that approach had two important caveats:
1) It would require a significant amount of work to keep the `openPorts` collection documents in sync given that the endpoint -> subnet association may change in different ways (e.g. by re-binding to another space, or if the space topology changes). In the case of a space topology change, in order to update the open ports collection,  we would need to figure out the endpoint from the subnet ID given the previous space topology and then resolve a new subnet ID for the endpoint using the new topology.
2) By resolving endpoints to subnet IDs early on, we lose an important bit of information: whether the charm opened a port to **all** endpoints or to specific endpoints (a subset of the known endpoints that may also be the entire set of known endpoints). For example: if the application defines endpoints `x, y and z`, the commands: `open-port 1234/tcp` and `open-port 1234/tcp --endpoints x,y,z` (the proposed extension for endpoints) are not _identical_. While the effective set of opened ports **is** the same, the **first** command implies that the port should also be opened to **all endpoints that might show up in the future** (e.g. after upgrading the charm), whereas the latter would retain the open port rules for the specified endpoints only (i.e. the charm will need to explicitly open the ports to the new endpoints if needed).

This PR attempts to address these limitations by:
- changing the layout/format of the documents in the `openedPorts` collection. In particular instead of having a **per-machine, per subnet** document, the new format uses a single document per machine. The new document keeps track of the opened port ranges **by unit and by endpoint**. In addition, an endpoint can be left **empty** to indicate that a port range applies to **all** endpoints.  This change avoids the need to sync the document for binding/space topology changes as we can defer the resolution of the subnet CIDRs to the place where the ingress rules get generated (the firewall worker).

To make it easy to work with the new format, the PR provides two interfaces (and their concrete implementations) in the `state` package: the `MachinePortRanges` and the `UnitPortRanges`. The state-provided `MachinePortRanges` implementation is backed by the document from the `openedPorts` collection and provides methods for querying the open port ranges for a unit (`ByUnit` and `ForUnit`). The obtained `UnitPortRanges` instances provide methods for querying and manipulating the port ranges for an individual unit. They are implemented as views to the parent machine port ranges instance and record port change (open or close) requests to it. The `MachinePortRanges` interface provides the `Changes` method for getting back a `ModelOperation` for applying any required modifications to the `openedPorts` collection. The same method is also defined on the `UnitPortChanges` type but in this case the set of changes only includes the changes for a particular unit. For example:

```go
machPortRanges, _ := machine.OpenedPortRanges()
machPortRanges.ForUnit("foo/0").Open("endpoint", network.MustParsePortRange("1337/tcp"))
machPortRanges.ByUnit()["foo/0"].Open("endpoint", network.MustParsePortRange("8080/tcp"))
machPortRanges.ForUnit("bar/0").Open("endpoint", network.MustParsePortRange("9999/tcp"))

// Apply all at once
modelOp := machPortRanges.Changes()
state.ApplyOperation(modelOp)

// OR apply individually (each successful ApplyOperation removes the pending changes for the 
// unit from the parent machine port ranges instance)
state.ApplyOperation(machPortRanges.ForUnit("foo/0").Changes())
state.ApplyOperation(machPortRanges.ForUnit("bar/0").Changes())
err := machPortRanges.Changes().Build(0) // returns ErrNoOperations as all changes have been applied!
```

Pre-2.9 controllers never seem to have ever made use of the open-port-in-subnet feature as this is not exposed via the uniter APIs that allow units to open/close ports. Note that some of the other APIs (e.g. the firewaller ones) do allow the caller to provide a subnet ID. However, at the end of the day, they all get stored to the DB as a single document with an _empty subnet ID_.

To this end, the facade changes from this PR emulate this behavior. Separate PRs will introduce new APIs for working with open ports in an endpoint-aware manner and the relevant workers will be updated to work with them as the other chunks of work begin to land.

Finally, the PR bumps the description package (see https://github.com/juju/description/pull/85) so we can use the new port format (which also transparently converts pre 2.9 descriptions to the new format) when exporting 2.9+ models and includes an upgrade step for converting pre-2.9 open port documents to the new format introduced by this PR.

For more information see the individual commit messages.

### With a 2.9 controller

```console
$ juju bootstrap lxd test-2.9 --no-gui
$ juju add-model model-29
$ juju deploy mysql
$ juju run --unit mysql/0 "open-port 1337/udp"

# Note: mysql will open 3306/tcp on its own
$ juju status --format oneline
- mysql/0: 10.55.158.86 (agent:idle, workload:active) 3306/tcp, 1337/udp

$ juju status --format oneline
- mysql/0: 10.55.158.86 (agent:idle, workload:active) 3306/tcp, 1337/udp
$ juju run --unit mysql/0 "opened-ports"
3306/tcp
1337/udp

# Connect to mongo and verify the contents of the 'openedPorts' collection
db>  db.openedPorts.find({},{"unit-port-ranges":1}).pretty()
{
        "_id" : "7f72ae21-9347-426c-846f-cb597150456e:0",
        "unit-port-ranges" : {
                "mysql/0" : {
                        "" : [
                                {
                                        "fromport" : 1337,
                                        "toport" : 1337,
                                        "protocol" : "udp"
                                },
                                {
                                        "fromport" : 3306,
                                        "toport" : 3306,
                                        "protocol" : "tcp"
                                }
                        ]
                }
        }
}
```

**Keep the controller up and running for the next steps**

### Upgrading a 2.8 controller

This test verifies that:
- A 2.8 model can use the modified port-related API methods of a 2.9 controller.
- The port documents get correctly converted to the new format after upgrading the model.

```console
$ /snap/bin/juju version
2.8.1-focal-amd64

$ /snap/bin/juju bootstrap lxd test-2.8 --no-gui
$ /snap/bin/juju add-model model-28
$ /snap/bin/juju deploy mysql
$ /snap/bin/juju run --unit mysql/0 "open-port 1337/udp"

# Upgrade the controller to 2.9 but do not upgrade model just yet

# Verify docs have been correctly converted
db>  db.openedPorts.find({},{"unit-port-ranges":1}).pretty()
{
        "_id" : "81e4b70e-488a-449f-80a3-ed9e19bc8cfa:0",
        "unit-port-ranges" : {
                "mysql/0" : {
                        "" : [
                                {
                                        "fromport" : 1337,
                                        "toport" : 1337,
                                        "protocol" : "udp"
                                },
                                {
                                        "fromport" : 3306,
                                        "toport" : 3306,
                                        "protocol" : "tcp"
                                }
                        ]
                }
        }
}

# Verify 2.8 model agent can still list opened ports
$ juju run --unit mysql/0 "opened-ports"
3306/tcp
1337/udp

# Upgrade model and verify everything is happy
$ juju upgrade-model
$ juju run --unit mysql/0 "opened-ports"
3306/tcp
1337/udp
```

### Migrating a 2.8 model to a 2.9 controller

This test verifies that the 2.9 controller correctly converts  the port list of the V1 machine description exported by a 2.8 controller  to the new port range format.

```console
$ /snap/bin/juju bootstrap lxd test-another-2.8 --no-gui
$ /snap/bin/juju add-model model-migrate-28
$ /snap/bin/juju deploy mysql
$ /snap/bin/juju run --unit mysql/0 "open-port 1337/udp"

# Commence migration and wait for it to complete
$ juju migrate model-migrate-28 test-2.9

# Check ports on 2.9 controller
$ juju switch test-2.9:admin/model-migrate-28
$  juju run --unit mysql/0 "opened-ports"
3306/tcp
1337/udp

db>  db.openedPorts.find({},{"unit-port-ranges":1}).pretty()
... # There should be an entry for the migrated model ID
{
        "_id" : "27359e6f-1305-4d8f-8f46-740661cd147e:0",
        "unit-port-ranges" : {
                "mysql/0" : {
                        "" : [
                                {
                                        "fromport" : 1337,
                                        "toport" : 1337,
                                        "protocol" : "udp"
                                },
                                {
                                        "fromport" : 3306,
                                        "toport" : 3306,
                                        "protocol" : "tcp"
                                }
                        ]
                }
        }
}
```

### Migrating a 2.9 model to a 2.9 controller

Continue from previous test! 
This test verifies that the 2.9 controller correctly parses the V2 machine description that uses the new port range format.

```console
# Fire up a new 2.9 controller
$ juju bootstrap lxd test-another-29 --no-gui

# Switch to the 2.9 model in the previous controller and verify its version
$ juju switch test-2.9:admin/model-29
$ juju status | grep model-29
model-29  test-2.9    lxd/default  2.9-beta1.1  unsupported  16:12:00+01:00

# Migrate 2.9 model to the new 2.9 controller and wait for migration to complete...
$ juju migrate model-29 test-another-29

$ juju switch test-another-29:admin/model-29
$ juju run --unit mysql/0 "opened-ports"
3306/tcp
1337/udp

db>  db.openedPorts.find({},{"unit-port-ranges":1}).pretty()
{
        "_id" : "7f72ae21-9347-426c-846f-cb597150456e:0",
        "unit-port-ranges" : {
                "mysql/0" : {
                        "" : [
                                {
                                        "fromport" : 3306,
                                        "toport" : 3306,
                                        "protocol" : "tcp"
                                },
                                {
                                        "fromport" : 1337,
                                        "toport" : 1337,
                                        "protocol" : "udp"
                                }
                        ]
                }
        }
}
```

## Expose application

This should be tried on a substrate that supports firewalls (e.g. aws)

```console
$ juju bootstrap aws test-aws

$ juju deploy mysql
$ juju run --unit mysql/0 "open-port 1337/udp"

$ juju expose mysql

# Verify that the right ingress rules have been added to the security group for the machine
# where the unit is running
```


state/upgrade/types.go
```diff
@@ -67,3 +67,21 @@ func (a OldAddress27) Upgrade(lookup network.SpaceInfos) (OldAddress27, error) {
...
+	Ports     []OldPortRangeDoc28 `bson:"ports"`
+	TxnRevno  int64               `bson:"txn-revno"`
+}
+
+// OldPortsDoc28 represents a port range entry document prior to the 2.9 schema changes.
+type OldPortRangeDoc28 struct {
+	UnitName string
+	FromPort int
+	ToPort   int
+	Protocol string
```
wallyworld, line 86: Should tag these

state/upgrades.go
```diff
@@ -2961,3 +2961,88 @@ func AddCharmhubToModelConfig(pool *StatePool) error {
...
+		defer closer()
+
+		var oldDocs []upgrade.OldPortsDoc28
+		err := col.Find(nil).All(&oldDocs)
+		if err != nil {
+			return errors.Trace(err)
+		}
+
+		// Roll up per-subnet docs (there should actually only be one
+		// per machine) and emit txn ops for removing them.
```
wallyworld, line 2986: Maybe a check and warning if there do happen to be > 1 doc per machine, even though it should not happen

api/firewaller/machine_test.go
```diff
@@ -161,3 +165,25 @@ func (s *machineSuite) TestIsManual(c *gc.C) {
 	c.Assert(answer, jc.IsTrue)
 
 }
+
+func mustOpenPortRanges(c *gc.C, st *state.State, u *state.Unit, endpointName string, portRanges []network.PortRange) {
+	unitPortRanges, portChangesFn, err := u.OpenedPortRanges()
```
howbazaar, line 170: Why have two values returned here? The usage of them looks quite disconnected.

Perhaps have the `portChangesFn` actually be a method on the `unitPortRanges`, something like `Operations`?

apiserver/facades/controller/firewaller/firewaller_test.go
```diff
@@ -104,19 +101,34 @@ func (s *firewallerSuite) TestGetAssignedMachine(c *gc.C) {
...
+
+func (s *firewallerSuite) mustOpenPorts(c *gc.C, unit *state.Unit, endpointName string, portRanges []network.PortRange) {
+	unitPortRanges, portChangesFn, err := unit.OpenedPortRanges()
+	c.Assert(err, jc.ErrorIsNil)
+
+	for _, pr := range portRanges {
+		unitPortRanges.Open(endpointName, pr)
+	}
+
+	c.Assert(s.State.ApplyOperation(portChangesFn()), jc.ErrorIsNil)
```
howbazaar, line 122: After reading code between here and the previous comment, perhaps a method called `Changes`?

state/machine_ports_ops_test.go
```diff
@@ -0,0 +1,346 @@
...
+import (
+	"github.com/juju/collections/set"
+	jc "github.com/juju/testing/checkers"
+	gc "gopkg.in/check.v1"
+
+	"github.com/juju/juju/core/network"
+)
+
+type MachinePortsOpsSuite struct {
+}
```
howbazaar, line 15: it is useful even using the testing.IsolationSuite just in case someone adds some logging later.

api/firewaller/machine_test.go
achilleasa: I was just lazy to add this because the correct implementation would need to change the machine_port_ops code to filter the registered open/close port modifications to apply by unit name (otherwise your change set could include the bits for other units as well).

Now that I think about it, it's not _that_ much code so I guess I can implement it today and update the tests as a separate commit.

state/upgrade/types.go
achilleasa: Should we always put bson tags in general? The original type in state before my changes did not (hence the odd lowercased key names in the `$set` queries). I retained this for the new type but should probably tag those as well to get better looking keys.

achilleasa: @wallyworld It looks like the same functionality for open/close/conflicts is tested multiple times (I counted at least 3 different suites in various places that test the same thing over and over) and I believe the rollup test to be adequate (I mean, the rollup logic is pretty trivial). Do you have any particular use case in mind that I may have missed?

That being said, I do have a card for converting the QA step from all these related PRs into functional tests (once all bits land) so I think we should be able to cover more things that way.

state/upgrades.go
achilleasa: Done

apiserver/facades/controller/firewaller/firewaller_test.go
achilleasa: Addressed in https://github.com/juju/juju/pull/11863/commits/e3ab5027dca6cd28af2070b0107ef831fdae1a6e

achilleasa:
!!build!! (TestStatusAllFormats)

state/upgrade/types.go
wallyworld: It was nothing in particular - just a (not always followed) convention that we should tag struct attributes for serialisation, even if the tags would match the default serialisation names. It means there's absolutely no ambiguity. But since this is just a type used for an upgrade step, it doesn't really matter so mych.

achilleasa:
Rebase and force-pushed to resolve conflicts

achilleasa:
$$merge$$

PR #11865, opened 2020-07-30
Add a juju repl CLI

wallyworld:
## Description of change

Add a new interactive Juju shell. It is invoked by running the juju command without any args.
The shell will include in the prompt the current controller/model/user. If there's no bootstrapped
controller, it will tell you to bootstrap one.
Running juju help still prints out the expected help text, but with an additional instruction on how to run the shell.
There's support for cmd history via ^R and the usual ^A, ^E, up, down keys etc. Todo is to add support for tab completion.

## QA steps

```sh
$ juju
Welcome to the Juju interactive shell.
Type "help" to see a list of available commands.
Type "q" or ^D or ^C to quit.

admin@ctrl:default$ status
Model    Controller  Cloud/Region       Version  SLA          Timestamp
default  ctrl        testlxd/localhost  2.8.2.1  unsupported  12:09:55+10:00

App      Version  Status  Scale  Charm    Store       Rev  OS      Notes
mariadb  10.1.40  active      1  mariadb  jujucharms    7  ubuntu  

Unit        Workload  Agent  Machine  Public address  Ports  Message
mariadb/0*  active    idle   0        10.115.246.142         ready

0        started  10.115.246.142  juju-946997-0  trusty      Running

admin@ctrl:default$ models
Controller: ctrl

Model       Cloud/Region       Type  Status     Machines  Units  Access  Last connection
controller  testlxd/localhost  lxd   available         1      -  admin   just now
default*    testlxd/localhost  lxd   available         1      1  admin   15 seconds ago
foo         testlxd/localhost  lxd   available         0      -  admin   2 hours ago

admin@ctrl:default$ switch foo
ctrl:admin/default -> ctrl:admin/foo
admin@ctrl:foo$ 
```

### DIFF
--- a/cmd/juju/commands/main.go
+++ b/cmd/juju/commands/main.go
@@ -13,6 +13,7 @@ import (
 	"text/template"
 
 	"github.com/juju/cmd"
+	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 	"github.com/juju/featureflag"
 	"github.com/juju/loggo"
@@ -142,6 +143,19 @@ func (m main) Run(args []string) int {
 		}
 	}
 
+	// See if we need to invoke the juju interactive shell.
+	// It is invoked by:
+	// $ juju
+	// We also run the repl command so that it prints help and exits
+	// if the user types:
+	// $ juju --help | -h | help
+	jujuArgs := set.NewStrings(args[1:]...)
+	helpArgs := set.NewStrings("help", "-h", "--help")
+	showHelp := jujuArgs.Intersection(helpArgs).Size() > 0
+	jujuArgs = jujuArgs.Difference(helpArgs)
+	if showHelp && len(args) == 1 || len(jujuArgs) == 0 {
+		return cmd.Main(newReplCommand(showHelp), ctx, nil)
+	}
 	jcmd := NewJujuCommand(ctx, jujuMsg)
 	return cmd.Main(jcmd, ctx, args[1:])
 }
--- a/cmd/juju/commands/repl.go
+++ b/cmd/juju/commands/repl.go
@@ -1,0 +1,228 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package commands
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"strings"
+
+	"github.com/chzyer/readline"
+	"github.com/juju/cmd"
+	"github.com/juju/errors"
+	"github.com/juju/gnuflag"
+	"github.com/juju/loggo"
+
+	jujucmd "github.com/juju/juju/cmd"
+	"github.com/juju/juju/cmd/modelcmd"
+	"github.com/juju/juju/jujuclient"
+)
+
+type replCommand struct {
+	cmd.CommandBase
+
+	store    jujuclient.ClientStore
+	showHelp bool
+
+	execJujuCommand func(cmd.Command, *cmd.Context, []string) int
+}
+
+func newReplCommand(showHelp bool) cmd.Command {
+	return &replCommand{
+		showHelp:        showHelp,
+		store:           jujuclient.NewFileClientStore(),
+		execJujuCommand: cmd.Main,
+	}
+}
+
+const replDoc = `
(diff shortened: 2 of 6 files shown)


apiserver/facades/schema.json
```diff
@@ -28215,7 +28215,7 @@
                             "$ref": "#/definitions/MachineNetworkConfigResults"
                         }
                     },
-                    "description": "GetContainerInterfaceInfo returns information to configure networking for a\ncontainer. It accepts container tags as arguments."
+                    "description": "GetContainerInterfaceInfo returns information to configure networking for a\ncontainer. It accepts container tags as arguments.\nTODO (manadart 2020-07-23): This method is not used and can be removed when\nnext this facade version is bumped.\nWe then don't need the parameterised prepareOrGet..."
```
howbazaar, line 28218: We should move the TODO statement so it doesn't show up in the schema.json.

cmd/juju/commands/main.go
```diff
@@ -142,6 +143,17 @@ func (m main) Run(args []string) int {
 		}
 	}
 
+	// See if we need to invoke the juju interactive shell.
+	// It is invoked by:
+	// $ juju
+	// $ juju --help | -h | help
```
howbazaar, line 149: I don't think we should be starting the repl shell if they ask for help.

cmd/juju/commands/repl.go
```diff
@@ -0,0 +1,211 @@
...
+`
+
+func (c *replCommand) Info() *cmd.Info {
+	return jujucmd.Info(&cmd.Info{
+		Name:    "juju",
+		Purpose: "Enter an interactive shell for running Juju commands",
+	})
+}
+
+func filterInput(r rune) (rune, bool) {
```
howbazaar, line 54: This function deserves a comment to explain the what and why.

cmd/juju/commands/repl.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+const replDoc = `
+When run without arguments, enter an interactive shell which can be
+used to run run any Juju command directly.
+Type q|Q or ^D or ^C to quit.
+Otherwise, the supported command usage is described below.
+`
+
+func (c *replCommand) Info() *cmd.Info {
```
howbazaar, line 47: Exported functions should have comments.

cmd/juju/commands/repl.go
```diff
@@ -0,0 +1,211 @@
...
+	if err != nil {
+		return errors.Trace(err)
+	}
+	for {
+		// loggo maintains global state so reset before each command.
+		loggo.ResetLogging()
+		_ = loggo.RegisterWriter(loggo.DefaultWriterName, defaultWriter)
+
+		jujuCmd := NewJujuCommand(ctx, "")
+		if c.showHelp {
```
howbazaar, line 113: I think the default prompt initial output should at the bare minimum tell the user they can type "help" for commands, and tell them how to exit.

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da
 	github.com/aws/aws-sdk-go v1.29.8
 	github.com/bmizerany/pat v0.0.0-20160217103242-c068ca2f0aac
+	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
```
howbazaar, line 18: I think we need to do a quick review of this library.

cmd/juju/commands/main.go
wallyworld: Ah, the comment was wrong. juju --help does not invoke the repl. That was a typo on my part

go.mod
wallyworld: It seems to do what it says on the tin when I looked at it - picks chars off stdin and buffers then into a line of text, with smart processing to handle navigation

cmd/juju/commands/repl.go
wallyworld: Good idea

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #11869, opened 2020-07-31
Hook up the deployer nested context.

howbazaar:
This work introduces the implementation of the unified machine
and unit agent for IAAS models. The deployer now runs the workers
for the units in the same process using a runner to run a primary
worker for each unit. This primary worker is a dependency engine
that runs a subset of the workers that the old unit agent ran.
There is no need for the units to now run upgrade workers as
that is all handled by the machine agent.

An upgrade step is added to disable all the service files running
for the old unit agents, and update the agent.conf file for the
machine agent to record the deployed units.

## QA steps

```sh
juju bootstrap lxd test
juju deploy cs:~jameinel/ubuntu-lite ubuntu
juju deploy filebeat
juju relate ubuntu filebeat
watch -c juju status --color
```
Wait until the units are deployed.
```sh
juju ssh 0
```
See that the file structure in `/var/lib/juju` is unchanged. Look at the `juju_engine_report` and observe that the `deployer` now shows the units as deployed, and the nested engine report is shown for each of the units.
Also note that `/etc/systemd/system` only has a machine agent, and no unit agents.

Upgrade steps testing
```sh
sudo snap refresh juju --channel 2.8
/snap/bin/juju bootstrap lxd upgrade --no-gui
juju deploy cs:~jameinel/ubuntu-lite ubuntu
juju deploy filebeat
juju relate ubuntu filebeat
watch -c juju status --color
```
Wait until deployed, then
```sh
# wait a bit for the controller to upgrade
juju upgrade-model --agent-version 2.9-beta1.1
```
then
```sh
juju ssh 0
```
Check `/etc/systemd/system` to observe that no unit agents exist, then run `juju_engine_report` to
observe that the unit workers are running in the machine agent.

## Documentation changes

We probably want to update the docs to remove references to the unit agents as separate entities for IAAS models.


cmd/jujud/agent/machine.go
```diff
@@ -1363,8 +1365,8 @@ func (a *MachineAgent) removeJujudSymlinks() (errs []error) {
 // running the tests and (2) get access to the *State used internally, so that
 // tests can be run without waiting for the 5s watcher refresh time to which we would
 // otherwise be restricted.
-var newDeployContext = func(st *apideployer.State, agentConfig agent.Config) deployer.Context {
-	return deployer.NewSimpleContext(agentConfig, st)
+var newDeployContext = func(config deployer.ContextConfig) (deployer.Context, error) {
```
tlm, line 1368: Silly question but is there a specific reason this needs to be a var and not just a package scoped func ?

worker/deployer/manifold.go
```diff
@@ -37,16 +46,51 @@ func Manifold(config ManifoldConfig) dependency.Manifold {
...
 	return w, nil
 }
+
+type apiShim struct {
+	st *apideployer.State
+}
+
+func (s *apiShim) Machine(tag names.MachineTag) (Machine, error) {
+	// Need to deal with typed nils.
+	machine, err := s.st.Machine(tag)
```
tlm, line 83: This can also get simplified down to `return s.st.Machine(tag)`

worker/deployer/manifold.go
```diff
@@ -37,16 +46,51 @@ func Manifold(config ManifoldConfig) dependency.Manifold {
...
+	if err != nil {
+		return nil, err
+	}
+	return machine, nil
+}
+
+func (s *apiShim) Unit(tag names.UnitTag) (Unit, error) {
+	unit, err := s.st.Unit(tag)
```
tlm, line 91: Same again.

worker/deployer/nested.go
```diff
@@ -50,10 +50,11 @@ type Logger interface {
...
 
 type nestedContext struct {
-	logger            Logger
-	agentConfig       agent.Config
-	baseUnitConfig    UnitAgentConfig
-	updateConfigValue func(string, string) error
+	logger Logger
+	agent  agent.Agent
+	// agentConfig is a snapshot of the current configuration.
+	agentConfig    agent.Config
```
tlm, line 56: Would it be worth naming this to `agentConfigSnapshot` based on the comment?

worker/deployer/manifold.go
howbazaar: Unfortunately it can't.

I hit this problem with the description package as well.

`s.st.Machine(tag)` returns a *state.Machine, and if the returned value is nil, it is a typed nil, meaning that the type of the interface is a *state.Machine and the value is nil, which means when you compare against nil with `shim.Machine(tag) == nil` the typed nil doesn't match. This is a trick for new players.

cmd/jujud/agent/machine.go
howbazaar: I'll double check, maybe not.

howbazaar: No reason at all. Removed the function and just pass in the package func where this was used.

howbazaar: I think in the past it changed the order of args, but now it doesn't, so no reason to keep it.

worker/deployer/nested.go
howbazaar: I don't think so. The comment was really a reminder for me as it was the cause of a bug. Now it should be obvious in the usage for the package as it is created with the `CurrentConfig` method which clearly mentions that it is a copy of the config.

howbazaar:
$$merge$$

cmd/jujud/agent/machine.go
howbazaar: Actually yes, it changes the return type.

howbazaar:
$$merge$$

howbazaar:
Weird that it never failed locally, but did in the merge machine.

$$merge$$

howbazaar:
$$merge$$

PR #11874, opened 2020-08-04
Introduce bash functions to stop and start unit workers. 

howbazaar:
The introspection endpoint gains a new endpoint to handle post
methods to request unit start and stop. This endpoint then uses
a new simple hub to publish the request.

The deployer now requires a simple hub to be passed in through the
manifold rather than creating one itself. The internal stop unit
message is now replaced with a centrally defined one. An additional
start unit method, as well as a status method, is added.

The scripts file that is written out into /etc/profile.d is updated
to remove the unit agent aspects, as now there is only a machine
agent. This also allows better argument passing from the other
defined functions.

## QA steps

```sh
juju bootstrap lxd test
juju deploy cs:~jameinel/ubuntu-lite ubuntu
juju deploy filebeat
juju relate ubuntu filebeat
juju add-unit ubuntu --to 0
juju ssh 0
# in another shell window start debug-log
juju debug-log
```
Wait for the model to finish deploying, and then start poking the introspection endpoint.
```
# confirm that the other functions work
juju_engine_report
juju_goroutines
juju_heap_profile
juju_metrics
# now for the new ones
juju_unit_status
juju_stop_unit ubuntu/0 filebeat/1
juju_unit_status
juju_engine_report
```
Observe that the units are shown as stopped in the `juju_unit_status`, and in the `juju_engine_report` the deployer has them as stopped.  `juju status` on the model from the outside will list the units as `lost`.

```
sudo service jujud-machine-0 restart
juju_unit_status
juju_engine_report
```
Observe that the stopped units are still stopped.
```
juju_start_unit ubuntu/0 filebeat/1
juju_unit_status
```
Observe that the units have started again, and `juju_status` shows the units as connected and idle.

## Documentation changes

Potentially worth documenting in the operational guides around managing the agent workers.

### DIFF
--- a/apiserver/facades/schema.json
+++ b/apiserver/facades/schema.json
@@ -23057,7 +23057,7 @@
                             "$ref": "#/definitions/SetMachineNetworkConfig"
                         }
                     },
-                    "description": "SetObservedNetworkConfig reads the network config for the machine identified\nby the input args. This config is merged with the new network config supplied\nin the same args and updated if it has changed."
+                    "description": "SetObservedNetworkConfig reads the network config for the machine\nidentified by the input args.\nThis config is merged with the new network config supplied in the\nsame args and updated if it has changed."
                 },
                 "SetStatus": {
                     "type": "object",
@@ -28589,7 +28589,7 @@
                             "$ref": "#/definitions/MachineNetworkConfigResults"
                         }
                     },
-                    "description": "GetContainerInterfaceInfo returns information to configure networking for a\ncontainer. It accepts container tags as arguments.\nTODO (manadart 2020-07-23): This method is not used and can be removed when\nnext this facade version is bumped.\nWe then don't need the parameterised prepareOrGet..."
+                    "description": "GetContainerInterfaceInfo returns information to configure networking for a\ncontainer. It accepts container tags as arguments."
                 },
                 "GetContainerProfileInfo": {
                     "type": "object",
@@ -28825,7 +28825,7 @@
                             "$ref": "#/definitions/SetMachineNetworkConfig"
                         }
                     },
-                    "description": "SetObservedNetworkConfig reads the network config for the machine identified\nby the input args. This config is merged with the new network config supplied\nin the same args and updated if it has changed."
+                    "description": "SetObservedNetworkConfig reads the network config for the machine\nidentified by the input args.\nThis config is merged with the new network config supplied in the\nsame args and updated if it has changed."
                 },
                 "SetPasswords": {
                     "type": "object",
--- a/worker/deployer/nested.go
+++ b/worker/deployer/nested.go
@@ -14,7 +14,6 @@ import (
 	"github.com/juju/loggo"
 	"github.com/juju/names/v4"
 	"github.com/juju/os/series"
-	"github.com/juju/pubsub"
 	"github.com/juju/utils/arch"
 	"github.com/juju/version"
 	"github.com/juju/worker/v2"
@@ -24,6 +23,7 @@ import (
 	"github.com/juju/juju/agent"
 	"github.com/juju/juju/agent/tools"
 	agenterrors "github.com/juju/juju/cmd/jujud/agent/errors"
+	message "github.com/juju/juju/pubsub/agent"
 	jujuversion "github.com/juju/juju/version"
 	jworker "github.com/juju/juju/worker"
 )
@@ -60,7 +60,7 @@ type nestedContext struct {
 	units  map[string]*UnitAgent
 	errors map[string]error
 	runner *worker.Runner
-	hub    *pubsub.SimpleHub
+	hub    Hub
 	unsub  func()
 }
 
@@ -69,6 +69,7 @@ type nestedContext struct {
 type ContextConfig struct {
 	Agent            agent.Agent
 	Clock            clock.Clock
+	Hub              Hub
 	Logger           Logger
 	UnitEngineConfig func() dependency.EngineConfig
 	SetupLogging     func(*loggo.Context, agent.Config)
@@ -82,6 +83,9 @@ func (c *ContextConfig) Validate() error {
 	}
 	if c.Clock == nil {
 		return errors.NotValidf("missing Clock")
+	}
+	if c.Hub == nil {
+		return errors.NotValidf("missing Hub")
 	}
 	if c.Logger == nil {
 		return errors.NotValidf("missing Logger")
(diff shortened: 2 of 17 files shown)


apiserver/facades/schema.json
```diff
@@ -23057,7 +23057,7 @@
                             "$ref": "#/definitions/SetMachineNetworkConfig"
                         }
                     },
-                    "description": "SetObservedNetworkConfig reads the network config for the machine identified\nby the input args. This config is merged with the new network config supplied\nin the same args and updated if it has changed."
+                    "description": "SetObservedNetworkConfig reads the network config for the machine\nidentified by the input args.\nThis config is merged with the new network config supplied in the\nsame args and updated if it has changed."
```
wallyworld, line 23060: I wonder how this got landed without the changes

worker/deployer/nested.go
```diff
@@ -60,7 +60,7 @@ type nestedContext struct {
 	units  map[string]*UnitAgent
 	errors map[string]error
 	runner *worker.Runner
-	hub    *pubsub.SimpleHub
+	hub    Hub
```
ycliuhw, line 63: It would be good to use interface for `addons` package as well

worker/deployer/nested.go
```diff
@@ -159,14 +168,51 @@ func NewNestedContext(config ContextConfig) (*nestedContext, error) {
...
+		c.logger.Errorf("data should be a Units structure")
+	}
+	for _, unitName := range units.Names {
+		if err := c.startUnit(unitName); err != nil {
+			c.logger.Errorf("%v", err)
+		}
+	}
+}
+
+func (c *nestedContext) unitStatusRequest(topic string, data interface{}) {
```
wallyworld, line 195: _ interface{}

to make it clear data is not used

I wonder if we want to allow for status of just specific units

worker/deployer/nested.go
```diff
@@ -301,18 +356,18 @@ func (c *nestedContext) DeployUnit(unitName, initialPassword string) error {
...
-	agent, found := c.units[unitName]
-	if !found {
-		var err error
-		agent, err = c.newUnitAgent(unitName)
-		if err != nil {
-			return errors.Trace(err)
-		}
-		c.units[unitName] = agent
+	agent := c.units[unitName]
+	if agent == nil {
```
wallyworld, line 363: the agent, ok pattern is preferred and more idiomatic IMO

worker/deployer/nested.go
```diff
@@ -321,6 +376,9 @@ func (c *nestedContext) startUnit(unitName string) error {
 func (c *nestedContext) stopUnitWorkers(unitName string) error {
 	// Assumes lock is held.
 	agent := c.units[unitName]
+	if agent == nil {
```
wallyworld, line 379: ditto

worker/deployer/nested.go
```diff
@@ -361,6 +420,29 @@ func (c *nestedContext) StopUnit(unitName string) error {
 	return nil
 }
 
+// StartUnit will start the workers for a stopped unit specified.
```
wallyworld, line 423: startUnit

worker/deployer/nested_test.go
```diff
@@ -239,19 +264,68 @@ func (s *NestedContextSuite) TestErrTerminateAgentFromAgentWorker(c *gc.C) {
...
+			},
+		})
+		close(responseHandled)
+	})
+	defer unsub()
+
 	ctx := s.newContext(c)
+	s.deployThreeUnits(c, ctx)
+	// And stop one.
+	handled := s.hub.Publish(message.StopUnitTopic, message.Units{
```
wallyworld, line 309: Would like to see some tests for when Start/StopUnitTopic etc is called twice

worker/introspection/script.go
```diff
@@ -131,6 +117,29 @@ juju_machine_lock () {
...
+  juju_agent units?action=status
+}
+
+juju_stop_unit () {
+  arr=("$@")
+  local -a args
+  for i in "${arr[@]}"; do
+    args+=("unit=$i")
+  done
+  juju_agent --post units action=stop "${args[@]}"
```
wallyworld, line 130: should check that the unit to stop is actually passed as an arg to the bash function
same with start
the handler will error with "missing unit" but nice/better to do it here also to avoid the unnecessary round trip

worker/introspection/worker_test.go
```diff
@@ -14,7 +14,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/juju/loggo"
+
```
ycliuhw, line 18: remove me

worker/introspection/worker.go
```diff
@@ -302,3 +330,92 @@ func (a ValueSort) Less(i, j int) bool {
...
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprintln(w, "missing unit")
+		return
+	}
+
+	h.hub.Publish(topic, agent.Units{Names: units})
+	if len(units) == 1 {
+		fmt.Fprintf(w, "requested unit %s to %s\n", units[0], action)
+	} else {
+		fmt.Fprintf(w, "requested units %s to %s\n", strings.Join(units, ", "), action)
```
wallyworld, line 382: Should we wait for the response that they have been started.
That way any error will be printed to the user immediately, otherwise they have no idea if the request errorred and would need to look at logs

worker/deployer/nested.go
ycliuhw: I think it could be ok and `agent == nil` now because we always set `c.units[unitName] = agent`.
Not sure if it's intended.

apiserver/facades/schema.json
howbazaar: No idea.

worker/deployer/nested.go
howbazaar: sure, and I don't think allowing status just for specific units is useful, this is only what units are there for the machine, there won't me many.

worker/introspection/worker.go
howbazaar: Sure.

worker/introspection/script.go
howbazaar: Done.

worker/deployer/nested_test.go
howbazaar: Sure...

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #11877, opened 2020-08-04
[dev] Introduce new API methods for endpoint aware port manipulation

achilleasa:
## Description of change

This PR is part of the work for enabling charms to open/close ports for particular application endpoints. The PR bumps the uniter API (to version 17) and introduces the following changes:
- The port range payload for the `CommitHookChanges` API has been augmented with an `Endpoint` field which allows clients to open ports for specific (or **all** if an empty endpoint is passed) application endpoints.
- Adds the `OpenedMachinePortRanges` API method which supersedes the `AllPorts` methods and returns the open machine port ranges grouped by unit and by application endpoint.

Note: as far as the `CommitHookChanges` payload changes are concerned, the change is backwards-compatible with pre-2.9 clients as the `Endpoint` field will be empty and thus the pre-2.9 behavior where open/close ports apply to all subnets (and by extension application endpoints) is preserved.

The following PRs will begin tweaking the uniter worker bits to use the new API methods.

## QA steps

Not needed; new API methods are not currently used.


api/uniter/uniter.go
```diff
@@ -395,6 +395,51 @@ func (st *State) AllMachinePorts(machineTag names.MachineTag) (map[network.PortR
 	return portsMap, nil
 }
 
+// OpenedMachinePortRanges returns all port ranges currently open on the given
+// machine, grouped by unit tag and application endpoint.
+// relation that applies.
+func (st *State) OpenedMachinePortRanges(machineTag names.MachineTag) (map[names.UnitTag]map[string][]network.PortRange, error) {
+	if st.BestAPIVersion() < 17 {
+		// OpenedMachinePortRanges() was introduced in UniterAPIV17.
+		return nil, errors.NotImplementedf("AllMachinePorts() (need V17+)")
```
hmlanigan, line 404: Copy Paste error?  s/AllMachinePorts/OpenedMachinePortRanges/

api/uniter/uniter.go
```diff
@@ -395,6 +395,51 @@ func (st *State) AllMachinePorts(machineTag names.MachineTag) (map[network.PortR
...
+		return nil, fmt.Errorf("expected 1 result, got %d", len(results.Results))
+	}
+	result := results.Results[0]
+	if result.Error != nil {
+		return nil, result.Error
+	} else if result.GroupKey != "endpoint" {
+		return nil, fmt.Errorf("expected open unit port ranges to be grouped by endpoint, got %s", result.GroupKey)
+	}
+
+	portRangeMap := make(map[names.UnitTag]map[string][]network.PortRange)
```
hmlanigan, line 424: A map of a map of a slice seems a bit complicated to deal with.  Is there a way to easier to use like has been done with the network addresses etc?  Or types?

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -392,8 +410,53 @@ func NewUniterAPIV4(context facade.Context) (*UniterAPIV4, error) {
...
+			return result.Results[i].UnitPortRanges[a].UnitTag < result.Results[i].UnitPortRanges[b].UnitTag
+		})
+	}
+	return result, nil
+}
+
 // AllMachinePorts returns all opened port ranges for each given
 // machine (on all networks).
+//
+// DEPRECATED: clients should switch to the OpenedMachinePortRanges API call.
```
hmlanigan, line 459: Why not just go all out and make it not available in V17?   The facade version is changing and there is something to move to, add comments in the OpenedMachinePortRanges.

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -403,11 +466,56 @@ func (u *UniterAPI) AllMachinePorts(args params.Entities) (params.MachinePortsRe
...
 	return result, nil
 }
 
+func (u *UniterAPI) getOneMachineOpenedPortRanges(canAccess common.AuthFunc, machineTag string) (state.MachinePortRanges, error) {
+	tag, err := names.ParseMachineTag(machineTag)
+	if err != nil {
+		return nil, apiservererrors.ErrPerm
+	}
+	if !canAccess(tag) {
+		return nil, apiservererrors.ErrPerm
```
hmlanigan, line 510: It'd be helpful annotate these ErrPerm, so we know which ErrPerm it is when tracking it down.

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -3579,16 +3639,16 @@ func (u *UniterAPI) commitHookChangesForOneUnit(unitTag names.UnitTag, changes p
...
-		// always opened in all subnets. To emulate this behavior, we
-		// simply open/close the requested port ranges for all endpoints.
 		for _, r := range changes.OpenPorts {
 			// Ensure the tag in the port open request matches the root unit name
 			if r.Tag != changes.Tag {
 				return apiservererrors.ErrPerm
 			}
-			unitPortRanges.Open("", corenetwork.PortRange{
+
+			// Pre-2.9 clients do not populate the new Endpoint
```
hmlanigan, line 3648: Pre-2.9 is clear now, but later it might be take more time to track down that this is clients using facade version 16 or earlier.  Perhaps both 2.9 / v 16 or earlier would be good.

state/machine_ports_ops.go
```diff
@@ -207,6 +212,32 @@ func (op *openClosePortRangesOperation) lookupUnitEndpoints() error {
...
 
+// validatePendingChanges ensures that the none of the pending open/close
+// entries specifies an endpoint that is not defined by the unit's charm
+// metadata.
+func (op *openClosePortRangesOperation) validatePendingChanges() error {
+	for unitName, pendingRangesByEndpoint := range op.mpr.pendingOpenRanges {
+		// Already verified; ignore error
+		appName, _ := names.UnitApplication(unitName)
+		for pendingEndpointName := range pendingRangesByEndpoint {
+			if pendingEndpointName != "" && !op.endpointsNamesByApp[appName].Contains(pendingEndpointName) {
```
hmlanigan, line 223: In what case is it valid for pendingEndpointName to be an empty string?

api/uniter/uniter.go
achilleasa: Nice catch; I will fix that

api/uniter/uniter.go
achilleasa: I couldn't find a better way to organize this information. I _could_ introduce a helper type here but then we would get references to api/uniter.$thing in the uniter's runner/context package which looks even worse (the uniter stores this information in the context)... So instead, I opted to use core types.

However, my next PR does introduce a helper type (essentially replaces ports.go in the runner/context) which gets initialized with this map and then offers an API for working with it in a sane manner.

apiserver/facades/agent/uniter/uniter.go
achilleasa: That's actually not a bad idea. I will do that. My next PR (see comment above) also removes the equivalent client method (and some others that we forgot to remove when we switched to CommitHookChanges)

apiserver/facades/agent/uniter/uniter.go
achilleasa: Just following the pattern seen in every other method here :D 

However, well behaving clients (i.e. the uniter as I don't think pylibjuju should be messing with these APIs) should never hit an ErrPerm. I could error.Trace them although I am not entirely sure whether the stack gets propagated when you convert into a ServerError.

apiserver/facades/agent/uniter/uniter.go
achilleasa: Sure thing.

state/machine_ports_ops.go
achilleasa: We follow the same pattern as with the default binding for applications. 

An empty endpoint signifies that a list of port ranges applies to **all** (present _and_ future) application endpoints for this unit. When you get down to the firewaller level, `all endpoints == all machine subnets` so we effectively get the same ingress rules as we do with 2.8 and earlier jujus.

hmlanigan: Ack.  Is this well documented in the comments already?  If not, a comment here would be great.

achilleasa: Yeap! In multiple locations (state and the facades). I have also changed pretty much all tests that manipulate ports (in the previous refactoring PRs) to pass a `allEndpoints` constant instead of an empty string to make it clear what it's all about.

apiserver/facades/agent/uniter/uniter.go
achilleasa: Actually, I cannot do this yet as client-side bits still use this API. I have created an API client/facade cleanup card to do this once all my other changes land.

api/uniter/uniter.go
```diff
@@ -395,6 +395,51 @@ func (st *State) AllMachinePorts(machineTag names.MachineTag) (map[network.PortR
 	return portsMap, nil
 }
 
+// OpenedMachinePortRanges returns all port ranges currently open on the given
+// machine, grouped by unit tag and application endpoint.
+// relation that applies.
```
hmlanigan, line 400: Partial sentence.  Does it still belong?

apiserver/facades/agent/uniter/uniter.go
```diff
@@ -392,8 +410,57 @@ func NewUniterAPIV4(context facade.Context) (*UniterAPIV4, error) {
...
+	return result, nil
+}
+
 // AllMachinePorts returns all opened port ranges for each given
 // machine (on all networks).
+//
+// DEPRECATED: clients should switch to the OpenedMachinePortRanges API call
+// when using the V17+ API.
+//
+// TODO(achilleasa): remove from V17 once all client references to this API
```
hmlanigan, line 462: 👍

api/uniter/uniter.go
achilleasa: Whoops :-(

achilleasa:
$$merge$$

PR #11882, opened 2020-08-06
[dev] Introduce new firewaller API for getting opened machine ports grouped by unit and by CIDR

achilleasa:
## Description of change

This PR introduces a new API call for the firewaller facade that allows the firewaller worker to obtain the list of open port ranges on a machine grouped by unit and by CIDR. The new call has the same signature with the similarly named call of the
Uniter (V17) facade (introduced in #11877) and shares the same response payload. However, while the uniter cares about the port ranges opened by units grouped by endpoint, the firewall worker requires the port ranges grouped by a subnet CIDR.

To produce the result for this API method, the implementation makes use of the following bits of information:
- the application endpoint to spaceID bindings for each machine unit deployed to the machine.
- the space ID to endpoint CIDRs mappings for the targeted machine.

Using this information, we can transitively resolve an endpoint (port ranges are stored in the DB grouped by unit and endpoint name) to a list of subnets and group the port ranges by CIDR. The special case of the wildcard (`""`) endpoint name is handled by expanding it to all known application endpoints and gathering the CIDRs that correspond to each one given their bound space. 

This call is _not_ yet wired into the firewaller. Once that's done (in a future PR), we will be able to lazily resolve the opened port ranges into the CIDRs needed for generating the firewall ingress rules.

## QA steps

No QA steps needed; the new API call is not wired yet.

### DIFF
--- a/api/firewaller/machine.go
+++ b/api/firewaller/machine.go
@@ -172,7 +172,7 @@ func (m *Machine) IsManual() (bool, error) {
 // OpenedMachinePortRanges queries the open port ranges for all units on this
 // machine and returns back a map where keys are unit names and values are maps
 // of the opened port ranges by the unit grouped by subnet CIDR.
-func (m *Machine) OpenedMachinePortRanges() (map[names.UnitTag]map[string][]network.PortRange, error) {
+func (m *Machine) OpenedMachinePortRanges() (map[names.UnitTag]network.GroupedPortRanges, error) {
 	if m.st.BestAPIVersion() < 6 {
 		// OpenedMachinePortRanges() was introduced in FirewallerAPIV6.
 		return nil, errors.NotImplementedf("OpenedMachinePortRanges() (need V6+)")
@@ -196,13 +196,13 @@ func (m *Machine) OpenedMachinePortRanges() (map[names.UnitTag]map[string][]netw
 		return nil, fmt.Errorf("expected open unit port ranges to be grouped by subnet CIDR, got %s", result.GroupKey)
 	}
 
-	portRangeMap := make(map[names.UnitTag]map[string][]network.PortRange)
+	portRangeMap := make(map[names.UnitTag]network.GroupedPortRanges)
 	for _, unitPortRanges := range result.UnitPortRanges {
 		unitTag, err := names.ParseUnitTag(unitPortRanges.UnitTag)
 		if err != nil {
 			return nil, errors.Trace(err)
 		}
-		portRangeMap[unitTag] = make(map[string][]network.PortRange)
+		portRangeMap[unitTag] = make(network.GroupedPortRanges)
 
 		for cidr, portRanges := range unitPortRanges.PortRangeGroups {
 			portList := make([]network.PortRange, len(portRanges))
--- a/api/firewaller/machine.go
+++ b/api/firewaller/machine.go
@@ -6,6 +6,7 @@ package firewaller
 import (
 	"fmt"
 
+	"github.com/juju/errors"
 	apiwatcher "github.com/juju/juju/api/watcher"
 	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/core/instance"
@@ -110,6 +111,10 @@ func (m *Machine) ActiveSubnets() ([]names.SubnetTag, error) {
 
 // OpenedPorts returns a map of network.PortRange to unit tag for all opened
 // port ranges on the machine for the subnet matching given subnetTag.
+//
+// TODO(achilleasa): remove from client once we complete the migration to the
+// new API.
+// DEPRECATED: use OpenedPortRanges instead.
 func (m *Machine) OpenedPorts(subnetTag names.SubnetTag) (map[network.PortRange]names.UnitTag, error) {
 	var results params.MachinePortsResults
 	var subnetTagAsString string
@@ -162,4 +167,50 @@ func (m *Machine) IsManual() (bool, error) {
 		return false, result.Error
 	}
 	return result.Result, nil
+}
+
+// OpenedMachinePortRanges queries the open port ranges for all units on this
+// machine and returns back a map where keys are unit names and values are maps
+// of the opened port ranges by the unit grouped by subnet CIDR.
+func (m *Machine) OpenedMachinePortRanges() (map[names.UnitTag]map[string][]network.PortRange, error) {
+	if m.st.BestAPIVersion() < 6 {
+		// OpenedMachinePortRanges() was introduced in FirewallerAPIV6.
+		return nil, errors.NotImplementedf("OpenedMachinePortRanges() (need V6+)")
+	}
+
+	var results params.OpenMachinePortRangesResults
+	args := params.Entities{
+		Entities: []params.Entity{{Tag: m.tag.String()}},
+	}
+	err := m.st.facade.FacadeCall("OpenedMachinePortRanges", args, &results)
+	if err != nil {
+		return nil, err
+	}
+	if len(results.Results) != 1 {
+		return nil, fmt.Errorf("expected 1 result, got %d", len(results.Results))
+	}
+	result := results.Results[0]
+	if result.Error != nil {
+		return nil, result.Error
+	} else if result.GroupKey != "cidr" {
+		return nil, fmt.Errorf("expected open unit port ranges to be grouped by subnet CIDR, got %s", result.GroupKey)
+	}
+
+	portRangeMap := make(map[names.UnitTag]map[string][]network.PortRange)
+	for _, unitPortRanges := range result.UnitPortRanges {
+		unitTag, err := names.ParseUnitTag(unitPortRanges.UnitTag)
+		if err != nil {
+			return nil, errors.Trace(err)
+		}
+		portRangeMap[unitTag] = make(map[string][]network.PortRange)
+
+		for cidr, portRanges := range unitPortRanges.PortRangeGroups {
+			portList := make([]network.PortRange, len(portRanges))
+			for i, pr := range portRanges {
+				portList[i] = pr.NetworkPortRange()
+			}
+			portRangeMap[unitTag][cidr] = portList
+		}
+	}
+	return portRangeMap, nil
 }
--- a/api/firewaller/machine.go
+++ b/api/firewaller/machine.go
@@ -7,13 +7,14 @@ import (
 	"fmt"
 
 	"github.com/juju/errors"
+	"github.com/juju/names/v4"
+
 	apiwatcher "github.com/juju/juju/api/watcher"
 	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/core/instance"
 	"github.com/juju/juju/core/life"
 	"github.com/juju/juju/core/network"
 	"github.com/juju/juju/core/watcher"
-	"github.com/juju/names/v4"
 )
 
 // Machine represents a juju machine as seen by the firewaller worker.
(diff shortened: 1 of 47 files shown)


achilleasa:
!!build!! (DNS timeout in unrelated RefreshClientSuite.TestLiveRefreshManyRequest)

api/firewaller/machine.go
```diff
@@ -163,3 +168,49 @@ func (m *Machine) IsManual() (bool, error) {
 	}
 	return result.Result, nil
 }
+
+// OpenedMachinePortRanges queries the open port ranges for all units on this
+// machine and returns back a map where keys are unit names and values are maps
+// of the opened port ranges by the unit grouped by subnet CIDR.
+func (m *Machine) OpenedMachinePortRanges() (map[names.UnitTag]map[string][]network.PortRange, error) {
```
manadart, line 175: `map[string][]network.PortRange` appears quite a lot. Might we make a type/alias for it?

achilleasa: Thought of it but was unsure how to call it. However, note that just as with the previous PR, the idea is to feed this into a helper type that exposes a friendly API so you are not going to be seeing much of it floating around in the code-base.

apiserver/facades/controller/firewaller/firewaller.go
```diff
@@ -535,3 +550,164 @@ func (f *FirewallerAPIV5) AreManuallyProvisioned(args params.Entities) (params.B
...
+
+	// Sort by unit tag so we return consistent results
+	sort.Slice(res, func(i, j int) bool {
+		return res[i].UnitTag < res[j].UnitTag
+	})
+
+	return res, nil
+}
+
+func (f *FirewallerAPIV6) resolveApplicationBindings(apps set.Strings) (map[string]map[string]string, error) {
```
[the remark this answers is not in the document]
achilleasa: Given that machines generally don't tend to have to have many applications deployed to them, I thought this call might be good enough in terms of performance (especially on larger models with many deployed applications) as opposed to pulling everything from the DB.

However, if the space topology changes (although I guess the bindings will be the ones changing more often so this might be a moot point), the firewall worker will be making a call for each machine in the model so perhaps we could get all endpoints once at the top of "OpenedMachinePortRanges" and re-use it for each machine. This also would mean that we need to move the client out of that per-machine wrapper (or clone it) so it can pass all machine tags in one go.

Thoughts?

state/machine.go
```diff
@@ -1609,14 +1609,15 @@ func (m *Machine) AddressesBySpaceID() (map[string][]corenetwork.SpaceAddress, e
...
 
-// SubnetsBySpaceID resolves the subnet assignments for each space the machine
-// is associated with and returns them back as a map where keys are space IDs
-// and each value is a slice of subnet IDs assigned to a particular space.
-func (m *Machine) SubnetsBySpaceID() (map[string][]string, error) {
+// SubnetCIDRsBySpaceID resolves the subnet assignments for each space the
+// machine is associated with and returns them back as a map where keys are
+// space IDs and each value is a slice of subnet CIDRs assigned to a particular
+// space.
+func (m *Machine) SubnetCIDRsBySpaceID() (map[string][]string, error) {
```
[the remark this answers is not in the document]
achilleasa: Can I get this information from SpaceInfos (or add a helper to do that?). If so, I don't mind changing my code to use SpaceInfos here and remove this method (it was added by me in the past and nothing else uses it).

apiserver/facades/controller/firewaller/firewaller.go
```diff
@@ -535,3 +550,164 @@ func (f *FirewallerAPIV5) AreManuallyProvisioned(args params.Entities) (params.B
...
+// To facilitate this conversion, the function consults the application
+// endpoint bindings for the unit in conjunction with the provided subnetCIDRs
+// by spaceID map. Using this information, each endpoint from the incoming
+// port range grouping is resolved to a space ID and the space ID is in turn
+// resolved into a list of subnet CIDRs (the wildcard endpoint is treated as
+// *all known* endpoints for this conversion step).
+//
+// After the above step, all opened port ranges are then grouped by subnet CIDR,
+// de-dupped and sorted before they are returned to the caller.
+func mapUnitPortsToSubnetCIDRs(unitName string, portRangesByEndpoint map[string][]network.PortRange, endpointBindings map[string]string, subnetCIDRsBySpaceID map[string][]string) map[string][]network.PortRange {
```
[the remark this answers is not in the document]
achilleasa: Nice catch; it was used but then I changed the return type and forgot to remove it.

api/firewaller/machine.go
achilleasa: After thinking about it, it might make sense to add this to core network as a `GroupedPortRanges` type. Then I can provide the UniquePortRanges helper for it and clean up the repeated code from the call-sites.

state/machine.go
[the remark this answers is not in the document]
achilleasa: One snag here is that we want the space -> []subnet assignments for a particular machine. Is it possible for a machine to only have addresses in a subset of the subnets allocated to a space?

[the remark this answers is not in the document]
achilleasa: On the other hand, we could grab all subnets anyway. Worst case we will generate a few bogus ingress rules (which is fine because we are not really connected to any of those subnets anyway)

api/firewaller/machine.go
manadart: :+1:

state/machine.go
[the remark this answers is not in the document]
manadart: What we're actually doing here is getting all the addresses, and making a query for each to get the subnet.

If you had a materialised `SpaceInfos`, you could just call `AllSubnetInfos` and use `GetByCIDR` for each address. If you are not already in possession of a `SpaceInfos`, it might not be worth it for single-NIC machines...

core/network/portrange.go
```diff
@@ -12,6 +12,34 @@ import (
 	"github.com/juju/errors"
 )
 
+// GroupedPortRanges represents a list of PortRange instances grouped by a
+// particular feature.
+type GroupedPortRanges map[string][]PortRange
+
+// UniquePortRanges returns the unique set of PortRanges in this group.
+func (grp GroupedPortRanges) UniquePortRanges() []PortRange {
```
manadart, line 20: Really nice clean-up out of this type and method.

cmd/juju/commands/ssh_container.go
```diff
@@ -217,6 +217,9 @@ func (c *sshContainer) getExecClient() (k8sexec.Executor, error) {
 		return nil, errors.Annotatef(mInfo.Error, "getting model information")
 	}
 	credentialTag, err := names.ParseCloudCredentialTag(mInfo.Result.CloudCredentialTag)
+	if err != nil {
```
manadart, line 220: Oops.

apiserver/facades/controller/firewaller/state.go
```diff
@@ -64,15 +66,23 @@ func (st stateShim) FirewallRule(service corefirewall.WellKnownServiceType) (*st
 	return api.Rule(service)
 }
 
-type Subnet interface {
-	ID() string
-	CIDR() string
+func (st stateShim) AllEndpointBindings() (map[string]map[string]string, error) {
```
manadart, line 69: So looking at `stateShim` it's always a red flag to have two members with one or more being a veneer for essentially the same type.

Quick hack, still with errors:
```diff
diff --git a/apiserver/facades/controller/firewaller/firewaller.go b/apiserver/facades/controller/firewaller/firewaller.go
index c99947c379..ca854478ec 100644
--- a/apiserver/facades/controller/firewaller/firewaller.go
+++ b/apiserver/facades/controller/firewaller/firewaller.go
@@ -82,7 +82,7 @@ func NewStateFirewallerAPIV3(context facade.Context) (*FirewallerAPIV3, error) {
 		cloudspec.MakeCloudSpecCredentialContentWatcherForModel(st),
 		common.AuthFuncForTag(m.ModelTag()),
 	)
-	return NewFirewallerAPI(stateShim{st: st, State: firewall.StateShim(st, m)}, context.Resources(), context.Auth(), cloudSpecAPI)
+	return NewFirewallerAPI(stateShim{st}, context.Resources(), context.Auth(), cloudSpecAPI)
 }
 
 // NewStateFirewallerAPIV4 creates a new server-side FirewallerAPIV4 facade.
@@ -395,7 +395,7 @@ func (f *FirewallerAPIV3) getSpaceInfos() (network.SpaceInfos, error) {
 		return f.spaceInfos, nil
 	}
 
-	si, err := f.st.SpaceInfos()
+	si, err := f.st.AllSpaceInfos()
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
diff --git a/apiserver/facades/controller/firewaller/state.go b/apiserver/facades/controller/firewaller/state.go
index 56cdf215b7..44048e350f 100644
--- a/apiserver/facades/controller/firewaller/state.go
+++ b/apiserver/facades/controller/firewaller/state.go
@@ -31,47 +31,35 @@ type State interface {
 
 	AllEndpointBindings() (map[string]map[string]string, error)
 
-	SpaceInfos() (network.SpaceInfos, error)
+	AllSpaceInfos() (network.SpaceInfos, error)
 }
 
 // TODO(wallyworld) - for tests, remove when remaining firewaller tests become unit tests.
 func StateShim(st *state.State, m *state.Model) stateShim {
-	return stateShim{st: st, State: firewall.StateShim(st, m)}
+	return stateShim{st, m}
 }
 
 type stateShim struct {
-	firewall.State
-	st *state.State
+	*state.State
+	*state.Model
 }
 
 func (st stateShim) ModelUUID() string {
-	return st.st.ModelUUID()
+	return st.ModelUUID()
 }
 
 func (st stateShim) GetMacaroon(entity names.Tag) (*macaroon.Macaroon, error) {
-	r := st.st.RemoteEntities()
+	r := st.RemoteEntities()
 	return r.GetMacaroon(entity)
 }
 
-func (st stateShim) FindEntity(tag names.Tag) (state.Entity, error) {
-	return st.st.FindEntity(tag)
-}
-
-func (st stateShim) WatchOpenedPorts() state.StringsWatcher {
-	return st.st.WatchOpenedPorts()
-}
-
 func (st stateShim) FirewallRule(service corefirewall.WellKnownServiceType) (*state.FirewallRule, error) {
-	api := state.NewFirewallRules(st.st)
+	api := state.NewFirewallRules(st.State)
 	return api.Rule(service)
 }
 
 func (st stateShim) AllEndpointBindings() (map[string]map[string]string, error) {
-	model, err := st.st.Model()
-	if err != nil {
-		return nil, errors.Trace(err)
-	}
-	allEndpointBindings, err := model.AllEndpointBindings()
+	allEndpointBindings, err := st.Model.AllEndpointBindings()
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
@@ -82,7 +70,3 @@ func (st stateShim) AllEndpointBindings() (map[string]map[string]string, error)
 	}
 	return res, nil
 }
-
-func (st stateShim) SpaceInfos() (network.SpaceInfos, error) {
-	return st.st.AllSpaceInfos()
-}
```
Might be possible to pear down this to something more crisp.

achilleasa: I have a cleanup card for the facade/client bits.  I can try refactoring the test suite as part of that work so we don't make this PR longer than it should be.

achilleasa:
@manadart Just realized that the json schema is missing and the linter did not complain!

achilleasa:
$$merge$$

PR #11885, opened 2020-08-10
New apiserver endpoint for running embedded CLI commands

wallyworld:
## Description of change

The controller gains a new "/commands" endpoint which provides a websocket. Sending json snippets to this socket runs named CLI commands directly against the controller. The json data request contains the command and args, as well as the username and credential (password or macaroon) - these are used to set up an in memory ClientStore which is passed to a CLI instance constructed the same as if it were invoked from a terminal. The CLI instance runs as normal, making the same API calls, logging into the controller using the supplied credential. So the /cli endpoint is just a shim used to invoke the CLI commands; it doesn't itself attempt to make any API calls against the controller.

Some tweaks were needed to how the Juju SuperCommand was constructed - logger is passed in explicitly instead of being set up internally (to avoid issues with loggo global variables). Plus, commands now support an embedded flag so that certain behaviours can be changed when running embedded. eg the Optional Controller commands which require a client or controller to be specified always use the single controller on which they are running so do not need the --client or --controller flags. And we always need to be sure that the passed in juju ClientStore is used; not all commands allowed it to be set and always created the file based version.

Still TODO in the next PR is to disable os file api calls as embedded commands do not support file access.

## QA steps

bootstrap a controller

Use the test utility here:
https://gist.github.com/wallyworld/4e8de75d5a439bd511e27d9bf5046c31

Also works with macaroons instead of a password, but you need to extra the macaroons from a juju login request. The Dashboard will have access to these.


apiserver/apiserver.go
```diff
@@ -783,6 +792,11 @@ func (srv *Server) endpoints() []apihttp.Endpoint {
 		tracked:         true,
 		unauthenticated: true,
 		noModelUUID:     true,
+	}, {
+		pattern:         "/cli",
```
howbazaar, line 796: I'm not overly enthused with the idea of calling the endpoint "/cli". Perhaps something like 'repl' ?

apiserver/apiserver.go
```diff
@@ -1015,3 +1029,47 @@ func (srv *Server) GetAuditConfig() auditlog.Config {
...
 }
+
+func (srv *Server) embeddedCLIHandler(w http.ResponseWriter, req *http.Request) {
+	var err error
+	var result params.StringResults
+	defer func() {
+		if err == nil {
+			err = sendStatusAndJSON(w, http.StatusOK, result)
+		}
+		if err != nil {
```
howbazaar, line 1040: Perhaps had a comment as to why this isn't just an else? It took me a minute to notice, so I think it would help future readers too.

apiserver/embeddedcommands.go
```diff
@@ -0,0 +1,231 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package apiserver
```
howbazaar, line 4: Is there a reason to have this in the apiserver package rather than a subpackage? If it was a subpackage it could be more easily brought into JIMM.

apiserver/embeddedcommands.go
```diff
@@ -0,0 +1,231 @@
...
+	"remove-unit",
+	"remove-user",
+	"rename-space",
+	"resolved",
+	"resolve",
+	"resources",
+	"resume-relation",
+	"retry-provisioning",
+	"revoke",
+	"run",
```
howbazaar, line 97: This is run action rather than exec?

apiserver/embeddedcommands.go
```diff
@@ -0,0 +1,231 @@
...
+	result := params.StringResults{}
+	if commands.AuthTag == "" {
+		return result, errors.NotSupportedf("CLI command for anonymous user")
+	}
+	userTag, err := names.ParseTag(commands.AuthTag)
+	if err != nil {
+		return result, errors.Trace(err)
+	}
+	if userTag.Kind() != names.UserTagKind {
+		return result, errors.NotSupportedf("CLI command for tag %q", userTag.Kind())
```
howbazaar, line 149: tag shouldn't be exposed to users, it should be internal

apiserver/apiserver.go
wallyworld: Strictly speaking, each API request is separate from the one before and causes a new CLI command instance to be created and run. So it doesn't behave as a repl in that sense, ie it's not an (interactive) loop

wallyworld: I though about "embedded" or "embeddedcli" but these seemed overly long.
The idea behind "cli" was to reflect the purpose (embedded cli for dashboard or any client really).

apiserver/embeddedcommands.go
wallyworld: I had it in a subpackage originally. Then when the tests were best written in the apiserver_test.go file I thought it a little weird to have a subpackage with just one go file. But I can move back to a subpackage.

apiserver/embeddedcommands.go
wallyworld: Yeah, "run" is "run-action" in new CLI, and equivalent to "exec" in the old CLI. So it exists either way.

apiserver/embeddedcommands.go
wallyworld: Issue here is that prams structs use tags.
model-tag is used also in the CLI args struct.
For this case, I could revert to using just UUID and username. It depends on whether we want to treat this http endpoint as an API call.

apiserver/apiserver.go
wallyworld: Changed to /commands for now

wallyworld:
$$merge$$

PR #11888, opened 2020-08-10
Remove charm.LXDProfile from state

SimonRichardson:
## Description of change

The following attempts to resolve a long outstanding issue in state. The
`charmDoc` writes directly to the database the `charm.LXDProfile`, which is
an external dependency and means that any update of that charm
dependency could potentially break our database.

The following attempts to fix the issue in a piecemeal fashion, by
tackling one field at a time. Here we're tackling the LXD profile from
the charm package.

The code is very mechanical and just requires to correctly marshal
between state, lxdprofile and charm profiles. Each correctly contained
to their vertical boundary.

I suspect Meta, Config and Actions will be the hardest to move, but baby
steps are more than welcome!

## QA steps

```sh
juju bootstrap lxd test
juju deploy lxd-profile-test
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1467964


SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

hmlanigan:
$$merge$$

PR #11890, opened 2020-08-10
[2.8] Sync remove juju script to the one that gets installed

achilleasa:
## Description of change

Sync the contents of the remove juju script in the juju source tree to the version installed to each machine.

### DIFF
--- a/scripts/removejujuservices.bash
+++ b/scripts/removejujuservices.bash
@@ -10,7 +10,11 @@ for path_to_unit in $(ls /etc/systemd/system/juju*); do
   systemctl stop "$unit"
   systemctl disable "$unit"
   systemctl daemon-reload
+  rm -f "$path_to_unit"
 done
+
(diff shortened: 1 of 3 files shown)


scripts/removejujuservices.bash
```diff
@@ -20,3 +24,9 @@ rm -rf /var/lib/juju/raft/*
 
 echo "removing /var/run/juju/*"
 rm -rf /var/run/juju/*
+
+has_juju_db_snap=$(snap info juju-db | grep installed)
```
hmlanigan, line 28: Suggested you change "installed" to "installed:".  If the description of the juju-db snap changes, you may end up with false positives.  Noticed when I ran the grep with juju instead of juju-db.

achilleasa: aha! thanks. I will update my PR accordingly

achilleasa:
$$merge$$

PR #11893, opened 2020-08-11
Add juju_leases introspection command.

howbazaar:
This branch adds a /leases endpoint in the introspection http handler.
The new juju_leases function wraps up calls to this function and allows
for the user to easily add application names (or partial names) or
part of a model UUID using -m.

The result is always formatted YAML. This conveniently sorts all the
keys for us, and would allow other tools to programaticly interrogate
the endpoint if need be.

## QA steps

Bootstrap a controller, and deploy some applications.

```sh
juju ssh -m controller 0
juju_leases
juju_leases -m default
juju_leases appname
```

## Documentation changes

Should be added to troubleshooting guide around leases.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1890070


howbazaar:
A followup PR will add the ability to revoke leadership of an app.

worker/introspection/leases.go
```diff
@@ -0,0 +1,185 @@
...
+		fmt.Fprintf(w, "error: %v\n", err)
+		return
+	}
+	w.Write(bytes)
+}
+
+type leases struct {
+	// TODO: handle pinning
+	controller map[string]leaseInfo
+	models     map[string]map[string]leaseInfo
```
wallyworld, line 63: perhaps a comment on what the map keys are

worker/introspection/leases.go
```diff
@@ -0,0 +1,185 @@
...
+	return result
+}
+
+func (h leaseHandler) filterModel(data *leases, partialModelUUID string) *leases {
+	result := &leases{
+		controller: make(map[string]leaseInfo),
+		models:     make(map[string]map[string]leaseInfo),
+	}
+	for modelUUID, value := range data.controller {
+		if strings.Contains(modelUUID, partialModelUUID) {
```
wallyworld, line 115: startsWith maybe instead?
I would expect partial UUIDs would always be the first few characters. Saves from unexpected matches

worker/introspection/leases.go
```diff
@@ -0,0 +1,185 @@
...
+func (h leaseHandler) filterApp(data *leases, partialAppNames []string) *leases {
+	result := &leases{
+		models: make(map[string]map[string]leaseInfo),
+	}
+	for modelUUID, apps := range data.models {
+		appInfo := make(map[string]leaseInfo)
+		for appName, value := range apps {
+			// If the appName matches any of the partial names, add it in.
+			for _, partial := range partialAppNames {
+				if strings.Contains(appName, partial) {
```
wallyworld, line 137: regexp match?
allows for "mysql.*" or ".*mysql" etc
we'd take the partial string and make a regexp of "^" + partial + "$" to make the default match the whole app name
as with partial uuid matches, we want to avoid surprises in the match

worker/introspection/leases.go
howbazaar: except the short hash we use almost everywhere is the last seven

worker/introspection/leases.go
howbazaar: Adding regexp just adds to problems IMO. Partial string matches are easy to explain, and easy to implement.

You can provide multiple matches, which will be easier than making people think about how to say that in a regexp.

worker/introspection/leases.go
howbazaar: added a comment.

howbazaar:
$$merge$$

howbazaar:
$$merge$$

PR #11897, opened 2020-08-11
Move upgrade step to move unit state from machine to unit agents

hmlanigan:
## Description of change

There is a window where the machine agent is trying to write unit state to the controller during upgrade, where running units can attempt to read relation data.  This would be more obvious on machines running many units as seen in production with many subordinates.

Move writing of a unit's state to be the unit's responsibility during upgrade.

## QA steps

Bootstrap 2.6.x or 2.7.x.  Deploy a config where units have many monitoring subordinates.  Or many units on the same machine. Upgrade juju. 

Failure if units try to install themselves again.  Relations lost. Storage is lost.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1890828


hmlanigan:
$$merge$$

PR #11900, opened 2020-08-12
[dev] Refactor ingress rule types into dedicated core/network/firewall package

achilleasa:
## Description of change

This PR is part of the work for allowing the firewall worker to manipulate port ranges in an endpoint-aware manner. It applies a couple of refactoring changes to serve as the scaffolding for the upcoming changes to the firewall worker.

The key refactoring change here is that the `IngressRule` definitions that used to live in the legacy `network` package have now been moved into the `core/network/firewall` package, refactored to use core network types (`PortRange`) and have been augmented with helper methods that are useful for performing ingress rule-related work. In addition, the PR introduces the `IngressRules` helper type (a `[]IngressRule`) with additional validation, sorting and comparison (Equal/Diff) methods. 

What's more, the known service definitions have been relocated from `core/firewall` into the `core/network/firewall` package while the original type definitions for ingress rules have been removed from the `network` package.

This change required a significant amount of _mostly mechanical changes_ to the providers (and some other bits). These are part of the last PR commit. Another benefit of this refactoring is that it was possible to remove the `juju/network` import from a whole bunch of files and switch to `juju/core/network` without the need to use an import alias.


core/network/firewall/rule.go
```diff
@@ -0,0 +1,203 @@
...
+		if !thisRule.EqualTo(other[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+// Diff returns a list of IngressRules to open and/or close so that this
+// set of ingress rules matches the target.
+func (r IngressRules) Diff(target IngressRules) (toOpen, toClose IngressRules) {
```
achilleasa, line 149: This bit was yanked out of the firewall worker.

core/network/firewall/rule_test.go
```diff
@@ -0,0 +1,245 @@
...
+		NewIngressRule(network.MustParsePortRange("80/tcp"), "192.168.0.0/24", "10.0.0.0/24"),
+		NewIngressRule(network.MustParsePortRange("90/tcp"), "10.0.0.0/24", "192.168.0.0/24"),
+	}
+
+	c.Assert(setA.EqualTo(setB), jc.IsTrue)
+	c.Assert(setA.EqualTo(setC), jc.IsFalse)
+	c.Assert(setB.EqualTo(setC), jc.IsFalse)
+}
+
+func (IngressRuleSuite) TestDiffOpenAll(c *gc.C) {
```
achilleasa, line 126: All diff tests where yanked out of the firewall worker.

provider/azure/instance.go
```diff
@@ -374,7 +375,7 @@ func (inst *azureInstance) IngressRules(ctx context.ProviderCallContext, machine
 
 		var portRange corenetwork.PortRange
 		if *rule.DestinationPortRange == "*" {
-			portRange.FromPort = 0
+			portRange.FromPort = 1
```
achilleasa, line 378: Changed this to 1 as specifying a 0 port here triggers a port range validation error

core/network/firewall/rule.go
```diff
@@ -0,0 +1,203 @@
...
+
+	"github.com/juju/errors"
+	"github.com/juju/utils/set"
+
+	"github.com/juju/juju/core/network"
+)
+
+// AllNetworksIPV4CIDR represents the zero address (quad-zero) CIDR for an IPV4
+// network.
+const AllNetworksIPV4CIDR = "0.0.0.0/0"
```
achilleasa, line 21: I am using this const in the provider code but kept "0.0.0.0/0" in tests to make sure this does not change. Also, having this will make it easy to figure out what firewall bits need to be changed for IPV6

provider/gce/google/conn_network.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/juju/errors"
 	"google.golang.org/api/compute/v1"
 
-	"github.com/juju/juju/network"
+	corefirewall "github.com/juju/juju/core/network/firewall"
```
achilleasa, line 15: Of course the internal package type for a gce ingress rule would be called `firewall`...

achilleasa:
Had to rebase to pick up the k8s_agent changes from develop HEAD or a test in that pkg keeps failing.

provider/manual/environ.go
```diff
@@ -366,15 +366,15 @@ func (e *manualEnviron) seriesAndHardwareCharacteristics() (_ *instance.Hardware
 	return e.hw, e.series, nil
 }
 
-func (e *manualEnviron) OpenPorts(ctx context.ProviderCallContext, rules []network.IngressRule) error {
+func (e *manualEnviron) OpenPorts(ctx context.ProviderCallContext, rules firewall.IngressRules) error {
```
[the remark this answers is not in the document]
achilleasa: I plan to replace the not-implemented ones (instance-related Open/ClosePorts) soon with an iptables implementation so we get instance firewall support across all providers. We can revisit this then.

provider/openstack/provider.go
```diff
@@ -523,15 +523,15 @@ func convertNovaAddresses(publicIP string, addresses map[string][]nova.IPAddress
 	return machineAddresses
 }
 
-func (inst *openstackInstance) OpenPorts(ctx context.ProviderCallContext, machineId string, rules []network.IngressRule) error {
+func (inst *openstackInstance) OpenPorts(ctx context.ProviderCallContext, machineId string, rules firewall.IngressRules) error {
```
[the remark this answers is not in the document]
achilleasa: I think it's being traced in the caller.

achilleasa:
$$merge$$

PR #11904, opened 2020-08-13
Remove incorrect assumption.  Shorten goose error message.

hmlanigan:
## Description of change
The assumption you cannot create an OpenStack instance with no network
defined if there is an vailable network is false.  Remove that logic.

Keep the output change to suggest a network to specify if
tryStartInstance fails due to missing network.

If StartInstance fails due to error from RunServer, print the cause, not the
entire thing which includes a []byte.  Instead print the error cause,
and log the entire message as Debug.

ERROR failed to bootstrap model: cannot start bootstrap instance: cannot run instance: request (http://10.20.20.1:8774/v2.1/servers) returned unexpected status: 400; error info: {"badRequest": {"code": 400, "message": "Invalid input for field/attribute networks. Value: None. None is not of type 'array'"}}
	This error was caused by juju attempting to create an OpenStack instance with no network defined.
	No network has been configured. The following internal networks are available:
		"test"

## QA steps

- Bootstrap microstack with no network specified on the command line, nor config - expect failure, see new error output
- Bootstrap microstack --debug with no network specified on the command line, nor config - expect failure, see big struct in debug output.
- Bootstrap microstack with no network specified on the command line, nor config - expect success
- Bootstrap serverstack with no network specified on the command line, nor config. - expect success
- Bootstrap serverstack with network specified - expect success

### DIFF
--- a/provider/openstack/provider.go
+++ b/provider/openstack/provider.go
@@ -1339,14 +1339,14 @@ func (e *Environ) startInstance(
 }
 
 func (e *Environ) userFriendlyInvalidNetworkError(err error) error {
-	msg := fmt.Sprintf("%s\n\t%s\n\t%s", err,
+	msg := fmt.Sprintf("%s\n\t%s\n\t%s", err.Error(),
 		"This error was caused by juju attempting to create an OpenStack instance with no network defined.",
 		"No network has been configured.")
 	networks, err := e.networking.FindNetworks(true)
 	if err != nil {
-		msg = fmt.Sprintf("%s\n\t%s\n\t\t%s", msg, "Error attempting to find internal networks:", err.Error())
+		msg += fmt.Sprintf("\n\t%s\n\t\t%s", "Error attempting to find internal networks:", err.Error())
 	} else {
-		msg = fmt.Sprintf("%s %s\n\t\t%q", msg, "The following internal networks are available: ", strings.Join(networks.Values(), ", "))
+		msg += fmt.Sprintf(" %s\n\t\t%q", "The following internal networks are available: ", strings.Join(networks.Values(), ", "))
 	}
 	return errors.New(msg)
 }
--- a/provider/openstack/provider.go
+++ b/provider/openstack/provider.go
@@ -1276,11 +1276,10 @@ func (e *Environ) startInstance(
 	server, err := tryStartNovaInstance(shortAttempt, e.nova(), opts)
 	if err != nil || server == nil {
 		logger.Debugf("cannot run instance full error: %q", err)
-		errCause := errors.Cause(err)
-		err := errors.Annotate(errCause, "cannot run instance")
+		err = errors.Annotate(errors.Cause(err), "cannot run instance")
 		// Improve the error message if there is no valid network.
 		if isInvalidNetworkError(err) {
-			err = errors.New(fmt.Sprintf("%s%s", err.Error(), e.invalidNetworkErrorMessageForUser()))
+			err = e.userFriendlyInvalidNetworkError(err)
 		}
 		// 'No valid host available' is typically a resource error,
 		// let the provisioner know it is a good idea to try another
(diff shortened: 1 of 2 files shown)


provider/openstack/provider.go
```diff
@@ -1275,11 +1275,12 @@ func (e *Environ) startInstance(
 
 	server, err := tryStartNovaInstance(shortAttempt, e.nova(), opts)
 	if err != nil || server == nil {
-		err := errors.Annotate(err, "cannot run instance")
+		logger.Debugf("cannot run instance full error: %q", err)
+		errCause := errors.Cause(err)
+		err := errors.Annotate(errCause, "cannot run instance")
```
achilleasa, line 1280: You can remove the `:` to avoid shadowing `err`

provider/openstack/provider.go
```diff
@@ -1275,11 +1275,12 @@ func (e *Environ) startInstance(
...
 		// Improve the error message if there is no valid network.
 		if isInvalidNetworkError(err) {
-			msg := "\n\tThis error was caused by juju attempting to create an OpenStack instance with no network defined.\n\tNo network has been configured, nor could juju find an internal OpenStack network to suggest."
-			err = errors.New(fmt.Sprintf("%s%s", err.Error(), msg))
+			err = errors.New(fmt.Sprintf("%s%s", err.Error(), e.invalidNetworkErrorMessageForUser()))
```
achilleasa, line 1283: This looks a bit strange... why not just pass `err` to `invalidNetworkErrorMessageForUser` (btw, perhaps `userFriendlyNetworkError` would be a more suitable name) and have it return an error back? Wouldn't just annotating the error with the user-friendly message do the trick?

provider/openstack/provider.go
hmlanigan: done

provider/openstack/provider.go
hmlanigan: Updating the method name. 

Due to the length of the error, annotating it will make it harder to read now to to fix imo.

provider/openstack/provider.go
```diff
@@ -1338,6 +1338,19 @@ func (e *Environ) startInstance(
 	}, nil
 }
 
+func (e *Environ) userFriendlyInvalidNetworkError(err error) error {
+	msg := fmt.Sprintf("%s\n\t%s\n\t%s", err,
```
achilleasa, line 1342: You probably want `err.Error()` as the Sprintf argument just in case the passed error type does not implement `fmt.Stringer`

provider/openstack/provider.go
```diff
@@ -1338,6 +1338,19 @@ func (e *Environ) startInstance(
...
+		"This error was caused by juju attempting to create an OpenStack instance with no network defined.",
+		"No network has been configured.")
+	networks, err := e.networking.FindNetworks(true)
+	if err != nil {
+		msg = fmt.Sprintf("%s\n\t%s\n\t\t%s", msg, "Error attempting to find internal networks:", err.Error())
```
achilleasa, line 1347:
```suggestion
		msg += fmt.Sprintf("\nError attempting to find internal networks: %s, err.Error())
```

provider/openstack/provider.go
```diff
@@ -1338,6 +1338,19 @@ func (e *Environ) startInstance(
...
+	} else {
+		msg = fmt.Sprintf("%s %s\n\t\t%q", msg, "The following internal networks are available: ", strings.Join(networks.Values(), ", "))
```
achilleasa, line 1349:
```suggestion
		msg += fmt.Sprintf("\nThe following internal networks are available:\n  %q", strings.Join(networks.Values(), ", "))
```

AFAICT the client uses double-spaces instead of tabs for indentation.

provider/openstack/provider.go
hmlanigan: done

provider/openstack/provider.go
hmlanigan: we're inconsistent with the spacing.  keeping inline with what the rest of the openstack provider does at this time.

hmlanigan:
$$merge$$

PR #11905, opened 2020-08-14
Instrospect revoke lease

howbazaar:
This work is to add an introspection endpoint to revoke a lease.

It is functionally there, but needs tests.

While testing interactively I did notice that the lease wasn't claimed for somewhere around five seconds. Perhaps something is polling. Sorry that it isn't fully complete.


wallyworld:
Superceded by #12332

PR #11906, opened 2020-08-14
Add all facades to the schema

SimonRichardson:
## Description of change

Output **all** the facades into the schema!

This will require pylibjuju updates, so we have to be very careful with
this. The jslib version of juju wants all the information though, so I'm
unsure what the best path forward here is.

## QA steps

```sh
make rebuild-schema
```

### DIFF
--- a/generate/schemagen/gen/generator.go
+++ b/generate/schemagen/gen/generator.go
@@ -78,23 +78,16 @@ func Generate(pkgRegistry PackageRegistry, linker Linker, client APIServer, opti
 		facades = append(facades, adminFacades...)
 	}
 
-	latest := make(map[string]facade.Details)
-	for _, facade := range facades {
-		if f, ok := latest[facade.Name]; ok && facade.Version < f.Version {
-			continue
-		}
-		latest[facade.Name] = facade
+	allFacades := make([]facade.Details, 0, len(facades))
+	for _, v := range facades {
+		allFacades = append(allFacades, v)
 	}
-	latestFacades := make([]facade.Details, 0, len(latest))
-	for _, v := range latest {
-		latestFacades = append(latestFacades, v)
-	}
-	sort.Slice(latestFacades, func(i, j int) bool {
-		return latestFacades[i].Name < latestFacades[j].Name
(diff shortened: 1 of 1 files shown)


hatched:
This schema file worked out of the box for me to generate all the old facades for the js libjuju. It created 209 facades. Some of these, like the `uniter` probably aren't necessary. Could that actually be used from outside of Juju? Maybe a blacklist of 'internal only' facades could help trim this down a bit. I was also surprised to see that the latest code still supported instances as far back as `ApplicationV1` 😄 .

hatched:
I noticed that there are a number of invalid records in this schema that aren't in previous versions. You'll note that `$ref` is invalid and only points to `#/definitions/`, not an actual definition. There are 95 of them, they seem to be isolated to the `Uniter` facade and two in `MeterStatus`.

```JSON
"State": {
  "type": "object",
  "properties": {
      "Params": {
          "$ref": "#/definitions/"
      }
  },
  "description": "State isn't on the v1 API."
},
```

wallyworld:
Do we really need all old facades included? It seems overly verbose. What's the use case we're solving? If a Juju client needs such info, I would think that we'd be targetting say Juju 2.8 (or 2.7 at the earliest) and therefore only a much smaller subset of facades would be needed. We could also generate the latest version as we do now and diffs for earlier versions - there's really only very small changes between facade versions.

hatched:
@wallyworld unfortunately JIMM doesn't support the latest facades so if we want to interact with it we need the older facade versions.

Additional information is in the bug report: https://bugs.launchpad.net/juju/+bug/1891404

wallyworld:
Thanks for the bug reference, that extra info is good to know.
IMO let's look at what we need to support (eg Juju versions 2.7 and 2.8) and JIMM and generate only the facades we need.  It would be good not to have the overhead of maintaining a significantly large file full of facade info we don't need, especially with the pylibjuju implications. Ideally we'd upgrade JIMM ASAP to use a more recent Juju version so we avoid generating as many facades.

SimonRichardson:
See: https://github.com/juju/juju/pull/11913

PR #11911, opened 2020-08-18
Rework charm placeholder

SimonRichardson:
## Description of change

The methods with in-state are not just restricted for just the charm
store, but also the charm hub.

It brings in additional changes around the charm url, but requires a
large seismic shift in terms of library and package management.

--- 

The deploy code will still work without a prefix for the moment. We
should follow this up in subsequent PRs so that the deploy code calls
`charm.EnsureSchema` and that all URLs are correctly identified and
classified.

## QA steps

```sh
juju bootstrap lxd test
juju deploy cs:redis
```


SimonRichardson:
https://trello.com/c/4auv79et/2848-refactor-charm-placeholder-functions-in-state

hmlanigan:
$$merge$$

PR #11912, opened 2020-08-18
[2.8] Ensure caas specific charm options pass validation checks

achilleasa:
## Description of change

K8s charms _may_ include CAAS-specific options (e.g. `juju-external-hostname`) which are not part of their config metadata and are not stored to the DB in the charm document. As a result,  when we attempt to deploy the charm again and we provide the CAAS-specific options, juju will load the charm metadata from the DB and the CAAS-specific options will fail the validation check since they are not included in the metadata.

To fix this problem, this PR changes the accessor for the charm config data in the `Charm` state model to query the model type and inject config metadata entries for the CAAS-specific (as obtained from the `caas` package) settings for CAAS models.

As this kind of validation occurs in multiple places (e.g. the application facade does this once in `applicationSetCharmConfigYAML` and then the same check is performed by the state's Application model in `UpdateCharmConfig`), I opted to patch that at the accessor level to avoid any surprises in other parts of the codebase.

## QA steps

```console
# Set up a test bundle with the following contents:
$ cat bundle.yaml
bundle: kubernetes
applications:
  mm-pd-bot:
    charm: cs:~mm-pd-bot-charmers/mm-pd-bot
    scale: 1
    options:
      image_path: stg-is.docker-registry.canonical.com/mm-pd-bot:v1
      tls_secret_name: mm-pd-bot-staging-canonical-com-tls
      # Those parameters are k8s specific
      juju-external-hostname: mm-pd-bot.staging.canonical.com

# Try the following on a 2.8.1 k8s instance
# The second command should fail!
$ juju deploy ./bundle.yaml
$ juju deploy ./bundle.yaml
ERROR cannot deploy bundle: cannot update options for application "mm-pd-bot": setting configuration from YAML: updating settings: unknown option "juju-external-hostname"

# Now set up a k8s instance from this branch
$ juju add-model foo

# All these should proceed without errors
$ juju deploy ./bundle.yaml
$ juju deploy ./bundle.yaml
$ juju deploy ./bundle.yaml

# NOTE: the test bundle is missing a required field so it will eventually error
# It is only used to check that the cass-specific option validation bug has been
# addressed.
```

## Bug reference
https://bugs.launchpad.net/juju/+bug/1891524

### DIFF
--- a/state/charm.go
+++ b/state/charm.go
@@ -16,6 +16,7 @@ import (
 	"gopkg.in/mgo.v2/bson"
 	"gopkg.in/mgo.v2/txn"
 
+	"github.com/juju/juju/caas"
 	"github.com/juju/juju/core/model"
 	"github.com/juju/juju/mongo"
 	mongoutils "github.com/juju/juju/mongo/utils"
@@ -528,7 +529,43 @@ func (c *Charm) Meta() *charm.Meta {
 
 // Config returns the configuration of the charm.
 func (c *Charm) Config() *charm.Config {
-	return c.doc.Config
+	model, err := c.st.Model()
+	if err != nil {
+		logger.Warningf("unable to query model type: %v", err)
+		return c.doc.Config
+	}
+
+	if model.Type() != ModelTypeCAAS {
+		return c.doc.Config
+	}
+
+	// Since the charm metadata comes from the database, it will not include
+	// any CAAS-specific configuration options and will therefore trigger
+	// validation errors if a CAAS-specific settings is provided.
+	//
+	// To avoid this, we inject the known CAAS options into the declared
+	// charm options.
+	//
+	patchedConfig := &charm.Config{
+		Options: make(map[string]charm.Option, len(c.doc.Config.Options)),
+	}
+	for k, v := range c.doc.Config.Options {
+		patchedConfig.Options[k] = v
+	}
+
+	// An error might be returned if we provide additional args
+	// to the function. We can safely ignore the error here.
+	caasCfgSchema, _ := caas.ConfigSchema(nil)
+	caasDefaults := caas.ConfigDefaults(nil)
+	for caasCfgKey, info := range caasCfgSchema {
+		patchedConfig.Options[caasCfgKey] = charm.Option{
+			Type:        string(info.Type),
+			Description: info.Description,
+			Default:     caasDefaults[caasCfgKey],
+		}
+	}
+
+	return patchedConfig
 }
 
 // Metrics returns the metrics declared for the charm.
--- a/state/charm_test.go
+++ b/state/charm_test.go
@@ -692,6 +692,49 @@ func (s *CharmSuite) TestAllCharms(c *gc.C) {
 	c.Assert(charms[2].URL(), gc.DeepEquals, curl2)
 }
 
+func (s *CharmSuite) TestOptionPatchingForCAASCharms(c *gc.C) {
+	st := s.Factory.MakeCAASModel(c, nil)
+	s.CleanupSuite.AddCleanup(func(*gc.C) { st.Close() })
+
+	// Add a deployed k8s charm
+	info := s.dummyCharm(c, "cs:kubernetes/dummy-1")
+	_, err := st.AddCharm(info)
+	c.Assert(err, jc.ErrorIsNil)
+
+	charms, err := st.AllCharms()
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(charms, gc.HasLen, 1)
+
+	// On a CAAS model, validation should succeed for CAAS-specific settings.
+	chCfg := charms[0].Config()
+	_, err = chCfg.ValidateSettings(charm.Settings{
+		"juju-external-hostname": "blah",
+		"juju-application-path":  "/dev/null",
+	})
+	c.Assert(err, jc.ErrorIsNil)
+}
+
+func (s *CharmSuite) TestOptionPatchingForIAASCharms(c *gc.C) {
+	st := s.Factory.MakeModel(c, nil)
+	s.CleanupSuite.AddCleanup(func(*gc.C) { st.Close() })
+
+	// Add a deployed charm
+	info := s.dummyCharm(c, "cs:quantal/dummy-1")
+	_, err := st.AddCharm(info)
+	c.Assert(err, jc.ErrorIsNil)
+
+	charms, err := st.AllCharms()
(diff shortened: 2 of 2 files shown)


achilleasa:
The fix comes with some side-effects which cause some of the existing tests to fail: because of the way the `charm` package generates default values (iterating the option list and populating a map with the default values), the CAAS-specific bits now appear in `juju config` output.

```
# Deploy bundle from QA steps
$ juju deploy bundle.yaml

# Mongo:
> db.settings.find().pretty()
...
{
        "_id" : "b49efcce-1209-4f71-86e2-a3158eeacc64:a#mm-pd-bot#application",
        "model-uuid" : "b49efcce-1209-4f71-86e2-a3158eeacc64",
        "settings" : {
                "kubernetes-ingress-ssl-redirect" : false,
                "kubernetes-ingress-ssl-passthrough" : false,
                "kubernetes-ingress-class" : "nginx",
                "juju-external-hostname" : "mm-pd-bot.staging.canonical.com",  <------ from bundle
                "kubernetes-ingress-allow-http" : false,
                "juju-application-path" : "/",  <----- default value
                "trust" : false
        },
        "version" : NumberLong(0),
        "txn-revno" : NumberLong(2),
        "txn-queue" : [
                "5f3bd6c5fe5c050016454c5b_00d7c2f0"
        ]
}
...

# 
> db.charms.find({},{"config":1}).pretty()
{
        "_id" : "b49efcce-1209-4f71-86e2-a3158eeacc64:cs:~mm-pd-bot-charmers/mm-pd-bot-2",
        "config" : {
                "options" : {
                        "image_path" : {
                                "type" : "string",
                                "description" : "The location of the image to use, e.g. \"registry.example.com/mm-pd-bot:v1\".\n\nThis setting is required.\n",
                                "default" : ""
                        },
                        "image_username" : {
                                "type" : "string",
                                "description" : "The username for accessing the registry specified in image_path.\n",
                                "default" : ""
                        },
                        "image_password" : {
                                "type" : "string",
                                "description" : "The password associated with image_username for accessing the registry specified in image_path.\n",
                                "default" : ""
                        },
                        "mm_pd_bot_cfg" : {
                                "type" : "string",
                                "description" : "The whole configuration file of the bot is passed as content to this setting.\n\nThis setting is required.\n",
                                "default" : null
                        },
                        "tls_secret_name" : {
                                "type" : "string",
                                "description" : "The name of the Kubernetes secret to be associated with the ingress resource.\n\nThis setting is ignored unless juju-external-hostname begins with \"https://\".\n",
                                "default" : ""
                        },
                        "ingress_whitelist_source_range" : {
                                "type" : "string",
                                "description" : "A comma-separated list of CIDRs to store in the ingress.kubernetes.io/whitelist-source-range annotation.\n\nThis can be used to lock down access to Mattermost PagerDuty Bot based on source IP address.\n",
                                "default" : ""
                        }
                }
        }
}
```

When you run `juju config`, you get the CAAS-bits in, but the `juju-external-hostname` value is missing:

```console
# An empty value is returned
$ juju config mm-pd-bot juju-external-hostname

# Deploy again. This bit was fixed by this PR.
$ juju deploy ./bundle.yaml
Resolving charm: cs:~mm-pd-bot-charmers/mm-pd-bot
Executing changes:
- set application options for mm-pd-bot

# Now the config value does indeed show up
juju config mm-pd-bot juju-external-hostname
mm-pd-bot.staging.canonical.com
```

@wallyworld any idea why the first `juju config` returns a blank and why it works after the 2nd deployment? The data seems to be in the settings collection in both cases. Also, do we want to expose these CAAS-specific values via `juju config`. If so, I need to update the broken tests (e.g. [this](https://github.com/juju/juju/blob/2.8/cmd/juju/application/bundle_test.go#L158-L159) one now gets the default values for the CAAS opts and fails)

wallyworld:
Some background:

There's a bunch of config which is considered application config rather than charm config, eg
- juju-external-hostname
- kubernetes-ingress-class
- kubernetes-ingress-ssl-passthrough
- trust
- and a bunch more

All except trust are k8s models only.

These can be specified under the same options block in bundles (or --config deploy arg) - when the key values map arrives server side, it is split up using
`func splitApplicationAndCharmConfig(modelType state.ModelType, inConfig map[string]string)`
or
`func splitApplicationAndCharmConfigFromYAML(modelType state.ModelType, inYaml, appName string)`

The split is done at deploy time or when `juju config` is used to update the app config. After the split, separate app speciic and charm specific schemas are used to validate the charm and application config. Both app and charm config are stored in the settings collection, but under different global keys:
- a#appname#application (for app config)
- a#appname#cs:somecharm (for charm config)

The `db.settings.find().pretty()` shows this (your example shows just the app config).

When the user runs `juju config foo` to print app config, both the above global keys are used to read the settings and they are displayed categorised:

```
$ juju config mariadb-k8s
application: mariadb-k8s
application-config:
  juju-application-path:
    default: /
    description: the relative http path used to access an application
    source: default
    type: string
    value: /
  juju-external-hostname:
    description: the external hostname of an exposed application
    source: user
    type: string
    value: 12.10.0.2.xip.io
  kubernetes-ingress-allow-ht
...
  trust:
    default: false
    description: Does this application have access to trusted credentials
    source: default
    type: bool
    value: false
charm: mariadb-k8s
settings:
  database:
    default: database
    description: The database name.
    source: default
    type: string
    value: database
  mysql_port:
    default: "3306"
    description: The mysql port
    source: default
    type: string
    value: "3306"
...
```

I'll have to dig around to see why a bundle deployment fails second time around. But the expectation is that any incoming key values should be split up just like at the time of original deploy.

state/charm.go
```diff
@@ -528,7 +529,43 @@ func (c *Charm) Meta() *charm.Meta {
 
 // Config returns the configuration of the charm.
 func (c *Charm) Config() *charm.Config {
-	return c.doc.Config
+	model, err := c.st.Model()
```
wallyworld, line 532: This isn't what you want to do.

What you need to do is have the Update() method on the application facade split the config into app and charm config the same as is done for SetApplicationsConfig() on that facade or func deployApplication() in apiserver/facades/client/application/application.go

achilleasa:
@wallyworld Thanks for the clarification. I will trace the code to see why this does not happen in the facade when the charm already exists in the DB. As this is clearly the wrong solution for the bug, I will close this PR and open a new one.

achilleasa:
Looks like, bundlechanges diffing causes the cli to invoke `Update` instead of `DeployApplication` when the charm is already present in the controller. Guess copying the split logic from DeployApplication should do the trick.

PR #11913, opened 2020-08-18
Facade schema groups

SimonRichardson:
## Description of change

This introduces facade schema groups to the schema gen. It defines these
restricted groups so that it makes it easy to understand what the schema
gen generates.

There are 4 groups:

 1. latest: this is the default group and the one that was previously
 running.
 2. all: this outputs all the facades, everything!
 3. client: this outputs just the facades for client interactions and
 probably should actually be the default!
 4. jimm: this is the JIMM facades.

We can tailor each group so that they're defined and restricted to a
given release. We could make all the facades a giant list with versions,
but then it becomes harder to understand because of the many versions.

### Client

```sh
go run github.com/juju/juju/generate/schemagen -admin-facades --facade-group=client \
                ./apiserver/facades/schema.json
```

### JIMM

```sh
go run github.com/juju/juju/generate/schemagen -admin-facades --facade-group=jimm \
                ./apiserver/facades/schema.json
```

### Compose multiple groups

```sh
go run github.com/juju/juju/generate/schemagen -admin-facades --facade-group=client,jimm \
                ./apiserver/facades/schema.json
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1891404

### DIFF
--- a/generate/schemagen/gen/generator.go
+++ b/generate/schemagen/gen/generator.go
@@ -43,11 +43,13 @@ type Option func(*options)
 
 type options struct {
 	adminFacades bool
+	facadeGroup  FacadeGroup
 }
 
 func newOptions() *options {
 	return &options{
 		adminFacades: false,
+		facadeGroup:  Latest,
 	}
 }
 
@@ -55,6 +57,13 @@ func newOptions() *options {
 func WithAdminFacades(adminFacades bool) Option {
 	return func(options *options) {
 		options.adminFacades = adminFacades
+	}
+}
+
+// WithFacadeGroup sets the facadeGroup on the option
+func WithFacadeGroup(facadeGroup FacadeGroup) Option {
+	return func(options *options) {
+		options.facadeGroup = facadeGroup
 	}
 }
 
@@ -78,23 +87,13 @@ func Generate(pkgRegistry PackageRegistry, linker Linker, client APIServer, opti
 		facades = append(facades, adminFacades...)
 	}
 
-	latest := make(map[string]facade.Details)
-	for _, facade := range facades {
-		if f, ok := latest[facade.Name]; ok && facade.Version < f.Version {
-			continue
-		}
-		latest[facade.Name] = facade
-	}
-	latestFacades := make([]facade.Details, 0, len(latest))
-	for _, v := range latest {
-		latestFacades = append(latestFacades, v)
-	}
-	sort.Slice(latestFacades, func(i, j int) bool {
-		return latestFacades[i].Name < latestFacades[j].Name
+	facades = Filter(opts.facadeGroup, facades, registry)
+	sort.Slice(facades, func(i, j int) bool {
+		return facades[i].Name < facades[j].Name
 	})
 
-	result := make([]FacadeSchema, len(latestFacades))
-	for i, facade := range latestFacades {
+	result := make([]FacadeSchema, len(facades))
+	for i, facade := range facades {
 		// select the latest version from the facade list
 		version := facade.Version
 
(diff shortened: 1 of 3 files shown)


generate/schemagen/gen/groups.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func allFacades(facades []facade.Details) []facade.Details {
+	return facades
+}
+
+func clientFacades(facades []facade.Details, registry Registry) []facade.Details {
+	required := map[string]struct{}{
+		"AllWatcher":      struct{}{},
+		"AllModelWatcher": struct{}{},
```
hatched, line 77: The client output appears to be missing these facades:
- Admin
- ModelSummaryWatcher
- Pinger

generate/schemagen/schemagen.go
```diff
@@ -25,7 +25,10 @@ import (
 )
 
 func main() {
-	var adminFacades = flag.Bool("admin-facades", false, "add the admin facades when generating the schema")
+	var (
+		adminFacades = flag.Bool("admin-facades", false, "add the admin facades when generating the schema")
+		facadeGroup  = flag.String("facade-group", "latest", "facade group to export (latest, all, client, jimm)")
```
hatched, line 30: Could this take a list? `facade-groups` Saves us from running it twice and merging the output.

```
go run github.com/juju/juju/generate/schemagen -admin-facades --facade-groups=client,jimm ./apiserver/facades/schema.json
```

SimonRichardson: Sure, was thinking about that...

generate/schemagen/gen/groups.go
SimonRichardson: Will add

generate/schemagen/gen/generator.go
```diff
@@ -78,23 +88,34 @@ func Generate(pkgRegistry PackageRegistry, linker Linker, client APIServer, opti
...
+	// Compose all the facade groups together.
+	var groupFacades [][]facade.Details
+	for _, group := range opts.facadeGroups {
+		groupFacades = append(groupFacades, Filter(group, facades, registry))
+	}
+
+	unique := make(map[string]facade.Details)
+	for _, list := range groupFacades {
+		for _, f := range list {
+			unique[fmt.Sprintf("%s:%d", f.Name, f.Version)] = f
```
hmlanigan, line 100: is it possible to get different Details with the same name and version?

generate/schemagen/gen/groups.go
```diff
@@ -0,0 +1,143 @@
...
+
+func jimmFacades(facades []facade.Details) []facade.Details {
+	required := map[string][]int{
+		"Bundle":              {1},
+		"Cloud":               {1, 2, 3, 4, 5},
+		"Controller":          {3, 4, 5, 6, 7, 8, 9},
+		"ModelManager":        {2, 3, 4, 5},
+		"ModelSummaryManager": {1},
+		"Pinger":              {1},
+		"UserManager":         {1},
```
hmlanigan, line 118: Will these versions need to be updated when a facade version is rev'd?  Please add a comment if needed, or not needed.

SimonRichardson: Yes, yes they will!

generate/schemagen/gen/generator.go
SimonRichardson: Yeah, to ensure we only have one of each, to prevent repetition in the schema.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #11914, opened 2020-08-19
Add abstraction for CLI filesystem access

wallyworld:
## Description of change

When running as embedded CLI commands for the Dashboard, we need to disallow access to the filesystem.
This PR introduces a Filesystem interface on the base command struct, which delegates to the os pkg calls when the commands are run from a terminal, but when embedded the various functions return an error.

A new test is added to the juju/cmd package which parses all of the CLi code and fails if there are any calls to filesystem related functions from the os package. This includes assigning an os function to a variable for later use, and where an import alias may be used for the os pkg.

A followup PR will disable various flags which allow for file access so that the user isn't even given the option in the first place.

## QA steps

bootstrap
run commands which access the filesystem, eg
juju model-config somefile.yaml
juju deploy . --resource foo=bar.zip
juju upgrade-charm foo --path . --resource foo=baz.zip

Now run the commands embedded using https://gist.github.com/wallyworld/4e8de75d5a439bd511e27d9bf5046c31 and ensure they fail.

### DIFF
--- a/cmd/modelcmd/filesystem.go
+++ b/cmd/modelcmd/filesystem.go
@@ -1,0 +1,89 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package modelcmd
+
+import (
+	"io"
+	"os"
+
+	"github.com/juju/errors"
+)
+
+// ReadSeekCloser is a minimal interface used for os.File.
+type ReadSeekCloser interface {
+	io.ReadCloser
+	io.Seeker
+}
+
+// Filesystem is an interface providing access to the filesystem,
+// either delegating to calling os functions or functions which
+// always return an error.
+type Filesystem interface {
+	Stat(name string) (os.FileInfo, error)
+	Open(name string) (ReadSeekCloser, error)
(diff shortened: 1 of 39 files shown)


cmd/modelcmd/filesystem.go
```diff
@@ -0,0 +1,89 @@
...
+// ReadSeekCloser is a minimal interface used for os.File.
+type ReadSeekCloser interface {
+	io.ReadCloser
+	io.Seeker
+}
+
+// Filesystem is an interface providing access to the filesystem,
+// either delegating to calling os functions or functions which
+// always return an error.
+type Filesystem interface {
```
ycliuhw, line 22: probably consider to gomock this for cmd tests in the future

ycliuhw (approved):
lgtm thanks

cmd/modelcmd/filesystem.go
wallyworld: We can mock later when it may be needed. The CLI tests pretty much all set up a fake Juju home and access files from there. So at the moment,, there's no need to mock anything as everything is done in a tmp dir.

wallyworld:
$$merge$$

PR #11916, opened 2020-08-19
[2.8] Ensure caas specific charm options pass validation checks (v2)

achilleasa:
## Description of change

This PR supersedes #11912.

When attempting to deploy a bundle multiple times, the juju cli (via `bundlechanges`) detects that charms are already present in the controller and instead of performing a `DeployApplication` call, it performs an `Update` call (same facade) to update the charm settings.

Prior to this PR, the `Update` codepath would simply try to unmarshal the YAML (if present) and try to validate it via the charm's Config() instance prior to calling `application.UpdateCharmConfig`. This approach has two issues:
- it does not support updating application settings
- validation would fail if the payload contained any application-specific settings (e.g. `juju-external-hostname` et. al for CAAS models).

To resolve this issue, this PR extracts the charm/application config parse logic from the deploy codepath and ensures that both `Update` and `DeployApplication` delegate the config parsing to it. Furthermore, the `Update` implementation will now call both `UpdateApplicationConfig` and `UpdateCharmConfig` depending on whether a non-empty config set was provided for each one.

## QA steps

```console
# Set up a test bundle with the following contents:
$ cat bundle.yaml
bundle: kubernetes
applications:
  mm-pd-bot:
    charm: cs:~mm-pd-bot-charmers/mm-pd-bot
    scale: 1
    options:
      image_path: stg-is.docker-registry.canonical.com/mm-pd-bot:v1
      tls_secret_name: mm-pd-bot-staging-canonical-com-tls
      # Those parameters are k8s specific
      juju-external-hostname: mm-pd-bot.staging.canonical.com

# Try the following on a 2.8.1 k8s instance
# The second command should fail!
$ juju deploy ./bundle.yaml
$ juju deploy ./bundle.yaml
ERROR cannot deploy bundle: cannot update options for application "mm-pd-bot": setting configuration from YAML: updating settings: unknown option "juju-external-hostname"

# Now set up a k8s instance from this branch
$ juju add-model foo

# All these should proceed without errors
$ juju deploy ./bundle.yaml
$ juju deploy ./bundle.yaml
$ juju deploy ./bundle.yaml

# Edit the bundle and change the "juju-external-hostname" value to something else
# Then, deploy again
$ juju deploy ./bundle.yaml

# Verify that the application settings have been correctly updated in mongo
> db.settings.find({_id: {$regex: ".*#application"}}).pretty()
{
        "_id" : "fe99c27b-8c60-406b-8dde-aeffc1413702:a#mm-pd-bot#application",
        "model-uuid" : "fe99c27b-8c60-406b-8dde-aeffc1413702",
        "settings" : {
                "kubernetes-ingress-class" : "nginx",
                "kubernetes-ingress-allow-http" : false,
                "kubernetes-ingress-ssl-redirect" : false,
                "trust" : false,
                "kubernetes-ingress-ssl-passthrough" : false,
                "juju-external-hostname" : "kaboom.staging.canonical.com",  <-------
                "juju-application-path" : "/"
        },
        "version" : NumberLong(1),
        "txn-revno" : NumberLong(3),
        "txn-queue" : [
                "5f3d0b3afe5c050010114c75_a1347752"
        ]

# NOTE: the test bundle is missing a required field so it will eventually error
# It is only used to check that the cass-specific option validation bug has been
# addressed.
```

## Bug reference
https://bugs.launchpad.net/juju/+bug/1891524

### DIFF
--- a/apiserver/facades/client/application/application.go
+++ b/apiserver/facades/client/application/application.go
@@ -492,13 +492,18 @@ func splitApplicationAndCharmConfigFromYAML(modelType state.ModelType, inYaml, a
 	outYaml string,
 	_ error,
 ) {
-	var allSettings map[string]map[string]interface{}
+	var allSettings map[string]interface{}
 	if err := goyaml.Unmarshal([]byte(inYaml), &allSettings); err != nil {
 		return nil, "", errors.Annotate(err, "cannot parse settings data")
 	}
-	settings, ok := allSettings[appName]
+	settings, ok := allSettings[appName].(map[interface{}]interface{})
 	if !ok {
-		return nil, "", errors.Errorf("no settings found for %q", appName)
+		// Application key not present; it might be 'juju get' output.
+		if _, err := charmConfigFromGet(allSettings); err != nil {
+			return nil, "", errors.Errorf("no settings found for %q", appName)
+		}
+
+		return nil, inYaml, nil
 	}
 
 	providerSchema, _, err := applicationConfigSchema(modelType)
@@ -509,8 +514,8 @@ func splitApplicationAndCharmConfigFromYAML(modelType state.ModelType, inYaml, a
 
 	appConfigAttrs := make(map[string]interface{})
 	for k, v := range settings {
-		if appConfigKeys.Contains(k) {
-			appConfigAttrs[k] = v
+		if key, ok := k.(string); ok && appConfigKeys.Contains(key) {
+			appConfigAttrs[key] = v
 			delete(settings, k)
 		}
 	}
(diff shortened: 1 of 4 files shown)


apiserver/facades/client/application/application_test.go
```diff
@@ -1953,9 +1984,9 @@ func (s *applicationSuite) TestApplicationUpdateAllParams(c *gc.C) {
 	// Check the minimum number of units.
 	c.Assert(app.MinUnits(), gc.Equals, minUnits)
 
-	// Check the settings: also ensure the YAML settings take precedence
-	// over strings ones.
-	expectedSettings := charm.Settings{"blog-title": "yaml-title"}
+	// Check the settings: also ensure the string settings take precedence
```
achilleasa, line 1987: This contradicts what we see in the deploy code [here](https://github.com/juju/juju/blob/2.8/apiserver/facades/client/application/application.go#L681-L713). AFAICT, the string settings override the YAML ones. 

I have changed the test to match what we actually see in the code.

@wallyworld Can you clarify whether this is the expected behavior?

apiserver/facades/client/application/application.go
```diff
@@ -492,13 +492,18 @@ func splitApplicationAndCharmConfigFromYAML(modelType state.ModelType, inYaml, a
...
-	var allSettings map[string]map[string]interface{}
+	var allSettings map[string]interface{}
 	if err := goyaml.Unmarshal([]byte(inYaml), &allSettings); err != nil {
 		return nil, "", errors.Annotate(err, "cannot parse settings data")
 	}
-	settings, ok := allSettings[appName]
+	settings, ok := allSettings[appName].(map[interface{}]interface{})
 	if !ok {
-		return nil, "", errors.Errorf("no settings found for %q", appName)
+		// Application key not present; it might be 'juju get' output.
```
achilleasa, line 501: Is this 'juju get' thing the output of a legacy juju version?

apiserver/facades/client/application/application_test.go
wallyworld: I think the code comments are wrong.

```
type ApplicationUpdate struct {
	SettingsStrings map[string]string  `json:"settings,omitempty"`
	SettingsYAML    string             `json:"settings-yaml"` // Takes precedence over SettingsStrings if both are present.
```

```
type ApplicationDeployV6 struct {
	Config           map[string]string              `json:"config,omitempty"`
	ConfigYAML       string                         `json:"config-yaml"` // Takes precedence over config if both are present.
```

In both cases, the YAML is processed first and the config key values then takes precedence.

wallyworld (changes requested):

apiserver/facades/client/application/application.go
```diff
@@ -492,13 +492,18 @@ func splitApplicationAndCharmConfigFromYAML(modelType state.ModelType, inYaml, a
...
+		if _, err := charmConfigFromGet(allSettings); err != nil {
```
wallyworld, line 502: I'm not sure "FromGet" is the most accurate description.
Maybe "FromConfigValues"?

wallyworld (approved):

apiserver/facades/client/application/application.go
achilleasa: This was the original name (charmConfigFromGetYaml). I will rename it as suggested to make it clear what it does.

achilleasa: This and the incorrect comment entries have been addressed via https://github.com/juju/juju/pull/11916/commits/1055b3d5d2ceb86446c3299f0dc04bce6ef30b66

achilleasa:
$$merge$$

achilleasa:
$$merge$$ (TestStatusAllFormats)

PR #11917, opened 2020-08-20
CLI cleanup for embedded cli and juju 1x

wallyworld:
## Description of change

When running `juju clouds` embedded, we don't need messaging about client clouds.
Also, we don't need to carry the code in the Juju client to warn about first run of Juju 2 on a client that did have a Juju 1 CLI. Firstly, it's client only and Juju 1 clients are no longer freely available, and secondly, there's <= 2 (soon to be 0) Juju 1.x installs at known customer sites anyway.

We we also importing providers/all in many places where it was not needed.

## QA steps

Run `juju clouds` standalone and embedded and check "client clouds" message is only on standalone.


wallyworld:
!!build!!

wallyworld:
$$merge$$

PR #11920, opened 2020-08-20
Target latest mongo

SimonRichardson:
## Description of change

The client tests rely on a certain mongo installed for github.
Unfortunately, github actions install their own default. We can't have
that, as we require what's packaged with the apt cache for the OS (there
is better verbiage for that statement!).

Anyway, we brute force our way through the mongo versions and 
remove them from the source list. The code is terrible and annoying at 
the same time. 

I'm still unsure how we requested ubuntu and we ended up with a
bunch of software on it that I didn't ask for!

## QA steps

github actions client tests pass!

### DIFF
--- a/.github/workflows/client-tests.yml
+++ b/.github/workflows/client-tests.yml
@@ -32,7 +32,9 @@ jobs:
       if: (matrix.os == 'ubuntu-latest')
       run: |
         # Remove the default mongo
-        sudo rm /etc/apt/sources.list.d/mongodb-org-4.2.list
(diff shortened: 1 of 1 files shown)


SimonRichardson:
https://trello.com/c/IA4ZbqIq/2863-fix-github-actions-client-tests

manadart (approved):

achilleasa (approved):

SimonRichardson:
$$merge$$

PR #11923, opened 2020-08-20
Add ResolveCharms to charms api and facade

hmlanigan:
## Description of change

As part of adding CharmHub charms to the deploy code.  Consolidate ResolveCharms into the charms api and facades.  This also moves resolution of the charm from the client to the controller for the CharmStore charms as well.  This functionality
 is needed to meet CharmHub requirements.  It also prevents a case where the juju client could find a charm, but the controller could not find it to download and deploy.   We can then exit sooner.

## QA steps

Unit tests should work.  Full QA once this new api call is used by the application deployer and upgrade charm.


api/charms/client.go
```diff
@@ -36,6 +36,55 @@ func (c *Client) IsMetered(charmURL string) (bool, error) {
...
+// ResolveCharms resolves the given charm URLs with an optionally specified
+// preferred channel.
+func (c *Client) ResolveCharms(curls []CharmToResolve) ([]ResolvedCharm, error) {
+	args := params.ResolveCharmsWithChannel{
+		Resolve: make([]params.ResolveCharmWithChannel, len(curls)),
+	}
+	for i, curl := range curls {
+		args.Resolve[i] = params.ResolveCharmWithChannel{
+			Reference: curl.URL.String(),
+			Channel:   string(curl.Channel),
```
SimonRichardson, line 62: Can we pass origin here, might be more useful in the future?

api/charms/client.go
```diff
@@ -36,6 +36,55 @@ func (c *Client) IsMetered(charmURL string) (bool, error) {
...
+			resolvedCharms[i] = ResolvedCharm{Error: r.Error}
+			continue
+		}
+		curl, err := charm.ParseURL(r.URL)
+		if err != nil {
+			resolvedCharms[i] = ResolvedCharm{Error: err}
+		}
+		resolvedCharms[i] = ResolvedCharm{
+			URL:             curl,
+			Channel:         csparams.Channel(r.Channel),
```
SimonRichardson, line 81: This should be origin, will help in the future.

apiserver/facades/client/charms/resolver.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return csClient, nil
+}
+
+type chResolver struct {
+	client *charmhub.Client
+}
+
+// TODO implement me
+func (c *chResolver) ResolveWithPreferredChannel(*charm.URL, csparams.Channel) (*charm.URL, csparams.Channel, []string, error) {
```
SimonRichardson, line 69:
```suggestion
func (c *chResolver) ResolveWithPreferredChannel(*charm.URL, Origin) (*charm.URL, Origin, []string, error) {
```

SimonRichardson (changes requested):
I think we should move to Origin and encapsulate the `cs.Channel` into the `Origin.Channel.Risk`.

hmlanigan:
@SimonRichardson: we need to look at Origin.Channel.Risk, the current values are good for the CharmHub, but are only a subset of the csparams.Channel.

SimonRichardson:
I think the validation function may only need to run for charmhub and not anything else?

apiserver/facades/client/charms/resolver.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+type chResolver struct {
+	client *charmhub.Client
+}
+
+// ResolveWithPreferredChannel call the CharmHub version of
+// ResolveWithPreferredChannel.
+func (c *chResolver) ResolveWithPreferredChannel(*charm.URL, params.CharmOrigin) (*charm.URL, params.CharmOrigin, []string, error) {
+
```
SimonRichardson, line 28: Doing exactly the same thing I'm currently doing :+1:

cmd/juju/application/bundle/bundle.go
```diff
@@ -156,6 +158,7 @@ func BuildModelRepresentation(
 		mod.Applications[appNames[i]].Options = options
 	}
 	// Lastly get all the application constraints.
+	sort.Strings(principalApps)
```
SimonRichardson, line 161: Excellent:+1:

SimonRichardson (approved):

hmlanigan:
$$merge$$

PR #11925, opened 2020-08-21
Fix k8s workload version

wallyworld:
## Description of change

On k8s charms, we were not displaying the workload version properly.
It is meant to be the docker image path, but an update to how that is modelled broke it.
And if the charm explicitly called application-version-set, we were ignoring that.

## QA steps

deploy a k8s charm
juju status shows Version as "mariadb:latest"

`juju exec --unit mariadb-k8s/0 "application-version-set 666"`
juju status shows Version as "666"

## Bug reference

https://bugs.launchpad.net/juju/+bug/1892216

### DIFF
--- a/apiserver/facades/client/client/status.go
+++ b/apiserver/facades/client/client/status.go
@@ -4,7 +4,6 @@
 package client
 
 import (
-	"fmt"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


wallyworld:
$$merge$$

PR #11933, opened 2020-08-25
Feature/new k8s api layer

ycliuhw:
### Checklist

----

## Description of change

Add a new API layer for k8s provider;
split from https://github.com/juju/juju/pull/11836


caas/kubernetes/provider/resources/applier.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+var logger = loggo.GetLogger("juju.kubernetes.provider.resources")
+
+type applier struct {
+	ops []operation
+}
+
+// NewApplier creates a new applier.
+func NewApplier() Applier {
```
wallyworld, line 21: I wonder if "Applier" could be ClusterOperation so it is named like our ModelOperation interface

caas/kubernetes/provider/resources/inteface.go
```diff
@@ -0,0 +1,41 @@
...
+
+//go:generate go run github.com/golang/mock/mockgen -package mocks -destination mocks/resources_mock.go github.com/juju/juju/caas/kubernetes/provider/resources Resource,Applier
+
+const (
+	// JujuFieldManager marks the resource changes were made by Juju.
+	JujuFieldManager = "juju"
+)
+
+// Resource defines methods for manipulating a k8s resource.
+type Resource interface {
```
wallyworld, line 20: My preference is for a service oriented approach where there's entities which are used as args to various service methods (could be varidac for create, delete etc). Entities encapsulate state but not interactions with the cluster.
So there'd be a Resource interface with Clone(), String().
There'd be a Service interface with Apply(), Get(), Delete() - these take one or more entity resources as args.
Then there'd be a DeploymentService, a StatefulsetService, a DaemonsetService etc. These service instances hold a k8s client reference.

k8s itself takes this approach where there's a client that is akin to the service instance and the domain entities are the Deployment or Secret or whatever.

hpidcock (approved):
LGTM thank-you

wallyworld (commented):
Looks good but I have a different view on approach which would be good to at least discuss before landing.

caas/kubernetes/provider/resources/inteface.go
wallyworld: I guess the Applier provides a different level of abstraction / operation grouping which is an alternative approach to take. So maybe my apprehension is misplaced.

hpidcock: Lets chat about this tomorrow.

ycliuhw: As discussed, we keep the current approach for now.

ycliuhw:
$$merge$$

PR #11951, opened 2020-09-01
Fix the storage processing for StatefulSet, Deployment and DaemonSet …

ycliuhw:
## Description of change

Fix the storage processing for StatefulSet, Deployment and DaemonSet resources in k8s application;


caas/kubernetes/provider/application/application.go
```diff
@@ -634,6 +705,7 @@ func (a *app) annotations(config caas.ApplicationConfig) annotations.Annotation
 }
 
 func (a *app) labels() map[string]string {
+	// TODO: add modelUUID for global resources?
```
hpidcock, line 708: we could just have model uuid for everything

hpidcock (approved):
Thanks, this is good work.

PR #11956, opened 2020-09-02
[dev] Use dedicated payloads for the uniter/firewaller OpenedMachinePortRanges calls

achilleasa:
## Description of change

Note: this PR supersedes #11953 and incorporates the review comments from that PR.

Prior to this PR, the `OpenedMachinePortRanges` call for the uniter and the firewaller facades used the same response payload (see #11882 for the firewaller implementation). To support the "expose application endpoints" work, the firewaller needs information about the per-unit opened port ranges per endpoint _and_ the subnet CIDRs that correspond to the space that each endpoint is bound to. As a result, the shared response payload for these calls proved to be inadequate.

This PR introduces the following changes:
- the uniter call has been renamed to `OpenedMachinePortRangesByEndpoint` to better indicate its purpose. It
now has its own dedicated payload type with one less level of indirection.
- the firewaller also gets its own dedicated payload which also includes the required CIDR details.
- the client for the `OpenedMachinePortRanges` firewaller API call has been removed as it was targeting a single machine at a time (something that will change with the upcoming firewaller changes). It will be re-introduced in a future PR with a signature that allows us to receive this information for a list of machine tags.

# QA steps
Not required, the changed API is not wired yet.

### DIFF
--- a/api/firewaller/machine.go
+++ b/api/firewaller/machine.go
@@ -6,7 +6,6 @@ package firewaller
 import (
 	"fmt"
 
-	"github.com/juju/errors"
 	"github.com/juju/names/v4"
 
 	apiwatcher "github.com/juju/juju/api/watcher"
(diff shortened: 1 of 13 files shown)


achilleasa:
@hmlanigan The bumps to V17 for the uniter facade and V6 for firewaller were introduced to the develop branch as part of this stream of work in #11877 and #11882. So it's fine to rename and/or tweak their payloads as they are only used by code in the develop branch.

achilleasa:
$$merge$$

PR #11961, opened 2020-09-03
Add CAASApplicationProvisioner facade.

hpidcock:
## Description of change

Adds CAASApplicationProvisioner facade to provide charm, provisioning info, password setting and unit garbage collection.

The purpose behind API and APIGroup for the apiserver facades is to keep testing of the API and it's required state to just the new facade and not the common facade code.

## QA steps

Unit Tests. Need worker to test thoroughly.

## Documentation changes

N/A

## Bug reference

N/A


cloudconfig/podcfg/image.go
```diff
@@ -15,6 +15,7 @@ import (
 const (
 	JujudOCINamespace = "jujusolutions"
 	JujudOCIName      = "jujud-operator"
+	JujuK8sAgentName  = "k8sagent"
```
ycliuhw, line 18: have we already split `k8sagent` from `jujud-operator`(the build process)?

cloudconfig/podcfg/image.go
```diff
@@ -51,6 +52,18 @@ func GetJujuOCIImagePath(controllerCfg controller.Config, ver version.Number, bu
 	return imageRepoToPath(controllerCfg.CAASImageRepo(), ver)
 }
 
+// GetJujuK8sOCIImagePath returns the jujud oci image path.
```
ycliuhw, line 55: comment needs to be changed

ycliuhw (approved):
lgtm, thanks!

cloudconfig/podcfg/image.go
hpidcock: Yep that was landed earlier

PR #11970, opened 2020-09-07
Add an explicit user to the consume offer params

wallyworld:
## Description of change

The "ApplicationOffers" facade is a controller facade which uses the authenticated user to determine who is performing an api call. For JAAS, the JIMM proxy connects to the controller as admin and so needs to pass in the user as a parameter. The first place where this has become an issue is the ConsumeOfferDetails API call.

This PR adds the user as an explicit arg to the ConsumeOfferDetails call and weaves that user through all the underlying facade methods. Other API calls like Offer() etc will likely need to be migrated in the future.

## QA steps

bootstrap a controller
offer an application
add a new user and grant access to the offer
as that user, with both a new juju client and an old juju client, consume the offer

### DIFF
--- a/api/applicationoffers/client.go
+++ b/api/applicationoffers/client.go
@@ -266,7 +266,17 @@ func (c *Client) GetConsumeDetails(urlStr string) (params.ConsumeOfferDetails, e
 
 	found := params.ConsumeOfferDetailsResults{}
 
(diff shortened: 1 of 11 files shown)


wallyworld:
[cmr consume api optional user](https://trello.com/c/zeZDbdVY/1152-cmr-consume-api-optional-user)

hpidcock (approved):

wallyworld:
$$merge$$

PR #11973, opened 2020-09-07
[dev] Support per endpoint expose parameters in application state model

achilleasa:
# Description of change

Due to a change in the approach for modeling expose parameters, this PR revises the changes from #11963 so that we can specify a list of spaces and/or CIDRs that can access exposed application ports on a **per endpoint** basis.

## QA steps

None needed. The `exposed` flag is still being set as usual and that's the only thing that the firewall worker considers when generating ingress rules.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,7 +47,7 @@ require (
 	github.com/juju/clock v0.0.0-20190205081909-9c5c9712527c
 	github.com/juju/cmd v0.0.0-20200108104440-8e43f3faa5c9
 	github.com/juju/collections v0.0.0-20200605021417-0d0ec82b7271
-	github.com/juju/description/v2 v2.0.0-20200903081432-fe0d99923df9
+	github.com/juju/description/v2 v2.0.0-20200907123847-ca234e95e1e7
 	github.com/juju/errors v0.0.0-20200330140219-3fe23663418f
 	github.com/juju/featureflag v0.0.0-20200423045028-e2f9e1cb1611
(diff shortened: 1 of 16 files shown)


go.mod
```diff
@@ -182,3 +182,5 @@ replace (
 	k8s.io/metrics v0.0.0 => k8s.io/metrics v0.18.6
 	k8s.io/sample-apiserver v0.0.0 => k8s.io/sample-apiserver v0.18.6
 )
+
+replace github.com/juju/description/v2 => /home/achilleas/Workspace/personal/golang/src/github.com/juju/description
```
SimonRichardson: Probably want to fix this :)

state/application.go
```diff
@@ -31,12 +31,39 @@ import (
...
+	// A list of CIDRs that should be able to reach the opened ports
+	// for an exposed application's endpoint.
+	ExposeToCIDRs []string `bson:"to-cidrs,omitempty"`
+}
+
+// AllowTrafficFromAnyNetwork returns true if the exposed endpoint parameters
+// include the 0.0.0.0/0 CIDR.
+func (exp ExposedEndpoint) AllowTrafficFromAnyNetwork() bool {
+	for _, cidr := range exp.ExposeToCIDRs {
+		if cidr == firewall.AllNetworksIPV4CIDR {
```
SimonRichardson, line 59: Shame we don't have a nice abstraction over a predicate here - `if predicate(cidr) {`. Then you can ask better questions about `CIDRs`

SimonRichardson (approved):

achilleasa:
$$merge$$

PR #11975, opened 2020-09-08
Increase AWS SDK retries from 3 to 10; fetch 1000 "offerings" at once

benhoyt:
### Checklist

 - [ ] Added [integration tests](https://github.com/juju/juju/tree/develop/tests) for the PR? **NOTE:** adding integration tests for this would be *very* heavy -- you have to bootstrap loads of stuff in parallel to get it to trigger, and it takes tens of minutes. IMO is change isn't very invasive and my local testing is enough.

## Description of change

Retries are turned on by default with the AWS SDK, but they're very
light: 3 retries with an initial 500ms delay and doubling backoff (so
500ms, 1s, 2s, give up). Whereas the retries in the amz library (in
EC2.query) are much heavier: 10 total attempts with an initial delay of
1s and doubling backoff (with max 1 minute, so 1s, 2, 4, 8, 16, 32, 60,
60, 60 -- about 4 minutes total). I've bumped up the AWS SDK retries to
roughly match the EC2.query ones. This fixes (or at least masks)
https://bugs.launchpad.net/juju/+bug/1888409

We were doing 39 AWS API requests per InstanceTypes call: 1 for zones,
16 for "offerings", 16 for types, 6 for spot price. The types and spot
price are already on maximum # of results per page, but I've increased
the "offerings" one from 100 to 1000. That means: 1 zones, 2 offerings,
16 types, 6 spot price = 25 total = 36% reduction in API calls.

## QA steps

You can verify it fails before the change (you get lots of `RequestLimitExceeded` errors) and works after the change by using Adam Stokes' `repro-juju.bash` script in parallel (I ran it in parallel x8). See: https://bugs.launchpad.net/juju/+bug/1888409

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1888409

### DIFF
--- a/provider/ec2/instancetypes.go
+++ b/provider/ec2/instancetypes.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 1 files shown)


benhoyt:
!!build!!

wallyworld (approved):

benhoyt:
$$merge$$

PR #11977, opened 2020-09-09
Fixes label bug in the addmission worker.

tlm:
LP: 1892285

### Checklist

----

## Description of change

The admission worker has been adding juju-app labels to created objects
from deploy objects. This has been causing K8's selector to incorrectly
target the wrong pods.

The fix here is to change the label we use so that selectors don't
break.

This change doesn't require any upgrade work.

## QA steps

```sh
cd tests
./main.sh caasadmission
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1892285

### DIFF
--- a/caas/kubernetes/provider/labels.go
+++ b/caas/kubernetes/provider/labels.go
@@ -3,6 +3,17 @@
 
 package provider
 
+import (
+	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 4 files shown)


wallyworld (approved):

tlm:
$$merge$$

PR #11979, opened 2020-09-09
Added CAAS application provisioner worker.

hpidcock:
## Description of change

Adds the CAAS application provisioner worker and application watcher with filtering.

## QA steps

Unit Tests

## Documentation changes

N/A

## Bug reference

N/A


worker/caasapplicationprovisioner/application.go
```diff
@@ -0,0 +1,311 @@
...
+
+func (a *appWorker) updateState(app caas.Application, force bool) error {
+	// Fetching the units here is to ensure happens-before consistency
+	// on the deletion of units.
+	observedUnits, err := a.facade.Units(a.name)
+	if err != nil {
+		return errors.Trace(err)
+	}
+	st, err := app.State()
+	if errors.IsNotFound(err) {
```
ycliuhw, line 184: return nil ?

ycliuhw (approved):
LGTM, 
just a few questions, thank you!

caas/kubernetes/provider/application/application.go
```diff
@@ -143,34 +142,52 @@ func (a *app) Ensure(config caas.ApplicationConfig) (err error) {
...
+				Name:        a.name,
+				Namespace:   a.namespace,
+				Labels:      a.labels(),
+				Annotations: a.annotations(config),
+			},
+			Spec: corev1.ServiceSpec{
+				Selector: a.labels(),
+				Type:     corev1.ServiceTypeClusterIP,
+			},
+		},
```
ycliuhw, line 188: We probably don't want to revert an already port opened svc to the initial version in Ensure here.
So the logic here should create first, and ignore the already exist error.
I am currently working on service stuff, so we can leave it as it is for now.

worker/caasapplicationprovisioner/application.go
hpidcock: Returning nil here would prevent the call to force cleanup remaining units.

caas/kubernetes/provider/application/application.go
hpidcock: Yep agreed, I'll leave this to you.

PR #11981, opened 2020-09-10
Ensure embedded commands do not exec commands

wallyworld:
## Description of change

Update the unit test to check that CLI commands do not directly exec things.
Adjust the super command set up to not run plugins in embedded mode.

## QA steps

From the CLI, run a plugin, eg juju metadata

### DIFF
--- a/cmd/package_test.go
+++ b/cmd/package_test.go
@@ -23,40 +23,56 @@ func Test(t *stdtesting.T) {
 	gc.TestingT(t)
 }
 
-var disallowedCalls = set.NewStrings(
-	"Chdir",
-	"Chmod",
-	"Chown",
-	"Create",
-	"Lchown",
-	"Lstat",
-	"Mkdir",
-	"Open",
-	"OpenFile",
-	"Remove",
-	"RemoveAll",
-	"Rename",
-	"TempDir",
-	"Stat",
-	"Symlink",
-	"UserCacheDir",
-	"UserConfigDir",
-	"UserHomeDir",
-)
+var disallowedCalls = map[string]set.Strings{
+	"os": set.NewStrings(
+		"Chdir",
+		"Chmod",
+		"Chown",
+		"Create",
+		"Lchown",
+		"Lstat",
+		"Mkdir",
+		"Open",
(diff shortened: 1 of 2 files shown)


cmd/package_test.go
```diff
@@ -103,22 +124,33 @@ func (s *OSCallTest) parseDir(fset *token.FileSet, calls map[string]set.Strings,
...
 			}
-			s.parseFile(f, fName, osImportAliases, calls)
 		}
 	}
 }
 
-func (*OSCallTest) parseFile(f *ast.File, fName string, osImportAliases set.Strings, calls map[string]set.Strings) {
+func (s *OSCallTest) parsePackageFunctions(ctx *callCheckContext, fName string, f *ast.File) {
+	osImportAliases := set.NewStrings(ctx.pkgName)
+	// Ensure we also capture os calls where the import tis aliased.
```
hpidcock, line 141: tis

hpidcock (approved):
LGTM; Although I have concerns about how shallow this checking is. Really this should be checking all the dependencies of packages and tracing function usage that way.

I can look at doing this in the future, its a little bit tricky, but might be worth making this into a separate tool.

Additionally we should be checking for net.Dial, as you could _theoretically_ dial a unix socket like docker, containerd etc and do some privilege escalation that way.

wallyworld:
The original idea wasn't to be totally pedantic with it, but recognise that juju commands can do things like accept file input for config parameters, or write output to a file etc, and disallow that type of access. In theory there could be something in a transitive dependency but the CLi code is somewhat limited in what it does - it parses args, connects to the Juju APi, prints results etc. And commands whose sole purpose is to read / write files, like add-cloud, add-credential etc, have been disallowed. And other commands have had options which touch files excluded. So the test is really to ensure we don't regress in these checks.

wallyworld:
$$merge$$

wallyworld:
[enhance CLI checks to excluded exec functions for embedded](https://trello.com/c/DDxQNDML/1168-enhance-cli-checks-to-excluded-exec-functions-for-embedded)

PR #11982, opened 2020-09-10
Only fetch pricing for one availability zone and Linux

benhoyt:
### Checklist

----

## Description of change

This should reduce the number of paginated API calls for
instanceTypeCosts() from 6 to 1, and provide much the same result when
we do cost-based comparisons (it's only an estimate using spot prices).

Also rename instanceTypeRegions to instanceTypeZones as these are
actually availability zones, not regions.

## QA steps

Kind of hard to test -- I wrote a little test script to log each API call and call instanceTypeCosts() directly and ran that from the command line to test that API calls were reduced and it provided the same price-ordering as output. Could add a unit test or two, but we wouldn't actually be testing much expect that we got the args correct.

## Bug reference

This is follow-on work from: https://bugs.launchpad.net/juju/+bug/1888409

### DIFF
--- a/provider/ec2/instancetypes.go
+++ b/provider/ec2/instancetypes.go
@@ -203,7 +203,7 @@ func (e *environ) supportedInstanceTypes(ec2Session ec2iface.EC2API, ctx context
 
 	// Query the instance type names for the region and credential in use.
 	var instTypeNames []*string
-	instanceTypeRegions := make(map[string]set.Strings)
+	instanceTypeZones := make(map[string]set.Strings)
 	var token *string
 	for {
 		offeringResults, err := ec2Session.DescribeInstanceTypeOfferings(&ec2.DescribeInstanceTypeOfferingsInput{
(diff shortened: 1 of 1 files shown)


wallyworld (approved):

wallyworld:
Would be good to land this in 2.8

benhoyt:
$$merge$$

benhoyt:
Okay, I'll put up another PR for 2.8 as well.

wallyworld:
Better to land in 2.8 first by rebasing the PR.

benhoyt:
Oh, sorry -- had already typed merge. What difference does doing it that way make?

jameinel:
It means you can land a clean PR into 2.8 based on 2.8, and then merge 2.8 into develop, and not end up with potential conflicts because you are landing "the same kind of code" in 2 different branches independently. If you *want* to be able to merge 2 stable branches and not have to track a list of growing divergence, then it is cleaner to merge into the older branch, and merge forward from the older branch to the newer branch. We can do some graph theory on it, but short answer is that it results in cleaner merges and fewer future conflicts.

wallyworld:
[Reduce # of AWS calls for instance types (e.g., lazily fetch instance types only if needed)](https://trello.com/c/MMpYiNsC/1157-lazily-fetch-instance-types-only-if-needed-to-reduce-of-aws-calls)

PR #11999, opened 2020-09-15
K8s labels dev

tlm:
### Checklist

----

## Description of change

Changes the labels used by Juju for Kubernetes.

Juju now adopts more idiomatic way of adding Kubernetes labels as per k8s docs https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

We are adopting k8s standard labels and also adding juju.is labels such as model.juju.is/name

This change will not retroactively change labels for existing models

## QA steps

Need to perform a lot of bootstrapping and upgrading older controllers.

- We are wanting to make sure that model's pre this change continue to use the older style of kubernetes labeling.
- Need to make sure that newly created models use the new style of labels
- All created objects are correctly labels include a managed-by: juju label

```sh
#Run Unit Tests
cd caas/kubernetes/provider
go test .
```

## Documentation changes

- Needs a discourse post explaining the change

## Bug reference

https://bugs.launchpad.net/juju/+bug/1888513

### DIFF
--- a/caas/kubernetes/provider/admissionregistration.go
+++ b/caas/kubernetes/provider/admissionregistration.go
@@ -25,6 +25,7 @@ func (k *kubernetesClient) getAdmissionControllerLabels(appName string) map[stri
 		utils.LabelsForApp(appName, k.IsLegacyLabels()),
 		utils.LabelsForModel(k.CurrentModel(), k.IsLegacyLabels()),
 	)
+
 	if !k.IsLegacyLabels() {
 		labels = utils.LabelsMerge(labels, utils.LabelsJuju)
 	}
@@ -173,7 +174,7 @@ func (k *kubernetesClient) ensureValidatingWebhookConfigurations(
 			ObjectMeta: metav1.ObjectMeta{
 				Name:        decideNameForGlobalResource(v.Meta, k.namespace),
 				Namespace:   k.namespace,
-				Labels:      k8slabels.Merge(v.Labels, k.getAdmissionControllerLabels(appName)),
+				Labels:      utils.LabelsMerge(v.Labels, k.getAdmissionControllerLabels(appName)),
 				Annotations: k8sannotations.New(v.Annotations).Merge(annotations),
 			},
 			Webhooks: v.Webhooks,
--- a/caas/kubernetes/provider/admissionregistration.go
+++ b/caas/kubernetes/provider/admissionregistration.go
@@ -21,10 +21,14 @@ import (
 )
 
 func (k *kubernetesClient) getAdmissionControllerLabels(appName string) map[string]string {
-	return map[string]string{
-		constants.LabelApplication: appName,
-		constants.LabelModel:       k.namespace,
+	labels := utils.LabelsMerge(
+		utils.LabelsForApp(appName, k.IsLegacyLabels()),
+		utils.LabelsForModel(k.CurrentModel(), k.IsLegacyLabels()),
+	)
+	if !k.IsLegacyLabels() {
+		labels = utils.LabelsMerge(labels, utils.LabelsJuju)
 	}
+	return labels
 }
 
 var annotationDisableNamePrefixKey = constants.AnnotationKey("disable-name-prefix")
@@ -47,7 +51,7 @@ func (k *kubernetesClient) ensureMutatingWebhookConfigurations(
 			ObjectMeta: metav1.ObjectMeta{
 				Name:        decideNameForGlobalResource(v.Meta, k.namespace),
 				Namespace:   k.namespace,
-				Labels:      k8slabels.Merge(v.Labels, k.getAdmissionControllerLabels(appName)),
+				Labels:      utils.LabelsMerge(v.Labels, k.getAdmissionControllerLabels(appName)),
 				Annotations: k8sannotations.New(v.Annotations).Merge(annotations),
 			},
 			Webhooks: v.Webhooks,
--- a/caas/kubernetes/provider/admissionregistration_test.go
+++ b/caas/kubernetes/provider/admissionregistration_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/juju/juju/caas"
 	"github.com/juju/juju/caas/kubernetes/provider"
 	k8sspecs "github.com/juju/juju/caas/kubernetes/provider/specs"
+	"github.com/juju/juju/caas/kubernetes/provider/utils"
 	"github.com/juju/juju/core/application"
 	"github.com/juju/juju/core/status"
 	"github.com/juju/juju/testing"
@@ -42,8 +43,11 @@ func (s *K8sBrokerSuite) assertMutatingWebhookConfigurations(c *gc.C, cfgs []k8s
 	numUnits := int32(2)
 	statefulSetArg := &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:   "app-name",
-			Labels: map[string]string{"juju-app": "app-name"},
+			Name: "app-name",
+			Labels: utils.LabelsMerge(
+				utils.LabelsForApp("app-name", false),
+				utils.LabelsJuju,
+			),
 			Annotations: map[string]string{
 				"juju-app-uuid":                  "appuuid",
 				"juju.io/controller":             testing.ControllerTag.Id(),
@@ -53,12 +57,12 @@ func (s *K8sBrokerSuite) assertMutatingWebhookConfigurations(c *gc.C, cfgs []k8s
 		Spec: appsv1.StatefulSetSpec{
 			Replicas: &numUnits,
 			Selector: &metav1.LabelSelector{
-				MatchLabels: map[string]string{"juju-app": "app-name"},
+				MatchLabels: utils.LabelsForApp("app-name", false),
 			},
 			RevisionHistoryLimit: int32Ptr(0),
 			Template: core.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Labels: map[string]string{"juju-app": "app-name"},
+					Labels: utils.LabelsForApp("app-name", false),
 					Annotations: map[string]string{
 						"apparmor.security.beta.kubernetes.io/pod": "runtime/default",
 						"seccomp.security.beta.kubernetes.io/pod":  "docker/default",
@@ -176,9 +180,13 @@ func (s *K8sBrokerSuite) TestEnsureMutatingWebhookConfigurationsCreate(c *gc.C) 
 
 	cfg1 := &admissionregistration.MutatingWebhookConfiguration{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        "test-example-mutatingwebhookconfiguration",
-			Namespace:   "test",
-			Labels:      map[string]string{"juju-app": "app-name", "juju-model": "test"},
+			Name:      "test-example-mutatingwebhookconfiguration",
+			Namespace: "test",
+			Labels: utils.LabelsMerge(
+				utils.LabelsForApp("app-name", false),
+				utils.LabelsForModel("test", false),
+				utils.LabelsJuju,
+			),
 			Annotations: map[string]string{"juju.io/controller": testing.ControllerTag.Id()},
 		},
 		Webhooks: []admissionregistration.MutatingWebhook{webhook1},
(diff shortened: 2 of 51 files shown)


tlm:
!!build!!

caas/kubernetes/provider/k8s.go
```diff
@@ -1813,14 +1856,17 @@ func (k *kubernetesClient) configureHeadlessService(
...
+				utils.LabelsForApp(appName, k.IsLegacyLabels()),
+				utils.LabelsJuju,
+			),
 			Annotations: k8sannotations.New(nil).
 				Merge(annotations).
 				Add("service.alpha.kubernetes.io/tolerate-unready-endpoints", "true").ToMap(),
 		},
 		Spec: core.ServiceSpec{
-			Selector:                 utils.LabelsForApp(appName),
+			Selector:                 utils.LabelsForApp(appName, k.IsLegacyLabels()),
```
wallyworld, line 1869: Oh I see, sometimes we do want just these.
Perhaps then have SelectorLabelsForApp() and a LabelsForApp() methods to avoid the caller needed to figure out what to do

caas/kubernetes/provider/modeloperator.go
```diff
@@ -100,6 +107,13 @@ func (m *modelOperatorBrokerBridge) Namespace() string {
 	return m.namespace()
 }
 
+func (m *modelOperatorBrokerBridge) IsLegacyLabels() bool {
+	if m.isLegacyLabels == nil {
+		return false
+	}
+	return m.isLegacyLabels()
```
wallyworld, line 114: return m.isLageacyLabels != nil && m.isLegacyLabels()

?

caas/kubernetes/provider/utils/labels.go
```diff
@@ -4,52 +4,131 @@
...
+func LabelSetToSelector(l labels.Set) labels.Selector {
+	return labels.SelectorFromValidatedSet(l)
 }
 
 // LabelsForApp returns the labels that should be on a k8s object for a given
 // application name
-func LabelsForApp(name string) map[string]string {
-	return map[string]string{
-		constants.LabelApplication: name,
+func LabelsForApp(name string, legacy bool) labels.Set {
```
wallyworld, line 56: If we have a SelectorLabelsForApp() as well, we can avoid the boilerpate used in many places to add the LabelsJuju

caas/kubernetes/provider/utils/labels.go
```diff
@@ -4,52 +4,131 @@
...
 // LabelForKeyValue returns a Kubernetes label set for the supplied key value.
-func LabelForKeyValue(key, value string) k8slabels.Set {
-	return k8slabels.Set{
+func LabelForKeyValue(key, value string) labels.Set {
+	return labels.Set{
 		key: value,
 	}
 }
 
+// LabelsMerge
```
wallyworld, line 74: fix me

caas/kubernetes/provider/utils/labels_test.go
```diff
@@ -0,0 +1,54 @@
...
+	jc "github.com/juju/testing/checkers"
+	gc "gopkg.in/check.v1"
+	core "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+
+	"github.com/juju/juju/caas/kubernetes/provider/utils"
+)
+
+type LabelSuite struct {
```
wallyworld, line 18: Are there enough tests here to cover the new methods added?
(sort of related the comment elsewhere about using hand crafted maps in tests)

caas/kubernetes/provider/operator.go
```diff
@@ -32,14 +32,24 @@ import (
 	"github.com/juju/juju/core/watcher"
 )
 
-func operatorLabels(appName string) map[string]string {
-	return map[string]string{constants.LabelOperator: appName}
-}
+const (
+	OperatorAppTarget = "application"
```
wallyworld, line 36: comment

wallyworld (approved):
Small things. I think we can do away with 

```
if !k.IsLegacyLabels() {
	labels = utils.LabelsMerge(selectorLabels, utils.LabelsJuju)
}
```

all over the place by introducing an extra method to get just selector labels for an app.

Other main issue is we should use hand crafted maps in tests to ensure we are getting exactly what we expect. There appears to be perhaps some missing unit test coverage of the new label functions and using hand crafted test data would cover that as well as making it so that the reader can see exactly the labels we expect.

caas/kubernetes/provider/admissionregistration.go
```diff
@@ -21,10 +21,15 @@ import (
 )
 
 func (k *kubernetesClient) getAdmissionControllerLabels(appName string) map[string]string {
-	return map[string]string{
-		constants.LabelApplication: appName,
-		constants.LabelModel:       k.namespace,
+	labels := utils.LabelsMerge(
```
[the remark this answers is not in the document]
tlm: @wallyworld I wasn't the one that added utils. It came through in a PR @hpidcock did. This was just the product of a re-base.

caas/kubernetes/provider/admissionregistration_test.go
```diff
@@ -42,8 +43,11 @@ func (s *K8sBrokerSuite) assertMutatingWebhookConfigurations(c *gc.C, cfgs []k8s
 	numUnits := int32(2)
 	statefulSetArg := &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:   "app-name",
-			Labels: map[string]string{"juju-app": "app-name"},
+			Name: "app-name",
+			Labels: utils.LabelsMerge(
+				utils.LabelsForApp("app-name", false),
+				utils.LabelsJuju,
+			),
```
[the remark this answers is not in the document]
tlm: I don't agree with the assertion. Hard coding expectation is what dragged this PR into a very long time as lots of tests had to be updated. It would be far better and easier to write a single set of tests that check the output from these methods matches a given set of expectations.

Because we use mock's it means we end up testing a lot of what we don't care about i.e above. It would be far better to do the work to assert that what we care about is there and not verbatim check the whole object.

caas/kubernetes/provider/admissionregistration_test.go
```diff
@@ -176,9 +180,13 @@ func (s *K8sBrokerSuite) TestEnsureMutatingWebhookConfigurationsCreate(c *gc.C)
...
-			Name:        "test-example-mutatingwebhookconfiguration",
-			Namespace:   "test",
-			Labels:      map[string]string{"juju-app": "app-name", "juju-model": "test"},
+			Name:      "test-example-mutatingwebhookconfiguration",
+			Namespace: "test",
+			Labels: utils.LabelsMerge(
+				utils.LabelsForApp("app-name", false),
+				utils.LabelsForModel("test", false),
+				utils.LabelsJuju,
+			),
```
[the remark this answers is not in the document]
tlm: See above.

caas/kubernetes/provider/configmap.go
```diff
@@ -18,9 +18,11 @@ import (
 )
 
 func (k *kubernetesClient) getConfigMapLabels(appName string) map[string]string {
-	return map[string]string{
-		constants.LabelApplication: appName,
+	labels := utils.LabelsForApp(appName, k.IsLegacyLabels())
+	if !k.IsLegacyLabels() {
+		labels = utils.LabelsMerge(labels, utils.LabelsJuju)
 	}
```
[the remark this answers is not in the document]
tlm: No it can't as we have then putting a blanket statement on all users of labels for app. It will then trickle down into our selectors for subordinate objects.

caas/kubernetes/provider/ingress.go
```diff
@@ -21,9 +21,11 @@ import (
 )
 
 func (k *kubernetesClient) getIngressLabels(appName string) map[string]string {
-	return map[string]string{
-		constants.LabelApplication: appName,
+	labels := utils.LabelsForApp(appName, k.IsLegacyLabels())
+	if !k.IsLegacyLabels() {
+		labels = utils.LabelsMerge(labels, utils.LabelsJuju)
 	}
+	return labels
```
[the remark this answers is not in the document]
tlm: It wouldn't unfortunately

caas/kubernetes/provider/k8s.go
tlm: It's not that easy. We could build out generic routines for all the object types that exist

caas/kubernetes/provider/utils/labels.go
tlm: Done & Done

caas/kubernetes/provider/utils/labels_test.go
tlm: Done & Done

tlm: Wrong one. Adding some more tests now. Let that slip through.

caas/kubernetes/provider/operator.go
tlm: Done & Done

caas/kubernetes/provider/modeloperator.go
tlm: This is just a protection mechanism for if the func is not defined on the bridge. Open to better suggestions

caas/kubernetes/provider/k8s.go
```diff
@@ -187,6 +191,14 @@ func newK8sBroker(
 		return nil, errors.NotValidf("modelUUID is required")
 	}
 
+	isLegacy, err := utils.IsLegacyModelLabels(
+		newCfg.Config.Name(), k8sClient.CoreV1().Namespaces())
+	if err != nil {
+		logger.Warningf("determining legacy label status for model %s: %v",
+			newCfg.Config.Name(), err)
+		isLegacy = true
```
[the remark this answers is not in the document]
tlm: It used to but there was a test case that cause it to break. Let me take another look.

caas/kubernetes/provider/admissionregistration_test.go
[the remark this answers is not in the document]
hpidcock: @wallyworld I agree with @tlm here, we shouldn't be retesting logic. Once a set of functionality has sufficient unit tests, its uses elsewhere are considered tested. By hard-coding the labels values in every unrelated test, we just shoot ourselves in the foot.

caas/kubernetes/provider/admissionregistration.go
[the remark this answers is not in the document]
hpidcock: @wallyworld the reason there is a utils package here is out of necessity. The caas/kubernetes/provider package is excessive in scope, and therefore any attempt at breaking it up will result in recursive imports. The utils package here serves as an intermediary for things that have no home but can't be in the provider package for the stated reason.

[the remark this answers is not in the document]
hpidcock: Eventually we will get rid of the utils package, but for the meantime the boundaries of what provider functionality should go where is not clear.

caas/kubernetes/provider/admissionregistration_test.go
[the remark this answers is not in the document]
hpidcock: Might be valuable to have one single test per-suite that validates the label map from the result of
```
utils.LabelsMerge(
	utils.LabelsForApp("app-name", false),
	utils.LabelsForModel("test", false),
	utils.LabelsJuju,
)
```

Then we can "trust" it for the rest of the tests in that suite.

caas/kubernetes/provider/k8s.go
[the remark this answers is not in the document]
hpidcock: Yeah I'd like to see this handled. Considering this changes how we update the Kubernetes state, it could be very dangerous if we operated on a namespace without first understanding it.

caas/kubernetes/provider/modeloperator.go
hpidcock: Why would this function not be defined?

tlm:
This has been approved to merge. Future PR will address some of the small points in this PR.

tlm:
$$merge$$

tlm:
$$merge$$

tlm:
$$merge$$

wallyworld:
Superceded by #12065

PR #12000, opened 2020-09-15
Fix static-analysis

SimonRichardson:
## Description of change

The function UserPermission was duplicated in both
`commoncrossmodel.Backend` and `Backend`, which emits a warning. Static
analysis should have picked this up, but for some reason, it hasn't. This
fixes the issue, but we need to work out why it wasn't caught.

## QA steps

```sh
make static-analysis
```


hmlanigan (approved):
I was going to do the same thing.  :-)

SimonRichardson:
$$merge$$

PR #12002, opened 2020-09-15
Add GetCharmURLOrigin to application facade

hmlanigan:
## Description of change

Variation on GetCharmURL to use for charm refresh.  Get the charm URL and the charm's origin.  Required for CharmHub integration.

When refreshing (upgrading) a charm, it is necessary to know the current origin which holds an ID required by the CharmHub.  It also contains the channel track data required to properly update to the correct channel if the channel was specified during Refresh.  If only a Risk specified with --channel, it applies to the Charm's current track.  Track/Risk would be required in --channel to change the track also.

The application facade version has already been rev'd for 2.9.

## QA steps

Not used yet.  Test when wired up with an api and caller.

### DIFF
--- a/apiserver/facades/client/application/application.go
+++ b/apiserver/facades/client/application/application.go
@@ -1329,6 +1329,9 @@ func (api *APIBase) GetCharmURL(args params.ApplicationGet) (params.StringResult
 	return params.StringResult{Result: charmURL.String()}, nil
 }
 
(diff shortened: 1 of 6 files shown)


SimonRichardson (approved):
Amazing :+1:

SimonRichardson:
$$merge$$

PR #12003, opened 2020-09-16
Handle errors in the info response

SimonRichardson:
## Description of change

When the info API returns an error, we should correctly handle it to
highlight the issue to a user of juju.

## QA steps

```sh
$ juju deploy bad
ERROR No charm or bundle with name 'bad'.
```

### DIFF
--- a/charmhub/info.go
+++ b/charmhub/info.go
@@ -5,6 +5,7 @@ package charmhub
 
 import (
 	"context"
+	"strings"
 
 	"github.com/juju/errors"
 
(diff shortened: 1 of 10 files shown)


charmhub/info.go
```diff
@@ -38,5 +39,16 @@ func (c *InfoClient) Info(ctx context.Context, name string) (transport.InfoRespo
...
+
+	if len(resp.ErrorList) > 0 {
+		var combined []string
+		for _, err := range resp.ErrorList {
+			if err.Message != "" {
+				combined = append(combined, err.Message)
+			}
+		}
+		return resp, errors.Errorf(strings.Join(combined, "\n"))
+	}
```
hmlanigan, line 51: This could be a helper function, I believe that each api call will end up with []APIError to deal with.

hmlanigan (commented):
QA works well.

hmlanigan (approved):
Nice!

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #12016, opened 2020-09-17
2.8 upgrade panic placeholder charms 1895954

jameinel:
### Checklist

----

## Description of change

The upgrade step to fix charm Meta 'limits' field didn't handle charms that weren't fully downloaded. When recording a new charm, we start by adding a placeholder (with no Meta), and then we populate that meta by downloading the charm.
This just checks for a nil value, and we know there is no upgrade that we need to do.

## QA steps

I'm not sure how to force one of these into actual data. I suppose you could block traffic from your controller to the charm store, and then try to deploy a charm. You could do that with 2.8.1 and then upgrade to 2.8.2, and see that it panics, but with this patch, it won't panic. (The unit test covers the data that we've seen in the wild.)

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1895954

### DIFF
--- a/scripts/win-installer/setup.iss
+++ b/scripts/win-installer/setup.iss
@@ -2,7 +2,7 @@
 ; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
 
 #define MyAppName "Juju"
-#define MyAppVersion "2.8.2"
+#define MyAppVersion "2.8.3"
 #define MyAppPublisher "Canonical, Ltd"
 #define MyAppURL "http://jujucharms.com/"
 #define MyAppExeName "juju.exe"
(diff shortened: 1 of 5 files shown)


manadart:
We don't want to propose against develop.

I also had this at the bottom of the method to reduce the warning logging:
```go
		if len(ops) == 0 {
			return nil
		}
		return errors.Trace(st.db().RunTransaction(ops))
```

jameinel:
I just realized the Warning is not correct, because placeholders *explicitly* have no meta, so we might check placeholder, and then skip silently

manadart (approved):
Yes.

manadart:
[2.8.2 Upgrade Panic](https://trello.com/c/sfypjVWV/2936-282-upgrade-panic)

petevg (approved):
This looks plausible to me. Approved pending passing tests.

manadart (approved):

jameinel:
$$merge$$

PR #12018, opened 2020-09-17
application Update problematic for charmhub integration and too many ways to update config

hmlanigan:
## Description of change

Update in the application facade is problematic for the CharmHub integration has it allows another way a charm to be upgraded.  Update is only used by the juju client to update charm and application config.  Its full capabilities appear to be a duplicate of SetCharm. A full duplicate of which is no longer required. Changing to only be part of the application facade prior to version 13.  Setting a charm-url via Update is now unsupported.

We update charm and application config with 1 api call if given key value pairs on the cli, and a 2nd way when deploying bundles and when a file is provided on the cli, a third way when deploying charms or upgrading them.  Too confusing.  Consolidating the first 2 ways into 1 from here forward.   The prior methods are left for client compatibility with older versions of juju.

## QA steps

There should be no change to current behavior.  Test client with old versions of juju.  Test old clients with new juju controller.

Deploy bundles with config and use the juju config command.

### DIFF
--- a/api/application/client.go
+++ b/api/application/client.go
@@ -950,9 +950,12 @@ func (c *Client) Consume(arg crossmodel.ConsumeApplicationArgs) (string, error) 
 	return localName, nil
 }
 
-// SetApplicationConfig sets configuration options on an application.
+// SetApplicationConfig sets configuration options on an application and the charm,
+// from the provided map.
+// Note: The name is misleading as charm config is also set.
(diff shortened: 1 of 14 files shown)


SimonRichardson (approved):

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #12020, opened 2020-09-18
Add openstack support for the allocate-public-ip constraint

wallyworld:
## Description of change

The openstack provider supports the allocate-public-ip constraint.
The now deprecated use-floating-ip model config is still supported, but if the new constraint is specified, that take precedence.
With this new constraint, we can now provision openstack machines with/without public IP as needed, rather than having to have a global, model wide setting for all machines.

The openstack provider used to use the value of use-floating-ip to determine whether any public IP addresses should be queried when getting instances. This is no longer possible since it can't be known on a per machine basis at the time of query whether to do that or not, especially when AllInstances() is called. So now we always attempt to fetch any public IP address for each instance.

The other providers which don't yet support this constraint have had it added to the unsupported list. Sadly, we don't surface this to the user if they try and deploy an app or add a machine; we just log a WARNING on the controller. ie unsupported constraints are not considered an error. For bootstrap, we actually have the constraint validator to use so I've added a WARNING there which the user sees written to their terminal when bootstrapping. But it will require invasive API changes to surface anything to the user when deploying or adding a machine.

## QA steps

juju bootstrap lxd --bootstrap-constraints="allocate-public-ip=true"
Creating Juju controller "ian3" on lxd/localhost
WARNING unsupported constraints: allocate-public-ip
...

juju bootstrap microstack --config use-floating-ip=true
WARNING Config attribute "use-floating-ip" is deprecated.
You can instead use the constraint "allocate-public-ip".
,,,

juju bootstrap microstack --bootstrap-constraints="allocate-public-ip=true"
...

In both cases confirm that a floating ip is used.

### DIFF
--- a/caas/kubernetes/provider/constraints.go
+++ b/caas/kubernetes/provider/constraints.go
@@ -17,6 +17,7 @@ var unsupportedConstraints = []string{
 	constraints.Arch,
 	constraints.InstanceType,
 	constraints.Spaces,
+	constraints.AllocatePublicIP,
 }
 
 // ConstraintsValidator returns a Validator value which is used to
--- a/caas/kubernetes/provider/precheck.go
+++ b/caas/kubernetes/provider/precheck.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/juju/errors"
+
 	"github.com/juju/juju/environs"
 	"github.com/juju/juju/environs/context"
 )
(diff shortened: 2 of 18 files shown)


wallyworld:
[openstack support for allocate-public-ip](https://trello.com/c/3rua06iq/1187-openstack-support-for-allocate-public-ip)

hpidcock (approved):
LGTM Thanks

wallyworld:
$$merge$$

PR #12026, opened 2020-09-18
Allow bundle rewrites

SimonRichardson:
## Description of change

The following ensure that we correctly resolve the charm URL in a
bundle, if the bundle contains a non-canonical URL.

For example the following is a bad bundle:

```yaml
applications:
  wordpress:
      charm: wordpress
```

Instead it should be:

```yaml
applications:
  wordpress:
      charm: cs:wordpress
```

That way we always know that the wordpress correctly comes from the
right store. The code is simple and works with the charm hub
integration.

## QA steps

Attempt to deploy a bundle without a `cs:` prefix.

```sh
juju deploy cs:~juju-qa/bundle/bundle-in-candidate-0
```

### DIFF
--- a/acceptancetests/assess_network_health.py
+++ b/acceptancetests/assess_network_health.py
@@ -140,7 +140,7 @@ class AssessNetworkHealth:
             self.existing_series.add(info['series'])
         for series in self.existing_series:
             try:
-                client.deploy('~juju-qa/network-health', series=series,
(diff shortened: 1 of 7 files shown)


SimonRichardson:
https://trello.com/c/TV0voILn/2942-fix-the-issue-where-acceptance-tests-fails-because-of-bundle-urls-missing-a-schema

SimonRichardson:
$$merge$$

PR #12033, opened 2020-09-22
Handle quoting in HardwareCharacteristics to allow strings with spaces

benhoyt:
### Checklist

 - [ ] Added [integration tests](https://github.com/juju/juju/tree/develop/tests) for the PR? *No integration tests, but I've added extensive unit tests to cover the new cases (and the existing tests cover backwards compatibility).*

----

## Description of change

This extends `HardwareCharacteristics.String()` and `ParseHardware()` to handle fields with spaces by quoting strings with spaces in them. For example, `root-disk-source="My Disk"` or `tags="ab c",def`. The `HardwareCharacteristics.String()` function is used when serializing the machine status across the wire (and comes through as a JSON string), so the stringify/parse functions really should be safely roundtrippable.

Note that any field can be quoted when parsing if you want, but `.String()` only adds quotes when the value has a space or comma in it. This means the change is fully backwards compatible with what was produced/parsed before, but it adds support for strings with spaces in.

We were concerned about double escaping/quoting for command line args, but that's not an issue. You just specify the data store (for example) with shell quotes, like so:

```sh
$ juju bootstrap vsphere vs --config 'datastore=Data Store 1' --show-log ...
```

## QA steps

To repro, rename our braixen Vsphere cluster's data store from `datastore1` to `Data Store 1` and then run the following (REMEMBER to rename it back when you're done):

```sh
$ juju bootstrap vsphere vs --config 'datastore=Data Store 1' --show-log --logging-config="<root>=DEBUG;juju.provider.vmware=TRACE" --constraints 'zones=aron.internal'
...
$ juju status -m controller 
Model       Controller  Cloud/Region  Version  SLA          Timestamp
controller  vs          vsphere/QA    2.8.4.1  unsupported  16:23:40+12:00

WARNING invalid hardware info arch=amd64 mem=3584M root-disk=8192M root-disk-source="Data Store 1" for machine {<nil> {<nil> started  22 Sep 2020 16:19:26+12:00 2.8.4.1 } 10.247.5.61 [10.247.5.61] juju-9d1212-0  {<nil> running poweredOn 22 Sep 2020 16:19:37+12:00  } {<nil> idle  22 Sep 2020 16:18:51+12:00  } bionic 0 map[ens192:{[10.247.5.61] 00:50:56:21:ab:af 10.247.0.1 []  true}] map[] mem=3584M zones=aron.internal arch=amd64 mem=3584M root-disk=8192M root-disk-source="Data Store 1" has-vote false map[]}
Machine  State    DNS          Inst id        Series  AZ  Message
0        started  10.247.5.61  juju-9d1212-0  bionic      poweredOn
```

Note: the above output was with a controller including this fix, but a client not including it (2.8 without this PR). But a similar error would happen if both were on 2.8 without the fix (closer to the error in the attached bug).

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1894100

### DIFF
--- a/core/instance/hardwarecharacteristics.go
+++ b/core/instance/hardwarecharacteristics.go
@@ -4,10 +4,12 @@
 package instance
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"strconv"
 	"strings"
+	"text/scanner"
 
 	"github.com/juju/utils/arch"
 )
@@ -40,10 +42,20 @@ type HardwareCharacteristics struct {
 	AvailabilityZone *string `json:"availability-zone,omitempty" yaml:"availabilityzone,omitempty"`
 }
 
+// quoteIfNeeded quotes s (according to Go string quoting rules) if it
+// contains a space or comma, otherwise returns the original string.
+func quoteIfNeeded(s string) string {
+	if strings.IndexAny(s, " ,") < 0 {
+		// No space in string, return as is
+		return s
+	}
+	return strconv.Quote(s)
+}
+
 func (hc HardwareCharacteristics) String() string {
 	var strs []string
 	if hc.Arch != nil {
-		strs = append(strs, fmt.Sprintf("arch=%s", *hc.Arch))
+		strs = append(strs, fmt.Sprintf("arch=%s", quoteIfNeeded(*hc.Arch)))
 	}
 	if hc.CpuCores != nil {
 		strs = append(strs, fmt.Sprintf("cores=%d", *hc.CpuCores))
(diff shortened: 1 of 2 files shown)


core/instance/hardwarecharacteristics.go
```diff
@@ -4,10 +4,12 @@
 package instance
 
 import (
+	"errors"
```
wallyworld, line 7: We prefer juju's errors package

core/instance/hardwarecharacteristics.go
```diff
@@ -99,51 +115,143 @@ func MustParseHardware(args ...string) HardwareCharacteristics {
...
-				continue
-			}
-			if err := hc.setRaw(raw); err != nil {
-				return HardwareCharacteristics{}, err
+		arg = strings.TrimSpace(arg)
+		for arg != "" {
+			var err error
+			arg, err = hc.parseField(arg)
+			if err != nil {
+				return hc, err
```
wallyworld, line 123: this should be errors.Trace(err)
(original code should have been this way)

core/instance/hardwarecharacteristics.go
```diff
@@ -99,51 +115,143 @@ func MustParseHardware(args ...string) HardwareCharacteristics {
...
-		err = hc.setRootDiskSource(str)
 	case "tags":
-		err = hc.setTags(str)
-	case "availability-zone":
-		err = hc.setAvailabilityZone(str)
+		// Tags is a multi-valued field (comma separated)
+		var values []string
+		values, rest, err = parseMulti(rest)
+		if err != nil {
+			return rest, fmt.Errorf("%s: %v", name, err)
```
wallyworld, line 146: errors.Errorf()

wallyworld (approved):
Just a few fixes to errors (use juju/errors)
Good to see that a test was done with an older client

benhoyt:
$$merge$$

PR #12034, opened 2020-09-22
Make `juju status` error a bit more helpful if bootstrap is ongoing

benhoyt:
### Checklist

----

## Description of change

I was confused by this terse error message from `juju status` when I started. When a bootstrap is in progress on another terminal tab, it just says "no API addresses". So I've expanded that a bit.

It's possible it'd be better to add a special `NoAPIAddresses` error type which the CLI looks for, instead of changing the message directly in `juju/api.go` -- please comment if so.

## QA steps

Run a bootstrap, and while it's running type `juju status` in another terminal window. It should say "ERROR: no API addresses" before this change and "ERROR: no controller API addresses; is bootstrap still in progress?" after.

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1896566

### DIFF
--- a/juju/api.go
+++ b/juju/api.go
@@ -58,7 +58,7 @@ func NewAPIConnection(args NewAPIConnectionParams) (_ api.Connection, err error)
 		return nil, errors.Annotatef(err, "cannot work out how to connect")
 	}
 	if len(apiInfo.Addrs) == 0 {
-		return nil, errors.New("no API addresses")
+		return nil, errors.New("no controller API addresses; is bootstrap still in progress?")
 	}
 	// Copy the cache so we'll know whether it's changed so that
 	// we'll update the entry correctly.
--- a/juju/api_test.go
+++ b/juju/api_test.go
@@ -161,7 +161,7 @@ func (s *NewAPIClientSuite) TestWithInfoNoAddresses(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 
 	st, err := newAPIConnectionFromNames(c, "noconfig", "", store, panicAPIOpen)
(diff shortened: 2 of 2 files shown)


benhoyt:
[Make `juju status` error a bit more helpful if bootstrap is ongoing](https://trello.com/c/qSEoJ4te/1193-make-juju-status-error-a-bit-more-helpful-if-bootstrap-is-ongoing)

juju/api.go
```diff
@@ -58,7 +58,7 @@ func NewAPIConnection(args NewAPIConnectionParams) (_ api.Connection, err error)
 		return nil, errors.Annotatef(err, "cannot work out how to connect")
 	}
 	if len(apiInfo.Addrs) == 0 {
-		return nil, errors.New("no API addresses")
+		return nil, errors.New("no controller API addresses; is bootstrap still in progress?")
```
wallyworld, line 61: Yeah, it would be better if the api layer avoid user facing messaging - that's the responsibility of the CLI. So returning a specific error type would be better.

jameinel (commented):
I don't think this was a string that people should be matching against, so I think it is ok to change it.
It is a little risky, so let's target this to develop instead of 2.8. That way we don't have a public string change in a .PATCH release.

benhoyt:
Thanks for the feedback. Closing in favor of https://github.com/juju/juju/pull/12040 -- which is against develop, and uses the technique of a new error type which the CLI looks for.

benhoyt:
[Make `juju status` error a bit more helpful if bootstrap is ongoing](https://trello.com/c/qSEoJ4te/1193-make-juju-status-error-a-bit-more-helpful-if-bootstrap-is-ongoing)

PR #12038, opened 2020-09-22
Create the charmhub refresher

SimonRichardson:
## Description of change

The following updates the charmhub refresher to correctly handle
charmhub vs charmstore function calls. Essentially because we don't have
macaroons then we don't want to end up falling into the
AddCharmWithAuthorization workflow. By removing the special case we can
correctly reason about what the path the charmhub refresher will take.

In an ideal world we would use a better pattern for dealing with this.
Maybe resort to a stragtegy, but generally just embedding the
functionality into two different types also resolves this.

## QA steps

Unsure here?

### DIFF
--- a/cmd/juju/application/refresher/refresher.go
+++ b/cmd/juju/application/refresher/refresher.go
@@ -13,6 +13,7 @@ import (
 	"github.com/juju/errors"
 	"github.com/juju/featureflag"
 
+	commoncharm "github.com/juju/juju/api/common/charm"
 	"github.com/juju/juju/cmd/juju/application/store"
 	"github.com/juju/juju/cmd/juju/application/utils"
 	corecharm "github.com/juju/juju/core/charm"
@@ -42,6 +43,8 @@ type RefresherConfig struct {
 	ForceSeries     bool
 }
 
+// RefresherFn defines a function alias to create a Refresher from a given
+// function.
 type RefresherFn = func(RefresherConfig) (Refresher, error)
 
 type factory struct {
@@ -58,6 +61,7 @@ func NewRefresherFactory(deps RefresherDependencies) RefresherFactory {
 	d.refreshers = []RefresherFn{
 		d.maybeReadLocal(deps.CharmAdder, defaultCharmRepo{}),
 		d.maybeCharmStore(deps.Authorizer, deps.CharmAdder, deps.CharmResolver),
+		d.maybeCharmHub(deps.CharmAdder, deps.CharmResolver),
 	}
 	return d
 }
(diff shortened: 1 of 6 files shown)


cmd/juju/application/refresher/refresher.go
```diff
@@ -116,15 +120,34 @@ func (d *factory) maybeReadLocal(charmAdder store.CharmAdder, charmRepo CharmRep
...
+
+func (d *factory) maybeCharmHub(charmAdder store.CharmAdder, charmResolver CharmResolver) func(RefresherConfig) (Refresher, error) {
+	return func(cfg RefresherConfig) (Refresher, error) {
+		return &charmHubRefresher{
+			baseRefresher: baseRefresher{
+				charmAdder:     charmAdder,
+				charmResolver:  charmResolver,
+				charmURL:       cfg.CharmURL,
+				charmRef:       cfg.CharmRef,
+				channel:        cfg.Channel,
```
hmlanigan, line 146: Can we pass an origin here instead?  To ensure the correct ID is run.  Or do you intend to change that in the follow ups?

cmd/juju/application/refresher/refresher.go
```diff
@@ -280,3 +351,51 @@ type defaultCharmRepo struct{}
...
+
+// Refresh a given charm hub charm.
+// Bundles are not supported as there is no physical representation in Juju.
+func (r *charmHubRefresher) Refresh() (*CharmID, error) {
+	newURL, origin, err := r.ResolveCharm()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	curl, _, err := store.AddCharmFromURL(r.charmAdder, newURL, origin, r.force)
```
hmlanigan, line 353: I'm wondering whether ignoring the returned origin is okay or not.  Where do you know that it has the correct revision and channel for the new charm to be install?

hmlanigan (approved):
A couple of questions, otherwise LGTM.

Verified that upgrading a charm for 2.8.3 and 2.9 controller with this client works.

cmd/juju/application/refresher/refresher.go
SimonRichardson: I have no idea, was going to ask you the exact same question

cmd/juju/application/refresher/refresher.go
SimonRichardson: Follow up

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
https://trello.com/c/G16mvD2d/2953-add-charmhub-refresh-scaffold

PR #12040, opened 2020-09-23
Make CLI error a bit more helpful if bootstrap is in progress

benhoyt:
### Checklist

----

## Description of change

I was confused by this terse error message from `juju status` when I started. When a bootstrap is in progress on another terminal tab, it just says "no API addresses". So I've expanded that a bit (via a new error type which the CLI looks for).

Previous closed PR: https://github.com/juju/juju/pull/12034 (that's against 2.8 and uses an approach that changed the API layer message, which we don't want).

Also a drive-by fix to `RedirectError.Error()` removing an unnecessary `fmt.Sprintf`.

## QA steps

Run a bootstrap, and while it's running type `juju status` in another terminal window. It should say "ERROR: no API addresses" before this change and "ERROR: no controller API addresses; is bootstrap still in progress?" after.

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1896566

### DIFF
--- a/cmd/modelcmd/base.go
+++ b/cmd/modelcmd/base.go
@@ -249,6 +249,9 @@ func (c *CommandBase) NewAPIRoot(
 	if redirErr, ok := errors.Cause(err).(*api.RedirectError); ok {
 		return nil, newModelMigratedError(store, modelName, redirErr)
 	}
+	if _, ok := errors.Cause(err).(*juju.NoAddressesError); ok {
+		return nil, errors.New("no controller API addresses; is bootstrap still in progress?")
+	}
 	return conn, errors.Trace(err)
 }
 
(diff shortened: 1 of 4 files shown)


benhoyt:
[Make `juju status` error a bit more helpful if bootstrap is ongoing](https://trello.com/c/qSEoJ4te/1193-make-juju-status-error-a-bit-more-helpful-if-bootstrap-is-ongoing)

cmd/modelcmd/base.go
```diff
@@ -249,6 +249,9 @@ func (c *CommandBase) NewAPIRoot(
 	if redirErr, ok := errors.Cause(err).(*api.RedirectError); ok {
 		return nil, newModelMigratedError(store, modelName, redirErr)
 	}
+	if _, ok := errors.Cause(err).(*juju.NoAddressesError); ok {
```
wallyworld, line 252: We normally also define a IsNoAddressesError() helper to use
And the test becomes c.Assert(err, jc.Satisfies, juju.IsNoAddressesError)
We could then get away with not exporting the error, just the helper

wallyworld (approved):
A small fix before landing

benhoyt:
$$merge$$

PR #12046, opened 2020-09-23
Handle no application info for workload status

SimonRichardson:
## Description of change

Although theoretically impossible in a normal workload setup, it has
shown up in tests. The idea that an application wouldn't exist, but have
a unit seems very implausible.
How ever implausible it's shown up in testing and if it can happen in
testing, it can show up in the wild.

The fix is really simple, if no application exists, we can safely ignore
the workload status and move on.

## QA steps

Remove the if nil check and the tests will fail.

## Bugs

https://bugs.launchpad.net/juju/+bug/1896658


hmlanigan (approved):
LGTM, ta!

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #12049, opened 2020-09-24
K8s SA roles

wallyworld:
## Description of change

We were restricting the primary k8s service account for a workload pod to only have one role.
This PR removes that restriction.
The common validation code is moved so that it is shared between the primary service account struct and the additional service account struct, ie ServiceAccountSpecV3
Also, setting the primary service account name no longer sets the role name - the user can specify their own or it defaults to the service account name. ie the same behaviour as for additional service accounts.

## QA steps

Deploy a k8s charm with a primary service account with > 1 role like in the bug.
Use kubectl to check the roles and role bindings.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1896076

### DIFF
--- a/caas/specs/rbac.go
+++ b/caas/specs/rbac.go
@@ -4,6 +4,9 @@
 package specs
 
 import (
+	"fmt"
+
+	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 )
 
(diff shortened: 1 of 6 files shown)


wallyworld:
[k8s service account roles](https://trello.com/c/yHzRjMxi/1195-k8s-service-account-roles)

caas/specs/rbac.go
```diff
@@ -78,36 +81,38 @@ func (sa ServiceAccountSpecV3) Validate() error {
...
+		// All good.
+		return nil
+	}
+	return errors.NewNotValid(nil, "either all or none of the roles should have a name set")
 }
 
 // PrimeServiceAccountSpecV3 defines spec for creating the prime RBAC resources for version 3.
 type PrimeServiceAccountSpecV3 struct {
-	name                 string
+	Name                 string
```
ycliuhw, line 106: we now allow charm to set primary `SA` name?

ycliuhw (approved):
lgtm thanks

caas/specs/rbac.go
wallyworld: ah good point, I think I should go back to making it unexported but not have the SetName() also set the role name

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #12058, opened 2020-09-28
Explain why an application name is not valid

SimonRichardson:
## Description of change

The following changes are rather trivial, but the idea is to explain to
the user why an application name isn't valid. It attempts to do this in
plain English where possible.

Rather than falling back to just telling the user about an unexpected
character, it first explains that uppercase characters aren't valid,
then numbers aren't allowed after hyphens. Then if there is something
else wrong, explain which character is at fault.

Hopefully, this will explain the real issue to the user when attempting
to interacting with juju.

## QA steps

```sh
$ juju bootstrap lxd test

$ juju deploy redis redis-01
ERROR invalid application name "redis-1", unexpected number(s) found after hyphen

$ juju deploy redis Redis
ERROR invalid application name "Redis", unexpected uppercase character

$ juju deploy redis red£is
ERROR invalid application name "red£is", unexpected character £
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1891418


cmd/juju/application/utils/names.go
```diff
@@ -0,0 +1,54 @@
...
+	return nil
+}
+
+// validApplicationNameRune works out if there is a valid application rune.
+func validApplicationNameRune(r rune) bool {
+	if (r >= 'a' && r <= 'z') || unicode.IsNumber(r) || r == '-' {
+		return false
+	}
+	return true
+}
```
petevg, line 54: I'm a little bit torn on this.

On the one hand, I'd rather that IsValidApplication returned a boolean and an error string, so that you'd have access to the reason for the failure everywhere that we call this. The code at this level of abstraction thus becomes much more compact: check for the error and return it if we find it, rather than doing forensics on why the error might be occurring.

That approach would put the hand translation of the regular expression in names/application.go closer to the regular expression itself. I'd even be tempted to factor out the regular expression. I'm not sure that the cpu cycles that it saves are worth the levels of indirection in telling the user about failures.

On the other hand, changing the return of IsValidApplication requires touching a lot of code, in a lot of packages. I understand the temptation to just put the wrapper here.

I like the change, overall. I just have mixed feelings about a wrapper being appropriate, here.

SimonRichardson: Changing `IsValidApplication` is a pure no-go. I might as well re-write juju. 
Instead, I could lift this into the names package, that way we can share why things may fail and hopefully it may become a pattern for other name tags in that repo.

petevg: @SimonRichardson +1 to moving this wrapper to names. That makes sense, and i think abstracts things more cleanly.

cmd/juju/application/utils/names.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/juju/names/v4"
+)
+
+var alphaChar = regexp.MustCompile("[A-Z]")
+var tailNumberSuffix = regexp.MustCompile("-[0-9]+$")
+
+// ValidateApplicationName takes a name and attempts to validate the application
+// name, before returning a reason why it's not valid.
+func ValidateApplicationName(name string) error {
+	if !names.IsValidApplication(name) {
```
achilleasa, line 21: You could invert the check to return early if it is valid and remove one level of indentation from the error checking code.

cmd/juju/application/utils/names.go
```diff
@@ -0,0 +1,54 @@
...
+		// If the application has uppercase characters, bail out and explain
+		// why.
+		if alphaChar.MatchString(name) {
+			return errors.Errorf("invalid application name %q, unexpected uppercase character", name)
+		}
+		// If the application ends up being suffixed by a number, then we want
+		// to mention it to users why.
+		if tailNumberSuffix.MatchString(name) {
+			return errors.Errorf("invalid application name %q, unexpected number(s) found after hyphen", name)
```
achilleasa, line 30:
```suggestion
			return errors.Errorf("invalid application name %q, unexpected number(s) found after last hyphen", name)
```

cmd/juju/application/utils/names.go
```diff
@@ -0,0 +1,54 @@
...
+		// the first one.
+		invalidRune := []rune(name[index:])[0]
+		return errors.Errorf("invalid application name %q, unexpected character %c", name, invalidRune)
+	}
+
+	return nil
+}
+
+// validApplicationNameRune works out if there is a valid application rune.
+func validApplicationNameRune(r rune) bool {
```
achilleasa, line 49: The name of this function is confusing when you consider the path that returns `true`. Should this actually be `invalidRuneForApplicationName` or something like that?

achilleasa (approved):

SimonRichardson:
$$merge$$

PR #12061, opened 2020-09-29
[dev] Fetch costs for ALL linux spot instances on ec2

achilleasa:
## Description of change

The instance cost lookup code for EC2 attempts to reduce the number of retrieved results by filtering spot instances and retaining the ones that can deploy linux instances. However, according to the API [docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSpotPriceHistory.html), besides `Linux/UNIX` there is also an `Linux/UNIX (Amazon VPC)` option.

The PR adds the second option to ensure that we don't accidentally leave out any VPC-specific instances from the cost list.

## QA steps

Bootstrap a controller and deploy a workload; check the instance types that you got from EC2

### DIFF
--- a/provider/ec2/instancetypes.go
+++ b/provider/ec2/instancetypes.go
@@ -372,8 +372,11 @@ func instanceTypeCosts(ec2Session ec2iface.EC2API, instTypeNames []*string, zone
 			// Only look at Linux results (to reduce total number of results;
 			// it's only an estimate anyway)
 			{
-				Name:   aws.String("product-description"),
-				Values: []*string{aws.String("Linux/UNIX")},
+				Name: aws.String("product-description"),
+				Values: []*string{
+					aws.String("Linux/UNIX"),
+					aws.String("Linux/UNIX (Amazon VPC)"),
+				},
 			},
 		},
 	}
--- a/provider/ec2/instancetypes.go
+++ b/provider/ec2/instancetypes.go
@@ -173,26 +173,29 @@ func archName(in string) string {
 	return in
 }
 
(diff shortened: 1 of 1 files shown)


provider/ec2/instancetypes.go
```diff
@@ -173,26 +174,39 @@ func archName(in string) string {
...
+	if len(e.instTypes) == 0 {
+		instTypes, err := e.collectSupportedInstanceTypes(ec2Session, ctx)
+		if err != nil {
+			return nil, err
+		}
+		e.instTypes = instTypes
+
+		var filtList []instances.InstanceType
+		for _, it := range e.instTypes {
+			if !it.Deprecated && it.Cost < math.MaxUint64 {
```
SimonRichardson, line 193: I think from what @jameinel was saying; we do want to include the `math.MaxUint64` because we're using spot instance types as a corollary for on-demand prices (which is totally wrong, but we have what we have), then you do want to include the instance types which have no cost.

That way you can say I want an instance type with 128GB, you'll end up with an instance type, that might not have a cost, but you do at least end up with an instance of said constraints.

The real fix here is to find better pricing API calls, yet I believe that's expensive in terms of the number of API calls.

SimonRichardson (changes requested):

provider/ec2/instancetypes.go
achilleasa: @SimonRichardson ah crap! that was a debug log from my tests and should not be in there :-(
Sorry, let me amend that.

SimonRichardson (approved):

achilleasa:
$$merge$$

jameinel:
This also needs to be landed in the 2.8 branch, as the change to do filtering was done in 2.8 so that we don't do as many API requests against AWS's API.

PR #12062, opened 2020-09-29
fix lp:1895040, run upgrade-charm operation after uniter loop restart if one is pending

hmlanigan:
## Description of change

During SetCharm, if there are new peer relations, they are added to the relation collection at the same time as the new charm URL is updated in the applications collection.  There is a race between these two events reaching the unit to be handled.  If the new peer relation is received, or acted upon, before the upgrade charm event, the unit will start an error loop.  The peer relation isn't found in the local copy read from disk of the charm's metadata.yaml.  

Upgrade charm is usually done when there is a lull in charm activity.  To not drastically change behavior, check to see if upgrade charm is in the queue when the uniter loop starts up.  If so run upgrade charm before continuing.  Other conditions for running upgrade-charm are: one is not already in progress, the unit has been started.

The verifyCharmProfileResolver cannot be used here as it creates an import circle.

## QA steps

```console
# run twice, once with an lxdprofile and once without.

# setup charms for test
$ cd /tmp
$ charm pull cs:~jameinel/ubuntu-lite-7
$ charm pull cs:~jameinel/ubuntu-lite

# Add the following text to the end of /tmp/ubuntu-lite-7/metadata.yaml
peers:
  peer:
    interface: ubuntu-peer

# 
$ juju deploy -n 5 /tmp/ubutu-lite

# wait for units to be up and running
$ juju upgrade-charm --path /tmp/ubutu-lite-7 ubuntu-lite
```

Run these steps on a prior version of 2.8, see the error, upgrade to this patch, the error should resolve.

Run charm upgrade with other charms, there should be no change in current behavior.
## Bug reference

https://bugs.launchpad.net/juju/+bug/1895040

### DIFF
--- a/worker/uniter/resolver/loop.go
+++ b/worker/uniter/resolver/loop.go
@@ -7,6 +7,7 @@ import (
 	"github.com/juju/charm/v7/hooks"
 	"github.com/juju/errors"
 
+	"github.com/juju/juju/core/lxdprofile"
 	"github.com/juju/juju/worker/fortress"
 	"github.com/juju/juju/worker/uniter/operation"
 	"github.com/juju/juju/worker/uniter/remotestate"
@@ -19,6 +20,7 @@ var logger interface{}
 // Logger represents the logging methods used in this package.
 type Logger interface {
 	Errorf(string, ...interface{})
+	Debugf(string, ...interface{})
 	Tracef(string, ...interface{})
 }
 
@@ -67,6 +69,13 @@ func Loop(cfg LoopConfig, localState *LocalState) error {
 
 	// Initialize charmdir availability before entering the loop in case we're recovering from a restart.
 	err := updateCharmDir(cfg.Logger, cfg.Executor.State(), cfg.CharmDirGuard, cfg.Abort)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	// If we're restarting the loop, ensure any pending charm upgrade is run
+	// before continuing.
+	err = checkCharmUpgrade(cfg.Logger, cfg.Watcher.Snapshot(), rf, cfg.Executor)
 	if err != nil {
 		return errors.Trace(err)
 	}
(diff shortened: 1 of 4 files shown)


worker/uniter/resolver/loop.go
```diff
@@ -167,3 +176,53 @@ func updateCharmDir(logger Logger, opState operation.State, guard fortress.Guard
...
+	if remote.CharmProfileRequired {
+		if remote.LXDProfileName == "" {
+			return nil
+		}
+		rev, err := lxdprofile.ProfileRevision(remote.LXDProfileName)
+		if err != nil {
+			return errors.Trace(err)
+		}
+		if rev != remote.CharmURL.Revision {
+			logger.Tracef("Charm profile required: current revision %d does not match new revision %d", rev, remote.CharmURL.Revision)
```
SimonRichardson, line 211: Should this be trace?

worker/uniter/uniter.go
```diff
@@ -297,48 +296,9 @@ func (u *Uniter) loop(unitTag names.UnitTag) (err error) {
...
-		charmURL = curl
-		app, err := u.unit.Application()
-		if err != nil {
-			return errors.Trace(err)
-		}
-		charmModifiedVersion, err = app.CharmModifiedVersion()
-		if err != nil {
-			return errors.Trace(err)
-		}
+	canApplyCharmProfile, charmURL, charmModifiedVersion, err := u.charmState()
```
SimonRichardson, line 299: Great, makes this loop less intimidating.

SimonRichardson (commented):
Code is good, doing Q&A now.

worker/uniter/resolver/loop.go
hmlanigan: It's the same as in verifyCharmProfileResolver.  Perhaps both should be Debug?

SimonRichardson: Yeah, seems like it would be valuable to know that a charm profile is required without going to a trace logging.

SimonRichardson (approved):
Q&A steps worked as described.

hmlanigan:
$$merge$$

PR #12065, opened 2020-09-30
Improve k8s labels

wallyworld:
## Description of change

This adds extra commits to #11999 (see the last 3 commits).

An extra label helper was added to avoid boilerplate dealing with app labels vs selectors.
A deleted test was added back and skipped so we can fix later.
Also replaced the function calls in tests with literal values.

## QA steps

See original PR

### DIFF
--- a/caas/kubernetes/provider/admissionregistration_test.go
+++ b/caas/kubernetes/provider/admissionregistration_test.go
@@ -18,7 +18,6 @@ import (
 	"github.com/juju/juju/caas"
 	"github.com/juju/juju/caas/kubernetes/provider"
 	k8sspecs "github.com/juju/juju/caas/kubernetes/provider/specs"
-	"github.com/juju/juju/caas/kubernetes/provider/utils"
 	"github.com/juju/juju/core/application"
 	"github.com/juju/juju/core/status"
 	"github.com/juju/juju/testing"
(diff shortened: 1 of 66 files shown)


caas/kubernetes/provider/admissionregistration_test.go
```diff
@@ -43,7 +44,7 @@ func (s *K8sBrokerSuite) assertMutatingWebhookConfigurations(c *gc.C, cfgs []k8s
 	statefulSetArg := &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:   "app-name",
-			Labels: map[string]string{"juju-app": "app-name"},
+			Labels: utils.LabelsForApp("app-name", false),
```
ycliuhw, line 47: For test assertion, it's better to use map[string]string{} rather than the computed result

ycliuhw (approved):
lgtm.
As discussed, we need to revert the labels in tests from computed one to a map of labels in the following pr.

caas/kubernetes/provider/utils/labels_test.go
```diff
@@ -0,0 +1,290 @@
...
+func (l *LabelSuite) TestSelectorLabelsForApp(c *gc.C) {
+	tests := []struct {
+		AppName        string
+		ExpectedLabels labels.Set
+		Legacy         bool
+	}{
+		{
+			AppName: "tlm-boom",
+			ExpectedLabels: labels.Set{
+				"app.kubernetes.io/name": "tlm-boom",
```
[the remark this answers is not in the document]
wallyworld: these are "selector" labels, used to label a pod to match them with a service

caas/kubernetes/provider/admissionregistration_test.go
wallyworld: done

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #12075, opened 2020-10-01
[dev] Enable IPV6 CIDR firewall rule support for compliant providers

achilleasa:
## Description of change

This PR builds on top of the work introduced in #12063 and enables support for IPV6 CIDRs in firewall rules for EC2 and openstack.

In addition, the PR includes a fix for a potential null error value dereference when checking for space support in lxd that was accidentally introduced in #12068.

### Verify IPv6 CIDR support

Run the following on EC2 and openstack. Then check the firewall rules (for ec2 you can use the UI)
and verify that both IPV4 and IPV6 CIDRs are present.

```console
$ juju deploy apache2
$ juju run --unit apache2/0 "open-port 8080/tcp"
$ juju expose apache2 --to-cidrs 192.168.0.0/24,2002::1234:abcd:ffff:c0a8:101/64
```

### Verify LXD fix

```console
$ juju bootstrap lxd test
$ juju spaces
```

### DIFF
--- a/provider/lxd/environ_network_test.go
+++ b/provider/lxd/environ_network_test.go
@@ -180,7 +180,7 @@ func (s *environNetSuite) TestSubnetDiscoveryFallbackForOlderLXDs(c *gc.C) {
 	// Even though ovs-br0 is returned by the LXD API, it is *not* bridged
 	// into the container we will be introspecting and so this subnet will
 	// not be reported back. This is a caveat of the fallback code.
-	srv.EXPECT().GetNetworkNames().Return([]string{"lo", "ovsbr0", "lxdbr0"}, nil)
+	srv.EXPECT().GetNetworkNames().Return([]string{"lo", "ovsbr0", "lxdbr0"}, nil).AnyTimes()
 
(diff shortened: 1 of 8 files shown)


provider/lxd/environ_network_test.go
```diff
@@ -241,6 +241,13 @@ func (s *environNetSuite) TestSubnetDiscoveryFallbackForOlderLXDs(c *gc.C) {
 	env := s.NewEnviron(c, srv, nil).(*environ)
 
 	ctx := context.NewCloudCallContext()
+
+	// Spaces should be supported
+	supportsSpaces, err := env.SupportsSpaces(ctx)
```
achilleasa, line 246: This bit was why I didn't catch this yesterday. The original implementation just returned "true, nil" so there was no test for it. I added the checks for Subnets() but forgot to add one for SupportsSpaces (which is the one that triggers the NPE when you run `juju spaces`)

hmlanigan (approved):
All 3 QA tests went well.

LGTM.

achilleasa:
$$merge$$

PR #12081, opened 2020-10-02
Pin the transent dependency

SimonRichardson:
## Description of change

The following dependency `google_auth` changes what it installs
depending on the version of python you have. I'm unsure of the real fix
here, as multidict doesn't seem to compile on anything by amd64. So
pinning it to use an older version seems a bit bruteforce'ish and we're
not getting the latest spec/security changes from that lib.

I think someone needs to come through and audit what is being installed
here and correctly pin them so we don't get blocked on a release.


manadart (approved):

SimonRichardson:
DO NOT MERGE...

PR #12088, opened 2020-10-05
[dev] Add integration tests for expose-application endpoints feature

achilleasa:
## Description of change

This PR adds two integration tests to verify that the _expose application endpoints_ feature works as expected. More specifically, the tests are designed to verify:
- that the `open-ports` and `opened-ports` hook tools work correctly when the `--endpoints` argument is provided or missing.
- that expose application `--to-cidrs` invocations work as expected (i.e. additive for different endpoints and last write wins when targeting specific endpoints multiple times)
- the juju cli emits an error when attempting to deploy a bundle containing both the `expose: true` flag and a list of endpoint-specific expose settings in an overlay.

These tests target ec2 specifically because that's one of the providers that supports ingress rules with IPV6 CIDRs. This allows us to additionally verify that the IPV6 rules are properly handled by ec2.

## QA steps

```console
$ ./main.sh -p aws expose_ec2
```

### DIFF
--- a/tests/main.sh
+++ b/tests/main.sh
@@ -32,19 +32,23 @@ import_subdir_files includes
 
 # If adding a test suite, then ensure to add it here to be picked up!
 TEST_NAMES="static_analysis \
+            agents \
             appdata \
             branches \
             caasadmission \
             cli \
             controller \
             deploy \
+            expose_ec2 \
             hooks \
-            hook_tools \
+            hooktools \
             machine \
             manual \
+            model \
+            network \
             relations \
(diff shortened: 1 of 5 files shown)


SimonRichardson:
Can you apply this patch: https://paste.ubuntu.com/p/5SpjZtYPnt/ it fixes the help document (although not your doing, would be good as a driveby).

SimonRichardson (approved):
Thanks for the changes, works :+1:

achilleasa:
$$merge$$

PR #12091, opened 2020-10-06
Fix jujuc for new k8s charms.

hpidcock:
## Description of change

Hook tool symlinks now use jujuc instead of k8sagent (as it doesn't support hook tools).

## QA steps

Deploy v2 charm and `juju run app/0 status-get`

## Documentation changes

N/A

## Bug reference

N/A


wallyworld (approved):

hpidcock:
$$merge$$

PR #12096, opened 2020-10-06
Azure encrypted disk

wallyworld:
## Description of change

When provisioning an Azure instance, it is now possible to ask for root disk encryption using a disk encryption set and a key vault.  The SDK API versions needed to be bumped so there was also some compatibility fallout to address.

There's 2 commits:
1. Enhance the start instance provisioning info api call to set root disk volume params if the "root-disk-source" constraint points to a suitable storage pool.
2. The Azure provider acts on the root disk start instance params to set up a disk encryption set and key vault.

You can either BYO disk encryption set or have Juju create a key vault and disk encryption set.
Right now, the key size and algorithm is not user configurable. It's RSA and 4096 bits.

The configurable parameters are:
 - encrypted (must be true for Juju to act on it)
 - disk-encryption-set-name (defaults to vault name prefix)
 - vault-name-prefix
 - vault-key-name (defaults to "disk-secret")
 - vault-user-id

If just the disk encryption set name is configured, Juju will use that existing disk encryption set and not create any addition resources. It can reside in a different resource group by specifying resgroup_name/des_name

If a vault name prefix is also configured, Juju will create a vault called vault_name_prefix-resource_group_name. Vault names need to be unique across a subscription.

If vault-user-id is configured, that user will be given full access to manage the vault and keys created by Juju.

TODO: allow the initial bootstrapped controller to have disk encryption specified

## QA steps

bootstrap Azure

To use a BYO disk encryption set
```
juju create-storage-pool byo azure encrypted=true disk-encryption-set-name="mydisksecrets"
juju add-machine --constraints="root-disk-source=byo"
```

```
juju create-storage-pool des azure encrypted=true vault-name-prefix=mysecrets vault-user-id=27f86c31-bca0-4a37-b233-0e5030107696
juju add-machine --constraints="root-disk-source=des"
```

The above storage pool could also have had disk-encryption-set-name or vault_key-name set.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1892485


hpidcock (approved):

wallyworld:
$$merge$$

PR #12115, opened 2020-10-09
Fix deployment of IAAS charms

wallyworld:
This PR reverts a charm validation check introduced when the k8s spike was merged.
An extra check of charm series was done based on the metadata version. The point at which the check was done was before the charm was available. However, the exact same check is done in state/AddCharm() so it is not possible to write a bad charm to state. So there's no real need to do this particular check client side.
The changes here are to essentially revert some of the recent changes, which were not really specifically unit tested, so there's no test fallout. The validateCharmSeriesWithName() method has the metadata format argument removed.

## QA steps

Deploy any IAAS charm... eg
juju deploy mysql


hmlanigan (approved):
I deployed cs, ch and local charms, and a cs bundle for QA.

wallyworld:
$$merge$$

PR #12122, opened 2020-10-13
Fix chicken and egg with model admission.

tlm:
This change introduces a fix for lp-1898718 where the model operator
pods can't restart in kubernetes after because they rely on validation
of their self

## Checklist

## QA steps

Acceptance test coming soon.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1898718

### DIFF
--- a/caas/kubernetes/provider/bootstrap.go
+++ b/caas/kubernetes/provider/bootstrap.go
@@ -566,8 +566,11 @@ func (c *controllerStack) createControllerStatefulset() error {
 	numberOfPods := int32(1) // TODO(caas): HA mode!
 	spec := &apps.StatefulSet{
 		ObjectMeta: v1.ObjectMeta{
-			Name:        c.resourceNameStatefulSet,
-			Labels:      c.stackLabels,
+			Name: c.resourceNameStatefulSet,
+			Labels: AppendLabels(
+				nil,
+				c.stackLabels,
+				LabelsModelOperatorDisableWebhook),
 			Namespace:   c.broker.GetCurrentNamespace(),
 			Annotations: c.stackAnnotations,
 		},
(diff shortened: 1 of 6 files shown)


jameinel:
!!build!!

caas/kubernetes/provider/bootstrap.go
```diff
@@ -566,8 +566,11 @@ func (c *controllerStack) createControllerStatefulset() error {
 	numberOfPods := int32(1) // TODO(caas): HA mode!
 	spec := &apps.StatefulSet{
 		ObjectMeta: v1.ObjectMeta{
-			Name:        c.resourceNameStatefulSet,
-			Labels:      c.stackLabels,
+			Name: c.resourceNameStatefulSet,
```
ycliuhw, line 569: after doing this, the model operator in the controller namespace will be doing nothing.
Probably, we just do not deploy the model operator for the controller model?

tlm: Hmmm seems like a lot of work for what could be used later on. There is nothing stopping people from deploying to controller model correct ?

Thoughts @wallyworld

wallyworld: People can still deploy charms to the controller model. They do it all the time in IAAS; for k8s it's also possible but the need is much less. Still having the model operator there for when needed is a good thing. Its purpose it to track non-juju created resources; we know about the controller stateful set so no need to track it

tests/suites/caasadmission/admission.sh
```diff
@@ -139,3 +139,20 @@ juju_app=$(kubectl --kubeconfig "${TEST_DIR}"/kube-sa.json get cm -n "${namespac
 
   echo "$juju_app" | check test-app
 }
+
+# Tests that after the model operator pod restarts it can come back up without
+# having to be validated by it's self.
```
wallyworld, line 144: itself

wallyworld (approved):

ycliuhw (approved):

wallyworld:
[reboot breaks microk8s deployments](https://trello.com/c/hAHbUTzA/1207-reboot-breaks-microk8s-deployments)

tlm:
$$merge$$

tlm:
$$merge$$

PR #12130, opened 2020-10-14
Fix tests runner bash

SimonRichardson:
There is a tension between running tests and running linters. The fix
previously worked great for linters, but failed hard for tests. This
fixes both paths.

We now have a run_linter command that turns out the ability to not have
to fail hard when you encounter an error, but run will continue to do
so.

Allow pipefail to run for linters, but not for other tests. We need to
correctly handle pipefail to ensure we have the correct modes for the
tests.

This happens because we have a tension between the expected output from
the different running items.

The real fix is to use the fact that the subshell can fail, but we don't
want the subshell to bring down the parent shell. Correctly doing this
is kinda tricky, but I believe this is working as expected now.

## QA Steps

```sh
make static-analysis
```

```sh
cd tests && ./main.sh controller
```


tests/suites/static_analysis/lint_go.sh
```diff
@@ -45,19 +45,19 @@ run_deadcode() {
...
+    echo -e "\\n${OUT}" >&2
     exit 1
   fi
 }
 
 run_misspell() {
   FILES="${2}"
   # shellcheck disable=SC2086
-  OUT=$(misspell -source=go 2>/dev/null ${FILES} || true)
+  OUT=$(misspell -i=gratuitious ${FILES} 2>/dev/null || true)
```
achilleasa, line 56: For anyone curious about why this exception is needed, see https://github.com/juju/juju/blob/2.8/network/netplan/netplan.go#L191-L193

achilleasa (approved):

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

> FAIL: status_internal_test.go:4616: StatusSuite.TestStatusAllFormats

PR #12136, opened 2020-10-14
Make the juju-fake-init/pebble binary like juju.

hpidcock:
## Description

This is to include the juju-fake-init/pebble binary in the juju k8s operator image. Eventually this will be replaced with the real binary, so this commit is only temporary for 2.9-beta1+

## QA steps

Build juju

## Documentation changes

N/A

## Bug reference

N/A


wallyworld (approved):

hpidcock:
$$merge$$

PR #12138, opened 2020-10-15
Fix bash tests part deux

SimonRichardson:
Turns out the tension between run and run_linter was too great. This
instead makes it more evident that the two should never be merged
together as they require different things from the set modes.

Some time should be used to potentially merge the run functions together
with maybe a MODE envar, but for now, we'll have duplicate functions
so we can make any tweaks without breaking the abstractions.

## QA steps

This takes some time, but is worth testing...

```sh
make go-install
cd tests && ./main.sh -l test deploy
```

```sh
make static-analysis
```

### DIFF
--- a/tests/includes/run.sh
+++ b/tests/includes/run.sh
@@ -22,7 +22,7 @@ run() {
   # Do not remove or none of the tests will report correctly!
   set +e
 
-  cmd_output=$("${CMD}" "$@" 2>&1)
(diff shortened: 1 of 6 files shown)


SimonRichardson:
@benhoyt might be worth checking this as well!

achilleasa (approved):
:shipit:

SimonRichardson:
$$merge$$

PR #12148, opened 2020-10-16
agent-stream isn't honoured with upgrade-juju

SimonRichardson:
This acceptance test expected that you can go from released stream to
devel stream using upgrade-juju --agent-stream devel. That isn't the
case and so the test is a bad test. There are numerous bugs around this
so we should really fix this at some point:

 - https://bugs.launchpad.net/juju/+bug/1742342
 - https://bugs.launchpad.net/juju/+bug/1863179
 - https://bugs.launchpad.net/juju/+bug/1875567
 - https://bugs.launchpad.net/juju/+bug/1879573


manadart (approved):

SimonRichardson:
$$merge$$

PR #12149, opened 2020-10-16
Remove double validation

SimonRichardson:
Seems like a simple copy and paste error when adding documentation.

Simply removing the offending code to clean that up, prevents the double
calling of the validation method.


manadart (approved):

SimonRichardson:
$$merge$$

PR #12154, opened 2020-10-18
Use the pool system state for getting api addresses

wallyworld:
When getting controller API addresses, the controller model state is needed for k8s models as that's where the controller cloud service entity is stored. We were constructing a new controller model state for each API call, but that's not needed as the state pool is already available on the facade context, and said state pool has a reference to SystemState().

This PR weaves use of the controller model through the facade layers, for both CAAS and IAAS models.

## QA steps

bootstrap both k8s and lxd controllers
deploy charms
enable ha on lxd controller
check local controller yaml for correct api addresses
juju show-controller and check api addresses
ssh into charm operator and check agent conf for correct api addresses


wallyworld:
[excess newstate() calls in k8s controller](https://trello.com/c/gFLNhmUh/1216-excess-newstate-calls-in-k8s-controller)

hpidcock (approved):
LGTM Thankyou

wallyworld:
$$merge$$

apiserver/facades/agent/caasapplication/application.go
```diff
@@ -40,13 +41,17 @@ type Facade struct {
 func NewStateFacade(ctx facade.Context) (*Facade, error) {
 	authorizer := ctx.Auth()
 	resources := ctx.Resources()
-	return NewFacade(resources, authorizer, &stateShim{ctx.State()}, ctx.StatePool().Clock())
+	return NewFacade(resources, authorizer,
+		&stateShim{ctx.StatePool().SystemState()},
```
manadart, line 45: This doesn't appear to required the shim wrapper. If not, including it hinders clarity somewhat...

wallyworld:
[extra logging for unit agent operations (include unit name) for k8s](https://trello.com/c/xONuuHqe/1224-extra-logging-for-unit-agent-operations-include-unit-name-for-k8s)

manadart:
This appears to have broken machine provisioning on Oracle. I only just found it via `git bisect`, so have not delved into how.

Machine agents are getting an unauthorised error when connecting, then shutting down permanently.

PR #12159, opened 2020-10-19
Fix data race in apiserver.(*embeddedCLIHandler).runEmbeddedCommands()

hmlanigan:
Create a juju constructor for a linereader Reader to avoid changing the structure values, timeout, after it's been used in a go routine.

## QA steps

No change to current behavior should be seen.  Ran ~/bin/stress-race.bash in apiserver: 6 times and counting with no issue.  Seen in first time without the change.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1900619


wallyworld (approved):

hmlanigan:
$$merge$$

PR #12164, opened 2020-10-20
[dev] Remove transitive dependency to apiserver in state package

achilleasa:
This is a maintenance PR that removes a transitive dependency between the `state` package and the `apiserver` package. As a result of the transitive dep, running the tests within the `state` package would trigger a recompile of a large part of the code-base. 

The issue was caused by the import of the `api/testing` package which some of the tests used for creating/comparing macaroons. The relevant code has been copied into state and the transitive dependency has been eliminated.

In addition, redundant imports of `provider/dummy` in the test suite code have been removed and replaced with a direct import to `storage/provider/dummy`.

## QA steps

Add this block in one of the files in the apiserver package (e.g. rest.go)

```go
func init(){
  panic("state imports apiserver")
}
```

Run the tests in the state package and verify that the test suites begin to execute without hitting the injected panic

```console
cd state
go test -check.v
```

### DIFF
--- a/state/charm_test.go
+++ b/state/charm_test.go
@@ -17,7 +17,6 @@ import (
 	"gopkg.in/macaroon.v2"
 	"gopkg.in/mgo.v2"
 
-	apitesting "github.com/juju/juju/api/testing"
 	"github.com/juju/juju/state"
 	"github.com/juju/juju/state/storage"
 	"github.com/juju/juju/testcharms"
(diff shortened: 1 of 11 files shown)


SimonRichardson (commented):
Code looks good, will run Q&A steps next

SimonRichardson (approved):
Finally running `go test -v ./state` doesn't use 100% of my CPU on all 24 cores!

achilleasa:
$$merge$$

achilleasa:
$$merge$$

PR #12165, opened 2020-10-20
Start to implement bootstrap simplestream test

SimonRichardson:
Using the `juju metadata generate-agents` we can now test that we
can deploy from a local jujud to a bootstrapped controller using
simplestreams.

The test is rather simple but annoyingly hard to articulate in a concise
manor.

---

Additionally, I improved the output from generating the agents command
by fixing the help text about what is expected and where and also
preventing the fallback to the public clouds if your agent tgz doesn't
match the correct format.

This ensures that you can see that the agent stream wasn't correctly
generated at generational time RATHER than at bootstrap time, which
is REALLY confusing!

## QA steps

It expects that jujud and juju-matadata are on your $PATH and that the 
python is around to create a static HTTP server.

```sh
cd tests && ./main.sh bootstrap
```


environs/bootstrap/bootstrap.go
```diff
@@ -452,10 +452,15 @@ func bootstrapIAAS(
...
-			// we always want the latest/newest available.
-			agentVersion, availableTools = availableTools.Newest()
+		if len(availableTools) != 0 {
+			if args.AgentVersion == nil {
+				// If agent version was not specified in the arguments,
+				// we always want the latest/newest available.
+				agentVersion, availableTools = availableTools.Newest()
+			}
+			for _, tool := range availableTools {
+				ctx.Infof("Located Juju agent version %s for %s at %s", tool.Version, bootstrapArch, tool.URL)
```
SimonRichardson, line 462: This is really useful if we want to know which available tools will be selected for the bootstrap and from what URL.

tests/includes/cleanup.sh
```diff
@@ -3,7 +3,7 @@ add_clean_func() {
 
     name=${1}
 
-    echo "${name}" > "${TEST_DIR}/cleanup"
+    echo "${name}" >> "${TEST_DIR}/cleanup"
```
SimonRichardson, line 6: This is REALLY required!

hmlanigan:
Alternative to creating the server: You can use juju metadata generate-agents and have local files.  Bootstrap with --metadata-source <path>

hmlanigan (approved):
QA went well.

tests/suites/bootstrap/streams.sh
```diff
@@ -30,8 +30,8 @@ run_simplestream_metadata() {
     wait_for "ubuntu" "$(idle_condition "ubuntu")"
 }
 
-test_stream() {
-    if [ -n "$(skip 'test_stream')" ]; then
+test_bootstrap_simplestream() {
```
hmlanigan, line 33: 👍

SimonRichardson:
$$merge$$

PR #12167, opened 2020-10-21
Fix provider/common tests if repo not checked out at */juju/juju

benhoyt:
Simple fix for test failures I was seeing locally with `provider/common` package. I home the juju repo checked out at `~/w/juju`, but it seems the tests expect it at `*/juju/juju` (probably most people have it at `~/go/src/github.com/juju/juju`?). Was probably happening only to me (and I guess I've never run the entire unit tests suite locally! I usually run by subdirectory or package).

Either way, the tests shouldn't depend on the checkout directory, so loosen up this regex a bit.


wallyworld (approved):

benhoyt:
$$merge$$

PR #12168, opened 2020-10-21
Fix a race in the worker/introspection tests TestUnitStatusTimeout

benhoyt:
What was happening is that sometimes (specifically in `-race` mode), the `Advance()` call in the test was happening before the select on `time.After()` in the worker code, and the `time.After()` never triggered. Fixed by using `WaitAdvance()` -- see https://github.com/juju/juju/wiki/Intermittent-failures

Additionally, there was no timeout on the testing `http.Client`, so the `Get()` never returned and the test didn't stop till the 40 minute overall timeout kicked in. Once I added an http.Client timeout the actual issue above was a lot more clear.

### DIFF
--- a/worker/introspection/worker_test.go
+++ b/worker/introspection/worker_test.go
@@ -355,7 +355,7 @@ two: stopped`[1:])
 
 func (s *introspectionSuite) TestUnitStatusTimeout(c *gc.C) {
 	unsub := s.hub.Subscribe(agent.UnitStatusTopic, func(string, interface{}) {
-		s.clock.Advance(10 * time.Second)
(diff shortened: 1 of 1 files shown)


wallyworld (approved):

benhoyt:
!!build!!

jameinel:
Just as a heads up, WaitAdvance is clearly better than just Advance. However, it isn't perfect in all cases, either. (I've known a few places where it wasn't just sufficient to use WaitAdvance. If you ever have the potential for >1 Waiter, and/or the waiter has an alternate path with a timeout it can get confused by the number of waiters.)
It is still better than a bare Advance, though.

benhoyt:
$$merge$$

PR #12173, opened 2020-10-21
Upgrade controller via streams

SimonRichardson:
Waiting for https://github.com/juju/juju/pull/12165 to land

-----

The following test ensures that we can upgrade our controller using
simplestreams from a known stable branch to the new juju.

There is a LOT of horse play here to get what we want to do. Namely that
we need to find a good stable version, luckily we can use
`--channel=<last version>/stable when installing a snap and that'll
give us a clean stable.

Then we need to perform more magic, as between major releases (2.8 and
2.9) we bumped the series version. So before we generate our metadata
agents we also need to create tar.gz files for both focal and bionic.
That way we can upgrade correctly, as 2.8 will expect that we can
upgrade to 2.9 bionic, even if they're the same binary!

Another magic trick is that we bootstrap with the new server url even
though we're not using that, we fallback to the canonical
simplestreams for selecting the agent binary - magic /s

In addition to this, we then do the normal things of upgrading the
controller and waiting for it to correctly upgrade.

Additional work for deploying should be added to ensure that we didn't
break things elsewhere.

# QA Steps

**This will update your local snap client if testing locally** 

It is expected to bring up a password prompt if testing locally, this is because
the snap wants to install/refresh juju.

This won't be a problem in jenkins.

```sh
./main.sh upgrade
```


cmd/plugins/juju-metadata/toolsmetadata.go
```diff
@@ -29,32 +29,41 @@ func newToolsMetadataCommand() cmd.Command {
...
+directory path must contain a "tools" subdirectory. The "tools" directory is 
+expected to contain a streams release subdirectory (released, devel, proposed) 
+containing agent binary tarballs. The output of the streams metadata will be
+written next to the release subdirectory.
+
+The stream for which metadata is generated is specified using the --stream 
+parameter (default is "released"). Metadata can be generated for any supported 
+stream - released, proposed, testing, devel.
+
+Tools tarballs can are located in either a sub directory called "releases"
```
hmlanigan, line 56: While you're here, can we changes Tools to Agents were possible?

SimonRichardson: Fixed in https://github.com/juju/juju/pull/12165

hmlanigan:
Tried qa - it didn't bring up a password prompt 

==> RUN OUTPUT: test_upgrade
    | ===> Using jujud version 2.9-rc1
    | error: access denied (try with sudo)
    | error: access denied (try with sudo)

hmlanigan (approved):
Now that we've got the python vs python3 issue resolved.  LGTM

SimonRichardson:
$$merge$$

PR #12176, opened 2020-10-21
Fix k8s charm v2 cleanup and upgrade.

hpidcock:
- Fixes charm upgrade on charm v2 k8s applications.
- Fixes application removal and caasapplicationprovisioner application worker teardown.

## QA steps

Deploy a charm, upgrade the charm (no changes), upgrade the charm (use a different oci-image resource), remove-application.

## Documentation changes

N/A

## Bug reference

N/A

### DIFF
--- a/api/caasapplicationprovisioner/client.go
+++ b/api/caasapplicationprovisioner/client.go
@@ -10,6 +10,7 @@ import (
 	"github.com/juju/version"
 
 	"github.com/juju/juju/api/base"
+	"github.com/juju/juju/api/common"
 	charmscommon "github.com/juju/juju/api/common/charms"
 	apiwatcher "github.com/juju/juju/api/watcher"
 	"github.com/juju/juju/apiserver/params"
@@ -106,16 +107,18 @@ func (c *Client) Life(appName string) (life.Value, error) {
 
 // ProvisioningInfo holds the info needed to provision an operator.
 type ProvisioningInfo struct {
-	ImagePath    string
-	Version      version.Number
-	APIAddresses []string
-	CACert       string
-	Tags         map[string]string
-	Constraints  constraints.Value
-	Filesystems  []storage.KubernetesFilesystemParams
-	Devices      []devices.KubernetesDeviceParams
-	Series       string
-	ImageRepo    string
+	ImagePath            string
+	Version              version.Number
+	APIAddresses         []string
+	CACert               string
+	Tags                 map[string]string
+	Constraints          constraints.Value
+	Filesystems          []storage.KubernetesFilesystemParams
+	Devices              []devices.KubernetesDeviceParams
+	Series               string
+	ImageRepo            string
+	CharmModifiedVersion int
+	CharmURL             *charm.URL
 }
 
 // ProvisioningInfo returns the info needed to provision an operator for an application.
@@ -136,14 +139,15 @@ func (c *Client) ProvisioningInfo(applicationName string) (ProvisioningInfo, err
 	}
 
 	info := ProvisioningInfo{
-		ImagePath:    r.ImagePath,
-		Version:      r.Version,
-		APIAddresses: r.APIAddresses,
-		CACert:       r.CACert,
-		Tags:         r.Tags,
-		Constraints:  r.Constraints,
-		Series:       r.Series,
-		ImageRepo:    r.ImageRepo,
+		ImagePath:            r.ImagePath,
+		Version:              r.Version,
+		APIAddresses:         r.APIAddresses,
+		CACert:               r.CACert,
+		Tags:                 r.Tags,
+		Constraints:          r.Constraints,
+		Series:               r.Series,
+		ImageRepo:            r.ImageRepo,
+		CharmModifiedVersion: r.CharmModifiedVersion,
 	}
 
 	for _, fs := range r.Filesystems {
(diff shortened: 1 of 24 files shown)


api/caasapplicationprovisioner/client.go
```diff
@@ -334,3 +346,9 @@ func (c *Client) UpdateUnits(arg params.UpdateApplicationUnits) (*params.UpdateA
 	}
 	return firstResult.Info, maybeNotFound(firstResult.Error)
 }
+
+// WatchApplication returns a NotifyWatcher that notifies of
+// changes to the application in the current model.
+func (c *Client) WatchApplication(appName string) (watcher.NotifyWatcher, error) {
```
wallyworld, line 352: Test?

worker/caasapplicationprovisioner/application.go
```diff
@@ -103,82 +108,134 @@ func (a *appWorker) loop() error {
 		return errors.Errorf("charm version 2 or greater required")
 	}
 
+	// Update the password once per worker start to avoid it changing to frequently.
```
wallyworld, line 111: too

worker/caasapplicationprovisioner/application.go
```diff
@@ -103,82 +108,134 @@ func (a *appWorker) loop() error {
...
+	a.password, err = utils.RandomPassword()
+	if err != nil {
+		return errors.Trace(err)
+	}
+	err = a.facade.SetPassword(a.name, a.password)
```
wallyworld, line 116: This should not be done every worker restart - it should be done once per deploy. ie when the application watcher first notices the application

worker/caasapplicationprovisioner/application.go
```diff
@@ -103,82 +108,134 @@ func (a *appWorker) loop() error {
...
+		switch appLife {
+		case life.Alive:
+			if appStateChanges == nil {
+				appStateWatcher, err := a.facade.WatchApplication(a.name)
+				if err != nil {
+					return errors.Annotatef(err, "failed to watch for changes to application %q", a.name)
+				}
+				err = a.catacomb.Add(appStateWatcher)
+				if err != nil {
+					appStateWatcher.Kill()
```
wallyworld, line 148: This isn't needed. The only error is ErrDying and the catacomb stops the watcher in that case

worker/caasapplicationprovisioner/application.go
```diff
@@ -103,82 +108,134 @@ func (a *appWorker) loop() error {
...
 				}
-			case life.Dead:
-				return a.dead(app)
+				replicaChanges = replicaWatcher.Changes()
+			}
+		case life.Dying:
+			err = a.dying(app)
+			if err != nil {
+				return errors.Trace(err)
+			}
```
wallyworld, line 185: Inline the err for briefity? etc

worker/caasapplicationprovisioner/application.go
```diff
@@ -103,82 +108,134 @@ func (a *appWorker) loop() error {
...
+				}
+				err = a.catacomb.Add(appStateWatcher)
+				if err != nil {
+					appStateWatcher.Kill()
+					return errors.Trace(err)
+				}
+				appStateChanges = appStateWatcher.Changes()
+			}
+			err = a.alive(app)
+			if err != nil {
```
wallyworld, line 154: Inline the err for briefity?

worker/caasapplicationprovisioner/application.go
```diff
@@ -103,82 +108,134 @@ func (a *appWorker) loop() error {
...
-					}
-					err = a.catacomb.Add(appWatcher)
-					if err != nil {
-						appWatcher.Kill()
-						return errors.Trace(err)
-					}
-					appChanges = appWatcher.Changes()
+				err = a.catacomb.Add(appWatcher)
+				if err != nil {
+					appWatcher.Kill()
```
wallyworld, line 164: ditto

worker/caasapplicationprovisioner/application.go
```diff
@@ -103,82 +108,134 @@ func (a *appWorker) loop() error {
...
+			done = true
+			return nil
+		}
+		return nil
+	}
+
+	for {
+		select {
+		case <-a.catacomb.Dead():
+			return a.catacomb.Err()
```
wallyworld, line 200: This should be

```
case <-a.catacomb.Dying():
	return a.catacomb.ErrDying()
```

worker/uniter/uniter.go
```diff
@@ -306,6 +309,15 @@ func (u *Uniter) loop(unitTag names.UnitTag) (err error) {
 		return err
 	}
 
+	// Check we are running the correct charm version.
+	if u.embedded && u.enforcedCharmModifiedVersion != -1 {
+		if charmModifiedVersion != u.enforcedCharmModifiedVersion {
+			u.logger.Infof("remote charm modified version (%d) does not match agent's (%d)",
```
wallyworld, line 315: I'm unclear on when this could happen.

wallyworld (approved):
A few small things, especially the catacomb fixes.
Could you explain how the unit agent charm modified version can get out of sync? Is the check just for safety?

worker/uniter/uniter.go
hpidcock: Pod starts up, user upgrades, unit agent starts with old CharmModifiedVersion but the Application in state is different. Uniter downloads the new charm when it should be running the old charm.

worker/caasapplicationprovisioner/application.go
hpidcock: I'm generally not a fan of inline statements but I'll do it for you :)

worker/caasapplicationprovisioner/application.go
hpidcock: I don't think this is a problem. The worker needs to know the password and I prefer it have ownership over the password lifecycle, since it is the thing the ensures the provider updates the Secret.

This password is only the application password used by the init container, so it isn't an issue if it changes briefly to be replaced in the Secret.

hpidcock:
$$merge$$

PR #12177, opened 2020-10-22
Try to fix intermittent TestAPIServerCanShutdownWithOutstandingNext

benhoyt:
This test is failing intermittently in -race mode, for example: https://jenkins.juju.canonical.com/job/Unit-RunUnitTests-race-amd64/6/consoleFull

Try increasing the timeout for stopping the server from 10s to 60s, and also add logging so we can see how long it usually takes (and if the MgoServer.Reset() is being retried).

Tracking issue: https://bugs.launchpad.net/juju/+bug/1900931

### DIFF
--- a/api/controller/legacy_test.go
+++ b/api/controller/legacy_test.go
@@ -240,13 +240,15 @@ func (s *legacySuite) TestAPIServerCanShutdownWithOutstandingNext(c *gc.C) {
 	go func() {
 		// Resetting the dummy environment will call Stop on the
 		// embedded API server.
+		start := time.Now()
 		dummy.Reset(c)
+		c.Logf("dummy.Reset() took %v", time.Since(start))
 		close(srvStopped)
 	}()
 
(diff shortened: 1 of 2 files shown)


api/controller/legacy_test.go
```diff
@@ -240,13 +240,15 @@ func (s *legacySuite) TestAPIServerCanShutdownWithOutstandingNext(c *gc.C) {
...
+		start := time.Now()
 		dummy.Reset(c)
+		c.Logf("dummy.Reset() took %v", time.Since(start))
 		close(srvStopped)
 	}()
 
 	select {
 	case <-srvStopped:
-	case <-time.After(testing.LongWait):
+	case <-time.After(time.Minute):
```
wallyworld, line 251: Let's add a comment and TODO to explain why the timeout is a minute rather than LongWait

wallyworld (approved):

PR #12182, opened 2020-10-22
[dev] Wait for watchers to be idle before evaluating TestStatusAllFormats expectations

achilleasa:
The TestStatusAllFormats test is occasionally failing on CI. This commit, under the assumption that the test fails due to a race in processing watcher events and updating the cached status, injects a `WaitForModelWatchersIdle` call before evaluating the expectation for the table-driven test cases used by the test suite.

The `expect` step (one of the concrete implementations of the `step` interface ) has been augmented with a `requireModelWatchersIdle` method that returns a boolean (always returns `true`) to indicate whether it requires a call to WaitModelWatchersIdle prior to executing the step. The suite runner checks whether each step implements this method and if it does, and it returns true, invokes `WaitForModelWatchersIdle`

### DIFF
--- a/cmd/juju/status/status_internal_test.go
+++ b/cmd/juju/status/status_internal_test.go
@@ -103,6 +103,10 @@ type agentStatusSetter interface {
 	SetAgentStatus(agent string, status corepresence.Status)
 }
 
+type idleModelWatcherRequirer interface {
+	requireModelWatchersIdle() bool
+}
+
 type context struct {
 	st            *state.State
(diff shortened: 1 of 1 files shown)


manadart (approved):
Ostensibly, this appears to work.

Without this patch, I was able to get a failure reasonably early in a loop, but I haven't seen a failure with it.

achilleasa:
$$merge$$ (I will be monitoring CI runs to see if the issue re-appears in the future)

PR #12184, opened 2020-10-22
Provisioner state address

wallyworld:
The provisioner worker was calling a StateAddresses() API but it doesn't need that info. Remove the API and fix the associated fallout.

The second commit is a drive by to move some trivial container helper methods from state to core. This removes one of the cases where the worker package imports state. Sadly it's one of many, so there's more to do.

## QA steps

bootstrap k8s and deploy a charm
juju show-controller and check the api address is correct

bootstrap lxd and deploy a charm
juju show-controller and check the api address is correct
juju enable-ha
juju show-controller and check the api addresses are correct
remove a controller machine
enable-ha again
check that a replacement controller comes up and show-controller is happy

## Bug reference

https://bugs.launchpad.net/juju/+bug/1866643

### DIFF
--- a/apiserver/facades/agent/provisioner/provisioner_test.go
+++ b/apiserver/facades/agent/provisioner/provisioner_test.go
@@ -1860,17 +1860,6 @@ func (s *withControllerSuite) TestAPIAddresses(c *gc.C) {
 	})
 }
 
-func (s *withControllerSuite) TestStateAddresses(c *gc.C) {
-	addresses, err := s.State.Addresses()
-	c.Assert(err, jc.ErrorIsNil)
-
-	result, err := s.provisioner.StateAddresses()
-	c.Assert(err, jc.ErrorIsNil)
-	c.Assert(result, gc.DeepEquals, params.StringsResult{
-		Result: addresses,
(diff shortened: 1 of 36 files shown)


ycliuhw (approved):
lgtm thanks

wallyworld:
[Provisioner shouldn't need to know StateAddresses](https://trello.com/c/HCZ0OmX8/1233-provisioner-shouldnt-need-to-know-stateaddresses)

wallyworld:
$$merge$$

apiserver/facades/agent/provisioner/provisioner_test.go
```diff
@@ -1860,17 +1860,6 @@ func (s *withControllerSuite) TestAPIAddresses(c *gc.C) {
 	})
 }
 
-func (s *withControllerSuite) TestStateAddresses(c *gc.C) {
```
manadart, line 1863: This breaks backward compatibility, even if unused.
I think the version needs to be bumped and the method stubbed for old versions as was done in:
https://github.com/juju/juju/pull/11706

PR #12189, opened 2020-10-23
Smoke tests: Do not tee the output

SimonRichardson:
Do not append when tee'ing to the output.log. Otherwise the following
grep will always match and so will never die!

## QA steps

Smoke tests pass.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -165,7 +165,7 @@ jobs:
 
         VERSION=$(juju version | cut -d "-" -f 1,2 | xargs -I% echo "%.1")
         while true; do
-            juju upgrade-model --agent-version="$VERSION" 2>&1 | tee -a output.log || true
+            juju upgrade-model --agent-version="$VERSION" 2>&1 | tee output.log || true
             RES=$(cat output.log | grep "upgrade in progress" || echo "NOT-UPGRADING")
             if [ "$RES" = "NOT-UPGRADING" ]; then
                 break


hmlanigan (approved):

SimonRichardson:
$$merge$$

.github/workflows/smoke.yml
```diff
@@ -165,7 +165,7 @@ jobs:
 
         VERSION=$(juju version | cut -d "-" -f 1,2 | xargs -I% echo "%.1")
         while true; do
-            juju upgrade-model --agent-version="$VERSION" 2>&1 | tee -a output.log || true
+            juju upgrade-model --agent-version="$VERSION" 2>&1 | tee output.log || true
```
petevg, line 168: I think that this loop needs to be refactored completely, rather than patched. This change will fix the immediate problem, but at the cost of losing our output.log, which might be useful when attempting to investigate a test failure in the future.

My understanding of the current loop goes something like this ...

1) We blast the upgrade command at the controller, presumably because we don't trust that it will work the first time. Do we need to adjust the wait above so that we don't need to do this?

2) We then rely on an error message about the upgrade in order to determine whether it has completed. Can we do something smarter instead? If we can't determine whether an upgrade has completed via juju status or debug output, I'm guessing that operators in the field are also going to run into trouble. I'm not sure that I want to tell them that the solution is just to spam the upgrade command until it stops complaining. :-/

petevg (changes requested):
I'd like to see a bigger refactor (see comment) rather than a patch of some iffy logic here.

Edit: backing off, as juju-wait-for isn't in the snap yet, and this loop is fine for now.

.github/workflows/smoke.yml
SimonRichardson: The real fix is to use `juju wait-for`, until that happens, we should roll with this until I can land that.

petevg (approved):

PR #12190, opened 2020-10-23
Expose juju wait-for in a snap

SimonRichardson:
Now that we're going to use juju wait-for for tests we should expose it
via a snap.

## QA steps

Run this from the root of juju and ensure you don't have an existing one 
in the path, if so remove it...

```sh
$ snapcraft --use-lxd
$ sudo snap install *.snap --dangerous --classic
```

In one terminal do the following:

```sh
$ juju bootstrap lxd test
$ juju wait-for model test
```

In another terminal and ensure it's running from a snap...

```sh
$ ps aux | grep "juju wait-for"
... /snap/juju/x1/bin/juju wait-for model test
```

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -47,6 +47,7 @@ parts:
       - github.com/juju/juju/cmd/jujuc
       - github.com/juju/juju/cmd/jujud
       - github.com/juju/juju/cmd/plugins/juju-metadata
+      - github.com/juju/juju/cmd/plugins/juju-wait-for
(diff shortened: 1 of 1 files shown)


hmlanigan (approved):
`wait-for` was included in the snap.  However it didn't work as expected,  returning when a model named `test` was created.  This PR only address the first.

SimonRichardson:
> However it didn't work as expected, returning when a model named test was created

That's the default query, happy to have better defaults or you. An supply your own...

SimonRichardson:
$$merge$$

PR #12209, opened 2020-10-28
Fix lp:1901752, add detail to error message.

hmlanigan:
Add detail to juju add-cloud error message to help user understanding of the problem.

## QA steps

```console
# 
# Create a foo.yaml for adding a cloud, see help.  Instead of "clouds:" on the first line, put "louds:"
$ juju add-cloud foo foo.yaml
ERROR "no clouds found in parsed yaml, please validate yaml keys
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1901743

### DIFF
--- a/cmd/juju/cloud/add.go
+++ b/cmd/juju/cloud/add.go
@@ -721,7 +721,7 @@ func (p *CloudFileReader) ReadCloudFromFile(cloudFile string, ctxt *cmd.Context)
 		return nil, errors.Trace(err)
 	}
 	if len(specifiedClouds) == 0 {
-		return nil, errors.New("no personal clouds are defined")
+		return nil, errors.New("no clouds found in parsed yaml, please validate yaml keys")
 	}
 
 	var newCloud jujucloud.Cloud


cmd/juju/cloud/add.go
```diff
@@ -721,7 +721,7 @@ func (p *CloudFileReader) ReadCloudFromFile(cloudFile string, ctxt *cmd.Context)
 		return nil, errors.Trace(err)
 	}
 	if len(specifiedClouds) == 0 {
-		return nil, errors.New("no personal clouds are defined")
+		return nil, errors.New("no clouds found in parsed yaml, please validate yaml keys")
```
petevg, line 724: Is the error message that we get in the two sets of lines above also going to be vague? I think it might make sense to wrap all three errors in something like: ```"Error parsing yaml: %s" % err```  (That's pseudo code Python style formatting -- obv. you'd want to do the Golang equivalent :-) )

petevg (changes requested):
I think that we should generalize this for all three error messages when parsing yaml, so that we're covering all of our bases.

petevg (approved):

hmlanigan:
$$merge$$

PR #12218, opened 2020-10-29
Update os series

SimonRichardson:
The following commit updates the default series to include groovy, it
drops support for all other series that are no longer supported (eoan)
and additionally brings the supported code inline with the juju/os
package.

## QA steps

The tests pass, we're interested in ensuring that we correctly handle the
series correctly when bootstrapping with in the tests.

### DIFF
--- a/core/series/supported.go
+++ b/core/series/supported.go
@@ -4,6 +4,8 @@
 package series
 
 import (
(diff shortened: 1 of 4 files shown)


wallyworld:
$$merge$$

PR #12219, opened 2020-10-29
Patch host arch for override priority

SimonRichardson:
The following ensures that we patch the Host Arch so when it falls back
on a different machine arch that isn't AMD64, it does actually pick up
the AMD64 constant and not the real host arch.

## QA steps

Test on another architecture.


wallyworld:
Supeceded by #12220 which targets the 2.8 branch.

PR #12229, opened 2020-10-30
Fix race in k8sagent/initialize tests

benhoyt:
The cmdtesting version of Context.Stderr is not concurrency-safe
because it's a plain old bytes.Buffer, so ctx.Infof (at least in
testing) is not concurrency-safe, and this was causing a data race
in bytes.Buffer.Write.

Per discussion, just "fix" by avoiding the log here (we didn't have it
before anyway).

Drive-by is updating the tests to use different byte strings for the
three different files. Initially I thought this may have been causing
it, but it's not. Seems safer/better anyway.


wallyworld (approved):

wallyworld:
$$merge$$

PR #12232, opened 2020-10-30
Test prior versions

SimonRichardson:
So the current test we have says, test from the last stable version and
see if you can upgrade. Example being from 2.9.1 to 2.9.2 or 2.8.6 to
2.9.0. So it ensures that you didn't break anything from the last known
stable release to your PR version.

This change ensures that we test from the last prior release. So drop a
minor, not a patch, which is almost the same, except you would test
going from 2.8.6 to 2.9.2, the later example would be the same.

It's known that this is more of an established route to upgrade from as
people will sometimes skip a .0 release and wait until .1 or .2 release.

This way we can ensure that we can upgrade a controller in both
scenarios.

The code is actually really simple, we can reuse what we already have
and just reuse the function with a few tweaks.

## QA steps

See https://github.com/juju/juju/pull/12173

```sh
cd tests && ./main.sh upgrade
```

### DIFF
--- a/tests/suites/upgrade/streams.sh
+++ b/tests/suites/upgrade/streams.sh
@@ -1,20 +1,47 @@
-run_simplestream_metadata() {
-    VERSION=$(jujud version)
-    JUJUD_VERSION=$(jujud_version)
-    STABLE_VERSION=$(last_stable_version "${JUJUD_VERSION}")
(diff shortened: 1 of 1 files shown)


hmlanigan (approved):
QA good.

hmlanigan:
$$merge$$

PR #12236, opened 2020-11-02
Ensure Azure mirrors used for agent metadata

wallyworld:
When bootstrapping on Azure, ensure the local mirrors are used for agent metadata and binaries.

The Azure cpc mirror metadata now uses the proper cloud endpoint instead of the now obsolete storage endpoint, however, out of date cloud images Azure metadata is still published to [cloud-images.canonical.com](https://cloud-images.ubuntu.com/releases/streams/v1). This meant that in this PR, updating the Azure set up to use the correct endpoint to look up agents breaks image metadata lookup.

So the PR also includes a fix so that not having public cloud metadata is not fatal to bootstrap at that point - the error is still raised, but at a slightly later stage and also now with more context instead of a generic message. ie instead of 
`no image metadata found`
it says
`no metadata for "focal" images in ap-southeast-2 with arches [amd64]`

## QA steps

juju bootstrap azure --show-log

Look for a line like below, which shows agent binary comes from the blobstore mirror.
```
11:18:24 INFO  juju.cloudconfig userdatacfg_unix.go:573 Fetching agent: curl -sSfw 'agent binaries from %{url_effective} downloaded: HTTP %{http_code}; time %{time_total}s; size %{size_download} bytes; speed %{speed_download} bytes/s ' --retry 10 -o $bin/tools.tar.gz <[https://jujuagents.blob.core.windows.net/juju-agents/agents/agent/2.9-rc1/juju-2.9-rc1-ubuntu-amd64.tgz]>
```

### DIFF
--- a/acceptancetests/jujupy/tests/test_status.py
+++ b/acceptancetests/jujupy/tests/test_status.py
@@ -627,7 +627,7 @@ class TestStatus(FakeHomeTestCase):
         self.assertEqual(e.state, failure)
 
     def test_check_agents_started_read_juju_status_error(self):
-        failures = ['no "centos7" images in us-east-1 with arches [amd64]',
+        failures = ['no metadata for "centos7" images in us-east-1 with arches [amd64]',
                     'sending new instance request: GCE operation ' +
                     '"operation-143" failed', '']
         for failure in failures:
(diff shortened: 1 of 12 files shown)


wallyworld:
[fix azure agent metadata to use mirrors](https://trello.com/c/o8muP6Wa/1255-fix-azure-agent-metadata-to-use-mirrors)

hpidcock (approved):
LGTM

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #12239, opened 2020-11-02
Improve upgrade-model error message with permission error

wallyworld:
When running upgrade-model with --build-agent, a check is done to see if the model being upgraded is the controller model.
But if the user is upgrading a different model they do have access to, but they don't have access to the controller model, the error message "permission denied" is not helpful.

## QA steps

bootstrap and create a user and give that user access to a model (not the controller model)
As that user, run juju upgrade-model --build-agent
Check the error message
ERROR --build-agent can only be used with the controller model but you don't have permission to access that model

### DIFF
--- a/cmd/juju/commands/upgrademodel.go
+++ b/cmd/juju/commands/upgrademodel.go
@@ -444,14 +444,20 @@ func (c *upgradeJujuCommand) upgradeModel(ctx *cmd.Context, implicitUploadAllowe
 	}
 
 	controllerModelConfig, err := controllerClient.ModelConfig()
-	if err != nil {
+	if err != nil && !params.IsCodeUnauthorized(err) {
 		return err
 	}
-	isControllerModel := cfg.UUID() == controllerModelConfig[config.UUIDKey]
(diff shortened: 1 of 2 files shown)


wallyworld:
[upgrade-model error message](https://trello.com/c/LJsMk7e8/1256-upgrade-model-error-message)

cmd/juju/commands/upgrademodel.go
```diff
@@ -444,14 +444,20 @@ func (c *upgradeJujuCommand) upgradeModel(ctx *cmd.Context, implicitUploadAllowe
 	}
 
 	controllerModelConfig, err := controllerClient.ModelConfig()
-	if err != nil {
+	if err != nil && params.ErrCode(err) != params.CodeUnauthorized {
```
ycliuhw, line 447: `!params.IsCodeUnauthorized(err)` would be better?

ycliuhw (approved):
lgtm

wallyworld:
$$merge$$

PR #12243, opened 2020-11-02
Juju plugin check

SimonRichardson:
If using snaps then juju-metadata won't exist, instead it will have sort
of forwarding from juju to juju metadata and so the which check won't
work.

Interestingly I'm surprised this even works with plugins as it expects
things to be on a given path.

## QA steps

```sh
(cd tests && ./main.sh upgrade)
```

## Bug reference

Additional work, but doesn't fix https://bugs.launchpad.net/juju/+bug/1899786

### DIFF
--- a/tests/includes/check.sh
+++ b/tests/includes/check.sh
@@ -15,6 +15,23 @@ check_dependencies() {
     fi
 }
 
+check_juju_dependencies() {
+    local dep missing
+    missing=""
+
+    for dep in "$@"; do
+        if ! juju "$dep" >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


hmlanigan:
==> Cleaning up
====> Cleaning up jujus
====> Introspection gathering
ERRORjq: error (at <stdin>:1): null (null) has no keys
 dial tcp 10.74.147.104:17070: i/o timeout
====> Introspection gathered
====> Removing offers

hmlanigan (approved):

SimonRichardson:
$$merge$$

PR #12247, opened 2020-11-03
Fix data race in model cache

wallyworld:
As seen here

https://jenkins.juju.canonical.com/job/Unit-RunUnitTests-race-arm64/7/testReport/junit/github/com_juju_juju_core_cache/TestPackage/

There's concurrent access to the removal message when recording changes and doing a sweep operation.
Refactor access to the removal message and stale attribute to avoid the race.

## QA steps

go test --race

### DIFF
--- a/core/cache/model.go
+++ b/core/cache/model.go
@@ -560,14 +560,9 @@ func (m *Model) removeBranch(ch RemoveBranch) error {
 func (m *Model) setDetails(details ModelChange) {
 	m.mu.Lock()
 
-	// If this is the first receipt of details, set the removal message.
-	if m.removalMessage == nil {
-		m.removalMessage = RemoveModel{
-			ModelUUID: details.ModelUUID,
-		}
-	}
-
-	m.setStale(false)
+	m.setRemovalMessage(RemoveModel{
+		ModelUUID: details.ModelUUID,
+	})
 	m.details = details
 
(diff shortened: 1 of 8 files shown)


wallyworld:
[TestSweepWithConcurrentUpdates race failure](https://trello.com/c/HG0Ilrrr/1258-testsweepwithconcurrentupdates-race-failure)

petevg (commented):
My understanding is that we're centralizing some logic and wrapping it in a lock. That makes sense to me, and I am +1. Want one of the more experienced engineers to give the actual +1, though!

benhoyt (approved):

wallyworld:
$$merge$$

core/cache/model.go
```diff
@@ -560,14 +560,9 @@ func (m *Model) removeBranch(ch RemoveBranch) error {
...
 
-	// If this is the first receipt of details, set the removal message.
-	if m.removalMessage == nil {
-		m.removalMessage = RemoveModel{
-			ModelUUID: details.ModelUUID,
-		}
-	}
-
-	m.setStale(false)
+	m.setRemovalMessage(RemoveModel{
```
manadart, line 563: There's a lock acquisition on the line above that deadlocks this call.

wallyworld: There are 2 muexes. One on the Model and one on the embedded Resident struct.
The one used in the setRemoveMessage is the Resident one.

PR #12250, opened 2020-11-03
fix a test; find an actual bug

hmlanigan:
Fixed the reboot monitor tests for 2.9, by using charm upgrade to determine the start hook isn't run at agent restart, and a new wait for when removing a subordinate.  The wait for waits for the number of subordinate units to reach a certain count.

Once the above was resolved.  It turns out that we had stopped Purging the reboot monitor when a unit was removed from a machine.  Call lost when the machine and unit agents were merged together.

## QA steps

```sh
(cd tests ; ./main.sh -s test_dispatching_script -v hooks)
```

### DIFF
--- a/tests/suites/hooks/reboot.sh
+++ b/tests/suites/hooks/reboot.sh
@@ -10,7 +10,7 @@ run_start_hook_fires_after_reboot() {
     # log level is WARNING.
     juju model-config -m "${model_name}" logging-config="<root>=INFO;unit=DEBUG"
 
-    juju deploy cs:~jameinel/ubuntu-lite-7
+    juju deploy cs:~jameinel/ubuntu-lite-6
     wait_for "ubuntu-lite" "$(idle_condition "ubuntu-lite")"
 
     # Ensure that the implicit start hook after reboot detection does not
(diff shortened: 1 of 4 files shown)


SimonRichardson (approved):

SimonRichardson:
$$merge$$

tests/suites/hooks/reboot.sh
```diff
@@ -27,17 +27,19 @@ run_start_hook_fires_after_reboot() {
     # Restart the unit agent and ensure that the implicit start hook still
     # does not fire
     echo "[+] ensuring that implicit start hook does not fire after restarting the unit agent"
-    juju run --unit ubuntu-lite/0 'sudo service jujud-unit-ubuntu-lite-0 restart'
+    # juju run --unit ubuntu-lite/0 'sudo service jujud-unit-ubuntu-lite-0 restart'
```
achilleasa, line 30: I was looking into why this particular test was failing earlier today (e.g. https://jenkins.juju.canonical.com/job/test-hooks-lxd/449/console) and I noticed this error from the `juju run` on this line:

```Failed to restart jujud-unit-ubuntu-lite-0.service: Unit jujud-unit-ubuntu-lite-0.service not found.```

I believe we are hitting some kind of race here where the service file has not been written yet  even though status shows the unit as being idle. How about if we add a small helper to wait for the service to appear before attempting to restart? Something like:

```bash
# Call like: wait_for_unit_service_to_appear "ubuntu-lite/0" "ubuntu-lite-0"
wait_for_unit_service_to_appear() {
  unit=$1
  svc=$2
  for i in 1 2 3; do
    svc_present=$(juju ssh "$unit" "ls /etc/systemd/system/jujud-unit-${svc}.service || echo 'missing'")
    if [ "$svc_present" != "missing" ]; then
      juju ssh "$unit" "sudo systemctl restart jujud-unit-${svc}"
      return
    fi
 
    # Wait for service file to appear
    sleep 1
  done

   echo $(red "Systemd service file for ${svc} not present")
   exit 1
}
```

I think the upgrade-charm bit below is a nice addition for regression checks but I also believe we should also keep the above test to ensure that reboot is not incorrectly handled when the service is externally restarted.

worker/deployer/nested.go
```diff
@@ -37,6 +38,12 @@ type Logger interface {
 	Tracef(string, ...interface{})
 }
 
+// RebootMonitorStatePurger is implemented by types that can clean up the
```
achilleasa, line 41: Nice catch! I wonder why these changes got left out in the refactoring since they were probably on both 2.8 and develop by the time the agent unification bits landed. Could this be a copy of the original file that lived in the refactoring branch? That might explain why the changes got lost...

PR #12253, opened 2020-11-04
Expose EC2 integration fixes

SimonRichardson:
We can't always depend on the series of the machine, so we should find
it correctly before testing against it.

See: https://jenkins.juju.canonical.com/job/test-expose-ec2/1/console

## QA steps

```sh
(cd tests && ./main.sh -p aws expose_ec2)
```

### DIFF
--- a/tests/suites/expose_ec2/expose_app.sh
+++ b/tests/suites/expose_ec2/expose_app.sh
@@ -107,17 +107,8 @@ assert_ingress_cidrs_for_port_range() {
 assert_export_bundle_output_includes_exposed_endpoints() {
     echo "==> Checking that export-bundle output contains the exposed endpoint settings"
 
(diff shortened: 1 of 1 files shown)


tests/suites/expose_ec2/expose_app.sh
```diff
@@ -109,7 +109,7 @@ assert_export_bundle_output_includes_exposed_endpoints() {
 
     got=$(juju export-bundle)
     exp=$(cat <<-EOF
-series: focal
+series: %s
```
achilleasa, line 112: Why not grab the series before this line and expand it here instead of doing a hack with printf?

achilleasa (approved):

SimonRichardson:
$$merge$$

PR #12254, opened 2020-11-04
[2.8] Fix race in start hook after reboot test

achilleasa:
This PR backports the fixes from #12250 and adds extra logic to the hook tests to wait for the systemd files to be written to disk prior to attempting to restart the unit's agent.

## QA steps

```sh
cd tests
./main.sh -v hooks
```

### DIFF
--- a/tests/includes/wait-for.sh
+++ b/tests/includes/wait-for.sh
@@ -188,3 +188,39 @@ wait_for_model() {
         sleep "${SHORT_TIMEOUT}"
     fi
 }
+
+# wait_for_systemd_service_files_to_appear blocks until the systemd service
+# file for a unit is written to disk.
+#
+# ```
+# wait_for_systemd_service_files_to_appear <unit_name>
+#
+# example:
+# wait_for_systemd_service_files_to_appear "ubuntu-lite/0"
+# ```
+wait_for_systemd_service_files_to_appear() {
+    local unit
+
+    unit=${1}
+    # shellcheck disable=SC2086
+    svc_file_path="/etc/systemd/system/jujud-unit-$(echo -n ${1} | tr '/' '-').service"
+
+    attempt=0
+    # shellcheck disable=SC2046,SC2143
+    while [ "$attempt" != "3" ]; do
+        echo "[+] (attempt ${attempt}) waiting for the systemd unit files for ${unit} to appear"
+
+        svc_present=$(juju ssh "${unit}" "ls ${svc_file_path} 2>/dev/null || echo -n 'missing'")
+        if [ "${svc_present}" != "missing" ]; then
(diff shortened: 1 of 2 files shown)


tests/includes/wait-for.sh
```diff
@@ -158,6 +158,38 @@ wait_for_machine_netif_count() {
...
+# wait_for_subordinate_count blocks until the number of subordinates
+# to the desired unit becomes equal to the desired value.
+#
+# ```
+# wait_for_subordinate_count <applicaiton name> <principal unit num> <count>
+#
+# example:
+# wait_for_subordinate_count mysql 0 3
+# ```
+wait_for_subordinate_count() {
```
achilleasa, line 170: This has been back-ported from #12250

tests/suites/hooks/reboot.sh
```diff
@@ -26,6 +26,7 @@ run_start_hook_fires_after_reboot() {
 
     # Restart the unit agent and ensure that the implicit start hook still
     # does not fire
+    wait_for_systemd_service_files_to_appear "ubuntu-lite/0"
```
achilleasa, line 29: This was causing issues on (slow) CI as there was a small race between the unit appearing as idle and the systemd unit files being written to disk. This change addresses the issue and allows us to ensure that the start hook does not fire when the agent gets restarted _externally_.

tests/suites/hooks/reboot.sh
```diff
@@ -39,6 +40,22 @@ run_start_hook_fires_after_reboot() {
       exit 1
     fi
 
+    # Ensure that the implicit start hook does not fire after upgrading the unit
+    juju upgrade-charm ubuntu-lite --revision 7
```
achilleasa, line 44: This was backported from #12250

tests/suites/hooks/reboot.sh
```diff
@@ -87,7 +104,8 @@ run_reboot_monitor_state_cleanup() {
     echo "[+] Verifying that reboot monitor state files are removed once a subordinate gets removed"
     juju remove-relation rsyslog-forwarder mysql
     wait_for "mysql" "$(idle_condition "mysql")"
-    sleep 5 # there is probably a better way to wait for the subordinate to be removed.
+
+    wait_for_subordinate_count "mysql"
```
achilleasa, line 108: Also backported from #12250

petevg (commented):
I am a little bit sad that we don't have cleaner methods for waiting for things. But definitely approve of hiding the messy bits in functions that can be made prettier later/when we have extended our abilities to wait for cloud states.

hmlanigan (approved):
LGTM, qa went well

achilleasa:
Doh! This was supposed to be targeting 2.8....

PR #12255, opened 2020-11-04
deploy_centos integration test

hmlanigan:
nw-deploy-centos7-amd64-aws both bootstraps with a centos client and deploys a centos charm.  Breaking into 2 pieces to better provide centos images to the model for deployment.  A separate pr will address using the centos client to bootstrap.

The new test, test_deploy_os for centos is snowflake test.  It has a specific AMI in a specific region and will only take some of the instance types that juju might choose to use.  It will require tweaking as things change.

I tried to update to use the centos8 image.  However centos.org does not list one in the aws marketplace today. I was not able to use the aws centos8 image to create an juju machine.  More investigation is needed.  There are other centos8 images, but do not want to trust ones from unknown places.

Add 'ec2' to list of provider names to be aws for integration tests when using an already bootstrapped controller.

Update the default and bootstrap series for bootstrap to when deploying on win or centos in the acceptance tests to bionic instead of trusty.

## QA steps

```console
$ juju bootstrap aws --credentials <juju qa creds> deploy-centos-test
$ (cd tests ; ./main.sh -l centos-deploy -v deploy test_deploy_os)

# should not run
$ (cd tests ; ./main.sh -p lxd -v deploy test_deploy_os)

==> Checking for dependencies
==> TEST BEGIN: test_deploy (tmp.MMu)
==> TEST SKIPPED: deploy_centos - tests for AWS only
==> TEST DONE: test_deploy (0s)
==> Cleaning up
====> Completed cleaning up jujus

==> Test result: success
==> Tests Removed: /home/heather/work/src/github.com/juju/juju/tests/tmp.MMu
==> TEST COMPLETE
```

### DIFF
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -46,6 +46,9 @@ bootstrap() {
         "aws")
             provider="aws"
             ;;
+        "ec2")
+            provider="aws"
+            ;;
         "localhost")
             provider="lxd"
             ;;
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -43,14 +43,10 @@ ensure() {
 bootstrap() {
     local provider name output model bootstrapped_name
     case "${BOOTSTRAP_PROVIDER:-}" in
-        "aws")
-            ;&
-        "ec2")
+        "aws" | "ec2")
             provider="aws"
             ;;
-        "localhost")
-            ;&
-        "lxd")
+        "localhost" | "lxd")
             provider="lxd"
             ;;
         "lxd-remote")
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -44,14 +44,12 @@ bootstrap() {
     local provider name output model bootstrapped_name
     case "${BOOTSTRAP_PROVIDER:-}" in
         "aws")
-            provider="aws"
-            ;;
+            ;&
         "ec2")
             provider="aws"
             ;;
         "localhost")
-            provider="lxd"
-            ;;
+            ;&
         "lxd")
             provider="lxd"
             ;;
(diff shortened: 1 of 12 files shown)


hmlanigan:
!!build!!

tests/suites/deploy/os.sh
```diff
@@ -0,0 +1,47 @@
...
+        cd .. || exit
+
+
+
+        case "${BOOTSTRAP_PROVIDER:-}" in
+            "ec2")
+                run "run_deploy_centos"
+                ;;
+            *)
+                echo "==> TEST SKIPPED: deploy_centos - tests for LXD only"
```
achilleasa, line 20: Shouldn't this echo 'tests for EC2 only"?

tests/includes/juju.sh
```diff
@@ -46,6 +46,9 @@ bootstrap() {
         "aws")
             provider="aws"
             ;;
+        "ec2")
```
achilleasa, line 49: Why do we need this bit?

hmlanigan: When running the tests with a already bootstrapped aws controller, the type comes up as ec2.

tests/suites/deploy/os.sh
hmlanigan: yes, I have another commit to fix this and some other aws/ec2 issues across the board.  should be up in a minute.

hmlanigan:
I found a bug, working on the fix.  Test doesn't run if the test has to bootstrap a controller

achilleasa (approved):
Had to tweak the QA steps a bit and accept the ToS for centos but the test worked as expected.

achilleasa:
Linter is unhappy with the fallthroughs in the case code but you can probably just add the #/bin/bash to the top of the file since we are using bash now to make it happy.

hmlanigan:
$$merge$$

achilleasa:
@hmlanigan FYI, tests are now using bash exclusively as the main runner (see main.sh and https://github.com/juju/juju/pull/12130 / https://github.com/juju/juju/pull/12138) so we can use bash-isms freely ;-)

PR #10480, opened 2019-08-02
dev - support charm channel in bundles

achilleasa:
## Description of change

This PR introduces support for bundles which can specify a per-application channel for fetching its charm. 

Prior to the changes introduced by this PR, the value of the `--channel` parameter was injected into the charmstore client and used to resolve metadata/URLs for **all** charms. So for example, if we attempted to deploy a bundle from the `edge` channel (e.g. `juju deploy --channel edge cs:$some-bundle-in-edge`), each charm within the bundle would be looked up from the edge channel and potentially cause issues like (1677404; see bug reference below).

The various resolve URL implementations used by the deploy/updatecharm/bundle diff commands have been updated to use the appropriate channel when resolving bundle/charm URLS. The lookup rules are summarized below:

- For **bundle** deployments, use the value of `--channel` flag (if provided)
- For **individual charm** deployments (e.g. `juju deploy mysql`), use the value of `--channel` flag (if provided)
- For deploying **charms within a bundle**, use the value of the [channel](https://github.com/juju/charm/blob/v6/bundledata.go#L198) application-level bundle attribute if one is provided.

## QA steps

```console 
$ juju bootstrap lxd test --no-gui

## Deploy remote charm from a specific channel
$ juju deploy --channel edge cs:~juju-qa/bionic/dummy-in-edge
Located charm "cs:~juju-qa/bionic/dummy-in-edge-0".
Deploying charm "cs:~juju-qa/bionic/dummy-in-edge-0".

## Deploy local bundle with different channels per charm
$ cat > charm-channels.yaml << EOT
applications:
  dummy-in-edge:
    charm: cs:~juju-qa/bionic/dummy-in-edge
    channel: edge
  other-dummy-in-beta:
    charm: cs:~juju-qa/bionic/other-dummy-in-beta
    channel: beta
  wordpress:
    charm: wordpress
EOT

$ juju deploy ./charm-channels.yaml --debug --dry-run

12:24:33 INFO  juju.cmd supercommand.go:79 running juju [2.7-beta1 4039b10d183fb7f4cd147601e30e57e658dc1dfb gc go1.12.6]
12:24:33 DEBUG juju.cmd supercommand.go:80   args: []string{"juju", "deploy", "./charm-channels.yaml", "--debug", "--dry-run"}
12:24:33 INFO  juju.juju api.go:67 connecting to API addresses: [10.65.47.252:17070]
[...]

PR #12262, opened 2020-11-05
Prevent inheritance of bundle channels

SimonRichardson:
Bundles can have their own channels that are separate from the
application charm channels. This feature was implemented in PR#10480[1].
When moving the deployer code out from the application package into a
more testable setup we ended up causing a regression.

If a user wants to specify their own channels for each charm then they
should pin each charm correctly inside that bundle and not inherit that
channel from the bundle.

The fix is really simple, we just don't pass the origin through and
expect each charm to correctly resolve itself.

 1. https://github.com/juju/juju/pull/10480

## QA steps

Setup microk8s

```sh
sudo snap install microk8s --classic 
sudo usermod -a -G microk8s $USER
su - $USER

microk8s status --wait-ready
microk8s.enable dns storage

make microk8s-operator-update
juju bootstrap microk8s micro
```

The following would previously fail without this patch, running this will
make this work now.

```sh
juju deploy cs:kubeflow-lite
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1902945

### DIFF
--- a/cmd/juju/application/deploy.go
+++ b/cmd/juju/application/deploy.go
@@ -318,18 +318,24 @@ type DeployCommand struct {
 	unknownModel bool
 }
 
+// TODO (stickupkid): Update/re-write the following doc for charmhub related
+// charm urls.
 const deployDoc = `
-A charm can be referred to by its simple name and a series can optionally be
-specified:
+A charm or bundle can be referred to by its simple name and a series or channel
+can optionally be specified:
 
-  juju deploy postgresql
-  juju deploy bionic/postgresql
   juju deploy cs:postgresql
   juju deploy cs:bionic/postgresql
-  juju deploy postgresql --series bionic
+  juju deploy cs:postgresql --series bionic
+  juju deploy cs:postgresql --channel edge
 
 All the above deployments use remote charms found in the Charm Store (denoted
-by 'cs') and therefore also make use of "charm URLs".
+by 'cs' prefix) and therefore also make use of "charm URLs".
+
+Specifying a channel source will use that as the source to look for the charm or
+bundle from the Charm Store. Using a channel with a bundle will refer to
+the bundle only, to override a charm source in a bundle requires setting
+channel per each required charm.
 
 A versioned charm URL will be expanded as expected. For example, 'mysql-56'
 becomes 'cs:bionic/mysql-56'.
--- a/cmd/juju/application/deploy.go
+++ b/cmd/juju/application/deploy.go
@@ -332,10 +332,11 @@ can optionally be specified:
 All the above deployments use remote charms found in the Charm Store (denoted
 by 'cs' prefix) and therefore also make use of "charm URLs".
 
-Specifying a channel source will use that as the source to look for the charm or
-bundle from the Charm Store. Using a channel with a bundle will refer to
-the bundle only, to override a charm source in a bundle requires setting
-channel per each required charm.
+If a channel is specified, it will be used as the source for looking up the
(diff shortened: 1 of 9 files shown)


cmd/juju/application/deployer/bundle.go
```diff
@@ -215,7 +215,11 @@ type charmstoreBundle struct {
 
 // String returns a string description of the deployer.
 func (d *charmstoreBundle) String() string {
-	return fmt.Sprintf("deploy charm store bundle: %s", d.bundleURL.String())
+	str := fmt.Sprintf("deploy charm store bundle: %s", d.bundleURL.String())
+	if d.origin == (commoncharm.Origin{}) {
+		return str
+	}
+	return fmt.Sprintf("%s with origin: %s", str, d.origin.CoreChannel().String())
```
achilleasa, line 222: Can you double-check with the other places where we log the origin and ensure we use consistent terminology across the board? IIRC we tend to use "channel: X" in this package.

SimonRichardson: Sure can!

achilleasa: Have a look at bundlehandler.go:338 and refresh.go:370

SimonRichardson:
!!build!!

cmd/juju/application/deployer/charm.go
```diff
@@ -484,3 +493,15 @@ func (c *charmStoreCharm) PrepareAndDeploy(ctx *cmd.Context, deployAPI DeployerA
 	c.origin = csOrigin
 	return c.deploy(ctx, deployAPI)
 }
+
+func isEmptyOrigin(origin commoncharm.Origin, source commoncharm.OriginSource) bool {
```
achilleasa, line 497: Why not add an `IsEmpty` to the `Origin` type? Sounds like a useful helper to have

SimonRichardson: I can do, but I'm not sure how valid it actually would be outside of this.

achilleasa: Ok, guess if it's only used here we can leave it there.

achilleasa (approved):
LGTM. 

We may also need to tweak the docs (and maybe the `--channel` flag description in deploy/upgrade) to make it clear how the channel is applied.

cmd/juju/application/deploy.go
```diff
@@ -318,18 +318,24 @@ type DeployCommand struct {
...
   juju deploy cs:bionic/postgresql
-  juju deploy postgresql --series bionic
+  juju deploy cs:postgresql --series bionic
+  juju deploy cs:postgresql --channel edge
 
 All the above deployments use remote charms found in the Charm Store (denoted
-by 'cs') and therefore also make use of "charm URLs".
+by 'cs' prefix) and therefore also make use of "charm URLs".
+
+Specifying a channel source will use that as the source to look for the charm or
```
achilleasa, line 335: Maybe reworded as:

```
If a channel is specified, it will be used as the source for looking up the charm or bundle from the Charm Store. When used in a bundle deployment context, the specified channel is only used for retrieving the bundle and is ignored when looking up the charms referenced by the bundle. However, each charm within a bundle is allowed to explicitly specify the channel used to look it up.
```

Or something along those lines?

SimonRichardson: Yes!

SimonRichardson:
$$merge$$

achilleasa:
QA works as expected

PR #12264, opened 2020-11-05
ensure_text is in six 1.12.0+, ensure we have a better version.

hmlanigan:
Sometimes we use the wrong version of six, ensure we have the correct one.  Older versions, pre 1.12.0 do not include 'ensure_text' which leads to:

AttributeError: 'module' object has no attribute 'ensure_text'

in CI tests.


SimonRichardson (approved):

hmlanigan:
$$merge$$

PR #12266, opened 2020-11-05
Unfortunately yaml and json differ in serialization.

jameinel:
I'm pretty sure this was not intended, but we are using these structures
to format JSON and YAML content in 'network-get' output. And because the
structures didn't have YAML tags, they were getting bad serialization.
For now, I'm enshrining the bad serialization, because we don't want to
break compatibility arbitrarily. But I did want to highlight it instead
of having it be accidental, and then have us break it if we renamed a
field.

## QA steps

There should be no change in structured output. This is simply so that we record the actual tags we are using.

## Documentation changes

None.

## Bug reference

None.

### DIFF
--- a/apiserver/params/network.go
+++ b/apiserver/params/network.go
@@ -1153,24 +1153,28 @@ type ProxyConfigResultsV1 struct {
 }
 
 // InterfaceAddress represents a single address attached to the interface.
+// The serialization is different between json and yaml because of accidental
+// differences in the past, but should be preserved for compatibility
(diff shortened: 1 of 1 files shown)


hpidcock (approved):
This is most unfortunate but good pickup, it would be nice if we could configure json and yaml pkg to error when the tag is missing. I'd really prefer an explicit choice with serialized names vs auto-formatted names.

jameinel:
$$merge$$

PR #12270, opened 2020-11-06
Charmhub download integration tests

SimonRichardson:
The following adds an integration test for downloading a charm and then
deploying. The test fails because there is an error in the underlying
charmhub.

## QA steps

Until they fix the API we're using staging and so we need to bootstrap our
own controller and ensure that we use the staging url.

```sh
$ (cd tests && ./main.sh charmhub)
```

### DIFF
--- a/tests/suites/charmhub/download.sh
+++ b/tests/suites/charmhub/download.sh
@@ -6,7 +6,7 @@ run_charmhub_download() {
 
     ensure "test-${name}" "${file}"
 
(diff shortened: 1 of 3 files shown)


manadart (approved):
Works fine.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

> FAIL: remoterelations_test.go:523: remoteRelationsSuite.TestLocalRelationsRemoved

PR #12274, opened 2020-11-06
Bring in the updated github.com/juju/txn

jameinel:
It had a bug where if it got an error while removing transactions that
would bubble up into a panic because it wasn't properly handling the
error state.

This also brings in updated "lru" as it was a transitive dependency, but
I don't think that is a problem for our purposes.

## QA steps

Everything should build without changes. It is non-trivial to force an error at exactly the right time to trigger the removeAll to fail.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1903202


SimonRichardson (approved):

jameinel:
$$merge$$

PR #12276, opened 2020-11-06
Fix lp:1891227 - openstack networks with no AZ

hmlanigan:
Some openstack configurations do no specify availability zones for networks.  When looking for an external network to create a FIP, if the private network has no AZ, look for the same with the external network.

Did a little refactoring on AllocatePublicIP to shorten the method and make it easier to understand.

## QA steps

We do not have an exact environment within the team to test this.  The unit test added, does produce the same result if the provider change is not met.  

Test for not change to Openstack bootstrap where AZ for networks are available.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1891227

### DIFF
--- a/provider/openstack/networking.go
+++ b/provider/openstack/networking.go
@@ -122,46 +122,11 @@ func newNetworking(e *Environ) Networking {
 }
 
 // AllocatePublicIP is part of the Networking interface.
-func (n *NeutronNetworking) AllocatePublicIP(instId instance.Id) (*string, error) {
-	extNetworkIds := make([]string, 0)
-	neutronClient := n.env.neutron()
-	externalNetwork := n.env.ecfg().externalNetwork()
-	if externalNetwork != "" {
-		// the config specified an external network, try it first.
-		netId, err := resolveNeutronNetwork(neutronClient, externalNetwork, true)
-		if err != nil {
-			logger.Debugf("external network %s not found, search for one", externalNetwork)
-		} else {
-			logger.Debugf("using external network %q", externalNetwork)
-			extNetworkIds = []string{netId}
-		}
-	}
-
-	if len(extNetworkIds) == 0 {
-		// Create slice of network.Ids for external networks in the same AZ as
-		// the instance's network, to find an existing floating ip in, or allocate
-		// a new floating ip from.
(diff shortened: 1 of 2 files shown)


provider/openstack/networking.go
```diff
@@ -202,12 +167,68 @@ func (n *NeutronNetworking) AllocatePublicIP(instId instance.Id) (*string, error
...
+		}
+
+		availabilityZones := netDetails.AvailabilityZones
+		if len(availabilityZones) == 0 {
+			// No availability zones is valid, check for empty string
+			// to ensure we still find the external network with no
+			// AZ specified.  lp: 1891227
+			availabilityZones = []string{""}
+		}
+		for _, az := range availabilityZones {
```
hmlanigan, line 210: Here is the bug fix change.

provider/openstack/networking.go
```diff
@@ -219,6 +240,9 @@ func getExternalNeutronNetworksByAZ(e *Environ, azName string) ([]string, error)
 				break
 			}
 		}
+		if azName == "" && len(network.AvailabilityZones) == 0 {
+			netIds = append(netIds, network.Id)
+		}
```
hmlanigan, line 245: 2nd half of the bug fix change.

provider/openstack/networking.go
```diff
@@ -202,12 +167,68 @@ func (n *NeutronNetworking) AllocatePublicIP(instId instance.Id) (*string, error
...
+		for _, az := range availabilityZones {
+			extNetIds, _ := getExternalNeutronNetworksByAZ(n.env, az)
+			if len(extNetIds) > 0 {
+				extNetworkIds = append(extNetworkIds, extNetIds...)
+			}
+		}
+
+		if len(extNetworkIds) == 0 {
+			return nil, errors.NotFoundf(
+				"could not find an external network in availability zone %s", netDetails.AvailabilityZones)
```
wallyworld, line 219: this error message could use some tweaking.
1. it is a iist of zones and messages says "zone"
2. for the az="" case (the bug fix), the message doesn't seem quite right

wallyworld (approved):
Just a small error message tweak

hmlanigan:
$$merge$$

PR #12279, opened 2020-11-09
[gce] Update compute auth setup

SimonRichardson:
The previous way of connecting to the compute service was deprecated via
this warning[1] in the documentation. This was spotted when
investigating the TLS handshake timeout[2].

The fix is to update to the lastest compute construction with the new
token and scopes.

 1. https://godoc.org/google.golang.org/api/compute/v1#New
 2. https://bugs.launchpad.net/juju/+bug/1899793

## QA steps

Correctly sort the credentials out for GCE then do the following:

```sh
juju bootstrap google test
juju deploy cs:mysql
```
And wait for a very long time...

## Bug reference

https://bugs.launchpad.net/juju/+bug/1899793

### DIFF
--- a/provider/gce/google/auth.go
+++ b/provider/gce/google/auth.go
@@ -5,11 +5,21 @@ package google
 
 import (
 	"context"
+	"net/http"
+	"time"
 
 	"github.com/juju/errors"
 	"golang.org/x/oauth2/google"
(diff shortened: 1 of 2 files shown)


manadart (approved):
Nice.

SimonRichardson:
$$merge$$

PR #12289, opened 2020-11-10
Cleanup error messages when shutting down units

jameinel:
When a K8s unit is stopped, we end up with a log line that looks like:
```
2020-11-10 10:06:36 INFO juju.worker.uniter uniter.go:289 unit "discourse/0" shutting down: %!s(<nil>)
```

The intent is that it isn't an error for a K8s unit to be stopped, and we shouldn't treat it as an error. I'm happy to bike shed what message we want to be giving, but we certainly don't want to be giving `%!s(<nil>)`.

## Checklist

## QA steps

I have not done my own QA, so I would appreciate someone doing explicit QA and confirming these steps (and update this message).

```sh
$ juju bootstrap microk8s
$ juju deploy cs:~juju/mariadb-k8s-3
$ juju config mariadb-k8s query-cache-type=ON
$ juju debug-log
# see that the log message indicates why we are stopping mariadb-k8s/0
```

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1903726

### DIFF
--- a/worker/uniter/uniter.go
+++ b/worker/uniter/uniter.go
@@ -269,8 +269,13 @@ func (u *Uniter) loop(unitTag names.UnitTag) (err error) {
 	defer func() {
 		// If this is a CAAS unit, then dead errors are fairly normal ways to exit
 		// the uniter main loop, but the parent operator agent needs to keep running.
+		errorString := "<unknown>"
+		if err != nil {
+			errorString = err.Error()
+		}
 		if errors.Cause(err) == ErrCAASUnitDead {
 			err = nil
+			errorString = "caas unit dead"
 		}
 		if u.runListener != nil {
 			u.runListener.UnregisterRunner(unitTag.Id())
@@ -278,7 +283,7 @@ func (u *Uniter) loop(unitTag names.UnitTag) (err error) {
 		if u.localRunListener != nil {
 			u.localRunListener.UnregisterRunner(unitTag.Id())
 		}
-		u.logger.Infof("unit %q shutting down: %s", unitTag.Id(), err)
+		u.logger.Infof("unit %q shutting down: %s", unitTag.Id(), errorString)
 	}()
 
 	if err := u.init(unitTag); err != nil {


SimonRichardson (approved):

worker/uniter/uniter.go
```diff
@@ -269,16 +269,21 @@ func (u *Uniter) loop(unitTag names.UnitTag) (err error) {
 	defer func() {
 		// If this is a CAAS unit, then dead errors are fairly normal ways to exit
 		// the uniter main loop, but the parent operator agent needs to keep running.
+		errorString := "<unknown>"
+		if err != nil {
+			errorString = err.Error()
+		}
 		if errors.Cause(err) == ErrCAASUnitDead {
```
achilleasa, line 276: We could change `ErrCAASUnitDead` error from "unit dead" to "caas unit dead" and set `errorString = ErrCAASUnitDead.String()`

worker/uniter/uniter.go
```diff
@@ -269,16 +269,21 @@ func (u *Uniter) loop(unitTag names.UnitTag) (err error) {
...
+			errorString = "caas unit dead"
 		}
 		if u.runListener != nil {
 			u.runListener.UnregisterRunner(unitTag.Id())
 		}
 		if u.localRunListener != nil {
 			u.localRunListener.UnregisterRunner(unitTag.Id())
 		}
-		u.logger.Infof("unit %q shutting down: %s", unitTag.Id(), err)
+		u.logger.Infof("unit %q shutting down: %s", unitTag.Id(), errorString)
```
achilleasa, line 286: Should this be an Errorf if errorString is not empty?

achilleasa (approved):
LGTM; left some minor comments

worker/uniter/uniter.go
jameinel: This will be triggered any time the loop exits, which might be just because the unit is being removed. So I don't think it is universally an error, which is why it is at INFO level.

jameinel:
$$merge$$

PR #12291, opened 2020-11-10
Prevent fetching of oci timing out snap install

SimonRichardson:
The following wraps the installing of the oci image in a timeout. As we
don't actually require the image to be installed when we bootstrap we
can be a bit fast and loose about it. If you don't have timeout
installed (BSD and macosx) then we exit and install the image using
manual intervention/bootstrap.

The code ensures that if it takes longer than we expect we just cancel
it, but importantly ignore the timeout error and return true. That way
we can still install the snap.

I've used 2 minutes because it can take a long time to install a snap on
slower systems, so we don't want to gobble that time up installing an
image that might not install anyway.

## QA steps

Ensure you have microk8s installed.
If you've built the snap before then it's wise to do `lxc delete snapcraft-juju --force`

```sh
snap install snapcraft --classic
snapcraft --use-lxd
sudo snap install *.snap --dangerous --classic
sudo snap logs juju
```

Ensure that we did in fact attempt to pull the images down.

### DIFF
--- a/snap/local/wrappers/fetch-oci
+++ b/snap/local/wrappers/fetch-oci
@@ -13,7 +13,7 @@ if ! which timeout; then
     exit 0
 fi
 
-timeout 2m sh <<EOT || echo true
+timeout 2m sh <<EOT || echo "Timed out waiting to install microk8s image."
 container_cmd=microk8s.ctr
 image_arg="--namespace k8s.io images"
 if ! which $container_cmd; then
--- a/snap/local/wrappers/fetch-oci
+++ b/snap/local/wrappers/fetch-oci
@@ -13,7 +13,6 @@ if ! which timeout; then
     exit 0
 fi
 
-timeout 2m sh <<EOT || echo "Timed out waiting to install microk8s image."
 container_cmd=microk8s.ctr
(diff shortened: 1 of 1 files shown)


snap/local/wrappers/fetch-oci
```diff
@@ -8,6 +8,12 @@ if ! which microk8s.kubectl; then
     exit 0
 fi
 
+if ! which timeout; then
+    echo "timeout is not installed."
+    exit 0
+fi
+
+timeout 2m sh <<EOT || echo true
```
petevg, line 16: `|| true` should be sufficient here, correct? Or echo "microk8s docker images downloading in bg" if you want to give a user some information.

petevg (approved):
This makes sense. Haven't had time to QA, and had a minor nitpick w/ the code. But this is a nice, elegant fix. :-)

snap/local/wrappers/fetch-oci
SimonRichardson: I think you're right, let me change this before it lands. It would be much better to add a message here.

manadart (approved):

SimonRichardson:
$$merge$$

PR #12292, opened 2020-11-10
Fix intermittent fail in TestWatchPorts with clock.Advance timing issue

benhoyt:
This is happening sometimes (quite rarely!) because the test clock's Advance method (called by the test) may sometimes be called before the TxnWatcher.loop() method has called w.clock.After(d) (state/watcher/txnwatcher.go:231).

Fix is to add a channel to StatePool that we can wait on for testing (some of the machinery is already there, with the existing txnWatcherStarting event published).

Note: this test is called TestWatchMachinePorts in the develop branch.

For an example, see https://jenkins.juju.canonical.com/job/Unit-RunUnitTests-centos7/76/ which shows the following:

```
[LOG] 0:00.337 DEBUG juju.clock advancing a clock that has nothing waiting: cf. https://github.com/juju/juju/wiki/Intermittent-failures
machine_ports_test.go:436:
    wc.AssertChange(expectChange)
testing/watcher.go:234:
    c.Fatalf("watcher did not send change")
... Error: watcher did not send change
```

## QA steps

To repro the failure, add `time.Sleep(1 * time.Second)` after line 227 before the clock.Now() in `state/watcher/txnwatcher.go` (oh, and uncomment the `s.WaitHubWatcherStarting(c)` line that fixes the test, of course).
```

### DIFF
--- a/state/testing/suite.go
+++ b/state/testing/suite.go
@@ -45,6 +45,7 @@ type StateSuite struct {
 	RegionConfig              cloud.RegionConfig
 	Clock                     *testclock.Clock
 	txnSyncNotify             chan struct{}
+	hubWatcherStarting        chan struct{}
 	modelWatcherIdle          chan string
 	modelWatcherMutex         *sync.Mutex
 }
@@ -64,10 +65,12 @@ func (s *StateSuite) SetUpTest(c *gc.C) {
 	s.BaseSuite.SetUpTest(c)
 
 	s.txnSyncNotify = make(chan struct{})
+	s.hubWatcherStarting = make(chan struct{})
 	s.modelWatcherIdle = nil
 	s.modelWatcherMutex = &sync.Mutex{}
 	s.PatchValue(&statewatcher.TxnPollNotifyFunc, s.txnNotifyFunc)
 	s.PatchValue(&statewatcher.HubWatcherIdleFunc, s.hubWatcherIdleFunc)
+	s.PatchValue(&statewatcher.HubWatcherStartingFunc, s.hubWatcherStartingFunc)
 
 	s.Owner = names.NewLocalUserTag("test-admin")
 	initialTime := s.InitialTime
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: state/testing/suite.go [hunk 1]
```diff
@@ -142,6 +145,19 @@ func (s *StateSuite) hubWatcherIdleFunc(modelUUID string) {
 	}
 }
 
+func (s *StateSuite) hubWatcherStartingFunc() {
+	s.hubWatcherStarting <- struct{}{}
```
jameinel, line 149: Do we want to do this or do we want:

s.hubWatcherStarted as the channel and then to do
```
close(s.hubWatcherStarted)
```
the main reason for close is if 0 or more units actually listen for the event, they can all get told, and it won't block the other goroutine. If you intentionally want the other routine to wait until we observe that it has started then what you have done is correct, except if there are 0 units that will ever listen for the event, then you have just caused this code to block indefinitely.
So you really need to wrap it in some sort of select{} that will see some other sort of event like tomb.Dying() or just a simple timeout so the test doesn't hang forever if you have a logic error.

state/watcher/hubwatcher.go
```diff
@@ -175,7 +182,10 @@ func newHubWatcher(hub HubSource, clock Clock, modelUUID string, logger Logger)
 func (w *HubWatcher) receiveEvent(topic string, data interface{}) {
 	switch topic {
 	case txnWatcherStarting:
-		// We don't do anything on a start.
+		if w.startingFunc != nil {
```
jameinel, line 185: I wonder if we want the passed in function, vs just always closing a channel that we always have that others can look to listen on. So something ilke:

```
// Started waits until the watcher has started and then returns.
// If the HubWatcher is already started, this returns immediately.
func HubWatcher Started() error {
  select {
    case <-tomb.Dying():
      return ErrDying ?(or 
    case <- s.started:
      return nil
}
```

I'm not sure we want the error return. We might just always return if we are dead or started, as it doesn't matter much and simplifies the code.

state/watcher/txnwatcher.go
```diff
@@ -29,9 +29,9 @@ type Clock interface {
 }
 
 const (
-	txnWatcherStarting   = "starting"
-	txnWatcherSyncErr    = "sync err"
-	txnWatcherCollection = "collection"
+	TxnWatcherStarting   = "starting"
```
wallyworld, line 32: if these are now exported should add a comment

wallyworld (approved):

benhoyt:
$$merge$$

PR #12293, opened 2020-11-10
Filter mongo addresses we try to use to exclude fan, public

wallyworld:
When dialling mongo, we pull out all of the possible controller api addresses from agent.conf and attempt to use them. Or more correctly, we pass them to the mgo client to use. This includes fan and public addresses. mgo client will try them all and pick the first one that connects. If that's the fan address, subsequent attempts to use that connection in a mongo session fail. 

We should be filtering on cloud local addresses. This is actually what's done in the peer grouper worker which maintains the mongo replicaset. So this PR simply uses the same logic. ie we configure the controller with the same addresses mongo itself uses to talk to itself in the cluster.

## QA steps

Unable to reproduce the actual failure. But enable-ha was tested on an aws bootstrap which did have fan addresses present.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1903557

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -898,6 +898,23 @@ func (c *configInternal) MongoInfo() (info *mongo.MongoInfo, ok bool) {
 	if !ok {
 		return nil, false
 	}
+	addrs := c.apiDetails.addresses
+	var netAddrs network.SpaceAddresses
+	for _, addr := range addrs {
+		host, _, err := net.SplitHostPort(addr)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


ycliuhw (approved):
lgtm thanks

wallyworld:
$$merge$$

wallyworld:
[enable-ha connect error](https://trello.com/c/uvb2K4Ap/1271-enable-ha-connect-error)

PR #12297, opened 2020-11-11
Include controller FQDN host in api address for k8s agents

wallyworld:
If the IP address of a k8s controller changes because the cluster itself changes, deployed agents have no way of communicating to the controller anymore as they use a single IP address of the controller. 
This PR includes in the set of available address in agent.conf the FQDN of the controller based on the cluster DNS, eg
controller-service.controller-test.svc.cluster.local

## QA steps

I didn't have a means to easily get microk8s a new ip address, but I hacked the api address code to only return the FQDN host and tested deploying a charm and confirmed that the agent.conf contained the host name and agent could talk to the controller.
I also verified that the non-hacked code wrote out both the IP address and FQDN of the controller, eg

```
apiaddresses:
- 10.152.183.204:17070
- controller-service.controller-ian.svc.cluster.local:17070
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1900930

### DIFF
--- a/state/address.go
+++ b/state/address.go
@@ -16,6 +16,7 @@ import (
 	"gopkg.in/mgo.v2/bson"
 	"gopkg.in/mgo.v2/txn"
 
+	k8sprovider "github.com/juju/juju/caas/kubernetes/provider"
 	"github.com/juju/juju/core/network"
 	"github.com/juju/juju/mongo"
 )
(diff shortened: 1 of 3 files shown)


state/address.go
```diff
@@ -310,23 +310,36 @@ func (st *State) apiHostPortsForCAAS(public bool) (addresses []network.SpaceHost
 		return addrsToHostPorts(publicAddrs...), nil
 	}
 
+	var hostAddresses network.SpaceAddresses
+	// Add in the FQDN of the controller service for agents to use as an option.
+	controllerName := controllerConfig.ControllerName()
+	if controllerName != "" {
+		hostAddresses = append(
+			hostAddresses, network.NewScopedSpaceAddress(
+				fmt.Sprintf("controller-service.controller-%s.svc.cluster.local", controllerName),
```
hpidcock, line 319: @wallyworld this should probably be in github.com/juju/juju/caas/kubernetes/provider/constants

state/address_test.go
```diff
@@ -376,6 +377,9 @@ type CAASAddressesSuite struct {
 var _ = gc.Suite(&CAASAddressesSuite{})
 
 func (s *CAASAddressesSuite) SetUpTest(c *gc.C) {
+	s.ControllerConfig = map[string]interface{}{
+		controller.ControllerName: "trump",
```
hpidcock, line 381: heh?

hpidcock (approved):

state/address.go
hpidcock: Also maybe a comment saying this is a caas todo hack

wallyworld:
$$merge$$

wallyworld:
[microK8s restart breaks juju](https://trello.com/c/DRKPUAy4/1274-microk8s-restart-breaks-juju)

PR #12299, opened 2020-11-11
Handle user charmstore urls

SimonRichardson:
The following ensures that we correctly handle when there is no prefix
for a charmstore url. The way the url path is handled in the library
means we need to do a bit more work upfront if we want to maintain
strict compatibility. This was not the original goal of the library as
the feature flag was intended to be lifted by now, but that's how it is.

## QA steps

Ensure you don't have the feature flag enabled for charmhub.

```sh
juju bootstrap lxd test
juju deploy ~juju-qa/network-health-10
```


manadart (approved):
Looks good, works fine.

SimonRichardson:
$$merge$$

PR #12300, opened 2020-11-11
Charmhub: Optimise channel map find and subset filtering.

SimonRichardson:
The following ensures that we're only requesting the fields that we
require when using info and find. Unfortunately this doesn't know about
channel-map metadata.yaml and config.yaml so that needs to be worked in
correctly.

There is a missing piece of work where we'll need to request the info for a given
channel that will pass back the metadata.yaml and config.yaml back. This is
found in `apiserver/facades/client/charms/reqpositories.go#L169`. This is
where we look up the info channel map by either the channel or the revision.

Once the new API is up to do that, we should change this to correctly implement
that lookup.

## QA steps

What currently works:

```sh
juju bootstrap lxd test
juju info ubuntu
juju find ubuntu
juju find
```


charmhub/filter.go
```diff
@@ -34,17 +34,20 @@ var defaultResultFilter = []string{
...
-	"result.media.url",
-	"result.media.width",
 	"result.publisher.display-name",
 	"result.store-url",
 	"result.summary",
 	"result.used-by",
 	"result.website",
 }
 
+var defaultMediaFilter = []string{
```
hmlanigan, line 44: I think we can drop the result.media in general.

charmhub/find.go
```diff
@@ -39,7 +39,14 @@ func (c *FindClient) Find(ctx context.Context, query string) ([]transport.FindRe
 		return nil, errors.Trace(err)
 	}
 
-	path, err = path.Query("fields", defaultFindFilter())
+	var fields string
+	if query != "" {
+		fields = defaultFindFilter()
+	} else {
+		fields = subsetFindFilter()
+	}
```
hmlanigan, line 47: Let's go for it and only request the Find data we use: names, bundle, version, publisher and summary?  We do not print anything different based on query vs all.

hmlanigan (approved):
QA is good.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

> FAIL: txnwatcher_test.go:281: TxnWatcherSuite.TestInsertThenRemove

PR #12301, opened 2020-11-11
[2.8] Request a NIC per unique space for MAAS instances

achilleasa:
Prior to this change, the constraints mapping code in the MAAS provider
would iterate the list of endpoint bindings and request an interface for
each individual endpoint even thought they might all be bound to the
same space.

As a result, when MAAS was asked to synthesize a KVM node with those
constraints it would generate multiple NICs labeled after each endpoint
thus triggering LP1902878.

Instead of having two separate code paths (v1 and v2) that intersected
(getBindings) and then diverged again, the fix introduced by this commit
extracts the space requirement calculation logic into a separate method
that is invoked by StartInstance and whose results are passed into
acquireNode/acquireNode2. This new method calculates the unique set of
endpoints from the application bindings (if provided) and appends any
positive space constraints. Then it calculates the negative space set
and checks for clashes with the positive set.

After this change, various bits of logic have been made redundant and
therefore removed. In addition, new tests have been added for the
introduced space requirement calculation logic and the existing
AcquireNode tests have been cleaned up as the validation logic is
applied before they are invoked.

## QA steps

Setup you virtual maas instance with a KVM instance and create 3 spaces: space1, space2, space3.
Assign each space to a different subnet and ensure that **none** of your already composed KVM pods
have a NIC in spaces 1 and 2 (the goal is to have MAAS synthesize a pod for us based on the constraints)

```sh
$ juju model-config logging-config='<root>=INFO;juju.provider.maas=TRACE'
$ juju deploy apache2 --bind space1 --constraints spaces=space2,^space3
```

Then check the provisioned KVM pod in MAAS web UI and verify that you have two NICs, one on space1 and one on space2.
If you check the model logs, you should be able to spot the constraints being passed to maas (IDs will probably be different in your case): "interfaces=space=1;space=2;not_subnets=space=3"

Try the same steps with juju from snap and you should see a NIC per endpoint name.

For consistency, we should try the same with a V1 MAAS.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1902878

### DIFF
--- a/provider/maas/constraints.go
+++ b/provider/maas/constraints.go
@@ -9,11 +9,9 @@ import (
 	"strings"
 
 	"github.com/juju/collections/set"
-	"github.com/juju/errors"
 	"github.com/juju/gomaasapi"
 
 	"github.com/juju/juju/core/constraints"
-	"github.com/juju/juju/core/network"
 	"github.com/juju/juju/environs/context"
 )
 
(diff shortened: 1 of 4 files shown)


provider/maas/constraints.go
```diff
@@ -133,172 +131,46 @@ func parseDelimitedValues(rawValues []string) (positives, negatives []string) {
...
-		params.Interfaces = interfaceSpecs
-	}
-	if len(negatives) > 0 {
-		negativeStrings := make([]string, len(negatives))
-		for i, space := range negatives {
-			negativeStrings[i] = space.SpaceProviderId
-		}
-		params.NotSpace = negativeStrings
+	if len(negativeSpaceIDs) != 0 {
+		params.NotSpace = negativeSpaceIDs.SortedValues()
```
achilleasa, line 172: Using SortedValues ensures consistent order for params which we need for tests.

provider/maas/environ.go
```diff
@@ -775,26 +775,89 @@ func (env *maasEnviron) spaceNamesToSpaceInfo(
...
+	// positive space set.
+	for epName, providerSpaceID := range endpointToProviderSpaceID {
+		// The default space is applied at deploy time to all endpoints
+		// that are not explicitly bound but should be ignored when
+		// collecting the spaces we need. Consider this edge case:
+		// an explicit space binding is provided for all endpoints as
+		// well as a default one that is *different* from all others.
+		// In this case, we shouldn't request for a NIC in the default
+		// space as it is not actually going to be used.
+		if epName == "" {
```
achilleasa, line 797: I have added this extra guard to prevent the creation of an unused interface in the edge case where all endpoints are explicitly bound to a space different than the default value.

jameinel:
I'm not sure if that is correct behavior. If you have a default binding to
foo, even if it isn't currently used we should still ensure that you have
access to foo. (we do that already for --constraints spaces=foo with no
other bindings)

It is an edge case but I think including the device is arguably correct.

wallyworld (approved):
Very nice

achilleasa:
> I'm not sure if that is correct behavior. If you have a default binding to foo, even if it isn't currently used we should still ensure that you have access to foo. (we do that already for --constraints spaces=foo with no other bindings) It is an edge case but I think including the device is arguably correct.

I guess that extra check is too conservative. I will push a commit to remove it. We may get an extra unused NIC in the edge case but users still have the flexibility to rebind an endpoint to the default space value.

manadart (approved):
Not bothering with MAAS 1.x.

1) It doesn't have pods.
2) We are about to drop support for it.

achilleasa:
$$merge$$

PR #12305, opened 2020-11-12
Try and keep incomplete actions when pruning

wallyworld:
When pruning actions by size, first try and prune only completed actions to achieve the goal.
Only if needed to get the collection under the required size, then prune any incomplete actions.

## QA steps

Hard to QA manually - the state unit tests for pruning mongo collections exercise the behaviour quite well.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1902727

### DIFF
--- a/state/prune.go
+++ b/state/prune.go
@@ -64,6 +64,18 @@ func pruneCollectionAndChildren(mb modelBackend, maxHistoryTime time.Duration, m
 	if err := p.pruneByAge(); err != nil {
 		return errors.Trace(err)
 	}
+	// First try pruning, excluding any items that
+	// have an age field that is not yet set.
+	// ie only prune completed items.
+	if err := p.pruneBySize(); err != nil {
+		return errors.Trace(err)
+	}
+	if ageField == "" {
+		return nil
+	}
+	// If needed, prune additional incomplete items to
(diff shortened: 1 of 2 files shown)


wallyworld:
[do not prune by size incomplete actions by default](https://trello.com/c/hmLgnmqb/1275-do-not-prune-by-size-incomplete-actions-by-default)

hpidcock (approved):
LGTM thanks

wallyworld:
$$merge$$

state/prune.go
```diff
@@ -64,6 +64,18 @@ func pruneCollectionAndChildren(mb modelBackend, maxHistoryTime time.Duration, m
 	if err := p.pruneByAge(); err != nil {
 		return errors.Trace(err)
 	}
+	// First try pruning, excluding any items that
+	// have an age field that is not yet set.
+	// ie only prune completed items.
+	if err := p.pruneBySize(); err != nil {
```
jameinel, line 70: I feel like this would have been cleaner if you passed in the ageField rather than having it as an attribute of 'p'. The fact that you have to mutate the value of p.ageField and call a function twice expecting different side effects seems like it would be better as a parameter.

PR #12307, opened 2020-11-12
Info unicode output

SimonRichardson:
One of the missing features was adding unicode support to the info
command so that when you supplied info it would naturally add the
correct types in the output.

The following code adds that for info. If required we can move it to the
output folder if others find it more useful. It's basic in its primitive
form but does exactly what we want by identifying a location in the
code to print a Unicode character out.

By default, it is set to "auto" which is the same as snap cli and that it
will work out if your system can support it. It seems rather sane so
I've lifted in from there as they don't expose that function for
detection.

Unfortunately, we don't have the same CLI structure to lift more of the
code which is a shame, but this is the way.

## QA steps

The interesting part is that there were channels when I tested this
yesterday, but that is no longer the case. I modified the code to show this
in a hardcoded way by editing the info command channels.

```sh
juju bootstrap lxd test
juju info mysql
```


cmd/juju/charmhub/info.go
```diff
@@ -91,6 +94,14 @@ func (c *infoCommand) Init(args []string) error {
 		return errors.Trace(err)
 	}
 	c.charmOrBundle = args[0]
+
+	switch c.unicode {
+	case "auto", "never", "always":
+	case "":
```
hmlanigan, line 100: how is this possible if the default is auto?  `juju info mysql --unicode "" ` is an error in my mind.

hmlanigan (commented):
I'm unable to get good qa results on this right now.  juju info ubuntu-qa doesn't give correct channel data to make this work due to a bug, and juju info mysql has no channel data displayed.   Neither is error caused by this PR.

hmlanigan (approved):
approving based on TestBundleChannelClosedWithUnicode

SimonRichardson:
$$merge$$

PR #12310, opened 2020-11-12
Charmhub not found errors

SimonRichardson:
The following exposes the "not found" errors for a given request. There
are other errors we could be exposing (403, 500 for example) and showing
to the user.

This is the first pass that we do that and in a subsequent PR we can
handle that by exposing an error message to the user if required
explaining that the charm/bundle could not be found.

The code is simple, we return a `RESTResponse` that has a status code 
`int` that we can check against the `http.StatusCode` before returning a
different error based on the Error from the api response. It's a bit
convoluted because the status code is in the header and the message is
in a json blob.

## QA steps

```sh
juju bootstrap lxd test
juju deploy meshuggah
```

### DIFF
--- a/charmhub/http.go
+++ b/charmhub/http.go
@@ -202,7 +202,7 @@ func (c *HTTPRESTClient) Post(ctx context.Context, path path.Path, body, result 
 
 	// Parse the response.
 	if err := httprequest.UnmarshalJSONResponse(resp, result); err != nil {
-		return RESTResponse{}, errors.Annotate(err, "charm hub client get")
+		return RESTResponse{}, errors.Annotate(err, "charm hub client post")
 	}
 	return RESTResponse{
 		StatusCode: resp.StatusCode,
--- a/charmhub/http.go
+++ b/charmhub/http.go
@@ -74,12 +74,17 @@ func (t *APIRequester) Do(req *http.Request) (*http.Response, error) {
 	return resp, nil
 }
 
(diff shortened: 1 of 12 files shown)


charmhub/http.go
```diff
@@ -194,9 +202,11 @@ func (c *HTTPRESTClient) Post(ctx context.Context, path path.Path, body, result
 
 	// Parse the response.
 	if err := httprequest.UnmarshalJSONResponse(resp, result); err != nil {
-		return errors.Annotate(err, "charm hub client get")
+		return RESTResponse{}, errors.Annotate(err, "charm hub client get")
```
hmlanigan, line 205: s/get/post/ -  while you're here.

hmlanigan (commented):
There was no change to user output when I ran 'juju find foo', wouldn't have expected any difference with deploy, as it does not use find.

SimonRichardson:
> There was no change to user output when I ran 'juju find foo'...

That's not actually true as everything is server side you'd have to rebootstrap/upgraded the controller to see the change.

#### Prior

```
➜ juju info bob
ERROR No charm or bundle with name 'bob'.
```

#### Post

```
➜ juju info bob
ERROR charm or bundle: No charm or bundle with name 'bob'. (not found)
```

hmlanigan (approved):
output resolved.  lgtm

hmlanigan:
merging so I can rebase for #12311 

$$merge$$

PR #12311, opened 2020-11-12
Add ResourcesClient for charmhub

hmlanigan:
Includes the new client, response structures and ListResourceRevisons.
A new api endpoint for charmhub from: http://api.snapcraft.io/docs/charms.html#list_resource_revisions

There may be changes up coming to have the same property names, avoid hash-sha256 vs hash-sha256.

## QA steps

No change to current charmhub client behavior.

### DIFF
--- a/charmhub/client.go
+++ b/charmhub/client.go
@@ -114,12 +114,13 @@ func (c Config) BasePath() (charmhubpath.Path, error) {
 
 // Client represents the client side of a charm store.
 type Client struct {
-	url            string
-	infoClient     *InfoClient
-	findClient     *FindClient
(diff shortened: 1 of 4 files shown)


SimonRichardson (commented):
I think I want #12310 to land first and then rebase.

Also, are we're missing filters here?

hmlanigan:
The API doc does not mention fields, though it is not approved yet, so that may change.  I do not have a way to verify the data yet. 

I will add fields/filters when adding resources array to the find and info results.

hmlanigan:
This client is not needed yet, if never

PR #12313, opened 2020-11-13
Retain Azure custom resource group contents

wallyworld:
When bootstrapping on Azure using a BYO resource group, and the bootstrap fails, Juju would purge the entire resource group. This would also happen if a model using a BYO resource group were destroyed.

To fix the issue, 2 key changes are implemented:
- on a failed bootstrap, cancel the existing ARM deployment; this will remove only juju resources that were provisioned udring bootstrap
- when destroying a model, first remove only Juju machines; this would also delete any machine resources like NICs etc; then loop over the remaining Juju resources and delete, catching any errors due to the resource being in use and try again

Any non-Juju tagged resources are left alone.

## QA steps

Ensure a bootstrap on Azure will fail (eg see bug)
juju bootstrap to an existing resource group (see bug)
Ensure bootstrap operation terminate and only the juju bootstrap machine and nic and public ip etc are removed
(any existing vnet etc in the resource group are retained)

bootstrap successfully and add a model using an existing resource group
destroy the model
again, check that any non-juju content is retained

## Bug reference

https://bugs.launchpad.net/juju/+bug/1904020

### DIFF
--- a/provider/azure/environ.go
+++ b/provider/azure/environ.go
@@ -21,6 +21,7 @@ import (
 	azurestorage "github.com/Azure/azure-sdk-for-go/storage"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	"github.com/juju/names/v4"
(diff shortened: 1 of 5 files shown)


provider/azure/environ.go
```diff
@@ -1765,41 +1786,142 @@ func (env *azureEnviron) deleteResourceGroup(ctx context.ProviderCallContext, sd
...
 	}
-	future, err := client.CreateOrUpdate(
-		sdkCtx,
-		resourceGroup,
-		"purge-resource-group",
-		deployment,
+
+	// Older APIs can ignore the filter above, so query the hard way just in case.
+	if len(resourceItems) == 0 {
+		result, err = resourcesClient.ListByResourceGroupComplete(sdkCtx, resourceGroup, "", "", nil)
```
manadart, line 1815: This could be extracted, shortening this lengthy method.

provider/azure/environ.go
```diff
@@ -1765,41 +1786,142 @@ func (env *azureEnviron) deleteResourceGroup(ctx context.ProviderCallContext, sd
...
+	// Stopping instances will also remove most of their dependent resources.
+	err = env.StopInstances(ctx, ids...)
 	if err != nil {
-		return errors.Annotatef(err, "purging resource group %q", resourceGroup)
+		return errors.Annotatef(err, "deleting machine instances %q", ids)
+	}
+
+	// Loop until all remaining resources are deleted.
+	remainingResources := otherResources
+	for len(remainingResources) > 0 {
```
manadart, line 1866: It it worth putting a ceiling on how long we will keep trying this?

manadart (commented):
Code looks good modulo a couple of suggestions. I have not done QA yet.

provider/azure/environ.go
wallyworld: Add for safety, in reality it will be at most 2 passes

ycliuhw (approved):
lgtm thanks

wallyworld:
$$merge$$

wallyworld:
[azure resource group fix](https://trello.com/c/znNO1AWM/1278-azure-resource-group-fix)

PR #12321, opened 2020-11-16
Account for failed hook retries in uniter worker tests

wallyworld:
The way uniter worker tests are done, it is possible for tests which use failed hooks to sneak in an extra retry before the test checks the expected vs observed hooks. This PR adds a little extra code to skip over duplicate failed hook retries when it does the matching.

Failure is logged as having extra hooks in UniterSuite.TestUniterInstallHook and UniterSuite.TestUniterStartHook  (the extra failed hook retry)

```
util_test.go:777:
    c.Fatalf("ran more hooks than expected")
... Error: ran more hooks than expected
```

## QA steps

Unit tests  pass as before. I hacked in a extra hook run for any failed hooks the tests may have queued up and they continued to pass (but failed without the fix or with the allowedRetryCount set to 0).

### DIFF
--- a/worker/uniter/util_test.go
+++ b/worker/uniter/util_test.go
@@ -207,17 +207,53 @@ func (ctx *context) matchHooks(c *gc.C) (match, cannotMatch, overshoot bool) {
 	defer ctx.mu.Unlock()
 	c.Logf("actual hooks: %#v", ctx.hooksCompleted)
 	c.Logf("expected hooks: %#v", ctx.hooks)
-	for i, e := range ctx.hooks {
-		if i >= len(ctx.hooksCompleted) {
+
+	// If hooks are automatically retried, this may cause stutter in the actual observed
+	// hooks depending on timing of the test steps. For the purposes of evaluating expected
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
LGTM. My only question (without digging into the surrounding code): are these hooks ordered in such a way that the concept of a "previous" hook is meaningful?

wallyworld:
> LGTM. My only question (without digging into the surrounding code): are these hooks ordered in such a way that the concept of a "previous" hook is meaningful?

For the purposes of how the tests are written, it's ok as is. The tests add a single failed hook, check it, and moved on. And also, the hooks which fail are install or start which prevent other hooks from being queued up regardless.

wallyworld:
$$merge$$

wallyworld:
[UniterSuite.TestUniterInstallHook](https://trello.com/c/mmAtNXmc/167-unitersuitetestuniterinstallhook)

wallyworld:
[UniterSuite.TestUniterStartHook](https://trello.com/c/JB6ZEmGX/169-unitersuitetestuniterstarthook)

PR #12320, opened 2020-11-16
Change couple of import jujuversion to match develop (easier merge)

benhoyt:
When resolving conflicts for https://github.com/juju/juju/pull/12319, I did a few too many imports of `import jujuversion "github.com/juju/juju/version"` when it wasn't necessary and was just done as `import "github.com/juju/juju/version"` on the develop branch. Update these to make forward merging less painful.

PR #12319, opened 2020-11-16
Pull in latest juju/os to support macOS Big Sur

benhoyt:
Unfortunately the [juju/os](https://github.com/juju/os) repo's master branch had a breaking change applied to it, specifically the removal of `version.DefaultSupportedLTS()` (it moved into `juju/juju`). So updating required the same refactoring as we'd done on develop to use the `DefaultSupportedLTS` function in `juju/juju/version`.

We're pulling in these changes from `juju/os`, so the reviewer should also look here: https://github.com/juju/os/compare/8e6dd7a2b438d448578b32d0115ca0daaabb6d1f...10720519c304290ca6b6e7586a60ce0015323cfe -- I don't *think* the other changes should be a problem, as we're not changing the default series.

These changes were from https://github.com/juju/juju/pull/12103/files and https://github.com/juju/juju/pull/12108/files cherry-picked, with some manual fix-ups and a few conflict resolutions due to 2.8/develop differences.

## QA steps

Run `juju status` on macOS Big Sur and see if it works. I can't test as I don't have access to macOS Big Sur, but we'll ask the bug reporter to check. The relevant fix itself is a straight-forward one-liner in https://github.com/juju/os/commit/10720519c304290ca6b6e7586a60ce0015323cfe.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1904252

PR #12324, opened 2020-11-16
Revert juju/os fix

benhoyt:
This reverts the commits from my Big Sur fixes https://github.com/juju/juju/pull/12320 and https://github.com/juju/juju/pull/12319 (in reverse order), including the `juju/os` module update. We don't want to update juju/os for 2.8, as it will deploy charms that default to focal using focal, whereas previously it would have been deployed to bionic as 2.8 didn't support focal. Ian's summary:

> This can cause accidental breakage, albeit unlikely. Consider the case where a charm supports focal, bionic, xenial. If focal is the default series for the charm, then
>
> `juju deploy foo`
>
> which fire up a focal machine, whereas before this fix it would have been bionic. So that's a change in behaviour we don't want in a point release.

So we're going to revert this approach and use a much smaller juju/os patch instead (with a v2 branch of juju/os for 2.9).


hmlanigan (approved):
looks good to me.

benhoyt:
$$merge$$

PR #12328, opened 2020-11-17
Use juju/os v1.0.0 on 2.8 to pull in macOS Big Sur fix

benhoyt:
Safer replacement for https://github.com/juju/juju/pull/12319, which we reverted as it pulled in too many juju/os changes. This updates to the new v1.0.0 version of juju/os, which is the previous version used on 2.8 (8e6dd7a2b438) with the one-line Big Sur fix added.

## QA steps

Run `juju status` on macOS Big Sur and see if it works. I can't test as I don't have access to macOS Big Sur, but we'll ask the bug reporter to check. The relevant fix itself is a straight-forward one-liner in https://github.com/juju/os/commit/10720519c304290ca6b6e7586a60ce0015323cfe.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1904252

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -62,7 +62,7 @@ require (
 	github.com/juju/mutex v0.0.0-20180619145857-d21b13acf4bf
 	github.com/juju/names/v4 v4.0.0-20200424054733-9a8294627524
 	github.com/juju/naturalsort v0.0.0-20180423034842-5b81707e882b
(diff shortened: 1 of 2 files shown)


wallyworld (approved):

benhoyt:
$$merge$$

PR #12329, opened 2020-11-17
Use new v2 version of github.com/juju/os on develop branch

benhoyt:
This should not be a functional change at all, as the juju/os commit that was being used (10720519c304) includes the Big Sur fix already. This is essentially just a package name update from github.com/juju/os to github.com/juju/os/v2.

We also pull in new versions of bundlechanges, charm, charmrepo to use juju/os/v2 so that the main juju code is using only juju/os/v2.

Note, go.mod still includes github.com/juju/os (master branch) due to an indirect import, but it's only for the charmrepo tests (it imports charmstore which still uses juju/os v1).

```
$ go mod why github.com/juju/os
# github.com/juju/os
github.com/juju/juju/apiserver/facades/client/application
github.com/juju/charmrepo/v6
github.com/juju/charmrepo/v6.test
gopkg.in/juju/charmstore.v5
gopkg.in/juju/charmstore.v5/internal/charmstore
gopkg.in/juju/charm.v6
github.com/juju/os

```


wallyworld (approved):
The dep change for gopkg.in/macaroon-bakery.v2 could potentially be an issue.
Can we confirm that the change are ok before landing, or else revert that dep change.

benhoyt:
Yeah, I'd checked those changes too -- you can see the diff at https://github.com/go-macaroon-bakery/macaroon-bakery/compare/6734dc66fe81...v2.2.0 and I think it's a legitimate "minor version" bump. It adds a new type `ContextBacking` to allow calling the functions with a `ctx` argument, but it doesn't change the signature of existing functions. It's only if you pass in a backing which implements the `ContextBacking` interface that it uses it. So looks like a well thought-out, backwards-compatible change.

benhoyt:
I've also done a basic lxd bootstrap and deploy locally, and they see to work fine.

benhoyt:
$$merge$$

benhoyt:
[Updating to juju/os/v2](https://trello.com/c/HnBSQzhS/1284-updating-to-juju-os-v2)

PR #12330, opened 2020-11-17
lxd: Not all remotes have a certificate

stgraber:
Remotes can be running on valid TLS certificates in which case no file will be found in `servercerts`.
This is perfectly fine and should not result in an error. Instead the remote simply should not get a server certificate set at all, allowing the tls client to use the system CA.


wallyworld (commented):
As discussed live, we'll add a unit test and target this fix to 2.8

wallyworld:
PR #12335 takes the fix from here, adds unit tests, and targets 2.8

PR #12339, opened 2020-11-18
Charmhub download arch

SimonRichardson:
Update the charmhub download command to be arch aware. This PR
does a lot of things, so it's worth listing them out.

 1. Introduces a common charmhub base command for dealing with arch
     and series support.
 2. Allows the downloading of arch, series, or both together.
 3. Allows the downloading of closed channels by referring to their sibling.
     edge looks to beta, then to candidate, and then finally to stable.

For points 1 and 2, the code is relatively simple but quite destructive in 
terms of looking at it from a diff point of view. We throw away the test caching
in favor of passing the dependencies in via the constructor. This makes the
code production quality 100% of the time.

In terms of point 3, we standardize the filtering of the arch and series
support and favoring HOF (high order functions) to perform the filtering.
This makes the code concise and easier to read.

Additionally, the featuring of closed channels should behave like an alias
to the next non-closed channel. To prevent the issue of locating a closed
channel and then to recalculate the next non-closed channel and applying
the whole filter process again seems more error-prone. Instead, we
pre-compute all the closed channel aliases and apply the filtering logic once.

Potentially we could optimize this by selecting the channels first and so for
larger channel maps we could just do this for a small selection, but at the
moment performance doesn't seem a problem as correctness is more
important.

### Setup with controller

```sh
$ juju info wordpress
name: wordpress
charm-id: PdZggeV24UW98kGssQHDkppCMuVp6bEU
summary: Wordpress, uses official Docker Wordpress image by default
publisher: mruffel
supports: kubernetes
tags: applications, blog
subordinate: false
store-url: https://staging-api.charmhub.io/wordpress
description: Wordpress, uses official Docker Wordpress image by default
relations:
  provides:
    website: http
  requires: {}
channels: |
  latest/stable:     9  2020-11-16  (9)  194kB
  latest/candidate:  ↑
  latest/beta:       ↑
  latest/edge:       ↑
```

#### Download from non-closed channels

```sh
$ juju download wordpress --channel="latest/stable"
Fetching charm "wordpress" using "stable" channel at revision 9
Install the "wordpress" charm with:
    juju deploy ./wordpress.charm
```

#### Download from closed channels

```sh
$ juju download wordpress --channel="latest/edge"
Fetching charm "wordpress" using "edge" channel at revision 9
Install the "wordpress" charm with:
    juju deploy ./wordpress.charm
```

#### Download by arch

```sh
$ juju download wordpress --channel="latest/edge" --arch="amd64"
Fetching charm "wordpress" using "edge" channel at revision 9
Install the "wordpress" charm with:
    juju deploy ./wordpress.charm
```

#### Download by series

```sh
$ juju download wordpress --channel="latest/edge" --series="kubernetes"
Fetching charm "wordpress" using "edge" channel at revision 9
Install the "wordpress" charm with:
    juju deploy ./wordpress.charm
```

#### Download by arch and series

```sh
$ juju download wordpress --channel="latest/edge" --arch=amd64 --series="kubernetes"
Fetching charm "wordpress" using "edge" channel at revision 9
Install the "wordpress" charm with:
    juju deploy ./wordpress.charm
```

#### Download by default series

This one is a bit different, in that if you pass empty string for the series
it will use the default series (which won't work with wordpress as it's k8s
only).

```sh
$ juju download wordpress --channel="latest/edge" --series=""
ERROR charm "wordpress" not found for "latest/edge" channel matching "focal" series
```

#### Error with invalid arch

```sh
$ juju download wordpress --channel="latest/edge" --arch="risc" --series="focal"
ERROR unexpected architecture flag value "risc", expected <all|amd64|arm64|armhf|i386|ppc64el|s390x>
```

#### Error with invalid series

```sh
juju download wordpress --channel="latest/edge" --arch="amd64" --series="focal"
ERROR charm "wordpress" not found for "latest/edge" channel matching "focal" series
```

### Setup without controller

You can run through all the above scenarios without a controller using
a flag.

```sh
$ juju download wordpress --charm-hub-url=https://api.staging.snapcraft.io --channel=edge
Fetching charm "wordpress" using "edge" channel at revision 9
Install the "wordpress" charm with:
    juju deploy ./wordpress.charm
```


cmd/juju/charmhub/download.go
```diff
@@ -55,27 +55,32 @@ See also:
 
 // NewDownloadCommand wraps downloadCommand with sane model settings.
 func NewDownloadCommand() cmd.Command {
-	return modelcmd.Wrap(&downloadCommand{},
-		modelcmd.WrapSkipModelInit,
-	)
+	return modelcmd.Wrap(&downloadCommand{
+		charmHubCommand: newCharmHubCommand(),
+		fallbackArch:    utilsarch.HostArch(),
```
achilleasa, line 60: I think using the host arch as a fallback might be problematic here. What if I am running the cli on an arm64 machine and my controller only supports amd64?

cmd/juju/charmhub/download.go
```diff
@@ -210,17 +222,29 @@ func (c *downloadCommand) Run(cmdContext *cmd.Context) error {
...
-	// If there is no series, then attempt to select the best one first.
-	if c.series == "" {
-		revision, found = c.locateRevisionByChannel(info.ChannelMap, charmChannel)
-	} else {
-		// We have a series go attempt to find the channel.
-		revision, found = c.locateRevisionByChannelAndSeries(info.ChannelMap, charmChannel, c.series)
+	var filterFn FilterInfoChannelMapFunc
+	switch {
+	case c.arch == ArchAll && c.series == SeriesAll:
+		filterFn = filterByArchitecture(c.fallbackArch)
```
achilleasa, line 228: See my previous comment. I think that having the host as the fallback arch is wrong here. Could we add a check for the number of results and error with a message like "found multiple arch matches for X (arch1, arch2, arch3); please explicitly specify which arch to download via the --arch flag"

cmd/juju/charmhub/download.go
SimonRichardson: Specify an arch via `--arch=amd64`

achilleasa: Perhaps then the `--arch` should be mandatory then?

cmd/juju/charmhub/download.go
```diff
@@ -244,14 +247,25 @@ func (c *downloadCommand) Run(cmdContext *cmd.Context) error {
...
 		if c.series != "" {
 			return errors.Errorf("%s %q not found for %q channel matching %q series", info.Type, c.charmOrBundle, c.channel, c.series)
 		}
 		return errors.Errorf("%s %q not found with in the channel %q", info.Type, c.charmOrBundle, c.channel)
 	}
 
+	if selectAll {
+		// Ensure we have just one architecture.
+		if !hasMatchingArch(revisions) {
+			return errors.Errorf("ambiguous architectures found, use --arch flag to specify architecture")
```
achilleasa, line 261: Can you include the list of found arches in the error?

achilleasa: Also perhaps rephrase the error as `multiple architectures (%s) found, use --arch flag to specify the one to download`

achilleasa (approved):
Nice!

SimonRichardson:
$$merge$$

PR #12352, opened 2020-11-23
Restrict setting arch constraint

SimonRichardson:
The aim here is to restrict an application architecture constraint once
it's been set. If the architecture is already set and it's the one being
passed in, we treat it as a no-op and move along. Removing the
application removes the constraint so can be re-applied the same as
other constraints.

Most of the time you won't be able to change an architecture unless
you have a cloud that reports back as being able to swap (manual).
In that case the error you'll generally hit is:

```sh
juju set-constraints mysql arch=amd64
ERROR invalid constraint value: arch=amd64
valid values are: [arm64]
```

The driver behind this is to prevent the deploying of a charm with a
constraint and changing the architecture under the charm with a
different constraint. This will cause a world of pain to try and
reconcile that correctly, re-downloading charms and their resources
along with any profiles that need applying... So the fix is to prevent
it and then if they have done the wrong thing, remove the application
and start again.

Architecture is becoming more pervasive so we need to ensure that we
don't allow foot guns in the codebase.

## QA steps

Bootstrap to a hybrid architecture cloud (manual, aws)

### Apply a constraint

This states that you can't change a constraint after the fact for
an application with an arch that we don't know about.

```sh
$ juju deploy redis
$ juju set-constraints redis arch=arm64
ERROR: changing architecture not supported
```

### Apply the same constraint

This should be fine as we're applying the same constraint again.

```sh
$ juju deploy redis --constraints arch=amd64
$ juju set-constraints redis arch=amd64
```

### Removing an application 

This should allow us to specify another architecture

```sh
$ juju deploy redis --constraints arch=arm64
$ juju remove-application redis
$ juju deploy redis --constraints arch=amd64
```

## Documentation changes

Attempting to update an application that has an existing architecture 
constraint with a new architecture fails if they're not the same.

### DIFF
--- a/state/application.go
+++ b/state/application.go
@@ -2837,13 +2837,33 @@ func (a *Application) SetConstraints(cons constraints.Value) (err error) {
 	} else if err != nil {
 		return err
 	}
+
 	if a.doc.Subordinate {
 		return ErrSubordinateConstraints
 	}
+
+	// If the architecture has already been set, do not allow the application
+	// architecture to change.
+	// Note: we don't apply this check in validate constraints as we have no
+	// way to know if it's a model or an application (former we do allow
+	// changes).
+	current, err := a.Constraints()
+	if !errors.IsNotFound(err) {
+		if err != nil {
+			return errors.Annotate(err, "unable to read constraints")
+		}
+		if current.Arch == nil && cons.Arch != nil {
+			return errors.NotSupportedf("changing architecture")
+		} else if current.Arch != nil && cons.Arch != nil && *current.Arch != *cons.Arch {
+			return errors.NotSupportedf("changing architecture (%s)", *current.Arch)
+		}
+	}
+
 	defer errors.DeferredAnnotatef(&err, "cannot set constraints")
 	if a.doc.Life != Alive {
 		return applicationNotAliveErr
 	}
+
 	ops := []txn.Op{{
 		C:      applicationsC,
 		Id:     a.doc.DocID,
--- a/state/application.go
+++ b/state/application.go
@@ -27,6 +27,7 @@ import (
 	"gopkg.in/mgo.v2/txn"
 
 	"github.com/juju/juju/core/application"
+	"github.com/juju/juju/core/arch"
 	"github.com/juju/juju/core/constraints"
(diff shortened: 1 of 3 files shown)


state/application.go
```diff
@@ -2837,13 +2837,33 @@ func (a *Application) SetConstraints(cons constraints.Value) (err error) {
...
+	current, err := a.Constraints()
+	if !errors.IsNotFound(err) {
+		if err != nil {
+			return errors.Annotate(err, "unable to read constraints")
+		}
+		if current.Arch == nil && cons.Arch != nil {
+			return errors.NotSupportedf("changing architecture")
+		} else if current.Arch != nil && cons.Arch != nil && *current.Arch != *cons.Arch {
+			return errors.NotSupportedf("changing architecture (%s)", *current.Arch)
+		}
```
achilleasa, line 2859: Extracting the common check might make the logic a bit easier to read

```suggestion
                if cons.Arch != nil {
		    if current.Arch == nil {
			return errors.NotSupportedf("changing architecture")
		    } else if *current.Arch != *cons.Arch {
			return errors.NotSupportedf("changing architecture (%s)", *current.Arch)
		    }
	      }
```

achilleasa (commented):
Running the QA steps.

state/application.go
```diff
@@ -2848,14 +2848,14 @@ func (a *Application) SetConstraints(cons constraints.Value) (err error) {
 	// Note: we don't apply this check in validate constraints as we have no
 	// way to know if it's a model or an application (former we do allow
 	// changes).
-	current, err := a.Constraints()
-	if !errors.IsNotFound(err) && cons.Arch != nil {
+	if current, err := a.Constraints(); !errors.IsNotFound(err) && (cons.Arch != nil && *cons.Arch != "") {
```
achilleasa, line 2851: `err` is shadowing the outer err here so the defer on L2863 will not work as expected if `Contstraints` returns NotFound

SimonRichardson: Good spot, I'll rename the errrrrrr

achilleasa (approved):
LGTM

SimonRichardson:
$$merge$$

PR #12359, opened 2020-11-24
Fixes an issue with CAData when importing k8s

tlm:
When we import a cluster with CAData file in the config we fill out the
k8s config struct with the raw CAData as well. This causes an error to
propogate from the k8s client-go code. This change stops that from
happening by using tmp variables to pass the data around.

## Checklist

## QA steps

Follow repo steps in bug report

## Bug reference

https://bugs.launchpad.net/juju/+bug/1905321

### DIFF
--- a/caas/kubernetes/clientconfig/k8s.go
+++ b/caas/kubernetes/clientconfig/k8s.go
@@ -132,21 +132,18 @@ func contextsFromConfig(config *clientcmdapi.Config) (map[string]Context, error)
 }
 
 func cloudsFromConfig(config *clientcmdapi.Config, cloudName string) (map[string]CloudConfig, error) {
-
 	clusterToCloud := func(cluster *clientcmdapi.Cluster) (CloudConfig, error) {
 		attrs := map[string]interface{}{}
 
-		// TODO(axw) if the CA cert is specified by path, then we
-		// should just store the path in the cloud definition, and
-		// rely on cloud finalization to read it at time of use.
-		if cluster.CertificateAuthority != "" {
+		k8sCAData := cluster.CertificateAuthorityData
+		if len(cluster.CertificateAuthorityData) == 0 && cluster.CertificateAuthority != "" {
 			caData, err := ioutil.ReadFile(cluster.CertificateAuthority)
 			if err != nil {
 				return CloudConfig{}, errors.Trace(err)
 			}
-			cluster.CertificateAuthorityData = caData
+			k8sCAData = caData
 		}
-		attrs["CAData"] = string(cluster.CertificateAuthorityData)
+		attrs["CAData"] = string(k8sCAData)
 
 		return CloudConfig{
 			Endpoint:   cluster.Server,
@@ -295,7 +292,6 @@ func readKubeConfigFile() (reader io.Reader, err error) {
 }
 
 func parseKubeConfig(data []byte) (*clientcmdapi.Config, error) {
-
 	config, err := clientcmd.Load(data)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


caas/kubernetes/clientconfig/k8s.go
```diff
@@ -132,21 +132,18 @@ func contextsFromConfig(config *clientcmdapi.Config) (map[string]Context, error)
...
 
 func cloudsFromConfig(config *clientcmdapi.Config, cloudName string) (map[string]CloudConfig, error) {
-
 	clusterToCloud := func(cluster *clientcmdapi.Cluster) (CloudConfig, error) {
 		attrs := map[string]interface{}{}
 
-		// TODO(axw) if the CA cert is specified by path, then we
-		// should just store the path in the cloud definition, and
-		// rely on cloud finalization to read it at time of use.
+		k8sCAData := cluster.CertificateAuthorityData
```
jameinel, line 138: Is this something we can put into a unit test that having CertificateAuthority is supported but doesn't modify CertificateAuthorityData?

Given the other error message, doesn't CertificateAuthorityData supercede CertificateAuthority? This code looks like it overwrites Data if CertificateAuthority is set.

The error that we got was
```
error: certificate-authority-data and certificate-authority are both specified for kubernetes. certificate-authority-data will override.
```

Which would hint that -data is primary, except it is an error so it just refuses.
(Note this also happens if you just kubectl with both available according to:
https://stackoverflow.com/questions/52302684/eks-error-certificate-authority-data-and-certificate-authority-are-both-specif
)

caas/kubernetes/clientconfig/k8s.go
```diff
@@ -132,21 +132,18 @@ func contextsFromConfig(config *clientcmdapi.Config) (map[string]Context, error)
...
 		if cluster.CertificateAuthority != "" {
 			caData, err := ioutil.ReadFile(cluster.CertificateAuthority)
 			if err != nil {
 				return CloudConfig{}, errors.Trace(err)
 			}
-			cluster.CertificateAuthorityData = caData
+			k8sCAData = caData
```
jameinel, line 144: Trying to dig through what certificate-authority-data is supposed to be, I came across this:
https://insujang.github.io/2019-12-18/kubernetes-authentication/

Which seems to say that -data is supposed to be base64 encoding of .../ca.crt? This seems to assume that it is the direct data.

This post:
https://codefarm.me/2019/02/01/access-kubernetes-api-with-client-certificates/

also says to generate the content:
```
# Generate the <CA-DATA>
sudo cat /etc/kubernetes/pki/ca.crt | base64 | tr -d '\n'
```
So you do take the raw .crt and base64 encode it again, and then remove newlines.

That would seem that we are incorrectly handling the content.

jameinel (commented):
I think the changes are good. I'm concerned about how we've been handling certificate-authority-data, as we seem to unilaterally override it with certificate-authority content, and it seems -data was never supposed to be the raw content anyway.

caas/kubernetes/clientconfig/k8s.go
tlm: 100% correct

caas/kubernetes/clientconfig/k8s.go
tlm: I've made some changes to cover this.

caas/kubernetes/clientconfig/k8s.go
tlm: The kubernetes client unpacks the base64 for us. So we haven't been storing bad data luckily.

wallyworld (approved):

wallyworld:
$$merge$$

PR #12361, opened 2020-11-24
Fix TestClientWatchAllAdminPermission intermittent failure

benhoyt:
Failure was:

```
client_test.go:1042:
    c.Assert(dApp, gc.NotNil)
... value *params.RemoteApplicationUpdate = (*params.RemoteApplicationUpdate)(nil)
```

And was happening sometimes (I got it to happen locally every 20-50
runs) because while normally there are 3 watcher changes that come
through (model, machine, remoteApplication), sometimes there are more,
for example: model, machine 1, machine2, remoteApplication, machine3. In
the latter case, it sends 3 "machine" deltas that kind of settle over
time: in the second one the HardwareCharacteristics field changed to
non-nil, in the third the InstanceId changed to a non-empty string.

The number and exact contents of changes that come through varies
depending on timing, so instead of waiting for exactly 3 changes, wait
for them to settle to the right value (up to a 10 second timeout).

Also do the same fix to TestClientWatchAllReadPermission, which I don't
think has been failing (at least not as often), but almost certainly has
the same issue.

## QA steps

Run the TestClientWatchAllAdminPermission 100 times until it fails (I could repro this locally every 30 or so runs).

### DIFF
--- a/apiserver/facades/client/client/client_test.go
+++ b/apiserver/facades/client/client/client_test.go
@@ -910,42 +910,65 @@ func (s *clientSuite) TestClientWatchAllReadPermission(c *gc.C) {
 		err := watcher.Stop()
 		c.Assert(err, jc.ErrorIsNil)
 	}()
-	deltas, err := watcher.Next()
-	c.Assert(err, jc.ErrorIsNil)
-	// Model and machine deltas returned.
-	c.Assert(len(deltas), gc.Equals, 2)
-	var d0 *params.MachineInfo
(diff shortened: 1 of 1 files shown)


wallyworld (approved):

benhoyt:
$$merge$$

benhoyt:
[clientSuite.TestClientWatchAllAdminPermission](https://trello.com/c/duRgOPWi/156-clientsuitetestclientwatchalladminpermission)

PR #12362, opened 2020-11-25
When upgrading to juju 3, models model be at least 2.9

wallyworld:
Do not allow models or controllers older than 2.9 to be upgraded to juju 3.
Same for migration - add precheck to prevent old model being migrated to a 3.x controller.

## QA steps

bootstrap a 2.8 controller
attempt to upgrade-controller to agent version 3.0.0 -> error

bootstrap a 2.8 controller
add a model
upgrade controller to 2.9
attempt to upgrade-controller to agent version 3.0.0 -> error

bootstrap a 2.8 controller
add a model
bootstrap a 3.0 controller
attempt to migrate model to new controller -> error

### DIFF
--- a/upgrades/model.go
+++ b/upgrades/model.go
@@ -1,0 +1,32 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package upgrades
+
+import (
+	"github.com/juju/errors"
+	"github.com/juju/version"
+)
+
+// MinMajorUpgradeVersion defines the minimum version all models
+// must be running before a major version upgrade.
+var MinMajorUpgradeVersion = map[int]version.Number{
+	3: version.MustParse("2.9.0"),
+}
+
+// UpgradeAllowed returns true if a major version upgrade is allowed
+// for the specified from and to versions.
+func UpgradeAllowed(from, to version.Number) (bool, version.Number, error) {
+	if from.Major == to.Major {
+		return true, version.Number{}, nil
+	}
+	// Downgrades not allowed.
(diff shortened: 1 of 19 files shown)


upgrades/model.go
```diff
@@ -0,0 +1,28 @@
...
+// minMajorUpgradeVersion defines the minimum version all models
+// must be running before a major version upgrade.
+var minMajorUpgradeVersion = map[int]version.Number{
+	3: version.MustParse("2.9.0"),
+}
+
+// UpgradeAllowed returns true if a major version upgrade is allowed
+// for the specified from and to versions.
+func UpgradeAllowed(from, to version.Number) (bool, version.Number, error) {
+	if from.Major >= to.Major {
```
hpidcock, line 20: this should just be if they are equal

hpidcock: unless we support downgrades?

hpidcock (approved):

upgrades/model.go
wallyworld: We do now support downgrade; it's been for minor versions eg 2.9 -> 2.8
But thinking about it, allowing a controller to be downgraded would be bad, so I'll fix, thanks.

wallyworld:
$$merge$$

wallyworld:
[do not allow upgrade to 3 unless all models are 2.9+](https://trello.com/c/2PjdKsdZ/1294-do-not-allow-upgrade-to-3-unless-all-models-are-29)

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #12363, opened 2020-11-25
Charmhub check charm placement

SimonRichardson:
The referesher client can now check for a charm placement using the new
API client.

We only do this in the charmhub side of things as that improves our
single responsibility side of things. CharmStore charms don't need to
check for placement. Charmstore charms follow the analogy of:

    a charm is a charm is a charm

This is no longer a case as architecture and series dictate that a charm
is a matrix of architecture, series and a charm. Charm is a binary blob.

The nice thing here is that we created a SOLID type that makes it easier
to decide on what gets called depending on what type is used.

### Homogenous workload check

Ensure that we can switch to a different charm from charmstore to
charmhub.

```sh
$ juju deploy cs:mysql
$ juju refresh mysql --switch ubuntu 
```

### Heterogeneous workload check

```sh
$ juju bootstrap aws test --config charm-hub-url="https://api.staging.snapcraft.io"
$ juju add-machine --constraints="arch=arm64" --series=xenial
$ juju add-machine --constraints="arch=amd64" --series=xenial
$ juju deploy cs:mysql --to 1
$ juju add-unit mysql --to 2
$ juju refresh mysql --switch ubuntu
Looking up metadata for charm-store charm "mysql" from channel stable
ERROR charm can not be placed in a heterogeneous environment
```

## Documentation changes

Charmhub charms require homogenous units and so we need to prevent
a charm landing in a heterogeneous workload.


api/charms/client.go
```diff
@@ -142,3 +142,23 @@ func (c *Client) AddCharmWithAuthorization(curl *charm.URL, origin apicharm.Orig
...
+	args := params.ApplicationCharmPlacements{
+		Placements: []params.ApplicationCharmPlacement{{
+			Application: applicationName,
+			CharmURL:    curl.String(),
+		}},
+	}
+	var result params.ErrorResults
+	if err := c.facade.FacadeCall("CheckCharmPlacement", args, &result); err != nil {
+		if errors.IsNotSupported(err) {
+			return nil
```
achilleasa, line 159: I think you should surface the error as an `errors.Unsupported` value and have the consumer handle it. The client shouldn't be doing any black magic and suppress error values.

apiserver/facades/client/charms/client.go
```diff
@@ -33,8 +35,8 @@ import (
 type API struct {
 	*charmscommon.CharmsAPI
 	authorizer   facade.Authorizer
-	backendState BackendState
-	backendModel BackendModel
+	backendState interfaces.BackendState
```
achilleasa, line 38: I know there is no need to do this but it would be cleaner to alias the import to something like `charminterfaces` so we know what is the purpose of the interface.

apiserver/facades/client/charms/client.go
```diff
@@ -531,3 +550,147 @@ func (a *API) IsMetered(args params.CharmURL) (params.IsMeteredResult, error) {
...
+	}
+
+	// The placement logic below only cares about charmhub charms. Once we have
+	// multiple architecture support for charmhub, we can remove the placement
+	// check.
+	if !charm.CharmHub.Matches(curl.Schema) {
+		return params.ErrorResult{}, nil
+	}
+
+	// Get the application if it's not found, just return without an error as
```
achilleasa, line 593:
```suggestion
	// Get the application. If it's not found, just return without an error as
```

apiserver/facades/client/charms/client.go
```diff
@@ -531,3 +550,147 @@ func (a *API) IsMetered(args params.CharmURL) (params.IsMeteredResult, error) {
...
+			Error: apiservererrors.ServerError(err),
+		}, nil
+	}
+
+	// We don't care for subordinates here.
+	if !app.IsPrincipal() {
+		return params.ErrorResult{}, nil
+	}
+
+	constraints, err := app.Constraints()
```
achilleasa, line 609: Do you get a blank constraints.Value here if the constraints are not found or a nil value? If it's the latter, if you get ErrNotFound you will skip the next if block and panic on L619

apiserver/facades/client/charms/client.go
```diff
@@ -531,3 +550,147 @@ func (a *API) IsMetered(args params.CharmURL) (params.IsMeteredResult, error) {
...
+
+	// If the application has an existing architecture constraint then we're
+	// happy that the constraint logic will prevent heterogenous application
+	// units.
+	if constraints.Arch != nil && *constraints.Arch != "" {
+		return params.ErrorResult{}, nil
+	}
+
+	// Unfortunately we now have to check instance data for all units to
+	// validate that we have homogeneous setup.
```
achilleasa, line 624:
```suggestion
	// validate that we have a homogeneous setup.
```

apiserver/facades/client/charms/client.go
```diff
@@ -531,3 +550,147 @@ func (a *API) IsMetered(args params.CharmURL) (params.IsMeteredResult, error) {
...
+
+		if machineArch == "" {
+			arches.Add(arch.DefaultArchitecture)
+		} else {
+			arches.Add(machineArch)
+		}
+	}
+
+	if arches.Size() > 1 {
+		// It is expected that charmhub charms are in a heterogeneous workload,
```
achilleasa, line 667:
```suggestion
		// It is expected that charmhub charms form a homogenous workload,
```

apiserver/facades/client/charms/client.go
```diff
@@ -531,3 +550,147 @@ func (a *API) IsMetered(args params.CharmURL) (params.IsMeteredResult, error) {
...
+		err := errors.Errorf("charm can not be placed in a heterogeneous environment")
+		return params.ErrorResult{
+			Error: apiservererrors.ServerError(err),
+		}, nil
+	}
+
+	return params.ErrorResult{}, nil
+}
+
+func (a *API) getMachineArch(machine interfaces.Machine) (string, error) {
```
achilleasa, line 678: Don't we have a typed value for arches?

cmd/juju/application/refresher/refresher.go
```diff
@@ -383,9 +384,23 @@ func (r *charmHubRefresher) Allowed(cfg RefresherConfig) (bool, error) {
...
+		}
+
+		if !cfg.Switch {
 			return true, nil
 		}
+
+		if err := r.charmAdder.CheckCharmPlacement(cfg.ApplicationName, curl); err != nil {
+			// If force is used then ignore the error, the user seems to know
+			// what they're doing.
+			if !cfg.Force {
```
achilleasa, line 398: Should we emit a warning here? Forcing such a placement will probably break the deployment so we should at least warn that we proceed even though the charm cannot be legally placed.

apiserver/facades/client/charms/client.go
SimonRichardson: constraints.Value is a struct so you would get an empty struct, essentially everything is nil inside, so you have to check HasArch everywhere.

achilleasa: Ah cool then. Thought it could be a pointer.

apiserver/facades/client/charms/client.go
```diff
@@ -664,7 +664,7 @@ func (a *API) checkCharmPlacement(arg params.ApplicationCharmPlacement) (params.
 	}
 
 	if arches.Size() > 1 {
-		// It is expected that charmhub charms are in a heterogeneous workload,
+		// It is expected that charmhub charms form a heterogeneous workload,
```
achilleasa, line 667: The comment before was about the __heterogenous__ bit. Shouldn't the expectation be that this must be a **homogeneous** workload?

cmd/juju/application/refresher/refresher.go
```diff
@@ -392,12 +392,13 @@ func (r *charmHubRefresher) Allowed(cfg RefresherConfig) (bool, error) {
...
 		}
 
-		if err := r.charmAdder.CheckCharmPlacement(cfg.ApplicationName, curl); err != nil {
+		if err := r.charmAdder.CheckCharmPlacement(cfg.ApplicationName, curl); err != nil && !errors.IsNotSupported(err) {
 			// If force is used then ignore the error, the user seems to know
 			// what they're doing.
 			if !cfg.Force {
 				return false, errors.Trace(err)
 			}
+			r.logger.Warningf("Charm placement check failed, using force may break deployment")
```
achilleasa, line 401: or "using --force"?

achilleasa (approved):
LGTM

SimonRichardson:
$$merge$$

PR #12370, opened 2020-11-26
[dev] Remove k8s specific bits from caas broker interface

achilleasa:
This PR attempts to decouple some of the k8s-specific aspects of the `caas.Broker` interface and is part of the work for introducing support for ECS. 

The k8s-specific interfaces have been moved into the kubernetes provider and cast checks are used to verify that the injected broker instances complies with the interfaces (e.g. ClusterMetadataChecker) expected by the provider implementation.

In addition, a new interface called `CredentialChecker` has been introduced to the `caas.Broker` interface. The various facades call its `CheckCloudCredentials` method to verify that the cloud credentials still work. The original implementation was k8s-specific (list namespaces) and has been moved into the provider.

## QA steps

```console
$ juju bootstrap microk8s k8s --no-gui
$ juju add-model test

# Deploy a legacy k8s charm
$ juju deploy cs:~charmed-osm/prometheus-k8s-32

# Deploy a charm that uses a sidecar container
$ git clone https://github.com/hpidcock/testing-charms
$ juju deploy ./testing-charms/cockroachdb --resource cockroachdb-image=cockroachdb/cockroach
# The latter should make it up to the install hook and error
```

### DIFF
--- a/caas/broker.go
+++ b/caas/broker.go
@@ -191,9 +191,6 @@ type Broker interface {
 	// ClusterVersionGetter provides methods to get cluster version information.
 	ClusterVersionGetter
 
-	// ClusterMetadataGetter provides an API for querying the cluster metadata.
-	ClusterMetadataGetter
-
 	// CredentialChecker provides an API for checking that the credentials
 	// used by the broker are functioning.
 	CredentialChecker
@@ -293,12 +290,6 @@ type StorageValidator interface {
 type ClusterVersionGetter interface {
 	// Version returns cluster version information.
 	Version() (*version.Number, error)
-}
-
-// ClusterMetadataGetter provides an API to query cluster metadata.
-type ClusterMetadataGetter interface {
-	// GetClusterMetadata returns metadata about host cloud and storage for the cluster.
-	GetClusterMetadata(storageClass string) (result *ClusterMetadata, err error)
 }
 
 // CredentialChecker provides an API for checking that the credentials
--- a/caas/broker.go
+++ b/caas/broker.go
@@ -9,7 +9,6 @@ import (
 	"github.com/juju/errors"
 	"github.com/juju/names/v4"
 	"github.com/juju/version"
-	core "k8s.io/api/core/v1"
 
 	"github.com/juju/juju/caas/specs"
 	"github.com/juju/juju/core/application"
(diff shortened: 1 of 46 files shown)


caas/kubernetes/provider/provider.go
```diff
@@ -28,11 +28,24 @@ import (
 	"github.com/juju/juju/environs/context"
 )
 
+// ClusterMetadataChecker provides an API to query cluster metadata.
```
hpidcock, line 31: All the interfaces moved from caas into caas/kubernetes/provider should probably be in caas/kubernetes instead.

cmd/juju/caas/add.go
```diff
@@ -54,8 +54,21 @@ type AddCloudAPI interface {
 	Close() error
 }
 
+// ClusterMetadataChecker provides an API to query cluster metadata.
+type ClusterMetadataChecker interface {
```
hpidcock, line 58: Yeah lets just put this interface into the caas/kubernetes package, that way it doesn't need to be re-declared here.

worker/caasmodeloperator/manifold.go
```diff
@@ -30,6 +30,7 @@ type ManifoldConfig struct {
 	// Logger to use in this worker
 	Logger Logger
 	// ModelUUID is the id of the model this worker is operating on
+
```
hpidcock, line 33: Remove line

caas/broker.go
```diff
@@ -170,31 +169,56 @@ type Broker interface {
...
+	environs.InstancePrechecker
+
+	// BootstrapEnviron defines methods for bootstrapping a controller.
+	environs.BootstrapEnviron
+
+	// ResourceAdopter defines methods for adopting resources.
+	environs.ResourceAdopter
+
+	// StorageValidator provides methods to validate storage.
+	StorageValidator
```
ycliuhw, line 183: the method `ValidateStorageClass` in `StorageValidator` is k8s spec as well

caas/broker.go
```diff
@@ -278,8 +295,22 @@ type ClusterVersionGetter interface {
 	Version() (*version.Number, error)
 }
 
-// ServiceGetterSetter provides the API to get/set service.
-type ServiceGetterSetter interface {
+// ClusterMetadataGetter provides an API to query cluster metadata.
+type ClusterMetadataGetter interface {
+	// GetClusterMetadata returns metadata about host cloud and storage for the cluster.
+	GetClusterMetadata(storageClass string) (result *ClusterMetadata, err error)
```
ycliuhw, line 301: `ClusterMetadata` is k8s spec thing as well.
Most of the code in `caas/metadata.go` could be moved to `caas/kubernetes`

caas/kubernetes/provider/provider.go
achilleasa: No problem. I will move those and also try to relocate some k8s constants (IIRC they are defined in the caas package) there.

caas/broker.go
achilleasa: Yes, I wanted to do that initially but there are lots of parts in the code that expect this to be there. I will make a second attempt today.

caas/broker.go
achilleasa: Can't we repurpose this for validating storage for caas in general? It accepts a map so we should be able to make it work with ecs, right?

achilleasa:
!!build!! (windows tests failed)

SimonRichardson:
@hpidcock @ycliuhw can you take another look, @achilleasa wants to baby sit this one in before he gets back.

caas/broker.go
ycliuhw: I mean the `ValidateStorageClass` method name is purely for k8s.
But that's ok we can change it later if we need.

ycliuhw (approved):
lgtm thanks

PR #12375, opened 2020-11-27
lxd: Implement disk and architecture constraints

stgraber:


stgraber:
I've been able to test the disk part as working properly (minus the unit issue I reported in https://bugs.launchpad.net/juju/+bug/1905827) for the architecture part, I only see amd64 as reportedly supported here, not sure if it's because of the provider code not knowing about supported arches for my cluster or if it's because I'm using local binaries and don't have an i386 or arm64 build around.

manadart:
@stgraber Thanks for this. Any objection to me cherry-picking the commit into a patch targeting the 2.9 branch?
We've just forked it off from develop, and this would not be released until Juju 3.0 otherwise.

stgraber:
Nope, that'd be great, thanks!

stgraber:
@manadart updated my branch to split things into better chunks and I've also fixed that LP bug I reported yesterday.
Basically Juju pinning an older (buggy) version of the LXD client is what was causing the version check to fail.

I've now refreshed it to current and cleaned up the one thing that broke as a result of it. Local tests now show proper limits in MiB.

stgraber:
Further updates to fix mocking after the lxd package update and also added some changes to make the arch constraint actually be useful. It now pulls all the supported arches from the cluster and ensures that you're using one of them.

Test deployment here now fails on not being able to find suitable agent binaries, which looks good.

manadart:
[Apply PR 12375 to 2.9 branch](https://trello.com/c/jPpZfdiJ/3185-apply-pr-12375-to-29-branch)

manadart:
Landed these commits with https://github.com/juju/juju/pull/12383.

PR #12381, opened 2020-11-30
Parse charm platform

SimonRichardson:
The following introduces the parsing of platform so that importing and
exporting works as expected for the description package. Originally the
description package expected `os/series/arch`, but in reality it should
be `arch/os/series`, with arch being the non-optional value making it
consistent between the lack of optional values.

The rules for platform are the following:

 1. Architecture is mandatory.
 2. OS is optional and can be dropped. Series is mandatory if OS wants
 to be displayed.
 3. Series is also optional.

To indicate something is missing `unknown` can be used in place. Using
Normalize can clean up the output of the Platform output to ensure
consistency.

Examples of this would be:

 1. `<arch>/<os>/<series>`
 2. `<arch>`
 3. `<arch>/<series>`
 4. `<arch>/unknown/<series>`

The code is rather simple as it just splits over a string and plucks the
right parts based on the amount visible.

## QA steps

Test pass, this is being used in #12376 and was a piece that could be broken off
as a separate isolated chunk of work.

### DIFF
--- a/core/charm/channel.go
+++ b/core/charm/channel.go
@@ -137,7 +137,7 @@ func ParseChannel(s string) (Channel, error) {
 func ParseChannelNormalize(s string) (Channel, error) {
 	ch, err := ParseChannel(s)
 	if err != nil {
-		return Channel{}, err
+		return Channel{}, errors.Trace(err)
 	}
 	return ch.Normalize(), nil
 }
--- a/core/charm/origin.go
+++ b/core/charm/origin.go
@@ -3,6 +3,13 @@
 
 package charm
 
+import (
+	"fmt"
+	"strings"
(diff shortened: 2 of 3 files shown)


manadart (approved):
Looks good. One question: What is the particular reason for OS being nullified if we have no series?

SimonRichardson:
$$merge$$

SimonRichardson:
> Looks good. One question: What is the particular reason for OS being nullified if we have no series?

We currently have no way what the best series is. We should probably introduce `latest`, but this is out of scope for this piece of work. Tracking a bug for it though -> https://bugs.launchpad.net/juju/+bug/1906252

PR #12384, opened 2020-11-30
Add ResourcesClient for charmhub

hmlanigan:
Includes the new client, response structures and ListResourceRevisons.
A new api endpoint for charmhub from: http://api.snapcraft.io/docs/charms.html#list_resource_revisions

There may be changes up coming to have the same property names, avoid hash-sha256 vs hash-sha256.

Will be used by `juju resources` in the future.

A few drive-by debug message fixes.

## QA steps

No change to current charmhub client behavior.

### DIFF
--- a/charmhub/info.go
+++ b/charmhub/info.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/juju/errors"
+	"github.com/kr/pretty"
 
 	"github.com/juju/juju/charmhub/path"
 	"github.com/juju/juju/charmhub/transport"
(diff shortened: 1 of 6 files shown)


hmlanigan:
This is PR #12311 reappearing.

charmhub/info.go
```diff
@@ -56,7 +57,7 @@ func (c *InfoClient) Info(ctx context.Context, name string) (transport.InfoRespo
 		}
 		return resp, resultErr
 	}
-
+	c.logger.Tracef("Info() unmarshalled: %s", pretty.Sprint(resp))
```
benhoyt, line 60: Should we wrap in `if c.logger.IsTraceEnabled() { ... }` to avoid pretty-printed a large object to a string in the normal case (when we're not tracing)?

benhoyt (approved):

hmlanigan:
$$merge$$

charmhub/info.go
hmlanigan: Had a discussion with some folks.  In general is not needed in the case.  Though there can be other cases were it is.

PR #12385, opened 2020-12-01
Consolidate exec actions

wallyworld:
Make juju exec and juju run use the same APIs and CLI UX.
This means that juju exec will now have the new, improved actions behaviour.
The exec and actions implementation both used actions at the backed, but totally different CLI code, and also slightly different api results when enqueuing tasks.

There's a lot of code moving around which blows out the diff size. Plus github is showing the move of exec.go and exec_test.go as a add / delete instead of a move.

The key parts in separate commits:
- remove the state code to support action UUIDs
- the api facade changes to make enqueuing juju exec actions behave the same as enqueuing charm actions
- moving exec from commands to actions package and factoring out common code for results handing and tests

Also, the actions tests were using a real clock, which was problematic to start with, but more so when trying to get the tests right. The real clock in tests is replaced with a test clock, which resulted in the test helpers needing to run the commands in a separate goroutine so the test clock cloud be advanced.

## QA steps

bootstrap and deploy a charm with actions
check things like
```
juju exec --machine 0 ps
juju exec --machine 0 ps --wait 1s
juju run ubuntu-lite/0 hello
juju run ubuntu-lite/0 hello --wait 1s
```
Note that actions now will print the receivers which did not completed when a timeout occurs (as a result of consolidating the run and exec code).


cmd/juju/action/exec.go
```diff
@@ -0,0 +1,242 @@
+// Copyright 2013 Canonical Ltd.
```
wallyworld, line 1: moved from cmd/juju/commands

cmd/juju/action/exec_test.go
```diff
@@ -0,0 +1,656 @@
+// Copyright 2013 Canonical Ltd.
```
wallyworld, line 1: moved from cmd/juju/commands

wallyworld:
[unitfy exec and run CLI and backend](https://trello.com/c/X5zZhkRO/1297-unitfy-exec-and-run-cli-and-backend)

cmd/juju/action/showoperation_test.go
```diff
@@ -303,14 +313,34 @@ timing:
...
+			fakeClient := s.makeFakeClient(
 				t.withAPIDelay,
 				t.withAPITimeout,
 				t.withAPIResponse,
 				t.withAPIError,
 			)
-			testRunOperationHelper(
-				c, s,
+
+			numExpectedimers := 0
```
benhoyt, line 324: Typo? `numExpectedTimers`?

benhoyt (approved):
LGTM. One nit comment. Also a question: why does the facade version not have to be incremented due to the facade return value changes?

benhoyt:
Also looks to be failing the tests on Windows: https://jenkins.juju.canonical.com/job/github-unit-tests-win2012/6621/testReport/junit/github/com_juju_juju_cmd_juju_action/TestPackage/

cmd/juju/action/showoperation_test.go
wallyworld: doh, ty

cmd/juju/action/exec_test.go
wallyworld: > LGTM. One nit comment. Also a question: why does the facade version not have to be incremented due to the facade return value changes?

This is for juju 3 and we are disallowing version 2 clients from using this api. You will need a v 3 client.

wallyworld:
$$merge$$

apiserver/params/actions.go
```diff
@@ -51,7 +51,7 @@ type Action struct {
 // EnqueuedActions represents the result of enqueuing actions to run.
 type EnqueuedActions struct {
 	OperationTag string         `json:"operation"`
-	Actions      []StringResult `json:"actions,omitempty"`
+	Actions      []ActionResult `json:"actions,omitempty"`
```
manadart, line 54: Shouldn't we still bump the facade version, even if we end up removing older versions for a 3.0 release?

PR #12393, opened 2020-12-03
Fix time.After usage causing memory leaks. 

hpidcock:
time.After internals are not cleaned up until the timer fires.
In certain usage of time.After where long timeouts + fast firing channels can cause a large number of timers to be created.

## QA steps

Run unit tests.

## Documentation changes

N/A

## Bug reference

N/A


wallyworld (approved):

hpidcock:
$$merge$$

PR #12396, opened 2020-12-03
Not found charm error message

SimonRichardson:
The not found charm error message was terrible, it offered no support
for users with a cryptic error about it not conforming to a URL.

The following change fixes that by checking that the file doesn't exist
and if it doesn't warn them about that, rather than a malformed url that
nobody cares about.

## QA steps

```sh
$ juju deploy ./bad.charm
ERROR no charm was found at "./bad.charm"
```


petevg (commented):
Hooray for better error messages! This looks good to me.

hmlanigan (approved):
LGTM, qa went well.  Nice to fix a completely misleading error message.

SimonRichardson:
$$merge$$

PR #12404, opened 2020-12-04
Update the upgrade step for 2.9 for charm origin

SimonRichardson:
Now that more fields have been added to charm-origin, the upgrade steps
should also be added to better reflect that.

There is a worrying part that I'm unsure how to solve and that is around
people who are already trying out RC1 and RC2 of 2.9. Officially this
won't re-run for them, but is that a bad thing?

 - The items that actually are mandatory are source, revision, channel
 and id. These were already in version 1 of the upgrade step.
 - Platform although important will fallback to the logic to find out
 what it should be if it's missing.
 - Type isn't required for charms and is a requirement for bundles,
 considering we don't store bundles this should be fine.

If what I say holds we should be ok.

## QA steps

Bootstrap a 2.8.6 and upgrade to 2.9
Deploy a charm...

### DIFF
--- a/state/constraints.go
+++ b/state/constraints.go
@@ -175,7 +175,7 @@ func readConstraints(mb modelBackend, id string) (constraints.Value, error) {
 	constraintsCollection, closer := mb.db().GetCollection(constraintsC)
 	defer closer()
 
-	doc := constraintsDoc{}
+	var doc constraintsDoc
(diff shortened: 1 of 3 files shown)


manadart (approved):
QA and code all good.

SimonRichardson:
$$merge$$

PR #12406, opened 2020-12-07
Feature/ecs provider

ycliuhw:
*Wireframe the ECS provider;*
Implemented features:

- Ensure() - deploy the task definition and service;
- Delete() - remove the whole stack when the application gets deleted;

## Checklist

 - [ ] ~Requires a [pylibjuju](https://github.com/juju/python-libjuju) change~
 - [ ] ~Added [integration tests](https://github.com/juju/juju/tree/develop/tests) for the PR~
 - [ ] ~Added or updated [doc.go](https://discourse.jujucharms.com/t/readme-in-packages/451) related to packages changed~

## QA steps

No QA steps, because the ECS integration is not complete yet.

## Documentation changes

No

## Bug reference

No


agent/agentbootstrap/bootstrap.go
```diff
@@ -388,7 +388,7 @@ func getEnviron(
 		Cloud:          cloudSpec,
 		Config:         modelConfig,
 	}
-	if cloudSpec.Type == cloud.CloudTypeCAAS {
```
achilleasa, line 391: Could we make this a method on `Type` instead?

caas/ecs/application.go
```diff
@@ -0,0 +1,637 @@
...
+	return fmt.Sprintf("%s-%s", a.name, storageName)
+}
+
+// getMountPathForFilesystem returns mount path.
+func getMountPathForFilesystem(idx int, appName string, fs jujustorage.KubernetesFilesystemParams) string {
+	if fs.Attachment != nil {
+		return fs.Attachment.Path
+	}
+	return fmt.Sprintf(
+		"%s/fs/%s/%s/%d",
```
achilleasa, line 98: Are we using host-based volumes here? If so, how are we going to deal with the possibility that two task instances of the same app get scheduled on the same node?

caas/ecs/application.go
```diff
@@ -0,0 +1,637 @@
...
+					{
+						Name:  aws.String("JUJU_K8S_APPLICATION_PASSWORD"),
+						Value: aws.String(config.IntroductionSecret),
+					},
+					{
+						Name:  aws.String("JUJU_K8S_CONTROLLER_ADDRESSES"),
+						Value: aws.String(config.ControllerAddresses),
+					},
+					{
+						Name:  aws.String("JUJU_K8S_CONTROLLER_CA_CERT"),
```
achilleasa, line 209: So does this actually work as-is when you use the API? Does the SDK transparently encode the KV pairs when calling to the ECS API? I had issues when using the cli with a json document.

caas/ecs/application.go
```diff
@@ -0,0 +1,637 @@
...
+			{
+				ContainerPath: aws.String("/charm/bin"),
+				SourceVolume:  aws.String("charm-data-bin"),
+			},
+		},
+	}
+
+	for _, v := range containers {
+		// TODO: https://aws.amazon.com/blogs/compute/amazon-ecs-and-docker-volume-drivers-amazon-ebs/
+		// to use EBS volumes, it requires some docker storage plugin installed in the
```
achilleasa, line 301: Are we going for EBS instead of instance storage? Is this so we can always re-attach storage (similar to k8s) when the tasks get rescheduled to other containers?

caas/ecs/application.go
```diff
@@ -0,0 +1,640 @@
...
+
+		ebsCfg, err := newEbsConfig(fs.Attributes)
+		if err != nil {
+			// This should never happen because it's been validated `storageProvider.ValidateConfig`.
+			return nil, nil, errors.NotValidf("storage attribute for %q", fs.StorageName)
+		}
+		vol := &ecs.Volume{
+			Name: aws.String(a.volumeName(fs.StorageName)),
+			DockerVolumeConfiguration: &ecs.DockerVolumeConfiguration{
+				Scope:         aws.String("shared"),
```
ycliuhw, line 135: These volumes are configured to `shared` for all the tasks for now.
We will need to test how to set here to `task` scope, so each task has its own volume.

caas/ecs/application.go
ycliuhw: We don't use any host-based volumes.
For juju related volumes, we use `DockerVolumeConfiguration` with local driver 

- https://docs.aws.amazon.com/AmazonECS/latest/developerguide/docker-volumes.html.
- https://github.com/juju/juju/pull/12406/files#diff-2ccee37d2965d4d7d708a406d87294f7be801db49dde62eec02686efbbee5709R257

For user-requested volumes, we use `EBS` 
- https://github.com/juju/juju/pull/12406/files#diff-2ccee37d2965d4d7d708a406d87294f7be801db49dde62eec02686efbbee5709R132

caas/ecs/application.go
ycliuhw: It works as expected.

caas/ecs/application.go
ycliuhw: Yes, We choose this as the default driver for EBS support.
- https://github.com/rexray/rexray

caas/ecs/application.go
ycliuhw: We don't use any host-based volumes.
For juju related volumes, we use `DockerVolumeConfiguration` with local driver 

- https://docs.aws.amazon.com/AmazonECS/latest/developerguide/docker-volumes.html.
- https://github.com/juju/juju/pull/12406/files#diff-2ccee37d2965d4d7d708a406d87294f7be801db49dde62eec02686efbbee5709R257

For user-requested volumes, we use `EBS` 
- https://github.com/juju/juju/pull/12406/files#diff-2ccee37d2965d4d7d708a406d87294f7be801db49dde62eec02686efbbee5709R132

caas/ecs/application.go
ycliuhw: It works as expected.

achilleasa: Nice!

caas/ecs/application.go
achilleasa: According to the docs: ```The scope for the Docker volume, which determines its lifecycle. Docker volumes that are scoped to a task are automatically provisioned when the task starts destroyed when the task is cleaned up. Docker volumes that are scoped as shared persist after the task stops.```. 

Does this mean that the storage gets reclaimed when a task gets moved to another node or clean up only applies when the task is _permanently_ destroyed (e.g. we scale down)?

ycliuhw: The ECS tasks are never restarted but always recreated if any `essential` containers died.
So it seems there is no point to keep the volume after the task died because there will be no chance to be re-attached again later at some point.
So we will need to test what happens it sets to the `task` scope.

agent/agentbootstrap/bootstrap.go
ycliuhw: It is a good idea.
I tried to move this to cloud.Type. But I found I have to touch hundreds of files (from `cmd` to `api/apiserver` then `state`).
Given this PR is already not small. It might be better to do this change in a separate PR.

ycliuhw:
[Add a new cloud type for ecs(might need tweak add-cloud cmd);](https://trello.com/c/QVXwRTsQ/1308-add-a-new-cloud-type-for-ecsmight-need-tweak-add-cloud-cmd)

ycliuhw:
[Implement ECS API integration under `caas/ecs` package](https://trello.com/c/XfUkGxAk/1309-implement-ecs-api-integration-under-caas-ecs-package)

worker/caasapplicationprovisioner/application.go
```diff
@@ -196,16 +196,16 @@ func (a *appWorker) loop() error {
 			// Respond to state changes
 			err = handleChange()
 			if err != nil {
-				return nil
+				return errors.Trace(err)
```
achilleasa, line 199: The prior behavior (mute errors) before your change looks quite dangerous... Should the changes in this method be backported to 2.9?

achilleasa (approved):
Overall it LGTM. Left some comments (there is a missing insert that needs to be fixed). I will go ahead and approve it so you can land it after you resolve the broken tests and apply any of my comments that make sense.

I will iterate on the provider once this patch lands and start populating the TODOs and try to figure out a way to grab the task ID via the ec2 metadata route as we discussed.

caas/ecs/constants/constants.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package constants
```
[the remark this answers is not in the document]
ycliuhw: To avoid importing `ecs` package when we need those constants from outside (workers for example).
If we can ensure no need to import these `StorageProviderType` etc constants later, then we can move those constants to `ecs` package.

worker/caasapplicationprovisioner/application.go
ycliuhw: yeah, I will backport these fixes to 2.9

caas/ecs/application.go
```diff
@@ -0,0 +1,706 @@
...
+			FamilyPrefix: aws.String(a.resourceName()),
+		},
+	)
+	err = a.handleErr(err)
+	if errors.IsNotFound(err) {
+		logger.Tracef("listing task definitions for family %q in cluster %q, err: %v", a.resourceName(), a.clusterName, err)
+		return nil
+	}
+	if err != nil {
+		errors.Trace(err)
```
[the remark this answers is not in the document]
ycliuhw: ops, fixing that!

caas/ecs/application.go
```diff
@@ -0,0 +1,706 @@
...
+			since = aws.TimeValue(t.StoppingAt)
+		}
+		return "", status.Terminated, since
+	}
+	jujuStatus = status.Unknown
+	healthStatus := aws.StringValue(t.HealthStatus)
+	switch healthStatus {
+	case "UNKNOWN":
+	case "RUNNING":
+		jujuStatus = status.Running
```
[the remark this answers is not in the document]
ycliuhw: Luckily, AWS provides all those utils like `aws.TimeValue`

caas/ecs/application.go
```diff
@@ -0,0 +1,706 @@
...
+	case "UNHEALTHY":
+		jujuStatus = status.Error
```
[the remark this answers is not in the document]
ycliuhw: yeah, we can set it to `ExecutionStoppedAt` for now.
We can change when we are able to get the real status later.

ycliuhw:
$$merge$$

caas/ecs/package_test.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package ecs_test
```
manadart, line 4: Is there an import loop that prevents us putting this in the _ecs_ package and using exported suite methods instead of _export_test.go_

achilleasa: We will moving stuff around in the following PRs so we can try changing this in one of those.

PR #12408, opened 2020-12-07
Charmhub find summary wordwrap

SimonRichardson:
This is a very simple word wrapping of a line. We know that the summary
has been chopped to one line via the one line function, meaning we can
just look for spaces and hyphens to workout when to wordwrap a line.
With these constraints/restrictions is makes is easy for us to clean up
the find table.

This won't handle pathological cases like long words, but it'll do for
now. It's much easier on the eye than the previous dumpster fire that
find was!

## QA steps

```sh
$ juju bootstrap lxd test
$ juju find
```

### DIFF
--- a/cmd/juju/charmhub/findwriter.go
+++ b/cmd/juju/charmhub/findwriter.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"unicode"
 
 	"github.com/juju/errors"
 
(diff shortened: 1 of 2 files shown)


manadart (approved):
The change works, but the tabular output is huge in width. It isn't really usable in shorter width terminals.

SimonRichardson:
> The change works, but the tabular output is huge in width. It isn't really usable in shorter width terminals.

I'll make architecture and supports optional columns, otherwise like you said it would be huge...

SimonRichardson:
$$merge$$

PR #12409, opened 2020-12-07
Columnise find to reduce output width

SimonRichardson:
The output find is rather large, now that we have more and more
populated fields, we're potentially showing to much information. Instead
of just removing the fields that look unimportant, we should tailor it
depending on the users needs.

This PR introduces columns for the tabular field. Using --columns flag
with the column alias seems like a decent approach. Other canonical
tools also do this, so it isn't something new (LXD).

With decent defaults it actually looks rather good as well.

### Default column output 

```sh
$ juju bootstrap lxd test
$ juju find kubernetes
```

### Tailoured

```sh
$ juju bootstrap lxd test
$ juju find --columns=nbvaS
```

### DIFF
--- a/cmd/juju/charmhub/find.go
+++ b/cmd/juju/charmhub/find.go
@@ -51,7 +51,8 @@ type findCommand struct {
 	out        cmd.Output
 	warningLog Log
 
-	query string
+	query   string
+	columns string
 }
 
 // Find returns help related info about the command, it implements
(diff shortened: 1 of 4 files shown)


manadart (approved):
Nice.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #12410, opened 2020-12-07
Charmhub download progress

SimonRichardson:
The following is lifted from snap[1] leaving the copyleft GPLv3 license
in-tact. It deviates from the snap version to implementing testing
without resorting to using export_test hacks.

Once we have this we can then start reporting download progress for juju
download to help people see the download timeframe.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju download ubuntu
```


apiserver/facades/client/charms/repositories.go
```diff
@@ -31,7 +31,7 @@ var logger = loggo.GetLogger("juju.apiserver.charms")
 // CharmHubClient represents the methods required of a
 // client to install or upgrade a CharmHub charm.
 type CharmHubClient interface {
-	DownloadAndRead(ctx context.Context, resourceURL *url.URL, archivePath string) (*charm.CharmArchive, error)
+	DownloadAndRead(ctx context.Context, resourceURL *url.URL, archivePath string, pb charmhub.ProgressBar) (*charm.CharmArchive, error)
```
achilleasa, line 34: Probably not worth it but I wonder whether we could convert the last two params into functional args (WithArchivePath, WithProgressBar). This might simplify the various DownloadXXX methods below.

cmd/output/progress/progress.go
```diff
@@ -0,0 +1,115 @@
...
+// * if no terminal is attached, or we think we're running a test, a
+//   minimalistic QuietMeter is returned.
+// * otherwise, an ANSIMeter is returned.
+//
+func MakeProgressBar(stdout io.Writer) Meter {
+	if terminal.IsTerminal(int(os.Stdin.Fd())) {
+		return &ANSIMeter{
+			terminal:    term{},
+			stdout:      stdout,
+			escapeChars: DefaultEscapeChars(),
```
achilleasa, line 63: Are we sure these are valid under windows? They seem fairly trivial escape sequences but I don't have access to a windows instance to test.

apiserver/facades/client/charms/repositories.go
SimonRichardson: We always need the archive path, we just don't need the progress bar, so I've made that optional.

achilleasa (approved):
LGTM and QA works as expected. Caveat: haven't tested on windows.

SimonRichardson:
$$merge$$

cmd/output/progress/progress.go
benhoyt: Looks like they're not -- or at least something in here is causing these tests to fail on Windows. For example: https://jenkins.juju.canonical.com/job/github-unit-tests-win2012/6726/testReport/github/com_juju_juju_cmd_output_progress/Test/

PR #12412, opened 2020-12-07
[dev] Ensure that assess_upgrade_series test can run against develop branch

achilleasa:
The develop branch deprecates --machine for juju run and breaks this
test. As a workaround, this commit changes the test to use juju ssh when
issuing reboot requests.

In addition, a 5 second delay is injected between the second (post
series upgrade) reboot and the check for the series reported by the
agent. This delay should be enough to allow the agent to report the
series and invalidate the cached status. I theorize that this is why
we see intermittent failures for this test against 2.8/2.9

## QA steps

The test should pass: 

```sh
make install
cd acceptancetests
./assess_upgrade_series.py lxd `which juju` ./tmp/art nw-upgradeseries-amd64-lxd
```

### DIFF
--- a/acceptancetests/assess_upgrade_series.py
+++ b/acceptancetests/assess_upgrade_series.py
@@ -8,6 +8,7 @@ import logging
 import os
 import subprocess
 import sys
+import time
 
 from deploy_stack import (
     BootstrapManager,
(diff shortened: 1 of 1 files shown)


hmlanigan (approved):
LGTM, qa went well.

achilleasa:
!!build!!

achilleasa:
$$merge$$

PR #12400, opened 2020-12-04
Make UpdateLatestRevisions controller facade work with charmhub

benhoyt:
Currently it only pulls new charm revision info from the charmstore;
make it pull from the new charmhub API as well (using Refresh). We
partition the list of the model's charms into charmstore vs charmhub
so we can call each API in bulk. Most of the new charmhub-specific
code is in the new charmhub.go file.

Change RefreshEntity.CreatedAt to a time.Time instead of a string. May
want to do this for other *At fields in there as well?

For testing, I've added a wrench charmrevision/shortinterval that
is checked when the controller starts up. If you create
/var/lib/juju/wrench/charmrevision on the controller with contents
"shortinterval" and then restart the controller it'll set the poll
interval to 10 seconds (!) instead of 24 hours.

Other things:
* Added a debug logger to the charmrevision worker so you can see when
  it fires.
* A few drive-by typo fixes.

## QA steps

```
# create the wrench and restart controller
juju ssh -m controller 0
sudo mkdir /var/lib/juju/wrench
echo shortinterval | sudo tee /var/lib/juju/wrench/charmrevision
sudo systemctl restart jujud-machine-0
```

```sh
export JUJU_DEV_FEATURE_FLAGS=charm-hub
juju add-model test --config charm-hub-url="https://api.staging.snapcraft.io"
juju deploy postgresql
juju deploy cs:mysql  # try a charmstore charm as well
juju debug-log -m controller --include-module juju.apiserver.charmrevisionupdater --include-module juju.worker.charmrevision
```

PR #12415, opened 2020-12-08
Add Juju-Metadata HTTP headers for charmhub Refresh requests

benhoyt:
We currently add various `Juju-Metadata` HTTP headers when sending charmstore requests, but not for the new charmhub API. This adds those headers to charmhub Refresh calls as well. The model metadata is added in charmhub Client config. And the charm metadata (series and arch) is added per-charm by Refresh() itself.

## QA steps

Hard to specifically look at the headers going out, but I've added unit tests to check them. For general QA steps for the charm revision updater, see QA steps at https://github.com/juju/juju/pull/12400

### DIFF
--- a/charmhub/client.go
+++ b/charmhub/client.go
@@ -39,6 +39,8 @@ const (
 	CharmHubServerVersion = "v2"
 	CharmHubServerEntity  = "charms"
 
+	MetadataHeader = "Juju-Metadata"
+
 	RefreshTimeout = 10 * time.Second
 )
 
--- a/charmhub/client.go
+++ b/charmhub/client.go
@@ -39,7 +39,7 @@ const (
 	CharmHubServerVersion = "v2"
 	CharmHubServerEntity  = "charms"
 
-	MetadataHeader = "Juju-Metadata"
+	MetadataHeader = "X-Juju-Metadata"
 
 	RefreshTimeout = 10 * time.Second
 )
--- a/charmhub/refresh.go
+++ b/charmhub/refresh.go
@@ -84,8 +84,31 @@ func (c *RefreshClient) Refresh(ctx context.Context, config RefreshConfig) ([]tr
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
+
+	// Add Juju-Metadata headers for the charms' unique series and
+	// architecture values, for example:
+	//
+	// Juju-Metadata: series=bionic
+	// Juju-Metadata: arch=amd64
+	// Juju-Metadata: series=focal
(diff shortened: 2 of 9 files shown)


cmd/jujud/agent/model/manifolds.go
```diff
@@ -269,7 +269,6 @@ func commonManifolds(config ManifoldsConfig) dependency.Manifolds {
 			NewFacade: charmrevision.NewAPIFacade,
 			NewWorker: charmrevision.NewWorker,
 			Logger:    config.LoggingContext.GetLogger("juju.worker.charmrevision"),
-			// No Logger defined in the charmrevision package.
```
benhoyt, line 272: Should have removed this in my previous commit, as I added the logger.

benhoyt:
[Send Juju-Metadata headers when using charmhub Refresh API](https://trello.com/c/htkxHsiW/1312-send-juju-metadata-headers-when-using-charmhub-refresh-api)

charmhub/refresh.go
```diff
@@ -84,8 +84,31 @@ func (c *RefreshClient) Refresh(ctx context.Context, config RefreshConfig) ([]tr
...
+
+	// Add Juju-Metadata headers for the charms' unique series and
+	// architecture values, for example:
+	//
+	// Juju-Metadata: series=bionic
+	// Juju-Metadata: arch=amd64
+	// Juju-Metadata: series=focal
+	headers := make(http.Header)
+	seriesAdded := make(map[string]bool)
+	archsAdded := make(map[string]bool)
```
wallyworld, line 96: We have a juju/collections/set that could be used here

charmhub/client.go
```diff
@@ -39,6 +39,8 @@ const (
 	CharmHubServerVersion = "v2"
 	CharmHubServerEntity  = "charms"
 
+	MetadataHeader = "Juju-Metadata"
```
wallyworld, line 42: We should originally have made this X-Juju-Metadata
Ideally we'd use the correct one for CH moving forward. Something to consider.

wallyworld (approved):
Looks good. Let's consider fixing the header name

charmhub/refresh.go
benhoyt: Used.

charmhub/client.go
benhoyt: Updated. Per discussion, I'll also ask Simon and Heather what the charmhub team wants.

benhoyt:
$$merge$$

PR #12420, opened 2020-12-08
Remove state from common client

SimonRichardson:
The common client constructor was bringing in state. We REALLY do NOT
want to include state in the apiserver. It's a slow march forward and
adding more state to the apiserver is counter productive.

This cleans this up and moves the metadata headers to charmhub directly
as we get this for free for others whom are using the client.

## QA steps

Tests pass, this is a mechanical change.

### DIFF
--- a/apiserver/common/charmhub.go
+++ b/apiserver/common/charmhub.go
@@ -4,19 +4,26 @@
 package common
 
 import (
-	"sort"
-
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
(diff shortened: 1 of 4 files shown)


hmlanigan (approved):

hmlanigan:
$$merge$$

benhoyt:
Simon, thanks for this. I was going to remove State from charmrevisionupdater and make it more testable today, but this is a head start. :-)

PR #12421, opened 2020-12-08
Charmhub download resource meta

SimonRichardson:
The following ensures that we get the metadata that we want. This is
abstracted out into a client so that both charmhub and the charmstore
can get the information in the same logic, without duplication.

## QA steps

TBA

```sh
QA steps here
```

### DIFF
--- a/apiserver/facades/client/resources/repository.go
+++ b/apiserver/facades/client/resources/repository.go
@@ -43,11 +43,95 @@ type CharmHub interface {
 	Refresh(ctx context.Context, config charmhub.RefreshConfig) ([]transport.RefreshResponse, error)
 }
 
+// ResourceClient requests the resource info for a given charm URL,
+// charm Origin, resource name and resource revision.
+type ResourceClient interface {
+	ResourceInfo(url *charm.URL, origin corecharm.Origin, name string, revision int) (charmresource.Resource, error)
+}
+
+type resourceClient struct {
+	client ResourceClient
+}
+
+// resolveResources determines the resource info that should actually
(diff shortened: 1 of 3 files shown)


apiserver/facades/client/resources/repository.go
```diff
@@ -43,43 +43,154 @@ type CharmHub interface {
...
+	if err != nil {
+		return nil, err
+	}
+	// TODO(ericsnow) Ensure that the non-upload resource revisions
+	// match a previously published revision set?
+	return resolved, nil
+}
+
+func (ch *charmHubClient) ResourceInfo(_ *charm.URL, origin corecharm.Origin, name string, revision int) (charmresource.Resource, error) {
+	charmOrigin := ch.id.Origin
```
SimonRichardson, line 159: I'm putting money on we could bulk this, but for now call it one by one.

apiserver/facades/client/resources/repository.go
```diff
@@ -112,14 +223,22 @@ func resourceFromRevision(rev transport.ResourceRevision) (charmresource.Resourc
...
 	}
+	fingerprint, err := charmresource.ParseFingerprint(rev.Download.HashSHA384)
+	if err != nil {
+		return charmresource.Resource{}, errors.Trace(err)
+	}
 	r := charmresource.Resource{
 		Meta: charmresource.Meta{
 			Name: rev.Name,
 			Type: resType,
+			// TODO (stickupkid): API is missing both of these fields.
```
hmlanigan, line 234: transport ResourceRevision now has the data, landed with 12422, however the data will not be filled in yet.

PR #12414, opened 2020-12-08
Charmhub resources part 2 - resource_get

hmlanigan:
Create a resource opener for charmhub.  This will allow resources to be downloaded from charmhub for use.

There are too many levels of indirection already in juju/resource.  Unfortunately I've made it a tiny bit worse here.  Near future work will cut this down to an easier to understand piece.

## QA steps

```console 
export JUJU_DEV_FEATURE_FLAGS="charm-hub"
juju bootstrap localhost
juju add-model six --config charm-hub-url="https://api.staging.snapcraft.io"

# Deploy a CharmHub charm with resources, set the region to trigger resource_get in the charm.
juju deploy postgresql --config aws_region="us-east-1"
# wait for the charm to install and try to get the resource successfully
# there will be a status message about installing the Wal-e snap
$ juju show-status-log postgresql/0
Time                   Type       Status       Message
....
08 Dec 2020 22:27:55Z  juju-unit  executing    running install hook
08 Dec 2020 22:28:52Z  workload   maintenance  Updating apt cache
08 Dec 2020 22:28:54Z  workload   maintenance  Installing wal-e snap
```

PR #12422, opened 2020-12-08
Add path and description to ResourceRevision

hmlanigan:
Add Path and Description to the transport ResourceRevision struct.  They will be used to create resource Meta for resources from charmhub.

Today they are not part of the charmhub api responses, so adding them in charmhub/resources.go would break charmhub deploy.  Left as a todo once the api knows about the object elements.

## QA steps

Test to ensure current work doesn't break.  However the resources facade piece isn't testable until #12414 lands.

```console
$ juju bootstrap localhost
$ juju add-model six --config charm-hub-url="https://api.staging.snapcraft.io"
$ juju deploy ubuntu
```


benhoyt (approved):

hmlanigan:
$$merge$$

PR #12424, opened 2020-12-09
charmrevisionupdater: get rid of *State and JujuConnSuite

benhoyt:
In the charmrevisionupdater facade, use interfaces instead of `*state.State` for testability; avoid `JujuConnSuite` and use gomock. Note that I've used gomock everywhere except the charmstore and charmhub mocks, where it was quite klunky -- I started porting those to gomock too, but a simple fake worked out better for this case.

For QA steps, see: https://github.com/juju/juju/pull/12400


apiserver/facades/controller/charmrevisionupdater/state.go
```diff
@@ -0,0 +1,76 @@
+// Copyright 2020 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
```
hmlanigan, line 2: To be consistent with juju precendent, this file should be called interface.go.  state brings to mind juju/state.

apiserver/facades/controller/charmrevisionupdater/charmhub_test.go
```diff
@@ -1,75 +1,56 @@
...
-		Tag:        names.NewMachineTag("99"),
-	}
-	var err error
-	s.updater, err = charmrevisionupdater.NewCharmRevisionUpdaterAPI(s.State, resources, authorizer)
-	c.Assert(err, jc.ErrorIsNil)
+	s.CleanupSuite.SetUpTest(c)
 
 	// Patch the charmhub initializer function
-	s.PatchValue(&charmrevisionupdater.NewCharmhubClient, func(st *state.State, metadata map[string]string) (charmrevisionupdater.CharmhubRefreshClient, error) {
+	s.PatchValue(&charmrevisionupdater.NewCharmhubClient, func(st charmrevisionupdater.State, metadata map[string]string) (charmrevisionupdater.CharmhubRefreshClient, error) {
```
hmlanigan, line 37: This test file does not need to live in the _test package.  Thus you can avoid using PatchValue.  Or find another way to avoid it.

apiserver/facades/controller/charmrevisionupdater/state_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/juju/names/v4"
+
+	"github.com/juju/juju/apiserver/facades/controller/charmrevisionupdater"
+	"github.com/juju/juju/cloud"
+	"github.com/juju/juju/controller"
+	"github.com/juju/juju/environs/config"
+	"github.com/juju/juju/state"
+)
+
+type mockState struct {
```
hmlanigan, line 22: Please use go mock for these, it is the preferred juju unit test pattern.  Overtime time this style of mocking becomes limited in what it can do.  We'd only use this style for something like a logger.  As well, we cannot verify inputs easily with it, nor have tests with different results for easy error scenarios.

state/charm.go
```diff
@@ -28,7 +28,13 @@ import (
 
 // MacaroonCache is a type that wraps State and implements charmstore.MacaroonCache.
 type MacaroonCache struct {
-	*State
+	Charmer
```
hmlanigan, line 31: Once #12414 lands, you'll need to rebase, as it's been done there.

apiserver/facades/controller/charmrevisionupdater/state.go
benhoyt: Renamed to interface.go. (Though I see there are quite a number of `state.go`'s around the place too. It is kinda intended to bring to mind juju/state, as it's an interface of a subset of `*state.State`).

state/charm.go
benhoyt: Updated.

apiserver/facades/controller/charmrevisionupdater/charmhub_test.go
benhoyt: Fixed. Made a new function for testing which takes these dependencies as arguments.

apiserver/facades/controller/charmrevisionupdater/testing/suite.go
```diff
@@ -76,7 +74,7 @@ type CharmSuite struct {
 	jcSuite *jujutesting.JujuConnSuite
 
 	charms map[string]*state.Charm
-	store  *mockStore
+	Store  *mockStore
```
benhoyt, line 77: I don't like the fact that `charmrevisionupdater/testing` still exists, but it's used by `apiserver/facades/client/client/status.go:TestUpdateRevisions`. I'll take a look later to see if that test can be updated (or even just removed).

state/charm.go
```diff
@@ -26,9 +26,13 @@ import (
 	jujuversion "github.com/juju/juju/version"
 )
 
+type MacaroonCacheState interface {
```
benhoyt, line 29: Taken from https://github.com/juju/juju/pull/12414/files#diff-1fd66e956344843e302c15e83ba915ac01e8a6285e3e5d7a01d8ee9f701ba51b

apiserver/facades/controller/charmrevisionupdater/testing/suite.go
```diff
@@ -164,10 +157,20 @@ func (s *CharmSuite) AddCharmWithRevision(c *gc.C, charmName string, rev int) *s
 }
 
 // AddService adds a service for the specified charm to state.
-func (s *CharmSuite) AddService(c *gc.C, charmName, serviceName string) {
+func (s *CharmSuite) AddApplication(c *gc.C, charmName, applicationName string) {
```
wallyworld, line 160: Change comment also

apiserver/facades/controller/charmrevisionupdater/updater.go
```diff
@@ -34,24 +33,50 @@ type CharmRevisionUpdater interface {
...
-	authorizer facade.Authorizer
+	state State
+
+	newCharmstoreClient NewCharmstoreClientFunc
+	newCharmhubClient   NewCharmhubClientFunc
 }
 
+type NewCharmstoreClientFunc func(st State) (charmstore.Client, error)
+type NewCharmhubClientFunc func(st State, metadata map[string]string) (CharmhubRefreshClient, error)
+
```
wallyworld, line 44: These don't need to be public

apiserver/facades/controller/charmrevisionupdater/updater_test.go
```diff
@@ -4,136 +4,126 @@
...
+	state.EXPECT().AllApplications().Return([]charmrevisionupdater.Application{
+		makeApplication(ctrl, "ch", "varnish", "charm-5", "app-1", 1),
+		makeApplication(ctrl, "cs", "varnish", "charm-6", "app-2", 2),
+	}, nil).AnyTimes()
+
+	updater, err := charmrevisionupdater.NewCharmRevisionUpdaterAPIState(state, newFakeCharmstoreClient, newFakeCharmhubClient)
 	c.Assert(err, jc.ErrorIsNil)
 
-	result, err = s.charmrevisionupdater.UpdateLatestRevisions()
+	result, err := updater.UpdateLatestRevisions()
```
wallyworld, line 92: We need to check something other than result.Error is nil I would have thought?
How can we check that UpdateLatestRevisions() was called. I see we expect a call to AllApplications(). Is there something else we can do in addition?
eg we set up the fake charm store to have wordpress and mysql charms with revisions. I can't see where we use those revisions in any checks

apiserver/facades/controller/charmrevisionupdater/updater_test.go
```diff
@@ -4,136 +4,126 @@
...
 
-	// No pending charm for wordpress.
-	curl = charm.MustParseURL("cs:quantal/wordpress")
-	_, err = s.State.LatestPlaceholderCharm(curl)
-	c.Assert(err, jc.Satisfies, errors.IsNotFound)
+	state.EXPECT().AddCharmPlaceholder(charm.MustParseURL("cs:mysql-23")).Return(nil)
+
+	result, err = updater.UpdateLatestRevisions()
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(result.Error, gc.IsNil)
```
wallyworld, line 128: Same here - the old code could check for no pending charms. What can we check for now?

apiserver/facades/controller/charmrevisionupdater/updater_test.go
```diff
@@ -4,136 +4,126 @@
...
+	updater, err := charmrevisionupdater.NewCharmRevisionUpdaterAPIState(state, nil, newFakeCharmhubClient)
 	c.Assert(err, jc.ErrorIsNil)
-}
 
-func (s *charmVersionSuite) TestNewCharmRevisionUpdaterAPIAcceptsStateManager(c *gc.C) {
-	endPoint, err := charmrevisionupdater.NewCharmRevisionUpdaterAPI(s.State, s.resources, s.authoriser)
+	result, err := updater.UpdateLatestRevisions()
 	c.Assert(err, jc.ErrorIsNil)
-	c.Assert(endPoint, gc.NotNil)
+	c.Assert(result.Error, gc.IsNil)
```
wallyworld, line 55: See other comments

wallyworld (commented):
I could be missing it, but it seems we need to add some checks to the tests - see comments

wallyworld (approved):

apiserver/facades/controller/charmrevisionupdater/updater.go
benhoyt: Done.

apiserver/facades/controller/charmrevisionupdater/testing/suite.go
benhoyt: Done.

apiserver/facades/controller/charmrevisionupdater/updater_test.go
benhoyt: Per voice discussion, this is actually correct as is. The result type only includes an Error field, so that's all there is to check for that. However, we know the code paths are executed because of the gomock `state.EXPECT().AddCharmPlaceholder()` calls -- charms that are going to be updated are added as placeholders to state. If that's not called (or called too many times or with the wrong args), gomock will fail the test.

benhoyt: As far as revisions go, that's the `-23` part of an assertion like the following:

```
state.EXPECT().AddCharmPlaceholder(charm.MustParseURL("cs:mysql-23")).Return(nil)
```

apiserver/facades/controller/charmrevisionupdater/updater_test.go
benhoyt: As above

apiserver/facades/controller/charmrevisionupdater/updater_test.go
benhoyt: As above

benhoyt:
!!build!!

benhoyt:
$$merge$$

benhoyt:
$$merge$$

apiserver/facades/controller/charmrevisionupdater/updater.go
```diff
@@ -34,24 +33,50 @@ type CharmRevisionUpdater interface {
...
-	st *state.State,
-	resources facade.Resources,
-	authorizer facade.Authorizer,
-) (*CharmRevisionUpdaterAPI, error) {
-	if !authorizer.AuthController() {
+func NewCharmRevisionUpdaterAPI(ctx facade.Context) (*CharmRevisionUpdaterAPI, error) {
+	if !ctx.Auth().AuthController() {
 		return nil, apiservererrors.ErrPerm
 	}
+	api := &CharmRevisionUpdaterAPI{
```
hmlanigan, line 52: Call NewCharmRevisionUpdaterAPIState here, rather than return the struct.

apiserver/facades/controller/charmrevisionupdater/updater_test.go
```diff
@@ -4,136 +4,230 @@
...
+		},
+	}, nil)
+
+	state := makeState(c, ctrl, nil)
+	state.EXPECT().AllApplications().Return([]charmrevisionupdater.Application{
+		makeApplication(ctrl, "ch", "postgresql", "charm-2", "app-2", 42),
+	}, nil).AnyTimes()
+	state.EXPECT().AddCharmPlaceholder(charm.MustParseURL("ch:postgresql-42")).Return(nil)
+
+	newClient := func(st charmrevisionupdater.State, metadata map[string]string) (charmrevisionupdater.CharmhubRefreshClient, error) {
```
hmlanigan, line 166: If you add client to the updaterSuite, you can do:

func (s *updaterSuite) newClient(_ charmrevisionupdater.State, _ map[string]string) (charmrevisionupdater.CharmhubRefreshClient, error) {
	return s.chClient, nil
}

and 

updater, err := charmrevisionupdater.NewCharmRevisionUpdaterAPIState(state, nil, s.newClient)

reduce code duplication.  Much of the juju go mock use, adds the mocks to the suite, to facilitate ease of use.

apiserver/facades/controller/charmrevisionupdater/interface_test.go
```diff
@@ -110,94 +105,46 @@ func makeResource(c *gc.C, name string, revision, size int, hexFingerprint strin
...
 }
 
-func (c *fakeCharmhubClient) Refresh(_ context.Context, config charmhub.RefreshConfig) ([]transport.RefreshResponse, error) {
-	// Sanity check that metadata headers are present
-	if c.metadata["model_uuid"] == "" {
-		return nil, errors.Errorf("model metadata not present")
+func (m charmhubConfigMatcher) Matches(x interface{}) bool {
+	config, ok := x.(charmhub.RefreshConfig)
+	if !ok {
+		return false
```
hmlanigan, line 123: I've started to add c *gc.C to the matchers, then I can do
 c.Assert(ok, gc.IsTrue, gc.Commentf("detailed reason")
You can stop the test there, and know exactly what the issue is, rather than the terse go mock response.

apiserver/facades/controller/charmrevisionupdater/updater.go
benhoyt: Done

apiserver/facades/controller/charmrevisionupdater/updater_test.go
benhoyt: Done (well, I don't want to move everything to fields on the struct, but I created a wrapper function which has the same effect of reducing the duplication).

apiserver/facades/controller/charmrevisionupdater/interface_test.go
benhoyt: I think this would be fine for the interface type conversion (as that's simply a bug in the code). Though I'm not sure you can do this in general, because the `Matches()` method has to return false if it doesn't match in case of matching multiple items. I decided to just leave this as is.

apiserver/facades/controller/charmrevisionupdater/testing/suite.go
benhoyt: I moved this into apiserver/facades/client/client where it's used.

benhoyt:
$$merge$$

apiserver/facades/controller/charmrevisionupdater/interface_test.go
benhoyt: For future reference: Ian noted that you can `c.Fail()` and still return false.

benhoyt:
[Make charmhub updater unit testable and not use *state.State directly](https://trello.com/c/SyL9jz1u/1313-make-charmhub-updater-unit-testable-and-not-use-statestate-directly)

apiserver/facades/controller/charmrevisionupdater/interface_test.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}
+	return true
+}
+
+func (charmhubConfigMatcher) String() string {
+	return "matches config"
+}
+
+func newFakeCharmstoreClient(st charmrevisionupdater.State) (charmstore.Client, error) {
```
manadart, line 151: I'm not particularly bullish on using GoMock for everything - a test suite might only need one or two trivial fakes; but I think consistency within a suite is worth adopting. 

This one would be trivial as a generated mock.

apiserver/facades/controller/charmrevisionupdater/interface_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+func (c *fakeCharmstoreClient) ResourceMeta(_ csparams.Channel, _ *charm.URL, _ string, revision int) (csparams.Resource, error) {
+	panic("not implemented")
+}
+
+func (c *fakeCharmstoreClient) ServerURL() string {
+	panic("not implemented")
+}
+
+type facadeContextShim struct {
```
manadart, line 198: I would do the following here:
- Pass `ctx.Auth()` to `NewCharmRevisionUpdaterAPIState`.
- Do `AuthController()` at the top of `UpdateLatestRevisions`.

This makes it more consistent with other APIs where the auth required might depend on the method called.

Then you don't need this mock; just the `authorizer`. `state` is never set for this in any case.

PR #12425, opened 2020-12-09
Ensure process limits are correctly set with using juju-db snap

wallyworld:
When using the juju-db snap for mongo, the necessary process limits were not being set.
This PR adds the necessary systemd service property overrides.
The ensureServer() func needed a little rearranging to ensure the limits were set for both the case where the snap was specified using the feature flag, and for where it comes with the series, eg focal

Also, there was a setupDataDirectory() method where the error was being ignored. Fixing that had a little test fallout.

## QA steps

juju bootstrap lxd --bootstrap-series=focal
ssh to controller and use ps to find the juju-db snap process
`cat /proc/<id>/limits`
to check the limits are as expected, eg LimitNOFILE=64000
also check /etc/systemd/system/snap.juju-db.daemon.service.d/overrides.conf is created

## Bug reference

https://bugs.launchpad.net/juju/+bug/1907139

### DIFF
--- a/mongo/mongo.go
+++ b/mongo/mongo.go
@@ -30,7 +30,6 @@ import (
 	"github.com/juju/juju/packaging"
 	"github.com/juju/juju/packaging/dependency"
 	"github.com/juju/juju/service"
-	"github.com/juju/juju/service/common"
 	"github.com/juju/juju/service/snap"
 )
 
@@ -426,6 +425,9 @@ type EnsureServerParams struct {
 	// DataDir is the machine agent data directory.
 	DataDir string
 
+	// LogDir is where mongo logs go.
+	LogDir string
+
 	// Namespace is the machine agent's namespace, which is used to
 	// generate a unique service name for Mongo.
 	Namespace string
@@ -484,9 +486,12 @@ func ensureServer(args EnsureServerParams, mongoKernelTweaks map[string]string) 
 		args.DataDir, args.StatePort,
 	)
 
-	setupDataDirectory(args, usingMongoFromSnap)
+	err = setupDataDirectory(args, usingMongoFromSnap)
+	if err != nil {
+		return zeroVersion, errors.Trace(err)
+	}
 
-	if err := installMongod(mongoDep, hostSeries, dataDir); err != nil {
+	if err := installMongod(mongoDep, usingMongoFromSnap, hostSeries, dataDir); err != nil {
 		// This isn't treated as fatal because the Juju MongoDB
 		// package is likely to be already installed anyway. There
 		// could just be a temporary issue with apt-get/yum/whatever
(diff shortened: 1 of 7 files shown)


hpidcock (approved):
Thanks.

wallyworld:
$$merge$$

wallyworld:
[juju-db snap process limits](https://trello.com/c/cCno5N2G/1314-juju-db-snap-process-limits)

mongo/mongo.go
```diff
@@ -681,25 +686,21 @@ func logVersion(mongoPath string) {
...
-	// as a snap <name>. snap install <name> will then do the Right Thing (TM).
-	files, err := ioutil.ReadDir(path.Join(dataDir, "snap"))
-	if err == nil {
-		for _, fullFileName := range files {
-			_, fileName := path.Split(fullFileName.Name())
-			if jujuDbLocalSnapPattern.MatchString(fileName) {
-				snapName = fullFileName.Name()
+
+	if usingMongoFromSnap {
+		jujuDbLocalSnapPattern := regexp.MustCompile(`juju-db_[0-9]+\.snap`)
```
jameinel, line 693: Is 'juju-db_12345' the only name? wouldn't it be 'juju-db_1.2' or something along those lines? I realize you didn't change the regex, I'm just surprised at it.

mongo/mongo.go
```diff
@@ -711,14 +712,31 @@ func installMongod(mongoDep packaging.Dependency, hostSeries, dataDir string) er
...
-	service, err := snap.NewService(snapName, ServiceName, conf, snap.Command, snapChannel, "", backgroundServices, prerequisites)
+	snapSvc, err := snap.NewService(snapName, ServiceName, conf, snap.Command, snapChannel, "", backgroundServices, prerequisites)
 	if err != nil {
 		return errors.Trace(err)
 	}
 
-	return service.Install()
+	if usingMongoFromSnap {
+		defer func() {
+			err = snapSvc.ConfigOverride()
```
jameinel, line 725: Shouldn't we check if we already have an error here? (eg snapSvc.Install() returned an error, why would we do ConfigOverride?)

mongo/mongo.go
```diff
@@ -711,14 +712,31 @@ func installMongod(mongoDep packaging.Dependency, hostSeries, dataDir string) er
...
+			if err == nil {
+				err = snapSvc.Restart()
+			}
+		}()
+	}
+
+	// If we are not forcing a mongo snap via a feature flag, install the
+	// package list (which may also include snaps for focal+) provided by
+	// the mongo dependency for our series.
+	if !featureflag.Enabled(feature.MongoDbSnap) {
```
jameinel, line 735: This feels clearer to understand without the negation:
```
if featureFlag.Enabled(feature.MongoDbSnap) {
  err = snapSvc.install()
} else {
  err = packaging.InstallDependency(mongoDep, hostSeries)
}
if err != nil {
  return errors.Trace(err)
}
if usingMongoFromSnap {
  err = snapSvc.ConfigOverride()
  if err == nil {
    err = snapSvc.Restart()
  }
}
return errors.Trace(err)
```

Thought?

mongo/mongo.go
```diff
@@ -681,25 +686,21 @@ func logVersion(mongoPath string) {
...
+		files, err := ioutil.ReadDir(path.Join(dataDir, "snap"))
+		if err == nil {
+			for _, fullFileName := range files {
+				_, fileName := path.Split(fullFileName.Name())
+				if jujuDbLocalSnapPattern.MatchString(fileName) {
+					snapName = fullFileName.Name()
+				}
 			}
 		}
 	}
```
jameinel, line 706: Presumably the below dependency on "core" will change with the new versions of mongo 4.4? (but I guess that would only be in develop)

jameinel (commented):
Just some comments. Mostly to make it easier to understand what is going on, I don't think there are any structural changes.

PR #12428, opened 2020-12-09
Progress bar test susceptible to timing failures

SimonRichardson:
The progress bar tests are susceptible to timing failures if running on
slow CI boxes. The fix is to pass in a clock to prevent that from
happening, so when we run the tests we can pass in our own mock clock
which we can target exactly the time we want to test against.

## QA steps

Tests pass, along with the following still working!

```sh
$ juju bootstrap lxd test
$ juju download ubuntu
```


achilleasa (approved):
Ah. Sorry I missed the use of the real clock in the tests.

SimonRichardson:
> Ah. Sorry I missed the use of the real clock in the tests.

Which makes me think that upstream doesn't test this on slow machines or how have they never run into this before :thinking:

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #12430, opened 2020-12-09
Charmhub resources meta part 2

SimonRichardson:
This supersedes #12421. Requires #12414 to land first (will rebase once landed)

The following ensures that we get the metadata that we want. This is
abstracted out into a client so that both charmhub and the charmstore
can get the information in the same logic, without duplication. Adds test 
for the charmhub repository resolving. It currently doesn't work as we're 
waiting for file path to be put back into the API after they removed it.

Additionally:

Also cleaned up the charmhub headers so that we get ALL the headers for
all scenarios. Previously it only looked in the context, but the
platform could actually be in the actions as well.

It probably should be another branch, but life is too short!

## QA steps

```sh
$ export JUJU_DEV_FEATURE_FLAGS="charm-hub"
$ juju bootstrap localhost
$ juju add-model six --config charm-hub-url="https://api.staging.snapcraft.io"

# Deploy a CharmHub charm with resources, set the region to trigger resource_get in the charm.
$ juju deploy postgresql --config aws_region="us-east-1"
# wait for the charm to install and try to get the resource successfully
# there will be a status message about installing the Wal-e snap
$ juju show-status-log postgresql/0
Time                   Type       Status       Message
....
08 Dec 2020 22:27:55Z  juju-unit  executing    running install hook
08 Dec 2020 22:28:52Z  workload   maintenance  Updating apt cache
08 Dec 2020 22:28:54Z  workload   maintenance  Installing wal-e snap
```

### DIFF
--- a/apiserver/facades/client/resources/repository.go
+++ b/apiserver/facades/client/resources/repository.go
@@ -146,9 +146,7 @@ func (ch *charmHubClient) ResolveResources(resources []charmresource.Resource) (
 }
 
 func (ch *charmHubClient) ResourceInfo(curl *charm.URL, origin corecharm.Origin, name string, revision int) (charmresource.Resource, error) {
-	charmOrigin := ch.id.Origin
-
-	if charmOrigin.ID == "" {
+	if origin.ID == "" {
 		return charmresource.Resource{}, errors.Errorf("empty charm ID")
 	}
 
@@ -159,14 +157,14 @@ func (ch *charmHubClient) ResourceInfo(curl *charm.URL, origin corecharm.Origin,
 		cfg charmhub.RefreshConfig
 		err error
 
-		refPlatform = charmhub.RefreshPlatform(charmOrigin.Platform)
+		refPlatform = charmhub.RefreshPlatform(origin.Platform)
 	)
-	if charmOrigin.Revision != nil {
-		cfg, err = charmhub.DownloadOneFromRevision(charmOrigin.ID, *charmOrigin.Revision, refPlatform)
+	if origin.Revision != nil {
+		cfg, err = charmhub.DownloadOneFromRevision(origin.ID, *origin.Revision, refPlatform)
 	} else if curl.Revision >= 0 {
-		cfg, err = charmhub.DownloadOneFromRevision(charmOrigin.ID, curl.Revision, refPlatform)
+		cfg, err = charmhub.DownloadOneFromRevision(origin.ID, curl.Revision, refPlatform)
 	} else {
-		cfg, err = charmhub.DownloadOneFromChannel(charmOrigin.ID, charmOrigin.Channel.String(), refPlatform)
+		cfg, err = charmhub.DownloadOneFromChannel(origin.ID, origin.Channel.String(), refPlatform)
 	}
 	if err != nil {
 		return charmresource.Resource{}, errors.Trace(err)
--- a/apiserver/facades/client/resources/repository.go
+++ b/apiserver/facades/client/resources/repository.go
@@ -150,27 +150,37 @@ func (ch *charmHubClient) ResourceInfo(curl *charm.URL, origin corecharm.Origin,
 		return charmresource.Resource{}, errors.Errorf("empty charm ID")
 	}
 
-	// We prefer the revision over the channel, so we attempt to locate the
-	// resource meta data for the identical one we've located. If we don't have
-	// a revision, then fall back to locating via the channel.
+	// Get all the resource for everything and just find out which one matches.
+	// The order is expected to be kept so when the response is looped through
+	// we get channel, then revision.
 	var (
-		cfg charmhub.RefreshConfig
-		err error
-
+		configs     []charmhub.RefreshConfig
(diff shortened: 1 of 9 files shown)


apiserver/facades/client/resources/repository.go
```diff
@@ -43,58 +43,183 @@ type CharmHub interface {
...
+		return c.client.ResourceInfo(id.URL, id.Origin, res.Name, res.Revision)
+	}
+	// The caller fully-specified a resource with a different resource
+	// revision than the one associated with the charm in the backend. So
+	// we use the provided info as-is.
+	return res, nil
+}
+
 // TODO hml 2020-12-2
 // charmHubClient needs to be "caching" like the charmstoreclient is??
```
hmlanigan, line 119: can we remove this TODO now?

apiserver/facades/client/resources/repository.go
```diff
@@ -43,58 +43,183 @@ type CharmHub interface {
...
 	}
-	results := make([]charmresource.Resource, len(chResources))
-	var i int
-	for _, res := range chResources {
-		results[i] = res
-		i += 1
+	return resolved, nil
+}
+
+func (ch *charmHubClient) ResourceInfo(curl *charm.URL, origin corecharm.Origin, name string, revision int) (charmresource.Resource, error) {
```
hmlanigan, line 152: Both the charmHubClient and charmStoreClient have id embedded in them, and ResourceInfo takes an origin as input.  Suggest ResourceInfo doesn't take a origin as input, or that clients do not need the id.

charmhub/refresh.go
```diff
@@ -9,8 +9,8 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/juju/collections/set"
 	"github.com/juju/errors"
+	"github.com/juju/utils/set"
```
hmlanigan, line 13: Use collections/set instead.

SimonRichardson: My IDE, every time!

apiserver/facades/client/resources/repository.go
SimonRichardson: I've moved stuff around, I'd rather not pass in the CharmID and be more explicit about URL and Origin here. Only because Metadata is hanging off CharmID and people would assume that it has an important role when it comes to locating a resource via `ResourceInfo`, which is not the case.

hmlanigan: Quick offline conversation.  We'll use the origin passed in consistently, and not the one embedded in the resourceClient.

apiserver/facades/client/resources/repository.go
```diff
@@ -43,71 +43,177 @@ type CharmHub interface {
...
+	return charmresource.Resource{}, errors.NotFoundf("charm resource %q at revision %d", name, revision)
 }
 
 // listResources composes, a map of details for each of the charm's
 // resources. Those details are those associated with the specific
 // charm revision. They include the resource's metadata and revision.
 // Found via the CharmHub api.
 func (ch *charmHubClient) listResources() (map[string]charmresource.Resource, error) {
-	cfg, err := charmhub.DownloadOneFromChannel(ch.id.URL.Name, ch.id.Origin.Channel.String(), charmhub.RefreshPlatform(ch.id.Origin.Platform))
+	charmOrigin := ch.id.Origin
```
hmlanigan, line 200: using the resourceClient embedded id instead of a passed in one.

hmlanigan (commented):
From QA: 
juju deploy postgresql --config aws_region="us-east-1"
Located charm "postgresql" in charm-hub, revision 208
ERROR listing resources for charm "ch:postgresql-208": Both id and name are present in the request for this Charm. It should be one or the other exclusively.

apiserver/facades/client/resources/repository.go
```diff
@@ -43,71 +43,177 @@ type CharmHub interface {
...
+	}
+	return resolved, nil
+}
+
+func (ch *charmHubClient) ResourceInfo(curl *charm.URL, origin corecharm.Origin, name string, revision int) (charmresource.Resource, error) {
+	if origin.ID == "" {
+		return charmresource.Resource{}, errors.Errorf("empty charm ID")
+	}
+
+	// We prefer the revision over the channel, so we attempt to locate the
```
hmlanigan, line 153: Why do we prefer revision over channel here?  With the charms, its the other way around?

For resources in charmstore, we use the newest revision of the resource, even if it's newer than what's in the charm revision.

SimonRichardson: This is because we know exactly which charm we have, so we want to locate the exact resource revision associated (match exactly for the case there has been an update in between deploying a charm and deploying a resource). This is the opposite when selecting a charm, we don't know what revision to look for, so use the channel (default release).

If the revision is found then fall back to the latest channel.

hmlanigan (approved):
The problem found in qa is resolved.  Full deploy of a resource will not work until some charmhub updates, which should be available shortly.  

LGTM

hmlanigan:
$$merge$$

PR #12431, opened 2020-12-09
[2.8] Set unit error state when relation created hooks fail

achilleasa:
The way that the uniter worker handles RunHook errors is by dropping errors to the floor in the resolver main loop and then entering the loop once again under the assumption that the resolver will check the **local state** for a pending RunHook operation (Operation == RunHook and Step == Pending) and move the unit into an error state.

This design assumes that the individual resolvers check that the current local state is in `Continue` mode (meaning that each resolver should go ahead and figure out if it needs to emit an operation) and return `ErrNoOperation` otherwise so that the main resolver can eventually apply the above logic and set the required error state.

Unfortunately, this particular check was not applied by the recently (2.8) added CreatedRelation resolver. As a result, if a `xxx-relation-created` hook failed, the uniter would keep invoking the aforementioned resolver which would happily return a `xxx-relation-created` hook operation causing the unit to get stuck in an infinite loop.

This PR addresses this problem thus allowing the unit to reach an error state as expected. 

## QA steps

```sh
# apply this patch
diff --git a/acceptancetests/repository/charms/ubuntu/metadata.yaml b/acceptancetests/repository/charms/ubuntu/metadata.yaml
index 1846048417..e5b7b19648 100644
--- a/acceptancetests/repository/charms/ubuntu/metadata.yaml
+++ b/acceptancetests/repository/charms/ubuntu/metadata.yaml
@@ -12,4 +12,6 @@ series:
   - bionic
   - eoan
   - focal
-
+peers:
+  peer:
+    interface: http


# Then create this file with the specified contents
$ cat acceptancetests/repository/charms/ubuntu/hooks/peer-relation-created
#!/bin/bash
set -e
# This should fail when executed by a non-leader unit!
relation-set -r $JUJU_RELATION_ID --app foo=bar

# Finally, run the following command and check that the non-leader unit reaches an error state
$ juju deploy -n2 ./acceptancetests/repository/charms/ubuntu

$ juju status
Model    Controller  Cloud/Region            Version  SLA          Timestamp
default  test        lxd-with-cache/default  2.8.7.1  unsupported  18:33:58Z

App     Version  Status  Scale  Charm   Store  Rev  OS      Notes
ubuntu           error       2  ubuntu  local    1  ubuntu

Unit       Workload  Agent  Machine  Public address  Ports  Message
ubuntu/0*  unknown   idle   0        10.59.233.117
ubuntu/1   error     idle   1        10.59.233.252          hook failed: "peer-relation-created" <---- look for this

Machine  State    DNS            Inst id        Series  AZ  Message
0        started  10.59.233.117  juju-be2af9-0  bionic      Running
1        started  10.59.233.252  juju-be2af9-1  bionic      Running
```

## Bug reference
https://bugs.launchpad.net/juju/+bug/1906706

### DIFF
--- a/worker/uniter/relation/resolver.go
+++ b/worker/uniter/relation/resolver.go
@@ -323,6 +323,11 @@ func (r *createdRelationsResolver) NextOp(
 		return nil, resolver.ErrNoOperation
 	}
 
+	// We should only evaluate the resolver logic if there is no other pending operation
+	if localState.Kind != operation.Continue {
+		return nil, resolver.ErrNoOperation
+	}
+
 	if err := r.stateTracker.SynchronizeScopes(remoteState); err != nil {
 		return nil, errors.Trace(err)
 	}
--- a/worker/uniter/relation/resolver_test.go
+++ b/worker/uniter/relation/resolver_test.go
@@ -1292,6 +1292,29 @@ func (s *relationCreatedResolverSuite) TestCreatedRelationResolverFordRelationNo
 	})
 }
 
+// This is a regression test for LP1906706
+func (s *relationCreatedResolverSuite) TestCreatedRelationsResolverWithPendingHook(c *gc.C) {
+	ctrl := gomock.NewController(c)
+	defer ctrl.Finish()
+
+	r := mocks.NewMockRelationStateTracker(ctrl)
+
+	localState := resolver.LocalState{
+		State: operation.State{
+			Installed: true,
+			Kind:      operation.RunHook,
+			Step:      operation.Pending,
+		},
+	}
+	remoteState := remotestate.Snapshot{
+		Life: life.Alive,
+	}
+
+	createdRelationsResolver := relation.NewCreatedRelationResolver(r)
+	_, err := createdRelationsResolver.NextOp(localState, remoteState, &mockOperations{})
+	c.Assert(errors.Cause(err), gc.Equals, resolver.ErrNoOperation, gc.Commentf("expected to get ErrNoOperation when a RunHook operation is pending"))
+}
+
 type mockRelationResolverSuite struct {
 	mockRelStTracker *mocks.MockRelationStateTracker
 	mockSupDestroyer *mocks.MockSubordinateDestroyer


jameinel:
Given the discussion, couldn't we also just 'exit 1' in the bash script and have the same effect? (iow, any non-zero relation-created hook just loops indefinitely?)

worker/uniter/relation/resolver.go
```diff
@@ -323,6 +323,11 @@ func (r *createdRelationsResolver) NextOp(
 		return nil, resolver.ErrNoOperation
 	}
 
+	// We don't need to do anything if there is a pending hook.
+	if localState.Kind == operation.RunHook && localState.Step == operation.Pending {
```
jameinel, line 327: Trying to look at the other examples, 'relationsResolver' uses:
```
        if localState.Kind != operation.Continue {
                return nil, resolver.ErrNoOperation
        }
```

I find it very strange that Kind is usually something like the hook being run, but here kind is a type of operation.

It just seems very unintuitive that the way to get a unit into an error is to return NoOperation.

Looking elsewhere,NoOperation tends to be surpressed. (if err != ErrNoOperation: return)

worker/uniter/relation/resolver_test.go
```diff
@@ -279,6 +279,37 @@ func (s *relationResolverSuite) TestNextOpNothing(c *gc.C) {
...
+		uniterAPICall("RelationsStatus", unitEntity, params.RelationUnitStatusResults{Results: []params.RelationUnitStatusResult{{RelationResults: []params.RelationUnitStatus{}}}}, nil),
+		uniterAPICall("State", unitEntity, unitStateResults, nil),
+	)
+	r := s.newRelationStateTracer(c, apiCaller, unitTag)
+	assertNumCalls(c, &numCalls, 4)
+
+	localState := resolver.LocalState{
+		State: operation.State{
+			Installed: true,
+			Kind:      operation.RunHook,
```
jameinel, line 301: This test seems very specific about the precise precondition that would result in NoOp. I'm not sure it encapsulates the idea of "a failed hook should give the right post-condition to result in recording an error".

jameinel (commented):
the test does cover the exact conditions of the change, it doesn't seem like it catches the conceptual 'failing hooks should set error status' to make it clear to me that this is the correct fix.

worker/uniter/relation/resolver_test.go
achilleasa: Ideally, this would be one of those odd uniter tests (the ones with their own "DSL" for describing the spec). 

However, given the [way](https://github.com/juju/juju/blob/2.8/worker/uniter/resolver.go#L142-L144) that the uniter changes to an error state  writing a proper test is cumbersome because you have to pretty much set up all resolvers (and mock the API calls that they may potentially make). So instead, I opted to explicitly test the "if hook pending then return ErrNoOperations" logic (which causes the main resolver loop to set the error state) as the lack of this check in the created relations resolver was the reason for the bug.

worker/uniter/relation/resolver.go
achilleasa: Yes... I really really wish we could allocate some dev time to rewrite the uniter and resolvers from scratch... It took me 1h to track down why the error state gets set with the regular relation hook errors but was not for the relation-created even though you could see the errors splatted in the logs! Take a look for example at https://github.com/juju/juju/blob/2.8/worker/uniter/resolver.go#L107-L137

achilleasa:
> Given the discussion, couldn't we also just 'exit 1' in the bash script and have the same effect? (iow, any non-zero relation-created hook just loops indefinitely?)

Yes. As long as the hook fails (for whatever reason), the uniter loop would attempt to resolve the next operation and the relation-created resolver would be like "sure, here is the same relation-created hook that has not yet been committed". So the charm would get stuck in an infinite loop for the same relation-created hook.

worker/uniter/relation/resolver.go
jameinel: Doesn't that just mean "if you get NoOpp try the next handler" not "report an error on the current thing because the thing failed"?

hmlanigan (approved):
QA went well.  LGTM, other comments have been addressed.

achilleasa:
$$merge$$

achilleasa:
$$merge$$

PR #12432, opened 2020-12-09
Arm64 unit test fixes with app constraint

SimonRichardson:
The following adds a new method to the connsuite[1] to ensure that when we
create a testing application that has the correct arch constraints.
The fact that the constraints weren't set would mean that it would
default to amd64. This is obviously not what we want and so forcing it
to be the host arch when adding the application solves this.

1. I know :(

## QA steps

Two ways:

 1. Check this code out on an amd64 box and run the following:

```sh
go test -v ./cmd/jujud/agent/. -check.v -check.f=MachineSuite.TestManageModelRunsInstancePoller
```

 2. Wait for the CI unit tests to run the arm64 ones after merging.

### DIFF
--- a/cmd/jujud/agent/machine_test.go
+++ b/cmd/jujud/agent/machine_test.go
@@ -281,7 +281,7 @@ func (s *MachineSuite) TestManageModelRunsInstancePoller(c *gc.C) {
 
 	// Add one unit to an application;
 	charm := s.AddTestingCharm(c, "dummy")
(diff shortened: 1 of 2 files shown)


manadart (approved):
All good.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

> FAIL: worker_test.go:978: workerSuite.TestDyingMachinesAreRemoved

PR #12451, opened 2020-12-14
Fix tests on CentOS (invert err test)

benhoyt:
@manadart [pointed out](https://github.com/juju/juju/pull/12341#issuecomment-741724626) that PR #12341 broke the unit tests on CentOS. Turned out to be a trivial logic bug that I'd introduced. Typed `if err != nil` one too any times, I guess. :-)

I've tested this locally by using gohack on `github.com/juju/os/v2` and overriding `hostSeries()` to return `centos8` -- breaks before the fix and works fine after, so hopefully it'll fix CentOS CI.

### DIFF
--- a/juju/testing/conn.go
+++ b/juju/testing/conn.go
@@ -432,7 +432,7 @@ func DefaultVersions(conf *config.Config) []version.Binary {
 	defaultSeries.Add(config.PreferredSeries(conf))
 
 	hostSeries, err := series.HostSeries()
-	if err != nil {
+	if err == nil {
 		defaultSeries.Add(hostSeries)
 	}
 


wallyworld (commented):
lgtm but target 2.8

benhoyt:
$$merge$$

benhoyt:
[Fix upgrade unit tests on CentOS (broken since commit f3602d4)](https://trello.com/c/Mc61hRXh/185-fix-upgrade-unit-tests-on-centos-broken-since-commit-f3602d4)

juju/testing/conn.go
```diff
@@ -432,7 +432,7 @@ func DefaultVersions(conf *config.Config) []version.Binary {
 	defaultSeries.Add(config.PreferredSeries(conf))
 
 	hostSeries, err := series.HostSeries()
-	if err != nil {
+	if err == nil {
```
manadart, line 435: I looked over this and missed it :)

benhoyt: Me too! I guess we're so used to seeing `if err != nil` ... and usually it's correct.

PR #12455, opened 2020-12-14
Sanitize charm origin

SimonRichardson:
The following aims to sanitize the charm origin before it leaves the API
server. Although not required, it ensures that we prevent the leaking of
the platform types out to the client.

The idea is that if we send "arch=amd64" and they send back "arch=all",
then we can replace "all" with "amd64" and the client is none the wiser.

The one difference is that we will recalculate the OS from the series if
the series in the platform is labeled as "all".

This will once and for all kill off the idea of all leaking out.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju deploy wiki-simple
$ juju deploy ubuntu
```

### DIFF
--- a/apiserver/facades/client/charms/repositories.go
+++ b/apiserver/facades/client/charms/repositories.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"net/url"
+	"strings"
 
 	"github.com/juju/charm/v8"
 	"github.com/juju/charmrepo/v6"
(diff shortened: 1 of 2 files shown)


apiserver/facades/client/charms/repositories.go
```diff
@@ -151,6 +170,42 @@ func (c *chRepo) findBundleDownloadURL(curl *charm.URL, origin corecharm.Origin)
...
+		os, err := series.GetOSFromSeries(requested.Series)
+		if err != nil {
+			return received, errors.Trace(err)
+		}
+		result.OS = os.String()
+	}
+	return result, nil
+}
+
+func sanitizeCoreCharmOrigin(received, requested corecharm.Origin) (corecharm.Origin, error) {
```
SimonRichardson, line 199: Fun with types :face_with_head_bandage:

apiserver/facades/client/charms/repositories.go
```diff
@@ -151,6 +170,42 @@ func (c *chRepo) findBundleDownloadURL(curl *charm.URL, origin corecharm.Origin)
...
+		result.OS = requested.OS
+	}
+	if received.Series == "all" {
+		result.Series = requested.Series
+
+		os, err := series.GetOSFromSeries(requested.Series)
+		if err != nil {
+			return received, errors.Trace(err)
+		}
+		result.OS = os.String()
```
hmlanigan, line 194: why overwrite OS if done at line 187? os must be lower case, and should be in the received, or the call would have failed.

apiserver/facades/client/charms/repositories.go
```diff
@@ -151,6 +170,42 @@ func (c *chRepo) findBundleDownloadURL(curl *charm.URL, origin corecharm.Origin)
...
+	}
+	return result, nil
+}
+
+func sanitizeCoreCharmOrigin(received, requested corecharm.Origin) (corecharm.Origin, error) {
+	a := apicharm.CoreCharmOrigin(received)
+	b := apicharm.CoreCharmOrigin(received)
```
hmlanigan, line 201: a & b are the same
s/received/requested/

hmlanigan (commented):
Adding comments, will qa after they are resolved.

hmlanigan (approved):
qa good.

SimonRichardson:
$$merge$$

PR #12457, opened 2020-12-14
do not close reader and try to read after

hmlanigan:
Reading from an io read closer after it has been closed causes a panic.  The resource operations repository was closing a reader then returning it to be read, eventually by a unitSetter, causing a panic.  Also let to the unit's resource doc not being set with an empty pendingID.  

Only seen on deploy, as this step put the resource into state to be found during subsequent add-units.

## QA steps

```console
$ juju bootstrap localhost resources
$ juju deploy cs:postgresql --config aws_region="us-east-1"

# wait for unit to be fully up and running
$ juju list-resources postgresql/0 --details --format yaml
- unitID: postgresql/0
  unit:
    resourceid: postgresql/wal-e
    ...
    fingerprint: 38b060a751ac96384cd9327eb1b1e36a21fdb71114be07434c0cc7bf63f6e1da274edebfe76f65fbd51ad2f14898b95b
    size: 0
   ...
  expected:
   ...
    fingerprint: 38b060a751ac96384cd9327eb1b1e36a21fdb71114be07434c0cc7bf63f6e1da274edebfe76f65fbd51ad2f14898b95b
    size: 0
    ...
  progress: -1
  unitnumber: 0
  revprogress: "0"
```

### DIFF
--- a/resource/repositories/repository.go
+++ b/resource/repositories/repository.go
@@ -12,8 +12,10 @@ import (
 	"github.com/juju/juju/resource"
 )
 
-// EntityRepository exposes the functionality needed to cache data from
-// the charm store. The operations apply to a single application (or unit).
+// EntityRepository exposes the functionality needed to access data from
+// a repository.  If the data is not in state, get it from charm hub or
+// charm store and store in state. The operations apply to a single application
+// (or unit).
 type EntityRepository interface {
 	// GetResource returns the resource data for the identified resource.
 	GetResource(name string) (resource.Resource, error)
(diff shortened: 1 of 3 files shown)


SimonRichardson (approved):
:clap:  for finding this out.

The idea that two different pieces of code look after the lifecycle of an io.Reader is terrible. I've run into this before with go, I wonder if we're leaking out the underlying type or not modeling correctly so that we can pass the Reader around?

hmlanigan:
$$merge$$

PR #12462, opened 2020-12-16
Handle Ctrl-C during bootstrap; pass ctx to waitForAgentInitialisation

benhoyt:
Bootstrap is a bit of a pain to stop/cancel half way through, even if it's obviously failing it keeps retrying stuff for a long time (eg: waitForAgentInitialization would take 10 hours before it gave up). Need to handle Ctrl-C and terminate/cleanup.

This PR threads context through parts of the bootstrap process and checks `ctx.Done()` in retry loops and at various points (except in WaitSSH, where we can actually select on `ctx.Done()` properly). We're doing it in this relatively hacky way for now to avoid wiring `ctx` up through *everything*, e.g., providers, the old AWS library, etc.

In any case, Ctrl-C is now handled in several parts of the bootstrap:

* BootstrapInstance
* WaitSSH
* waitForAgentInitialization
* K8s Deploy() step

Sample output of interrupting `waitForAgentInitialization`:

```
# set up minikube as per https://discourse.charmhub.io/t/using-juju-with-minikube/3975/3
$ juju bootstrap kubernetes k8s
Creating Juju controller "k8s" on kubernetes
WARNING bootstrapping to an unknown kubernetes cluster should be used with option --config controller-service-type. See juju help bootstrap
Fetching Juju Dashboard 0.3.0
Creating k8s resources for controller "controller-k8s"
Starting controller pod
Bootstrap agent now started
Contacting Juju controller at 10.99.3.123 to verify accessibility...
    # hangs due to load balancer error
^C Ctrl-C pressed, stopping bootstrap and cleaning up resources
    # max 3 second delay here
ERROR contacting controller (cancelled): dial tcp 10.99.3.123:17070: i/o timeout
    # during cleanup, I pressed Ctrl-C a couple more times just for kicks
^C Ctrl-C pressed, cleaning up failed bootstrap
^C Ctrl-C pressed, cleaning up failed bootstrap
    # takes about 20s to clean up (on minikube)
```

For reference, https://github.com/juju/juju/commit/c946eff768c2b5f536da1fc71321b68a273db727 is the commit where Ctrl-C "handling" was introduced (basically to ignore it).

### DIFF
--- a/caas/kubernetes/provider/bootstrap.go
+++ b/caas/kubernetes/provider/bootstrap.go
@@ -4,6 +4,7 @@
 package provider
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/url"
@@ -139,7 +140,7 @@ type controllerStack struct {
 
 type controllerStacker interface {
 	// Deploy creates all resources for controller stack.
-	Deploy() error
+	Deploy(ctx context.Context) error
 }
 
 func controllerCorelation(broker *kubernetesClient) (string, error) {
@@ -307,14 +308,27 @@ func (c *controllerStack) doCleanUp() {
 	}
 }
 
-// Deploy creates all resources for controller stack.
-func (c *controllerStack) Deploy() (err error) {
-	// TODO(benhoyt) - add ctx.Done() check between each step
+// Deploy creates all resources for the controller stack.
+func (c *controllerStack) Deploy(ctx context.Context) (err error) {
 	// creating namespace for controller stack, this namespace will be removed by broker.DestroyController if bootstrap failed.
 	nsName := c.broker.GetCurrentNamespace()
 	c.ctx.Infof("Creating k8s resources for controller %q", nsName)
 	if err = c.broker.createNamespace(nsName); err != nil {
 		return errors.Annotate(err, "creating namespace for controller stack")
+	}
+
+	// The given context is checked for cancellation manually between each
+	// step (not ideal, but it avoids wiring context absolutely everywhere).
+	isDone := func(ctx context.Context) bool {
+		select {
+		case <-ctx.Done():
+			return true
+		default:
+			return false
+		}
+	}
+	if isDone(ctx) {
+		return errors.Errorf("cancelled")
 	}
 
 	defer func() {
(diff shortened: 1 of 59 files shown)


caas/kubernetes/provider/bootstrap.go
```diff
@@ -307,15 +308,29 @@ func (c *controllerStack) doCleanUp() {
...
+	isDone := func(ctx context.Context) bool {
+		select {
+		case <-ctx.Done():
+			return true
+		default:
+			return false
+		}
+	}
+	if isDone(ctx) {
+		return errors.Errorf("cancelled")
```
wallyworld, line 331: Let's define an ErrCancelled to use in all the places

caas/kubernetes/provider/k8s.go
```diff
@@ -409,11 +409,11 @@ func (k *kubernetesClient) Create(envcontext.ProviderCallContext, environs.Creat
 
 // Bootstrap deploys controller with mongoDB together into k8s cluster.
 func (k *kubernetesClient) Bootstrap(
-	ctx environs.BootstrapContext,
+	ctx context.Context,
+	cmdCtx environs.BootstrapContext,
 	callCtx envcontext.ProviderCallContext,
```
wallyworld, line 414: I think BoostrapContext should be extended to gain a Context() method to return the context.Context.
It already has 
```
	GetStdin() io.Reader
	GetStdout() io.Writer
	GetStderr() io.Writer
	Infof(format string, params ...interface{})
	Verbosef(format string, params ...interface{})
...
```
so we'd be adding to that. Also reduces the churn in method signatures.

cmd/juju/commands/bootstrap.go
```diff
@@ -843,15 +844,25 @@ See `[1:] + "`juju kill-controller`" + `.`)
...
 	defer close(interrupted)
 	ctx.InterruptNotify(interrupted)
 	defer ctx.StopInterruptNotify(interrupted)
+	cancelCtx, cancel := context.WithCancel(context.Background())
 	go func() {
 		for range interrupted {
-			ctx.Infof("Interrupt signalled: waiting for bootstrap to exit")
+			// Space prefix is intentional, so output appears as
+			// "^C Ctrl-C pressed" instead of "^CCtrl-C pressed".
+			_, _ = fmt.Fprintln(ctx.GetStderr(), " Ctrl-C pressed, stopping bootstrap and cleaning up resources")
```
wallyworld, line 858: Either add a space or perhaps a newline would be better to keep the ^C separate from the Juju message

provider/azure/environ.go
```diff
@@ -262,7 +263,7 @@ func (env *azureEnviron) Bootstrap(
 		wg.Add(1)
 		go func(id string) {
 			defer wg.Done()
-			sdkCtx := stdcontext.Background()
+			sdkCtx := context.Background()
```
wallyworld, line 266: We now have a context that we're passing in via the Bootstrap method args, so we need to now use that one and now create a separate one here, ie we want this azure bootstrap process to react to the passed in ctx getting cancelled.
And as I said elsewhere, adding Context() to BootstrapContext means we don't need to change the Bootstrap() method signature at all, we just do ctx.Context() instead of stdcontext.Background()

provider/azure/environ.go
```diff
@@ -1696,13 +1697,13 @@ func (env *azureEnviron) DestroyController(ctx context.ProviderCallContext, cont
...
 
-func (env *azureEnviron) deleteControllerManagedResourceGroups(ctx context.ProviderCallContext, controllerUUID string) error {
+func (env *azureEnviron) deleteControllerManagedResourceGroups(ctx envcontext.ProviderCallContext, controllerUUID string) error {
 	filter := fmt.Sprintf(
 		"tagName eq '%s' and tagValue eq '%s'",
 		tags.JujuController, controllerUUID,
 	)
 	client := resources.GroupsClient{env.resources}
-	sdkCtx := stdcontext.Background()
+	sdkCtx := context.Background()
```
wallyworld, line 1706: Ultimately we also would want to add Context() to ProviderCallContext just as for BootstrapContext

provider/common/bootstrap.go
```diff
@@ -250,6 +253,13 @@ func BootstrapInstance(
 		if zone == "" || environs.IsAvailabilityZoneIndependent(err) {
 			return nil, "", nil, errors.Annotate(err, "cannot start bootstrap instance")
 		}
+
+		select {
+		case <-ctx.Done():
+			return nil, "", nil, errors.Annotatef(err, "starting bootstrap instance (cancelled)")
```
wallyworld, line 259: existing message, but let's change from "bootstrap instance" to "controller"

wallyworld (commented):
A couple of suggestions inline.
There's a bit of churn with import package renames which might go away somewhat if we make context part of BootstrapContext. But maybe not everywhere. Standardising on "context" for stdlib context and "envcontext" for ours seems ok, although elsewhere we tend to do the opposite for errors, ie we alias "errors" to "stderrors" and import "juju/errors" without the alias.

caas/kubernetes/provider/bootstrap.go
benhoyt: Done as `bootstrap.Cancelled()` / `bootstrap.IsCancelled()` in https://github.com/juju/juju/pull/12475

caas/kubernetes/provider/k8s.go
benhoyt: Done in https://github.com/juju/juju/pull/12475

cmd/juju/commands/bootstrap.go
benhoyt: Done as a newline in https://github.com/juju/juju/pull/12475

provider/azure/environ.go
benhoyt: I think this should actually be a separate context, as it's cleanup and shouldn't be cancelled if the bootstrap is cancelled (the cleanup *should* be run if the bootstrap is cancelled).

provider/common/bootstrap.go
benhoyt: Fixed in https://github.com/juju/juju/pull/12475

benhoyt:
Closing this in favor of the refactored version at https://github.com/juju/juju/pull/12475 - that adds a `Context()` method to the existing `BootstrapContext` instead of adding a `context.Context` parameter everywhere.

PR #12470, opened 2020-12-17
Force remove relation with bad data

wallyworld:
It seems there can be a case where a relation claims to have a non-zero number of units in scope, but there aren't actually any corresponding database records. This happened on a controller with a cross model relation to another controller; the cause is likely a disconnect between the controllers at just the precise wrong time. See bug https://bugs.launchpad.net/juju/+bug/1871898

The fix here is based on looking at a database dump from the affected controller. remove-relation --force is enhanced to forcibly remove a relation if there are no units in scope, even if the relation doc claims there are.

## QA steps

Almost impossible to deterministically reproduce the same scenario as in the bug.
remove-relation verified to work as expected

## Bug reference

https://bugs.launchpad.net/juju/+bug/1871898

### DIFF
--- a/api/watcher/watcher_test.go
+++ b/api/watcher/watcher_test.go
@@ -379,6 +379,8 @@ func (s *watcherSuite) TestRelationStatusWatcher(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 	assertChange(life.Alive, false, "")
 
+	err = rel.Refresh()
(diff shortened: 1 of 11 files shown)


wallyworld:
[remove-relation --force corner case](https://trello.com/c/Gsyco4Fg/1323-remove-relation-force-corner-case)

hpidcock (approved):
Thanks

wallyworld:
$$merge$$

PR #12474, opened 2020-12-17
[2.9] Populate CIDR field when faking NICs in instancepoller

achilleasa:
When the environment **does not** support the Networking interface, the instancepoller worker iterates the addresses reported as part of the instance info and creates fake NICs based on the discovered addresses.

This commit ensures that the CIDR for the generated NICs gets populated using the network details associated with each instance address to prevent the code in networkingcommon from emitting validation errors when it encounters a blank CIDR value.

Note that from 2.9 onwards, all providers support the Networking interface and can provide a NIC list so this is not an issue anymore (hence the lack of QA steps). I stumbled on this issue while testing the WIP for the Packet provider which has not yet implemented this interface.

### DIFF
--- a/core/network/address.go
+++ b/core/network/address.go
@@ -380,6 +380,15 @@ func NewScopedProviderAddress(value string, scope Scope) ProviderAddress {
 	return ProviderAddress{MachineAddress: NewScopedMachineAddress(value, scope)}
 }
 
+// NewScopedProviderAddressWithNetwork creates a new ProviderAddress by
+// embedding the result of NewScopedMachineAddress and populating its CIDR
+// value. No space information is populated.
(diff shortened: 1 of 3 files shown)


jameinel (approved):
If I understand correctly, this code already had a CIDR, we're just passing it through rather than dropping it on the floor. In which case, I'm very happy to see this change.

achilleasa:
> If I understand correctly, this code already had a CIDR, we're just passing it through rather than dropping it on the floor. In which case, I'm very happy to see this change.

Exactly. If the field is populated, we pass its value along to avoid triggering a validation error. As mentioned in the PR this is only the case for providers that do not support the Networking interface (since 2.9 all do AFAICT) so this fix only applies when creating new providers.

achilleasa:
$$merge$$

PR #12475, opened 2020-12-18
Handle Ctrl-C during bootstrap (refactored)

benhoyt:
*NOTE: this is a refactored version of #12462: this version adds a `Context()` method to the existing `environs.BootstrapContext` interface so we're only passing a single "context" around, and to reduce churn.*

Bootstrap is a bit of a pain to stop/cancel half way through, even if it's obviously failing it keeps retrying stuff for a long time (eg: waitForAgentInitialization would take 10 hours before it gave up). Need to handle Ctrl-C and terminate/cleanup.

This PR adds a `context.Context` to the existing `BootstrapContext` interface, and checks `ctx.Done()` in retry loops and at various points (except in WaitSSH, where we can actually select on `ctx.Done()` properly). We're doing it in this relatively hacky way for now to avoid wiring context up through *everything*, e.g., providers, the old AWS library, etc.

In any case, Ctrl-C is now handled in several parts of the bootstrap:

* BootstrapInstance
* WaitSSH
* waitForAgentInitialization
* K8s Deploy() step

I've tested both a working bootstrap and a `^C`-interrupted bootstrap on lxd, k8s (minikube), and AWS clouds.

Sample output of interrupting `waitForAgentInitialization`:

```
# set up minikube as per https://discourse.charmhub.io/t/using-juju-with-minikube/3975/3
$ juju bootstrap kubernetes k8s
Creating Juju controller "k8s" on kubernetes
WARNING bootstrapping to an unknown kubernetes cluster should be used with option --config controller-service-type. See juju help bootstrap
Fetching Juju Dashboard 0.3.0
Creating k8s resources for controller "controller-k8s"
Downloading images
Starting controller pod
Bootstrap agent now started
Contacting Juju controller at 10.97.223.174 to verify accessibility...
    # hangs due to load balancer error
^C
Ctrl-C pressed, stopping bootstrap and cleaning up resources
    # max 3 second delay here
ERROR contacting controller (cancelled): dial tcp 10.97.223.174:17070: i/o timeout
    # during cleanup, I pressed Ctrl-C a couple more times just for kicks
^C
Ctrl-C pressed, stopping bootstrap and cleaning up resources

Ctrl-C pressed, cleaning up failed bootstrap
^C
Ctrl-C pressed, cleaning up failed bootstrap
```

For reference, https://github.com/juju/juju/commit/c946eff768c2b5f536da1fc71321b68a273db727 is the commit where Ctrl-C "handling" was introduced (basically to ignore it).

### DIFF
--- a/apiserver/common/modelwatcher_test.go
+++ b/apiserver/common/modelwatcher_test.go
@@ -4,6 +4,7 @@
 package common_test
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/juju/cmd/cmdtesting"
@@ -103,7 +104,7 @@ func (*modelWatcherSuite) TestModelConfigFetchError(c *gc.C) {
 func testingEnvConfig(c *gc.C) *config.Config {
 	env, err := bootstrap.PrepareController(
 		false,
-		modelcmd.BootstrapContext(cmdtesting.Context(c)),
+		modelcmd.BootstrapContext(context.Background(), cmdtesting.Context(c)),
 		jujuclient.NewMemStore(),
 		bootstrap.PrepareParams{
 			ControllerConfig: testing.FakeControllerConfig(),
--- a/apiserver/facades/agent/fanconfigurer/fanconfigurer_test.go
+++ b/apiserver/facades/agent/fanconfigurer/fanconfigurer_test.go
@@ -4,6 +4,7 @@
 package fanconfigurer_test
 
 import (
+	"context"
 	"fmt"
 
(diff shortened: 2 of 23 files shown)


wallyworld (approved):
This will be great, thanks.

benhoyt:
[Clean teardown after ^C during bootstrap](https://trello.com/c/bwkE8nWb/1320-clean-teardown-after-c-during-bootstrap)

benhoyt:
$$merge$$

PR #12479, opened 2021-01-04
[dev] Implement identity discovery for ECS tasks

achilleasa:
This PR adds support for ECS task identity discovery within the context of a k8sagent instance running inside an ECS task.

ECS injects the URI to the ECS metadata service in the environment of each task. The PR modifies k8sagent to first scan the environment variables for the presence of the ECS metadata URI. If found, the agent queries it and populates the appropriate identity details for the task instance that the k8sagent is executing under.

Note that the agent will not be able to properly introduce itself as the relevant API assumes that all applications follow the k8s stateful deployment semantics (pods named 'xxx-ordinal') and will return an error when it tries to parse the identity value sent in by the agent.

### DIFF
--- a/cmd/k8sagent/initialize/config.go
+++ b/cmd/k8sagent/initialize/config.go
@@ -4,9 +4,16 @@
 package initialize
 
 import (
+	"encoding/json"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
 	"os"
+	"path"
 	"strings"
 
+	"github.com/juju/errors"
 	"github.com/juju/names/v4"
 	"github.com/juju/utils/v2/shell"
 	"github.com/juju/version"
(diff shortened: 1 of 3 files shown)


cmd/k8sagent/initialize/config.go
```diff
@@ -130,13 +137,64 @@ func defaultConfig() agent.Config {
 	return &configFromEnv{}
 }
 
-type identityFunc func() identity
+type identityFunc func() (identity, error)
+
+func defaultIdentity() (identity, error) {
```
ycliuhw, line 142: Probably we can consider moving `func defaultIdentity(cloudType) (identity, error) {
` to `caas` package, and `identityFromECSMetadataService` -> `caas/kubernetes`, `identityFromK8sMetadata ` -> `caas/ecs` package.

To achieve this, we could inject `cloudType` in init container as an environment variable.

achilleasa: Not quite sure if that would be a good idea. Granted, they are broker-specific but on the other hand they are meant to be invoked within a pod/task context as both implementations examine envvars (ok, the ecs one does an http get) to figure out the identity. As they are probably only going to be executed inside the k8sagent it makes sense to keep this logic encapsulated in there.

ycliuhw: ok, we can refactor if we need to later.

ycliuhw (approved):

ycliuhw:
$$merge$$

PR #12482, opened 2021-01-05
Service-less bundle yaml

SimonRichardson:
Services have been in a state of deprecation for a very long time,
although there was always a way to use services and applications, albeit
not in the same file. This removes services completely to make the
bundle logic simpler. Users of Juju should be using applications for
their keys in the bundle yaml. 2.9 release will explicitly call out
that services are being deprecated and 3.0 will just blindly ignore
them.

If a bundle has a services key and you attempt to deploy it, then juju
will error out about no applications present. If you have both services
and applications, then the services key will be ignored. It expects that
you know what you're doing or it was a mistake when you edited the
bundle yaml itself.

The code itself is a mechanical change of just bringing in the changes
from the upstream repos.

## QA steps

Save the following bundle.yaml

```sh
services:
    ubuntu-lite:
        charm: cs:~jameinel/ubuntu-lite-7
        num_units: 1
```

```sh
$ juju bootstrap lxd test
$ juju deploy ./bundle.yaml
ERROR cannot deploy bundle: the provided bundle has the following errors:
at least one application must be specified
```

## Documentation changes

Removes the ability to deploy a bundle with a services key.


achilleasa (approved):

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

Something fishy going on with actions

> FAIL	github.com/juju/juju/cmd/juju/action	2700.067s

PR #12483, opened 2021-01-05
[dev] Use 'juju exec' instead of 'juju run' for acceptance tests

achilleasa:
The "juju run" command on the develop branch is now used to execute
actions. The previous functionality is now provided by the "juju exec"
command instead. This commit updates the acceptance tests to use the new
command.

## QA steps

~You will need a **python 2** virtual env for this~. The tests have been updated to work with python3

```sh
make install
cd acceptancetests
./deploy_job.py lxd `which juju` ./tmp/art nw-deploy-lxd
```

### DIFF
--- a/acceptancetests/deploy_stack.py
+++ b/acceptancetests/deploy_stack.py
@@ -414,18 +414,19 @@ def copy_remote_logs(remote, directory):
             logging.warning(repr(e))
 
 
-def assess_juju_run(client):
-    responses = client.run(('uname',),
+# NOTE: Since juju 3, "juju run" has been renamed to "juju exec"
+def assess_juju_exec(client):
+    responses = client.exec_cmds(('uname',),
                            applications=['dummy-source', 'dummy-sink'])
-    for machine in responses:
-        if machine.get('ReturnCode', 0) != 0:
-            raise ValueError('juju run on machine %s returned %d: %s' % (
-                machine.get('MachineId'),
-                machine.get('ReturnCode'),
-                machine.get('Stderr')))
+    for response in responses.values():
+        if response.get('results').get('return-code'):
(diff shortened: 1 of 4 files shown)


SimonRichardson (changes requested):
I'm going to be really bad here and say python 3... python 2 is dead and so should our support for tests that don't support python 3.

Otherwise, the code is fine.

acceptancetests/deploy_stack.py
```diff
@@ -414,18 +414,19 @@ def copy_remote_logs(remote, directory):
...
+    responses = client.exec_cmds(('uname',),
                            applications=['dummy-source', 'dummy-sink'])
-    for machine in responses:
-        if machine.get('ReturnCode', 0) != 0:
-            raise ValueError('juju run on machine %s returned %d: %s' % (
-                machine.get('MachineId'),
-                machine.get('ReturnCode'),
-                machine.get('Stderr')))
+    for response in responses.values():
+        if response.get('results').get('return-code', 0) != 0:
```
petevg, line 422: It would probably be more Pythonic to say:

```
if machine.get('ReturnCode'):
    ...
```

Both the default return of .get, which is None, and 0 are Falsey values, while -1 and any number great than zero are Truthy values. You can thus just do a True/False check and be done with it.

petevg (approved):
Minor comment with a check that looked a bit wordy for my tastes. Changes aren't necessary, though, and this generally LGTM.

SimonRichardson (approved):
Other than @petevg nit's I think I'm fine with this.

acceptancetests/deploy_stack.py
achilleasa: oh ok. let me patch that.

achilleasa:
$$merge$$

PR #12484, opened 2021-01-05
Fix oplog TestStops on arm64 by inserting row before starting tailer

benhoyt:
TestStops was intermittently failing on arm64. I was able to fire up an
arm64 EC2 instance and repro it there, every 50 runs or so. I'm still
not sure why this fixes it, but with this change I can't reproduce it
over thousands of runs. Plus, inserting the row before starting the
tailer is consistent with what TestHonoursInitialTs does.

I also added a cleanup task to drop the fake oplog collection at the
end of the test. This doesn't actually seem to do anything, but I added
it because initially I thought the failure was related to
TestHonoursInitialTs not cleaning up after itself, and TestStops failing
as a result of some interaction. So the AddCleanup here doesn't actually
fix anything, but still seems like the right thing to do.

## QA steps

Create arm64 EC2 instance, clone github.com/juju/juju, install deps, get it building. Run test over and over to get it to fail, for example:

```sh
go test ./mongo -check.f TestStops -test.v -check.v -test.count=100
```

A failure (happens before the fix) will look like this:

```
----------------------------------------------------------------------
FAIL: oplog_test.go:103: oplogSuite.TestStops

[LOG] 0:00.000 DEBUG juju.testing starting mongo in /tmp/test-mgo138077763
[LOG] 0:00.000 DEBUG juju.testing using mongod at: "/usr/bin/mongod" (version=3.6.8)
[LOG] 0:00.087 DEBUG juju.testing started mongod pid 53913 in /tmp/test-mgo138077763 on port 39687
oplog_test.go:114:
    s.getNextOplog(c, tailer)
oplog_test.go:289:
    c.Fatal("timed out waiting for oplog doc")
... Error: timed out waiting for oplog doc

[LOG] 0:12.110 DEBUG juju.testing killing mongod pid 53913 in /tmp/test-mgo138077763 on port 39687 with killed
OOPS: 0 passed, 1 FAILED
--- FAIL: Test (12.12s)
```

### DIFF
--- a/mongo/oplog_test.go
+++ b/mongo/oplog_test.go
@@ -114,10 +114,10 @@ func (s *oplogSuite) TestStops(c *gc.C) {
 	_, session := s.startMongo(c)
 
 	oplog := s.makeFakeOplog(c, session)
+	s.insertDoc(c, session, oplog, &mongo.OplogDoc{Timestamp: 1})
+
 	tailer := mongo.NewOplogTailer(mongo.NewOplogSession(oplog, nil), time.Time{})
 	defer tailer.Stop()
-
-	s.insertDoc(c, session, oplog, &mongo.OplogDoc{Timestamp: 1})
 	s.getNextOplog(c, tailer)
 
 	err := tailer.Stop()
@@ -271,6 +271,10 @@ func (s *oplogSuite) makeFakeOplog(c *gc.C, session *mgo.Session) *mgo.Collectio
 		MaxBytes: 1024 * 1024,
 	})
 	c.Assert(err, jc.ErrorIsNil)
+	s.AddCleanup(func(c *gc.C) {
+		err := oplog.DropCollection()
+		c.Assert(err, jc.ErrorIsNil)
+	})
 	return oplog
 }
 


benhoyt:
!!build!!

tlm (approved):

benhoyt:
$$merge$$

jameinel:
We do certainly have a race where the oplog loop may or may not be started by the time we return from NewOplogIterator (as it is being launched in a goroutine), so it may not actually be up and active by the time it gets to Insert without doing the insert earlier. The structure of the code doesn't immediately look like we would miss anything.

However, the internal NextOpLog code looks to timeout every 1s and retry the search again after that, and testing.LongTimeout says that it waits 10s to see something (and you can see in the logs that it was 10s after the last log time by the time you got the error).

benhoyt:
@jameinel Thanks. I've been prodding on this for a while today to try to get to the bottom of it, but I think I'm going to give up now. I've narrowed it down to where I think it's a race/bug in the mgo driver's Tail method (or mongo itself, though that seems less likely). If I print out the entire collection when it times out after 10s, the row we're looking for is there in plain sight. I thought it might be the timestamp query filtering we're doing, but even if I remove that for the test, it doesn't help. So it seems like it's in Tail itself. Given this is only happening with the "fake oplog" I think we've spent enough time on it.

benhoyt:
[CI: fix oplogSuite.TestStops](https://trello.com/c/3pAkRP8M/1328-ci-fix-oplogsuiteteststops)

PR #12496, opened 2021-01-11
Use k8s in-cluster creds when inside same cluster as controller.

hpidcock:
If a model is inside the same Kubernetes cluster/cloud as the controller, the controller will use in-cluster credentials to connect to the k8s cluster.

This creates a ServiceAccount for the controller and uses a ClusterRoleBinding to give it the cluster-admin built-in role.

## QA steps

Bootstrap microk8s, GKE, AKS etc.
Add model with external k8s cluster (different cluster to the controller).
Add model with same k8s cluster (as the controller).

## Documentation changes

N/A

## Bug reference

N/A

### DIFF
--- a/api/common/cloudspec/cloudspec.go
+++ b/api/common/cloudspec/cloudspec.go
@@ -81,15 +81,16 @@ func (api *CloudSpecAPI) MakeCloudSpec(pSpec *params.CloudSpec) (environscloudsp
 		credential = &credentialValue
 	}
 	spec := environscloudspec.CloudSpec{
-		Type:             pSpec.Type,
-		Name:             pSpec.Name,
-		Region:           pSpec.Region,
-		Endpoint:         pSpec.Endpoint,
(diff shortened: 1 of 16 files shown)


wallyworld (approved):
Tested on microk8s, and after destroying a controller, the cluster role binding was left behind (maybe also role, forgot to check)

```
$ kubectl get clusterrolebindings
controller-ian    ClusterRole/cluster-admin
```

Also, I notice that cmd/juju/cloud/show.go is updatd to include the new attr but juju show-cloud does not seem to print it.

hpidcock:
$$merge$$

hpidcock:
$$merge$$

hpidcock:
[Kubernetes provider use in-cluster credentials when models share cloud with controller.](https://trello.com/c/Gd6ZCELc/1340-kubernetes-provider-use-in-cluster-credentials-when-models-share-cloud-with-controller)

PR #12501, opened 2021-01-12
Switch environ/ec2 "Instances" methods from amz package to aws-sdk-go 

benhoyt:
This switches the various Environ methods that call EC2's `DescribeInstances` API from the legacy amz package to the aws-sdk-go package. The rationale is that aws-sdk-go has better (more forgiving) retry support, whereas amz only retries on rate-limit responses.

We want to switch to aws-sdk-go for all API calls, but that's too big a chunk of work to bite off for now, so this is my attempt to do it piecemeal, starting for now with instances.

It does mean the mock for `DescribeInstances` is a bit weird. We currently use the amz `ec2test` sub-package to create instances, and the tests rely on those instances being started, so for the `DescribeInstances` mock function to work, we proxy it through to the `Instances` amz call, and convert the request/response types.

## QA steps

To exercise some of the code paths (run `juju debug-log` in another terminal window), first bootstrap with `juju bootstrap aws aws` and check that deploy and open/close ports works. With `firewall-mode=global`:

```
# test on a model with firewall-mode=global
juju add-model fwglobal --config firewall-mode=global
juju model-config firewall-mode  # prints 'global'
juju deploy mediawiki
juju deploy mysql
juju relate mysql mediawiki:db

juju expose mediawiki
# check in AWS console that port 80 is allowed in instance's security group
# also check that http://{ip_address}/ is accessible
# debug log says: opened port ranges [80/tcp from 0.0.0.0/0,::/0] in environment

juju unexpose mediawiki
# port 80 now not allowed in AWS console
# also check that http://{ip_address}/ is not accessible
# debug log says: closed port ranges [80/tcp from 0.0.0.0/0,::/0] in environment

juju destroy-model fwglobal
```

With `firewall-mode=instance` (the default):

```
# test on a model with firewall-mode=instance (the default)
juju add-model fwinstance
juju model-config firewall-mode  # prints 'instance'
juju deploy mediawiki
juju deploy mysql
juju relate mysql mediawiki:db

juju expose mediawiki
# check in AWS console that port 80 is allowed in instance's security group
# also check that http://{ip_address}/ is accessible
# debug log says: opened port ranges [80/tcp from 0.0.0.0/0,::/0] on "machine-0"

juju unexpose mediawiki
# port 80 now not allowed in AWS console
# also check that http://{ip_address}/ is not accessible
# debug log says: closed port ranges [80/tcp from 0.0.0.0/0,::/0] on "machine-0"

juju destroy-model fwinstance
```

Another test I did is to test the auth error handling. I did this by attempting to update the AWS credential on the controller. This calls `checkIAASModelCredential()` which ends up calling `environ.Instances()`, which has the aws-sdk-go error handling. It seems to be handled correctly as a credentials error:

```
$ juju update-credential aws --controller aws
Credential invalid for:
  controller:
    
The provided credentials could not be validated and 
may not be authorized to carry out the request.
Ensure that your account is authorized to use the Amazon EC2 service and 
that you are using the correct access keys. 
These keys are obtained via the "Security Credentials"
page in the AWS console.
: AWS was not able to validate the provided access credentials (AuthFailure)
  default:
    
The provided credentials could not be validated and 
may not be authorized to carry out the request.
Ensure that your account is authorized to use the Amazon EC2 service and 
that you are using the correct access keys. 
These keys are obtained via the "Security Credentials"
page in the AWS console.
: AWS was not able to validate the provided access credentials (AuthFailure)
Failed models may require a different credential.
Use ‘juju set-credential’ to change credential for these models before repeating this update.
```

I also did some local hacking to convince myself that `maybeConvertCredentialError` works the same for both `*ec2.Error` and `awserr.Error`. Tested it by manually calling `env.Instances()`, which calls `groupByName()` using amzec2, then commenting out that so it calls `DescribeInstances` using sdkec2. Both work fine and handle code=AuthFailure correctly.

### DIFF
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	sdkec2 "github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/juju/clock"
 	"github.com/juju/collections/set"
 	"github.com/juju/errors"
@@ -66,10 +67,20 @@ var (
 	_ environs.FirewallFeatureQuerier = (*environ)(nil)
 )
 
+type ec2Client interface {
+	DescribeAvailabilityZones(*sdkec2.DescribeAvailabilityZonesInput) (*sdkec2.DescribeAvailabilityZonesOutput, error)
+	DescribeInstances(*sdkec2.DescribeInstancesInput) (*sdkec2.DescribeInstancesOutput, error)
+	DescribeInstanceTypeOfferings(*sdkec2.DescribeInstanceTypeOfferingsInput) (*sdkec2.DescribeInstanceTypeOfferingsOutput, error)
+	DescribeInstanceTypes(*sdkec2.DescribeInstanceTypesInput) (*sdkec2.DescribeInstanceTypesOutput, error)
+	DescribeSpotPriceHistory(*sdkec2.DescribeSpotPriceHistoryInput) (*sdkec2.DescribeSpotPriceHistoryOutput, error)
+}
+
 type environ struct {
 	name  string
 	cloud environscloudspec.CloudSpec
 	ec2   *ec2.EC2
+
+	ec2Client ec2Client
 
 	// ecfgMutex protects the *Unlocked fields below.
 	ecfgMutex    sync.Mutex
@@ -185,8 +196,7 @@ func (e *environ) ConstraintsValidator(ctx context.ProviderCallContext) (constra
 		[]string{constraints.InstanceType},
 		[]string{constraints.Mem, constraints.Cores, constraints.CpuPower})
 	validator.RegisterUnsupported(unsupportedConstraints)
-	ec2Session := EC2Session(e.cloud.Region, e.ec2.AccessKey, e.ec2.SecretKey)
-	instanceTypes, err := e.supportedInstanceTypes(ec2Session, ctx)
+	instanceTypes, err := e.supportedInstanceTypes(ctx)
 
 	if err != nil {
 		return nil, errors.Trace(err)
@@ -349,8 +359,7 @@ func (e *environ) PrecheckInstance(ctx context.ProviderCallContext, args environ
 		return nil
 	}
 	// Constraint has an instance-type constraint so let's see if it is valid.
-	ec2Session := EC2Session(e.cloud.Region, e.ec2.AccessKey, e.ec2.SecretKey)
-	instanceTypes, err := e.supportedInstanceTypes(ec2Session, ctx)
+	instanceTypes, err := e.supportedInstanceTypes(ctx)
 	if err != nil {
 		return errors.Trace(err)
 	}
@@ -466,8 +475,7 @@ func (e *environ) StartInstance(
 		return nil, errors.Trace(err)
 	}
 
-	ec2Session := EC2Session(e.cloud.Region, e.ec2.AccessKey, e.ec2.SecretKey)
-	instanceTypes, err := e.supportedInstanceTypes(ec2Session, ctx)
+	instanceTypes, err := e.supportedInstanceTypes(ctx)
 	if err != nil {
 		return nil, wrapError(err)
 	}
@@ -1065,11 +1073,12 @@ func (e *environ) Instances(ctx context.ProviderCallContext, ids []instance.Id) 
 				need = append(need, string(ids[i]))
 			}
 		}
-		filter := ec2.NewFilter()
-		filter.Add("instance-state-name", aliveInstanceStates...)
-		filter.Add("instance-id", need...)
-		e.addModelFilter(filter)
-		err = e.gatherInstances(ctx, ids, insts, filter)
+		filters := []*sdkec2.Filter{
+			makeFilter("instance-state-name", aliveInstanceStates...),
+			makeFilter("instance-id", need...),
+			makeModelFilter(e.uuid()),
+		}
+		err = e.gatherInstances(ctx, ids, insts, filters)
 		if err == nil || err != environs.ErrPartialInstances {
 			break
 		}
@@ -1097,9 +1106,10 @@ func (e *environ) gatherInstances(
 	ctx context.ProviderCallContext,
 	ids []instance.Id,
 	insts []instances.Instance,
-	filter *ec2.Filter,
+	filters []*sdkec2.Filter,
 ) error {
-	resp, err := e.ec2.Instances(nil, filter)
+	req := &sdkec2.DescribeInstancesInput{Filters: filters}
+	resp, err := e.ec2Client.DescribeInstances(req)
 	if err != nil {
 		return maybeConvertCredentialError(err, ctx)
 	}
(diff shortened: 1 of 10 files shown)


provider/ec2/environ.go
```diff
@@ -66,11 +67,21 @@ var (
 	_ environs.FirewallFeatureQuerier = (*environ)(nil)
 )
 
+type ec2Client interface {
```
benhoyt, line 70: I've switched from using the full `ec2iface.EC2API` (which has ~1100 methods!) to this much smaller interface of just the stuff we currently need. We used to mock just the methods we needed using interface embedding in `mock_test.go`, but that causes a runtime nil panic if you miss something, whereas this is compile-time type safe. So this approach is better and clearer IMO.

provider/ec2/environ.go
```diff
@@ -1592,17 +1614,16 @@ func (e *environ) allInstanceIDs(ctx context.ProviderCallContext, filter *ec2.Fi
...
+	req := &sdkec2.DescribeInstancesInput{Filters: filters}
+	resp, err := e.ec2Client.DescribeInstances(req)
 	if err != nil {
 		return nil, errors.Annotate(maybeConvertCredentialError(err, ctx), "listing instances")
 	}
 	var insts []instances.Instance
 	for _, r := range resp.Reservations {
-		for i := range r.Instances {
-			inst := r.Instances[i]
-			// TODO(wallyworld): lookup the details to fill in the instance type data
```
benhoyt, line 1604: I'm not sure what this comment meant, but seems like we don't need it anymore?

provider/ec2/environ.go
wallyworld: Can we do this as well to ensure there's no typos in ec2Client

```
  var _ ec2Client = (ec2iface.EC2API)(nil)
```

provider/ec2/environ.go
```diff
@@ -66,11 +67,21 @@ var (
...
+	DescribeInstances(*sdkec2.DescribeInstancesInput) (*sdkec2.DescribeInstancesOutput, error)
+	DescribeInstanceTypeOfferings(*sdkec2.DescribeInstanceTypeOfferingsInput) (*sdkec2.DescribeInstanceTypeOfferingsOutput, error)
+	DescribeInstanceTypes(*sdkec2.DescribeInstanceTypesInput) (*sdkec2.DescribeInstanceTypesOutput, error)
+	DescribeSpotPriceHistory(*sdkec2.DescribeSpotPriceHistoryInput) (*sdkec2.DescribeSpotPriceHistoryOutput, error)
+}
+
 type environ struct {
 	name  string
 	cloud environscloudspec.CloudSpec
 	ec2   *ec2.EC2
```
wallyworld, line 81: Add a comment that ec2.EC2 is legacy and we are migrating away from it yada yada

provider/ec2/environ.go
```diff
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	sdkec2 "github.com/aws/aws-sdk-go/service/ec2"
```
wallyworld, line 15: how about leave this unaliased and alias "gopkg.in/amz.v3/ec2" as legacyec2 ?

provider/ec2/environ.go
```diff
@@ -1546,19 +1553,21 @@ func (e *environ) AllInstancesByState(ctx context.ProviderCallContext, states ..
...
-	return e.allInstances(ctx, filter)
+	filters := []*sdkec2.Filter{
+		makeFilter("instance-state-name", states...),
+		makeFilter("instance.group-id", group.Id),
+	}
+	return e.allInstances(ctx, filters)
 }
 
 // ControllerInstances is part of the environs.Environ interface.
 func (e *environ) ControllerInstances(ctx context.ProviderCallContext, controllerUUID string) ([]instance.Id, error) {
```
wallyworld, line 1564: This is only used in tests IIANM.  Most of those are JujuConnsuite which means the dummy provider is in use (the sooner that dies in a big fire the better). The one relevant ec2 test is TestBootstrapAndDeploy which is on a LiveTests suite - this suite will use a real AWS if run with credentials.
Just mentioning it as it would be worth considering how best to remove this method off the provider given it doesn't seem to earn its keep in terms of being maintained etc if not used in production.

provider/ec2/environ.go
wallyworld: Comment was from 2013!!! 
We used to have a Metadata() method on instances.
There's 2 of these comments in the ec2 package, they can be deleted

provider/ec2/instance.go
```diff
@@ -119,3 +119,109 @@ func (inst *ec2Instance) IngressRules(ctx context.ProviderCallContext, machineId
 	}
 	return ranges, nil
 }
+
+// AWK SDK version of ec2Instance.
+//
+// TODO(benhoyt) - get rid of ec2Instance and rename this to ec2Instance when
```
wallyworld, line 125: A little extra churn upfront but my preference would be to rename the other one as legacyInstance or something and start off with the new one being named correctly. Then we just delete the legacy instances as we go and eventually it will all be done with no further churn needed

provider/ec2/instance.go
```diff
@@ -119,3 +119,109 @@ func (inst *ec2Instance) IngressRules(ctx context.ProviderCallContext, machineId
...
+	return string(inst.Id())
+}
+
+var _ instances.Instance = (*sdkEC2Instance)(nil)
+
+func (inst *sdkEC2Instance) Id() instance.Id {
+	return instance.Id(*inst.i.InstanceId)
+}
+
+func (inst *sdkEC2Instance) Status(ctx context.ProviderCallContext) instance.Status {
```
wallyworld, line 142: should give these public methods a comment

wallyworld (approved):
Looks good.
We need to ensure that open/close port works with both global and instance firewall modes.
So QA will need to be a little more than bootstrap and deploy a charm.
Best to add the QA steps done to the PR to be sure we've covered everything.

provider/ec2/environ.go
benhoyt: Comment: sounds good, will add.

Regarding this:

```
var _ ec2Client = (ec2iface.EC2API)(nil)
```

Won't assigning the real EC2 client to an ec2Client type achieve the same thing? This happens in https://github.com/juju/juju/blob/eff28cbb84b570ad384c8d3868d3e3aad817cf42/provider/ec2/instancetypes.go#L63

provider/ec2/environ.go
benhoyt: Sounds good, though I prefer something like `amzec2` as it's more specific.

provider/ec2/environ.go
benhoyt: Yeah, doing a quick search and it does look like that's the case (it's only used in tests). I'll keep it in mind; doesn't seem in scope for this PR though.

provider/ec2/environ.go
benhoyt: Per discussion on Mattermost, it's not any "type safer", but the `var _ ec2Client = (ec2iface.EC2API)(nil)` line makes it explicit to the reader what we're trying to implement (and gives an error message closer to the definition if you get something wrong).

benhoyt:
!!build!!

benhoyt:
$$merge$$

benhoyt:
[Convert provider/ec2 "instances" operation to use awk-sdk-go instead of amz package](https://trello.com/c/kYQdP8Qf/1334-convert-provider-ec2-instances-operation-to-use-awk-sdk-go-instead-of-amz-package)

PR #12504, opened 2021-01-12
[2.9-python3-acceptancetests] Force python3 when running acceptance tests

achilleasa:
This PR applies a few changes to make some of the tests suitable for running using python3. More specifically:
- decode byte slices prior to testing regexes
- specify 'rb' access mode when reading binary files (python3 attempts to decode contents as utf-8 and explodes)

The PR targets a special branch `2.9-python3-acceptancetests` which allows us to land this set of changes, have the CI run all acceptance tests and fix the ones that break due to python2/python3 incompatibilities in a followup PR.

## QA steps

- make install from this branch
- install the latest 2.9 rc snap
- ./assess_upgrade.py lxd `which juju` /tmp/art nw-upgrade-juju-amd64-lxd --stable-juju-bin /snap/juju/current/bin/juju

### DIFF
--- a/acceptancetests/assess
+++ b/acceptancetests/assess
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # create_juju_test_env is a developer-centric tool. It creates an environment for running
 # acceptance tests locally under LXD.
--- a/acceptancetests/assess_juju_output.py
+++ b/acceptancetests/assess_juju_output.py
@@ -34,6 +34,7 @@ __metaclass__ = type
 
 log = logging.getLogger("assess_juju_output")
 
+
 def verify_juju_status_contains_display_name(client_status):
     fields_to_test = [
         "instance-id",
@@ -45,6 +46,7 @@ def verify_juju_status_contains_display_name(client_status):
         except KeyError:
             err = "juju status excludes {}".format(field)
             raise JujuAssertionError(err)
+
 
 def verify_juju_status_attribute_of_charm(charm_details):
     """Verify the juju-status of the deployed charm
@@ -106,7 +108,6 @@ def assess_juju_status(client, series):
     """
     deploy_charm_with_subordinate_charm(client, series)
     status = client.get_status()
-    verify_juju_status_fields(status)
     charm_details = status.get_applications()['dummy-sink']
     verify_juju_status_attribute_of_charm(charm_details)
     verify_juju_status_attribute_of_subordinate_charm(charm_details)
--- a/acceptancetests/assess_juju_output.py
+++ b/acceptancetests/assess_juju_output.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
     Deploy a charm and subordinate charm and verify juju-status attribute of
     the deployed charms.
--- a/acceptancetests/assess_model_migration_versions.py
+++ b/acceptancetests/assess_model_migration_versions.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """Test migrating models between controllers of increasing versions."""
 
 from __future__ import print_function
--- a/acceptancetests/assess_model_migration_versions.py
+++ b/acceptancetests/assess_model_migration_versions.py
@@ -11,6 +11,14 @@ from distutils.version import (
 import logging
 import sys
 
+from deploy_stack import (
+    BootstrapManager,
+    get_random_string,
+    )
+from utility import (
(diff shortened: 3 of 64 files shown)


acceptancetests/jujupy/client.py
```diff
@@ -563,9 +563,12 @@ def describe_substrate(env):
 def get_stripped_version_number(version_string):
     return get_version_string_parts(version_string)[0]
 
-
 def get_version_string_parts(version_string):
     # strip the series and arch from the built version.
+    # Note:
```
achilleasa, line 568: When running the `assess_upgrade` test, this func gets called with **both** a bytes version of `version_string` as well as a string one.

acceptancetests/assess
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python33
```
petevg, line 1: Typo! This should be python3.

petevg (changes requested):
Caught a small typo. Everything else looks fine, or at least as expected, if the plan is to switch over to Python3, and make fixes as you catch failures.

petevg:
As a general note, it might be interesting to run these files through a linter. You'd probably get lots of noise, but it should also catch a lot of the most straightforwardly broken things, like missing parens after print and such.

acceptancetests/assess
achilleasa: Whoops... Forgot to properly terminate my find and replace regex :D

achilleasa:
> As a general note, it might be interesting to run these files through a linter. You'd probably get lots of noise, but it should also catch a lot of the most straightforwardly broken things, like missing parens after print and such.

There are quite a few linter warnings (most are of the too-long-lines kind though) and I am trying to fix as many as I see (that's how I fixed the missing parenthesis one) but I am not generally running the linter against all files; just the ones open in my vim buffers. I guess since I will spend a bit of time on porting these I might as well fix all warnings.

acceptancetests/assess_juju_output.py
```diff
@@ -106,7 +108,6 @@ def assess_juju_status(client, series):
     """
     deploy_charm_with_subordinate_charm(client, series)
     status = client.get_status()
-    verify_juju_status_fields(status)
```
SimonRichardson, line 109: I'm guessing this actually solves a test?

acceptancetests/assess_model_migration_versions.py
```diff
@@ -67,7 +67,7 @@ def assess_model_migration_versions(stable_bsm, devel_bsm, args):
                 'version-migration',
                 resource_contents)
             migration_target_client = migrate_model_to_controller(
-                test_stable_model, devel_client)
+                test_stable_model, stable_client, devel_client)
```
SimonRichardson, line 70: Ah, this is because it's using the wrong client?

acceptancetests/assess_wallet.py
```diff
@@ -123,7 +123,7 @@ def assert_set_wallet(client, name, limit, error_strings):
     try:
         _try_setting_wallet(client, name, limit)
     except JujuAssertionError as e:
-        if error_strings['pass'] not in e.message:
+        if error_strings['pass'] not in str(e):
```
SimonRichardson, line 126: Is this correct, does `JujuAssertionError` implement `__str__` correctly?

SimonRichardson (approved):
I have a few questions, but not many changes...

acceptancetests/assess_model_migration_versions.py
achilleasa: `migrate_model_to_controller` signature is `(model, src_client, dst_client)` (no defaults!). And both this and the invocation below it used 2 args so I changed the args based on the comment above it.

acceptancetests/assess_wallet.py
achilleasa: JujuAssertionError inherits from the built-in AssertionError (which I believe should implement that). It has no other visible fields in the docs so I hope it does :D

acceptancetests/assess_juju_output.py
achilleasa: Not really sure what it does but given that:
a) it's missing
b) the bits below check specific status fields anyway

it seemed like the best approach would be to get rid of it.

achilleasa:
@petevg As you are off today and I won't be able to land (even though Simon already approved the extra bits) until you also approve the changes I will go ahead and click the dismiss review button so I can land this.

achilleasa:
$$merge$$

achilleasa:
$$merge$$ (transient error; PR only changes acceptance tests)

PR #12507, opened 2021-01-13
When removing a relation from uniter state, ignore missing units

wallyworld:
A customer somehow got to a point where units were removed but the relation state manager by a unit agent still had references to those orphaned units. This caused errors when trying to remove the relation from the unit state since the agent thought the relation still had members.

This PR adds extra logic to exclude removed units from consideration when deleting a relation from a unit agent's local state.

Note that this operation to update local unit agent state happens **after** the relation has been removed from the model and the unit has left scope; everything in the juju model related to the relation and unit is gone. It seems the check for related unit in the Members list was a sanity check. Even if the unit agent was down, deleting related units would stall until the agent cam back up. The use of --force could ram through the deletion of related units and the relation though. However, there's no point trying to run a relation-departed hook because there's nothing in the juju model that's relevant any more: no relation, no relation unit data bag. If we did attempt to run a relation-departed hook, the relation id that would be passed in would return not found if the hook attempted to do anything with it (relation-get or relation-set etc) as would the departing unit name as it doesn't exist anymore. The use of --force is something that is not normal; it is done when something has gone wrong anyway; this PR adds additional logic to deal with more of the fall out of that.

## QA steps

I couldn't reproduce the issue no matter what I did - manually stopping a unit agent, removing some related units and the relation, restarting the unit agent etc.
So QA was a regression check - create a relation between apps with multiple units, remove a unit, remove the relation etc.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1910958


wallyworld:
[Unit fails to start complaining there are members in the relation](https://trello.com/c/5rqEvIrw/1342-unit-fails-to-start-complaining-there-are-members-in-the-relation)

worker/uniter/relation/statemanager_test.go
```diff
@@ -169,11 +173,48 @@ func (s *stateManagerSuite) TestRemoveFailHasMembers(c *gc.C) {
...
+
+func (s *stateManagerSuite) TestRemoveIgnoresMissingUnits(c *gc.C) {
+	defer s.setupMocks(c).Finish()
+	stateTwo := relation.State{RelationId: 99}
+	stateTwo.Members = map[string]int64{"foo/1": 0}
+	s.expectState(c, stateTwo)
+	s.expectSetStateEmpty(c)
+	s.mockUnitGetter.EXPECT().Unit(names.NewUnitTag("foo/1")).Return(nil, &params.Error{Code: "not found"})
+
+	mgr, err := relation.NewStateManager(s.mockUnitRW, loggo.GetLogger("test"))
```
benhoyt, line 192: I guess we're not worried about testing that it logged a warning?

benhoyt (approved):
Looks good to me!

worker/uniter/relation/statemanager_test.go
wallyworld: i added an extra check

wallyworld:
$$merge$$

PR #12513, opened 2021-01-14
Minor improvements to scripts/generate-password

barryprice:
scripts/generate-password: Allow the operator to optionally hash a given `oldpassword` instead of generating a random one.

## QA steps

Break a deployed unit by modifying its `apipassword` and `oldpassword` fields in the on-disk `agent-conf` and restarting the agent process.

You'll observe that the agent becomes lost, and the on-disk log notes something like:

```
2021-01-13 02:58:55 ERROR juju.worker.apicaller connect.go:204 Failed to connect to controller: invalid entity name or password (unauthorized access)
```

Next, use this script to either:

1. Generate a new password and hash (the original behaviour):

```
$ go run main.go aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee postgresql/0
oldpassword: kYOOLc2IDyOQn9MMt8N1TN/5
db.units.update({"_id": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee:postgresql/0"}, {$set: {"passwordhash": "IGCgVJSwt8xYTaM0VEq/XtwR"}})
```

or

2. (now available with this change), Just create a new hash for the existing, modified on-disk `oldpassword`:
```
$ go run main.go aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee postgresql/0 examplepassword
oldpassword: examplepassword
db.units.update({"_id": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee:postgresql/0"}, {$set: {"passwordhash": "A4Fha2Ctuq2iGPikTVPQBSPd"}})
```
Doing either of the above and then restarting the agent process again should see the unit recover.

### DIFF
--- a/scripts/generate-password/main.go
+++ b/scripts/generate-password/main.go
@@ -15,7 +15,7 @@ import (
 
 func main() {
 	gnuflag.Usage = func() {
-		fmt.Fprintf(os.Stderr, "Usage: %s [modeluuid agent|--user <username>]\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "Usage: %s <modeluuid> <agent> [<password>] | --user <username>\n", os.Args[0])
 		gnuflag.PrintDefaults()
 	}
 	user := gnuflag.String("user", "", "supply a username to generate a password instead of modeluuid and agent")
@@ -23,6 +23,7 @@ func main() {
 	args := gnuflag.Args()
 	var modelUUID string
 	var agent string
+	var passwd string
 	if *user == "" {
 		if len(args) < 2 {
 			gnuflag.Usage()
@@ -30,10 +31,15 @@ func main() {
 		}
 		modelUUID = args[0]
 		agent = args[1]
-	}
-	passwd, err := utils.RandomPassword()
-	if err != nil {
-		log.Fatal(err)
+		if len(args) > 2 {
+			passwd = args[2]
+		} else {
+			var err error
+			passwd, err = utils.RandomPassword()
+			if err != nil {
+				log.Fatal(err)
+			}
+		}
 	}
 	if *user != "" {
 		salt, err := utils.RandomSalt()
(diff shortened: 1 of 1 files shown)


jameinel:
Old password is meant to be the fallback, where the agent will try to use it to connect, and then if it succeeds, it will then negotiate a new password. Which is way it was used vs apipassword.
We also don't really want to use a generic password, because we have an expectation about the strength of agent passwords. (If you make a password long enough, you don't have to do as much bcrypt testing to avoid possibly hash password smashing.)

So I'd rather not change those bits. But I'd like to understand your needs, and make sure we are covering them.

barryprice:
> Old password is meant to be the fallback, where the agent will try to use it to connect, and then if it succeeds, it will then negotiate a new password. Which is way it was used vs apipassword.
> We also don't really want to use a generic password, because we have an expectation about the strength of agent passwords. (If you make a password long enough, you don't have to do as much bcrypt testing to avoid possibly hash password smashing.)
> 
> So I'd rather not change those bits. But I'd like to understand your needs, and make sure we are covering them.

Ah, thanks for explaining - before I knew this tool existed, I was able to fix the problem by calculating the correct hash for `apipassword` and updating the corresponding mongodb document. It sounds like, if I'd hashed the value of `oldpassword` instead and used that to update the mongodb value, that would have worked too. I think the former is less confusing, but perhaps I'm missing something.

Allowing the operator to supply the current value of `apipassword` (or `oldpassword`) just halves the amount of work required - they now only need to paste the mongodb query, and no longer need to manually modify the corresponding `agent.conf` as well.

barryprice:
Thanks! That makes a lot more sense now.

I've amended the description above, and updated the diff to revert the output field name to `oldpassword`.

hpidcock:
[Land PR "Minor improvements to scripts/generate-password"](https://trello.com/c/b1iQX5pz/1475-land-pr-minor-improvements-to-scripts-generate-password)

jameinel (approved):
I think this still makes it easier to use than what we have, sorry about the delay in responding.

jameinel:
$$merge$$

PR #12510, opened 2021-01-13
[2.9-python3-acceptancetests] fix more python2 to python3 inconsistencies

achilleasa:
This is a followup to #12504 which addresses some of the things that I missed.

PR #12516, opened 2021-01-14
[2.9-python3-acceptancetests] Even more python3 compat fixes for acceptance tests

achilleasa:
Yet another followup to #12504 and #12510.

### DIFF
--- a/acceptancetests/assess_cloud_display.py
+++ b/acceptancetests/assess_cloud_display.py
@@ -42,7 +42,9 @@ def remove_display_attributes(cloud):
     # Delete None values, which are "errors" from parsing the yaml E.g. output
     # can show values which we show to the customers but should actually not
     # parsed and compared
-    for key in cloud.keys():
+    # NOTE(achilleasa): list() makes a copy of the keys to avoid a dict size
+    # changed while iterating error.
+    for key in list(cloud.keys()):
         if cloud[key] is None:
             del cloud[key]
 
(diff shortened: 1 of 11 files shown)


acceptancetests/assess_cloud_display.py
```diff
@@ -58,12 +60,13 @@ def remove_display_attributes(cloud):
 def get_clouds(client):
     cloud_list = yaml.safe_load(client.get_juju_output(
         'clouds', '--format', 'yaml', '--local', include_e=False))
-    for cloud_name, cloud in cloud_list.items():
-        if cloud['defined'] == 'built-in':
-            del cloud_list[cloud_name]
-            continue
+    cloud_list_without_builtins = {
```
achilleasa, line 63: This works different in python3. See comment on L47 about using `list` to get a copy of the keys when mutating the dict while it is being iterated.

acceptancetests/utility.py
```diff
@@ -436,11 +440,19 @@ def assert_dict_is_subset(sub_dict, super_dict):
     :raises JujuAssertionError: when sub_dict items are missing.
     :return: True when when sub_dict is a subset of super_dict
     """
-    if not all(item in super_dict.items() for item in sub_dict.items()):
+    if not is_subset(sub_dict, super_dict):
```
achilleasa, line 443: The original code didn't really work as intended (see checks in assess_storage). This is a recursive variant that does the trick.

SimonRichardson (approved):

achilleasa:
$$merge$$

PR #12519, opened 2021-01-15
[2.9-python3-acceptancetest] Yet even more python3 compat fixes for acceptance tests

achilleasa:
Followup to #12516. Also, fixes an incorrect assertion in `assess_persistent_storage`.

### DIFF
--- a/acceptancetests/assess_persistent_storage.py
+++ b/acceptancetests/assess_persistent_storage.py
@@ -49,7 +49,7 @@ def get_storage_list(client):
         'storage', '--format', 'json', include_e=False)
     # Bug 1708340 https://bugs.launchpad.net/juju/+bug/1708340
     # juju storage --format json output will be empty if no storage exists.
-    if raw_output == b'':
+    if raw_output == '':
         storage_list = storage_output = ''
     else:
         storage_output = json.loads(raw_output)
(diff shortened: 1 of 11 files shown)


acceptancetests/assess_persistent_storage.py
```diff
@@ -165,7 +165,7 @@ def assert_app_removal_msg(client, charm_name):
     # merge_stderr=True is required
     app_removal_output = client.get_juju_output(
         'remove-application', charm_name, '--show-log',
-        include_e=False, merge_stderr=True)
+        include_e=False, merge_stderr=True).decode('utf-8')
```
SimonRichardson, line 168: Maybe put `.decode('utf-8')` into the `get_juju_output` method?

achilleasa: In some cases we pipe that to `json.loads`. It's easier to patch at the call site where we need to run a regex against that.
I think we should just make things work under python3 and slap a big "caveat emptor: use the bash framework for writing new tests" warning in the readme.

SimonRichardson: Fine :+1:

SimonRichardson (approved):

acceptancetests/assess_persistent_storage.py
achilleasa: Hmmm... ok, let me change that everywhere :-(

achilleasa:
$$merge$$

achilleasa:
$$merge$$ (go snap installation failed)

PR #12527, opened 2021-01-19
[Charmhub] - Pass architecture for charmstore charms, allowing migration.

SimonRichardson:
The following ensures that we at least attempt to handle missing
architectual origins in charm-store charms. This is not mandatory and
shouldn't reflect what actually is installed 100%, instead should be an
indicator or what was used at install time.

## QA steps

bundle.yaml:

```yaml
series: focal
applications:
  csubuntu:
    charm: cs:ubuntu-16
    num_units: 1
```

```
$ juju bootstrap localhost dst
$ juju bootstrap localhost src
$ juju add-model move
$ juju deploy bundle.yaml
# once it's settled
$ juju migrate move dst
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1911800

### DIFF
--- a/state/migration_import.go
+++ b/state/migration_import.go
@@ -1455,6 +1455,10 @@ func (i *importer) getApplicationPlatform(a description.Application, url *charm.
 			return platform
 		}
 
+		// Attempting to find the right architecture is quite difficult,
+		// especially if we're in a heterogenous deployment. In that case we'll
+		// most likely guess wrong, so we now use the fact that just select
+		// the first available as that's what they most probably started with.
 		var arch string
 		for _, hw := range hardwareCharacteristics {
 			if hw.Arch == nil {
--- a/state/migration_import.go
+++ b/state/migration_import.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/juju/charm/v8"
+	"github.com/juju/collections/set"
 	"github.com/juju/description/v2"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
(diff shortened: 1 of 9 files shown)


hmlanigan:
Can we resolve this the other way, add a platform to CS applications imported to a 2.9 controller?

The current resolution means that applications deployed from the CharmStore and migrated to 2.9 will have different charm-origins, that applications deployed from the CharmStore directly to a 2.9 model.

It will also break behavior in the charm revision updater which expects CH charms to have platforms.  https://github.com/juju/juju/blob/5036cb3452bab9ea429694a3629f7145d35ca4b3/apiserver/facades/controller/charmrevisionupdater/updater.go#L196

SimonRichardson:
@hmlanigan I don't believe we can, this happens to exported models, way, way before an import ever happens. So either we add this and try and improve the resolution or we just don't send it for CS charms.

**But** I do think that adding an origin when importing is a good idea as well. I'll amend that to this PR

state/migration_import.go
```diff
@@ -1345,20 +1388,99 @@ func deduceCharmOrigin(url *charm.URL) (*CharmOrigin, error) {
...
+}
+
+// Attempt to locate the architecture, if it's found in the URL then use
+// that, otherwise fallback to the arch constraint.
+func (i *importer) getApplicationPlatform(a description.Application, url *charm.URL, units []description.Unit) *Platform {
+	var platform *Platform
+	if url != nil && url.Architecture != "" {
+		platform = &Platform{
+			Architecture: url.Architecture,
+			Series:       url.Series,
```
hmlanigan, line 1441: This only works for 2.9 to 2.9 migrations, migrations from 2.8 may not have urls with series.  Fallback of a.Series?

hmlanigan (commented):
Commented on issues found in QA.  Otherwise looks good.

SimonRichardson:
I should be able to steal it from a unit.

hmlanigan (approved):

hmlanigan:
$$merge$$

PR #12532, opened 2021-01-19
[Charmhub] - Migrate download charm to refresh API

SimonRichardson:
With the up and coming changes to the info API it is required that we
move to refresh API. The following changes are mainly mechanical, in
such that we're calling a different API but the results end up being the
same.

It should be noted that the request via info API, sole purpose is to check if
the entity is a bundle or a charm. Once the refresh API supports bundles
we can remove this block of code.

#### With controller

```sh
$ juju bootstrap lxd test
$ juju download ubuntu
```

#### Without controller

```sh
$ juju download --charmhub-url=https://api.staging.charmcraft.io ubuntu
```

### DIFF
--- a/charmhub/refresh.go
+++ b/charmhub/refresh.go
@@ -127,6 +127,11 @@ type refreshOne struct {
 	instanceKey string
 }
 
+// InstanceKey returns the underlying instance key.
+func (c refreshOne) InstanceKey() string {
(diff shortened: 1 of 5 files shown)


hmlanigan (approved):

hmlanigan:
$$merge$$

PR #12536, opened 2021-01-20
[2.9] Fix panic when serializing Proxier to YAML

achilleasa:
The original implementation of ConnectionProxyInfo() for k8s invoked
k8sproxy.GetControllerProxy which returns (*k8sproxy.Proxy, error) and
returned the result back to the caller after coercing the first value
into a proxy.Proxy interface.

As a result, if k8sproxy.GetControllerProxy returns an error, the nil
k8sproxy.GetControllerProxy value gets converted into a typed nil. Given
the way that k8sproxy.GetControllerProxy is implemented, you can
actually invoke MarshalYAML on a (*k8sproxy.Proxy)(nil) and trigger a
panic (null pointer dereference) when the marshal logic attempts to call
yaml.Marshal(&p.config).

## QA steps

```sh
Run assess_caas_deploy_charms.py on gke
```

## Bug reference
Caught by CI: https://jenkins.juju.canonical.com/job/nw-deploy-bionic-gke/1501/console

### DIFF
--- a/caas/kubernetes/provider/connection.go
+++ b/caas/kubernetes/provider/connection.go
@@ -4,16 +4,27 @@
 package provider
 
 import (
+	"github.com/juju/errors"
+
 	k8sproxy "github.com/juju/juju/caas/kubernetes/provider/proxy"
 	"github.com/juju/juju/proxy"
 )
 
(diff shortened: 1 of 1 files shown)


SimonRichardson (approved):

achilleasa:
$$merge$$

PR #12548, opened 2021-01-22
Wireup parallel actions

wallyworld:
The parallel and execution group attributes are used when running unit and machine actions.

Machine actions are run in parallel by default in separate go routines. The machine agent lock is wired through and a named lock is acquired if needed for non-parallel actions. The lock name is unique to running actions so does not affect the other users of the machine lock.
In addition, the relevant unit tests were reworked to use gomock since this allows much easier testing of the parallel execution.
The worker which runs the actions no longer exits if there's an error beginning or running an action - the action is marked as failed, allowing other actions to still run.

Unit actions also use the parallel and execution group to determine if the machine lock should be acquired. Note that for a given unit agent, actions for that agent are still run sequentially, since the state machine worker picks of the next task (hook, action etc) one at a time. But if units are co-located on the same machine, actions for these different units are no longer restricted to running sequentially globally for the machine.
The uniter changes included having to load the action to run before the execution started, not just pass in the action id. This was so that the action could be inspected to see if the machine lock was needed when setting up the executor.

## QA steps

bootstrap and deploy a simple charm with a couple of actions, one marked as parallel, the other not.
The action script should be something like
`for i in {1..10}; do echo "x"; sleep 1; done`

Add a second unit of the charm to the same machine, eg juju add-unit foo --to 1

Execute parallel actions on the two units

juju run foo/0 --background parallel-action && juju run foo/1 --background parallel-action
juju operations
juju show-task X

The start times of both actions should be the same.
Do the same but with an action without parallel=true in the metadata.
The second action should show a start time after the first one finishes.

Test juju exec similarly - at the moment, all juju exec scripts now run in parallel.

juju exec --machine 0 "for i in {1..10}; do echo "x"; sleep 1; done"  \
&& juju exec --machine 0 "for i in {1..10}; do echo "x"; sleep 1; done"

juju operations --machines 0
juju show-task X


wallyworld:
[wire up parallel actions](https://trello.com/c/ES026LEE/1352-wire-up-parallel-actions)

worker/machineactions/worker.go
```diff
@@ -103,30 +109,59 @@ func (h *handler) Handle(_ <-chan struct{}, actionsSlice []string) error {
...
-		results, err := h.config.HandleAction(action.Name(), action.Params(), action.Parallel(), action.ExecutionGroup())
-		if err != nil {
-			finishErr = h.config.Facade.ActionFinish(actionTag, params.ActionFailed, nil, err.Error())
-		} else {
-			finishErr = h.config.Facade.ActionFinish(actionTag, params.ActionCompleted, results, "")
-		}
-		if finishErr != nil {
-			return errors.Trace(finishErr)
-		}
+		// TODO(wallyworld) - use a thread pool to put a cap on the max number of parallel actions.
```
benhoyt, line 112: You shouldn't need a thread pool for this - this is pretty simple to do with a couple of extra lines of code using a buffered "limiter" channel with a size of the max concurrency you want. See: https://play.golang.org/p/PMW0rQv0wjO

worker/machineactions/worker.go
```diff
@@ -103,30 +109,59 @@ func (h *handler) Handle(_ <-chan struct{}, actionsSlice []string) error {
...
+					group = fmt.Sprintf("%s-%s", group, g)
+					worker = fmt.Sprintf("%s (exec group=%s)", worker, g)
+				}
+				spec := machinelock.Spec{
+					Cancel:  abort,
+					Worker:  worker,
+					Comment: fmt.Sprintf("action %s", action.ID()),
+					Group:   group,
+				}
+				releaser, err := h.config.MachineLock.Acquire(spec)
```
benhoyt, line 144: Is there a timeout waiting for this lock, or is it indefinite? I guess indefinite makes sense in the context of running arbitrary actions?

worker/machineactions/worker.go
```diff
@@ -103,30 +109,59 @@ func (h *handler) Handle(_ <-chan struct{}, actionsSlice []string) error {
...
+				if err != nil {
+					actionErr = errors.Annotatef(err, "could not acquire machine execution lock for exec action %s", action.Name())
+					return
+				}
+				defer releaser()
+			}
+			err = h.config.Facade.ActionBegin(actionTag)
```
benhoyt, line 151: Just curious what this does? There's no equivalent `ActionEnd` operation?

benhoyt (approved):

cmd/juju/action/exec.go
```diff
@@ -92,6 +94,13 @@ instead of the workload. On IAAS models, --operator has no effect.
 Commands run for applications or units are executed in a 'hook context' for
 the unit.
 
+Commands run on machines via the --machine argument are run in parallel by default.
+If you want commands to be run sequentially in order of submission, use --parallel=false.
+Such commands will first acquire a global execution lock on the host machine before running,
```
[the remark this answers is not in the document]
wallyworld: lol, 80 chars in 2021 :-)
Nonetheless, I've trimmed the line lengths.

worker/machineactions/worker.go
wallyworld: Yeah, we don't set a timeout which means the underlying mutex waits indefinitely.

worker/machineactions/worker.go
wallyworld: Good suggestion, done

worker/machineactions/worker.go
wallyworld: ActionFinish() is the equivalent which is called at the end of the go routine

benhoyt:
I just finished some manual QA on this, and couldn't find anything wrong. Some notes:

* You need `--background` on the `juju exec` commands too, right? Otherwise the juju client seems to wait and you can't get parallelism.
* Presumably the order of which action starts first when you run multiple parallel actions or execs with `--background` is not specified? I found sometimes it's the second one you started, sometimes the first.

For the record, I downloaded the [`ubuntu` charm](https://jaas.ai/ubuntu) locally, unzipped it, and added the following `actions.yaml` (from Ian) with three actions and three different combos of parallel/execution-group:

```yaml
hello:
  description: Hello World.
  parallel: true
  execution-group: test
  params:
    who:
      type: string
      description: The person to say hello to.
goodbye:
  description: Goodbye World.
  parallel: true
  params:
    who:
      type: string
      description: The person to say goodbye to.
gday:
  description: Gday World.
  parallel: false
  params:
    who:
      type: string
      description: The person to say gday to.
```

And then copied the following script to `actions/hello`, `actions/goodbye`, and `actions/gday` in the charm dir (also had to `chmod +x` each of those):

```bash
#!/bin/bash
action-set result-map.time-completed="$(date)" result-map.message="Hello $(action-get who)!"
# action-fail "so sorry"
action-log hello sailor
echo Goodbye cruel world
echo This message goes to stderr >&2
for i in {1..10}
do
   action-log "Welcome $i times"
   sleep 1
done

exit 0
```

The `juju run` operations all worked as expected. I also tried concurrent `juju exec`s, with `--parallel=false` to confirm that changed it back to serial mode, and two with `--execution-group=foo` and two with `--execution-group=bar` to ensure that worked correctly. All good!

benhoyt:
One more comment: a maxConcurrency of 100 seems somewhat arbitrary (and high) to me. I guess it depends on whether things are I/O bound vs CPU bound, but what about a default of [`runtime.NumCPU()`](https://golang.org/pkg/runtime/#NumCPU)?

wallyworld:
> One more comment: a maxConcurrency of 100 seems somewhat arbitrary (and high) to me. I guess it depends on whether things are I/O bound vs CPU bound, but what about a default of [`runtime.NumCPU()`](https://golang.org/pkg/runtime/#NumCPU)?

100 is arbitrary yeah. i felt numcpu was too low since i think many run scripts would be I/O bound and limiting to just numcpu would unnecessarily limit concurrency. juju agents themselves run 1000s of goroutines so 100 seemed ok.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
!!build!!

wallyworld:
!!build!!

PR #12552, opened 2021-01-26
Improve error reporting for invalid metrics arguments

flimzy:
When the `collect-metrics` runner returns an error, it's not properly reported. This should fix that.

## QA steps

```sh
juju collect-metrics foo/234
```
The output should be:

```
failed to collect metrics: unit-foo-234 not found
```

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1913063

### DIFF
--- a/cmd/juju/metricsdebug/collectmetrics.go
+++ b/cmd/juju/metricsdebug/collectmetrics.go
@@ -163,7 +163,7 @@ func (c *collectMetricsCommand) Run(ctx *cmd.Context) error {
 	resultChannel := make(chan string, len(runResults))
 	for _, result := range runResults {
 		r := result
-		if r.Error != nil {
+		if err := r.Error; err != nil {
 			fmt.Fprintf(ctx.Stdout, "failed to collect metrics: %v\n", err)
 			resultChannel <- "invalid id"
 			continue


flimzy:
I have not added any tests for this. I attempted to add one [here](https://github.com/juju/juju/blob/develop/cmd/juju/metricsdebug/collectmetrics_test.go#L33), but I was unable to trigger the failure, presumably due to the `runnerClient` instance being used in testing not causing the same error I see when connecting to a k8s controller.

I would appreciate guidance on the proper way to test this, with either a unit test, or an integration test.

cmd/juju/metricsdebug/collectmetrics.go
```diff
@@ -163,7 +163,7 @@ func (c *collectMetricsCommand) Run(ctx *cmd.Context) error {
 	resultChannel := make(chan string, len(runResults))
 	for _, result := range runResults {
 		r := result
-		if r.Error != nil {
+		if err := r.Error; err != nil {
 			fmt.Fprintf(ctx.Stdout, "failed to collect metrics: %v\n", err)
```
achilleasa, line 167: The general assumption is that anything written to STDOUT is meant to be consumed by users/tools so it's a good practice  to dump errors to STDERR.

```suggestion
			_, _ = fmt.Fprintf(ctx.Stderr, "failed to collect metrics: %v\n", err)
```

flimzy: Sounds like something for a separate PR. WDYT?

achilleasa: In my view it does fit the context of this set of changes given that sending the error message to the appropriate sink is definitely an improvement.

flimzy: I can do that, of course.  But it changes the scope of this PR significantly from "Stop ignoring one specific error value" to "send a large number of errors to STDERR instead of STDOUT".

achilleasa (approved):
LGTM

achilleasa:
$$merge$$

PR #12554, opened 2021-01-26
Charmhub expose extras

SimonRichardson:
The following allows for partial refresh requests. This PR is laying out
the groundwork for changes to come for how we resolve charms from
the charmhub API.

The refresh requests will be supported by the charmhub API in that if 
you have a partial platform for OS and Series, passing NA for those will
return an error with suggestions about what is available. We can pass
those to the user when resolving the charm, but for now we won't expose
them in this PR.

The idea with the code is that we don't expose the NA to outside the
library, instead of a series or OS is empty, we then swap it out. This
ensures the purity of the client API.

## QA steps

Tests pass, nothing has changed for Juju internally.

### DIFF
--- a/charmhub/transport/error.go
+++ b/charmhub/transport/error.go
@@ -11,8 +11,9 @@ import (
 
 // APIError represents the error from the CharmHub API.
 type APIError struct {
-	Code    string `json:"code"`
-	Message string `json:"message"`
+	Code    string        `json:"code"`
+	Message string        `json:"message"`
+	Extra   APIErrorExtra `json:"extra"`
 }
 
 // APIErrors represents a slice of APIError's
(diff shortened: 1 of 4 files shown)


charmhub/transport/error.go
```diff
@@ -11,8 +11,9 @@ import (
 
 // APIError represents the error from the CharmHub API.
 type APIError struct {
-	Code    string `json:"code"`
-	Message string `json:"message"`
+	Code    string        `json:"code"`
+	Message string        `json:"message"`
+	Extra   APIErrorExtra `json:"extra"`
```
achilleasa, line 16: Do we need an `omitempty` here?

achilleasa (approved):
LGTM

charmhub/transport/error.go
SimonRichardson: We don't send it, so you don't need to.

SimonRichardson:
$$merge$$

PR #12570, opened 2021-01-29
Fix action CLI display issue (ordering of tasks)

wallyworld:
The juju run CLI commands (actions and exec commands) were still using the over-the-wire params structs when calling the client side API. A side effect of this was that operation IDs were handled as strings (from tags) and so ordering was messed up. For numeric IDs we want natural sort order but we were getting straight alphabetical.

A large part of the fix here is to provide new bespoke structs for the api layer to use with the CLI. These structs use IDs not tags.
Along with the structs, a bunch of unmarshalling helper functions are provided. There are some deprecated structs and methods (already removed in 3.0) used with the legacy action UUIDs which still use params structs in the 2.9 branch.

The functional change is to rely on the fact that there are now IDs and perform a natural sort of the operations.

Tests were updated not to use JujuConnSuite (nor mongo) anymore (use the sub api caller pattern instead).
Plus extra (missing) tests were added for the Run() APIs.

## QA steps

deploy a charm with actions
```
juju deploy foo -n 2
juju exec --machine 0 "pwd"
juju exec --machine 0 "pwd" --background
juju exec --application ubuntu-lite "pwd"
```
exec several more commands to get the operation count > 0

check UX messages for wording missteps
```
juju operations --machines 0
juju operations
```
check ordering and correct display of task IDs, summary, receive names etc
```
juju show-task X --format yaml
```
now run some actions
```
juju run ubuntu-lite/0 hello
juju run ubuntu-lite/0 hello --background
```
again use list-operations and show-task to check output


wallyworld:
[actions/operations CLI API / UX](https://trello.com/c/jSvZWfpY/1355-actions-operations-cli-api-ux)

api/action/types.go
```diff
@@ -0,0 +1,260 @@
...
+type Action struct {
+	ID         string
+	Receiver   string
+	Name       string
+	Parameters map[string]interface{}
+}
+
+// ActionResult is the result of running an action.
+type ActionResult struct {
+	*Action
```
hpidcock, line 50: this is inherently dangerous, if Action is optional, which it looks like it is, embedding Action could cause a programming error where code doesn't nil check the embedded Action, instead might just nil check a ptr to ActionResult.

Two ways to fix this, Action is embedded but not a ptr or Action is a property rather than embedded.

cmd/juju/action/action.go
```diff
@@ -27,40 +27,43 @@ type APIClient interface {
...
 
 	// Operation fetches the operation with the specified id.
-	Operation(id string) (params.OperationResult, error)
+	Operation(id string) (action.Operation, error)
+
+	// WatchActionProgress reports on logged action progress messages.
+	WatchActionProgress(actionId string) (watcher.StringsWatcher, error)
+
+	// These methods are to support legacy actions with UUIDs.
+	// (deprecated and will be removed in Juju 3.0).
```
hpidcock, line 58: Comment each method with this rather than as a block.

cmd/juju/action/cancel.go
```diff
@@ -88,35 +88,36 @@ func (c *cancelCommand) Run(ctx *cmd.Context) error {
...
+		return errors.New("no actions were canceled")
 	}
 
 	type unCanceledAction struct {
-		ActionTag names.ActionTag
-		Result    *params.ActionResult
+		ID     string
+		Result *actionapi.ActionResult
 	}
 	var unCanceledActions []unCanceledAction
```
hpidcock, line 113: Can we please rename this variable. unCanceledActions isn't quite right and just looks wrong.

Perhaps failedCancels.

cmd/juju/action/listoperations.go
```diff
@@ -265,7 +256,14 @@ func (s byId) Swap(i, j int) {
 	s[i], s[j] = s[j], s[i]
 }
 func (s byId) Less(i, j int) bool {
-	return s[i].OperationTag < s[j].OperationTag
+	// We expect IDs to be ints but legacy
+	// action still use UUIDs.
+	id1, err1 := strconv.Atoi(s[i].ID)
```
hpidcock, line 261: Can we make this smarter to separate/group old UUID actions and new monotonic actions.

If _i_ is a UUID and _j_ is a monotonic-id then _i_ is less than _j_.
If _j_ is a UUID and _i_ is a monotonic-id then _j_ is less than _i_.

hpidcock (approved):
Just a few things I'd like addressed. :hamburger:

cmd/juju/action/action.go
wallyworld: ok

cmd/juju/action/cancel.go
wallyworld: ok

cmd/juju/action/listoperations.go
wallyworld: UUID actions were last created in 2.7 and are likely pruned by now, but easy enough to add some extra logic

api/action/types.go
wallyworld: changed to *Action as an attribute

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #12585, opened 2021-02-02
[Charmhub] - Normalize channel

SimonRichardson:
We should always normalize the channel passed into the refresh config.
This is because "latest" is a special keyword and that it can and will
be omitted when doing matches.

The code just cleans up how we construct the params conversion so that
we do it once. It leads to better readability, but more verbose code. I
think this is fine for now.

There is still work to do around picking the right method for
refreshing, but that can come later.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju add-model two --config charmhub-url="https://api.staging.charmhub.io"
$ juju deploy ubuntu
$ juju deploy charmed-kubernetes
```

### DIFF
--- a/apiserver/facades/client/charms/charmhubrepo.go
+++ b/apiserver/facades/client/charms/charmhubrepo.go
@@ -148,6 +148,7 @@ func (c *chRepo) ResolveWithPreferredChannel(curl *charm.URL, origin params.Char
 	// were passed in.  ResolveWithPreferredChannel is called for both charms to be
 	// deployed, and charms which are being upgraded.  Only charms being upgraded
 	// will have an ID and Hash.  Those values should only ever be updated in
+	// chRepro FindDownloadURL.
 	resOrigin := params.CharmOrigin{
 		Source:       origin.Source,
 		Type:         string(entity.Type),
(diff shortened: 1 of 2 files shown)


hmlanigan (commented):
looks good so far in test.  will finalize review after merge conflicts resolved.

apiserver/facades/client/charms/charmhubrepo.go
```diff
@@ -120,19 +140,24 @@ func (c *chRepo) ResolveWithPreferredChannel(curl *charm.URL, origin params.Char
...
-		WithSeries(origin.Series).
-		WithArchitecture(origin.Architecture).
+		WithSeries(input.Platform.Series).
+		WithArchitecture(input.Platform.Architecture).
 		WithRevision(revision)
 
 	// Create a resolved origin.  Keep the original values for ID and Hash, if any
 	// were passed in.  ResolveWithPreferredChannel is called for both charms to be
 	// deployed, and charms which are being upgraded.  Only charms being upgraded
 	// will have an ID and Hash.  Those values should only ever be updated in
```
hmlanigan, line 150: Apparently I left a dangling sentence in the comments, could you please add:
     // chRepro FindDownloadURL.
for me?

hmlanigan (approved):

PR #12588, opened 2021-02-02
Remove arch/os from deploy suggestions on Refresh failure.

hmlanigan:
Reword suggestions for what can be deployed when juju receives a ErrorCodeRevisionNotFound as part of Resolving a CharmHub charm.

Drive by to make the list of Risks public, for use in other places.  Rather than copying over.

## QA steps

QA via the unit tests.  The API responses have changed, so not as easy to reproduce the case where this suggestion is displayed.

### DIFF
--- a/apiserver/facades/client/charms/charmhubrepo.go
+++ b/apiserver/facades/client/charms/charmhubrepo.go
@@ -390,23 +390,35 @@ func matchesType(originType string, expected transport.Type) bool {
 }
 
 func composeSuggestions(releases []transport.Release, origin params.CharmOrigin) []string {
-	var suggestions []string
+	channelSeries := make(map[string][]string, 0)
 	for _, release := range releases {
 		platform := release.Platform
-		arch, os, series := platform.Architecture, platform.OS, platform.Series
+		arch, series := platform.Architecture, platform.Series
 		if arch == "all" {
 			arch = origin.Architecture
 		}
 		if arch != origin.Architecture {
 			continue
 		}
-		if os == "all" {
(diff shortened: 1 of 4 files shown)


apiserver/facades/client/charms/charmhubrepo.go
```diff
@@ -328,23 +328,24 @@ func matchesType(originType string, expected transport.Type) bool {
...
-		}
 		if series == "all" {
 			series = origin.Series
 		}
-		suggestions = append(suggestions, fmt.Sprintf("%s with %s/%s/%s", release.Channel, arch, os, series))
+		channelSeries[release.Channel] = append(channelSeries[release.Channel], series)
+	}
+	var suggestions []string
+	for channel, values := range channelSeries {
+		suggestions = append(suggestions, fmt.Sprintf("%s with %s", channel, strings.Join(values, ", ")))
```
SimonRichardson, line 348: We should sort the order of the suggestions. I'd say put latest first then sort everything after it in a natural ordering.

The tests are failing because this can be any order because of using a map to index channels.

SimonRichardson (changes requested):

apiserver/facades/client/charms/charmhubrepo.go
hmlanigan: updated.

SimonRichardson (approved):

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #12594, opened 2021-02-03
[Charmhub] - `Juju info` updates based on upcoming api changes.

hmlanigan:
The CharmHub Info api endpoint will shortly change what data is available and will be returned.  Make changes to juju info so that they do not create cause failures.  Data previously requested by juju will no longer available to the InfoResponse: Resources in the InfoChannelMap, and download hashes and url.

Also contained are some drive by fixes found while working on the rest:  deleting unused code,using a type more instead of the strings for charmhub response data.

## QA steps

There should be no difference or errors seen in 'juju info' output

### DIFF
--- a/api/charmhub/data.go
+++ b/api/charmhub/data.go
@@ -106,7 +106,6 @@ func convertChannels(in map[string]params.Channel) map[string]Channel {
 			Size:       v.Size,
 			Version:    v.Version,
 			Platforms:  convertPlatforms(v.Platforms),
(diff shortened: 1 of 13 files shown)


SimonRichardson (approved):
Nice changes... I like that transport type has been moved :clap:

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #12598, opened 2021-02-04
Remove capped collections

wallyworld:
Mongo 4.4 doesn't support writes to capped collections with server side transactions.

We use 2 capped collections - txnlogs and model logs. These are switched to be normal collections.
A new state worker is added to prune the txnlogs collection according to the size used when initialising the original capped collection.
A new primary controller worker is added to prune the model logs, using the shared prune logic already in place for actions and status history. The cancel channel is wired through and used in the prune loop. Since the logs collections are in a different database ("logs" vs "juju"), the pruning logic was tweaked to accept an actual collection not just a name.

In addition, the NewModel() method was starting the state workers for its state. But is is unnecessary since the model is used to set up the data model info mongo and then discarded; only state instances created from the state pool need to have the workers started.

Also, the controller config backend now needs the context resources passed in for use with the watch method. This had a little fallout with existing uses of that code.

A subsequent PR will deal with any needed changes upgrading from 2.9.

## QA steps

I bootstrapped with the collection sizes set to the smallest possible value
juju bootstrap --debug --config model-logs-size=1M --config max-txn-log-size=1M --config max-status-history-size=1M --config max-status-history-age=5m --config max-action-results-age=5m

Deploying and relating charms with lots of logging, I still struggled to get the collections to fill up very quickly.
```
juju:PRIMARY> for (var c in stats) { print(stats[c]['ns'] + ": " + getReadableFileSizeString(stats[c]['size']) + " (" + getReadableFileSizeString(stats[c]['storageSize']) + ")"); }
juju.txns: 287.7 kB (144.0 kB)
juju.txns.log: 76.0 kB (60.0 kB)
juju.controllers: 13.4 kB (56.0 kB)
juju.managedStoredResources: 13.1 kB (32.0 kB)
juju.toolsmetadata: 9.9 kB (16.0 kB)
...

use logs
juju:PRIMARY> for (var c in stats) { print(stats[c]['ns'] + ": " + getReadableFileSizeString(stats[c]['size']) + " (" + getReadableFileSizeString(stats[c]['storageSize']) + ")"); }
logs.logs.a4e63657-9009-4714-8e69-ebf48d1cce8b: 279.6 kB (140.0 kB)
logs.logs.8994f686-6d18-4c92-8b6a-8e5bdde1f508: 1.3 kB (16.0 kB)
```

What could be done was to hack the action results prune interval to 1 minute instead of 24 hours. I ran some actions on one of the charms and waited 5 minutes and confirmed that the older results were pruned.
The same for charm status history.

```
machine-0: 12:37:57 INFO juju.state operations age pruning (controller): 3 rows deleted
machine-0: 12:41:57 INFO juju.state statuseshistory age pruning (controller): 9 rows deleted

```


wallyworld:
[mongo 4.4 capped collections](https://trello.com/c/p7Cohsr8/1373-mongo-44-capped-collections)

hpidcock (approved):

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #12607, opened 2021-02-09
Clean up model migration code

SimonRichardson:
The model migration code has a lot of code that is no longer used. It
was commented out 2 years ago and never fixed. It makes it a lot harder
to debug what's actually happening.

We should cut our losses and run if somebody wanted to implement the
broken code they'd probably be better doing it from scratch anyway (sunk
cost fallacy comes into play here).

## QA steps

Follow acceptance tests setup and run the following:

```sh
$ ./assess_model_migration.py lxd $GOPATH/bin/juju /tmp/artifacts
```


achilleasa (approved):

SimonRichardson:
$$merge$$

PR #12609, opened 2021-02-09
EC2 fix panic

SimonRichardson:
This was a fun one to locate. There are two issues here:

 1. Using a slice with holes always leads to bad expectations
 2. Expecting an interface to be backed by one instance seems bad

The following PR ensures that we collect the zones from both `amzInstance`
and `sdkInstance`. As we seem to be transitioning between libraries we
need to ensure that we correctly handle those locations.

There aren't any unit tests in here so I did add some to at least cover
us for those scenarios.

----

I wish there were a better lint guard around this as it's bad
practice to just convert the type without checking.

## QA steps

It expects that you have setup the acceptance tests from the README.

```sh
$ ./assess_model_migration.py aws $GOPATH/bin/juju /tmp/artifacts
```

### DIFF
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -269,16 +269,30 @@ func (e *environ) AvailabilityZones(ctx context.ProviderCallContext) (corenetwor
 func (e *environ) InstanceAvailabilityZoneNames(ctx context.ProviderCallContext, ids []instance.Id) ([]string, error) {
 	instances, err := e.Instances(ctx, ids)
 	if err != nil && err != environs.ErrPartialInstances {
-		return nil, err
+		return nil, errors.Trace(err)
 	}
-	zones := make([]string, len(instances))
-	for i, inst := range instances {
+
+	return gatherAvailabilityZones(instances), nil
+}
+
+func gatherAvailabilityZones(instances []instances.Instance) []string {
+	zones := make([]string, 0)
+	for _, inst := range instances {
 		if inst == nil {
 			continue
 		}
-		zones[i] = inst.(*amzInstance).AvailZone
+		switch t := inst.(type) {
+		case *amzInstance:
+			zones = append(zones, t.AvailZone)
+		case *sdkInstance:
+			if t.i != nil && t.i.Placement != nil && t.i.Placement.AvailabilityZone != nil {
+				zones = append(zones, *t.i.Placement.AvailabilityZone)
+			}
+		default:
+			logger.Tracef("unexpected instance type %T when getting availability zones", inst)
+		}
 	}
-	return zones, err
+	return zones
 }
 
 // DeriveAvailabilityZones is part of the common.ZonedEnviron interface.
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -289,7 +289,7 @@ func gatherAvailabilityZones(instances []instances.Instance) []string {
 				zones = append(zones, *t.i.Placement.AvailabilityZone)
 			}
 		default:
-			logger.Tracef("unexpected instance type %T when getting availability zones", inst)
+			logger.Errorf("unexpected instance type %T when getting availability zones", inst)
 		}
 	}
 	return zones
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -266,7 +266,7 @@ func (e *environ) AvailabilityZones(ctx context.ProviderCallContext) (corenetwor
 
 // InstanceAvailabilityZoneNames returns the availability zone names for each
 // of the specified instances.
-func (e *environ) InstanceAvailabilityZoneNames(ctx context.ProviderCallContext, ids []instance.Id) ([]string, error) {
+func (e *environ) InstanceAvailabilityZoneNames(ctx context.ProviderCallContext, ids []instance.Id) (map[instance.Id]string, error) {
 	instances, err := e.Instances(ctx, ids)
 	if err != nil && err != environs.ErrPartialInstances {
 		return nil, errors.Trace(err)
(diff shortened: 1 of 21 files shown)


provider/ec2/environ.go
```diff
@@ -269,16 +269,30 @@ func (e *environ) AvailabilityZones(ctx context.ProviderCallContext) (corenetwor
...
+	zones := make([]string, 0)
+	for _, inst := range instances {
 		if inst == nil {
 			continue
 		}
-		zones[i] = inst.(*amzInstance).AvailZone
+		switch t := inst.(type) {
+		case *amzInstance:
+			zones = append(zones, t.AvailZone)
+		case *sdkInstance:
```
SimonRichardson, line 287: We sometimes ended up getting this instance and not the `amzInstance` causing it to panic.

provider/ec2/environ.go
```diff
@@ -269,16 +269,30 @@ func (e *environ) AvailabilityZones(ctx context.ProviderCallContext) (corenetwor
...
+			if t.i != nil && t.i.Placement != nil && t.i.Placement.AvailabilityZone != nil {
+				zones = append(zones, *t.i.Placement.AvailabilityZone)
+			}
+		default:
+			logger.Tracef("unexpected instance type %T when getting availability zones", inst)
```
achilleasa, line 292: We should probably log this at an error level or it may break when we change our ec2 deps in the future and we won't be able to tell.

achilleasa (approved):

achilleasa:
!!build!!

provider/ec2/environ.go
```diff
@@ -269,16 +269,30 @@ func (e *environ) AvailabilityZones(ctx context.ProviderCallContext) (corenetwor
...
 			continue
 		}
-		zones[i] = inst.(*amzInstance).AvailZone
+		switch t := inst.(type) {
+		case *amzInstance:
+			zones = append(zones, t.AvailZone)
+		case *sdkInstance:
+			if t.i != nil && t.i.Placement != nil && t.i.Placement.AvailabilityZone != nil {
+				zones = append(zones, *t.i.Placement.AvailabilityZone)
+			}
```
jameinel, line 290: I feel like there should be some sort of log messaging here.

jameinel (commented):
I think this is fallout from Ben's work to start moving away from goaws to the official go aws sdk. 

I am surprised there isn't a lint that checks where you are casting without an ", ok" check.

I am also a little surprised that we aren't always getting sdk instead of amz if that is actually what changed here.

provider/ec2/environ.go
```diff
@@ -269,16 +269,30 @@ func (e *environ) AvailabilityZones(ctx context.ProviderCallContext) (corenetwor
...
+		return nil, errors.Trace(err)
 	}
-	zones := make([]string, len(instances))
-	for i, inst := range instances {
+
+	return gatherAvailabilityZones(instances), nil
+}
+
+func gatherAvailabilityZones(instances []instances.Instance) []string {
+	zones := make([]string, 0)
```
benhoyt, line 279: I think we normally spell this `var zones []string`. See https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices and also:

```
~/w/juju$ rg 'var \w+ \[\]string' | wc -l
439
~/w/juju$ rg ':= make\(\[\]string, 0\)' | wc -l
21
```

benhoyt: That said, I think this change of appending only the non-nil ones breaks the contract in the `ErrPartialInstances`, because in the case where `e.Instances` returns partial instances, the length of the returned slice may be less than the input `ids` slice, and the slots in the input slice won't match up to the slots in the returned slice -- seems important for proper usage.

Here are the doc comments for `ZonedEnviron.InstanceAvailabilityZoneNames` and `Environ.Instances` confirming this:

```
	// InstanceAvailabilityZoneNames returns the names of the availability
	// zones for the specified instances. The error returned follows the same
	// rules as Environ.Instances. [see below]

	// Instances returns a slice of instances corresponding to the
	// given instance ids.  If no instances were found, but there
	// was no other error, it will return ErrNoInstances.  If
	// some but not all the instances were found, the returned slice
	// will have some nil slots, and an ErrPartialInstances error
	// will be returned.
```

Note also that the original version returned `err` at the end (the error from `e.Instances`), not `nil`, so as to pass on `ErrPartialInstances` if necessary. Callers that don't handle partial instances will just have the ordinary check for `err != nil`, and will never look at the slice in the case where it has nils / "holes".

We should probably have a unit test for the partial instances case (thanks for adding those tests, BTW).

provider/ec2/environ_test.go
```diff
@@ -407,3 +409,37 @@ func (*Suite) TestGetValidSubnetZoneMapIntersectionSelectsCorrectIndex(c *gc.C)
...
+					AvailabilityZone: ptrString("bbb"),
+				},
+			},
+		},
+	}
+	az := gatherAvailabilityZones(instances)
+	c.Assert(az, gc.DeepEquals, []string{"aaa", "bbb"})
+}
+
+func ptrString(s string) *string {
```
benhoyt, line 443: This is fine, though note that there's also the SDK function `aws.String` for this.

benhoyt (commented):
Thanks for finding and fixing this, and sorry for the fallout. Previously when we only had `amzInstance` (called `ec2Instance` then) I guess the type assertion without `, ok` was a reasonable thing to do.

I'm surprised we weren't getting this all the time: `InstanceAvailabilityZoneNames` always calls `e.Instances`, which calls `gatherInstances`, which always provides `*sdkInstance`. So it would never have been `*amzInstance` here. Maybe `InstanceAvailabilityZoneNames` always failed?

I've also searched for other occurrences of type assertions to `*amzInstance` or `*sdkInstance`, and that's the only one.

Another approach would be to add a method `AvailabilityZoneName() *string` on each of the two instance types and then `gatherAvailabilityZones` would check for this as an interface check and call it. But on balance I think your approach is simpler and more explicit.

See my note about handling of `ErrPartialInstances` -- I think that's part of the contract here and should be reverted. Was that part of this actually causing problems, or was the panic just from the failed type assertion?

provider/ec2/environ.go
SimonRichardson: Yeah, I hate it. There are so many things wrong with this. I've made a change that will drop this weird slice and favoring a map with `instance.Id` as the key and the zone names as the values. That way it doesn't matter the ordering or if there are missing slots.

Doing the work highlighted that each provider could potentially do this wrong and lose the nuance of the original intention. As highlighted in my original commit.

That problem should be gone for ever.

manadart (commented):
Nice one. Map return makes more sense than a swiss-cheese slice.

provider/ec2/environ.go
SimonRichardson: See #12614

manadart (approved):

SimonRichardson:
$$merge$$

PR #12614, opened 2021-02-10
Errcheck linter fixes

SimonRichardson:
Follow the fallout from #12609 it became apparent that we could employ linters
to help guard and solve against panics at runtime. The problem is that in order to
use the linter we need to enable `errcheck`.

The `errcheck` didn't currently run, so the following effort is to make juju `errcheck`
friendly. The next stage after this is landed is to then enable `check-type-assertions`.

If somebody was so inclined they could make the following run for tests as well!

----

The `errcheck` found a few gotchas in the code base, either we weren't handling
the error correctly or just blindly ignoring it.

As this is an ongoing experiment I've created a new yaml file to allow us to experiment
with linters whilst not breaking the initial setup.

## QA steps

```sh
make static-analysis
```

### DIFF
--- a/tests/suites/static_analysis/lint_go.sh
+++ b/tests/suites/static_analysis/lint_go.sh
@@ -5,6 +5,7 @@ run_go() {
       exit 1
   fi
   golangci-lint run -c .github/golangci-lint.config.yaml
+  golangci-lint run -c .github/golangci-lint.config.experimental.yaml
 }
 
 run_go_tidy() {
--- a/tests/suites/static_analysis/lint_go.sh
+++ b/tests/suites/static_analysis/lint_go.sh
@@ -1,7 +1,7 @@
 run_go() {
   VER=$(golangci-lint --version | tr -s ' ' | cut -d ' ' -f 4 | cut -d '.' -f 1,2)
-  if [ "${VER}" != "1.35" ]; then
-      (>&2 echo -e "\\nError: golangci-lint version does not match 1.35")
+  if [ "${VER}" != "1.36" ]; then
+      (>&2 echo -e "\\nError: golangci-lint version does not match 1.36. Please upgrade/downgrade to the right version.")
       exit 1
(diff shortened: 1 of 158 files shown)


tests/suites/static_analysis/lint_go.sh
```diff
@@ -1,7 +1,7 @@
 run_go() {
   VER=$(golangci-lint --version | tr -s ' ' | cut -d ' ' -f 4 | cut -d '.' -f 1,2)
-  if [ "${VER}" != "1.35" ]; then
-      (>&2 echo -e "\\nError: golangci-lint version does not match 1.35")
+  if [ "${VER}" != "1.36" ]; then
+      (>&2 echo -e "\\nError: golangci-lint version does not match 1.36. Please upgrade/downgrade to the right version.")
```
hmlanigan, line 4: 👍

benhoyt (approved):
Nice. Seems good to me, and caught a couple of real bugs. Does make `defer file.Close()` type of things a bit more annoying (do I log or ignore here?), but the trade-off seems worth it.

tlm:
> Nice. Seems good to me, and caught a couple of real bugs. Does make `defer file.Close()` type of things a bit more annoying (do I log or ignore here?), but the trade-off seems worth it.

Possible we could make some helpers for these situations?

tlm:
Looks good Simon

manadart (approved):
The shear volume of black-holed errors is concerning, particularly ones associated with critical path things like pub-sub and worker start-up.

We've tightened it up some though, and this allows us to ensure that we will not ignore any new errors inadvertently, so it's a win.

SimonRichardson:
$$merge$$

petevg:
Looks like the merge bot fails when we make changes in the .github folder.

Fine to override for now, but we definitely need to investigate if we start having trouble elsewhere.

PR #12617, opened 2021-02-11
Fixes permission for model operator.

tlm:
A bug was introduced in the 2.9 branch where local run juju binaries in
a kube cluster default to using the service account attached to their
pod. This change wasn't ported to the model operator so it was using a
service account without any permissions.

This change give the model operator it's own service account with the
need permissions to run.

## Checklist

## QA steps

The easiest way to verify the fix is bootstrap juju and check the model operator logs for errors. The more thorough test all though not needed is to follow the steps outline in the bug report below.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1915320

### DIFF
--- a/caas/kubernetes/provider/modeloperator.go
+++ b/caas/kubernetes/provider/modeloperator.go
@@ -13,6 +13,7 @@ import (
 	"github.com/juju/names/v4"
 	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
+	rbac "k8s.io/api/rbac/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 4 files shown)


tlm:
[LP: 1915320 fix model operator bug from 2.9 work](https://trello.com/c/L0V5oJDP/1384-lp-1915320-fix-model-operator-bug-from-29-work)

caas/kubernetes/provider/modeloperator.go
```diff
@@ -400,6 +502,127 @@ func modelOperatorService(
...
+		Rules: []rbac.PolicyRule{
+			{
+				APIGroups: []string{"admissionregistration.k8s.io"},
+				Resources: []string{"mutatingwebhookconfigurations"},
+				Verbs: []string{
+					"create",
+					"delete",
+					"get",
+					"list",
+					"update",
```
wallyworld, line 580: "patch" as well?
how do we know we have covered everything in the various verb slices?

caas/kubernetes/provider/modeloperator.go
```diff
@@ -400,6 +502,127 @@ func modelOperatorService(
 	}
 }
 
+func ensureModelOperatorRBAC(
```
wallyworld, line 505: I think we're missing rollback / cleanup of resources if something fails part way through?

wallyworld (approved):
New code seems to fit with what's there already.
A couple of questions in-line.

caas/kubernetes/provider/modeloperator.go
tlm: We don't do any patching at the moment in the current code path.

caas/kubernetes/provider/modeloperator.go
tlm: Yes because we don't expect this to fail. Should we add it as we don't have it around a lot of the other code paths.

wallyworld: The other k8s ensure methods have cleanup so would be good

tlm:
$$merge$$

tlm:
$$merge$$

PR #12625, opened 2021-02-15
When upgrading to juju 3.0, keep using the exiting mongo

wallyworld:
When upgrading to Juju 3.0 from 2.9, there will be an existing mongo installed which needs to be retained.
This PR re-packages previous logic from 2.9 and earlier to perform this check. The core functional change for 3.0 is the snippet in ensureServer() which checks for an older mongo.

## QA steps

boostrap juju 2.6
upgrade all models to 2.9-rc6
then
juju upgrade-controller to this branch

juju should keep running and ssh into the controller machine to see that the system mongo is retained and there's no mongo snap installed

compile a local copy of this branch with SnapTrack=4.4 and SnapRisk=candidate 
rinse and repeat with an initial juju 2.9 bootstrapped to focal (which will use mongo snap 4.0 which should also be retained)

### DIFF
--- a/mongo/legacy.go
+++ b/mongo/legacy.go
@@ -1,0 +1,301 @@
+// Copyright 2021 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package mongo
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+
+	"github.com/juju/errors"
+
+	"github.com/juju/juju/network"
+	"github.com/juju/juju/service"
+	"github.com/juju/juju/service/common"
+)
+
+// TODO(juju4): remove support for upgrading from older mongos
+
+// maybeUseLegacyMongo returns nil if there's a juju that's running
+// on a mongo that's not from the juju-db snap.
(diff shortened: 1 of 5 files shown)


mongo/legacy.go
```diff
@@ -0,0 +1,301 @@
...
+			}
+		}
+		return errors.Trace(err)
+	}
+	// Coming from an early trusty install is not supported since
+	// we now only support wired tiger.
+	if mongodVersion.Major == 2 {
+		return errors.NotSupportedf("mongo %v", mongodVersion)
+	}
+	logVersion(mongoPath)
```
wallyworld, line 44: The logic from here on is essentially copied from what was there in juju 2.9 to support mongo from apt as a systemd service.

wallyworld:
[keep using existing apt mongo when upgrading from 2.9 to 3](https://trello.com/c/jzl7Ju3s/1388-keep-using-existing-apt-mongo-when-upgrading-from-29-to-3)

ycliuhw (approved):
lgtm, thanks

wallyworld:
$$merge$$

PR #12627, opened 2021-02-15
Ensure relation units

SimonRichardson:
The following fixes an issue when a subordinate hasn't been correctly
related on all keys then it would commit suicide. This was caused when
attempting to introduce the error checking linter into juju code base
was incorrectly done (by me!). The fix is simple, only return on an
error, not on all errors or nils.

The REALLY interesting part of this piece of work is that if you return
nil from relationsChanged and you do not iterate over all the keys then
all subordinates will commit suicide.

There are no unit tests for this code, it is one big mudball of
integrations from state and internal state. The only way I caught this
was by inspecting a full integration test and digging deeper.

The architecture of this code is hard to understand and is built with
duct tape, we should revisit at some point.

## QA steps

The following should work and the nrpe unit should start.

```sh
$ juju bootstrap lxd test
$ juju deploy cs:nrpe
$ juju deploy cs:nagios
$ juju deploy cs:ubuntu
$ juju add-relation nrpe:monitors nagios:monitors
```

### DIFF
--- a/worker/uniter/remotestate/watcher.go
+++ b/worker/uniter/remotestate/watcher.go
@@ -826,6 +826,11 @@ func (w *RemoteStateWatcher) leadershipChanged(isLeader bool) {
 func (w *RemoteStateWatcher) relationsChanged(keys []string) error {
 	w.mu.Lock()
 	defer w.mu.Unlock()
+
+	// NOTE (stickupkid): Any return nil or early exit during the iteration of
+	// the keys that is non-exhaustive can cause units (subordinates) to
+	// commit suicide.
+
 	for _, key := range keys {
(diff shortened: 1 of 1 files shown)


worker/uniter/remotestate/watcher.go
```diff
@@ -840,7 +845,9 @@ func (w *RemoteStateWatcher) relationsChanged(keys []string) error {
 		} else if err != nil {
 			return errors.Trace(err)
 		} else {
```
achilleasa, line 847: Maybe merge the `else` and the nested `if err:...`?

achilleasa (approved):

SimonRichardson:
$$merge$$

PR #12634, opened 2021-02-17
Fix uniter loop when a relation is removed out of band

wallyworld:
If a unit agent restarts with a relation hook queued up, but the relation is removed in the interim, the unit agent gets in a restart loop. This PR tweaks the hook prepare and commit operations to allow the hook to be skipped in such cases.

In 2.8, there's no unit tests for the statetracker component - these do exist in 2.9 and the relevant unit tests will be updated there.

## QA steps

deploy a cross model scenario across 2 controllers
eg mariadb and mediawiki
in the offering controller,
juju remove-offer mariadb --force -y
check logs on the consuming model - there should be an initial error but no restart loop of the unit agent

### DIFF
--- a/worker/uniter/relation/statetracker.go
+++ b/worker/uniter/relation/statetracker.go
@@ -16,6 +16,7 @@ import (
 	"github.com/juju/juju/core/life"
 	"github.com/juju/juju/core/relation"
 	"github.com/juju/juju/worker/uniter/hook"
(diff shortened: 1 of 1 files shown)


ycliuhw (approved):
lgtm thanks

wallyworld:
$$merge$$

PR #12638, opened 2021-02-17
Bundle export with channel

SimonRichardson:
Charmhub relies on channels as a way to identify the correct way to
install a given charm, we need to ensure that this is correctly exported
with the bundle.

Interestingly channel was never outputted for charmstore or local
charms, which could have been a cause of potential bugs when to
replicate an existing deployment based on revision alone.

As a driveby:

When deploying charmhub charms, only the application name should be used
to identify the charm (this is actually wrong as well, but more on that
later), so the additional information that's within the URL is only
used to identify a charm and nothing more.

The fix here is that we correctly output just the charm application name
so you can identify it correctly.

---

The real issue is that the application name could have changed since the
charm was originally deployed and instead, we should allow charmhub
identifiers to be used instead. The problem I have with this is that it
becomes very hard to identify what the charm is, we're also leaking out
charmhub implementation details, but that might be an unnecessary evil?

## QA steps

```sh
$ juju bootstrap lxd test
$ juju deploy cs:ubuntu --channel=edge
$ juju deploy ubuntu --channel="latest/candidate"
$ juju export-bundle
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1913631

### DIFF
--- a/apiserver/facades/client/bundle/bundle_test.go
+++ b/apiserver/facades/client/bundle/bundle_test.go
@@ -5,6 +5,7 @@ package bundle_test
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/juju/description/v2"
 	"github.com/juju/names/v4"
@@ -758,7 +759,7 @@ func (s *bundleSuite) TestExportBundleWithApplication(c *gc.C) {
 
 	result, err := s.facade.ExportBundle()
 	c.Assert(err, jc.ErrorIsNil)
-	expectedResult := params.StringResult{nil, `
+	expectedResult := params.StringResult{Result: `
 series: trusty
 applications:
   ubuntu:
@@ -801,7 +802,7 @@ func (s *bundleSuite) TestExportBundleWithTrustedApplication(c *gc.C) {
 
 	result, err := s.facade.ExportBundle()
 	c.Assert(err, jc.ErrorIsNil)
-	expectedResult := params.StringResult{nil, `
+	expectedResult := params.StringResult{Result: `
 series: trusty
 applications:
   ubuntu:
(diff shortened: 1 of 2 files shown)


apiserver/facades/client/bundle/bundle_test.go
```diff
@@ -1868,7 +1889,41 @@ relations:
...
+
+	result, err := s.facade.ExportBundle()
+	c.Assert(err, jc.ErrorIsNil)
+
+	output := `
+series: xenial
+applications:
+  mysql:
+    charm: mysql
+    num_units: 1
```
achilleasa, line 1910: Shouldn't we have a channel here? Or is it empty to indicate "use whatever"? If we do know the channel (even if it's the default one) shouldn't we always include it so that the bundle can be re-deployed as-is (with the same channels) even if the charm authors monkey around with channels between deployments?

SimonRichardson: I'll fix, it's essentially it's the test code.

apiserver/facades/client/bundle/bundle_test.go
```diff
@@ -1158,14 +1166,21 @@ applications:
...
 	series := "xenial"
 	if model.Type() == "caas" {
 		series = "kubernetes"
 	}
+	var charmURL string
+	if strings.HasPrefix(name, "ch:") {
+		charmURL = name
+		name = name[3:]
+	} else {
+		charmURL = "cs:" + name
```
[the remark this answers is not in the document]
SimonRichardson: The problem is the whole premise on a lot of these tests is the global factory for getting a local charm off from the filesystem to represent the charm from the store. It's really nasty. For now I'd leave it and tackle that problem further down the line.

achilleasa (approved):

SimonRichardson:
$$merge$$

PR #12640, opened 2021-02-17
[2.8] Backport python3 acceptancetest changes

achilleasa:
This backports all the python3 compatibility fixes and miscellaneous acceptance test bugfixes from 2.9 to 2.8.

PR #12642, opened 2021-02-17
[2.8] Use python 3.6 compat fix for handling maas cli output

achilleasa:
The original fix for independently handling maas cli stdout/stderr (#12640) only
works on python 3.7 whereas the CI boxes currently use 3.6. This commit
changes the arg list to subprocess.run so that it also works with older
python versions.


petevg (approved):
This looks sensible to me.

achilleasa:
$$merge$$

PR #12643, opened 2021-02-17
Expose charm channel in juju status

SimonRichardson:
The following makes it possible to see the channel of the charm when
deploying a charm. This not only makes it good for charmhub charms, but
should really be done for charmstore charms. Although the channels are
different it highlights there could be an issue if you're dependant on
a given charm being from a given channel.

The code itself is really simple, just push the channel through the
apiserver and let it land in the status.

Originally revision wasn't shown for charmhub charms, but ongoing
discussion states that it probably should.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju deploy ubuntu --channel=latest/edge
```

Output should be the following:

```sh
$ juju status
Model  Controller  Cloud/Region  Version    SLA          Timestamp
three  test        lxd/default   2.9-rc6.5  unsupported  16:33:01Z

App     Version  Status  Scale  Charm   Store     Channel  Rev  OS      Message
ubuntu  16.04    active      1  ubuntu  charmhub  edge      17  ubuntu  ready

Unit       Workload  Agent  Machine  Public address  Ports  Message
ubuntu/0*  active    idle   0        10.132.183.237         ready

0        started  10.132.183.237  juju-9b3551-0  xenial      Running

```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1913631

### DIFF
--- a/apiserver/facades/schema.json
+++ b/apiserver/facades/schema.json
@@ -14543,6 +14543,8 @@
                     "additionalProperties": false,
                     "required": [
                         "charm",
+                        "charm-version",
+                        "charm-profile",
                         "series",
                         "exposed",
                         "life",
@@ -14553,8 +14555,6 @@
                         "meter-statuses",
                         "status",
                         "workload-version",
-                        "charm-version",
-                        "charm-profile",
                         "endpoint-bindings",
                         "public-address"
                     ]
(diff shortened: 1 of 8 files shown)


hmlanigan:
QA is good, works well with charm store charms also

hmlanigan (approved):
LGTM

SimonRichardson:
$$merge$$

PR #12644, opened 2021-02-18
Fix service type translation for k8s applications

wallyworld:
When a k8s app was deployed with a user specified service type, the translation into the k8s ServiceType value was missing. So the pod was incorrectly set up with the juju value and it failed.
This PR fixes that and also adds validation. It also caters for any already deployed applications that may have worked by accident by specifying the k8s value.
The existing unit tests were testing the implemented behaviour, not the correct behaviour.

## QA steps

bootstrap on microk8s
juju deploy ks8-app --config kubernetes-service-type=cluster
kubectl get all to see the service type of the app is ClusterIP

juju deploy ks8-app --config kubernetes-service-type=xxxx
ERROR service type "xxxx" not valid

## Bug reference

https://bugs.launchpad.net/juju/+bug/1914074

### DIFF
--- a/apiserver/facades/client/application/application.go
+++ b/apiserver/facades/client/application/application.go
@@ -563,6 +563,14 @@ func caasPrecheck(
 			return errors.Errorf("block storage %q is not supported for k8s charms", s.Name)
 		}
 	}
+	serviceType, ok := args.Config[k8s.ServiceTypeConfigKey]
+	if ok {
+		switch caas.ServiceType(serviceType) {
+		case caas.ServiceCluster, caas.ServiceLoadBalancer, caas.ServiceExternal, caas.ServiceOmit:
+		default:
+			return errors.NotValidf("service type %q", serviceType)
+		}
+	}
 
 	cfg, err := model.ModelConfig()
 	if err != nil {
--- a/apiserver/facades/client/application/application.go
+++ b/apiserver/facades/client/application/application.go
@@ -563,13 +563,9 @@ func caasPrecheck(
 			return errors.Errorf("block storage %q is not supported for k8s charms", s.Name)
 		}
 	}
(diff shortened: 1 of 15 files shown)


wallyworld:
[incorrect k8s service type](https://trello.com/c/de3wHnDg/1394-incorrect-k8s-service-type)

caas/kubernetes/provider/k8s.go
```diff
@@ -1794,6 +1794,14 @@ func (k *kubernetesClient) deleteVolumeClaims(appName string, p *core.Pod) ([]st
...
+	// There was no validation at deploy so the raw k8s types may have been used.
+	switch string(in) {
+	case string(core.ServiceTypeClusterIP),
+		string(core.ServiceTypeLoadBalancer),
+		string(core.ServiceTypeExternalName):
+		return core.ServiceType(in), nil
+	}
 	serviceType := defaultServiceType
 	if in != "" {
 		switch in {
```
ycliuhw, line 1807: Could we grab validation a new function, then it could be used for apiserver/facades/client/application/application.go

ycliuhw (approved):

apiserver/facades/client/application/application.go
```diff
@@ -563,6 +563,14 @@ func caasPrecheck(
 			return errors.Errorf("block storage %q is not supported for k8s charms", s.Name)
 		}
 	}
+	serviceType, ok := args.Config[k8s.ServiceTypeConfigKey]
+	if ok {
```
ycliuhw, line 567: Here

caas/kubernetes/provider/k8s.go
wallyworld: added upgrade step so we can now reuse this method without the fallback

wallyworld:
$$merge$$

PR #12647, opened 2021-02-18
[2.8] Fix acceptance test bug in get_uptime

achilleasa:
The original regex was too greedy and would fail to match the uptime
value thus causing get_uptime to return 0. As a result, the
assess_container_networking test assumed that the controller machine had
not rebooted yet and tried to block until the machine's ssh port became
unavailble (which never did thus causing the test to timeout).

## QA steps

```console
$ cd acceptancetests
$ ./assess_container_networking.py maas `which juju` /tmp/art nw-nethealth --series bionic  --machine-type=lxd --space-constraint=space1; echo "EXIT $?"
```

### DIFF
--- a/acceptancetests/assess_container_networking.py
+++ b/acceptancetests/assess_container_networking.py
@@ -318,7 +318,7 @@ def _assess_container_networking(client, types, hosts, containers):
 
 
 def get_uptime(client, host):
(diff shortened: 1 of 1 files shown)


SimonRichardson (approved):

achilleasa:
$$merge$$

achilleasa:
$$merge$$

PR #12677, opened 2021-02-24
Remove all model security groups on destroy

wallyworld:
Earlier version of Juju did not always delete all EC2 security groups when destroying a model.
This has since been fixed, but there's a possibility orphaned security groups can still occur, according to the bug report below.

This PR add 2 enhancements to EC2 to make things more robust:
- on destroy model, query all security groups tagged against the model and delete them (not just the model group, also include any others that may have been orphaned when the corresponding instance was deleted)
- log warnings if an instance is terminated but does not appear in the expected state (this is a possible reason why subsequent deletion of the security group may be skipped)

Also do little error message cleanup to remove some stuttering.

## QA steps

deploy an EC2 model with a few charms
remove an app and verify that the instance security groups are deleted
destroy the model and verify that all model tagged security groups are deleted
destroy the controller and ensure all remaining security groups are deleted

## Bug reference

https://bugs.launchpad.net/juju/+bug/1720571

### DIFF
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -1485,7 +1485,7 @@ func (e *environ) AdoptResources(ctx context.ProviderCallContext, controllerUUID
 	if err != nil {
 		return errors.Trace(err)
 	}
-	groupIds, err := e.modelSecurityGroupIDs(ctx)
+	groups, err := e.modelSecurityGroups(ctx)
 	if err != nil {
 		return errors.Trace(err)
 	}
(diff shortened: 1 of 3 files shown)


ycliuhw (approved):
lgtm, ty

wallyworld:
$$merge$$

wallyworld:
[ec2 security group deletion](https://trello.com/c/KiZO9avu/1410-ec2-security-group-deletion)

PR #12680, opened 2021-02-24
Better help message for existing applications

SimonRichardson:
When deploying an existing application again via `juju deploy` the help
message is a bit sad, as it doesn't explain how you get out of that
position and move forward.

The following aim so to do that by using an error type all the way from
state and catching that error code in the charm deployer.

Interestingly the state tests don't change as we're backwards compatible
with the old error message (nice!) and the charm one isn't exactly easy
to trigger.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju deploy mysql
$ juju deploy mysql
Located charm "mysql" in charm-hub, revision 58
Deploying "mysql" from charm-hub charm "mysql", revision 58 in channel stable
ERROR cannot add application "mysql": application already exists:
deploy application using an alias name, or use remove-application to remove the existing one and try again
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1915223

### DIFF
--- a/cmd/juju/application/deployer/charm.go
+++ b/cmd/juju/application/deployer/charm.go
@@ -22,6 +22,7 @@ import (
 	commoncharm "github.com/juju/juju/api/common/charm"
 	"github.com/juju/juju/api/resources/client"
 	app "github.com/juju/juju/apiserver/facades/client/application"
+	"github.com/juju/juju/apiserver/params"
 	"github.com/juju/juju/cmd/juju/application/store"
 	"github.com/juju/juju/cmd/juju/application/utils"
 	"github.com/juju/juju/cmd/juju/common"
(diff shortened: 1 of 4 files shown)


cmd/juju/application/deployer/charm.go
```diff
@@ -252,7 +253,20 @@ func (d *deployCharm) deploy(
...
 		EndpointBindings: d.bindings,
 	}
-	return errors.Trace(deployAPI.Deploy(args))
+
+	err = deployAPI.Deploy(args)
+	if err == nil {
+		return nil
+	}
+
+	if pErr, ok := errors.Cause(err).(*params.Error); ok {
```
achilleasa, line 262: How about moving this check into the client and have that return an `errors.AlreadyExists` that can be tested for and wrapped here? I would prefer we didn't deal with `params.Error` at this level.

SimonRichardson: Sure makes sense

cmd/juju/application/deployer/charm.go
```diff
@@ -259,11 +258,8 @@ func (d *deployCharm) deploy(
 		return nil
 	}
 
-	if pErr, ok := errors.Cause(err).(*params.Error); ok {
-		switch pErr.Code {
-		case params.CodeAlreadyExists:
-			return errors.Wrapf(err, errors.Errorf("\ndeploy application using an alias name, or use remove-application to remove the existing one and try again"), err.Error())
-		}
+	if errors.IsAlreadyExists(err) {
+		return errors.Wrapf(err, errors.Errorf("\ndeploy application using an alias name, or use remove-application to remove the existing one and try again"), err.Error())
```
achilleasa, line 262: Minor nit; I think you can get rid of the `\n` so it reads as one line after the `:` when the error is printed out.

achilleasa (approved):
LGTM; just a minor nit for the line-feed in the error (feel free to ignore)

cmd/juju/application/deployer/charm.go
SimonRichardson: The error message is huge otherwise...

SimonRichardson:
$$merge$$

PR #12681, opened 2021-02-24
[2.8] Retry ec2 security group lookups and improve instance lookup warning message

achilleasa:
This PR wraps the security group lookup code in a retry loop so that we can work around the eventual consistency model that applies to most/all EC2 resources.

In addition, the PR rephrases the warning message emitted while attempting to look up the instance information for an instance that was just created (so we can grab the volume ID of the root disk) as EC2 eventual consistency can cause the EC2 API to initially return a `NotFound` error. To avoid user confusion, the message is now logged at DEBUG level and the retry loop gets **immediately** aborted if any **other** error type is reported.

Finally, the PR addresses a go 1.15 compilation warning wrt. rune concatenation.

For more information on EC2 eventual consistency see: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html#eventual-consistency

## QA steps

```sh
$ juju bootstrap aws test --debug --no-gui
# Look for a message like:
DEBUG: instance i-0e043391f669e3b18 is not available yet; retrying fetch of instance information
$ juju add-machine

# Watch status until the machine is ready
$ watch juju status
```

## Bug reference
- https://bugs.launchpad.net/juju/+bug/1915057
- https://bugs.launchpad.net/juju/+bug/1914098

### DIFF
--- a/provider/ec2/environ.go
+++ b/provider/ec2/environ.go
@@ -2124,19 +2124,39 @@ var zeroGroup ec2.SecurityGroup
 // groupName or with filter by vpc-id and group-name, depending on whether
 // vpc-id is empty or not.
 func (e *environ) securityGroupsByNameOrID(groupName string) (*ec2.SecurityGroupsResp, error) {
+	var (
+		groups []ec2.SecurityGroup
+		filter *ec2.Filter
+	)
+
 	if chosenVPCID := e.ecfg().vpcID(); isVPCIDSet(chosenVPCID) {
 		// AWS VPC API requires both of these filters (and no
 		// group names/ids set) for non-default EC2-VPC groups:
-		filter := ec2.NewFilter()
+		filter = ec2.NewFilter()
 		filter.Add("vpc-id", chosenVPCID)
 		filter.Add("group-name", groupName)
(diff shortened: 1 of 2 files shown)


provider/ec2/environ.go
```diff
@@ -976,13 +976,20 @@ func tagRootDisk(e *ec2.EC2, ctx context.ProviderCallContext, tags map[string]st
...
+			// should retry until it appears or we run out of
+			// attempts.
+			if strings.HasSuffix(ec2ErrCode(err), ".NotFound") {
+				logger.Debugf("instance %v is not available yet; retrying fetch of instance information", inst.InstanceId)
+				continue
 			}
-			logger.Infof("retrying fetch of instances")
-			continue
+
+			// No need to retry for other error types.
```
achilleasa, line 988: The original implementation _retried anyway_ but failing for other kinds of errors seems more reasonable to me, especially if we get something like an expired credential error.

@jameinel Is there any particular reason why this was written this way initially?

wallyworld (approved):

wallyworld:
$$merge$$

PR #12682, opened 2021-02-24
Ensure we use the same user-agent version

SimonRichardson:
The user-agent we send from Juju differs from client to client. The
following starts a trend to change that so that we're consistent with
how we're querying the APIs.

This becomes apparent when the charmstore shim API has a different
User-Agent version from the same Juju version. This is a bit weird and
inconsistent.

## QA steps

To really ensure that we're hitting the API with the right User-Agent, you
can either use Wireshark, but certs become an issue, or you can use this
very simple hack for now.

```sh
$ juju bootstrap lxd test --config="charmstore-url=http://0.0.0.0:8080"
$ juju status -m controller
$ lxc exec <name> bash
$ cat > server.py
#!/usr/bin/env python3
import http.server as SimpleHTTPServer
import socketserver as SocketServer
import logging

PORT = 8080

class GetHandler(
        SimpleHTTPServer.SimpleHTTPRequestHandler
        ):

    def do_GET(self):
        logging.error(self.headers)
        SimpleHTTPServer.SimpleHTTPRequestHandler.do_GET(self)


Handler = GetHandler
httpd = SocketServer.TCPServer(("", PORT), Handler)

httpd.serve_forever()
$ chmod +x server.py
$ ./server.py
$ juju deploy cs:mysql
ERROR:root:Host: 0.0.0.0:8080
User-Agent: Juju/2.9-rc6
Bakery-Protocol-Version: 3
Cookie:
Accept-Encoding: gzip
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1913754

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -360,6 +360,8 @@ github.com/juju/charm/v8 v8.0.0-20210115142305-1eb0c22cff4e h1:qIK2VhlqGdlQyp3kI
 github.com/juju/charm/v8 v8.0.0-20210115142305-1eb0c22cff4e/go.mod h1:3gMi15p+v4CxEquVduhcOoPLWM7IaRB+OZB6bT3Glww=
 github.com/juju/charmrepo/v6 v6.0.0-20201118043529-e9fbdc1a746f h1:esltimJsCcUSaC9ahxBpC70Gumqnnmgm0Ah+BLVQBTY=
 github.com/juju/charmrepo/v6 v6.0.0-20201118043529-e9fbdc1a746f/go.mod h1:4V0vrJRD/0oxG+D9j/53elHpXiZ1FoBIXdJGm3Jq4Js=
+github.com/juju/charmrepo/v6 v6.0.0-20210224160253-14d76381037d h1:42OeuoFgQht+ddnFb0FyBVWDPoKaQbFqZr39HsECBZE=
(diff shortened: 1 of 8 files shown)


go.sum
```diff
@@ -360,6 +360,8 @@ github.com/juju/charm/v8 v8.0.0-20210115142305-1eb0c22cff4e h1:qIK2VhlqGdlQyp3kI
 github.com/juju/charm/v8 v8.0.0-20210115142305-1eb0c22cff4e/go.mod h1:3gMi15p+v4CxEquVduhcOoPLWM7IaRB+OZB6bT3Glww=
 github.com/juju/charmrepo/v6 v6.0.0-20201118043529-e9fbdc1a746f h1:esltimJsCcUSaC9ahxBpC70Gumqnnmgm0Ah+BLVQBTY=
 github.com/juju/charmrepo/v6 v6.0.0-20201118043529-e9fbdc1a746f/go.mod h1:4V0vrJRD/0oxG+D9j/53elHpXiZ1FoBIXdJGm3Jq4Js=
+github.com/juju/charmrepo/v6 v6.0.0-20210224160253-14d76381037d h1:42OeuoFgQht+ddnFb0FyBVWDPoKaQbFqZr39HsECBZE=
```
achilleasa, line 363: There is something fishy going on with the lines above the addition...

achilleasa (approved):
Curious about the contents of go.sum but other than that it LGTM

SimonRichardson:
$$merge$$

PR #12684, opened 2021-02-24
[2.9] Improve remove cloud output

achilleasa:
When attempting to remove a cloud, juju would treat unknown clouds differently for the client and the controller. More specifically, if the cloud is unknown to the **client**, an INFO message is displayed and no error is reported. On the other hand, if the cloud is unknown to the **controller**, we get a NotFound error back which the client emits and subsequently proceeds to exit the command with a **1** status code.

This can be quite confusing for users in the case where the cloud is only known by the client and not the controller where the juju CLI would emit both a success and an error message (see bug details).

This PR attempts to fix this by checking the error type returned by the controller and ensuring that the client emits a suitable INFO message when the cloud is not known to the controller.

## QA steps

```console

$ juju remove-cloud bogus --client --controller test
No cloud called "bogus" exists on this client
No cloud called "bogus" exists on controller "test"

# Add an lxd cloud to the client, controller and both (just point to your lxd's endpoint)
$ juju add-cloud client-only --client
$ juju add-cloud controller-only --controller test
$ juju add-cloud both --client --controller test

# Remove each one of the clouds and ensure that the output makes sense
$ juju remove-cloud client-only --client --controller test
$ juju remove-cloud controller-only --client --controller test 
$ juju remove-cloud both --client --controller test
```

## Documentation changes

*Please replace with any notes about how it affects current user workflow, CLI, or API.*

## Bug reference
https://bugs.launchpad.net/juju/+bug/1914027

### DIFF
--- a/api/cloud/cloud.go
+++ b/api/cloud/cloud.go
@@ -4,6 +4,8 @@
 package cloud
 
 import (
+	"strings"
+
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	"github.com/juju/names/v4"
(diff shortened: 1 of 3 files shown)


api/cloud/cloud.go
```diff
@@ -373,7 +375,20 @@ func (c *Client) RemoveCloud(cloud string) error {
 	if err != nil {
 		return errors.Trace(err)
 	}
-	return result.OneError()
+
+	if err = result.OneError(); err == nil {
+		return nil
+	}
+	if pErr, ok := errors.Cause(err).(*params.Error); ok {
```
achilleasa, line 382: @SimonRichardson I shamelessly "borrowed" this code from your PR ;-)

hmlanigan (approved):
QA went well.   I have a related drive-by suggestion for you.  The output of the QA commands was inconsistent  using both "the client" and "this client":

No cloud called "controller-only" exists on this client
Removed details of cloud "both" from the client

Let's choose one option.

wallyworld:
$$merge$$

PR #12691, opened 2021-02-25
Add coverage to make file

SimonRichardson:
The following introduces code coverage to the tests, this should allow
us to see what's missing and what we can remove in terms of tests
without loss to the project.
## QA steps

```sh
COVERAGE_CHECK=1 make run-tests
```

### DIFF
--- a/.github/workflows/client-tests.yml
+++ b/.github/workflows/client-tests.yml
@@ -98,6 +98,10 @@ jobs:
       if: (matrix.os == 'ubuntu-latest')
       run: |
         # Jenkins can perform the full jujud testing.
-        go test -v ./cmd/juju/... -check.v
+        go test -v ./cmd/juju/... -check.v -coverprofile=coverage.txt -covermode=atomic
         go test -v ./cmd/plugins/... -check.v
       shell: bash
+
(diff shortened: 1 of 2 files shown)


.github/workflows/client-tests.yml
```diff
@@ -98,6 +98,10 @@ jobs:
       if: (matrix.os == 'ubuntu-latest')
       run: |
         # Jenkins can perform the full jujud testing.
-        go test -v ./cmd/juju/... -check.v
+        go test -v ./cmd/juju/... -check.v -coverprofile=coverage.txt -covermode=atomic
```
achilleasa, line 101: Have we verified that running this both as a GH action and on jenkins for the same SHA does not clobber the coverage results?

SimonRichardson: Not yet, but consider this is only an experiment, we can fix that if we move it to jenkins.

achilleasa (approved):
Go for it

SimonRichardson:
$$merge$$

petevg:
Merging, because the failed merge seems to be a bug in the plugin we're using rather than a defect in the code.

PR #12692, opened 2021-02-25
Fix internal microk8s patch

ycliuhw:
Fix internal microk8s patch for assess_caas_deploy_kubeflow because we run these tests on ephemeral nodes on public cloud now;


acceptancetests/jujupy/k8s_provider/microk8s.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, bs_manager, cluster_name=None, timeout=1800):
         self.default_storage_class_name = 'microk8s-hostpath'
 
     def _ensure_cluster_stack(self):
```
ycliuhw, line 52: we built microk8s in AMI for ephemeral nodes now

wallyworld (approved):

PR #12694, opened 2021-02-25
[2.8] Release openstack machine before attempting to start a new instance after a timeout

achilleasa:
Prior to this change, if the openstack provider got a timeout while waiting for a machine to finish building it would not terminate the instance before retrying the StartInstance attempt.

This PR ensures that those machines get terminated before `StartInstance` returns with an error.

## QA steps
As this issue manifests under high load you will need to try deploying something to a loaded openstack instance.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1914829

### DIFF
--- a/provider/openstack/provider.go
+++ b/provider/openstack/provider.go
@@ -1236,8 +1236,9 @@ func (e *Environ) startInstance(
 				// If we got an error back (eg. StillBuilding)
 				// we need to terminate the instance before
 				// retrying to avoid leaking resources.
+				logger.Warningf("Unable to retrieve details for created instance %q: %v; attempting to terminate it", server.Id, err)
 				if termErr := e.terminateInstances(ctx, []instance.Id{instance.Id(server.Id)}); termErr != nil {
-					logger.Debugf("Failed to delete instance in BUILD state, %q", termErr)
+					logger.Errorf("Failed to delete instance %q: %v; manual cleanup required", server.Id, termErr)
 				}
 				server = nil
 				break
@@ -1254,7 +1255,7 @@ func (e *Environ) startInstance(
 				}
 				logger.Infof("Deleting instance %q in ERROR state%v", server.Id, faultMsg)
 				if err = e.terminateInstances(ctx, []instance.Id{instance.Id(server.Id)}); err != nil {
-					logger.Debugf("Failed to delete instance in ERROR state, %q", err)
+					logger.Errorf("Failed to delete instance in ERROR state %q: %v; manual cleanup required", server.Id, err)
 				}
 				server = nil
 				err = errors.New(faultMsg)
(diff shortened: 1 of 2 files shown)


provider/openstack/provider.go
```diff
@@ -1233,6 +1233,12 @@ func (e *Environ) startInstance(
 			var serverDetail *nova.ServerDetail
 			serverDetail, err = waitForActiveServerDetails(client, server.Id, 5*time.Minute)
 			if err != nil || serverDetail == nil {
+				// If we got an error back (eg. StillBuilding)
+				// we need to terminate the instance before
+				// retrying to avoid leaking resources.
+				if termErr := e.terminateInstances(ctx, []instance.Id{instance.Id(server.Id)}); termErr != nil {
+					logger.Debugf("Failed to delete instance in BUILD state, %q", termErr)
```
achilleasa, line 1240: This is best effort so we just log the termination error (if any) and allow the original error (`err`) to be returned by this function.

achilleasa:
!!build!! (timeout starting build)

provider/openstack/provider.go
```diff
@@ -1233,6 +1233,12 @@ func (e *Environ) startInstance(
 			var serverDetail *nova.ServerDetail
 			serverDetail, err = waitForActiveServerDetails(client, server.Id, 5*time.Minute)
 			if err != nil || serverDetail == nil {
+				// If we got an error back (eg. StillBuilding)
+				// we need to terminate the instance before
+				// retrying to avoid leaking resources.
+				if termErr := e.terminateInstances(ctx, []instance.Id{instance.Id(server.Id)}); termErr != nil {
```
hmlanigan, line 1239: Suggest adding an Info to day we're deleting the instance, like ln 1255.

provider/openstack/provider.go
hmlanigan, line 1240: Suggest including server.ID, to aid in cleanup on failure.

It's not necessarily true that the machine is in the BUILD state here.

hmlanigan (approved):
Very difficult to QA without an o7k under heavy load available.  The modified unit test does fail without the provider.go change, and passes with it.

Noted a few logging suggestions

provider/openstack/provider.go
achilleasa: Good idea. If you get here the machine is either in the BUILD state and we timed out or we got an API error. I will revise the message tomorrow and remove the BUILD part.

provider/openstack/provider.go
hmlanigan: Do any of the provider's handle that case? 

The instance is leaked, as the instance id isn't saved until after it leaves the provider.

achilleasa:
$$merge$$

PR #12700, opened 2021-02-26
[2.9] Revert back to summarized "juju help" output

achilleasa:
At some point, the `juju help` command would output the basic juju commands and prompt the user to run `juju help commands` to get the full list of commands. At some point this functionality was removed in favor of listing the _full_ set of commands when running `juju help`.

This PR:
- switches back to a short help text variant
- revises the help text (put k8s front and center, clean up provider list and reference juju.is for docs) 
- re-arranges the start command list so the order reflects the flow of commands that a new Juju user would execute when getting started with Juju. 

Additionally, the help text for the REPL has been embedded into each own section (the how to use the REPL details have been omitted as they will be rendered when entering REPL-mode) and the show-full-command-list instructions have also been moved into their own dedicated section (at the end) to make them clearly visible.

Finally, the PR fixes a small bug in the REPL implementation as a drive-by.

## QA steps

```console
# Try "juju help", "juju -h", "juju --help"
$ juju help
Juju provides easy, intelligent application orchestration on top of Kubernetes,
cloud infrastructure providers such as Amazon, Google, Microsoft, Openstack,
MAAS (and more!), or even your local machine via LXD.

See https://juju.is for getting started tutorials and additional documentation.

Starter commands:

    bootstrap           Initializes a cloud environment.
    add-model           Adds a hosted model.
    deploy              Deploys a new application.
    status              Displays the current status of Juju, applications, and units.
    add-unit            Adds extra units of a deployed application.
    relate              Adds a relation between two applications.
    expose              Makes an application publicly available over the network.
    models              Lists models a user can access on a controller.
    controllers         Lists all controllers.
    whoami              Display the current controller, model and logged in user name.
    switch              Selects or identifies the current controller and model.
    add-k8s             Adds a k8s endpoint and credential to Juju.
    add-cloud           Adds a user-defined cloud to Juju.
    add-credential      Adds or replaces credentials for a cloud.

Interactive mode:

When run without arguments, Juju will enter an interactive shell which can be
used to run any Juju command directly.

Help commands:

    juju help           This help page.
    juju help <command> Show help for the specified command.

For the full list of supported commands run:

    juju help commands

$ juju help commands
....
1 billion lines of output
...

# List commands in REPL
$ juju
Welcome to the Juju interactive shell.
Type "help" to see a list of available commands.
Type "q" or ^D or ^C to quit.

no controllers registered$ help
...
1 billion lines of output
....
```

## Bug reference
https://bugs.launchpad.net/juju/+bug/1912760

### DIFF
--- a/cmd/juju/commands/main.go
+++ b/cmd/juju/commands/main.go
@@ -63,41 +63,41 @@ func init() {
 //  Main
 
 var jujuDoc = `
-Juju provides easy, intelligent application orchestration on top of cloud
-infrastructure providers such as Amazon EC2, MaaS, OpenStack, Windows, Azure,
-or your local machine.
+Juju provides easy, intelligent application orchestration on top of Kubernetes,
+cloud infrastructure providers such as Amazon, Google, Microsoft, Openstack,
+MAAS (and more!), or even on your local machine via LXD.
 
-See https://discourse.jujucharms.com/ for ideas, documentation and FAQ.
+See https://juju.is for getting started tutorials and additional documentation.
 `
 
 var usageHelp = `
-Usage: juju [help] <command>
+Juju provides easy, intelligent application orchestration on top of Kubernetes,
+cloud infrastructure providers such as Amazon, Google, Microsoft, Openstack,
+MAAS (and more!), or even your local machine via LXD.
 
-Summary:
-Juju is model & application management software designed to leverage the power
-of existing resource pools, particularly cloud-based ones. It has built-in
-support for cloud providers such as Amazon EC2, Google GCE, Microsoft
-Azure, OpenStack, and Rackspace. It also works very well with MAAS and
-LXD. Juju allows for easy installation and management of workloads on a
-chosen resource pool.
+See https://juju.is for getting started tutorials and additional documentation.
 
-See https://jujucharms.com/docs/stable/help for documentation.
+Starter commands:
 
-Common commands:
-
-    add-cloud           Adds a user-defined cloud to Juju.
-    add-credential      Adds or replaces credentials for a cloud.
-    add-model           Adds a hosted model.
-    add-relation        Adds a relation between two applications.
-    add-unit            Adds extra units of a deployed application.
-    add-user            Adds a Juju user to a controller.
+    add-k8s             Adds a k8s endpoint and credential to Juju.
     bootstrap           Initializes a cloud environment.
-    controllers         Lists all controllers.
+    add-model           Adds a hosted model.
     deploy              Deploys a new application.
+    status              Displays the current status of Juju, applications, and units.
+    add-unit            Adds extra units of a deployed application.
+    relate              Adds a relation between two applications.
     expose              Makes an application publicly available over the network.
     models              Lists models a user can access on a controller.
-    status              Displays the current status of Juju, applications, and units.
+    controllers         Lists all controllers.
+    whoami              Display the current controller, model and logged in user name. 
     switch              Selects or identifies the current controller and model.
+    add-cloud           Adds a user-defined cloud to Juju.
+    add-credential      Adds or replaces credentials for a cloud.
+
+REPL mode:
+
+When run without arguments, Juju will enter an interactive shell which can be
+used to run any Juju command directly.
 
 Help commands:
     
--- a/cmd/juju/commands/main.go
+++ b/cmd/juju/commands/main.go
@@ -70,6 +70,44 @@ or your local machine.
 See https://discourse.jujucharms.com/ for ideas, documentation and FAQ.
 `
 
+var usageHelp = `
+Usage: juju [help] <command>
+
+Summary:
+Juju is model & application management software designed to leverage the power
+of existing resource pools, particularly cloud-based ones. It has built-in
+support for cloud providers such as Amazon EC2, Google GCE, Microsoft
+Azure, OpenStack, and Rackspace. It also works very well with MAAS and
+LXD. Juju allows for easy installation and management of workloads on a
+chosen resource pool.
+
+See https://jujucharms.com/docs/stable/help for documentation.
+
+Common commands:
+
+    add-cloud           Adds a user-defined cloud to Juju.
+    add-credential      Adds or replaces credentials for a cloud.
+    add-model           Adds a hosted model.
+    add-relation        Adds a relation between two applications.
+    add-unit            Adds extra units of a deployed application.
+    add-user            Adds a Juju user to a controller.
+    bootstrap           Initializes a cloud environment.
+    controllers         Lists all controllers.
+    deploy              Deploys a new application.
+    expose              Makes an application publicly available over the network.
+    models              Lists models a user can access on a controller.
+    status              Displays the current status of Juju, applications, and units.
+    switch              Selects or identifies the current controller and model.
+
+Help commands:
+    
+    juju help           This help page.
+    juju help <command> Show help for the specified command.
+
+For the full list of supported commands run: 
+    
+    juju help commands`[1:]
+
 const (
 	cliHelpHint = `See "juju --help"`
 )
@@ -221,6 +259,8 @@ func NewJujuCommandWithStore(
 			}
 		},
 	})
+	jcmd.AddHelpTopic("basics", "Basic Help Summary", usageHelp)
+
 	jujuRegistry = &jujuCommandRegistry{
 		store:           store,
 		commandRegistry: jcmd,
--- a/cmd/juju/commands/main.go
+++ b/cmd/juju/commands/main.go
@@ -79,7 +79,6 @@ See https://juju.is for getting started tutorials and additional documentation.
 
 Starter commands:
 
-    add-k8s             Adds a k8s endpoint and credential to Juju.
     bootstrap           Initializes a cloud environment.
     add-model           Adds a hosted model.
     deploy              Deploys a new application.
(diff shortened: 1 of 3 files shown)


cmd/juju/commands/main.go
```diff
@@ -70,6 +70,44 @@ or your local machine.
...
+    add-user            Adds a Juju user to a controller.
+    bootstrap           Initializes a cloud environment.
+    controllers         Lists all controllers.
+    deploy              Deploys a new application.
+    expose              Makes an application publicly available over the network.
+    models              Lists models a user can access on a controller.
+    status              Displays the current status of Juju, applications, and units.
+    switch              Selects or identifies the current controller and model.
+
+Help commands:
```
achilleasa, line 102: I slightly tweaked this section and added a dedicated section for listing all commands.

cmd/juju/commands/main.go
```diff
@@ -70,6 +70,44 @@ or your local machine.
...
+Juju is model & application management software designed to leverage the power
+of existing resource pools, particularly cloud-based ones. It has built-in
+support for cloud providers such as Amazon EC2, Google GCE, Microsoft
+Azure, OpenStack, and Rackspace. It also works very well with MAAS and
+LXD. Juju allows for easy installation and management of workloads on a
+chosen resource pool.
+
+See https://jujucharms.com/docs/stable/help for documentation.
+
+Common commands:
```
achilleasa, line 86: We probably want to review these at some point, perhaps split them into more sections (based on functionality?) and include some of the charmhub-related commands which will be available with the 2.9 release.

cmd/juju/commands/repl.go
```diff
@@ -97,8 +96,12 @@ func (c *replCommand) Run(ctx *cmd.Context) error {
 		if err := jujuCmd.Init([]string{"help"}); err != nil {
 			return errors.Trace(err)
 		}
+
+		if err := jujuCmd.Run(ctx); err != nil {
+			return err
+		}
 		fmt.Fprintln(ctx.Stdout, replDoc)
```
achilleasa, line 103: This ensures that the REPL output appears **below** the summarized help text (see QA steps output)

cmd/juju/commands/main.go
```diff
@@ -70,6 +70,44 @@ or your local machine.
...
+
+    add-cloud           Adds a user-defined cloud to Juju.
+    add-credential      Adds or replaces credentials for a cloud.
+    add-model           Adds a hosted model.
+    add-relation        Adds a relation between two applications.
+    add-unit            Adds extra units of a deployed application.
+    add-user            Adds a Juju user to a controller.
```
SimonRichardson, line 93: If we have this, I don't think `add-user` should be here. I see `add-user` as more of a power user command, one that needs better understanding and limitations than, something you'd reach for right away.

SimonRichardson (commented):
```
$ juju
Welcome to the Juju interactive shell.
Type "help" to see a list of available commands.
Type "q" or ^D or ^C to quit.

admin@test:default$ help
...
```

:point_up: Prints the old help out, not the abridged version.

achilleasa:
> ```
> $ juju
> Welcome to the Juju interactive shell.
> Type "help" to see a list of available commands.
> Type "q" or ^D or ^C to quit.
> 
> admin@test:default$ help
> ...
> ```
> 
> ☝️ Prints the old help out, not the abridged version.

That is correct (I think). In REPL mode help should show the command list (as per the help text above).

wallyworld:
FWIW the change being reverted was done after feedback from Real Kinetic (the first set of consultants who were engaged to provide feedback on the Juju experience).

jameinel:
So the ultimate goal is to get a bit closer to a mix between the two:
https://docs.google.com/document/d/15nBMWD90rj7xpUAOXb0LOzJCXxqYaNHUyA-EwPiEHkc/edit#heading=h.gzupgu85v3nx

The problem is that any time you are output 250 lines, you aren't being helpful. The actual CLI design that was worked on with UX was to have a section with a few commonly used commands with their description, and then a categorization of types of commands with just their names. (And a couple other things like 'juju help --all' instead of 'juju help --commands')

I'd like us to at least me making progress towards that.  While I think Real Kinetic mentioned "it is hard to find what I want with limited output", I don't think they actually came back around with "this is useful now". And I think the feedback from CS is "200 lines output drowns the signal in the noise".

If you actually read our current "juju help" output, it reads very much like "heres something you might want, let me spit it out", but Options are intermingled with text, it starts with talking about the shell, but that isn't the context you want as a new user coming to Juju.
The previous text isn't ideal, but I do feel it is significantly better than what we have now, and is better progress towards helping end users.

cmd/juju/commands/main.go
```diff
@@ -70,6 +70,44 @@ or your local machine.
...
 `
 
+var usageHelp = `
+Usage: juju [help] <command>
+
+Summary:
+Juju is model & application management software designed to leverage the power
+of existing resource pools, particularly cloud-based ones. It has built-in
+support for cloud providers such as Amazon EC2, Google GCE, Microsoft
+Azure, OpenStack, and Rackspace. It also works very well with MAAS and
```
jameinel, line 80: We definitely need to mention Kubernetes here.
"It also works very well" is also a bit of a weak statement, though I don't have immediate ways to fix it.

cmd/juju/commands/main.go
```diff
@@ -70,6 +70,44 @@ or your local machine.
...
+LXD. Juju allows for easy installation and management of workloads on a
+chosen resource pool.
+
+See https://jujucharms.com/docs/stable/help for documentation.
```
jameinel, line 84: This should be https://juju.is/docs/help

cmd/juju/commands/main.go
```diff
@@ -70,6 +70,44 @@ or your local machine.
...
+
+Common commands:
+
+    add-cloud           Adds a user-defined cloud to Juju.
+    add-credential      Adds or replaces credentials for a cloud.
+    add-model           Adds a hosted model.
+    add-relation        Adds a relation between two applications.
```
jameinel, line 91: add-relation is superseded by "relate"

cmd/juju/commands/main.go
```diff
@@ -70,6 +70,44 @@ or your local machine.
 See https://discourse.jujucharms.com/ for ideas, documentation and FAQ.
 `
 
+var usageHelp = `
+Usage: juju [help] <command>
+
+Summary:
```
jameinel, line 76: No need for the categories here. Just start with "Juju is ..."

achilleasa:
I pushed a new commit to to address some of @jameinel comments and to nudge the help text a bit closer to the CLI alignment doc contents. Please take a look at the extended commit for details and at the updated QA steps output for how the revised help text looks and let me know what you think.

cmd/juju/commands/main.go
```diff
@@ -63,13 +63,51 @@ func init() {
...
+    relate              Adds a relation between two applications.
+    expose              Makes an application publicly available over the network.
+    models              Lists models a user can access on a controller.
+    controllers         Lists all controllers.
+    whoami              Display the current controller, model and logged in user name. 
+    switch              Selects or identifies the current controller and model.
+    add-cloud           Adds a user-defined cloud to Juju.
+    add-credential      Adds or replaces credentials for a cloud.
+
+REPL mode:
```
jameinel, line 97: REPL seems very tech focused. I would probably call this "Interactive mode:"

cmd/juju/commands/main.go
```diff
@@ -63,13 +63,51 @@ func init() {
...
+var usageHelp = `
+Juju provides easy, intelligent application orchestration on top of Kubernetes,
+cloud infrastructure providers such as Amazon, Google, Microsoft, Openstack,
+MAAS (and more!), or even your local machine via LXD.
+
+See https://juju.is for getting started tutorials and additional documentation.
+
+Starter commands:
+
+    add-k8s             Adds a k8s endpoint and credential to Juju.
```
jameinel, line 82: I would probably put add-k8s next to add-cloud. Especially with the work to bootstrap to "known kubernetes" so you wouldn't have to always start with the add step.

jameinel (commented):
Make sure to update your PR description, but I think this is clearly better than what we have in 2.9 now, and we can look to evolve in going forward.

jameinel (approved):

cmd/juju/commands/main.go
petevg: The term repl makes me happy. It's so cheerful. But I agree that we should probably call it "interactive" or "shell" mode.

achilleasa: Changed to "Interactive Mode"

cmd/juju/commands/main.go
achilleasa: Moved it to the bottom of the list as suggested.

cmd/juju/commands/main.go
```diff
@@ -63,13 +63,51 @@ func init() {
...
+Juju provides easy, intelligent application orchestration on top of Kubernetes,
+cloud infrastructure providers such as Amazon, Google, Microsoft, Openstack,
+MAAS (and more!), or even on your local machine via LXD.
 
-See https://discourse.jujucharms.com/ for ideas, documentation and FAQ.
+See https://juju.is for getting started tutorials and additional documentation.
 `
 
+var usageHelp = `
+Juju provides easy, intelligent application orchestration on top of Kubernetes,
```
achilleasa, line 74: I rephrased this block to make it a bit more "catchy" and to hint that _we have getting started tutorials too_. Let me know if we need any changes to the wording.

achilleasa:
$$merge$$

petevg:
Merging, as the failure to merge appears to be that darn github bug.

PR #12704, opened 2021-03-01
Re-write multi-model test in bash

SimonRichardson:
The multi-model test in python would fail because it would lose track
of which model to query. Considering we're attempting to remove these
python tests in favour of the bash ones, I took the time to do that.

The test is actually really easy and just creates a series of models and
ensures you can switch and talk to each one independently.

## QA steps

```sh
(cd tests && ./main.sh model)
```


tests/includes/juju.sh
```diff
@@ -83,7 +83,7 @@ bootstrap() {
     output=${1}
     shift
 
-    rnd=$(head /dev/urandom | tr -dc a-z0-9 | head -c 8; echo '')
+    rnd=$(rnd_str)
```
achilleasa, line 86: How about `$RANDOM` since we use bash?

SimonRichardson: Because I want letters... ;-)

achilleasa: Why not EMOJIs! 🎆 🍕 🍺 ?

achilleasa (approved):

SimonRichardson:
$$merge$$

PR #12722, opened 2021-03-05
Prevent charm series and systems (bases)

SimonRichardson:
The following brings in the charm change that prevents series and
systems from being used at the same time. We should aim for 
conceptual purity within the modeling of our systems and having 
series and systems prevent that.

So that `series: [focal]` could deviate from `bases: {name: ubuntu,
channel: "16.04/stable"}`. 
This could cause chaos in the future and we should prevent it from 
ever happening.

----

Hyrum's Law states:

> With a sufficient number of users of an API,
> it does not matter what you promise in the contract:
> all observable behaviors of your system
> will be depended on by somebody.

----

See spec: https://discourse.charmhub.io/t/charm-metadata-v2/3674/12
And following google spec doc.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju deploy mysql
$ juju deploy snappass-test --series=focal --channel edge
ERROR storing charm "snappass-test": charm "snappass-test" declares both series and systems
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,7 @@ require (
 	github.com/juju/ansiterm v0.0.0-20180109212912-720a0952cc2a
 	github.com/juju/blobstore/v2 v2.0.0-20210302045357-edd2b24570b7
 	github.com/juju/bundlechanges/v4 v4.0.0-20210223105356-e3037fe2412c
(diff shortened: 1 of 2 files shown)


achilleasa (approved):

SimonRichardson:
$$merge$$

PR #12724, opened 2021-03-07
Use gopkg.in/macaroon-bakery.v3

wallyworld:
Another round of import churn.

lxd upstream does not use go mod so we can't use a versioned module path for any common juju upstream repos.
So `github.com/go-macaroon-bakery/macaroon-bakery/v3` needs to be changed to `gopkg.in/macaroon-bakery.v3`.

Also a small compatbility fix since lxd upstream has diverged from what we pinned to previously. The Juju fix was one line but then a bunch of test mocks needed to be regenerated, adding to the bloat.

## QA steps

juju bootstrap lxd
juju deploy mariadb

juju bootstrap aws
juju deploy mariadb ==to new:lxd


wallyworld:
[more upstream import fun](https://trello.com/c/mYV5YTvJ/1429-more-upstream-import-fun)

tlm (approved):
lgtm

wallyworld:
$$merge$$

PR #12733, opened 2021-03-08
Fix python exception in unregister acceptance test.

hmlanigan:
Fix python exception in unregister acceptance test.   Found with the unregister test which looks for an expected failure.  The same type of check was used in 1 other tests, though not hit as true exception handling looking for errors.

## QA steps

```console 
# Should now succeed
(cd acceptancetests/; ./assess unregister)

# No change expected to 
(cd acceptancetests/;  ./assess model_migration)
```


petevg (approved):
This works. I think that you might be able to do some golfing on it, but errors in CalledProcessError are annoying, and I don't feel like looking up how to do it with fewer characters. It's only a line, anyway :-)

hmlanigan:
$$merge$$

PR #12735, opened 2021-03-09
Silences proxy log messages from klog.

tlm:
On TCP connection resets klog generates a bit of noise that is not a
good user experience. In this commit we filtering out these log messages
to Juju debug.

This isn't an underlying error as it's expected by the Kubernetes
libraries.

## Checklist

## QA steps

Bootstrap to microk8s with the idea being the no errors message should be displayed on a successful bootstrap.

### DIFF
--- a/caas/kubernetes/provider/klog.go
+++ b/caas/kubernetes/provider/klog.go
@@ -4,8 +4,20 @@
 package provider
 
 import (
+	"strings"
+
 	"github.com/go-logr/logr"
 	"github.com/juju/loggo"
+)
+
+type KlogMessagePrefixes []string
+
+var (
+	klogIgnorePrefixes = KlogMessagePrefixes{
+		"an error occurred forwarding",
+		"error copying from remote stream to local connection",
+		"error copying from local connection to remote stream",
+	}
 )
 
 // klogAdapter is an adapter for Kubernetes logger onto juju loggo. We use this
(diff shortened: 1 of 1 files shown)


caas/kubernetes/provider/klog.go
```diff
@@ -30,9 +36,14 @@ func (k *klogAdapter) Enabled() bool {
 func (k *klogAdapter) Error(err error, msg string, keysAndValues ...interface{}) {
 	if err != nil {
 		k.Logger.Errorf(msg+": "+err.Error(), keysAndValues...)
-	} else {
-		k.Logger.Errorf(msg, keysAndValues...)
+		return
+	}
+
+	if strings.HasPrefix(msg, klogPrefixIgnoreProxy) {
+		k.Logger.Debugf(msg, keysAndValues)
```
wallyworld, line 43: Are they really just noise?
Maybe even Trace would be sufficient?

wallyworld (approved):

caas/kubernetes/provider/klog.go
benhoyt, line 43: Should this be `keysAndValues...` like the existing call?

tlm: I don't think so

tlm:
$$merge$$

tlm:
[Reduce the amount of klog noise that is in Juju proxy for Kubernetes.](https://trello.com/c/po8vg2sD/1431-reduce-the-amount-of-klog-noise-that-is-in-juju-proxy-for-kubernetes)

PR #12604, opened 2021-02-08
[2.9] Do not attempt to set proxy settings for windows/centos workloads

achilleasa:
This PR ensures that the proxy-updater worker does not attempt to set apt proxy settings on series that do not support apt.

## QA steps

Try this on maas after ensuring that you have pulled the centos7 image 

```console
$ make install
$ juju bootstrap maas test --no-gui

# Add a centos and a bionic machine and wait for them to come up.
$ juju add-machine --series=centos7
$ juju add-machine --series=bionic

# Dial up the logging and change the proxy settings
$ juju model-config logging-config='<root>=DEBUG;juju.worker.proxyupdater=TRACE'
$ juju model-config http-proxy=http://bogus:1337

# Tail the logs for both machines and check that the centos proxy-updater worker shows the 
# following lines:
 juju.worker.proxyupdater no snap proxies on CentOS
 juju.worker.proxyupdater no apt proxies on CentOS

Ensure that /etc/apt/apt.conf.d/95-juju-proxy-settings on the bionic host has been updated as expected
```

## Bug reference
https://bugs.launchpad.net/juju/+bug/1901069

PR #12746, opened 2021-03-10
[2.9] Fix broken centos proxy tests

achilleasa:
After the fix from https://github.com/juju/juju/pull/12604 landed, the
proxy worker will not attempt to setup a snap proxy on centos/windows
systems. Therefore, the snap-related tests should be skipped on centos
as well as windows or they will fail waiting for a proxy change that
will never happen.

Many thanks to @SimonRichardson for  tracking down the issue and providing a patch!

## QA steps
Run the worker/proxyupdater tests on *centos* (or let the CI do it for you!)

### DIFF
--- a/worker/proxyupdater/proxyupdater_test.go
+++ b/worker/proxyupdater/proxyupdater_test.go
@@ -4,6 +4,7 @@
 package proxyupdater_test
 
 import (
+	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


petevg (approved):
This looks good to me. I think that the kubeflow tests breaking is a separate issue. (Though one that we need to fix ...)

hmlanigan:
$$merge$$

PR #12750, opened 2021-03-11
Refactor series support to that juju itself can set its own supported series

wallyworld:
juju was relying on upstream `github.com/juju/os` package to provide
- available series/versions for ubuntu, macos, windows etc
- default lts

This is the wrong place for it as a given juju codebase should be able to specify what it supports and add new versions without caring about changing an upstream library.

This PR adds in the content of `github.com/juju/os` as `core/os` and `github.com/juju/os/series` is used to update `core/series`. Essentially, Juju now defines all the OS types and series it supports in tree. A number of public functions from the library are now in tree.

What the upstream `juju/os` library does do is provide apis to find out what the latest available series/version are on the host os. This is still used and augments what Juju has. eg on Ubuntu the /usr/share/distroinfo/ubuntu.csv file is read.

The PR is mostly import changes and copied in code. "eoan" is removed and "groovy" and "hirsute" are added as allowed workload series.

## QA steps

juju bootstrap
(default series should be focal)
juju bootstrap --bootstrap-series=bionic

deploy workloads to bionic, focal, groovy


core/os/os.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2015 Canonical Ltd.
```
wallyworld, line 1: All the `core/os` code is copied from upstream `github.com/juju/os`

core/series/supportedseries.go
```diff
@@ -50,18 +56,415 @@ func seriesForTypes(path string, now time.Time, requestedSeries, imageStream str
...
 	source := series.NewDistroInfo(path)
-	supported := NewSupportedInfo(source, defaultSeries)
+	supported := NewSupportedInfo(source, allSeriesVersions)
 	if err := supported.Compile(now); err != nil {
 		return nil, errors.Trace(err)
 	}
 
 	return supported, nil
 }
+
```
wallyworld, line 74: This big block of new code is copied from `github.com/juju/os/series`

wallyworld:
[untangle supported series](https://trello.com/c/kQ3HSShI/1436-untangle-supported-series)

core/series/supported.go
```diff
@@ -292,104 +286,120 @@ var ubuntuSeries = map[SeriesName]SeriesVersion{
...
 	Focal: {
 		WorkloadType: ControllerWorkloadType,
 		Version:      "20.04",
 		LTS:          true,
 		Supported:    false,
 	},
+	Groovy: {
+		WorkloadType: ControllerWorkloadType,
+		Version:      "20.10",
+		Supported:    true,
```
SimonRichardson, line 299: This is wrong, this should be false.

If we don't support focal by default, as we had to make changes to juju to support focal correctly then we shouldn't support groovy either.

SimonRichardson (changes requested):
Additionally, we should drop reading in the distro csv file for supported support. We can and probably should read it for the names and the version, but anything else should be just ignored. It's not true that we support something just because it's out. That means that at 2.0 that reads in the distro csv file will support focal by default and we know that's a lie.

core/series/supported.go
wallyworld: The focal support issue was more so for bootstrap behaviour - what LTS would be chosen by default.
Nonetheless, I've made groovy = false

SimonRichardson (approved):

wallyworld:
$$merge$$

PR #12758, opened 2021-03-12
[2.8] No panic missing bundle app body

SimonRichardson:
The change prevents a panic if the body of a base bundle 
application is missing and used in conjunction with overlay. 
The fix is to error out early to prevent panic out when attempting
to merge the overlays.

## QA steps

Create a bundle.yaml

```yaml
series: focal
applications:
  apache2:
--- # overlay 1
applications:
  apache2:
    num_units: 2
```

```sh
$ juju bootstrap lxd test
$ juju deploy ./bundle.yaml
ERROR cannot deploy bundle: base application "apache2" has no body
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1918742

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,7 +35,7 @@ require (
 	github.com/juju/ansiterm v0.0.0-20180109212912-720a0952cc2a
 	github.com/juju/blobstore/v2 v2.0.0-20210302045357-edd2b24570b7
 	github.com/juju/bundlechanges v1.0.1-0.20210209114844-acd700ed48fe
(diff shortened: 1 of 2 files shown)


achilleasa (approved):

achilleasa:
@SimonRichardson it might be useful to clarify that this issue is caused by empty bodies in **base** bundles which is a user-error and not something that juju would include in exported bundles.

SimonRichardson:
$$merge$$

PR #12754, opened 2021-03-11
[2.8] Record hostname when machine agent starts

achilleasa:
This PR provides a solution for scenarios such us the one described in the referenced bug. More specifically, given that (provider) instance IDs are typically not related to the machine's host name and that some providers do not let us query the machine host name, `juju show-machine` output will include an IP address in the `dns-name` field.

The PR introduces a consistent cross-provider way of recording and reporting the machine host names. To achieve this, the machine agents will now query the machine host name and report it at the same time that they request from the controller to record the agent start time. While the machiner facade already had a `RecordAgentStartTime` method (added in 2.8), it accepted a slice of `params.Entity` values which meant that we could not push additional information to the controller.

To this end, the PR adds a new API call named `RecordAgentStartInformation` which supersedes `RecordAgentStartTime` and bumps the machiner facade version to **5**. Given that the machiner facade versions have not diverged in the 2.9 branch, this change can easily be forward ported. The machiner API client utilizes facade version checks to select the appropriate method to call.

In a similar fashion to the `AgentStartTime` field, the `Hostname` field will always get populated after the agent restart. As a result there is no need to modify the `juju/description` package or worry about migrations.

To report back the recorded machine hostname, the payload for the client/status (MachineStatus in FullStatus) had to be augmented with an extra field. In this case, the status API has already diverged in 2.9 so I opted to simply add the extra (it's optional anyway) field to the status response **without bumping the facade version** so we can merge the changes forward. 

Finally, the CLI has been tweaked to include the host name (if not empty) in `juju show-machine` output.

## QA steps

```console
# Bootstrap with the current 2.8 juju snap
$ /snap/bin/juju bootstrap lxd test --no-gui

$ juju switch controller
$ juju show-machine 0
model: controller
machines:
  "0":
    juju-status:
      current: started
      since: 11 Mar 2021 12:52:21Z
      version: 2.8.9
    dns-name: 10.232.217.231
    ip-addresses:
    - 10.232.217.231
    instance-id: juju-1ee6f9-0
    machine-status:
      current: running
      message: Running
      since: 11 Mar 2021 12:52:31Z
    modification-status:
      current: idle
      since: 11 Mar 2021 12:52:08Z
    series: bionic
    network-interfaces:
      eth0:
        ip-addresses:
        - 10.232.217.231
        mac-address: 00:16:3e:24:20:9e
        gateway: 10.232.217.1
        is-up: true
    hardware: arch=amd64 cores=0 mem=0M
    controller-member-status: has-vote

[...]

PR #12760, opened 2021-03-12
[2.8] Include hostname in multi-watcher machine deltas

achilleasa:
This is a followup PR to #12754 to ensure that the agent-reported machine host name is included in the deltas produced by the multi-watcher/all-watcher API calls so this information can be made available via pylibjuju.

Similar to the client/status facade changes in the linked PR, the watcher facade has not been bumped as hostname is an optional field.

No further QA required; the changes will be tested as part of an upcoming PR against pylibjuju.


SimonRichardson (approved):

achilleasa:
$$merge$$

PR #12766, opened 2021-03-15
Enable ha integration test

SimonRichardson:
The following adds a generic enable-ha integration test, along with
removing ha to ensure that we are correctly going back from ha as 
well.

The code is actually simple and not that hard to follow, just waiting
for the machines to appear is most of the code.

## QA steps

```sh
(cd tests && ./main.sh controller)
```

### DIFF
--- a/tests/suites/controller/enable_ha.sh
+++ b/tests/suites/controller/enable_ha.sh
@@ -1,9 +1,9 @@
 wait_for_controller_machines() {
-    ammount=${1}
+    amount=${1}
 
     attempt=0
     # shellcheck disable=SC2143
-    until [[ "$(juju machines -m controller --format=json | jq -r ".machines | .[] | .[\"juju-status\"] | select(.current == \"started\") | .current" | wc -l | grep "${ammount}")" ]]; do
+    until [[ "$(juju machines -m controller --format=json | jq -r ".machines | .[] | .[\"juju-status\"] | select(.current == \"started\") | .current" | wc -l | grep "${amount}")" ]]; do
         echo "[+] (attempt ${attempt}) polling machines"
         juju machines -m controller 2>&1 | sed 's/^/    | /g'
         sleep "${SHORT_TIMEOUT}"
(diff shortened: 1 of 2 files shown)


tests/suites/controller/enable_ha.sh
```diff
@@ -0,0 +1,62 @@
+wait_for_controller_machines() {
+    ammount=${1}
```
hmlanigan, line 2: ammount,  typo?

tests/suites/controller/enable_ha.sh
```diff
@@ -0,0 +1,62 @@
...
+
+    attempt=0
+    until [[ "$(juju machines -m controller --format=json | jq -r ".machines | .[] | .[\"juju-status\"] | select(.current == \"started\") | .current" | wc -l | grep "${ammount}")" ]]; do
```
hmlanigan, line 6: This ensures that the machines have started.  Check for ha-status = ha-enabled on all 3 machines to ensure HA is up and running.

$ juju show-controller
  controller-machines:
    "0":
      instance-id: juju-a2cba4-0
      ha-status: ha-enabled
      ha-primary: true
    "1":
      instance-id: juju-a2cba4-1
      ha-status: ha-enabled
    "2":
      instance-id: juju-a2cba4-2
      ha-status: ha-enabled

hmlanigan: Or "controller-member-status": "has-vote" from juju machines.

hmlanigan (changes requested):

tests/suites/controller/enable_ha.sh
SimonRichardson: Do we care for this test? I know what you mean in the general case, but for this test we know we'll have exactly 3 machines for a controller...?

hmlanigan: We do, just because the machines deploy, doesn't mean that enable HA was successful.  The delays in the field have been after the machines were started.

hmlanigan (approved):
Thank you for the changes.  LGTM.  Eventually we should consider lowering the time allowed, or create a new test which has shorter timeframe.  To help catch items like lp: 184337

SimonRichardson:
$$merge$$

PR #12768, opened 2021-03-15
Use system instead of core for writing proxy settings.

SimonRichardson:
The following drops the need to install snap core when deploying a machine. Instead, we can use system to set these values. The motivation behind these charges, installing via snap can sometimes take 4-5 minutes and installing the core charm that is superfluous just to set proxy values seems overkill.

Whilst working on this, I also solved two other problems:

 1. `/server/snap/` was in a bit of a mess, it didn't have great tests and hadn't been thought out very well in terms of how commands should be called. We now have better tests around this.
 2. Retries of snap installs should be specific to call locations and with a retry strategy for the service, this is now done.

Additionally, this brings in the changes to `juju/packaging` as well to also call system over core.

### Ensure we don't install core.

Ensure using focal image to get the juju-db installed via snap.

```sh
$ juju bootstrap lxd test --bootstrap-series=focal --build-agent
```

Login to the container and ensure that core is no longer installed, use `tail -f`
on `/var/log/cloud-init-output.log`.

```sh
2021-03-22 13:55:17 WARNING juju.mongo service.go:326 configuring mongod  with --noauth flag enabled
2021-03-22 13:55:17 INFO juju.packaging manager.go:103 installing "juju-db" via "snap"
2021-03-22 13:55:17 INFO juju.packaging.manager utils.go:64 Running: snap install  --channel 4.0/stable juju-db
```

#### Setup the store proxy

```sh
$ sudo snap install snap-store-proxy
$ sudo snap-proxy config proxy.domain="<domain>"
$ snap-proxy check-connections
http: https://dashboard.snapcraft.io: OK
http: https://login.ubuntu.com: OK
http: https://api.snapcraft.io: OK
All connections appear to be accessible
$ sudo snap-proxy generate-keys
$ sudo snap-proxy register
```

#### Bootstrap using store proxy

```sh
$  juju bootstrap lxd test --no-gui --debug --snap-store-proxy-url="http://<proxy-ip>"
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1920033


service/snap/app.go
```diff
@@ -0,0 +1,152 @@
...
+// expect.
+func (p ConfinementPolicy) Validate() error {
+	switch p {
+	case StrictPolicy, ClassicPolicy, DevModePolicy, JailModePolicy:
+		return nil
+	}
+	return errors.NotValidf("%s confinement", p)
+}
+
+func (p ConfinementPolicy) String() string {
```
achilleasa, line 42: Need a comment for exported identifiers.

service/snap/app.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// NewApp creates a new application from a given name.
+func NewNamedApp(name string) *App {
+	return &App{
+		name: name,
+	}
+}
+
+func NewApp(name string, channel string, policy ConfinementPolicy, services []BackgroundService, prerequisites []Installable) *App {
```
achilleasa, line 62: Need a comment for exported identifiers.

service/snap/snap.go
```diff
@@ -39,6 +39,13 @@ var (
 	snapNameRe = regexp.MustCompile("^[a-z0-9][a-z0-9-]{0,39}[^-]$")
 )
 
+// Runnable expects to be able to run a given command with a series of arguments
+// and return the output and/or error from that executing command.
+type Runnable interface {
+	Execute(name string, args ...string) (string, error)
+	Clock() clock.Clock
```
achilleasa, line 46: Why would a `Runnable` thing give you access to a `clock.Clock`?  The `Clock` method seems out of place in this interface.

service/snap/snap.go
```diff
@@ -359,28 +303,16 @@ func (s Service) Installed() (bool, error) {
...
-	command := fmt.Sprintf("%v install --channel=%v %v %v",
-		s.executable,
-		s.app.Channel,
-		confinementPolicy,
-		s.app.Name,
-	)
-	logger.Infof("preparing command: %v", command)
-	commands = append(commands, command)
+	deps := s.app.Prerequisites()
+	commands := make([]string, 1+len(deps))
```
achilleasa, line 307: You could always change this to `commands := make([]string, 0, 1+len(deps))` and just append ;-)

service/snap/snap.go
```diff
@@ -390,7 +322,8 @@ func (s Service) ConfigOverride() error {
 	if len(s.conf.Limit) == 0 {
 		return nil
 	}
-	for _, backgroundService := range s.app.BackgroundServices {
+
+	for _, backgroundService := range s.app.BackgroundServices() {
 		unitOptions := systemd.ServiceLimits(s.conf)
```
achilleasa, line 327: Irrelevant to the changeset but why aren't these 3 lines pulled out of the loop?

service/snap/snap.go
achilleasa: Actually, you are using the append pattern below.

service/snap/snap.go
```diff
@@ -536,6 +462,33 @@ func (s *Service) execThenExpect(commandArgs []string, expectation string) error
...
 
-func (s *Service) runCommand(args ...string) (string, error) {
-	return utils.RunCommand(s.executable, args...)
+func (s Service) runCommand(args ...string) (string, error) {
+	return s.runnable.Execute(s.executable, args...)
+}
+
+func (s Service) runCommandWithRetry(args ...string) (res string, err error) {
+	if resErr := retry.Call(retry.CallArgs{
+		Clock: s.runnable.Clock(),
```
achilleasa, line 471: I think clocks should be injected via a different route here rather than the runnable instance (see my comment above).

service/snap/snap.go
SimonRichardson: Agreed, was being cheeky here.

service/snap/snap.go
SimonRichardson: Ouch, missed that one...

achilleasa (commented):
Thanks for the changes. LGTM; running the QA steps now.

achilleasa (approved):
QA works as expected. Also tried adding a bionic machine and deploying --to lxd:X

SimonRichardson:
$$merge$$

PR #12771, opened 2021-03-16
Upgrade series graph/FSM

SimonRichardson:
<strike>Requires #12763 to land, will rebase after landing.</strike>

The previous logic for upgrade series was really bad, essentially it
allowed you to jump up the order to reach the error state without
actually ensuring that you could access all the states.

The following code changes that and introduces an FSM. The FSM ensures
that you can go to only the states that are supplied in the child
states.

The FSM is modeled as a DAG with each state show how it can move along
the edges to each vertex. This tightens up a lot of the logic around a
upgrade series and ensures that you have to move to each state
correctly.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju upgrade-series 0 complete
```


manadart (approved):
Nice.

SimonRichardson:
$$merge$$

PR #12772, opened 2021-03-16
[2.8] Add link-local field to netplan.Interface

achilleasa:
When juju parses an existing netplan configuration, it uses string YAML unmarshaling (i.e. unmarshaling fails if we encounter a field that is not present in the struct we unmarshal to) to ensure that we don't accidentally drop an entry when we write the netplan configuration back.

This PR adds `link-local` to the list of supported fields (see https://netplan.io/reference/) to allow the bridge operations to work as expected in the scenario described in the following bug report.

## Bug reference
https://bugs.launchpad.net/juju/+bug/1919144

### DIFF
--- a/network/netplan/netplan.go
+++ b/network/netplan/netplan.go
@@ -44,6 +44,10 @@ type Interface struct {
 	RoutingPolicy  []RoutePolicy `yaml:"routing-policy,omitempty"`
 	// Optional doesn't have to be *bool because it is only used if True
 	Optional bool `yaml:"optional,omitempty"`
+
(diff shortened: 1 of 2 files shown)


jameinel (approved):
Do we need a test suite update of 'common netplan configs' that we have encountered in the wild, and proof that we are able to injest them?

achilleasa:
@jameinel That's a great idea. I will push an extra commit with the suite.

achilleasa:
Turns out we already have one. I will add an extra test for an ethernet device that includes`link-local`.

achilleasa:
$$merge$$

PR #12773, opened 2021-03-17
Pull resources for app from store

hpidcock:
Uses the resource opener from resource/resourceadapters to pull the resource if not available.

## QA steps

You'll need to apply the patch from the [first comment](https://github.com/juju/juju/pull/12773#issuecomment-800783793)

Then you can `juju deploy snappass-test --series=focal --channel=edge`

## Documentation changes

N/A

## Bug reference

N/A

### DIFF
--- a/resource/resourceadapters/opener.go
+++ b/resource/resourceadapters/opener.go
@@ -21,14 +21,41 @@ import (
 //
 // The caller owns the State provided. It is the caller's
 // responsibility to close it.
-func NewResourceOpener(st *corestate.State, unitName string) (opener resource.Opener, err error) {
-	return newInternalResourceOpener(&stateShim{st}, unitName)
+func NewResourceOpener(st ResourceOpenerState, unitName string) (opener resource.Opener, err error) {
+	return newInternalResourceOpener(st, unitName, "")
 }
 
-func newInternalResourceOpener(st ResourceOpenerState, unitName string) (opener resource.Opener, err error) {
-	unit, err := st.Unit(unitName)
+// NewResourceOpenerForApplication returns a new resource.Opener for the given app.
+//
+// The caller owns the State provided. It is the caller's
+// responsibility to close it.
+func NewResourceOpenerForApplication(st ResourceOpenerState, applicationName string) (opener resource.Opener, err error) {
+	return newInternalResourceOpener(st, "", applicationName)
+}
+
+// NewResourceOpenerState wraps a State pointer for passing into NewResourceOpener/NewResourceOpenerForApplication.
+func NewResourceOpenerState(st *corestate.State) ResourceOpenerState {
+	return &stateShim{st}
+}
+
+func newInternalResourceOpener(st ResourceOpenerState, unitName string, applicationName string) (opener resource.Opener, err error) {
+	unit := Unit(nil)
+	if unitName != "" {
+		unit, err = st.Unit(unitName)
+		if err != nil {
+			return nil, errors.Annotate(err, "loading unit")
+		}
+	}
+
+	if applicationName == "" {
+		if unit == nil {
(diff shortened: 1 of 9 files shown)


hpidcock:
```
diff --git a/charmhub/refresh.go b/charmhub/refresh.go
index 51cad7041b..a4a57bc8bf 100644
--- a/charmhub/refresh.go
+++ b/charmhub/refresh.go
@@ -116,6 +116,20 @@ func (c *RefreshClient) Refresh(ctx context.Context, config RefreshConfig) ([]tr
                return nil, errors.Trace(err)
        }

+       // HAACK
+       // TODO(benhoyt) - total hack for demo, remove
+       // The Charmhub API seems to be returning type="file" incorrectly
+       for _, result := range resp.Results {
+               if result.Entity.Name != "snappass-test" {
+                       continue
+               }
+               for i, res := range result.Entity.Resources {
+                       c.logger.Errorf("Refresh(): replacing type %q with \"oci-image\"", res.Type)
+                       result.Entity.Resources[i].Type = "oci-image"
+               }
+       }
+       // END HAACK
+
        c.logger.Tracef("Refresh() unmarshalled: %s", pretty.Sprint(resp.Results))
        return resp.Results, config.Ensure(resp.Results)
 }
diff --git a/go.mod b/go.mod
index 727e2b809a..3ab05da0a9 100644
--- a/go.mod
+++ b/go.mod
@@ -36,7 +36,7 @@ require (
        github.com/juju/ansiterm v0.0.0-20180109212912-720a0952cc2a
        github.com/juju/blobstore/v2 v2.0.0-20210302045357-edd2b24570b7
        github.com/juju/bundlechanges/v4 v4.0.0-20210223105356-e3037fe2412c
-       github.com/juju/charm/v8 v8.0.0-20210305131337-93c70ae08c0b
+       github.com/juju/charm/v8 v8.0.0-20210304015923-e7ca8ffa7716
        github.com/juju/charmrepo/v6 v6.0.0-20210309083204-29c3dbf03675
        github.com/juju/clock v0.0.0-20190205081909-9c5c9712527c
        github.com/juju/cmd v0.0.0-20200108104440-8e43f3faa5c9
```

resource/resourceadapters/opener.go
```diff
@@ -21,14 +21,36 @@ import (
...
 
-func newInternalResourceOpener(st ResourceOpenerState, unitName string) (opener resource.Opener, err error) {
-	unit, err := st.Unit(unitName)
+func NewResourceOpenerForApplication(st ResourceOpenerState, applicationName string) (opener resource.Opener, err error) {
+	return newInternalResourceOpener(st, "", applicationName)
+}
+
+func NewResourceOpenerState(st *corestate.State) ResourceOpenerState {
+	return &stateShim{st}
+}
```
wallyworld, line 34: Should probably comment these

resource/resourceadapters/opener.go
```diff
@@ -38,13 +60,18 @@ func newInternalResourceOpener(st ResourceOpenerState, unitName string) (opener
 
 	var resourceClient ResourceRetryClientGetterFn
 
-	curl, _ := unit.CharmURL()
+	charmURL := (*charm.URL)(nil)
```
wallyworld, line 63: Maybe not worth it, but could have an interface with CharmURL() and Tag() (which both app and unit implement) and pass to an extracted function of the code below to avoid the if unit != nil stuff

wallyworld (approved):

resource/resourceadapters/opener.go
hpidcock: ahaha I did this to begin with then deleted it because I thought you of all people wouldn't like it.

hpidcock:
$$merge$$

PR #12783, opened 2021-03-18
Replace systems with bases for metadata v2

hpidcock:
Updates charm/v8 to latest and applies refactors to use bases/assumes and remove systems/platforms/architectures etc.

Fixes closing pebble connections.

## QA steps

You can use the testing charms from [hpidcock/testing-charms](https://github.com/hpidcock/testing-charms)

`juju deploy cockroachdb.charm --resource cockroachdb-image=cockroachdb/cockroach:v20.1.4 --storage database=1G`

## Documentation changes

N/A

## Bug reference

Also fixes https://github.com/canonical/pebble/issues/15

## Related

https://github.com/canonical/operator/pull/490

### DIFF
--- a/worker/uniter/pebblepoller_test.go
+++ b/worker/uniter/pebblepoller_test.go
@@ -106,12 +106,17 @@ func (s *pebblePollerSuite) TestStart(c *gc.C) {
 	doRestart("a")
 
 	workertest.CleanKill(c, worker)
+
+	for k, v := range clients {
+		c.Assert(v.closed, jc.IsTrue, gc.Commentf("client %s not closed", k))
+	}
 }
 
 type fakePebbleClient struct {
 	sysInfo pebbleclient.SysInfo
 	err     error
 	mut     sync.Mutex
+	closed  bool
 }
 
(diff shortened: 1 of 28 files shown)


worker/uniter/pebblepoller_test.go
```diff
@@ -130,3 +130,6 @@ func (c *fakePebbleClient) TriggerStart() {
 	c.err = nil
 	c.sysInfo.BootID = utils.MustNewUUID().String()
 }
+
+func (c *fakePebbleClient) CloseIdleConnections() {
```
wallyworld, line 134: Can we add something to make sure this is called as expected in the defer

wallyworld (approved):

worker/uniter/pebblepoller_test.go
benhoyt: Or alternatively: 1) create the client once outside the poll loop, or 2) add `DisableKeepAlive: true` to the Pebble client config. I'd probably prefer the latter as I don't 100% trust keepalive in a long-running context like this. :-)

hpidcock:
$$merge$$

PR #12787, opened 2021-03-18
Make errors from API2Resource users friendly by include relevent

hmlanigan:
Found during test of a new charmstore compatibility API endpoint.  The error message seen when juju gets empty resource data is unhelpful, especially when deploying a bundle:

ERROR cannot deploy bundle: bad metadata: resource missing filename

Annotate the error message to include the charm and resource name, if available:

ERROR cannot deploy bundle: cs:openstack-dashboard-507 resource "theme": bad metadata: resource missing filename

### DIFF
--- a/charmstore/client.go
+++ b/charmstore/client.go
@@ -277,7 +277,7 @@ func (c Client) listResources(ch CharmID) ([]resource.Resource, error) {
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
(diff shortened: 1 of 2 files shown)


SimonRichardson (changes requested):
Can we add a test for this, so we don't regress?

SimonRichardson (approved):

hmlanigan:
$$merge$$

PR #12788, opened 2021-03-19
Enable RBAC in microk8s;

ycliuhw:
Enable RBAC in microk8s;

### DIFF
--- a/acceptancetests/jujupy/k8s_provider/base.py
+++ b/acceptancetests/jujupy/k8s_provider/base.py
@@ -85,6 +85,8 @@ class Base(object):
     default_storage_class_name = None
     kubeconfig_cluster_name = None
 
+    enable_rbac = None
+
     def _ensure_cluster_stack(self):
         """ensures or checks if stack/infrastructure is ready to use.
         - ensures kubectl/apiserver is functioning.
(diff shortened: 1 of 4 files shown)


ycliuhw:
[CI(non-kubeflow)-Enable RBAC microk8s](https://trello.com/c/eKH58MJW/1455-cinon-kubeflow-enable-rbac-microk8s)

acceptancetests/jujupy/k8s_provider/base.py
```diff
@@ -114,7 +116,7 @@ def _tear_down_substrate(self):
         """
         raise NotImplementedError()
 
-    def __init__(self, bs_manager, cluster_name=None, timeout=1800):
+    def __init__(self, bs_manager, cluster_name=None, enable_rbac=False, timeout=1800):
```
wallyworld, line 119: Maybe this should be true by default

wallyworld (approved):

acceptancetests/jujupy/k8s_provider/base.py
wallyworld: I guess that means we'd need to make the jenkins job do "--disable-rbac" instead of "--enable-rbac"

Was thinking it's better to be more restrictive by default, but it probably doesn't matter

ycliuhw: Given that kubeflow still does not enable RBAC yet, we should probably disable it for now unless we know we want to enable it.

PR #12792, opened 2021-03-19
Remove the force when destroying models

SimonRichardson:
The idea is to test the model teardown as an implicit step. Although
it's annoying to see your test failing because of a tear down issue.
It's important for juju to not fail in real life because of tear down
issue and should be solved.

## QA steps

Run some of the integration tests to see the damage.

```sh
(cd tests && ./main.sh model)
```


hpidcock (approved):
LGTM - tests passed locally

hpidcock:
$$merge$$

PR #12815, opened 2021-03-24
Allow setting of VM hardware version

gabriel-samfira:
This change adds a new model level flag, that allows operators to
set a newer compatibility version for the instances that get spawned
by juju.

The version set via model config must be of equal or higher version
than that of the source template, and it must be supported by the
remote compute resource it will be spawned on. Attempting to set a
version that does not conform to the mentioned rules, will be ignored,
and the VM will use the same HW compatibility version as the source
template.

## Checklist

## QA steps

```sh
juju model-config force-vm-hardware-version=12
juju add-machine
```
Check if the VM compatibility version is set to 12 in your vsphere deployment, for the newly spawned VM.

## Documentation changes

No changes required to CLI or API. This change adds a single model config option.

### DIFF
--- a/provider/vsphere/internal/vsphereclient/client.go
+++ b/provider/vsphere/internal/vsphereclient/client.go
@@ -614,9 +614,7 @@ func (c *Client) maybeUpgradeVMHardware(
 	}
 
 	if vmVersion > args.ForceVMHardwareVersion {
-		// Not downgrading VM.
-		c.logger.Warningf("selected HW version is lower than VM hardware. Not downgrading.")
-		return nil
+		return errors.Errorf("selected HW (%d) version is lower than VM hardware", args.ForceVMHardwareVersion)
 	}
 
 	supportedMaxVersion, err := c.getMaxSuportedVersion(ctx, args.ComputeResource)
@@ -626,8 +624,8 @@ func (c *Client) maybeUpgradeVMHardware(
 	if supportedMaxVersion < args.ForceVMHardwareVersion {
 		// Requested HW version is newer than what the destination supports
 		// Not upgrading VM hardware.
-		c.logger.Warningf("target does not support selected hardware version: %d", args.ForceVMHardwareVersion)
-		return nil
+		return errors.Errorf(
+			"hardware version %d not supported by target (max version %d)", args.ForceVMHardwareVersion, supportedMaxVersion)
 	}
 
 	version := fmt.Sprintf("vmx-%d", args.ForceVMHardwareVersion)
--- a/provider/vsphere/internal/vsphereclient/client.go
+++ b/provider/vsphere/internal/vsphereclient/client.go
@@ -550,7 +550,7 @@ func (c *Client) getMaxSuportedVersion(ctx context.Context, cr *mo.ComputeResour
 	}
 	opt, err := methods.QueryConfigOption(ctx, c.client, &req)
 	if err != nil {
-		return 0, err
+		return 0, errors.Trace(err)
 	}
 
 	if opt.Returnval == nil {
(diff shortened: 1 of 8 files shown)


hpidcock:
[External PR - Allow setting of VM hardware version](https://trello.com/c/6wO6P5Kg/1473-external-pr-allow-setting-of-vm-hardware-version)

hpidcock:
@gabriel-samfira is this still WIP?

gabriel-samfira:
@hpidcock Apologies for the delay. For some reason, not getting emails on new comments. I have added tests to this PR. Please take a look when time permits.

hpidcock:
!!build!!

hpidcock (approved):
I'm happy with this unless @hmlanigan has any feedback

hpidcock:
@hmlanigan do you think we need to extend the provider specific documentation to provide reference to these options https://juju.is/docs/vsphere-cloud

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -538,6 +540,108 @@ func megabytesToKiB(diskMiB uint64) uint64 {
...
+	if cr == nil || cr.EnvironmentBrowser == nil {
+		return 0, errors.Errorf("invalid compute resource")
+	}
+	ref := cr.EnvironmentBrowser.Reference()
+	req := types.QueryConfigOption{
+		This: ref,
+	}
+	opt, err := methods.QueryConfigOption(ctx, c.client, &req)
+	if err != nil {
+		return 0, err
```
SimonRichardson, line 553:
```suggestion
		return 0, errors.Trace(err)
```

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -538,6 +540,108 @@ func megabytesToKiB(diskMiB uint64) uint64 {
...
+}
+
+func (c *Client) getVMHardwareVersion(ctx context.Context, srcVM *object.VirtualMachine) (int64, error) {
+	if srcVM == nil {
+		return 0, errors.Errorf("source VM may not be nil")
+	}
+	var templateVM mo.VirtualMachine
+	err := srcVM.Properties(ctx, srcVM.Reference(), []string{"config.version"}, &templateVM)
+	if err != nil {
+		return 0, err
```
SimonRichardson, line 574:
```suggestion
		return 0, errors.Trace(err)
```

provider/vsphere/internal/vsphereclient/client.go
```diff
@@ -538,6 +540,108 @@ func megabytesToKiB(diskMiB uint64) uint64 {
...
+	if err != nil {
+		return 0, errors.Trace(err)
+	}
+	return parsed, nil
+}
+
+func (c *Client) parseVMXVersion(version string) (int64, error) {
+	fields := strings.Split(version, "-")
+	if len(fields) != 2 {
+		return 0, fmt.Errorf("invalid VMX version: %s", version)
```
SimonRichardson, line 590: Can we use `errors.Errorf`? Using the `errors` package will give us a stack trace

hmlanigan:
@hpidcock, yes, I'll update the doc once 2.9 is out.  Good catch.  We also need to fix this bug: https://bugs.launchpad.net/juju/+bug/1884794

hmlanigan (commented):
QA:

If I use a hw ver higher than the max, the machine fails with a nice error message.  However retry-provisioning doesn't work, the machine-status is "provisioning error". 

```console
Machine  State    DNS            Inst id        Series  AZ  Message
...
2        down                    pending        focal       hardware version 19 not supported by target (max version 17)
```

There seems to some disagreement about in which cases retry-provisioning should work: https://bugs.launchpad.net/juju/+bug/1843134/comments/2 and a few bugs on the topic.

hmlanigan:
$$merge$$

PR #12816, opened 2021-03-24
Equinix provider

jasmingacic:
Implementation of Equinix Metal Juju Provider (previously Packet Bare Metal)


provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+		}
+
+		var instArch string
+		switch {
+		case strings.HasSuffix(plan.Name, ".x86"):
+			instArch = arch.AMD64
+		case strings.HasSuffix(plan.Name, ".arm"):
+			instArch = arch.ARM64
+		default:
+			continue nextPlan
```
SimonRichardson, line 466: Juju doesn't generally use labels, we should instead extract the switch into a method/function and continue on that.

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+
+		}
+
+		mem, err := parseMemValue(plan.Specs.Memory.Total)
+		if err != nil {
+			continue
```
SimonRichardson, line 472: Log the error.

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+	isInvalid := notAvailable || plan.Pricing == nil ||
+		plan.Specs == nil ||
+		plan.Specs.Memory == nil ||
+		len(plan.Specs.Cpus) == 0 || plan.Specs.Cpus[0].Count == 0
+
+	return !isInvalid
+}
+
+func parseMemValue(v string) (uint64, error) {
+	var scaler = uint64(1)
```
SimonRichardson, line 509:
```suggestion
	scaler := uint64(1)
```

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+
+	val, err := strconv.ParseUint(v, 10, 64)
+	return val * scaler, err
+}
+
+func (e *environ) findInstanceSpec(controller bool, allImages []*imagemetadata.ImageMetadata, instanceTypes []instances.InstanceType, ic *instances.InstanceConstraint) (*instances.InstanceSpec, error) {
+	// version, err := series.UbuntuSeriesVersion(ic.Series)
+	// if err != nil {
+	// 	return nil, errors.Trace(err)
+	// }
```
SimonRichardson, line 523: Delete me.

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+				if err != nil || ic.Series != series {
+					continue nextImage
+				}
+			case "windows":
+				series, err := series.WindowsVersionSeries(os.Version)
+				if err != nil || ic.Series != series {
+					continue nextImage
+				}
+			default:
+				continue nextImage
```
SimonRichardson, line 552: Extract switch into method/function so it's more testable.

SimonRichardson (changes requested):

SimonRichardson:
@jasmingacic Also can we have a better description. I would like to see the reason for certain implementations, Q&A steps in how to manually use this.

jasmingacic:
@SimonRichardson Thanks for the input. Are there any specific implementations you want to get more information?

SimonRichardson:
@jasmingacic I'd expect to see what works or more even importantly what doesn't work in the implementation. What needs more work etc.

cloud/fallback-public-cloud.yaml
```diff
@@ -369,10 +322,27 @@ clouds:
         endpoint: https://identity.api.rackspacecloud.com/v2.0
       hkg:
         endpoint: https://identity.api.rackspacecloud.com/v2.0
+  joyent:
```
achilleasa, line 325: This has been removed since 2.9.

cloud/fallback-public-cloud.yaml
```diff
@@ -1,56 +1,9 @@
...
-        endpoint: https://ec2.ap-northeast-3.amazonaws.com
-      me-south-1:
-        endpoint: https://ec2.me-south-1.amazonaws.com
-      sa-east-1:
-        endpoint: https://ec2.sa-east-1.amazonaws.com
-  ecs:
-    type: ecs
-    description: Amazon Elastic Container Service
-    auth-types: [ access-key ]
+    auth-types: [access-key]
```
achilleasa, line 6: Why were the ec2 regions removed here?

cloud/fallback-public-cloud.yaml
jasmingacic: fixed

cloud/fallback-public-cloud.yaml
jasmingacic: fixed

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+		if err != nil {
+			return errors.Trace(err)
+		}
+	}
+
+	return e.Destroy(ctx)
+}
+
+func (e *environ) InstanceTypes(context.ProviderCallContext, constraints.Value) (instances.InstanceTypesWithCostMetadata, error) {
+	var i envinstance.InstanceTypesWithCostMetadata
```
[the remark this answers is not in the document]
jasmingacic: cannot convert nil (untyped nil value) to instances.InstanceTypesWithCostMetadata

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+	// Install additional dependencies that are present in ubuntu images
+	// but not in the versions built by equinix.
+	//
+	// NOTE(achilleasa): this is a hack and is only meant to be used
+	// temporarily; we must ensure that equinix mirrors the official
+	// ubuntu cloud images.
+	if _, err := series.UbuntuSeriesVersion(args.InstanceConfig.Series); err == nil {
+		cloudCfg.AddScripts(
+			"apt-get update",
+			"DEBIAN_FRONTEND=noninteractive apt-get --option=Dpkg::Options::=--force-confdef --option=Dpkg::Options::=--force-confold --option=Dpkg::Options::=--force-unsafe-io --assume-yes --quiet install dmidecode snapd lxd",
```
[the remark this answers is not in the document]
jasmingacic: There is a TODO couple lines above that @achilleasa left

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+	for k, v := range args.InstanceConfig.Tags {
+		packetTags = append(packetTags, fmt.Sprintf("%s=%s", k, v))
+	}
+
+	device := &packngo.DeviceCreateRequest{
+		Hostname:     e.name,
+		Facility:     []string{e.cloud.Region},
+		Plan:         spec.InstanceType.Name,
+		OS:           spec.Image.Id,
+		ProjectID:    e.cloud.Credential.Attributes()["project-id"],
```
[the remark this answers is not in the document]
jasmingacic: It is being checked when initializing

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+
+			requestedPublicAddr = requestedPublicAddr || net.Public
+			requestedPrivateAddr = requestedPrivateAddr || !net.Public
+
+			// Packet requires us to request at least a /31 for IPV4
+			// addresses and a /127 for IPV6 ones.
+			cidrSize := 31
+			if net.AddressFamily != 4 {
+				cidrSize = 127
+			}
```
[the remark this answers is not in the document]
jasmingacic: This is provider specific and we had to add it

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+	amd64 := arch.AMD64
+	mem, err := strconv.ParseUint(d.Plan.Specs.Memory.Total[:len(d.Plan.Specs.Memory.Total)-2], 10, 64)
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	var cpus uint64 = 1
+	if inst.Plan != nil && inst.Plan.Specs != nil && len(inst.Plan.Specs.Cpus) > 0 {
+		cpus = uint64(inst.Plan.Specs.Cpus[0].Count)
+	}
```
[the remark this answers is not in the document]
jasmingacic: I believe this was done because of some legacy in the API. This will probably go away.

jasmingacic:
I reverted my last commit as it introduced some issues and I didn't want to block @achilleasa

provider/equinix/environ_network.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/juju/juju/core/instance"
+	"github.com/juju/juju/core/network"
+	corenetwork "github.com/juju/juju/core/network"
+	"github.com/juju/juju/environs"
+	"github.com/juju/juju/environs/context"
+	"github.com/juju/juju/environs/instances"
+	"github.com/juju/names/v4"
+)
+
+var _ environs.Networking = (*environ)(nil)
```
achilleasa, line 20: I thought that we agreed to split the PR into two parts and land the `Networking` interface implementation as a separate PR.

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+		cp := d
+		if isListContained(packetTags, cp) {
+			toReturn = append(toReturn, &equnixDevice{e, &cp})
+		}
+	}
+
+	return toReturn, nil
+}
+
+func isListContained(tags []string, d packngo.Device) bool {
```
achilleasa, line 91: Please use `set.Strings` as suggested in the WIP PR instead of custom function. You can create two sets and check if `a.Intersection(B).IsEmpty()`

provider/equinix/environ.go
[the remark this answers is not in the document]
achilleasa: Wasn't there code in the WIP PR to populate this list?

provider/equinix/environ_network.go
jasmingacic: Yes we did, but when I was submitting the PR I realized that there are only two functions implemented in this file. Having a separate PR for so little felt like a overkill. 

If you want I can take this out of this PR

achilleasa: Yes, please remove that to reduce the PR size and make it easier to review for others. It is not required to test the basic provider behavior (bootstrap-deploy-teardown); the Networking interface enables support for more advanced features such as spaces and can be tested separately.

provider/equinix/environ.go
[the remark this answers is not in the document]
jasmingacic: not that I can recall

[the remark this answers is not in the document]
achilleasa: Oh I see now. 

Can you move the implementation of `e.supportedInstanceTypes()` here and change the call to `e.supportedInstanceTypes` in `StartInstance` to call `e.InstanceTypes` instead and just grab the `InstanceTypes` field from the result (`InstanceTypesWithCostMetadata` includes both the costs and the instance types). The code already pulls the cost values so you can also populate the relevant field in `InstanceTypesWithCostMetadata`

jasmingacic:
As suggested I added subnets in a separate PR #12847

provider/equinix/environ_network.go
jasmingacic: done

jasmingacic:
For some reason I cannot successfully fix the conflict above

provider/equinix/environ.go
jasmingacic: done

provider/equinix/environ.go
jasmingacic: I'd rather leave these values as they are

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+	for _, t := range tags {
+		tagFound := false
```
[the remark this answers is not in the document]
jasmingacic: I'd rather leave these values as they are

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+}
+
+func (e *environ) AllInstances(ctx context.ProviderCallContext) ([]instances.Instance, error) {
+	return e.getPacketInstancesByTag(map[string]string{"juju-model-uuid": e.Config().UUID()})
+}
+
+// if values tag and state are left empty it will return all instances
+func (e *environ) getPacketInstancesByTag(tags map[string]string) ([]instances.Instance, error) {
+	toReturn := []instances.Instance{}
+	packetTags := []string{}
```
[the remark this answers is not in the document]
jasmingacic: I'd rather leave these values as they are. I need those arrays already initialized

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+	"github.com/juju/juju/environs/simplestreams"
+	"github.com/juju/juju/provider/common"
+	"github.com/juju/juju/storage"
+	"github.com/juju/juju/tools"
+	"github.com/juju/os/series"
+	"github.com/juju/retry"
+	"github.com/juju/schema"
+	"github.com/juju/utils/arch"
+	"github.com/juju/version"
+	"github.com/lxc/lxd/shared/logger"
```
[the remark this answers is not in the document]
jasmingacic: @achilleasa what should be used here?

provider/equinix/environ.go
jasmingacic: The function only contains initializations and this logic. Is it really necessary to extract this into another function

provider/equinix/environ.go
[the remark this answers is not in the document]
jasmingacic: done

provider/equinix/environ.go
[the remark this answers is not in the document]
jasmingacic: @achilleasa I can't recall why we did this. Should we remove it?

provider/equinix/environ.go
```diff
@@ -0,0 +1,649 @@
...
+	}
+
+	// In order to spin up a new device, we must specify at least one
+	// public and one private address.
+	if !requestedPrivateAddr {
+		// Allocate a private address from the default address pool.
+		device.IPAddresses = append(device.IPAddresses, packngo.IPAddressCreateRequest{
+			Public:        false,
+			AddressFamily: 4,
+			CIDR:          31,
```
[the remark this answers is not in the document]
jasmingacic: I'm not sure if this will remain like this. So let us not change it until we are certain

provider/equinix/environ.go
SimonRichardson: Yes, because it will be more testable. Large functions/methods means your coverage for this is going to be low.

provider/equinix/environ.go
```diff
@@ -134,7 +134,7 @@ func (e *environ) ControllerInstances(ctx context.ProviderCallContext, controlle
 	if err != nil {
 		return nil, err
 	}
-	instanceIDs := []instance.Id{}
+	instanceIDs := make([]instance.Id, len(insts))
 	for _, i := range insts {
```
achilleasa, line 138: If you pre-allocate the slice, you should change your for loop as follows:

```go
for i, inst := range insts {
  instanceIDs[i] = inst.Id()
}
```

Alternatively, if you want to use append here, you can chance the slice allocation line to:

```
instanceIDs := make([]instance.Id, 0, len(insts))
```

provider/equinix/environ.go
```diff
@@ -335,7 +334,7 @@ func (e *environ) StartInstance(ctx context.ProviderCallContext, args environs.S
 	}
 
 	// Render the required tags for the instance.
-	packetTags := []string{}
+	packetTags := make([]string, len(args.InstanceConfig.Tags))
```
achilleasa, line 337: Same here: 

```
packetTags := make([]string, 0, len(args.InstanceConfig.Tags))
```

Otherwise, you will have `len(args.InstanceConfig.Tags)` empty values followed by the ones you append.

provider/equinix/environ.go
```diff
@@ -0,0 +1,665 @@
...
+	return i, nil
+}
+
+func (e *environ) Instances(ctx context.ProviderCallContext, ids []instance.Id) ([]instances.Instance, error) {
+	toReturn := []instances.Instance{}
+
+	tags := set.NewStrings("juju-model-uuid=" + e.Config().UUID())
+
+	for _, id := range ids {
+		//TODO handle case when some of the instanes are missing
```
achilleasa, line 196: This needs to be addressed in this PR. 

First, you need to pre-allocate a slice for the return values (instead of appending as the code below does). Then change the for loop (L195) to also use an index and use it to populate the appropriate index in the return value. Also, define a counter for missing instances outside the loop. If the client explicitly reports that a device is not found (i.e. **not** another kind of error occurred) just increment the counter and continue.

L206 can then be changed to something like this:

```
if missingInstanceCount > 0 {
   if missingInstanceCount == len(toReturn) {
      return nil, environs.ErrNoInstances
    } 

    return toReturn, environs.ErrPartialInstances
}
```

provider/equinix/environ.go
jasmingacic: I don't see point in allocating packetTags in this specific case. Maybe I should revert it to packetTags := []string{}

provider/equinix/environ.go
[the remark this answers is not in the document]
achilleasa: See the comment on L318. The images provided by equinix are custom and not the official ubuntu images. This has licensing issues (i.e calling ubuntu something that is not ubuntu) as well as practical issues; in particular, the images lack the lxd and snapd which are assumed to be present. In addition, the images include an odd script replacement for `/sbin/initctl` (it's an `exit 0` script!) which trips the init system detection logic used by juju (if initctl is installed and returns 0 it is assumed to be the active init system as per the initctl docs) thus requiring the extra patch in L312.

provider/equinix/environ.go
achilleasa: @SimonRichardson I wouldn't mind in this particular case as the logic here is self-contained and should be easy to test using a mocked packet client.

provider/equinix/environ.go
```diff
@@ -0,0 +1,665 @@
...
+			instances.InstanceType{
+				Id:       plan.ID,
+				Name:     plan.Name,
+				CpuCores: uint64(plan.Specs.Cpus[0].Count),
+				Mem:      mem,
+				Arches:   []string{instArch},
+				// Scale per hour costs so they can be represented as an integer for sorting purposes.
+				Cost: uint64(plan.Pricing.Hour * 1000.0),
+				// TODO: returned by packet's API but not exposed by the packngo client
+				// Deprecated: plan.Legacy,
```
achilleasa, line 499: @displague Any progress on adding this particular field to the client?

provider/equinix/environ.go
```diff
@@ -0,0 +1,665 @@
...
+		for _, os := range oss {
+			if !isDistroSupported(os, ic) {
+				continue
+			}
+
+			for _, p := range os.ProvisionableOn {
+				if p == it.Name {
+					image := &imagemetadata.ImageMetadata{
+						Id:   os.Slug,
+						Arch: arch.AMD64,
```
achilleasa, line 550: What about other arches? L478 indicates that arm64 is also supported.

provider/equinix/environ.go
```diff
@@ -0,0 +1,665 @@
...
+				return fmt.Errorf("device %s provisioning failed", id)
+			}
+			return nil
+		},
+		IsFatalError: func(err error) bool {
+			return common.IsCredentialNotValid(err)
+		},
+		Attempts: 180,
+		Delay:    5 * time.Second,
+		Clock:    clock.WallClock,
```
achilleasa, line 622: I would recommend passing `Clock` as an argument to the function. This will allow you to substitute it with a fake clock (from the clock/test sub-package) when testing this function in your tests.

provider/equinix/environ.go
```diff
@@ -0,0 +1,665 @@
...
+func (e *environ) AllInstances(ctx context.ProviderCallContext) ([]instances.Instance, error) {
+	return e.getPacketInstancesByTag(map[string]string{"juju-model-uuid": e.Config().UUID()})
+}
+
+// if values tag and state are left empty it will return all instances
+func (e *environ) getPacketInstancesByTag(tags map[string]string) ([]instances.Instance, error) {
+	toReturn := []instances.Instance{}
+	packetTags := []string{}
+
+	for k, v := range tags {
```
achilleasa, line 94: There is no need to prepare the slice and then create a set out of it. You could simplify that as follows:

```suggestion
        deviceTags := set.NewStrings()
	for k, v := range tags {
	   deviceTags.Add(fmt.Sprintf("%s=%s", k, v))
	 }
```

provider/equinix/environ.go
jasmingacic: What do you suggest we do about it?

achilleasa: Can we glean any arch-related information from `p` here? If not, and if packet guarantees that you can deploy distros to arm64 machines, you could probably generate two entries here? (Ideally, to future-proof the implementation, you could make the supported arches as a package-level variable and iterate it here creating an ImageMetadata entry for each supported arch.)

jasmingacic: we can't get much from `p` this is all the information I get from packngo.OS 
```
type OS struct {
	Name            string   `json:"name"`
	Slug            string   `json:"slug"`
	Distro          string   `json:"distro"`
	Version         string   `json:"version"`
	ProvisionableOn []string `json:"provisionable_on"`
}
``` 
Where `ProvisionableOn` is just an array of string with plan names  "c1.small.x86",  "baremetal_1",  "c1.xlarge.x86",
Maybe if we were to extract that data from plan which has property
```
 "cpus": [
        {
          "count": 1,
          "type": "Intel(R) Xeon(R) CPU E3-1578L v5 @ 2.00GHz"
        }
      ],
``` 
which we would have to normalize to extract correct architecture

achilleasa: Ok, so AFAICT, in one of the other methods we extract the arch from the plan name. You could extract that code into a helper that given the plan name, it returns an `arch.Arch` value and use the result (if any) to populate the `Arch` field in the generated `ImageMetadata` value. Would that do the trick?

jasmingacic: would this work
```
func getArchitectureFromPlan(p string) string {
	planSplit := strings.Split(p, ".")
	var architecture string
	if len(planSplit) > 3 {
		architecture = planSplit[2]
	}
	switch architecture {
	case "x86":
		return arch.AMD64
	case "arm":
		return arch.ARM64
	default:
		return arch.AMD64
	}
}
```

provider/equinix/environ.go
```diff
@@ -188,25 +187,34 @@ func (e *environ) InstanceTypes(context.ProviderCallContext, constraints.Value)
...
 
 	tags := set.NewStrings("juju-model-uuid=" + e.Config().UUID())
 
-	for _, id := range ids {
+	for i, id := range ids {
 		//TODO handle case when some of the instanes are missing
-		d, _, err := e.equnixClient.Devices.Get(string(id), nil)
-		if err != nil {
+		d, resp, err := e.equnixClient.Devices.Get(string(id), nil)
+		if err != nil && resp.Request.Response.StatusCode == 404 {
```
achilleasa, line 198: I believe `resp` is a pointer so this should probably read:

`if err != nil && resp != nil && resp.Request.Response.StatusCode == 404 {`

provider/equinix/environ.go
```diff
@@ -188,25 +187,34 @@ func (e *environ) InstanceTypes(context.ProviderCallContext, constraints.Value)
...
+			logger.Warningf("instance %s not found", string(id))
+			missingInstanceCount = missingInstanceCount + 1
+		} else if err != nil {
```
achilleasa, line 201: You need to `continue`. Otherwise you will panic trying to access pointer `d` a few lines below...

provider/equinix/environ.go
```diff
@@ -188,25 +187,34 @@ func (e *environ) InstanceTypes(context.ProviderCallContext, constraints.Value)
...
+			toReturn[i] = &equnixDevice{e, d}
 		}
 	}
-	if len(toReturn) == 0 {
-		return nil, environs.ErrNoInstances
+
+	if missingInstanceCount > 0 {
+		if missingInstanceCount == len(toReturn) {
+			return nil, environs.ErrNoInstances
+		}
```
achilleasa, line 214: This is wrong. If you don't return `ErrNoInstances` you should return `return toReturn, environs.ErrPartialInstances` here

provider/equinix/environ.go
```diff
@@ -188,25 +187,34 @@ func (e *environ) InstanceTypes(context.ProviderCallContext, constraints.Value)
...
 	}
-	return toReturn, nil
+
+	return toReturn, environs.ErrPartialInstances
```
achilleasa: This should be `return toReturn, nil`

provider/equinix/environ.go
achilleasa: Oh, nevermind; this got fixed in the followup commit...

provider/equinix/environ.go
```diff
@@ -0,0 +1,693 @@
...
+}
+
+func (e *environ) Instances(ctx context.ProviderCallContext, ids []instance.Id) ([]instances.Instance, error) {
+	toReturn := make([]instances.Instance, len(ids))
+	var missingInstanceCount int
+
+	tags := set.NewStrings("juju-model-uuid=" + e.Config().UUID())
+
+	for i, id := range ids {
+		//TODO handle case when some of the instanes are missing
```
achilleasa, line 196: The TODO can now be removed.

provider/equinix/environ.go
achilleasa: If you do revert it please change it into `var packetRags []string`. No need to allocate an empty slice; append will take care of this for you.

provider/equinix/environ.go
displague: Legacy is exposed - https://github.com/packethost/packngo/blob/master/plans.go#L25 in packngo 0.13.0

provider/equinix/environ.go
```diff
@@ -0,0 +1,693 @@
...
+	plans, _, err := e.equnixClient.Plans.List(opt)
+	if err != nil {
+		return nil, errors.Annotate(err, "retrieving supported instance types")
+	}
+
+	var instTypes []instances.InstanceType
+nextPlan:
+	for _, plan := range plans {
+		if !validPlan(plan, e.cloud.Region) {
+			logger.Infof("Plan %s not valid in facility %s", plan.Name, e.cloud.Region)
```
achilleasa, line 478: This should be logged at DEBUG level. Also, from a quick run against EWR1, I see that multiple plans get rejected as non-valid. Any ideas why?

```
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan Type 2A4 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan Type HUA not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan c1.bloomberg.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan c3.medium.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan c3.small.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan cpe1.g1.4028gr not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan cpe1.m2.r640 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan cpe1.s1.r730 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan cpe1.x1.supermicro not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan f3.large.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan f3.medium.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan g2.large.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan m2.xlarge.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan m3.large.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan s3.xlarge.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan Standard not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan Performance not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan Custom Storage Appliance not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan t3.small.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan x3.xlarge.x86 not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan x.large.arm not valid in facility ewr1
13:12:45 INFO  juju.provider.equnix environ.go:478 Plan x.small.x86 not valid in facility ewr1
```

Looks like only the `t1.small.x86` is available. Is that expected?

displague: `ewr1` has become an expansion site, reservations only.

`dc13` may be a better choice for On-Deman provisioning.

We may benefit from adapting this from a "facility" based deployment to a "metro" based deployment, to take advantage of broader availability. https://metal.equinix.com/developers/docs/locations/metros/#metros-quick-reference

achilleasa:
Closing this as it is superseded by #12983.

PR #12819, opened 2021-03-25
Fix agent lookup on upgraded 2.8 controllers

wallyworld:
When upgrading a 2.8 controller to 2.9, existing 2.8 models on that controller would try to ask for agents with version `2.8.9-ubuntu-amd64`. However, Juju only knows about series agents for 2.8. So in the metadata lookup and agent binary download, a compatibility step is added to look for any agent with a matching series and use that.

## QA steps

On 2.8.9
juju bootstrap
juju switch default
juju add-machine --series=bionic

WIth 2.9 client
^^^^ this would fail previously
check the machine gets provisioned and is shown as started

### DIFF
--- a/acceptancetests/assess_upgrade.py
+++ b/acceptancetests/assess_upgrade.py
@@ -72,7 +72,7 @@ def assess_upgrade_from_stable_to_develop(args, stable_bsm, devel_client):
             stable_client, base_dir, 'released')
         setup_agent_metadata(
             stream_server, args.devel_juju_agent,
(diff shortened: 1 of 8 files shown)


wallyworld:
[os based agent metadata part 3](https://trello.com/c/SYmfScWS/1476-os-based-agent-metadata-part-3)

hpidcock (approved):
Tested, LGTM

wallyworld:
$$merge$$

PR #12821, opened 2021-03-25
[2.9] Ensure openstack unit tests pass on non amd64 hosts

achilleasa:
This PR fixes two issues with the openstack unit tests:
- The `ensureAMDImages` method used by several tests in the localServerSuite used to iterate the supported series when uploading fake tool binaries. Since 2.9, tool binaries include the OS instead of the series thus causing the fake tool upload step to fail due to validation errors. It has been updated to use OS instead of series.
- When no bootstrap constraint is specified, juju uses the arch of the host running the client. As a result, the expectation in TestStartInstanceHardwareCharacteristics (as written, the test seems to be meant to test that an amd64 instance would be spinned up) would fail on non-amd64 machines. This was easily fixed by specifying an arch constraint

## QA steps

```console
# Run the openstack unit tests on a non-amd64 box
$ cd provider/openstack
$ go test
```

### DIFF
--- a/provider/openstack/local_test.go
+++ b/provider/openstack/local_test.go
@@ -513,7 +513,7 @@ func (s *localServerSuite) TestStartInstanceHardwareCharacteristics(c *gc.C) {
 	env := s.ensureAMDImages(c)
 	err := bootstrapEnv(c, env)
 	c.Assert(err, jc.ErrorIsNil)
-	_, hc := testing.AssertStartInstanceWithConstraints(c, env, s.callCtx, s.ControllerUUID, "100", constraints.MustParse("mem=1024"))
+	_, hc := testing.AssertStartInstanceWithConstraints(c, env, s.callCtx, s.ControllerUUID, "100", constraints.MustParse("mem=1024 arch=amd64"))
 	c.Check(*hc.Arch, gc.Equals, "amd64")
 	c.Check(*hc.Mem, gc.Equals, uint64(2048))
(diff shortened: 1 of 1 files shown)


provider/openstack/local_test.go
```diff
@@ -513,7 +513,7 @@ func (s *localServerSuite) TestStartInstanceHardwareCharacteristics(c *gc.C) {
 	env := s.ensureAMDImages(c)
 	err := bootstrapEnv(c, env)
 	c.Assert(err, jc.ErrorIsNil)
-	_, hc := testing.AssertStartInstanceWithConstraints(c, env, s.callCtx, s.ControllerUUID, "100", constraints.MustParse("mem=1024"))
+	_, hc := testing.AssertStartInstanceWithConstraints(c, env, s.callCtx, s.ControllerUUID, "100", constraints.MustParse("mem=1024 arch=amd64"))
```
SimonRichardson, line 516: I wonder why this just flagged up now? I would have thought we would have defaulted to amd64?

SimonRichardson (approved):

provider/openstack/local_test.go
achilleasa: https://github.com/juju/juju/blob/2.9/environs/bootstrap/bootstrap.go#L433-L445

It's been there for quite a while, hence my question about how this test could have been working before on non-amd64 boxes.

achilleasa:
!!build!!

achilleasa:
$$merge$$

PR #12827, opened 2021-03-26
[2.8] Unsubscribe from hub when closing state pool

achilleasa:
This PR fixes a goroutine leak in our tests. When the state pool opens, it subscribes to the hub (each subscriber runs in a separate goroutine). However, the subscribers were not terminated when the state pool closed thus causing a goroutine leak when running tests.

## QA steps

Just run the tests. The CI should do this for you.


SimonRichardson (approved):
Nice, I wonder if we ever do this in real code?

achilleasa:
$$merge$$

PR #12830, opened 2021-03-29
Fixes bug where model operator perms overwrite

tlm:
A bug was introduced where model operator cluster role and role binding
permissions would overwrite each other because they shared the same
name. This change makes it so they have unique names now.

## Checklist

## QA steps

Bootstrap Juju to a kubernetes cluster. Add two or models. On the first model added remove the model operator pod. Check the new pod coming up does not have any errors in the logs.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1921667

### DIFF
--- a/caas/kubernetes/provider/modeloperator.go
+++ b/caas/kubernetes/provider/modeloperator.go
@@ -65,6 +65,10 @@ type ModelOperatorBroker interface {
 	// exists in the targets cluster.
 	EnsureServiceAccount(*core.ServiceAccount) ([]func(), error)
 
+	// Model returns the name of the current model being deployed to for the
+	// broker
+	Model() string
+
 	// Namespace returns the current default namespace targeted by this broker.
 	Namespace() string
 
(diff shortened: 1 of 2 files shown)


caas/kubernetes/provider/modeloperator.go
```diff
@@ -523,18 +536,23 @@ func modelOperatorService(
 	}
 }
 
+func modelOperatorGlobalScopedName(model, operatorName string) string {
+	return fmt.Sprintf("%s-%s", model, operatorName)
```
wallyworld, line 540: this will be a bit meh if model is ""
(which the code expects cause it does a model nil check)
so let's add a small tweak here

wallyworld (approved):

tlm:
[model operator cluster role binding scoped names](https://trello.com/c/8Waqkfpz/1480-model-operator-cluster-role-binding-scoped-names)

tlm:
$$merge$$

PR #12831, opened 2021-03-29
Integration fixes

SimonRichardson:
The following changes ensure that the verbose setting when running
the integration tests to allow the real-time viewing of what's being currently 
run. Previously this was hidden until the very end of the test, which made
it really hard to develop against it locally when you have long integration
tests.

The code is simple, send the logs to a file and then using a background
process, just tail the log file. Although simple, it wasn't easy to get right
as first, as we had to ensure backward compatibility and ensure that
pipefail worked as designed.

As a driveby I fixed the inconsistency issue of one bracket vs two, now
that we're using bash.

## QA steps

Without verbose, this should be rather quiet:

```sh
(cd tests && ./main.sh smoke)
```

With verbose this should be really chatty:

```sh
(cd tests && ./main.sh -v smoke)
```

### DIFF
--- a/tests/includes/verbose.sh
+++ b/tests/includes/verbose.sh
@@ -11,11 +11,6 @@ set_verbosity() {
 		set -eu
 		set -o pipefail
 		;;
-	11)
-		# You asked for it!
-		set -eux
-		set -o pipefail
-		;;
 	*)
 		echo "Unexpected verbose level" >&2
 		exit 1
--- a/tests/includes/verbose.sh
+++ b/tests/includes/verbose.sh
@@ -5,13 +5,16 @@ set_verbosity() {
 	case "${VERBOSE}" in
 	1)
 		set -eu
+		set -o pipefail
 		;;
 	2)
 		set -eu
+		set -o pipefail
 		;;
 	11)
 		# You asked for it!
 		set -eux
+		set -o pipefail
 		;;
 	*)
 		echo "Unexpected verbose level" >&2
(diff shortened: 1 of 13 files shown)


hmlanigan (commented):
QA with `(cd tests && ./main.sh -v charmhub)`

The farther along in the test run, the more times a log line was repeated. 1, to 2, to 3 and finally 10:  https://pastebin.canonical.com/p/zQWgw9ZnNC/

In the pastern, some output is done wrong in the scripts, though unrelated to this PR I believe.  E.g `Success: "%s" > "%s" 4 0` instead of `Success: 4 > 0`

No issues with `(cd tests && ./main.sh smoke)` or `(cd tests && ./main.sh -v deploy)`  that jumped out.

tests/includes/verbose.sh
```diff
@@ -5,13 +5,16 @@ set_verbosity() {
 	case "${VERBOSE}" in
 	1)
 		set -eu
+		set -o pipefail
 		;;
 	2)
 		set -eu
+		set -o pipefail
 		;;
 	11)
```
hmlanigan, line 14: Unrelated: you can't set VERBOSE to 11 with just a flag to main.sh. What is it intended for?

SimonRichardson: I'll remove, it was from a very old method

SimonRichardson:
> Success: "%s" > "%s" 4 0 instead of Success: 4 > 0

@hmlanigan fix that as well.

hmlanigan:
====> Destroyed juju model test-charmhub-info
includes/run.sh: line 39: 216930 Killed                  tail -f "${TEST_DIR}/${TEST_CURRENT}.log" 2> /dev/null
===> [ ✔ ] Success: charmhub info (5s)

hmlanigan (approved):
The extra message is a bit confusing, but otherwise works

SimonRichardson:
$$merge$$

PR #12832, opened 2021-03-29
[2.9] Additional cross platform test fixes

achilleasa:
This PR includes a series of fixes that ensure that none of our unit tests depends on the OS that the tests are running on. 

More specifically, the changes in this PR hardcode the `Release` field of the `testing.CurrentVersion` helper to `ubuntu` instead of populating it with the value that corresponds to the host OS where the tests are running on. This is required as the vast majority of our test expectations (pretty much anything that uploads fake tools) assumes that we are running on an ubuntu host.

In addition, the PR updates some of the workers (e.g. upgrade, caasoperator etc.) to include hooks for allowing tests to patch the result of `os.HostOSTypeName` calls and therefore ensure that tests are not OS-dependent.

Finally, the PR includes a forward port of #12827 to avoid leaking goroutines when closing the StatePool during testing.

NOTE: these tests are failing on 2.9 because of the switch in versioning which now uses the OS type instead of a specific release type.

## QA steps

Run `make test` in a centos environment. I used an lxd instance for my tests but spinning up a centos box on ec2 and running the tests there (+ the typical CI boilerplate to setup the box) should do the trick.

### DIFF
--- a/cmd/juju/commands/upgrademodel_test.go
+++ b/cmd/juju/commands/upgrademodel_test.go
@@ -479,9 +479,14 @@ func (s *UpgradeBaseSuite) checkToolsUploaded(c *gc.C, vers version.Binary, agen
 	r.Close()
 	c.Check(err, jc.ErrorIsNil)
 	expectContent := version.Binary{
-		Number:  agentVersion,
-		Arch:    arch.HostArch(),
-		Release: coreos.HostOSTypeName(),
+		Number: agentVersion,
+		Arch:   arch.HostArch(),
+		// All tests that upload fake agent tool binaries assume that
+		// we are always running on ubuntu. Unfortunately, this
+		// assumption breaks tests when running the suites on non-ubuntu
+		// platform (e.g. centos). Instead of trying to auto-detect the
+		// current OS, we force the release to "ubuntu".
+		Release: strings.ToLower(coreos.Ubuntu.String()),
 	}
 	checkToolsContent(c, data, "jujud contents "+expectContent.String())
 }
--- a/cmd/jujud/agent/bootstrap_test.go
+++ b/cmd/jujud/agent/bootstrap_test.go
@@ -123,7 +123,16 @@ func (s *BootstrapSuite) SetUpTest(c *gc.C) {
 	s.makeTestModel(c)
 
 	// Create fake tools.tar.gz and downloaded-tools.txt.
+	//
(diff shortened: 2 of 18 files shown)


cmd/juju/commands/upgrademodel_test.go
```diff
@@ -479,9 +479,14 @@ func (s *UpgradeBaseSuite) checkToolsUploaded(c *gc.C, vers version.Binary, agen
...
-		Arch:    arch.HostArch(),
-		Release: coreos.HostOSTypeName(),
+		Number: agentVersion,
+		Arch:   arch.HostArch(),
+		// All tests that upload fake agent tool binaries assume that
+		// we are always running on ubuntu. Unfortunately, this
+		// assumption breaks tests when running the suites on non-ubuntu
+		// platform (e.g. centos). Instead of trying to auto-detect the
+		// current OS, we force the release to "ubuntu".
+		Release: strings.ToLower(coreos.Ubuntu.String()),
```
wallyworld, line 489: This is better fixed in the test setup as per #12834

cmd/jujud/agent/bootstrap_test.go
```diff
@@ -123,7 +123,16 @@ func (s *BootstrapSuite) SetUpTest(c *gc.C) {
 	s.makeTestModel(c)
 
 	// Create fake tools.tar.gz and downloaded-tools.txt.
+	//
+	// CurrentVersion() currently hardcodes Release to "ubuntu" to match
+	// the expectations of other test suites wrt. the underlying host where
+	// the tests are running. However, this test is an exception as we
+	// are testing the bootstrap flow which uses the actual OS type to
+	// select agent binaries. Therefore, we need to specify the appropriate
+	// release value here.
```
wallyworld, line 132: Bootstrap is only supported on ubuntu so we should test with agent metadata set to ubuntu rather than the host os on which the test is run

wallyworld:
#12834 fixes some of the remaining root causes of the non-ubuntu failures.
I've been working through the fallout from the switch to os based stream metadata and am happy to keep doing so to address the remaining issues.

achilleasa:
Closing this as it overlaps with the changes in #12834 that have already landed.

@wallyworld note that some of the workers (see last commit from this PR) make use of HostOSTypeName and cause the tests to fail on centos. The last commit allows tests to override that so you may want to cherry-pick some of those changes.

PR #12835, opened 2021-03-30
Use charm url name when migrating charms.

SimonRichardson:
During a migration it is expected that a charm name in the metadata is
the same one in the URL. This isn't always the case. In-fact it might
never be the case for CharmHub charms, as the charmhub store does a lot
of charm url re-writing.

To this end, we allow the passing in of the charm name during a
migration and fallback to the one in metadata if it's empty.

---

The code is TERRIBLE in this file. There is no logical separation, just
massive freaking function of code. It's impossible to test at a unit
level and as kat-co pointed out years ago this handler setup should be
broken up.

I've added my 2-cents to the "fixme" at the top of the file and we could
use something like a Middleware pattern for this, so we can test the
various pathways to have a better and more secure API server.

---

Fixes: LP#1921722

## QA steps

```sh
$ juju bootstrap lxd src
$ juju bootstrap lxd dst
$ juju switch src
$ juju add-model src-model
$ juju deploy ubuntu
$ juju deploy cs:~logrotate-charmers/logrotate-charm-5
$ juju add-relation ubuntu logrotated # notice this isn't logrotate, but logrotated!
$ juju migrate src-model dst
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1921722

### DIFF
--- a/api/migrationtarget/client.go
+++ b/api/migrationtarget/client.go
@@ -75,12 +75,17 @@ func (c *Client) Activate(modelUUID string) error {
 // to add the charm binary to the model specified.
 func (c *Client) UploadCharm(modelUUID string, curl *charm.URL, content io.ReadSeeker) (*charm.URL, error) {
 	args := url.Values{}
+	args.Add("name", curl.Name)
 	args.Add("schema", curl.Schema)
 	args.Add("arch", curl.Architecture)
 	args.Add("user", curl.User)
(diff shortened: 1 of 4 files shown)


manadart (approved):
Looks good; WAD.

SimonRichardson:
$$merge$$

PR #12837, opened 2021-03-31
Fix the remaining tests which fail on centos

wallyworld:
centos unit test runs had some failures.

## QA steps

run the failing test suites in centos


wallyworld:
$$merge$$

PR #12842, opened 2021-04-01
Fix upgrade series agent version handling

wallyworld:
When doing an upgrade-series, there's no need to copy across agent binaries to the new series as everything is now "ubuntu".
So this operation is removed.

When, when upgrading from 2.8, the controller agent itself still is recorded as running the original "series" agent. This is because it is the old 2.8 agent that write out the new agent symlink prior to exiting. However, this breaks upgrade series on the upgraded controller so add a check to the upgrade worker so that older agent binaries are copied. The code to do this is a cut down copy of what was deleted from upgrade series.

## QA steps

Using Juju 2.8, bootstrap controller, switch to controller model, eg:
juju bootstrap lxd --bootstrap-series bionic
juju switch controller

Using Juju 2.9, this branch (gh pr checkout 12842):
juju deploy cs:ubuntu --series bionic --to 0
juju status
juju ssh 0 "ls -l /var/lib/juju/tools"
(both machine-0 and ubuntu-0 should point to the ubuntu tools)

juju switch default
juju upgrade-model
juju deploy cs:ubuntu --series bionic
juju upgrade-series 0 prepare focal
check logs for errors
juju status
juju ssh 0 "ls -l /var/lib/juju/tools"
(both machine-0 and ubuntu-0 should point to the ubuntu tools)

### DIFF
--- a/worker/upgrader/upgrader.go
+++ b/worker/upgrader/upgrader.go
@@ -12,7 +12,10 @@ import (
 
 	"github.com/juju/errors"
 	"github.com/juju/names/v4"
+	"github.com/juju/os/v2/series"
 	"github.com/juju/utils/v2/arch"
+	"github.com/juju/utils/v2/fs"
+	"github.com/juju/utils/v2/symlink"
 	"github.com/juju/version/v2"
 	"github.com/juju/worker/v2/catacomb"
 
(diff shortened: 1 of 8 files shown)


worker/upgrader/upgrader.go
```diff
@@ -116,6 +119,68 @@ func AllowedTargetVersion(
 	return true
 }
 
+func (u *Upgrader) maybeCopyAgentBinary(dataDir, hostSeries string) (err error) {
```
wallyworld, line 122: moved from systemdServiceManager above

benhoyt (approved):
Looks good! Per discussion, I tested locally on lxd and tweaked the QA steps to match.

wallyworld:
[upgrade-series fix](https://trello.com/c/gQkutKff/1490-upgrade-series-fix)

wallyworld:
$$merge$$

hpidcock:
$$merge$$

hpidcock:
$$merge$$

hpidcock:
$$merge$$

hpidcock:
$$merge$$

PR #12845, opened 2021-04-01
[2.9] Restore space support for manual machines

achilleasa:
This PR fixes a networking-related bug which prevents bootstraps of manual machines on 2.9 and also enhances the fidelity of address-related information that gets captured by the machiner worker.

At some point along its execution, the machiner worker retrieves a list of observed network interfaces and sends it to the controller. When using netlink as our address source, it reports interface addresses (on manual machines) as `net.IPNet` values which represent the interface addresses in _CIDR notation-, e.g. `10.1.2.99/24`.

The controller, uses this information to populate the known subnet details as a workaround given that the manual provider does not support native subnet discovery.  However, storing a non-network CIDR is stored as the subnet CIDR causes issues when the machiner contacts the controller to report the machine addresses visible to the worker.

In particular, the machiner worker periodically makes an API call to `SetMachineAddresses`. The facade implementation first attempts to convert the input arguments into a`SpaceAddresses list`. However, that particular code path eventually calls
`core/network/SubnetInfos.GetByUnderlayCIDR` with the incorrect subnet CIDR which fails validation thus causing an error to be returned back to the machiner worker. The error then causes the machiner to get stuck in a restart loop.

To fix this issue, a new helper called `NetworkCIDRFromIPAndNetmask` has been added to the `core/network` package. This allows us to properly convert host addresses in CIDR notation into network CIDRS (e.g. 10.1.2.3/24 -> 10.1.2.0/24) and avoid the validation errors.

## QA steps

Deploy a bionic MAAS machine with NICs in two spaces. In the example below, the machine is accessible at 10.0.0.75

```console
$ juju bootstrap manual/ubuntu@10.0.0.75 manual-test --debug --no-gui

$ juju switch controller

# You should get a list of spaces/subnets
$ juju subnets
juju subnets
subnets:
  10.0.0.0/24:
    type: ipv4
    status: in-use
    space: alpha
    zones: []
  10.42.0.0/24:
    type: ipv4
    status: in-use
    space: alpha
    zones: []

$ juju add-space space1 10.42.0.0/24
added space "space1" with subnets 10.42.0.0/24

$ juju spaces
Name    Space ID  Subnets
alpha   0         10.0.0.0/24
space1  1         10.42.0.0/24

$ juju deploy cs:wordpress --to 0 --bind space1 --series bionic

# The charm should deploy as expected
```

## Bug reference
https://bugs.launchpad.net/juju/+bug/1922098

### DIFF
--- a/api/common/network.go
+++ b/api/common/network.go
@@ -186,8 +186,9 @@ func interfaceAddressToNetworkConfig(
 	}
 
 	if ipNet := addr.IPNet(); ipNet != nil && ipNet.Mask != nil {
-		config.CIDR = ipNet.String()
+		config.CIDR = network.NetworkCIDRFromIPAndMask(ip, ipNet.Mask)
 	}
+
 	config.Address = ip.String()
 	if configType != string(network.ConfigLoopback) {
 		config.ConfigType = string(network.ConfigStatic)
--- a/worker/machiner/machiner.go
+++ b/worker/machiner/machiner.go
@@ -125,16 +125,29 @@ func setMachineAddresses(tag names.MachineTag, m Machine) error {
 	}
 	var hostAddresses corenetwork.ProviderAddresses
 	for _, addr := range addrs {
-		var ip net.IP
+		var (
+			ip      net.IP
+			netmask net.IPMask
+		)
+
 		switch addr := addr.(type) {
 		case *net.IPAddr:
 			ip = addr.IP
 		case *net.IPNet:
 			ip = addr.IP
+			netmask = addr.Mask
 		default:
 			continue
 		}
-		address := corenetwork.NewProviderAddress(ip.String())
+
+		// If we discovered a netmask for the address, include a CIDR
+		// when constructing the provider address.
+		var addrOpts []func(corenetwork.AddressMutator)
+		if cidr := corenetwork.NetworkCIDRFromIPAndMask(ip, netmask); cidr != "" {
+			addrOpts = append(addrOpts, corenetwork.WithCIDR(cidr))
+		}
+		address := corenetwork.NewProviderAddress(ip.String(), addrOpts...)
+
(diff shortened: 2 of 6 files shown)


api/common/network.go
```diff
@@ -186,8 +186,14 @@ func interfaceAddressToNetworkConfig(
 	}
 
 	if ipNet := addr.IPNet(); ipNet != nil && ipNet.Mask != nil {
-		config.CIDR = ipNet.String()
+		config.CIDR = network.NetworkCIDRFromIPAndMask(ip, ipNet.Mask)
+	} else {
+		// NOTE(achilleasa): DefaultMask() can return a nil mask; if
+		// that's the case, NetworkCIDRFromIPAndMask will simply return
+		// an empty CIDR.
+		config.CIDR = network.NetworkCIDRFromIPAndMask(ip, ip.DefaultMask())
```
achilleasa, line 194: I might be trying to be too smart here but I think they will typically be /24 so this should work as expected.

worker/machiner/machiner.go
```diff
@@ -125,16 +125,30 @@ func setMachineAddresses(tag names.MachineTag, m Machine) error {
...
-		var ip net.IP
+		var (
+			ip      net.IP
+			netmask net.IPMask
+		)
+
 		switch addr := addr.(type) {
 		case *net.IPAddr:
 			ip = addr.IP
+			netmask = ip.DefaultMask() // may return nil if not ipv4 address
```
achilleasa, line 136: I might be trying to be too smart here but I think they will typically be /24 so this should work as expected.

manadart (commented):
Code looks good; will run the QA next week.

manadart (approved):
Works well.

achilleasa:
$$merge$$

hpidcock:
$$merge$$

PR #12854, opened 2021-04-08
Fix for LP1921557 sni in Juju login.

tlm:
Juju login has been failing with recent changes to Juju's certificate
management. Specifically there are two problems that have come up.

The first is the Juju controller selecting the right certificate when ip
connections are being initiated that don't have an SNI name set. This
has been fixed by returning Juju's ip certificate with empty SNI names.

The second is Juju returning the CA cert in it's certificate chain.
Normally it's not considered valid to return a root CA in a chain but
for the way Juju works when logging in it needs to in order to trust the
CA and add it to the config file.

## Checklist

## QA steps

Follow the steps outline in the bug report below. Also perform the below check with openssl.

```sh
openssl s_client --connect <controller-ip>:17070 -servername <controller-ip>
You need to confirm here that a certificate is returned covering the ip sans of the controller and that CA cert is returned in the chain.
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1921557

### DIFF
--- a/apiserver/facades/controller/caasoperatorprovisioner/provisioner_test.go
+++ b/apiserver/facades/controller/caasoperatorprovisioner/provisioner_test.go
@@ -257,7 +257,7 @@ func (s *CAASProvisionerSuite) TestIssueOperatorCertificate(c *gc.C) {
 		[]byte(certInfo.PrivateKey)...))
 	c.Assert(err, jc.ErrorIsNil)
 	c.Assert(len(signers), gc.Equals, 1)
-	c.Assert(len(certs), gc.Equals, 1)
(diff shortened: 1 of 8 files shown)


wallyworld (approved):
LGTM so long as the logic allows older juju clients to still work against this controller

tlm:
[1921557 2.9 login to new controller fails TLS](https://trello.com/c/Q7NeZpFV/1491-1921557-29-login-to-new-controller-fails-tls)

tlm:
$$merge$$

tlm:
$$merge$$

PR #12866, opened 2021-04-13
Workaround for k8s dashboard URL with k8s client proxy

wallyworld:
The change to route CLI requests for k8s via a local client proxy broke the `juju dashboard` command since the CLI considers the API address it connects to as `localhost` and that's what's printed as the dashboard URL.

This PR is a workaround for the majority of cases where the controller API address is reachable (the dashboard is served from the controller). If the API address is `localhost`, the dashboard command looks up the actual controller IP address and prints that instead. This won't work for minikube or kind where the client proxy is needed, but fixes the regression on microk8s and most other k8s clouds.

## QA steps

bootstrap microk8s
```
$ juju gui
Dashboard 0.3.0 for controller "foo" is enabled at:
  https://10.152.183.43:17070/dashboard
Your login credential is:
  username: admin
  password: 95b82c68f82c7e3e42297082c81183a6
```

Also tested on LXD for completeness.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1923141

### DIFF
--- a/cmd/juju/gui/gui.go
+++ b/cmd/juju/gui/gui.go
@@ -6,6 +6,7 @@ package gui
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


tlm (approved):
lgtm

wallyworld:
$$merge$$

PR #12879, opened 2021-04-15
Ensure refresh uses version

SimonRichardson:
The following ensures that we handle the version correctly when
installing a charm. Currently, we get a series from the charm URL and we
actually need the version. The following code gets the version from the
series and drops the channel risk on the floor.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju add-model seven --config charmhub-url="https://api.staging.charmhub.io"
$ juju deploy facundo-snappass-test  
```

### DIFF
--- a/apiserver/facades/client/charms/charmhubrepo.go
+++ b/apiserver/facades/client/charms/charmhubrepo.go
@@ -73,7 +73,7 @@ func (c *chRepo) ResolveWithPreferredChannel(curl *charm.URL, origin corecharm.O
 			fallthrough
 		case transport.ErrorCodeInvalidCharmBase:
 			logger.Tracef("Invalid charm platform %q %v - Default Platforms: %v", curl, origin, resErr.Extra.DefaultBases)
-			platform, err := c.selectNextPlatform(resErr.Extra.DefaultBases, origin)
+			platform, err := c.selectNextBase(resErr.Extra.DefaultBases, origin)
 			if err != nil {
 				return nil, corecharm.Origin{}, nil, errors.Annotatef(err, "refresh")
 			}
@@ -114,6 +114,7 @@ func (c *chRepo) ResolveWithPreferredChannel(curl *charm.URL, origin corecharm.O
 			return nil, corecharm.Origin{}, nil, errors.NotValidf("series for %s", curl.Name)
 		}
 
+		logger.Tracef("Refresh again with %q %v", curl, input)
 		res, err = c.refreshOne(curl, input)
 		if err != nil {
 			return nil, corecharm.Origin{}, nil, errors.Annotatef(err, "refresh retry")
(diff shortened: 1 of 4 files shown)


hmlanigan:
Looks okay in the microk8s world.  The charm installed and is active.

hmlanigan:
As @SimonRichardson found, we need to track down why this charm was able to look like it's deploying on lxd, which it shouldn't

hmlanigan (approved):
LGTM,

hmlanigan:
$$merge$$

apiserver/facades/client/charms/charmhubrepo.go
```diff
@@ -370,14 +371,24 @@ func refreshConfig(curl *charm.URL, origin corecharm.Origin) (charmhub.RefreshCo
...
+
 	var (
 		cfg charmhub.RefreshConfig
-		err error
 
 		platform = charmhub.RefreshPlatform{
 			Architecture: origin.Platform.Architecture,
 			OS:           origin.Platform.OS,
-			Series:       origin.Platform.Series,
+			Series:       version,
```
jameinel, line 391: Why are we passing Series here vs Base?
It feels like we are spending a lot of effort to map Platform back to series in order to refresh with series, but Charmhub is passing us back
```
$ curl -XPOST -s https://api.charmhub.io/v2/charms/refresh -H 'Content-type: application/json' -d '{"context": [], "actions": [{"name": "facundo-snappass-test", "base": {"architecture": "amd64", "name": "NA", "channel": "NA"}, "action": "install", "instance-key": "a-test"}]}'  | jq
{
  "error-list": [],
  "results": [
    {
      "charm": null,
      "error": {
        "code": "invalid-charm-platform",
        "extra": {
          "default-bases": [
            {
              "architecture": "amd64",
              "channel": "20.04",
              "name": "ubuntu"
            }
          ],
          "default-platforms": [
            {
              "architecture": "amd64",
              "os": "ubuntu",
              "series": "focal"
            }
          ]
        },
        "message": "Instance key 'a-test' invalid 'platform' in 'install' for charm_id=C3cqgiAPo0liBr4vwGUzTFl297c5SpMs name=facundo-snappass-test"
      },
      "id": "C3cqgiAPo0liBr4vwGUzTFl297c5SpMs",
      "instance-key": "a-test",
      "name": "facundo-snappass-test",
      "released-at": null,
      "result": "error"
    }
  ]
}
```

Which does have platform, but it also has bases, which we can pass back in.

hmlanigan: This is part of the boundary of where we're using series vs base in juju.  Eventually RefreshPlatform must become RefreshBase, but currently that's hidden inside of charmhub pkg.

Though it brings up a good question of why the change charmhub wasn't enough for this, perhaps that should have been tweaked instead.

hmlanigan: Perhaps this should have been addressed in here: https://github.com/juju/juju/blob/d25ae7239f01873221fc75cf510b553168c1cdff/charmhub/refresh.go#L443

SimonRichardson: I can fix this tomorrow morning

apiserver/facades/client/charms/charmhubrepo.go
```diff
@@ -472,6 +483,17 @@ func selectReleaseByArchAndChannel(releases []transport.Release, origin corechar
 	return Release{}, errors.NotFoundf("release")
 }
 
+func channelTrack(channel string) string {
+	// Base channel can be found as either just the version `20.04` (focal)
+	// or as `20.04/latest` (focal latest). We should future proof ourself
+	// for now and just drop the risk on the floor.
+	parts := strings.Split(channel, "/")
```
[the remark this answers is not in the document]
SimonRichardson: I'll validate based on the risk tomorrow morning

PR #12884, opened 2021-04-16
Fix ssh with sidecar charms and containers.

hpidcock:
Allow `juju ssh` with sidecar charms.

## QA steps

`juju ssh unit/0` should ssh to charm container
`juju ssh --container my-container unit/0` should ssh to the specified container which is in charm metadata.yaml
`juju ssh --container missing-container unit/0` should fail nicely and mention the available containers

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1923847


hpidcock:
[2.9-rc11 - ssh to sidecar charm fails](https://trello.com/c/GVqTD19U/1529-29-rc11-ssh-to-sidecar-charm-fails)

apiserver/facades/client/charms/client.go
```diff
@@ -154,40 +154,6 @@ func NewCharmsAPI(
 	}, nil
 }
 
-// CharmInfo returns information about the requested charm.
```
wallyworld, line 157: I thought this was still used then I saw we had a common func so good to see this go

wallyworld (approved):
Did you test scp also?

apiserver/facades/client/charms/client.go
hpidcock: > Did you test scp also?

I'll do that before I land

hpidcock:
$$merge$$

PR #12891, opened 2021-04-19
Update charn/v8 and remove use of juju/version

wallyworld:
Bring in a new charm/v8 upstream which uses juju/version/v2 so we can also remove juju/version from juju also.

## QA steps

unit tests


SimonRichardson (approved):

wallyworld:
$$merge$$

PR #12896, opened 2021-04-19
k8s charm url might not have a series

SimonRichardson:
The k8s charm url might not have the k8s url in it during a set charm,
if that's the case we need to check the meta data to ensure that it
actually is.

I've remove the if from the if/else as I think we should always check
it, but I might be wrong there?

## QA steps

```sh
TBA
```

### DIFF
--- a/state/application.go
+++ b/state/application.go
@@ -1456,7 +1456,7 @@ func (a *Application) SetCharm(cfg SetCharmConfig) (err error) {
 		return errors.Trace(err)
 	}
 	if cfg.Charm.Meta().Deployment != currentCharm.Meta().Deployment {
-		if currentCharm.Meta().Deployment == nil || currentCharm.Meta().Deployment == nil {
+		if cfg.Charm.Meta().Deployment == nil || currentCharm.Meta().Deployment == nil {
 			return errors.New("cannot change a charm's deployment info")
 		}
 		if cfg.Charm.Meta().Deployment.DeploymentType != currentCharm.Meta().Deployment.DeploymentType {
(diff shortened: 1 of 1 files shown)


state/application.go
```diff
@@ -1466,15 +1466,27 @@ func (a *Application) SetCharm(cfg SetCharmConfig) (err error) {
...
 	// For old style charms written for only one series, we still retain
 	// this check. Newer charms written for multi-series have a URL
 	// with series = "".
-	if cfg.Charm.URL().Series != "" {
-		if cfg.Charm.URL().Series != a.doc.Series {
-			return errors.Errorf("cannot change an application's series")
+	if charmURLSeries := cfg.Charm.URL().Series; charmURLSeries != "" {
+		// If a doc series is kubernetes, but the url isn't kubernetes, for
+		// example the url could be the latest LTS, we therefore check if it
+		// the charm is kubernetes from the meta series and meta container.
```
hmlanigan, line 1476: nit: "if it the charm is"

state/application.go
```diff
@@ -1492,6 +1504,7 @@ func (a *Application) SetCharm(cfg SetCharmConfig) (err error) {
 			}
 			return errors.Errorf("only these series are supported: %v", supportedSeries)
 		}
+
```
hmlanigan, line 1507: nit: odd spot for a new line

state/application.go
```diff
@@ -1456,7 +1456,7 @@ func (a *Application) SetCharm(cfg SetCharmConfig) (err error) {
 		return errors.Trace(err)
 	}
 	if cfg.Charm.Meta().Deployment != currentCharm.Meta().Deployment {
-		if currentCharm.Meta().Deployment == nil || currentCharm.Meta().Deployment == nil {
+		if cfg.Charm.Meta().Deployment == nil || currentCharm.Meta().Deployment == nil {
```
hmlanigan, line 1459: 👍

PR #12899, opened 2021-04-19
CAAS bundle deployments

SimonRichardson:
As we don't have the kubernetes in the charm URL we have to then
fallback to the charm metadata. In this instances we can do that and we
can inspect it.

The great thing about this change is that is suppliments the existing
workflow of checking the computed series, so this should work with old
style charms. The worst case is that you might end up with checking the
same series twice, but that's always been the case.

## QA steps

bundle.yaml:
```
bundle: kubernetes
applications:
  snappass:
    charm: facundo-snappass-test
    scale: 1
```

```console
$ juju bootstrap microk8s micro

$ juju deploy ./bundle.yaml
Located charm "facundo-snappass-test" in charm-hub
Executing changes:
- upload charm facundo-snappass-test from charm-hub for series kubernetes
- deploy application snappass from charm-hub with 1 unit on kubernetes using facundo-snappass-test
  added resource redis-image
  added resource snappass-image
Deploy of bundle completed.
```

### DIFF
--- a/cmd/juju/application/deployer/bundlehandler.go
+++ b/cmd/juju/application/deployer/bundlehandler.go
@@ -42,6 +42,7 @@ import (
 	"github.com/juju/juju/core/instance"
 	"github.com/juju/juju/core/lxdprofile"
 	"github.com/juju/juju/core/model"
+	"github.com/juju/juju/core/series"
 	"github.com/juju/juju/environs/config"
 	"github.com/juju/juju/resource/resourceadapters"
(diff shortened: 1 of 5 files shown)


hmlanigan:
tiny change to a commit message to get the checks running

hmlanigan:
$$merge$$

hmlanigan (approved):
QA was good.

hmlanigan:
$$merge$$

PR #12901, opened 2021-04-19
charm Format helper

hmlanigan:
Add a helper to core charm to determining what metadata format the charm is using.

## QA steps
no change to unit test results


wallyworld (approved):

core/charm/format.go
```diff
@@ -0,0 +1,28 @@
...
+package charm
+
+import "github.com/juju/charm/v8"
+
+// MetadataFormat of the parsed charm.
+type MetadataFormat int
+
+// MetadataFormat are the different versions of charm metadata supported.
+const (
+	FormatUnknown MetadataFormat = iota
```
jameinel, line 13: If there is any chance that this can leak out of the package (into a database, into a wire protocol, etc), I'm not a fan of "iota", because it makes it seem like the numbers aren't significant, when they very much are.

wallyworld: fwiw this is not used on the wire or in a db

hmlanigan: No, it will not leak into the db or across the wire.  We act on it immediately.  The charm goes over the wire, or in the db.

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #12910, opened 2021-04-21
Update snapcraft.yaml

pedroleaoc:
- Update to be aligned with the new terminology

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -2,17 +2,13 @@ name: juju
 version: 3.0-beta1
 summary: Juju - a model-driven operator lifecycle manager for K8s and machines
 description: |
-  A model-driven **universal operator lifecycle manager** for multi cloud and hybrid cloud application management on K8s and machines.
+  Juju is an Open Source Charmed Operator Framework. It helps you move from configuration management to application management and has two main components:
 
-  **What is an operator lifecycle manager?**
-  Kubernetes operators are containers with operations code, that drive your applications on K8s. Juju is an operator lifecycle manager that manages the installation, integration and configuration of operators on the cluster. Juju also extends the idea of operators to traditional application management on Linux and Windows servers, or cloud instances.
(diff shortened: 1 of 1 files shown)


SimonRichardson:
@pedroleaoc I'm actually guessing you want to target the 2.9 branch and not develop.

jameinel:
just targeting 2.9 brings in all of develop, will need to be rebased.

snap/snapcraft.yaml
```diff
@@ -2,17 +2,13 @@ name: juju
...
+  Juju is an Open Source Charmed Operator Framework. It helps you move from configuration management to application management and has two main components:
 
-  **What is an operator lifecycle manager?**
-  Kubernetes operators are containers with operations code, that drive your applications on K8s. Juju is an operator lifecycle manager that manages the installation, integration and configuration of operators on the cluster. Juju also extends the idea of operators to traditional application management on Linux and Windows servers, or cloud instances.
+  ● **Charmed Operator Lifecycle Manager (OLM)** - a hybrid-cloud application management and orchestration system that helps you from Day 0 to Day 2. Deploy, configure, scale, **integrate**, maintain and manage Kubernetes native, container-native and VM-native applications -- and the relations between them.
+  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;● **Charmed Operators, packaged as “Charms”**, are software that encapsulate a single application and all the code and know-how it takes to operate it on Kubernetes or machines.
```
jameinel, line 8: I didn't think snapcraft.yaml description was meant to be HTML.

This, in particular, seems to be using a Unicode bullet point, and &nbsp; to do indentation, but then uses **Foo** which would be markdown bold, etc.
At the very least I don't think we'd want to mix them.
I asked to see whether description is meant to be markdown or html or something else.

jameinel: Facundo was kind enough to track this down:
https://snapcraft.io/docs/store-listing-branding

It seems for the website, general Markdown is supported. Which we should do rather than html/unicode characters.

pedroleaoc: I thought that most markdown implementations would support basic HTML tags. Good point about the bullet point, thanks!
I will update it to pure MD.

SimonRichardson:
Closing this because of no activity, please reopen if you want to move this forward.

PR #12911, opened 2021-04-21
improve bundle messaging output

hmlanigan:
Add channel if available to "Locating charm" message while deploy a bundle.

When deploying a bundle using the same charm charm multiple times, output makes it appear that the same charm is located more than once.  We do print out a revision for charmstore charms, so add the channel to the output if available and there is no revision.  Unfortunately, at the point in time of message display, the name used for the application is not available.

## QA steps

```console
$ juju deploy juju-qa-bundle-test
Located bundle "juju-qa-bundle-test" in charm-hub, revision 2
Located charm "juju-qa-test" in charm-hub, channel 2.0/stable
Located charm "juju-qa-test" in charm-hub, channel candidate
Located charm "ntp" in charm-hub, channel stable
Located charm "ntp" in charm-hub, channel stable
Executing changes:
- upload charm juju-qa-test from charm-hub for series bionic from channel 2.0/stable with architecture=amd64
- deploy application juju-qa-test from charm-hub on bionic with 2.0/stable
  added resource foo-file
- set annotations for juju-qa-test
- upload charm juju-qa-test from charm-hub for series focal from channel latest/candidate with architecture=amd64
- deploy application juju-qa-test-focal from charm-hub on focal with latest/candidate using juju-qa-test
  added resource foo-file
- set annotations for juju-qa-test-focal
- upload charm ntp from charm-hub for series bionic from channel stable
- deploy application ntp from charm-hub on bionic with stable
- set annotations for ntp
- upload charm ntp from charm-hub for series focal from channel stable
- deploy application ntp-focal from charm-hub on focal with stable using ntp
- set annotations for ntp-focal
- add new machine 0
- add new machine 1
- add relation ntp:juju-info - juju-qa-test:juju-info
- add relation ntp-focal:juju-info - juju-qa-test-focal:juju-info
- add unit juju-qa-test/0 to new machine 0
- add unit juju-qa-test-focal/0 to new machine 1
Deploy of bundle completed.
```

### DIFF
--- a/cmd/juju/application/deployer/charm.go
+++ b/cmd/juju/application/deployer/charm.go
@@ -552,9 +552,11 @@ func formatLocatedText(curl *charm.URL, origin commoncharm.Origin) string {
 	if repository == "" || repository == commoncharm.OriginLocal {
 		return fmt.Sprintf("Located local charm %q, revision %d", curl.Name, curl.Revision)
 	}
-	var revision string
+	var next string
 	if curl.Revision != -1 {
-		revision = fmt.Sprintf(", revision %d", curl.Revision)
+		next = fmt.Sprintf(", revision %d", curl.Revision)
+	} else if str := origin.CharmChannel().String(); str != "" {
+		next = fmt.Sprintf(", channel %s", str)
 	}
-	return fmt.Sprintf("Located charm %q in %s%s", curl.Name, repository, revision)
+	return fmt.Sprintf("Located charm %q in %s%s", curl.Name, repository, next)
 }
--- a/cmd/juju/application/deployer/bundlehandler_test.go
+++ b/cmd/juju/application/deployer/bundlehandler_test.go
@@ -367,13 +367,13 @@ func (s *BundleDeployRepositorySuite) TestDeployKubernetesBundleSuccessWithCharm
 	s.assertDeployArgs(c, mariadbCurl.String(), "mariadb", "focal")
 
 	c.Check(s.output.String(), gc.Equals, ""+
-		"Located charm \"gitlab-k8s\" in charm-hub\n"+
-		"Located charm \"mariadb-k8s\" in charm-hub\n"+
+		"Located charm \"gitlab-k8s\" in charm-hub, channel new/edge\n"+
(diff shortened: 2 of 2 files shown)


cmd/juju/application/deployer/charm.go
```diff
@@ -550,11 +550,13 @@ func isEmptyOrigin(origin commoncharm.Origin, source commoncharm.OriginSource) b
 func formatLocatedText(curl *charm.URL, origin commoncharm.Origin) string {
 	repository := origin.Source
 	if repository == "" || repository == commoncharm.OriginLocal {
-		return fmt.Sprintf("Located local charm %q, revision %d", curl.Name, curl.Revision)
+		return fmt.Sprintf("Located local charm %q, next %d", curl.Name, curl.Revision)
```
wallyworld, line 553: typo

wallyworld (commented):
The QA output still has
`Located charm "ntp" in charm-hub, channel stable`
printed twice. Can we suppress this duplicate message?

SimonRichardson:
> Can we suppress this duplicate message?

That means we'd have to hold state for this. Seems a bit of a waste.

SimonRichardson (commented):
I wonder if we move the `formatLocatedText` (which outputs the message) to after it has an origin, then we can add the arch to the output as well.

hmlanigan:
@wallyworld We may never be able to put enough data on the line to make it unique.  And the fact is, we are locating it twice.  One problem is that while resolving the charms we do not have easy access to the name of the application to be used with the charm.  This is a day one "problem" when deploying bundles.

@SimonRichardson the arch is displayed when we "upload", is that enough?

I'm wondering when the lines get so long with so much data as to be unreadable?  Do we need to replicated every bit of data for:

- Located
- uploading
- deploying

SimonRichardson (approved):
> And the fact is, we are locating it twice. 

This is true. For that reason, we should highlight it. This is just an edge case, we can always improve it at a later stage.

cmd/juju/application/deployer/charm.go
hmlanigan: thanks, got caught by IDE refactor, caught one of them, just not the 2nd.

hmlanigan:
$$merge$$

PR #12912, opened 2021-04-22
Series-less charms

SimonRichardson:
The following program attempts to locate series-less charms on
charmhub. These charms will not have a series or a map of containers.

This script is useful to identify problem charms that made the charmhub
cut. These charms are either old or should be updated to have a series
within the metadata.

The code is nothing special, just iterates the API.

## QA steps

```sh
go run ./scripts/charmhub/locate-series-less-charms.go
```

### DIFF
--- a/scripts/charmhub/locate-series-less-charms.go
+++ b/scripts/charmhub/locate-series-less-charms.go
@@ -1,0 +1,55 @@
+package main
+
+import (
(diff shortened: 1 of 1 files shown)


manadart (approved):

SimonRichardson:
$$merge$$

PR #12914, opened 2021-04-22
Fix gosigma revision in go mod so it is a real revision

wallyworld:
The gosigma revision we were referencing in go.mod was incorrect - it didn't exist.
To fix this, we needed to make upstream changes to terms-client and charmrepo.

This PR pulls in those upstream deps.

## QA steps

bootstrap

## Bug reference

https://bugs.launchpad.net/juju/+bug/1923064


SimonRichardson (approved):

SimonRichardson (changes requested):
Actually, it broke the schema stuff, so holding for now.

SimonRichardson (approved):

SimonRichardson:
$$merge$$

PR #12924, opened 2021-04-23
Handle k8s v1 charm inside a bundle

SimonRichardson:
Deploying a v1 charm metadata inside a k8s bundle looses the ability to
know that the computed series for a charm is actually a kubernetes
charm.

The code just checks to ensure that the computed series also contains
kubernetes so that we can deploy correctly. This is a workaround for v1
charm metadata and if we ever remove support for it, we can safely drop
kubernetes series.

## QA steps

Save the following bundle yaml locally:

```yaml
bundle: kubernetes
applications:
  snappass:
    charm: snappass-test
    scale: 1
```

```sh
$ juju bootstrap microk8s
$ juju deploy ./bundle.yaml
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1925717

### DIFF
--- a/core/charm/computedseries.go
+++ b/core/charm/computedseries.go
@@ -22,7 +22,7 @@ func ComputedSeries(c charm.CharmMeta) ([]string, error) {
 
 	// We use a set to ensure uniqueness and a slice to ensure that we
 	// preserve the order of elements as they appear in the manifest.
-	seriesSlice := []string(nil)
+	var seriesSlice []string
 	seriesSet := set.NewStrings()
 
 	for _, base := range manifest.Bases {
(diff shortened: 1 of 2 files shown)


core/charm/computedseries.go
```diff
@@ -36,5 +36,10 @@ func ComputedSeries(c charm.CharmMeta) ([]string, error) {
 			seriesSlice = append(seriesSlice, series)
```
achilleasa, line 36: Are we using the slice instead of the set (with a .Values/SortedValues() in the end) here to maintain the order as seen in the bases?

SimonRichardson: Yes

achilleasa: Doh... didn't read the comment either!

achilleasa (approved):
Deploy works as expected

SimonRichardson:
$$merge$$

PR #12927, opened 2021-04-27
Add scope to trust CLI for k8s models

wallyworld:
Add a scope option to the trust command - when trusting charms on k8s models, the user is required to use `--scope=cluster` to confirm that the charm will be granted access to the entire cluster.
This is a stop gap until the permissions model is refined.

## QA steps

deploy a k8s charm
```
$ juju trust snappass-test
ERROR 'juju trust' currently grants full access to the cluster itself.
Set the scope to 'cluster' using '--scope=cluster' to confirm this choice.

$ juju trust snappass-test --scope=clu
ERROR scope "clu" not valid

$ juju trust snappass-test --scope=cluster
$

$ juju trust snappass-test --remove
$
```


wallyworld:
[juju trust CLI scope](https://trello.com/c/cl5LlutJ/1550-juju-trust-cli-scope)

cmd/juju/application/trust.go
```diff
@@ -19,19 +20,30 @@ const (
 	trustSummary = `Sets the trust status of a deployed application to true.`
 	trustDetails = `Sets the trust configuration value to true.
 
+On k8s models, the trust operation currently grants the charm full access to the cluster.
+Until the permissions model refined to grant more granular role based access, the use of
```
jameinel, line 24:
```suggestion
Until the permissions model is refined to grant more granular role based access, the use of
```

cmd/juju/application/trust.go
```diff
@@ -19,19 +20,30 @@ const (
...
+
 Examples:
     juju trust media-wiki
+    juju trust metallb --scope=cluster
 
 See also:
     config
+`
+	clusterScopeError = `'juju trust' currently grants full access to the cluster itself.
+Set the scope to 'cluster' using '--scope=cluster' to confirm this choice.
```
jameinel, line 35: I have some concern that this functionally puts us into a forced cli with '--scope', and we haven't even designed how we want categories to work. And what would the syntax be if there are multiple things that could be trusted.

cmd/juju/application/trust.go
```diff
@@ -62,3 +75,19 @@ func (c *trustCommand) Init(args []string) error {
 	trustOptionPair = fmt.Sprintf("%s=%t", application.TrustConfigOptionName, !c.removeTrust)
 	return c.parseSet([]string{trustOptionPair})
 }
+
+func (c *trustCommand) Run(ctx *cmd.Context) error {
+	modelType, err:= c.ModelType()
+	if err != nil {
+		return errors.Trace(err)
+	}
+	if modelType == model.CAAS && !c.removeTrust {
```
jameinel, line 84: This should really be validated on the API side and not in the CLI, shouldn't it?
Checking ModelType in the CLI feels like an anti pattern.
I guess we can't really since we aren't actually calling an API around trust, we are just setting trust: true in config.

jameinel (approved):
I think this is a bit of a hack, but it does solve the short term issue and gives us room to get it correct with an actual spec.
We'll need to make sure docs are updated to match this, and reach out to Jon Seager.

cmd/juju/application/trust.go
wallyworld: The model info (UUID and type) is stored client side after making an API call to get it if the current client doesn't know about the model yet.

cmd/juju/application/trust.go
wallyworld: Agree, but the clear direction we were given is that this is what was wanted in the dot zero release.

snap/snapcraft.yaml
```diff
@@ -19,7 +19,7 @@ description: |
    - https://discourse.charmhub.io/
    - https://github.com/juju/juju
 
-confinement: classic
+confinement: strict
```
jameinel, line 22: this doesn't seem quite right as part of this PR

wallyworld: oops, didn't mean to push that

wallyworld:
$$merge$$

PR #12928, opened 2021-04-27
Suppress charm name in status oci image path

wallyworld:
k8s charms deployed from the rocks or jujucharm repos have the image path set to
`<repo-urll>/<charm_name>/<resource_name>`.
 Charm name is somewhat redundant and takes up value space. So we'll replace with "ch:" to indicate the named resource comes from the store repo.

## QA steps

```
juju  deploy cs:~juju/mariadb-k8s
juju deploy ambassador
juju status
...
App            Version                  Status  Scale  Charm          Store       Channel  Rev  OS          Address         Message
ambassador     res:oci-image@ac02890    active      1  ambassador     charmhub    stable    13  kubernetes                  
mariadb-k8s    res:mysql_image@e6ea77e  active      2  mariadb-k8s    charmstore  stable     3  kubernetes  10.152.183.134  
...
```

### DIFF
--- a/cmd/juju/status/output_tabular.go
+++ b/cmd/juju/status/output_tabular.go
@@ -165,6 +165,24 @@ func printApplications(tw *ansiterm.TabWriter, fs formattedStatus) {
 			if len(parts) > 1 && len(version) > maxVersionWidth {
 				version = strings.Join(parts[1:], "/")
 			}
+			// Charms deployed from the rocks or jujucharm repos have the
+			// image path set to <repo-url>/<charm_name>/<resource_name>. Charm name
+			// is somewhat redundant and takes up value space. So we'll replace
+			// with "ch:" to indicate the named resource comes from the store repo.
+			parts = strings.Split(version, "/")
+			if len(parts) > 1 && (len(version) > maxVersionWidth || parts[0] == app.Charm) {
+				prefix := ""
+				switch app.CharmOrigin {
+				case "charmhub":
+					prefix = "ch"
+				case "charmstore":
+					prefix = "cs"
+				}
+				if prefix != "" {
+					parts[1] = prefix + ":" + parts[1]
+					version = strings.Join(parts[1:], "/")
+				}
+			}
 		}
 		// Don't let a long version push out the version column.
 		if len(version) > maxVersionWidth {
--- a/cmd/juju/status/output_tabular.go
+++ b/cmd/juju/status/output_tabular.go
@@ -162,26 +162,29 @@ func printApplications(tw *ansiterm.TabWriter, fs formattedStatus) {
 				}
 			}
 			parts := strings.Split(version, "/")
-			if len(parts) > 1 && len(version) > maxVersionWidth {
-				version = strings.Join(parts[1:], "/")
-			}
 			// Charms deployed from the rocks or jujucharm repos have the
-			// image path set to <repo-url>/<charm_name>/<resource_name>. Charm name
+			// image path set to <repo-url>/<user>/<charm_name>/<resource_name>. Charm name
 			// is somewhat redundant and takes up value space. So we'll replace
-			// with "ch:" to indicate the named resource comes from the store repo.
-			parts = strings.Split(version, "/")
-			if len(parts) > 1 && (len(version) > maxVersionWidth || parts[0] == app.Charm) {
+			// with "res:" to indicate the named resource comes from the store repo.
+			fromJujuRepo := strings.Contains(app.Version, "jujucharms.") || strings.Contains(app.Version, "rocks.")
+			if fromJujuRepo && len(parts) > 2 && (len(version) > maxVersionWidth || parts[1] == app.Charm) {
 				prefix := ""
 				switch app.CharmOrigin {
-				case "charmhub":
-					prefix = "ch"
-				case "charmstore":
-					prefix = "cs"
+				case "charmhub", "charmstore":
+					prefix = "res:"
 				}
 				if prefix != "" {
-					parts[1] = prefix + ":" + parts[1]
-					version = strings.Join(parts[1:], "/")
+					parts[2] = prefix + parts[2]
+					version = strings.Join(parts[2:], "/")
 				}
+			}
+			// For qualified images, if they are too long, we still
+			// want to see the core image name, but we can compromise
+			// on the namespace part and replace that with "...".
+			parts = strings.Split(version, "/")
+			if len(parts) > 1 && len(version) > maxVersionWidth {
+				parts[0] = ellipsis
+				version = strings.Join(parts, "/")
 			}
 		}
 		// Don't let a long version push out the version column.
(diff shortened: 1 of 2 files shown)


wallyworld:
[improve k8s app version for store charms](https://trello.com/c/WG8Aete5/1551-improve-k8s-app-version-for-store-charms)

SimonRichardson (commented):
I don't think we should have `cs:` or `ch:`. `ch` is an internal representation that we shouldn't leak outside of juju. I know that it does work when deploying, but in reality, I do wonder if we should prevent that.

I'd rather just drop the prefix and let people use `Store` to find out where the image is from.

cmd/juju/status/output_tabular.go
```diff
@@ -165,6 +165,24 @@ func printApplications(tw *ansiterm.TabWriter, fs formattedStatus) {
...
+			// with "ch:" to indicate the named resource comes from the store repo.
+			parts = strings.Split(version, "/")
+			if len(parts) > 1 && (len(version) > maxVersionWidth || parts[0] == app.Charm) {
+				prefix := ""
+				switch app.CharmOrigin {
+				case "charmhub":
+					prefix = "ch"
+				case "charmstore":
+					prefix = "cs"
+				}
```
jameinel, line 180: So what do we get if the CharmOrigin isn't either of these? eg: '--resource oci-image=jameinel/mysql'

It seems like this would also truncate those to just 'mysql'. Which would confuse "oci-image=mysql:latest' with "jameinel/mysql:latest".

I wonder if we can have a better way to only do the path stripping if it is a resource, and have a clear way to say it is a resource. 

`resource:oci-image@deadbee`

is probably a bit long, but does convey it clearly. And I do know that Mark has specifically stated he doesn't like to see 'ch:' anywhere.

`r:` feels like it is unclear what you mean by r. `res:oci-image@deadbee` does make efficient use of the space, but does still use a name that isn't defined.

cmd/juju/status/status_internal_test.go
```diff
@@ -5338,9 +5338,11 @@ func (s *StatusSuite) TestFormatTabularCAASModelTruncatedVersion(c *gc.C) {
 				},
 			},
 			"bar": {
+				Charm: "bar",
+				CharmOrigin: "charmhub",
 				Scale:   1,
 				Address: "54.32.1.3",
-				Version: "registry.jujucharms.com/mine/image:0.5",
```
jameinel, line 5343: these URLs are actually incomplete as it is 'registry.jujucharms.com/user/charm/resource'
I don't think it matters for testing purposes, but it would be good to have correct URLs.

cmd/juju/status/status_internal_test.go
```diff
@@ -5376,10 +5378,10 @@ func (s *StatusSuite) TestFormatTabularCAASModelTruncatedVersion(c *gc.C) {
...
                                  
 
-App  Version            Status  Scale  Charm  Store  Channel  Rev  OS  Address    Message
-bar  mine/image:0.5               0/1                           0      54.32.1.3  
-baz  image:0.6                    0/1                           0      54.32.1.4  
-foo  image:tag@3046a3d            0/1                           0      54.32.1.2  
+App  Version            Status  Scale  Charm  Store     Channel  Rev  OS  Address    Message
+bar  ch:image:0.5                 0/1         charmhub             0      54.32.1.3  
+baz  image:0.6                    0/1                              0      54.32.1.4  
+foo  image:tag@3046a3d            0/1                              0      54.32.1.2
```
jameinel, line 5384: From what I can understand from here:
https://docs.docker.com/engine/reference/commandline/images/

It looks like "image@deadbeef" works or "image:tag" but not "image:tag@digest"

Specifically this content:
```
before (<image-name>[:<tag>], <image id> or <image@digest>) - filter images created before given id or references
since (<image-name>[:<tag>], <image id> or <image@digest>) - filter images created since given id or references
```

jameinel (commented):
I'd rather not do ch: or cs:, and probably only strip the parts if it comes from the store.

cmd/juju/status/status_internal_test.go
wallyworld: I tested "image:tag@digest" manually and it worked.
The docker library has methods pull pulling out a tag separate to the digest.

cmd/juju/status/status_internal_test.go
wallyworld: > I don't think we should have `cs:` or `ch:`. `ch` is an internal representation that we shouldn't leak outside of juju. I know that it does work when deploying, but in reality, I do wonder if we should prevent that.
> 
> I'd rather just drop the prefix and let people use `Store` to find out where the image is from.

The reason for doing it was to signal that the displayed path comes from the resource definition inside a charm rather than a oci image path from a different registry. Without the "ch:", the user sees just "mysql_image" which is the resource name defined in charm metadata and is pretty meaningless without the hint as to where it comes from. Compare this to displaying "mariadb" which is reasonably interpreted as an upstream oci image name.

cmd/juju/status/output_tabular.go
wallyworld: It's difficult to use a very short text string to properly display something meaningful with all the possible permutations. There's  official upstream images from dicker.io or quay.io etc; there's images from someone's namespace on a repo; there's images from a canonical repo which use the charm declared resource name and which obfuscates the original image source. Anything other than the image name and sha very quickly uses the available width.
I think "res:" is probably the best compromise.

wallyworld:
> I don't think we should have `cs:` or `ch:`. `ch` is an internal representation that we shouldn't leak outside of juju. I know that it does work when deploying, but in reality, I do wonder if we should prevent that.
> 
> I'd rather just drop the prefix and let people use `Store` to find out where the image is from.

Replaced with "res:" as per jam's suggestion

cmd/juju/status/status_internal_test.go
wallyworld: As per another comment, "res:" is used to indicate that the image path is a charm resource.
Test also updated to use a correct jujucharms url

tlm (approved):
LGTM

wallyworld:
$$merge$$

cmd/juju/status/output_tabular.go
```diff
@@ -162,8 +162,29 @@ func printApplications(tw *ansiterm.TabWriter, fs formattedStatus) {
 				}
 			}
 			parts := strings.Split(version, "/")
+			// Charms deployed from the rocks or jujucharm repos have the
+			// image path set to <repo-url>/<user>/<charm_name>/<resource_name>. Charm name
+			// is somewhat redundant and takes up value space. So we'll replace
+			// with "res:" to indicate the named resource comes from the store repo.
+			fromJujuRepo := strings.Contains(app.Version, "jujucharms.") || strings.Contains(app.Version, "rocks.")
+			if fromJujuRepo && len(parts) > 2 && (len(version) > maxVersionWidth || parts[1] == app.Charm) {
```
manadart, line 170: I really think we should attempt wherever we can, to pare down complexity, which would tend away from adding this more logic at this level of indentation to an already large method.

PR #12929, opened 2021-04-27
Adds support for missing Kubernetes vars to hook.

tlm:
Hook contexts have been missing Kubernetes var's. As per lp-1892255.
This change introduces platform scoped env vars that can be harvested
from the enviornment.

## Checklist

## QA steps

See bug lp bug below.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1892255

### DIFF
--- a/worker/uniter/runner/context/context.go
+++ b/worker/uniter/runner/context/context.go
@@ -951,8 +951,19 @@ func (ctx *HookContext) ActionData() (*ActionData, error) {
 // such that it can know what environment it's operating in, and can call back
 // into context.
 // Implements runner.Context.
-func (ctx *HookContext) HookVars(paths Paths, remote bool, getEnv GetEnvFunc) ([]string, error) {
+func (ctx *HookContext) HookVars(
+	paths Paths,
+	remote bool,
+	getEnv GetEnvFunc,
+	osEnv OSEnvFunc,
+) ([]string, error) {
 	vars := ctx.legacyProxySettings.AsEnvironmentValues()
+
+	// We add all the host env's here because hooks are starting to expect that
+	// they can see this. This in response to lp1892255. We are doing this early
+	// so our overrides come out on top.
+	vars = append(vars, osEnv()...)
+
 	// TODO(thumper): as work on proxies progress, there will be additional
 	// proxy settings to be added.
 	vars = append(vars,
@@ -1017,6 +1028,7 @@ func (ctx *HookContext) HookVars(paths Paths, remote bool, getEnv GetEnvFunc) ([
 	if ctx.workloadName != "" {
 		vars = append(vars, "JUJU_WORKLOAD_NAME="+ctx.workloadName)
 	}
+
 	return append(vars, OSDependentEnvVars(paths, getEnv)...), nil
 }
 
--- a/worker/uniter/runner/context/env.go
+++ b/worker/uniter/runner/context/env.go
@@ -16,6 +16,9 @@ import (
 // based off existing variables.
 type GetEnvFunc func(key string) string
 
+// OSEnvFunc returns all env vars for the host os
+type OSEnvFunc func() []string
+
 // OSDependentEnvVars returns the OS-dependent environment variables that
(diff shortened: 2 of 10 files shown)


worker/uniter/runner/runner_test.go
```diff
@@ -59,7 +59,7 @@ exit 42
 	c.Assert(err, jc.ErrorIsNil)
 
 	c.Assert(result.Code, gc.Equals, 42)
-	c.Assert(strings.ReplaceAll(string(result.Stdout), "\n", ""), gc.Equals, paths.GetCharmDir()+"BAR")
+	c.Assert(strings.ReplaceAll(string(result.Stdout), "\n", ""), gc.Equals, paths.GetCharmDir())
```
wallyworld, line 62: So we need to change this test to use a KUBERNETES_FOO env var, not just delete the check :-)

worker/uniter/runner/context/context.go
```diff
@@ -1017,6 +1017,8 @@ func (ctx *HookContext) HookVars(paths Paths, remote bool, getEnv GetEnvFunc) ([
 	if ctx.workloadName != "" {
 		vars = append(vars, "JUJU_WORKLOAD_NAME="+ctx.workloadName)
 	}
+
+	vars = append(vars, PlatformEnvVars(getEnv)...)
```
wallyworld, line 1021: can we alter a relevant unit test to add in some k8s env vars to test this

worker/uniter/runner/context/env.go
```diff
@@ -34,6 +48,19 @@ func OSDependentEnvVars(paths Paths, getEnv GetEnvFunc) []string {
 	return nil
 }
 
+// PlatformEnvVars provides a quick way to detect any platform specific env vars
+// that should be used by the caller. This was originally written to pass
+// through Kubernetes env vars.
+func PlatformEnvVars(getEnv GetEnvFunc) []string {
+	pVars := []string{}
+	for _, e := range platformEnvKeys {
```
wallyworld, line 56: Probably ok as a quick fix but we are changing the behaviour of a released juju, ie juju run now would no longer include all env vars like it did with os.Environ(). Not sure if we shouldn't retain the existing juju run behaviour. Also, by filtering just an KUBERNETES_ vars, we would be potentially suppressing any env vars juju might in the future apply to the container config in the pod spec?

wallyworld (approved):
I think maybe we should just pass through everything

tlm:
[LP 1892255 Kubernetes ENV Vars](https://trello.com/c/OiAUAFid/1546-lp-1892255-kubernetes-env-vars)

worker/uniter/runner/runner.go
```diff
@@ -421,9 +432,16 @@ func (runner *runner) runCharmHookWithLocation(hookName, charmLocation string, r
 			v, _ := env[k]
 			return v
 		}
+		osEnv = func() []string {
```
jameinel, line 435: It doesn't feel great that we reimplement all of this for getEnv and now osEnv in 2 places.
And our HookVars function is growing enough attributes that its probably time to start refactoring it into a class instead of just a function.

worker/uniter/runner/context/env_test.go
```diff
@@ -230,7 +230,7 @@ func (s *EnvSuite) TestEnvUbuntu(c *gc.C) {
 				c.Errorf("unexpected get env call for %q", k)
 			}
 			return ""
-		})
+		}, func() []string { return []string{} })
```
jameinel, line 233: having lots of these noop funcs definitely has me a bit concerned.

jameinel (approved):
I'd like to see this land to fix the bug, but then some time taken to consider how we could refactor this code to make it more manageable. Especially odd that we use OSEnvFunc which gives us all the env vars and then have a *separate* function that gives us just the ones we ask for.

So I'd like to see a cleaner fix for 2.9.1, but this should land for today.

tlm:
$$merge$$

PR #12931, opened 2021-04-27
Prevent starvation when asking for random values

SimonRichardson:
The previous request of a 100M file would fail because the head
couldn't get enough characters to fulfill the request and die. Weirdly
you can get the test to pass sometimes and not others. 

The fix for this is to request a large enough size file without the usage 
of head. Using dd works nicely here and we just pipe it through base64 
so we get a clean output in juju status.

## QA steps

```sh
(cd tests && ./main.sh resources)
```


achilleasa (approved):
LGTM

SimonRichardson:
$$merge$$

PR #12933, opened 2021-04-28
Fix/lp 1924677

ycliuhw:
*Currently we increment the CharmModifiedVersion whenever we open resource in controller or `resource-get` was run which is not right. And it causes the new sidecar charms deployed from charmhub to restart immediately after deploy;*

## Checklist

 - [ ] ~Requires a [pylibjuju](https://github.com/juju/python-libjuju) change~
 - [ ] ~Added [integration tests](https://github.com/juju/juju/tree/develop/tests) for the PR~
 - [ ] ~Added or updated [doc.go](https://discourse.jujucharms.com/t/readme-in-packages/451) related to packages changed~

## QA steps

```console
$ juju deploy ch:snappass-test

$ mkubectl -nt1 get statefulset.apps/snappass-test -o yaml | grep observedGeneration
```

## Documentation changes

No

## Bug reference

https://bugs.launchpad.net/juju/+bug/1924677


apiserver/facades/controller/caasapplicationprovisioner/provisioner.go
```diff
@@ -704,7 +704,7 @@ func (a *API) ApplicationOCIResources(args params.Entities) (params.CAASApplicat
 			if v.Type != charmresource.TypeContainerImage {
 				continue
 			}
-			reader, err := resources.OpenResource(v.Name)
+			reader, err := resources.OpenResource(v.Name, false)
```
jameinel, line 707: This doesn't feel right. It seems to me that OpenResource should always be false, and SetResource should mostly (always) be true.

If you are just *reading* the resource, you aren't changing it, thus I'm not sure we need a way to supply it. My quick grep says you always passed false, but maybe I missed a case. That would help keep the api clean.

apiserver/resources.go
```diff
@@ -33,7 +33,7 @@ type ResourcesBackend interface {
 	GetPendingResource(applicationID, name, pendingID string) (resource.Resource, error)
 
 	// SetResource adds the resource to blob storage and updates the metadata.
-	SetResource(applicationID, userID string, res charmresource.Resource, r io.Reader) (resource.Resource, error)
+	SetResource(applicationID, userID string, res charmresource.Resource, r io.Reader, incrementCharmModifiedVersion bool) (resource.Resource, error)
```
wallyworld, line 36: Small nit, often for bools it's clearer to define a type alias and use constants for `IncrementCharmVersion` and `DoNotIncrementCharmVersion`. Just a suggeston to maker the api semantics clearer

apiserver/facades/controller/caasapplicationprovisioner/provisioner.go
wallyworld: +1, this is called during the provisioning step in `(a *appWorker) alive()` so I think the resources should already have been set up. And Open should always be read only.

ycliuhw:
!!build!!

apiserver/facades/controller/caasapplicationprovisioner/provisioner.go
ycliuhw: Inside the OpenResource we do `set` the resource. 
But I moved this toggle into the `resources.OpenResource` where it calls `set`.

caas/kubernetes/provider/application/trust.go
```diff
@@ -85,6 +89,9 @@ func (a *app) applyClusterRoles(trust bool) error {
 
 	api := a.client.RbacV1().ClusterRoles()
 	role, err := api.Get(context.Background(), a.qualifiedClusterName(), metav1.GetOptions{})
+	if k8serrors.IsNotFound(err) {
+		return errors.NotFoundf("cluster role %q", a.serviceAccountName())
```
wallyworld, line 93: a.qualifiedClusterName()

state/resources_persistence_staged.go
```diff
@@ -92,7 +92,7 @@ func (staged StagedResource) Activate() error {
 				logger.Errorf("can't read existing resource during activate: %v", errors.Details(err))
 				return nil, errors.Trace(err)
 			}
-			if hasNewBytes {
+			if hasNewBytes && incrementCharmModifiedVersion == IncrementCharmModifiedVersion {
```
wallyworld, line 95: since this is a bool, we can just do
`if hasNewBytes && incrementCharmModifiedVersion {`

state/resources_state_resource.go
```diff
@@ -170,11 +170,22 @@ func (st resourceState) GetPendingResource(applicationID, name, pendingID string
 
 // TODO(ericsnow) Separate setting the metadata from storing the blob?
 
+type IncrementCharmModifiedVersionType bool
```
wallyworld, line 173: add a comment

state/resources_state_resource.go
```diff
@@ -217,7 +228,7 @@ func (st resourceState) AddPendingResource(applicationID, userID string, chRes c
 	}
 	rLogger.Debugf("adding pending resource %q for application %q (ID: %s)", chRes.Name, applicationID, pendingID)
 
-	if _, err := st.setResource(pendingID, applicationID, userID, chRes, nil); err != nil {
+	if _, err := st.setResource(pendingID, applicationID, userID, chRes, nil, true); err != nil {
```
wallyworld, line 231: IncrementCharmModifiedVersion

state/resources_state_resource.go
```diff
@@ -227,7 +238,7 @@ func (st resourceState) AddPendingResource(applicationID, userID string, chRes c
 // UpdatePendingResource stores the resource in the Juju model.
 func (st resourceState) UpdatePendingResource(applicationID, pendingID, userID string, chRes charmresource.Resource, r io.Reader) (resource.Resource, error) {
 	rLogger.Tracef("updating pending resource %q (%s) for application %q", chRes.Name, pendingID, applicationID)
-	res, err := st.setResource(pendingID, applicationID, userID, chRes, r)
+	res, err := st.setResource(pendingID, applicationID, userID, chRes, r, true)
```
wallyworld, line 241: IncrementCharmModifiedVersion

worker/caasapplicationprovisioner/application.go
```diff
@@ -402,7 +406,11 @@ func (a *appWorker) ensureTrust(app caas.Application) error {
 	}
 
 	a.logger.Debugf("updating application %q trust to %v", a.name, desiredTrust)
-	if err := app.Trust(desiredTrust); err != nil {
+	err = app.Trust(desiredTrust)
+	if errors.IsNotFound(err) {
+		return nil
```
wallyworld, line 411: maybe log a warning?

This is for the case where the app gets removed before we can act on the trust request right? I think in similar cases elsewhere do also ignore the not found, is that right?

resource/repositories/operations.go
```diff
@@ -133,7 +133,7 @@ func GetResource(args GetResourceArgs) (resource.Resource, io.ReadCloser, error)
 		return resource.Resource{}, nil, errors.Trace(err)
 	}
 
-	res, reader, err = opRepo.set(data.Resource, data)
+	res, reader, err = opRepo.set(data.Resource, data, state.DoNotIncrementCharmModifiedVersion)
```
wallyworld, line 136: Do we have a test that GetResource() does not update charm modified version?

wallyworld (approved):
Thanks for fixing this.
To confirm - the only place we care about not doing the increment is in OpenResource(). Seems weird we would have originally done the increment given it should be considered read only.

state/resources_state_resource.go
manadart, line 173: Comments for these. What they mean, what they're used for.

resource/repositories/repository.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/juju/errors"
 
 	"github.com/juju/juju/resource"
+	"github.com/juju/juju/state"
```
manadart, line 13: I think we're better off moving the new type out of state if using it necessitates this import.

resource/repositories/operations.go
ycliuhw: we don't have a test in this package at all hmm

resource/repositories/repository.go
ycliuhw: we do have import in the other file inside this package. So it might be ok here.
and those constants are really just an argument type for an exported function. 
Whenever we need those constants imported, we already have the function imported.

worker/caasapplicationprovisioner/application.go
ycliuhw: This could happen when trust watcher fired event before ensureServer gets called. 
So I changed to return `dependency.ErrBounce` to restart the worker to retry

state/resources_persistence_staged.go
ycliuhw: incrementCharmModifiedVersion is bool but the type of it is `IncrementCharmModifiedVersionType`.
so if hasNewBytes && incrementCharmModifiedVersion will get type mismatch error.

ycliuhw:
> Thanks for fixing this.
> To confirm - the only place we care about not doing the increment is in OpenResource(). Seems weird we would have originally done the increment given it should be considered read only.

And the 2nd place is `UnitResourcesHandler` for uniter to fetch resource.

ycliuhw:
$$merge$$

ycliuhw:
[2.9-rc11 - Glitch with K8s sidecar deploy: remote charm modified version does not match agent's](https://trello.com/c/ayT3C4DP/1535-29-rc11-glitch-with-k8s-sidecar-deploy-remote-charm-modified-version-does-not-match-agents)

PR #12934, opened 2021-04-28
Fix ssh to a k8s unit in the controller model

wallyworld:
On k8s, ssh to a unit in the controller model failed because the legacy label detection was wrong.
We were using a label selector created from the namespace name, not the model name. For non-controller models, these are the same so it doesn't matter, but it matters a lot for the controller model.

## QA steps

bootstrap k8s
juju deploy mariadb-k8s
juju ssh mariadb-k8s/0

## Bug reference

https://bugs.launchpad.net/juju/+bug/1926394

### DIFF
--- a/caas/kubernetes/provider/operator.go
+++ b/caas/kubernetes/provider/operator.go
@@ -45,11 +45,11 @@ func GetOperatorPodName(
 	podAPI typedcorev1.PodInterface,
 	nsAPI typedcorev1.NamespaceInterface,
 	appName string,
-	namespace string,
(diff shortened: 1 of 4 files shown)


wallyworld:
[fix ssh to k8s unit in controller model](https://trello.com/c/DyYIStNT/1554-fix-ssh-to-k8s-unit-in-controller-model)

tlm (approved):
lgtm

wallyworld:
$$merge$$

PR #12937, opened 2021-04-29
Actions with errors queuing them can be listed

wallyworld:
If an action is run with bad parameters, the CLI prints the error but only the parent operation record was being stored.
The list-operations API queries actions matching the search criteria and collates the parent operation ids. So in the case ot an error validating action params, the failed operation would be in the DB but never show up.

The change here is two fold:
1. always record an action record even if there's an error
 (if there's an error, do not create the corresponding action notification entry)
2. mark the parent operation as being in error state

The allows the list-operations and show-task commands to work as expected for actions with errors. The time the action was enqueued is recorded but the action and operation are in error state with the action having the error message.

## QA steps

juju bootstrap
deploy a charm with an action taking an int parameter, eg
```
test:
  description: test
  params:
    count:
      type: integer
      description: test
```

Run the action with invalid parameters
```
juju run ubuntu-lite/0 test count=aa
Running operation 1 with 1 task
ERROR validation failed: (root).count : must be of type integer, given "aa"
```

List the operations
```
$ juju operations
Id  Status  Started  Finished  Task IDs  Summary
 1  error                      2         test run on unit-ubuntu-lite-0

$ juju show-task 2 --format yaml
id: "2"
message: 'validation failed: (root).count : must be of type integer, given "aa"'
status: error
timing:
  enqueued: 2021-04-29 11:32:03 +1000 AEST
unit: ubuntu-lite/0

$ juju show-operation 1
summary: test run on unit-ubuntu-lite-0
status: error
fail: 'validation failed: (root).count : must be of type integer, given "aa"'
action:
  name: test
  parameters:
    count: aa
timing:
  enqueued: 2021-04-29 11:32:03 +1000 AEST
tasks:
  "2":
    host: ubuntu-lite/0
    status: error
    timing:
      enqueued: 2021-04-29 11:32:03 +1000 AEST
    message: 'validation failed: (root).count : must be of type integer, given "aa"'
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1926496

### DIFF
--- a/apiserver/facades/client/action/operation.go
+++ b/apiserver/facades/client/action/operation.go
@@ -82,30 +82,45 @@ func (a *ActionAPI) enqueue(arg params.Actions) (string, params.ActionResults, e
 
 	tagToActionReceiver := common.TagToActionReceiverFn(a.state.FindEntity)
 	response := params.ActionResults{Results: make([]params.ActionResult, len(arg.Actions))}
+	errorRecorded := false
 	for i, action := range arg.Actions {
 		currentResult := &response.Results[i]
 		actionReceiver := action.Receiver
+		var (
+			actionErr    error
+			enqueued     state.Action
+			receiver     state.ActionReceiver
+			receiverName string
+		)
 		if strings.HasSuffix(actionReceiver, "leader") {
 			app := strings.Split(actionReceiver, "/")[0]
-			receiverName, err := getLeader(app)
-			if err != nil {
-				currentResult.Error = common.ServerError(err)
-				continue
+			receiverName, actionErr = getLeader(app)
+			if actionErr != nil {
+				currentResult.Error = common.ServerError(actionErr)
+				goto failOperation
 			}
 			actionReceiver = names.NewUnitTag(receiverName).String()
 		}
-		receiver, err := tagToActionReceiver(actionReceiver)
-		if err != nil {
-			currentResult.Error = common.ServerError(err)
-			continue
+		receiver, actionErr = tagToActionReceiver(actionReceiver)
+		if actionErr != nil {
+			currentResult.Error = common.ServerError(actionErr)
+			goto failOperation
 		}
-		enqueued, err := receiver.AddAction(operationID, action.Name, action.Parameters)
-		if err != nil {
-			currentResult.Error = common.ServerError(err)
-			continue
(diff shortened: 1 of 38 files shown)


wallyworld:
[actions with validation errors](https://trello.com/c/RQcib2qa/1557-actions-with-validation-errors)

state/watcher.go
```diff
@@ -3445,19 +3445,6 @@ func (f *controllerIdFilter) match(key interface{}) bool {
 	return false
 }
 
-// WatchActionResults starts and returns a StringsWatcher that
```
wallyworld, line 3448: unused

state/unit.go
```diff
@@ -2804,10 +2804,25 @@ type ActionSpecsByName map[string]charm.ActionSpec
 // AddAction adds a new Action of type name and using arguments payload to
 // this Unit, and returns its ID.  Note that the use of spec.InsertDefaults
 // mutates payload.
-func (u *Unit) AddAction(operationID, name string, payload map[string]interface{}) (Action, error) {
+func (u *Unit) AddAction(operationID, name string, payload map[string]interface{}) (_ Action, err error) {
 	if len(name) == 0 {
```
ycliuhw, line 2808: It might be worth to wrap this into a func then it can be called from here and machine.AddAction.

apiserver/facades/client/action/operation.go
```diff
@@ -82,30 +82,45 @@ func (a *ActionAPI) enqueue(arg params.Actions) (string, params.ActionResults, e
...
 		}
-
 		response.Results[i] = common.MakeActionResult(receiver.Tag(), enqueued, false)
+		continue
+
+	failOperation:
+		// If we failed to enqueue the action, record the error on the operation.
+		if !errorRecorded {
+			errorRecorded = true
+			err = a.model.FailOperation(operationID, actionErr)
```
ycliuhw, line 121: so do we only record the first action error to the operation?

ycliuhw (approved):
lgtm, just two small questions. ty

apiserver/facades/client/action/operation.go
wallyworld: yeah, 99% of the time it will be just one action run across several targets. that's what the juju run cli invokes

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #12944, opened 2021-05-03
Improve destroy model messages if aborted or timed out

wallyworld:
When destroy model is run, cleaning up all cloud resources can take a while. The CLI will block and report progress, but the maximum wait time can be exceeded or the user can hit Ctrl^C. In such cases, the messages printed to the console are improved to tell the user there may be cloud resources that need to be cleaned up.

This is in response to bug 1910810 where deleting a k8s namespace can take a long time. We can't do much about that, but if the user hits Ctrl^C or uses a reasonable timeout like 5m, then the messaging is hopefully better.

Ultimately we may need to consider removing the model from the juju database even if there are cloud resources still in existence if the user supplies the --force option.

## QA steps

deploy a k8s model
destroy model with a short timeout
```
$ juju destroy-model test -y --timeout=1s
Destroying model
Waiting for model to be removed, 1 application(s)...
The following resources have not yet been removed:
 - 1 application(s)
Because the destroy model operation did not finish, there may be cloud resources left behind.
ERROR timeout after 1s
```

or destroy a model and Ctrl^C before it is done

```
$ juju destroy-model test -y
Destroying model
Waiting for model to be removed, 1 application(s)...^C

destroy model is still running in the background...
The following resources have not yet been removed:
 - 1 application(s)
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1910810

### DIFF
--- a/cmd/juju/model/destroy.go
+++ b/cmd/juju/model/destroy.go
@@ -98,6 +98,7 @@ Examples:
 
     juju destroy-model test
     juju destroy-model -y mymodel
+    juju destroy-model -y mymodel --timeout 5m
     juju destroy-model -y mymodel --destroy-storage
     juju destroy-model -y mymodel --release-storage
     juju destroy-model -y mymodel --force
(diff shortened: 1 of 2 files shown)


hpidcock:
It seems like we need a standard way to report back to the user what resources were created in the cloud for the model/application/unit/machine, that way when destroy-model or destroy-controller fail, we can list what cloud resources remain.

manadart (approved):
Code and QA all good.

wallyworld:
$$merge$$

wallyworld:
> It seems like we need a standard way to report back to the user what resources were created in the cloud for the model/application/unit/machine, that way when destroy-model or destroy-controller fail, we can list what cloud resources remain.

This is a long standing desire - what's there is VM centric and limited.

PR #12955, opened 2021-05-07
destroy-model --force can clean up juju model

wallyworld:
Sometimes cleaning up cloud resources when destroying a model can get stuck. In particular, on k8s, sometimes the model namespace gets stuck in terminating. In such a case, you just want to force through the removal of the Juju model records and clear the model from the Juju controller.

`destroy-model` is enhanced in two ways:
1. the timeout set by the user is passed through with the destroy api call (previously it was used client only)
2. the model undertaker will use timeout=0 and force as a signal to cleanup the model from the controller even if the cloud destroy operations fail due to a timeout. If the timeout is > 0, the operation fails without the Juju model being removed.

The destroy checks above only take place if the model timeout is not nil, which will be the case for older Juju clients.

## QA steps

```sh
$ juju bootstrap microk8s test
...
$ juju add-model test
Added 'test' model on microk8s/localhost with credential 'microk8s' for user 'admin'
$ juju deploy snappass-test
Located charm "snappass-test" in charm-hub, revision 9
Deploying "snappass-test" from charm-hub charm "snappass-test", revision 9 in channel stable
$ juju destroy-model test -y --timeout=10ms
Destroying model
Waiting for model to be removed, 1 application(s)...
The following resources have not yet been removed:
 - 1 application(s)
Because the destroy model operation did not finish, there may be cloud resources left behind.
Re-run 'destroy-model --timeout=0 --force' to clean up the Juju model database records
even with potentially orphaned cloud resources.
ERROR timeout after 10ms
$
$ juju destroy-model test -y --timeout=0 --force
Destroying model
Because the destroy model operation did not finish, there may be cloud resources left behind.
$
$ juju models
Controller: test

Model       Cloud/Region        Type        Status     Access  Last connection
controller  microk8s/localhost  kubernetes  available  admin   just now

```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1910810


wallyworld:
[allow destroy model timeout](https://trello.com/c/2cz8dbGA/1572-allow-destroy-model-timeout)

PR #12956, opened 2021-05-10
Fixes proxy address bug in controllers.yaml

tlm:
With the introduction of the Kubernetes proxy our update controllers
yaml logic was including the temp proxy address in the endpoints list.
This is not ideal because it's temporary. This change fixes that by not
inlcuding the connected address as part of the output anymore.

## Checklist

## QA steps

1. Bootstrap a Kubernetes controller that requires proxying. (Minikube, Microk8s).
2. Check the controllers.yaml file created to make sure there is no mention of localhost and a random port in the endpoints list
3. Run auxiliary Juju commands to make sure that subsequent logins don't add this extra information.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1926595

### DIFF
--- a/api/apiclient.go
+++ b/api/apiclient.go
@@ -160,10 +160,10 @@ type state struct {
 	// authorize macaroon based login requests.
 	bakeryClient *httpbakery.Client
 
-	// proxy is the proxy used for this connection when not nil. If's expected
+	// proxier is the proxier used for this connection when not nil. If's expected
 	// the proxy has already been started when placing in this var. This struct
 	// will take the responsibility of closing the proxy.
-	proxy jujuproxy.Proxier
+	proxier jujuproxy.Proxier
 }
 
 // RedirectError is returned from Open when the controller
@@ -216,22 +216,6 @@ func Open(info *Info, opts DialOpts) (Connection, error) {
 		ctx1, cancel := utils.ContextWithTimeout(dialCtx, opts.Clock, opts.Timeout)
 		defer cancel()
 		dialCtx = ctx1
-	}
-
-	if info.Proxier != nil {
-		if err := info.Proxier.Start(); err != nil {
-			return nil, errors.Annotate(err, "starting proxy for api connection")
-		}
-
-		switch p := info.Proxier.(type) {
-		case jujuproxy.TunnelProxier:
-			info.Addrs = []string{
-				fmt.Sprintf("%s:%s", p.Host(), p.Port()),
-			}
-		default:
-			info.Proxier.Stop()
-			return nil, errors.New("unknown proxier provided")
-		}
 	}
 
 	dialResult, err := dialAPI(dialCtx, info, opts)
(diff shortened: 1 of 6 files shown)


api/state.go
```diff
@@ -210,14 +210,7 @@ func (st *state) setLoginResult(p loginResultParams) error {
...
 	st.modelAccess = p.modelAccess
 
-	hostPorts, err := addAddress(p.servers, st.addr)
-	if err != nil {
-		if clerr := st.Close(); clerr != nil {
-			err = errors.Annotatef(err, "error closing state: %v", clerr)
-		}
-		return err
-	}
-	st.hostPorts = hostPorts
```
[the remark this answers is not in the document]
tlm: I have fixed the PR so that we are only ignoring the socket address when there is a proxy case involved (juju internal proxy).

We are doing a seperate PR to make the python lib understand the proxy information.

api/apiclient.go
```diff
@@ -662,6 +667,7 @@ func dialAPI(ctx context.Context, info *Info, opts0 DialOpts) (*dialResult, erro
 		return nil, errors.Trace(err)
 	}
 	logger.Infof("connection established to %q", dialInfo.urlStr)
+	dialInfo.proxier = info.Proxier
```
wallyworld, line 670: Maybe include something about the proxier in the above info message

api/state.go
```diff
@@ -210,14 +210,30 @@ func (st *state) setLoginResult(p loginResultParams) error {
...
+	// to host ports
+	if !st.IsProxied() {
+		hostPorts, err = addAddress(p.servers, st.addr)
+		if err == nil {
+			goto SETHOSTPORTS
+		}
+		if clerr := st.Close(); clerr != nil {
+			err = errors.Annotatef(err, "error closing state: %v", clerr)
+		}
+		return err
```
wallyworld, line 224: we can easily avoid the goto

```
if err != nil {
  close
  return err
}
```

api/apiclient.go
```diff
@@ -1299,6 +1305,10 @@ func (s *state) IPAddr() string {
 	return s.ipAddr
 }
 
+func (s *state) IsProxied() bool {
```
wallyworld, line 1308: either comment or lowercase
does it need to be exported?

proxy/testing/proxy.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2021 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package testing
+
+type MockProxier struct {
```
wallyworld, line 6: ideally gomock for new mocks

wallyworld (approved):

api/apiclient.go
tlm: Done & Done

proxy/testing/proxy.go
tlm: These don't need to be mocked. They are far to simple and introduces an unneeded dependency.

api/state.go
tlm: Done & Done

tlm:
$$merge$$

PR #12958, opened 2021-05-11
Improve add-relation error message for terminated offers

wallyworld:
When a consumed offer is terminated on the offering side, attempts to relate to that SAAS application rightly fail.
This PR improves the error message.

Also a driveby fix for destroy model timeout.

## QA steps

See bug. Relate to an offer and then remove it from the offering model.
Then attempt to relate again.

```
$ juju relate mediawiki:db mariadb
ERROR Offer "mariadb" has been removed from the remote model.
To relate to a new offer with the same name, first run
'juju remove-saas mariadb' to remove the SAAS record from this model.
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1928020

### DIFF
--- a/cmd/juju/model/destroy.go
+++ b/cmd/juju/model/destroy.go
@@ -166,7 +166,7 @@ func (c *destroyCommand) Init(args []string) error {
 	if c.destroyStorage && c.releaseStorage {
 		return errors.New("--destroy-storage and --release-storage cannot both be specified")
 	}
-	if c.timeout <= 0 {
+	if c.timeout < 0 {
 		return errors.New("timeout must be zero or greater")
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/juju/model/destroy.go
```diff
@@ -166,7 +166,7 @@ func (c *destroyCommand) Init(args []string) error {
 	if c.destroyStorage && c.releaseStorage {
 		return errors.New("--destroy-storage and --release-storage cannot both be specified")
 	}
-	if c.timeout <= 0 {
+	if c.timeout < 0 {
 		return errors.New("timeout must be zero or greater")
```
tlm, line 170: Does this line need to be updated now?

tlm (approved):
lgtm

cmd/juju/model/destroy.go
wallyworld: no, error message was correct

PR #12960, opened 2021-05-11
Users CLI help directive

SimonRichardson:
The help directive for 'juju users' didn't mention that model-name is an
optional argument if you don't want to get the controller users.

The fix is simple, just add that information to the cmd.Info.Args value.

## QA steps

```sh
$ juju help users
Usage: juju users [options] [model-name]
....
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1912407


manadart (approved):

SimonRichardson:
$$merge$$

PR #12974, opened 2021-05-13
Ignore headless services when reporting k8s addresses

wallyworld:
As reported here
https://discourse.charmhub.io/t/how-does-juju-collect-ingress-address/4521

Juju was sometimes reporting the service address(es) of the headless service for an application, depending on the order of the services in the ListServices k8 API call. When this happened, if no record was yet stored in mongo, the reported address would be None. If there was already an address recorded, there would be a txn conflict because the UID had changed.

The fix here is twofold. First, allow last write wins on the service record update. The UID is not expected to change, but if it does, retry the txn. Second, filter out the headless services when querying the cluster. Also, a service could have an IP and a hostname - record both as juju assigns them different types.

## QA steps

bootstrap k8s
As per https://gitlab.com/endikap100/charmbugloadbalancernone
deploy
```
description: test loadbalancer in juju
bundle: kubernetes
applications:
  nginx1:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx2:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx3:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx4:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx5:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx6:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx7:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx8:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx9:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx10:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx11:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx12:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx13:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx14:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
  nginx15:
    charm: '../charms/nginx/nginx.charm'
    scale: 1
```

juju status will report either the loadbalancer ip addresses or nothing (not None).
logs will not have any txn conflict errors

## Bug reference

https://bugs.launchpad.net/juju/+bug/1928282

### DIFF
--- a/caas/kubernetes/provider/k8s.go
+++ b/caas/kubernetes/provider/k8s.go
@@ -586,28 +586,29 @@ func (k *kubernetesClient) getStorageClass(name string) (*k8sstorage.StorageClas
 	return storageClasses.Get(name, v1.GetOptions{})
 }
 
-func getLoadBalancerAddress(svc *core.Service) string {
+func getLoadBalancerAddresses(svc *core.Service) []string {
 	// different cloud providers have a different way to report back the Load Balancer address.
 	// This covers the cases we know about so far.
+	var addr []string
 	lpAdd := svc.Spec.LoadBalancerIP
 	if lpAdd != "" {
-		return lpAdd
+		addr = append(addr, lpAdd)
 	}
 
 	ing := svc.Status.LoadBalancer.Ingress
 	if len(ing) == 0 {
-		return ""
+		return addr
 	}
 
-	// It usually has only one record.
-	firstOne := ing[0]
-	if firstOne.IP != "" {
-		return firstOne.IP
+	for _, ingressAddr := range ing {
+		if ingressAddr.IP != "" {
+			addr = append(addr, ingressAddr.IP)
+		}
+		if ingressAddr.Hostname != "" {
+			addr = append(addr, ingressAddr.Hostname)
+		}
 	}
-	if firstOne.Hostname != "" {
-		return firstOne.Hostname
-	}
-	return lpAdd
+	return addr
 }
 
 func getSvcAddresses(svc *core.Service, includeClusterIP bool) []network.ProviderAddress {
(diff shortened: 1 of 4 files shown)


wallyworld:
[incorrect k8s services addresses](https://trello.com/c/aXPyvfwN/1581-incorrect-k8s-services-addresses)

caas/kubernetes/provider/k8s.go
```diff
@@ -657,12 +658,23 @@ func (k *kubernetesClient) GetService(appName string, mode caas.DeploymentMode,
...
+		svc    *core.Service
+	)
 	// We may have the stateful set or deployment but service not done yet.
 	if len(servicesList.Items) > 0 {
-		service := servicesList.Items[0]
-		result.Id = string(service.GetUID())
-		result.Addresses = getSvcAddresses(&service, includeClusterIP)
+		for _, s := range servicesList.Items {
+			// Ignore any headless service for this app.
+			if s.Name == appName || s.Name == appName+"-service" {
```
ycliuhw, line 669: do we have an existing helper function for generating the svc name from appName?

ycliuhw (approved):
lgtm, just one question

caas/kubernetes/provider/k8s.go
wallyworld: There's one for legacy service names (with the juju prefix). There's also a method on the controller stack for the controller name but it's not accessible here. I think the safest option is simply to exclude services that are named "foo-endpoints", which is how the headless services are called. That way we cater for all the other possibilities.

wallyworld:
$$merge$$

PR #12978, opened 2021-05-13
Sidecar charms do not need operator storage

wallyworld:
When deploying a k8s charm, juju checks to see if the cluster supports storage in case the charm needs it. Charms for juju 2.8 or greater do not need it. Neither do new sidecar charms, but we were not accounting for that.

Also a  driveby fix to the juju ssh help text.

## QA steps

bootstrap microk8s, add a model
```
$ juju model-config operator-storage=""
$ juju deploy snappass-test
$
$ juju deploy cs:~juju/mariadb-k8s
Located charm "mariadb-k8s" in charm-store, revision 3
Deploying "mariadb-k8s" from charm-store charm "mariadb-k8s", revision 3 in channel stable
ERROR deploying this Kubernetes application requires a suitable storage class.
None have been configured. Set the operator-storage model config to specify which storage class should be used to allocate operator storage.
See https://discourse.jujucharms.com/t/getting-started/152.
```

## Bug reference

https://bugs.launchpad.net/bugs/1927027
https://bugs.launchpad.net/juju/+bug/1927743

### DIFF
--- a/apiserver/facades/client/application/application_unit_test.go
+++ b/apiserver/facades/client/application/application_unit_test.go
@@ -1119,7 +1119,7 @@ func (s *ApplicationSuite) TestDeployCAASModelNoOperatorStorage(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 	c.Assert(result.Results, gc.HasLen, 1)
 	msg := result.OneError().Error()
-	c.Assert(strings.Replace(msg, "\n", "", -1), gc.Matches, `deploying a Kubernetes application requires a suitable storage class.*`)
+	c.Assert(strings.Replace(msg, "\n", "", -1), gc.Matches, `deploying this Kubernetes application requires a suitable storage class.*`)
 }
 
 func (s *ApplicationSuite) TestDeployCAASModelCharmNeedsNoOperatorStorage(c *gc.C) {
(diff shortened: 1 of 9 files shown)


apiserver/facades/client/application/application_unit_test.go
```diff
@@ -1119,7 +1119,7 @@ func (s *ApplicationSuite) TestDeployCAASModelNoOperatorStorage(c *gc.C) {
 	c.Assert(err, jc.ErrorIsNil)
 	c.Assert(result.Results, gc.HasLen, 1)
 	msg := result.OneError().Error()
-	c.Assert(strings.Replace(msg, "\n", "", -1), gc.Matches, `deploying a Kubernetes application requires a suitable storage class.*`)
+	c.Assert(strings.Replace(msg, "\n", "", -1), gc.Matches, `deploying this Kubernetes application requires a suitable storage class.*`)
```
jameinel, line 1122: good catch

apiserver/facades/client/application/application_unit_test.go
```diff
@@ -1146,6 +1146,28 @@ func (s *ApplicationSuite) TestDeployCAASModelCharmNeedsNoOperatorStorage(c *gc.
 	c.Assert(result.Results[0].Error, gc.IsNil)
 }
 
+func (s *ApplicationSuite) TestDeployCAASModelSidecarCharmNeedsNoOperatorStorage(c *gc.C) {
+	s.model.modelType = state.ModelTypeCAAS
+	delete(s.model.cfg, "operator-storage")
+	s.backend.charm = &mockCharm{
+		meta:     &charm.Meta{},
+		manifest: &charm.Manifest{Bases: []charm.Base{{}}},
```
jameinel, line 1154: I'm kind of surprised we just needed a Bases declaration and not a Containers declaration. Because v2 machine charm would look like this.

cmd/juju/commands/ssh.go
```diff
@@ -59,6 +59,11 @@ where it is available. This is done by inserting them between the target and
 a possible remote command. Refer to the ssh man page for an explanation 
 of those options.
 
+For k8s charms, the --container argument is used to identity a specific
+container in the pod. For charms which run the workload in a separate pod
```
jameinel, line 63: the default is the charm container, either way, isn't it?

jameinel (approved):
I'm a little surprised to see the SSH changes with the storage check.
And I'm a little surprised that Bases is the right check, but AFAICT it is correct today, and if it becomes incorrect we can fix it when it does.

wallyworld:
[false sidecar charm storage check](https://trello.com/c/wjKCt8lC/1584-false-sidecar-charm-storage-check)

cmd/juju/commands/ssh.go
wallyworld: It is but in reality, the --container argument is only really relevant for v2 charms. I'll reword a little

apiserver/facades/client/application/application_unit_test.go
wallyworld: The format v2 check is (currently) premised on having bases defined. It's irrelevant how v2 is determined. All we care about is whether the charm is v2 or not. No v2 charm requires operator storage so we just use that to short circuit the check.

hpidcock (approved):

wallyworld:
$$merge$$

PR #12979, opened 2021-05-14
lxd models can be added to k8s controller

wallyworld:
Adding a lxd cloud and then a model to a k8s controller failed because the provisioner worker was assuming the host controller tag is a machine. However, we are moving towards using controller agent tags (which is what we use for k8s) so both need to be allowed. The tag is just used to generate a nonce so there's nothing restricting what it can be.

Also, add lxd, mass, openstack clouds to the k8s whitelist so --force is not needed.

## QA steps

As per the bug, bootstrap microk8s and add a lxd cloud and then a lxd model.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1927193

### DIFF
--- a/cloud/whitelist.go
+++ b/cloud/whitelist.go
@@ -53,6 +53,7 @@ func (w *WhiteList) Check(existing, new string) error {
 // CurrentWhiteList returns current clouds whitelist supported by Juju.
 func CurrentWhiteList() *WhiteList {
 	return &WhiteList{map[string]set.Strings{
+		"kubernetes":          set.NewStrings(lxdnames.ProviderType, "maas", "openstack"),
 		lxdnames.ProviderType: set.NewStrings(lxdnames.ProviderType, "maas", "openstack"),
 		"maas":                set.NewStrings("maas", "openstack"),
 		"openstack":           set.NewStrings("openstack"),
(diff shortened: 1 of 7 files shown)


wallyworld:
[add lxd cloud to k8s controller](https://trello.com/c/PCAVnu9N/1583-add-lxd-cloud-to-k8s-controller)

cloud/whitelist.go
```diff
@@ -53,8 +54,9 @@ func (w *WhiteList) Check(existing, new string) error {
 // CurrentWhiteList returns current clouds whitelist supported by Juju.
 func CurrentWhiteList() *WhiteList {
 	return &WhiteList{map[string]set.Strings{
-		lxdnames.ProviderType: set.NewStrings(lxdnames.ProviderType, "maas", "openstack"),
-		"maas":                set.NewStrings("maas", "openstack"),
-		"openstack":           set.NewStrings("openstack"),
+		k8sconstants.CAASProviderType: set.NewStrings(lxdnames.ProviderType, "maas", "openstack"),
```
ycliuhw, line 57: so this allows us to add lxd cloud without --force?
do we wanna enable `lxdnames.ProviderType` now?

ycliuhw (approved):
lgtm, just one question

cloud/whitelist.go
wallyworld: yeah, --force no longer needed as this PR allows remote lxd clouds to be added successfully

wallyworld:
$$merge$$

ycliuhw:
[lp:1927193 names.MachineTag, vs names.ControllerAgentTag in compute-provisioner worker Edit](https://trello.com/c/M4Sc80ok/1569-lp1927193-namesmachinetag-vs-namescontrolleragenttag-in-compute-provisioner-worker-edit)

PR #12984, opened 2021-05-18
Fix upgrade of operator mode charms

wallyworld:
v1 k8s charms can have mode=operator. These charms could not be upgraded due to a bug in pod-spec-set. We still need to allow pod-spec-set with nil so that charm modified version is incremented.
This fixes that bug and also removes the feature flag so such charms can be deployed as needed.

## QA steps

Deploy https://github.com/davigar15/kdu-proxy-operator
juju upgrade-charm

## Bug reference

https://bugs.launchpad.net/juju/+bug/1928074


ycliuhw (approved):
lgtm thanks

wallyworld:
$$merge$$

PR #12986, opened 2021-05-18
[2.9] Prevent contention error when switching charms that drop peer relations

achilleasa:
Juju allows switching to a charm that would otherwise break an existing peer relation as long as there is only one unit present. However, doing so leaves stale documents in the relations collection.

As a result, if deployed a single-unit application, then switched to a charm without peer relations and then tried to switch back to a charm with peer relations we would get in a state where the SetCharm call would attempt to insert a new document into the relations collection but its DocMissing assertion would fail since the doc was still there. As a result the txn runner would eventually give up and return a "contention error".

To address this, this commit first modifies the check-if-relations-will-be-broken logic so that it allows us to change the charm 
even if that means that an existing **peer** relation will be broken. Note that the upgrade will still be blocked if any established **non-peer** relation would be broken (the second commit in this PR relaxes the check based on @jameinel's feedback below).

Secondly, an extra check has been added to the changeCharmOps method which performs a diff between the existing peer relations and the peer relations that the new charm defines and removes any stale relations that were previously left behind after the switch.

The issue also affects 2.8 (with a 2.9 client) so we should back-port the fix just in case we release another 2.8 in the future.

## QA steps

```console
$ juju deploy ceph-mon

# Login to mongo and check that "db.relations.find().pretty()" shows a single document

# Then, switch to a charm that does not define any peer relations
$ juju refresh ceph-mon --switch influxdb

# Login to mongo and check that "db.relations.find().pretty()" shows NO documents

# Then, switch back to the original charm
$ juju refresh ceph-mon --switch ceph-mon

# Login to mongo and check that "db.relations.find().pretty()" shows NO documents

# Add a second unit so now we have a peer relation with > 1 members and try to switch again
# This should work as expected
$ juju add-unit ceph-mon
$ juju refresh ceph-mon --switch influxdb

# Then relate the app to another app and try to switch again. This time you should get an error
$ juju refresh ceph-mon --switch ceph-mon
$ juju relate ceph-mon $something-compatible
$ juju refresh ceph-mon --switch influxdb
Added charm-hub charm "influxdb", revision 23 in channel stable, to the model
ERROR cannot upgrade application "ceph-mon" to charm "ch:amd64/xenial/influxdb-23": would break relation "ceph-mon:$non-peer-endpoint"
```

### DIFF
--- a/state/application.go
+++ b/state/application.go
@@ -930,7 +930,8 @@ func (a *Application) checkRelationsOps(ch *Charm, relations []*Relation) ([]txn
 
 	isPeerToItself := func(ep Endpoint) bool {
 		// We do not want to prevent charm upgrade when endpoint relation is
-		// peer-scoped and there is only one unit of this application.
+		// peer-scoped and there is only one unit of this application or no
+		// units at all.
 		// Essentially, this is the corner case when a unit relates to itself.
 		// For example, in this case, we want to allow charm upgrade, for e.g.
 		// interface name change does not affect anything.
@@ -940,7 +941,7 @@ func (a *Application) checkRelationsOps(ch *Charm, relations []*Relation) ([]txn
 			// We are only interested in thinking further if we can get units.
 			return false
 		}
-		return len(units) == 1 && isPeer(ep)
+		return len(units) <= 1 && isPeer(ep)
 	}
 
 	// All relations must still exist and their endpoints are implemented by the charm.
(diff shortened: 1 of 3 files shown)


jameinel (commented):
At a high level, I think we just want `isPeer` instead of `isPeerToItself`.

I can't think of a good reason to prevent an application that wants to get rid of its peer relation from being able to do so.  Something that was coordinated via the peer relation clearly is no longer coordinated that way, and that's ok, because that is how the new charm code is meant to work.

hmlanigan:
@achilleasa unable to QA so far, due to various I/o timeouts I haven't been able to bootstrap today.

achilleasa:
@hmlanigan no worries. When you get a chance to try the QA steps can you please also run them against #12993 (backport to 2.8)?

hmlanigan:
@achilleasa I got an upgrade-charm hook failure on 'relation-set' during my testing:
https://paste.ubuntu.com/p/2QhWWZ8jmX/

If you refresh ceph-mon to influxdb while in blocked/idle, the upgrade-charm hook fails on get_list_database()

hmlanigan:
After consultations, I did a more real life test for QA.  Editing the ubuntu charm to have a peer relation, deploying it.  Then upgrading to a ubuntu charm version without a peer relation.  Adding units and a relation to ntp after.

hmlanigan (approved):
QA is good.  Approving

achilleasa:
$$merge$$

PR #12995, opened 2021-05-19
Updated the show-action command to include the default value of action parameters

jnsgruk:
This pull request adds the ability for the `juju show-action` command to also display the default value for any parameters where they are specified/non-empty.

## QA steps

```sh
$ juju bootstrap microk8s micro
$ juju deploy jnsgruk-test
$ juju show-action jnsgruk-test example-action
An example action with some parameters.

Arguments
param-blank-default:
  type: string
  description: this parameter has a blank default value.
  default: ""
param-default-value:
  type: string
  description: this parameter has a default value.
  default: somevalue
param-no-default:
  type: string
  description: this parameter has no default.
```

Note that the `jnsgruk-test` charm essentially does nothing but [specify some actions](https://github.com/jnsgruk/test-charm/blob/show-action-test/actions.yaml) and start a blank ubuntu container.
## Bug reference

https://bugs.launchpad.net/juju/+bug/1908714

### DIFF
--- a/cmd/juju/action/show.go
+++ b/cmd/juju/action/show.go
@@ -102,21 +102,11 @@ func (c *showCommand) Run(ctx *cmd.Context) error {
 			if !ok {
 				continue
 			}
-
-			if infoMap["default"] != nil {
-				args[argName] = actionArg{
-					Type:        infoMap["type"],
-					Description: infoMap["description"],
-					Default:     infoMap["default"],
-				}
-			} else {
-				// Don't include the default value if it's unspecified
-				args[argName] = actionArg{
-					Type:        infoMap["type"],
-					Description: infoMap["description"],
-				}
+			args[argName] = actionArg{
+				Type:        infoMap["type"],
+				Description: infoMap["description"],
+				Default:     infoMap["default"],
 			}
-
 		}
 	}
 
--- a/cmd/juju/action/show.go
+++ b/cmd/juju/action/show.go
@@ -102,11 +102,21 @@ func (c *showCommand) Run(ctx *cmd.Context) error {
 			if !ok {
 				continue
 			}
-			args[argName] = actionArg{
-				Type:        infoMap["type"],
-				Description: infoMap["description"],
-				Default:     infoMap["default"],
+
+			if infoMap["default"] != nil {
+				args[argName] = actionArg{
+					Type:        infoMap["type"],
+					Description: infoMap["description"],
+					Default:     infoMap["default"],
+				}
+			} else {
+				// Don't include the default value if it's unspecified
+				args[argName] = actionArg{
+					Type:        infoMap["type"],
+					Description: infoMap["description"],
+				}
 			}
+
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


jameinel:
!!build!!

cmd/juju/action/show.go
```diff
@@ -102,10 +102,21 @@ func (c *showCommand) Run(ctx *cmd.Context) error {
 			if !ok {
 				continue
 			}
-			args[argName] = actionArg{
-				Type:        infoMap["type"],
-				Description: infoMap["description"],
+
+			if infoMap["default"] != nil {
```
hmlanigan, line 106: You can check that the key exists and get the value if it exists in one step with something like:

			action := actionArg{
				Type:        infoMap["type"],
				Description: infoMap["description"],
			}
			if value, ok := infoMap["default"]; ok {
				action.Default = value
			}
			args[argName] = action

cmd/juju/action/show.go
```diff
@@ -102,10 +102,21 @@ func (c *showCommand) Run(ctx *cmd.Context) error {
...
+				args[argName] = actionArg{
+					Type:        infoMap["type"],
+					Description: infoMap["description"],
+					Default:     infoMap["default"],
```
jameinel, line 110: setting an interface{} to 'nil' is the same as not setting it, so I think you can just have
`args[argName] = actionArg{Type=,Description=,Default=}`

without doing any other checking.

cmd/juju/action/show_test.go
```diff
@@ -87,9 +87,14 @@ Arguments
...
   type: boolean
   description: take a full backup
+  default: true
 name:
   type: string
   description: snapshot name
+prefix:
+  type: string
+  description: prefix to snapshot name
+  default: ""
```
jameinel, line 97: I think we are now missing the case where there is no default, are we?

cmd/juju/action/show.go
jnsgruk: Fixed per @jameinel comment in the next commit, thanks!

cmd/juju/action/show.go
jnsgruk: Much nicer, fixed. Thanks!

cmd/juju/action/show_test.go
jnsgruk: I adjusted the [test actions](https://github.com/juju/juju/blob/1b2093b01b91e6020ac0084c2f6030bf194daf5c/cmd/juju/action/package_test.go#L80) in `package_test.go` so I think we're covered. We have a test for no default, a blank default and a default value

jnsgruk: See the "name" field above? :)

hmlanigan (approved):
QA is good.  Changes and tests LGTM.

hmlanigan:
!!build!!

cmd/juju/action/show_test.go
jameinel: yep, sorry I missed it.

jameinel (approved):

hmlanigan:
$$merge$$

PR #12996, opened 2021-05-20
Allow detached caas storage to be removed

wallyworld:
The juju storage CLI command was configured not to work on caas models.
But we can delete k8s storage, so long as it is detached first.

## QA steps

bootstrap k8s
juju deploy mariadb-k8s -n 3
$ juju remove-storage database/0 
failed to remove database/0: cannot destroy storage "database/0": storage is attached
$ juju remove-storage database/0 --force
ERROR forced detachment of storage on container models not supported

scale back to 2 units so the removed unit's storage gets detached
$ juju scale-application mariadb-k8s 2

$ juju remove-storage database/2
removing database/2

use juju status ---storage and kubectl get pv to see the store is gone

## Bug reference

https://bugs.launchpad.net/juju/+bug/1928755

### DIFF
--- a/cmd/juju/storage/remove.go
+++ b/cmd/juju/storage/remove.go
@@ -14,6 +14,7 @@ import (
 	jujucmd "github.com/juju/juju/cmd"
 	"github.com/juju/juju/cmd/juju/common"
 	"github.com/juju/juju/cmd/modelcmd"
+	"github.com/juju/juju/core/model"
 )
 
 // NewRemoveStorageCommandWithAPI returns a command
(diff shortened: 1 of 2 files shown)


wallyworld:
[remove k8s storage](https://trello.com/c/GO0cS9zz/1597-remove-k8s-storage)

cmd/juju/storage/remove.go
```diff
@@ -140,7 +150,7 @@ func (c *removeStorageCommand) Run(ctx *cmd.Context) error {
 			anyFailed = true
 		}
 	}
-	if anyAttached {
+	if anyAttached && c.modelType != model.CAAS {
```
ycliuhw, line 153: do we need a message for CAAS model?

ycliuhw (approved):
lgtm, just one question
thanks

cmd/juju/storage/remove.go
wallyworld: we already print the failure above. this message is just to tell the user they can use --force, which is not applicable to caas models

wallyworld:
$$merge$$

PR #13001, opened 2021-05-20
Charmhub: Improve download logging on failure

SimonRichardson:
The following changes aim to improve debugging when attempting to
download a charm. In order to not leak the client implementation out 
to the user, we hide the error status code. With this logging, a user 
can use `juju debug-log` to at least inspect it if it's required/persistent.

## QA steps

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1928182


hmlanigan (approved):
👍

hmlanigan:
$$merge$$

PR #13002, opened 2021-05-20
CLI: Fix panic in upgrade-series with older controllers

SimonRichardson:
The upgrade-series command expected that all controllers would return a
`*params.Error`. This isn't always the case, so the code here assumes that
we correctly handle that case.

## QA steps

```sh
$ snap install juju --classic --channel=2.3
$ juju bootstrap lxd test
$ juju deploy ubuntu --series=xenial
```
Build this PR:

```sh
$ juju upgrade-series 0 prepare bionic
...
ERROR upgrade-series is not supported.
Please upgrade your controller to perform the operation.
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1929100

### DIFF
--- a/cmd/juju/machine/upgradeseries.go
+++ b/cmd/juju/machine/upgradeseries.go
@@ -321,7 +321,15 @@ func getMachineID(fullStatus *params.FullStatus, id string) (string, error) {
 // or the info was malformed, we will fall back to the underlying error
 // and not provide any hints.
 func (c *upgradeSeriesCommand) displayProgressFromError(ctx *cmd.Context, err error) error {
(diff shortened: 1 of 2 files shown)


achilleasa (approved):
LGTM; is it a 2.9-specific problem or do we need to back-port the fix to the 2.8 client?

jameinel (approved):
I believe we are moving away from defaulting to 2.8 for bug fixes, but we are still intending to do a 2.8.11 and it could be worthwhile to land a simple fix like this into 2.8.

But otherwise the fix appears good for at least avoiding the panic.

SimonRichardson:
$$merge$$

PR #13003, opened 2021-05-21
Only run upgrade-charm hook once when k8s charm upgraded

wallyworld:
When a k8s charm is upgraded, the provisioner rewrites the pod spec causing a rolling update. As each unit agent starts up, it needs to detect this has happened so it can run the upgrade hook. It does this by seeing if the charm dir exists. But it was being done in the wrong place, and the upgrade operation was not properly wrapped by a `resolverOpFactory`. This meant that the local state charm url was not set, and so the next pass through the resolve loop thought an upgrade was needed and ran the hook a second time. We already have a checkCharmUpgrade() method so the charm dir check just needs to be moved there.

## QA steps

deploy snappass-test
juju upgrade-charm
juju status-log to check that only one upgrade-charm hook is run

do the same for a pod spec charm like mariadb-k8s

## Bug reference

https://bugs.launchpad.net/juju/+bug/1928776

### DIFF
--- a/worker/uniter/resolver/loop.go
+++ b/worker/uniter/resolver/loop.go
@@ -4,6 +4,7 @@
 package resolver
 
 import (
+	corecharm "github.com/juju/charm/v8"
 	"github.com/juju/charm/v8/hooks"
 	"github.com/juju/errors"
 
@@ -44,6 +45,7 @@ type LoopConfig struct {
 	Abort         <-chan struct{}
 	OnIdle        func() error
 	CharmDirGuard fortress.Guard
+	CharmDir      string
 	Logger        Logger
 }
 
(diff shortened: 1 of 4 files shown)


wallyworld:
[juju refresh hooks order problem](https://trello.com/c/IDJDHAol/1600-juju-refresh-hooks-order-problem)

ycliuhw (approved):
LGTM, thanks

wallyworld:
$$merge$$

worker/uniter/resolver/loop.go
```diff
@@ -197,7 +199,13 @@ func checkCharmUpgrade(logger Logger, remote remotestate.Snapshot, rf *resolverO
 		return nil
 	}
 
-	if *local.CharmURL == *remote.CharmURL {
+	_, err := corecharm.ReadCharmDir(charmDir)
+	haveCharmDir := err == nil
+	if !haveCharmDir {
+		logger.Debugf("start to re-download charm %v because charm dir %q has gone which is usually caused by operator pod re-scheduling", remote.CharmURL, charmDir)
```
SimonRichardson, line 205: This log message is in the wrong location. It should be after the early return.

PR #13005, opened 2021-05-21
Fix/lp 1928796

ycliuhw:
This PR does not fix lp:1928796 yet but it could make the fix easier later.
It moves `github.com/juju/bundlechanges/v4` to `core/bundle/changes` for easier keeping logics consistent in `deploy` command for bundles and charms.

TODO: we will have to move `github.com/juju/bundlechanges/v5` to `core/bundle/changes` for develop branch next.


achilleasa:
In my opinion, we should not move bundlechanges into juju at this point in time. 

If we move something into `core` this implies that it is is something re-usable. Currently, only the juju client uses bundlechanges so there is no point really in having it in `core` vs a client sub-package.

In addition, we are currently working on the spec for server-side (transactional) bundle deployments (it's a 3.0 feature but we _may_ want to backport to a 2.9.x so we have an upgrade path that allows us to safely drop the client-side implementation in 3.0). We want to retain the flexibility to bring in (into a facade sub-package) bundlechanges when the feature is implemented and _change/refactor_ it so that it simplifies the server-side deploy implementation (e.g clean it up so we don't have IAAS- and CAAS-specific changes, support `remove X/replace X` changes in the future e.t.c).

SimonRichardson:
> . Currently, only the juju client uses bundlechanges

Note: It's also used in the apiserver call `bundle.GetChanges`.

I also think we shouldn't be adding this at the moment mainly to see how the serverside bundle spec is going to pan out. How we're going to address support of both serverside and backward compatible 2.x changes are going to be interesting.

ycliuhw:
[1928796 Unable to deploy bundle with sidecar and pod_spec charms](https://trello.com/c/B4e0r8Po/1598-1928796-unable-to-deploy-bundle-with-sidecar-and-podspec-charms)

SimonRichardson (approved):
Having spoken to @achilleasa and @wallyworld we've come to the decision to move this in. 

The 3.0 branch will still need doing, but will likely have large changes following the work on moving deployment to server-side, which we'll just have to manage.

**Note**: You've got some broken tests around mocking charms due to the library changes. Probably need to regenerate the mocks.

ycliuhw:
$$merge$$

PR #13016, opened 2021-05-25
Shorter timeout for both microk8s.status and images pulling;

ycliuhw:
Shorter timeout for both microk8s.status and images pulling because the systemd timeout config for `snap.juju.fetch-oci.service` sets to `60s`.

```console
$ systemctl cat snap.juju.fetch-oci.service | grep TimeoutStartSec
TimeoutStartSec=60

```

### DIFF
--- a/snap/local/wrappers/fetch-oci
+++ b/snap/local/wrappers/fetch-oci
@@ -25,7 +25,7 @@ if ! which $container_cmd; then
 fi
 
 echo "Wait for microk8s to be ready if needed."
-wait_result=$(microk8s.status --wait-ready --timeout 30 2>&1)
+wait_result=$(microk8s.status --wait-ready --timeout 15 2>&1)
 if [ $? -ne 0 ]; then
     echo "${wait_result}"
     exit 0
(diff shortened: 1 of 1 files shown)


ycliuhw:
["connect-plug-peers": error: cannot communicate with server](https://trello.com/c/hYFGtbIG/1609-connect-plug-peers-error-cannot-communicate-with-server)

wallyworld (approved):
Ideally we'd increase the `snap.juju.fetch-oci.service` timeout.
40s isn't really long enough to pull the oci image.
and microk8s often takes longer than 15s to be ready

If there's no choice, then not erroring is a better option, but would be good not to have to shorten these timeouts

ycliuhw:
> Ideally we'd increase the `snap.juju.fetch-oci.service` timeout.
> 40s isn't really long enough to pull the oci image.
> and microk8s often takes longer than 15s to be ready
> 
> If there's no choice, then not erroring is a better option, but would be good not to have to shorten these timeouts

I was not able to find a way to config this value from snapd docs, but I have asked snapd guys in discourse for help.
This PR should make the `snap install juju` happen even timeout reached.
I will extend the `TimeoutStartSec` if we got any response from snapd guys.

ycliuhw:
$$merge$$

PR #13018, opened 2021-05-25
Charmhub: Download error should include status

SimonRichardson:
To help understand why something failed to download, it might help if we
always include the status code message. That way it becomes easier to
diagnose the underlying error without the need to hit debug-log every time.

## QA steps

See tests.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1929058

### DIFF
--- a/charmhub/download.go
+++ b/charmhub/download.go
@@ -209,5 +209,5 @@ func (c *DownloadClient) downloadFromURL(ctx context.Context, resourceURL *url.U
 
 	// Server error, nothing we can do other than inform the user that the
 	// archive was unaviable.
-	return nil, errors.Errorf("unable to locate archive with status: %s", resp.Status)
+	return nil, errors.Errorf("unable to locate archive (store API responded with status: %s)", resp.Status)
 }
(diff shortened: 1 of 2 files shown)


charmhub/download.go
```diff
@@ -209,5 +209,5 @@ func (c *DownloadClient) downloadFromURL(ctx context.Context, resourceURL *url.U
 
 	// Server error, nothing we can do other than inform the user that the
 	// archive was unaviable.
-	return nil, errors.Errorf("unable to locate archive")
+	return nil, errors.Errorf("unable to locate archive with status: %s", resp.Status)
```
achilleasa, line 212: Perhaps to make it obvious that it's an upstream dep at fault here?

```suggestion
	return nil, errors.Errorf("unable to locate archive (store API responded with status: %s)", resp.Status)
```

SimonRichardson: yeah I like that

achilleasa (approved):
LGTM

SimonRichardson:
$$merge$$

PR #13023, opened 2021-05-26
Charmhub: Expose http.Client option

SimonRichardson:
The following changes intends to allow the hoisting of the http.Client away
from the edge and up into the facade stack. The change is required, to
prevent the creation of an http.Client every time we instantiate a charmhub
client.

The code is very mechanical and is the first step in a long line of changes
towards the following bugs:

 - https://bugs.launchpad.net/juju/+bug/1928182
 - https://bugs.launchpad.net/juju/+bug/1899793

## QA steps

```sh
$ juju bootstrap lxd test
$ juju deploy postgresql
```

## Bug reference

 - https://bugs.launchpad.net/juju/+bug/1928182
 - https://bugs.launchpad.net/juju/+bug/1899793

PR #13024, opened 2021-05-26
GCE: Allow injection of global http client

SimonRichardson:
The following code changes starts to lift the http client out from the
edge and into the constructor of the environ. Future commits should
allow us to pass in a http client on the construction of the environ
from the dependency engine. This is just the first step of many to do
that.

## See

 - https://github.com/juju/juju/pull/13023

## QA steps

```sh
$ juju bootstrap google test 
```

## Bug reference

### DIFF
--- a/provider/gce/environ.go
+++ b/provider/gce/environ.go
@@ -4,11 +4,13 @@
 package gce
 
 import (
+	stdcontext "context"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/juju/errors"
+	"github.com/juju/utils"
 	"google.golang.org/api/compute/v1"
 
 	jujucloud "github.com/juju/juju/cloud"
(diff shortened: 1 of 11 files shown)


provider/gce/environ.go
```diff
@@ -150,14 +152,27 @@ func (e *environ) SetCloudSpec(spec environscloudspec.CloudSpec) error {
...
+		sslVerification = utils.NoVerifySSLHostnames
+	}
+
 	connectionConfig := google.ConnectionConfig{
 		Region:    spec.Region,
 		ProjectID: credential.ProjectID,
+
+		// TODO (Stickupkid): Pass the http.Client through on the construction
+		// of the environ.
+		HTTPClient: utils.GetHTTPClient(sslVerification),
```
hpidcock, line 167: Yeah agree this is a bit ugly.

provider/gce/environ.go
```diff
@@ -150,14 +152,27 @@ func (e *environ) SetCloudSpec(spec environscloudspec.CloudSpec) error {
...
 	}
 
+	// TODO (stickupkid): Pass the context through the method call.
+	ctx := stdcontext.Background()
```
hpidcock, line 171: This really should be a context.TODO(), same same, but here we are expecting context to be wired through eventually.

hpidcock (approved):

provider/gce/environ.go
```diff
@@ -4,11 +4,13 @@
 package gce
 
 import (
+	stdcontext "context"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/juju/errors"
+	"github.com/juju/utils"
```
wallyworld, line 13: Should be utils/v2

wallyworld (changes requested):
Please fix import before landing

SimonRichardson:
@wallyworld can you re-look, you're blocking a merge.

hpidcock:
$$merge$$

PR #13025, opened 2021-05-26
Reset back the start-timeout to 3m;

ycliuhw:
The `start-timeout` was wrongly reverted to the old value when we bring forward from 2.8.


hpidcock (approved):

ycliuhw:
["connect-plug-peers": error: cannot communicate with server](https://trello.com/c/hYFGtbIG/1609-connect-plug-peers-error-cannot-communicate-with-server)

PR #13032, opened 2021-05-28
Fix upgrade-model agent lookup for k8s controllers

wallyworld:
When upgrading a k8s model, the juju upgrade-model CLI would make a call to the controller to check what agent binaries are there. The call failed because the returned agent metadata includes a URL used to stream the binaries, and this URL is the controller `host:port`. The lookup of the controller URL needs to be done using the system state, but only the model state was being used. So we fix that here. 

Also, the jujud docker image selection requires that there be simplestreams metadata as a safety check so that only official images are used. We add bypass so that if the `developer-mode` feature flag is on, this additional check is no longer done, making it trivial to test upgrades for developers.

## QA steps

push the 2.8.10 juju-operator image and a 2.9 image built off this PR to a docker namespace eg `foo`
docker login
make push-release-operator-image

juju bootstrap microk8s --config caas-image-repo=foo
juju add-model test
juju deploy <a charm>
juju upgrade-controller --verbose
juju upgrade-model --verbose

The controller and model should upgrade to 2.9.4. Previously the model upgrade would fail.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1929887

### DIFF
--- a/apiserver/facades/agent/provisioner/provisioner.go
+++ b/apiserver/facades/agent/provisioner/provisioner.go
@@ -134,7 +134,7 @@ func NewProvisionerAPI(ctx facade.Context) (*ProvisionerAPI, error) {
 		return nil, errors.Annotate(err, "instantiating network config API")
 	}
 
-	urlGetter := common.NewToolsURLGetter(model.UUID(), st)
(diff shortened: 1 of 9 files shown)


ycliuhw (approved):
LGTM, thanks!

PR #13036, opened 2021-05-31
Use mongo storage for external identity macaroon keys

wallyworld:
The external macaroon authenticator was using in-memory storage for macaroons keys. This PR instead wires up a mongo root key store with an expiry time of 24 hours (local login macaroons also expire after 24 hours).

## QA steps

Needs to be tested in a set up with a public controller.

## Bug

https://bugs.launchpad.net/juju/+bug/1930319

### DIFF
--- a/apiserver/stateauthenticator/context.go
+++ b/apiserver/stateauthenticator/context.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"sync"
+	"time"
 
 	"github.com/go-macaroon-bakery/macaroon-bakery/v3/bakery"
 	"github.com/go-macaroon-bakery/macaroon-bakery/v3/bakery/checkers"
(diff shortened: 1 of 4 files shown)


wallyworld:
[external auth macaroon key](https://trello.com/c/HTSc80Je/1613-external-auth-macaroon-key)

apiserver/stateauthenticator/context.go
```diff
@@ -245,10 +246,15 @@ func (ctxt *authContext) externalMacaroonAuth(identClient identchecker.IdentityC
 
 var errMacaroonAuthNotConfigured = errors.New("macaroon authentication is not configured")
 
+const (
+	// TODO make this configurable via model config.
```
benhoyt, line 250: Will we realistically need this or come back to it? I reckon we should either remove this comment if we're likely never going to do it, otherwise wire it up now.

apiserver/stateauthenticator/context.go
```diff
@@ -278,24 +284,27 @@ func newExternalMacaroonAuth(st *state.State, clock clock.Clock, identClient ide
...
+	store, err := st.NewBakeryStorage()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+	store = store.ExpireAfter(expiryTime)
 	if identClient == nil {
 		identClient = &auth
 	}
-	indentBakery := identchecker.NewBakery(identchecker.BakeryParams{
+	identBakery := identchecker.NewBakery(identchecker.BakeryParams{
```
benhoyt, line 295: Good catch! An "indent bakery" ... is that what "go fmt" does? :-)

benhoyt (approved):

apiserver/stateauthenticator/context.go
wallyworld: There's been requests to make it configurable at various times. It's not super critical to do at this moment though. I'm +/- 0 on retaining the TODO - we have so many of them already scattered throughout - 1747 at last count :-) What's one more?

wallyworld:
$$merge$$

PR #13039, opened 2021-06-01
Fixes error with ensure ClusterRole

tlm:
When ensuring a cluster role if the labels had changed then update would
fail with an already exists error. As reported in lp1929909.

## Checklist

## QA steps

Upgrade from 2.8 to 2.9 or use regression test added with this PR. See the bug below for more information.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1929909

### DIFF
--- a/caas/kubernetes/provider/bootstrap.go
+++ b/caas/kubernetes/provider/bootstrap.go
@@ -31,6 +31,7 @@ import (
 	"github.com/juju/juju/caas"
 	"github.com/juju/juju/caas/kubernetes/provider/constants"
 	k8sproxy "github.com/juju/juju/caas/kubernetes/provider/proxy"
+	"github.com/juju/juju/caas/kubernetes/provider/resources"
 	k8sutils "github.com/juju/juju/caas/kubernetes/provider/utils"
 	providerutils "github.com/juju/juju/caas/kubernetes/provider/utils"
 	"github.com/juju/juju/cloud"
@@ -704,9 +705,11 @@ func (c *controllerStack) ensureControllerServiceAccount() error {
 		return errors.Trace(err)
 	}
 
-	crb := &rbacv1.ClusterRoleBinding{
+	// name cluster role binding after the controller namespace.
+	clusterRoleBindingName := c.broker.GetCurrentNamespace()
+	crb := resources.NewClusterRoleBinding(clusterRoleBindingName, &rbacv1.ClusterRoleBinding{
 		ObjectMeta: v1.ObjectMeta{
-			Name:        c.broker.GetCurrentNamespace(), // name cluster role binding after the controller namespace.
+			Name:        clusterRoleBindingName,
 			Labels:      providerutils.LabelsForModel("controller", false),
 			Annotations: c.stackAnnotations,
 		},
@@ -720,9 +723,9 @@ func (c *controllerStack) ensureControllerServiceAccount() error {
 			Name:      "controller",
 			Namespace: c.broker.GetCurrentNamespace(),
 		}},
-	}
+	})
 
-	_, crbCleanUps, err := c.broker.ensureClusterRoleBinding(crb)
+	crbCleanUps, err := crb.Ensure(c.ctx.Context(), c.broker.client())
 	c.addCleanUp(func() {
 		logger.Debugf("deleting %q cluster role binding", crb.Name)
 		for _, v := range crbCleanUps {
--- a/caas/kubernetes/provider/k8s_test.go
+++ b/caas/kubernetes/provider/k8s_test.go
@@ -3945,10 +3945,10 @@ func (s *K8sBrokerSuite) TestEnsureServiceWithServiceAccountNewClusterRoleCreate
 		s.mockStatefulSets.EXPECT().Get(gomock.Any(), "juju-operator-app-name", v1.GetOptions{}).
 			Return(nil, s.k8sNotFoundError()),
 		s.mockServiceAccounts.EXPECT().Create(gomock.Any(), svcAccount, v1.CreateOptions{}).Return(svcAccount, nil),
-		s.mockClusterRoles.EXPECT().Create(gomock.Any(), cr, v1.CreateOptions{}).Return(cr, nil),
-		s.mockClusterRoleBindings.EXPECT().List(gomock.Any(), v1.ListOptions{LabelSelector: "app.kubernetes.io/managed-by=juju,app.kubernetes.io/name=app-name,model.juju.is/name=test"}).
-			Return(&rbacv1.ClusterRoleBindingList{Items: []rbacv1.ClusterRoleBinding{}}, nil),
-		s.mockClusterRoleBindings.EXPECT().Create(gomock.Any(), crb, v1.CreateOptions{}).Return(crb, nil),
+		s.mockClusterRoles.EXPECT().Get(gomock.Any(), cr.Name, gomock.Any()).Return(cr, nil),
+		s.mockClusterRoles.EXPECT().Update(gomock.Any(), cr, gomock.Any()).Return(cr, nil),
+		s.mockClusterRoleBindings.EXPECT().Get(gomock.Any(), crb.Name, gomock.Any()).Return(crb, nil),
+		s.mockClusterRoleBindings.EXPECT().Update(gomock.Any(), crb, gomock.Any()).Return(crb, nil),
 		s.mockSecrets.EXPECT().Create(gomock.Any(), secretArg, v1.CreateOptions{}).Return(secretArg, nil),
 		s.mockStatefulSets.EXPECT().Get(gomock.Any(), "app-name", v1.GetOptions{}).
 			Return(nil, s.k8sNotFoundError()),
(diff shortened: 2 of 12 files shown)


caas/kubernetes/provider/utils/claim.go
```diff
@@ -0,0 +1,97 @@
+// Copyright 2021 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package utils
```
wallyworld, line 4: Maybe better in a `claim` package rather than utils?

caas/kubernetes/provider/utils/claim.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/juju/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+
+	"github.com/juju/juju/caas/kubernetes/provider/constants"
+)
+
+// Claim represents an assertion over a generic Kubernetes object to assert
+// ownership. These are used in Juju for cluster scoped resources to assert that
+// that Juju is not going to take ownership of an object that was not created by
+// it self.
```
wallyworld, line 18: itself

caas/kubernetes/provider/utils/claim.go
```diff
@@ -0,0 +1,97 @@
...
+type Claim interface {
+	// Assert defines the assertion to run. Returns true if a claim is asserted
+	// over the provided object or if an error occurred where a claim can not be
+	// made
```
wallyworld, line 22: made.

(and a few below as well)

caas/kubernetes/provider/utils/claim.go
```diff
@@ -0,0 +1,97 @@
...
+// If no claims are provided or no claim returns true false is returned. The
+// first claim to error stops execution
+func ClaimAggregateOr(claims ...Claim) Claim {
+	return ClaimFn(func(obj interface{}) (bool, error) {
+		for _, claim := range claims {
+			if r, err := claim.Assert(obj); err != nil {
+				return r, err
+			} else if r {
+				return r, err
+			}
```
wallyworld, line 52: Can this be inlined above
`if r, err := claim.Assert(obj); err != nil || r {`

caas/kubernetes/provider/utils/claim.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+		return false, nil
+	})
+}
+
+// claimHasAJujuLabel is a throw everything against the wall and see what sticks
+// assertion. It itterates all labels of the object trying to find a key that
+// has the lowercase word "juju". We use this because our labeling at one stage
+// is a bit hit and miss and no consitancy to fall back on.
+func claimHasJujuLabel(obj interface{}) (bool, error) {
```
wallyworld, line 62: This is only really for juju 2.8 right?
We should add a todo to drop for 3.0

caas/kubernetes/provider/rbac.go
```diff
@@ -357,44 +362,84 @@ func (k *kubernetesClient) listRoles(selector k8slabels.Selector) ([]rbacv1.Role
...
-			return nil, cleanups, errors.AlreadyExistsf("cluster role %q", cRole.GetName())
+		return nil, cleanups, errors.Annotatef(
+			err,
+			"ensuring cluster role %s",
+			cRole.Name,
+		)
+	}
+
+	if len(claims) != 0 {
+		hasClaim, err := utils.ClaimAggregateOr(claims...).Assert(existing)
```
wallyworld, line 424: We could just drop the `len(claims)` check and have `ClaimAggregator().Assert()` return `true` if there are no claims. This would make it cleaner here.

wallyworld (commented):
Looks good - some questions about the implementation to consider before progressing.

caas/kubernetes/provider/bootstrap.go
```diff
@@ -679,6 +680,8 @@ func (c *controllerStack) ensureControllerConfigmapAgentConf() error {
 }
 
 func (c *controllerStack) ensureControllerServiceAccount() error {
+	ctx := context.TODO()
```
wallyworld, line 683: `c.ctx` has a method to get the `context.Context()`
no need to create a new one here

caas/kubernetes/provider/bootstrap.go
```diff
@@ -704,7 +707,7 @@ func (c *controllerStack) ensureControllerServiceAccount() error {
 		return errors.Trace(err)
 	}
 
-	crb := &rbacv1.ClusterRoleBinding{
+	crb := resources.NewClusterRoleBinding(c.broker.GetCurrentNamespace(), &rbacv1.ClusterRoleBinding{
```
wallyworld, line 710: we have
`c.broker.GetCurrentNamespace(), // name cluster role binding after the controller namespace.`
below.
Maybe assign to a var with the comment and use the var

caas/kubernetes/provider/k8s_test.go
```diff
@@ -3945,10 +3945,10 @@ func (s *K8sBrokerSuite) TestEnsureServiceWithServiceAccountNewClusterRoleCreate
 		s.mockStatefulSets.EXPECT().Get(gomock.Any(), "juju-operator-app-name", v1.GetOptions{}).
 			Return(nil, s.k8sNotFoundError()),
 		s.mockServiceAccounts.EXPECT().Create(gomock.Any(), svcAccount, v1.CreateOptions{}).Return(svcAccount, nil),
-		s.mockClusterRoles.EXPECT().Create(gomock.Any(), cr, v1.CreateOptions{}).Return(cr, nil),
-		s.mockClusterRoleBindings.EXPECT().List(gomock.Any(), v1.ListOptions{LabelSelector: "app.kubernetes.io/managed-by=juju,app.kubernetes.io/name=app-name,model.juju.is/name=test"}).
-			Return(&rbacv1.ClusterRoleBindingList{Items: []rbacv1.ClusterRoleBinding{}}, nil),
-		s.mockClusterRoleBindings.EXPECT().Create(gomock.Any(), crb, v1.CreateOptions{}).Return(crb, nil),
+		s.mockClusterRoles.EXPECT().Get(gomock.Any(), cr.Name, gomock.Any()).Return(cr, nil),
```
wallyworld, line 3948: `gomock.Any()` is fine for the context, but we know the expected value of the 3rd arg right?

(here and elsewhere)

caas/kubernetes/provider/modeloperator.go
```diff
@@ -30,19 +32,14 @@ import (
 // commands. This interfaces is scoped down to the exact components needed by
 // the ensure model operator routines.
 type ModelOperatorBroker interface {
+	// Client returns the Kubernetes client to use for model operator actions
```
wallyworld, line 35: actions.

caas/kubernetes/provider/modeloperator.go
```diff
@@ -549,6 +528,12 @@ func ensureModelOperatorRBAC(
 	labels map[string]string,
 ) (string, []func(), error) {
 	cleanUpFuncs := []func(){}
+	ctx := context.TODO()
```
wallyworld, line 531: since we have a context on the caller of this method, we should pass it in

caas/kubernetes/provider/modeloperator.go
```diff
@@ -549,6 +528,12 @@ func ensureModelOperatorRBAC(
...
+
+	k8sClient := broker.Client()
+	if k8sClient == nil {
+		return operatorName, cleanUpFuncs, errors.New("kubernetes client cannot be nil")
+	}
```
wallyworld, line 536: this check should be done when the broker is constructed
`broker.Client()` should never return nil

caas/kubernetes/provider/rbac.go
```diff
@@ -47,6 +48,7 @@ func RBACLabels(appName, model string, global, legacy bool) map[string]string {
 func (k *kubernetesClient) ensureServiceAccountForApp(
 	appName string, annotations map[string]string, rbacDefinition k8sspecs.K8sRBACSpecConverter,
 ) (cleanups []func(), err error) {
+	ctx := context.TODO()
```
wallyworld, line 51: as per other comment

caas/kubernetes/provider/resources/clusterrole.go
```diff
@@ -90,6 +90,43 @@ func (r *ClusterRole) Delete(ctx context.Context, client kubernetes.Interface) e
 	return nil
 }
 
+func (r *ClusterRole) Ensure(
```
wallyworld, line 93: comment (both places)

caas/kubernetes/provider/resources/clusterrole.go
```diff
@@ -102,3 +139,20 @@ func (r *ClusterRole) ComputeStatus(_ context.Context, _ kubernetes.Interface, n
 	}
 	return "", status.Active, now, nil
 }
+
+func (r *ClusterRole) Update(ctx context.Context, client kubernetes.Interface) error {
```
wallyworld, line 143: comment (here and on ClusterRoleBinding)

wallyworld (approved):
A few small things to tweak before landing, mainly the use of the context we already have and the client nil handling

caas/kubernetes/provider/k8s_test.go
tlm: no we shouldn't be testing get opts here as they could be dynamic on context. We should just be checking pre and post as the function docs not implementation. This is why mocks are bad!

tlm:
$$merge$$

PR #13040, opened 2021-06-01
Integration tests: Extend timeout

SimonRichardson:
We seem to be failing a lot waiting for a machine to provision or an
application to reach idle. Either way, we're struggling to reach the
goal state. I don't think waiting longer than 15minutes is advisable,
so let's set that for now.

## QA steps

```sh
(cd test && ./main.sh charmhub)
```


manadart (approved):

SimonRichardson:
$$merge$$

PR #13041, opened 2021-06-02
Update apt mirrors on all nodes when model value changes

wallyworld:
The apt-mirror model config now updates the `/etc/apt/sources.list` whenever it changes. Previously, the only time apt-mirror was used was when the machine instance was started.
As with instance startup, setting the apt-mirror value will update both the apt and security archives.

This change now also allow the apt mirrors to be updated inside k8s containers. In particular, the container running the charm now gets the apt mirror values from the model config.

Instead of making the apt updates scripts manually in juju, we now rely on the juju/package upstream to do that, so there's a bunch of deleted code also.

## QA steps

juju bootstrap aws
juju add-machine 
juju add-machine lxd:0
ssh to both machines
cat /etc/apt/sourcs.list and see the archives mirrors are stock
juju model-config apt-mirror=http://foo
juju add-machine 
juju add-machine lxd:0
ssh to all 4 machines
cat /etc/apt/sourcs.list and see the archives mirrors are set to http://foo

bootstrap k8s
juju model-config apt-mirror=http://foo
deploy snappass-test
ssh to the charm container
cat /etc/apt/sourcs.list and see the archives mirrors are set to http://foo

## Bug reference

https://bugs.launchpad.net/juju/+bug/1929399

### DIFF
--- a/api/proxyupdater/proxyupdater.go
+++ b/api/proxyupdater/proxyupdater.go
@@ -80,6 +80,8 @@ type ProxyConfiguration struct {
 	APTProxy    proxy.Settings
 	SnapProxy   proxy.Settings
 
+	AptMirror string
+
 	SnapStoreProxyId         string
 	SnapStoreProxyAssertions string
 	SnapStoreProxyURL        string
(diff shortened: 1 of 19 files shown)


ycliuhw (approved):
LGTM thanks

wallyworld:
$$merge$$

wallyworld:
[apt-mirror on k8s](https://trello.com/c/Sb8O8nGe/1612-apt-mirror-on-k8s)

PR #13048, opened 2021-06-03
Data race state/logdb tests

hmlanigan:
DataRace between creating the BufferedLoggerSuite mockLogger and testclock, and the clock being used in a go routine started by the testclock's AfterFunc.  We can't guarantee when the go routine finishes in the test, so the next test can overwrite the mockLogger and testclock in use by the prior test.

Create local mockLogger and testclock for each test instead.

## QA steps

The real test will be to no longer see this data race in Unit-RunUnitTests-race-amd64.

I had to contrive a delay in the go routine called by AfterFunc to reproduce locally.  Then change the timeout in BufferedLoggerSuite.waitFlush to allow the tests to pass.

```code
~/gohack/github.com/juju/clock/testclock$ diff -c  clock.go  clock.go.orig
*** clock.go	2021-06-03 12:29:20.469892823 +0000
--- clock.go.orig	2021-06-02 21:48:01.791345270 +0000
***************
*** 82,91 ****
  // AfterFunc is part of the clock.Clock interface.
  func (clock *Clock) AfterFunc(d time.Duration, f func()) clock.Timer {
  	return clock.addAlarm(d, nil, func() {
! 		go func() {
! 			time.Sleep(time.Second * 10)
! 			f()
! 		}()
  	})
  }

--- 82,88 ----
  // AfterFunc is part of the clock.Clock interface.
  func (clock *Clock) AfterFunc(d time.Duration, f func()) clock.Timer {
  	return clock.addAlarm(d, nil, func() {
! 		go f()
  	})
  }
```

and 
```code
$ git diff
diff --git a/state/logdb/buf_test.go b/state/logdb/buf_test.go
index 1ab0771293..aa5b3fcd70 100644
--- a/state/logdb/buf_test.go
+++ b/state/logdb/buf_test.go
@@ -31,7 +31,7 @@ func (s *BufferedLoggerSuite) waitFlush(c *gc.C, mock *mockLogger) []state.LogRe
        select {
        case records := <-mock.called:
                return records
-       case <-time.After(coretesting.LongWait):
+       case <-time.After(coretesting.LongWait*2):
        }
        c.Fatal("timed out waiting for logs to be flushed")
        panic("unreachable")
```

Run `go test -race -v ./state/logdb/...` multiple times.

### DIFF
--- a/state/logdb/buf_test.go
+++ b/state/logdb/buf_test.go
@@ -21,20 +21,15 @@ var _ = gc.Suite(&BufferedLoggerSuite{})
 
 type BufferedLoggerSuite struct {
 	testing.IsolationSuite
-
-	mock  mockLogger
-	clock *testclock.Clock
 }
 
 func (s *BufferedLoggerSuite) SetUpTest(c *gc.C) {
 	s.IsolationSuite.SetUpTest(c)
-	s.mock = mockLogger{}
-	s.clock = testclock.NewClock(time.Time{})
 }
 
-func (s *BufferedLoggerSuite) waitFlush(c *gc.C) []state.LogRecord {
+func (s *BufferedLoggerSuite) waitFlush(c *gc.C, mock *mockLogger) []state.LogRecord {
 	select {
-	case records := <-s.mock.called:
+	case records := <-mock.called:
 		return records
 	case <-time.After(coretesting.LongWait):
 	}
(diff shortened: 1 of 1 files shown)


hmlanigan:
https://jenkins.juju.canonical.com/job/Unit-RunUnitTests-race-amd64/611/console

achilleasa (approved):
LGTM

hmlanigan:
$$merge$$

PR #13053, opened 2021-06-07
Integration-tests: Add lxd-remote tests

SimonRichardson:
The following ensures that we also handle lxd and lxd-remote providers.

## QA steps

This should fix the CI integration manual test.


achilleasa (approved):

SimonRichardson:
$$merge$$

PR #13054, opened 2021-06-07
Integration tests: Server IP address

SimonRichardson:
The server IP address for the metadata service has to be routable, but
not the loopback one to work in every scenario. In this case we just
make this more robust by iterating over the `hostname -I` command and
finding the first one that works correctly.

This should be enough for juju to correctly bootstrap from a local file
system.

## QA steps

```sh
(cd tests && ./main.sh bootstrap)
```


manadart (approved):
Works.

SimonRichardson:
$$merge$$

PR #13056, opened 2021-06-08
Remove churn deploying bundles with default config values

wallyworld:
When deploying over the top of an existing model a bundle that contains charm config values set to the charm default, juju would claim that config had changed when it hadn't. Any default charm values were being removed from the model representation to compare, instead of unset values.

## QA steps

juju deploy a simple bundle with a config option set to the charm default
```
applications:
    mariadb:
        charm: cs:mariadb
        num_units: 1
        options:
          dataset-size: 50%
```
`juju export-bundle --filename foo.yaml`
`juju deploy ./foo.yaml --verbose --dry-run --map-machines existing`

There should be no changes printed.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1929908

### DIFF
--- a/cmd/juju/application/bundle.go
+++ b/cmd/juju/application/bundle.go
@@ -1518,7 +1518,7 @@ func applicationConfigValue(key string, valueMap interface{}) (interface{}, erro
 	if !found {
 		return nil, errors.Errorf("missing application config value 'source' for key %q", key)
 	}
-	if source != "user" {
+	if source == "unset" {
 		return nil, nil
 	}
 	value, found := vm["value"]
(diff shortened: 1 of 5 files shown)


hpidcock (approved):
Just need to fix allwatcher tests

wallyworld:
$$merge$$

wallyworld:
[juju deploy ./bundle.yaml on an existing model causes all units to run config-changed](https://trello.com/c/sHbC4imt/1625-juju-deploy-bundleyaml-on-an-existing-model-causes-all-units-to-run-config-changed)

PR #13063, opened 2021-06-10
Fix how default app constraint it set

wallyworld:
During 2.9 release candidate stages, a change was made to, at the time of deploying an app, set the app arch constraint to "amd64" if not otherwise set. This was the wrong place to do it. We should only record user intent and fill in any defaults at the appropriate time. The approach used broke constraints handling: arch is not compatible with instance-type and because we were setting arch to "amd64", if the user had specified "instance-type=foo", juju complained about ambiguous constraints (see bug).

The app constraint is only used to inform a unit constraint if none is other wise specified when adding a unit. And the unit constraint is only used to set up the machine constraint when provisioning a machine. So this PR moves the default arch logic to the Constraints() method on unit; moreover, it only fills in default arch if instance-type is not set, since instance type fixes the arch anyway. If we do need to fill in a default constraint to use to provision a machine, we actually look at the unit's charm manifest and use the platform arch if there is one. This ensures the correct machine is provisioned.

Also, there were some bundle deploy tests that appeared to be quite wrong, meaning we were testing for incorrect behaviour. Specifically, lines like this in the test output are wrong:
```
- set constraints for mysql to ""
- set constraints for wordpress to ""
```

The other thing is that apps from migrated models do not get this special treatment as far as I can see so they would be missing this added default arch.

We can still improve error messaging though. Consider an amd64 only charm deployed to an arm64 instance type

`juju deploy somecharm --constraints="instance-type=a1.medium"`

deploy "succeeds" but status subsequently shows

```
failed to start machine 0 (chosen architecture arm64 for image "ami-0ca133384d447f902" not present in [amd64]), retrying in 10s (9 more attempts)
```

which is not very inormative

## QA steps

As per bug, bootstrap azure and
`juju deploy mysql --constraints="instance-type=Standard_D4_v4"`

## Bug reference

https://bugs.launchpad.net/juju/+bug/1931504

### DIFF
--- a/apiserver/facades/client/application/application_test.go
+++ b/apiserver/facades/client/application/application_test.go
@@ -404,7 +404,7 @@ func (s *applicationSuite) TestApplicationDeployWithStorage(c *gc.C) {
 		},
 	}
 
-	cons := constraints.MustParse("arch=amd64")
+	var cons constraints.Value
 	args := params.ApplicationDeploy{
 		ApplicationName: "application",
 		CharmURL:        curl.String(),
@@ -485,7 +485,7 @@ func (s *applicationSuite) TestApplicationDeployDefaultFilesystemStorage(c *gc.C
 		URL: curl.String(),
 	}, s.openRepo)
 	c.Assert(err, jc.ErrorIsNil)
-	cons := constraints.MustParse("arch=amd64")
+	var cons constraints.Value
 	args := params.ApplicationDeploy{
 		ApplicationName: "application",
 		CharmURL:        curl.String(),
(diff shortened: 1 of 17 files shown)


wallyworld:
[Instance type constraint throws "ambiguous constraints" error](https://trello.com/c/oHyYm04F/1636-instance-type-constraint-throws-ambiguous-constraints-error)

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
> It would be really good to fix the error message from the invalid constraint value.

As discussed live, this can be fixed in a separate PR

wallyworld:
$$merge$$

PR #13064, opened 2021-06-11
Fixes controller correlation for pre 2.9

tlm:
During upgrades to 2.9 Juju cannot find controller namespaces that are pre 2.9

## Checklist

## QA steps

Run unit tests as 2.8 -> 2.9 upgrading for k8s is still not 100%

### DIFF
--- a/caas/kubernetes/provider/bootstrap.go
+++ b/caas/kubernetes/provider/bootstrap.go
@@ -24,7 +24,9 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/juju/juju/agent"
 	agenttools "github.com/juju/juju/agent/tools"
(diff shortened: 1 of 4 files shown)


caas/kubernetes/provider/bootstrap.go
```diff
@@ -152,22 +154,63 @@ type controllerStacker interface {
 	Deploy() error
 }
 
-func controllerCorelation(broker *kubernetesClient) (string, error) {
-	// ensure controller specific annotations.
-	controllerUUIDKey := k8sutils.AnnotationControllerIsControllerKey(false)
-	_ = broker.addAnnotations(controllerUUIDKey, "true")
+// findControllerNamespace is used for finding a controllers namespace based on
```
wallyworld, line 157: controller's

caas/kubernetes/provider/bootstrap.go
```diff
@@ -152,22 +154,63 @@ type controllerStacker interface {
...
+// it's model name and controller uuid. This function really shouldn't exists
+// and should be looked to removed in 3.0. We have it hear as we are still
+// trying to use Kubernetes annotations as database selectors in some parts of
+// Juju.
+func findControllerNamespace(
+	client kubernetes.Interface,
+	modelName string,
```
wallyworld, line 164: If we are finding the controller namespace, there's no need to pass in a model name.
The caller of this method already ensures that this method is only ever called with model name equal to `environsbootstrap.ControllerModelName`. So we can drop the model name param and use that const

wallyworld (approved):
Need to fix the apostrophe abuse plus we don't need to pass in model name as a param as it's always equal to `environsbootstrap.ControllerModelName`

tlm:
$$merge$$

tlm:
$$merge$$

PR #13068, opened 2021-06-14
Integration tests: Fix intermittent failure

SimonRichardson:
Ensure we capture all the output from stdout before streaming to grep
via pipe. This shouldn't happen, but does happen. Because of newlines
(\n) we can end up not capturing all of the output. The fix is to send
it to a variable and then check that information.

## QA steps

```sh
(cd tests && ./main.sh cli)
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1931587


achilleasa (approved):

SimonRichardson:
$$merge$$

PR #13070, opened 2021-06-14
Integration tests: Bootstrap series

SimonRichardson:
The natural bootstrap series for the integration tests are focal.
Unfortunately for us, focal can't be used inside multiple nested
containers, we have to resort to using bionic.
With this in mind, we allow the setting of a bootstrap-series for the
main deployments, when not using that flow, we need to ensure that we
also, pass that through to the new cases.

The change although simple should be enough to get some of the
integration tests working in the CI infra.

## QA steps

```sh
(cd tests && ./main.sh bootstrap)
```


manadart (approved):
Gets us more green. Nicer if upstream cleaned house.

SimonRichardson:
$$merge$$

PR #13077, opened 2021-06-15
EC2: Use juju/http client

SimonRichardson:
The following updates the ec2 client factory to use the juju/http
client instead of the one offered by the library. We can reduce a lot of
the client boilerplate by just surfacing the only client.

More work is required to lift it out of the provider into a manifold,
but that work can be done at a later PR.

### Tests should pass.

```sh
go test -v ./provider/ec2/...
```

### Bootstrap to ec2

```sh
$ juju bootstrap aws test
```

## Bug reference

### DIFF
--- a/agent/agentbootstrap/bootstrap.go
+++ b/agent/agentbootstrap/bootstrap.go
@@ -4,6 +4,7 @@
 package agentbootstrap
 
 import (
+	stdcontext "context"
 	"fmt"
 	"path/filepath"
 
@@ -390,9 +391,9 @@ func getEnviron(
 		Config:         modelConfig,
 	}
 	if cloudSpec.Type == cloud.CloudTypeCAAS {
-		return caas.Open(provider, openParams)
+		return caas.Open(stdcontext.TODO(), provider, openParams)
 	}
-	return environs.Open(provider, openParams)
+	return environs.Open(stdcontext.TODO(), provider, openParams)
 }
 
 func initRaft(agentConfig agent.Config) error {
--- a/apiserver/common/credentialcommon/modelcredential.go
+++ b/apiserver/common/credentialcommon/modelcredential.go
@@ -76,7 +76,7 @@ func checkCAASModelCredential(brokerParams environs.OpenParams) (params.ErrorRes
 }
 
 func checkIAASModelCredential(openParams environs.OpenParams, backend PersistentBackend, callCtx context.ProviderCallContext, checkCloudInstances bool) (params.ErrorResults, error) {
-	env, err := newEnv(openParams)
+	env, err := newEnv(callCtx, openParams)
 	if err != nil {
 		return params.ErrorResults{}, errors.Trace(err)
 	}
--- a/apiserver/common/credentialcommon/modelcredential.go
+++ b/apiserver/common/credentialcommon/modelcredential.go
@@ -4,6 +4,8 @@
 package credentialcommon
 
 import (
+	stdcontext "context"
+
 	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 	"github.com/juju/names/v4"
@@ -63,7 +65,7 @@ func ValidateNewModelCredential(backend PersistentBackend, callCtx context.Provi
 }
 
 func checkCAASModelCredential(brokerParams environs.OpenParams) (params.ErrorResults, error) {
-	broker, err := newCAASBroker(brokerParams)
+	broker, err := newCAASBroker(stdcontext.TODO(), brokerParams)
 	if err != nil {
 		return params.ErrorResults{}, errors.Trace(err)
 	}
--- a/apiserver/facades/client/modelmanager/modelmanager.go
+++ b/apiserver/facades/client/modelmanager/modelmanager.go
@@ -8,6 +8,7 @@
 package modelmanager
 
 import (
+	stdcontext "context"
 	"fmt"
 	"sort"
 	"time"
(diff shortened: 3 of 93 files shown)


SimonRichardson:
This isn't quite ready for review, but I want CI to find all other issues at the same time.

agent/agentbootstrap/bootstrap.go
```diff
@@ -390,9 +391,9 @@ func getEnviron(
 		Config:         modelConfig,
 	}
 	if cloudSpec.Type == cloud.CloudTypeCAAS {
-		return caas.Open(provider, openParams)
+		return caas.Open(stdcontext.TODO(), provider, openParams)
```
achilleasa, line 394: This should be `stdcontext.Background()`. The TODO one should only be used in tests.

apiserver/common/credentialcommon/modelcredential.go
```diff
@@ -63,7 +65,7 @@ func ValidateNewModelCredential(backend PersistentBackend, callCtx context.Provi
 }
 
 func checkCAASModelCredential(brokerParams environs.OpenParams) (params.ErrorResults, error) {
-	broker, err := newCAASBroker(brokerParams)
+	broker, err := newCAASBroker(stdcontext.TODO(), brokerParams)
```
achilleasa, line 68: `stdcontext.Background()`

apiserver/facades/client/modelmanager/modelmanager.go
```diff
@@ -621,7 +622,7 @@ func (m *ModelManagerAPI) newCAASModel(
 	if err != nil {
 		return nil, errors.Annotate(err, "getting controller config")
 	}
-	broker, err := m.getBroker(environs.OpenParams{
+	broker, err := m.getBroker(stdcontext.TODO(), environs.OpenParams{
```
achilleasa, line 625: `stdcontext.Background()`

caas/kubernetes/provider/provider.go
```diff
@@ -44,7 +45,7 @@ var providerInstance = kubernetesEnvironProvider{
 	cmdRunner:          defaultRunner{},
 	builtinCloudGetter: attemptMicroK8sCloud,
 	brokerGetter: func(args environs.OpenParams) (caas.ClusterMetadataChecker, error) {
-		return caas.New(args)
+		return caas.New(stdcontext.TODO(), args)
```
achilleasa, line 48: `stdcontext.Background()`

cmd/juju/caas/add.go
```diff
@@ -718,7 +719,7 @@ func (c *AddCAASCommand) newK8sClusterBroker(cloud jujucloud.Cloud, credential j
 		}
 		openParams.ControllerUUID = ctrlUUID
 	}
-	return caas.New(openParams)
+	return caas.New(stdcontext.TODO(), openParams)
```
achilleasa, line 722: `stdcontext.Background()`

cmd/juju/caas/update.go
```diff
@@ -159,7 +160,7 @@ func (c *UpdateCAASCommand) newK8sClusterBroker(cloud jujucloud.Cloud, credentia
 		}
 		openParams.ControllerUUID = ctrlUUID
 	}
-	return caas.New(openParams)
+	return caas.New(stdcontext.TODO(), openParams)
```
achilleasa, line 163: `stdcontext.Background()`

achilleasa (changes requested):
Please change all `stdcontext.TODO()` references to `stdcontext.Background()` in non-test code.

cmd/plugins/juju-metadata/imagemetadata.go
```diff
@@ -26,12 +27,12 @@ import (
 	"github.com/juju/juju/version"
 )
 
-func prepare(context *cmd.Context, controllerName string, store jujuclient.ClientStore) (environs.Environ, error) {
+func prepare(ctx *cmd.Context, controllerName string, store jujuclient.ClientStore) (environs.Environ, error) {
```
achilleasa, line 30: We should investigate how much effort it would require to thread a context through `cmd.Context` so we can pass it through.

provider/ec2/environ.go
```diff
@@ -2530,7 +2537,24 @@ func (e *environ) SetCloudSpec(spec environscloudspec.CloudSpec) error {
 		return errors.Trace(err)
 	}
 
-	e.ec2Client = EC2Session(e.cloud.Region, e.ec2.AccessKey, e.ec2.SecretKey)
+	// Allow the passing of a client func through the context. This allows
+	// passing the client from outside of the environ, one that allows for
+	// custom http.Clients.
+	if value := ctx.Value(AWSClientContextKey); value == nil {
```
achilleasa, line 2543: This seems like a horrible hack to me :-(

I really wish we could just take the time and move all tests into the ec2 package so we could simply call the constructor as a proper unit-test would do...

cmd/plugins/juju-metadata/imagemetadata.go
achilleasa: The BootstrapContext also has a context.Context field.

achilleasa (approved):

provider/ec2/environ.go
SimonRichardson: I see this as one step forward, but necessarily the final step. I'm not that enamored in how we create an environ, to begin with. I think if we can keep iterating on this a little more, I think we will get to a place of just passing in a value via the constructor.

SimonRichardson: https://github.com/juju/juju/pull/13077/commits/cb4f27d285fe892b253b254ce3271c25d8d443aa

SimonRichardson:
$$merge$$

PR #13078, opened 2021-06-15
lp 1928796: sidecar charm via bundle

hmlanigan:
Juju verifies the series of a charm and application in bundle data many times in an attempt to fail early if there is an issue.   However this makes it difficult to ensure that all places series is determined do it the correct way for any given point in time.

Take in the charm package change to not assign a series of Kubernetes to any application listed in a Kubernetes bundle.  Allow juju itself to determine which series an application in a bundle uses.  2.8 style Kubernetes charms use "kubernetes" as the series, but 2.9 sidecar charms would use "focal" or the like based on what's in their manifest.yaml.  

Pull NewCharmAtPath and NewCharmAtPathForceSeries in from charmrepo, to update how the series is determined for a local sidecar charm.

Remove code to add "kubernetes" to the list of supported series, it's duplicating code in ComputedSeries.

## QA steps

Follow the bug reproducer steps. 

Deploy a non-kubernetes bundle.  Deploy other bundles with local charms including manifest files.

Upgrade local charms.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1928796


SimonRichardson (commented):
```yaml
bundle: kubernetes
applications:
  qa:
    charm: ./testcharms/charm-hub/charms/juju-qa-test-v2/juju-qa-test-v2.charm
    num_units: 1
```

```sh
$ juju deploy ./bundle.yaml
ERROR cannot deploy bundle: series not specified and charm does not define any
```

I think we could just fallback to the model default series here.

hmlanigan:
@SimonRichardson I could not reproduce what you reported.  I tried the bundle with and with type: kubernetes and on both a localhost and microk8s controller. 

I did some clean up in bundlehandler.go while investigating.  It's all in the last commit.

SimonRichardson:
So I can replicate this with a bad charm, so ignore me. `charmcraft build` a charm and then remove the manifest file from the zip file (it was an old charm I was testing with) and I rebuilt it and it works fine.

SimonRichardson (approved):

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #13079, opened 2021-06-15
Fix LP: 1930761 - use pre 2.9 juju vsphere templates without an arch tag

hmlanigan:
Juju 2.9 tries to match VM templates by series and arch.  However templates created prior to juju 2.9 do not have an arch tag.  Prefer templates which match the requested arch, but do not fail if a matching template, by series, exists.

## QA steps

``` console
sudo snap refresh juju --channel 2.8/stable
/snap/bin/juju bootstrap vsphere
/snap/bin/juju deploy ubuntu

juju add-unit ubuntu
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1930761

### DIFF
--- a/provider/vsphere/vm_template.go
+++ b/provider/vsphere/vm_template.go
@@ -130,6 +130,7 @@ func (v *vmTemplateManager) getVMArch(ctx context.Context, vmObj *object.Virtual
 }
 
 func (v *vmTemplateManager) getImportedTemplate(ctx context.Context, series string, arches []string) (*object.VirtualMachine, string, error) {
+	logger.Tracef("getImportedTemplate for series %q, arches %q", strings.Join(arches, ", "))
 	seriesTemplatesFolder := v.seriesTemplateFolder(series)
 	seriesTemplates, err := v.client.ListVMTemplates(ctx, path.Join(seriesTemplatesFolder, "*"))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


provider/vsphere/vm_template.go
```diff
@@ -138,29 +139,26 @@ func (v *vmTemplateManager) getImportedTemplate(ctx context.Context, series stri
...
 	var vmTpl *object.VirtualMachine
 	var arch string
 	if len(arches) > 0 {
 		for _, item := range seriesTemplates {
 			arch, err = v.getVMArch(ctx, item)
 			if err != nil {
-				if errors.IsNotFound(err) {
-					continue
-				}
+				logger.Debugf("failed getVMArch for template %q: %s", item.InventoryPath, err)
```
jameinel, line 150: Do we want to ignore all errors here?
I can understand the IsNotFound but what other errors are we getting that we are essentially just turning into a Warning.

wallyworld (approved):

jameinel (commented):
I'm happy with the general changes, I'm concerned that we're changing all errors into just a debug message that is going to be hidden by default.

provider/vsphere/vm_template.go
hmlanigan: The arch found could not be a string, or the client call to get the data failed.

hmlanigan:
$$merge$$

PR #13083, opened 2021-06-16
state: Add UserConnections to the all watcher

mhilton:
Add all watcher events for notifying on user connections. This implements
the feature requested in https://bugs.launchpad.net/juju/+bug/1929524.

## QA steps

 1. Create a controller with this code version.
 2. Connect a client that calls `WatchAllModels` on the `Controller` facade and displays the results.
 3. Connect to a model on the controller (juju status)
 4. Verify the event stream includes a userConnection event with the connection time around step 3.

## Documentation changes

This adds a new event type into the AllWatcher event stream. The "userConnection" events will be fired every time a user connects to a model.

## Bug reference

- https://bugs.launchpad.net/juju/+bug/1929524

### DIFF
--- a/apiserver/params/multiwatcher.go
+++ b/apiserver/params/multiwatcher.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/juju/charm/v9"
@@ -109,6 +110,8 @@ func (d *Delta) UnmarshalJSON(data []byte) error {
 		d.Entity = new(RemoteApplicationUpdate)
 	case "unit":
 		d.Entity = new(UnitInfo)
+	case "userConnection":
+		d.Entity = new(UserConnectionInfo)
 	default:
 		return errors.Errorf("Unexpected entity name %q", entityKind)
 	}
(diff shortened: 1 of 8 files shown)


core/multiwatcher/types.go
```diff
@@ -29,6 +30,7 @@ const (
 	RelationKind          = "relation"
 	RemoteApplicationKind = "remoteApplication"
 	UnitKind              = "unit"
+	UserConnectionKind    = "user-connection"
```
SimonRichardson, line 33: Can we be consistent here and do `userConnection`?

SimonRichardson:
We'll have to ensure that pylibjuju can also support this in case it starts showing up. The following change is just to add it to the [delta types](https://github.com/juju/python-libjuju/blob/master/juju/delta.py#L112-L123).

apiserver/params/multiwatcher.go
```diff
@@ -447,3 +450,17 @@ func (i *BranchInfo) EntityId() EntityId {
 		Id:        i.Id,
 	}
 }
+
+type UserConnectionInfo struct {
```
alesstimec, line 454: godoc, please

apiserver/params/multiwatcher.go
```diff
@@ -447,3 +450,17 @@ func (i *BranchInfo) EntityId() EntityId {
...
+	ModelUUID      string    `json:"model-uuid"`
+	Username       string    `json:"user-name"`
+	LastConnection time.Time `json:"last-connection,omitempty"`
+}
+
+func (i *UserConnectionInfo) EntityId() EntityId {
```
alesstimec, line 460: godoc, please

alesstimec (approved):

SimonRichardson (changes requested):
We need to retarget this against 2.9

mhilton:
I'm closing this to replace with #13086

PR #13086, opened 2021-06-17
state: Add UserConnections to the all watcher

mhilton:
Add all watcher events for notifying on user connections. This implements
the feature requested in https://bugs.launchpad.net/juju/+bug/1929524.

## Checklist

## QA steps

 1. Create a controller with this code version.
 2. Connect a client that calls WatchAllModels on the Controller facade and displays the results.
 3. Connect to a model on the controller (juju status)
 4. Verify the event stream includes a userConnection event with the connection time around step 3.

## Documentation changes

This adds a new event type into the AllWatcher event stream. The "userConnection" events will be fired every time a user connects to a model.

## Bug reference

 - https://bugs.launchpad.net/juju/+bug/1929524


wallyworld (approved):
This looks pretty good.
The thing we need to be sure of though is does the replacement of UserPermissions with Users in the ModelUpdate struct cause issues with older machine agents that may expect the old data. I haven't dug right into it - we may be ok - but it worth being sure. We would need to check the api side unmarshalling of the struct and where it's used. If there's an issue, we may need to keep both values.

hpidcock:
> This looks pretty good.
> The thing we need to be sure of though is does the replacement of UserPermissions with Users in the ModelUpdate struct cause issues with older machine agents that may expect the old data. I haven't dug right into it - we may be ok - but it worth being sure. We would need to check the api side unmarshalling of the struct and where it's used. If there's an issue, we may need to keep both values.

ModelUpdate params struct doesn't look to have had UserPermissions in it. I think this change is safe, I'll try and land it for 2.9.8 rather than 2.9.7.

worker/modelcache/worker.go
```diff
@@ -421,7 +430,7 @@ func (c *cacheWorker) translateModel(d multiwatcher.Delta) interface{} {
 		Config:          value.Config,
 		Status:          coreStatus(value.Status),
 		// TODO: constraints, sla
-		UserPermissions: value.UserPermissions,
+		UserPermissions: userPermissions,
```
hpidcock, line 433: This here is the relevant change that maintains compatibility.

hpidcock:
@mhilton I'm unable to get WatchAllModels to show the behavior you expect, can you verify on your end please.

PR #13091, opened 2021-06-18
Migrate the ec2 provider to use the new aws sdk

wallyworld:
We need to be able to provision ec2 instances with encrypted root disks. The current sdk (gopkg.in/amz.v3/aws) doesn't support the required functionality and is somewhat out of date now. There's a new official SDK https://aws.github.io/aws-sdk-go-v2. This PR migrates the ec2 provider to use that instead.

The main churn is adapting to the new data types used to send and receive api call results. And there's a different upstream error struct so slightly different error messages, but the core error codes are the same.

Some of the ec2 provider - the instance type discovery - had been implemented with v1 of the official SDK. So all of that was moved across to v2 as well. The tests for this code which used a custom mock were deleted and now everything is provided by a single test server under `internal/testing`.

This new testing backend is simpler in concept than that used before as provided by the amz.v3 upstream - it stubs out the client API calls directly, instead of using a http transport, which totally should not the concern of the unit tests. The new testing server is based on what was done in amz.v3.

There is a little removed code also - back in the juju 2.0 RC days, allowance was made for incorrect public cloud metadata, which has long since been fixed.

## QA steps

bootstrap ontp aws
deploy a charm with storage
juju status --storage --format yaml to ensure storage becomes provisioned and attached
add a unit and aain check storage
remove the application and see that the volumes become detached but still provisioned
deploy another app with storage
remove the application with --destroy-storage and see the volumes get deleted
add a unit with --attach-storage and reuse a detached volume from a removed unit
check that zone placement directives work also


wallyworld:
[migrate ec2 provider to new sdk](https://trello.com/c/Ty6nXTQI/1644-migrate-ec2-provider-to-new-sdk)

wallyworld:
!!build!!

hpidcock (approved):
LGTM, was a long read but thank-you, I appreciate this change.

The test implementations are definitely better than mocks, so that is nice to see.

hpidcock:
$$merge$$

PR #13097, opened 2021-06-22
[2.8] Ensure all DB errors when inserting logs are always logged to logsink

achilleasa:
This improves visibility in cases such as LP1930899 where
underprovisioned mongodb instance prevented log entries from being
persisted with no additional error output.

As the DB error was returned back to the caller (log sender worker), the
worker would keep restarting thus preventing models from logging
anything.

While this commit does not fix the underlying problem (underprovisioned
mongo) it should at least make the cause of the problem more obvious to
users that are trying to diagnose similar issues by inspecting the
logsink.log file contents.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1930899

(NOTE: the PR does not fix ^ but helps diagnose the root cause of the issue)

### DIFF
--- a/apiserver/logsink.go
+++ b/apiserver/logsink.go
@@ -176,6 +176,19 @@ func (s *agentLoggingStrategy) WriteLog(m params.LogRecord) error {
 		Message:  m.Message,
 	}}), "logging to DB failed")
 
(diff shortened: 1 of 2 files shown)


SimonRichardson (approved):

achilleasa:
$$merge$$

PR #13099, opened 2021-06-22
Packaging: Update packaging dependency.

SimonRichardson:
The following brings in the new juju/packaging dependency. The updated
library allows for better retries for apt, yum, zypper and snap. With
the latter working as previously it never did before hand. Zypper
doesn't allow retries at the moment, because there aren't any retry
codes that we can use from the zypper CLI.

The issue was originally spotted in CI integration tests during bootstrap.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju bootstrap lxd old --bootstrap-series=bionic
$ juju bootstrap aws ec2test
```


achilleasa (approved):

hpidcock:
!!build!!

SimonRichardson:
$$merge$$

PR #13104, opened 2021-06-23
fix lp:1931708 -  run unit state upgrade step from machine agent.

hmlanigan:
Combining the machine and unit agents in 2.9 meant that prior upgrade steps requiring a unit agent would no longer be run, as the unit agent no longer has an upgrade-step worker.  

There is only one pre juju 2.9 upgrade step that requires the unit agent.  Revert a prior fix to run as unit agents, instead of the machine agent.  

Given that the unit agents are now stopped when the machine agent runs its upgrade steps.  There is little value in threading an upgrade-steps worker in the nested unit agent.

## QA steps

```console
# Setup a controller and model with 2.7
sudo snap refresh juju --channel 2.7/stable
/snap/bin/juju bootstrap localhost upgrade-me
/snap/bin/juju deploy wordpress
/snap/bin/juju deploy mysql
/snap/bin/juju add-relation wordpress mysql

# upgrade the controller and model, ensure you go directly from 2.7.x to the test code.
juju upgrade-model --agent-version 2.9.7.1

# The wordpress agent should not go into an error state, nor should the install hook be run after upgrade.
juju show-status-log wordpress/0
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1931708

### DIFF
--- a/upgrades/steps_28.go
+++ b/upgrades/steps_28.go
@@ -111,12 +111,20 @@ func moveUnitAgentStateToController(ctx Context) error {
 	// Lookup the names of all unit agents installed on this machine.
 	agentConf := ctx.AgentConfig()
 	ctxTag := agentConf.Tag()
-	if ctxTag.Kind() != names.UnitTagKind {
-		logger.Infof("skipping agent %q, not a unit", ctxTag.String())
+	if ctxTag.Kind() != names.MachineTagKind {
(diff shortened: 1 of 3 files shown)


manadart (approved):
Nicely done.

hmlanigan:
$$merge$$

PR #13107, opened 2021-06-24
Github actions: Update homebrew formulae

SimonRichardson:
The following adds a new github action to update the homebrew formulae
based on a newly published tag or on a 12-hour schedule (cron).

This should ensure we always have the latest tag in homebrew.

@hpidcock might be worth a look?

## QA steps

Unsure on this one, try it and see?

### DIFF
--- a/.github/workflows/update-brew-formulae.yml
+++ b/.github/workflows/update-brew-formulae.yml
@@ -1,0 +1,16 @@
+name: "Update homebrew tap"
+on:
+  release:
+    types: [published]
+  schedule:
+    - cron:  '0 */12 * * *'
+jobs:
+  update-brew-tap:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


petevg (approved):
I'm not the world's greatest expert on github actions, but this makes sense to me.

.github/workflows/update-brew-formulae.yml
```diff
@@ -0,0 +1,16 @@
+name: "Update homebrew tap"
+on:
+  release:
+    types: [published]
+  schedule:
```
hpidcock, line 5: What is the reason for the cron job? Why not just on release: published?

SimonRichardson: From reading around, it's not always triggered, so it's used to be sure.

SimonRichardson:
$$merge$$

PR #13115, opened 2021-06-28
Integration tests: Fix juju exec query

SimonRichardson:
The following ensures that we correctly handle returns from `juju exec`
where the exit code no longer informs us if we've correctly completed an
action. Instead, we have to grep stdout and ensure we're doing the right
thing.

## QA steps

```
$ (cd tests && ./main.sh controller) 
```


hpidcock (approved):

hpidcock:
!!build!!

hpidcock:
$$merge$$

hpidcock:
$$merge$$

PR #13116, opened 2021-06-28
Packaging: Update packaging dependency.

SimonRichardson:
The following updates the packaging dependency to handle udev failures
from snap installs. The effective workaround is to retry a second time.
We can take advantage of #13099 changes, which allows retries on mount
failures for the same udev failures.

Note: this should only really manifest itself for nested privileged
containers... esp. our CI and integration testing.

## QA steps

Tests pass.

```sh
$ juju bootstrap lxd test
```


hpidcock (approved):

hpidcock:
$$merge$$

PR #13118, opened 2021-06-29
Provider: Goose bump to latest lib version

SimonRichardson:
The following moves to goose/v4 from goose/v3. The changes are required
to remove the hardcoded dependency on loggo in a library. Instead, we
should use an interface to log out information.

This should allow for loggo updates to iterate without updating every
package that juju depends on.

## QA steps

Install [microstack](https://microstack.run/docs/using-juju)

```sh
$ juju bootstrap --bootstrap-series=$OS_SERIES \
  --metadata-source=~/simplestreams \
  --model-default network=test \
  --model-default external-network=external \
  --model-default use-floating-ip=true \
  microstack microstack
$ juju deploy ubuntu
```
Ensure logging works as previously.


hmlanigan (approved):
LGTM

hmlanigan:
$$merge$$

PR #13119, opened 2021-06-29
Logs: Ensure we close goroutine on cancel

SimonRichardson:
The following cleans up the StreamDebugLogs goroutine if
the callee wants to cancel the streaming of the logs.

I noticed the TODO whilst threading through a new debug-log parameter.

## QA steps

Test pass.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1644084

### DIFF
--- a/api/common/logs.go
+++ b/api/common/logs.go
@@ -4,6 +4,7 @@
 package common
 
 import (
+	"context"
 	"fmt"
 	"net/url"
 	"time"
@@ -104,14 +105,7 @@ type LogMessage struct {
 
 // StreamDebugLog requests the specified debug log records from the
 // server and returns a channel of the messages that come back.
-func StreamDebugLog(source base.StreamConnector, args DebugLogParams) (<-chan LogMessage, error) {
-	// TODO(babbageclunk): this isn't cancellable - if the caller stops
-	// reading from the channel (because it has an error, for example),
-	// the goroutine will be leaked. This is OK when used from the command
-	// line, but is a problem if it happens in jujud. Change it to accept
-	// a stop channel and use a read deadline so that the client can stop
-	// it. https://pad.lv/1644084
-
+func StreamDebugLog(ctx context.Context, source base.StreamConnector, args DebugLogParams) (<-chan LogMessage, error) {
 	// Prepare URL query attributes.
 	attrs := args.URLQuery()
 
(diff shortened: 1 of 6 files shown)


api/common/logs.go
```diff
@@ -95,7 +96,7 @@ type LogMessage struct {
 
 // StreamDebugLog requests the specified debug log records from the
 // server and returns a channel of the messages that come back.
-func StreamDebugLog(source base.StreamConnector, args DebugLogParams) (<-chan LogMessage, error) {
+func StreamDebugLog(ctx context.Context, source base.StreamConnector, args DebugLogParams) (<-chan LogMessage, error) {
 	// TODO(babbageclunk): this isn't cancellable - if the caller stops
```
jameinel, line 100: shouldn't this comment be updated?

worker/migrationmaster/worker.go
```diff
@@ -582,7 +583,10 @@ func (w *Worker) transferLogs(targetInfo coremigration.TargetInfo, modelUUID str
 
 	throwWrench := latestLogTime == utcZero && wrench.IsActive("migrationmaster", "die-after-500-log-messages")
 
-	logSource, err := w.config.Facade.StreamModelLog(latestLogTime)
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
+	logSource, err := w.config.Facade.StreamModelLog(ctx, latestLogTime)
```
jameinel, line 589: So the net sum of all these changes is just this, right? That if the model migration worker exits before finishing streaming, it will kill the associated goroutine.
Is there a reason to do this around contexts rather than Tombs (which predated context, which is why it is everywhere). I don't have a problem with context, just wondering if we would already have tomb objects around that we weren't properly utilizing.

jameinel (commented):
I do wish that there was a reasonable way to test this, but there probably isn't.
But we certainly need to update the now out of date comment from Christian.

api/common/logs.go
SimonRichardson: Yes.

worker/migrationmaster/worker.go
SimonRichardson: I find they serve slightly different purposes. I tend to think that `context.Context` gives a better API for passing them, where tombs feel like they're internal implementations.

manadart (approved):
Looks good.

SimonRichardson:
$$merge$$

PR #13121, opened 2021-06-29
Add test by cloud to bash integration tests

hmlanigan:
Add test by cloud to bash integration tests.  Allows for testing using clouds which have to be added, such as vSphere and OpenStack.  Renamed the "microk8s" provider to "k8s", as it's a cloud name.  Only specifying a provider will work if the provider and cloud name are the same.  Otherwise, use the cloud name to run the tests.

Caveats

- vSphere is setup to configure the controller to use pre-made vm templates as seen in the Boston vSphere only.
- OpenStack clouds must have images configured such that it is not necessary to provide image data at bootstrap time on the cli.

## QA steps

```console
# setup to bootstrap the Boston vsphere using the juju-ci folder and credentials.
(cd tests; ./main.sh -c <vsphere>  deploy deploy-charms)

# setup to bootstrap an openstack.
(cd tests; ./main.sh -c <openstack>  deploy deploy-charms)

# the k8s only test should still run by specifying a cloud to use:
(cd tests; ./main.sh -c microk8s  caasadmission)

# should not run, not a k8s provider
(cd tests; ./main.sh -p aws  caasadmission)
```

### DIFF
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -1,3 +1,4 @@
+#!/bin/bash -e
 # juju_version will return only the version and not the architecture/substrait
 # of the juju version.
 # This will use any juju on $PATH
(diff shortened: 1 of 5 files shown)


tests/includes/juju.sh
```diff
@@ -188,15 +218,31 @@ juju_bootstrap() {
 		;;
 	esac
 
+	image_metadata=""
+
+	if [[ ${BOOTSTRAP_PROVIDER} == "vsphere" ]]; then
+		echo "====> Creating image simplestream metadata for juju ($(green "${version}:${cloud}"))"
+
+	  image_streams_dir=/tmp/image-streams
```
SimonRichardson, line 226: I'd rather keep everything in `${TEST_DIR}`, so this would become `"${TEST_DIR}"/image-streams`. We just need to ensure that we make the directory first.

tests/main.sh
```diff
@@ -172,10 +176,27 @@ while getopts "hH?vAs:a:x:rl:p:S:" opt; do
...
+	    exit 1
+	  fi
+	  CLOUD="${OPTARG}"
+	  REGION=$(juju clouds --client --format=json | jq -r ".[\"${CLOUD}\"] | .regions | keys[0]")
+	  export BOOTSTRAP_REGION="${REGION}"
+	  export BOOTSTRAP_CLOUD="${CLOUD}"
+	  ;;
+	R)
+	  export BOOTSTRAP_REGION="${OPTARG}"
+	  ;;
```
SimonRichardson, line 199: This is great, but can we fix the indentation, use `shfmt -l -s -w tests/**/*.sh` from the root of juju.

SimonRichardson (changes requested):
Some nits, but otherwise great changes.

tests/main.sh
hmlanigan: My IDE is not cooperating, I was hoping there was a fmt command we could use!

SimonRichardson (approved):

hmlanigan:
$$merge$$

PR #13122, opened 2021-06-29
Add RootDiskSource as unsupportedConstraints, match the docs.

hmlanigan:
Juju constraints documentation indicates that root disk sources is unsupported for many of the providers, ensure that the juju constraint validator for each agrees.  Added root-disk-source to unsupported list where missing.

https://discourse.charmhub.io/t/juju-constraints/1160

### DIFF
--- a/caas/kubernetes/provider/constraints.go
+++ b/caas/kubernetes/provider/constraints.go
@@ -15,6 +15,7 @@ var unsupportedConstraints = []string{
 	constraints.InstanceType,
 	constraints.Spaces,
 	constraints.AllocatePublicIP,
+	constraints.RootDiskSource,
 }
 
 // ConstraintsValidator returns a Validator value which is used to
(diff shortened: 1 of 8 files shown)


hpidcock (commented):
RootDiskSource is used by the common provider code for selecting a root disk storage pool.

So I think this change doesn't make sense, but I might be wrong.

wallyworld:
As per Harry's comment, all providers use the root-disk-source constraint to look up a storage pool used to set the `RootDisk` attribute of `StartInstanceParams`. Not all providers honour this attribute - currently only Azure and EC2 (in the upcoming 2.9.7 release). There's doc for Azure already but EC2 doc needs doing also to coincide with the next release.

hmlanigan:
@wallyworld, @hpidcock please update the constraints doc, https://discourse.charmhub.io/t/juju-constraints/1160, and I'll close this PR.

hpidcock:
> @wallyworld, @hpidcock please update the constraints doc, https://discourse.charmhub.io/t/juju-constraints/1160, and I'll close this PR.

I've had a crack at rectifying this in that document, should be clearer now.

PR #13123, opened 2021-06-29
Update Pebble to include panic fixes in layer parsing code

benhoyt:
This pulls in https://github.com/canonical/pebble/pull/51:

* Fix panic when command is not set in combined layer
* Fix panic when service object/dict is null
* Also disallow use of empty string as service name

## QA steps

Make a layer with any of the above issues in it and call `add_layer` from a charm (see steps at https://github.com/canonical/pebble/issues/50) and it will now fail with an error message instead of panicking.


hpidcock (approved):
:100:

hpidcock:
$$merge$$

PR #13131, opened 2021-07-01
Merge 2.9

hmlanigan:
Changes:

- #13117
- #13127 
- #13121 

Conflicts:

- version/version.go
- snap/snapcraft.yaml
- scripts/win-installer/setup.iss
- apiserver/facades/agent/uniter/uniter_test.go
- apiserver/facades/agent/uniter/uniter.go
- apiserver/facades/agent/provisioner/mocks/package_mock.go

New commit to update Spaces signature method in the equinix provider, which does not exist in 2.9.

### DIFF
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -1,3 +1,4 @@
+#!/bin/bash -e
 # juju_version will return only the version and not the architecture/substrait
 # of the juju version.
 # This will use any juju on $PATH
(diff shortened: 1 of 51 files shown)


tests/includes/juju.sh
```diff
@@ -36,44 +37,41 @@ ensure() {
...
 
-		provider="${manual_name}"
-		;;
-	"microk8s")
-		provider="microk8s"
+		cloud="${manual_name}"
 		;;
 	*)
-		echo "Unexpected bootstrap provider (${BOOTSTRAP_PROVIDER})."
+		echo "Unexpected bootstrap cloud (${BOOTSTRAP_PROVIDER})."
```
jameinel, line 74: This seems confused that we would set unknown cloud but the  variable is PROVIDER.

tests/includes/juju.sh
```diff
@@ -188,15 +218,31 @@ juju_bootstrap() {
...
 	# does not run, goes directly to cleanup.
 	#shellcheck disable=SC2086
 	juju bootstrap ${series} \
 		--build-agent=${BUILD_AGENT} \
-		"${provider}" "${name}" -d "${model}" "$@" 2>&1 | OUTPUT "${output}"
+		"${cloud}" "${name}" -d "${model}" ${image_metadata} "$@" 2>&1 | OUTPUT "${output}"
 
 	echo "${name}" >>"${TEST_DIR}/jujus"
+
+	if [[ ${BOOTSTRAP_PROVIDER} == "vsphere" ]]; then
```
jameinel, line 241: It would be nice if we had a more generic section for this something like
```
post_bootstrap_initialization() {
switch ${BOOTSTRAP_PROVIDER} {
}
```

I say that because there is a particular type of operation we are trying to do, and it would be good to capture that rather than slowly accreting more and more if blocks.

jameinel (approved):

hmlanigan:
Added a card to address the comments next week, after holiday

hmlanigan:
$$merge$$

PR #13132, opened 2021-07-02
Action: Task exit code

SimonRichardson:
The following propagates the task result exit code up to the CLI so that
a user calling `juju exec` will know that an error occurred in one or
more tasks. Prior to this commit, users would have to inspect every
result which is a bit disappointing for bash style scripting against the
CLI. With this change it should be possible to do the following:

    $ juju exec --all 'exit 1'
    $ echo $? # should see 1

This makes juju a lot more powerful to sys or ops users.

The code is simple, inspect the result for an exit code. If it exists
propagate it to the CLI and return an error stating the fact that a task
failed.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju exec -m controller 'exit 1'
```


cmd/juju/action/common.go
```diff
@@ -31,17 +31,21 @@ import (
...
-const leaderSnippet = "(" + names.ApplicationSnippet + ")/leader"
+const (
+	// leaderSnippet is a regular expression for unit ID-like syntax that is used
+	// to indicate the current leader for an application.
+	leaderSnippet = "(" + names.ApplicationSnippet + ")/leader"
+)
 
-var validLeader = regexp.MustCompile("^" + leaderSnippet + "$")
+var (
+	validLeader = regexp.MustCompile("^" + leaderSnippet + "$")
```
achilleasa, line 41: I don't see any other use of `leaderSnippet` in the file. Why not just inline it when calling `MustCompile`?

cmd/juju/action/common.go
```diff
@@ -31,17 +31,21 @@ import (
...
 
-// nameRule describes the name format of an action or keyName must match to be valid.
-var nameRule = charm.GetActionNameRule()
+	// nameRule describes the name format of an action or keyName must match to be valid.
+	nameRule = charm.GetActionNameRule()
 
-//resultPollTime is how often to poll the backend for results.
-var resultPollTime = 2 * time.Second
+	//resultPollTime is how often to poll the backend for results.
```
achilleasa, line 46: minor nit
```suggestion
	// resultPollTime is how often to poll the backend for results.
```

cmd/juju/action/common.go
```diff
@@ -146,15 +150,25 @@ func (c *runCommandBase) processOperationResults(ctx *cmd.Context, results *acti
...
-	return c.waitForTasks(ctx, tasks, info)
+	exitCode, err := c.waitForTasks(ctx, tasks, info)
+	if err != nil {
+		return errors.Trace(err)
+	} else if exitCode > 0 {
+		var plural string
+		if numTasks > 1 {
+			plural = "s"
+		}
+		return errors.Errorf("task%s failed with exit code: %d", plural, exitCode)
```
achilleasa, line 161: Does a non-zero exit code mean that some tasks failed or that all tasks failed? If it's the former we should probably rephrase the error for the multiple task case to something like "some or all tasks failed with exit code: %d"

cmd/juju/action/common.go
```diff
@@ -197,18 +212,33 @@ func (c *runCommandBase) waitForTasks(ctx *cmd.Context, tasks []enqueuedAction,
...
+		// If any of the actions have a error exit code, then inform the user
+		// that their exec failed.
+		if resultExitCode < 1 {
+			continue
+		}
+
+		// If the number of tasks is just 1, show the underlying exit code. We
+		// can't do this for multiple tasks as we don't know what the
+		// appropriate code to show (highest or lowest)?
+		if len(tasks) == 1 {
```
achilleasa, line 234: I believe we should be explicit with our errors and indicate which tasks failed and with what exit code. It will make the error much more verbose but at least it tells us what failed so we can retry it.

cmd/juju/action/common.go
SimonRichardson: No idea, I was just cleaning up lint errors.

cmd/juju/action/common.go
SimonRichardson: `tasks` vs `task` should cover this.

cmd/juju/action/common.go
SimonRichardson: We don't know essentially and exit code is optional.

cmd/juju/action/common.go
achilleasa: As written, the message does not make it clear if one task failed or whether all failed. So the user runs 4 tasks and one fails, they might assume that all failed with exit X.

cmd/juju/action/common.go
achilleasa: Don't we know (per task) since we iterate the result list? Do you happen to know if the optional exit code is by design and if so what was the use case that had that as a requirement?

SimonRichardson: In fact, the stderr has ALL this information:

```
juju exec -m controller --all 'cat /etc/systemd/system/juju-db.service'
Running operation 29 with 3 tasks
  - task 30 on machine-0
  - task 31 on machine-1
  - task 32 on machine-2

Waiting for task 30...
Waiting for task 31...
Waiting for task 32...
cat: /etc/systemd/system/juju-db.service: No such file or directory

ERROR tasks failed with exit code: 1
```

cmd/juju/action/common.go
```diff
@@ -197,18 +220,26 @@ func (c *runCommandBase) waitForTasks(ctx *cmd.Context, tasks []enqueuedAction,
...
-		d := formatActionResult(result.task, actionResult, c.utc)
-		d["id"] = result.task // Action ID is required in case we timed out.
-		info[result.receiverId()] = d
+		resultData, resultExitCode := formatActionResult(result.task, actionResult, c.utc)
+		resultData["id"] = result.task // Action ID is required in case we timed out.
+		info[result.receiverId()] = resultData
+
+		// If any of the actions have a error exit code, then inform the user
+		// that their exec failed.
+		if resultExitCode < 1 {
```
achilleasa, line 235: Why not `== 0` ?

cmd/juju/action/common.go
```diff
@@ -146,15 +149,34 @@ func (c *runCommandBase) processOperationResults(ctx *cmd.Context, results *acti
...
+	if err != nil {
+		return errors.Trace(err)
+	} else if len(failed) > 0 {
+		var plural string
+		if len(failed) > 1 {
+			plural = "s"
+		}
+		list := make([]string, 0, len(failed))
+		for k, v := range failed {
+			list = append(list, fmt.Sprintf(" - id %q with return code %d", k, v))
```
achilleasa, line 162: s/return/exit?

Also, it could be helpful to include the machine in the error, e.g `-id %q on machine %q with exit code %d`

cmd/juju/action/exec_test.go
```diff
@@ -626,27 +626,37 @@ Running operation 1 with 1 task
   - task 1 on machine-0
 
 Waiting for task 1...
+`[1:]
+
+	errStr := `
+the following task failed:
+ - id "1" with return code 42
+
```
achilleasa, line 634: For completeness, can you also add an error with multiple tasks?

cmd/juju/action/common.go
SimonRichardson: It isn't always a machine, so the error message get's very complicated.

cmd/juju/action/common.go
SimonRichardson: Result exit code is optional, so we can return `-1`.

achilleasa: Ah ok then; thought it would just be zero if not tracked.

cmd/juju/action/common.go
achilleasa: What else can it be? The task ID by itself isn't really useful as the user would presumably have to look it up somehow, right?

SimonRichardson: Anything that uses names, but in reality it's probably machine or unit.

wallyworld: The receiver of an action (ie where a task can run) is either a machine or unit.
The task ID is useful as it can be used with show-task.
Perhaps the error message could be enhanced to hint that `show-task` can be used to inspect the details on any failed task IDs

wallyworld (approved):

SimonRichardson:
$$merge$$

PR #13133, opened 2021-07-02
Kubernetes: Wait for pod ready timeout

SimonRichardson:
The following moves the timeout deadline from 1 minute to 10 minutes. If
a pod hasn't become ready in that time, I think it's safe to say it
won't!

The change is simple, just add a longer duration.

## QA steps

See: https://discourse.charmhub.io/t/error-getting-started-with-juju-microk8s/3855/3

## Bug reference

https://bugs.launchpad.net/juju/+bug/1934494

### DIFF
--- a/caas/kubernetes/tunnel.go
+++ b/caas/kubernetes/tunnel.go
@@ -29,6 +29,11 @@ import (
 	"github.com/juju/juju/caas/kubernetes/pod"
 )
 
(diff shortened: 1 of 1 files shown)


achilleasa (approved):

SimonRichardson:
$$merge$$

PR #13135, opened 2021-07-06
Fix parsing of custom k8s image repos with a port

wallyworld:
A small fix for an issue found by QA folks when setting up a custom docker image repo. If the repo URL configured using `caas-image-repo` has a port, the parsing fails.

## QA steps

There's a unit test which ensures the expected behaviour. There's no 2.9.8 images but do this

` juju bootstrap microk8s test --config caas-image-repo=docker.io:5000/jujusolutions`

and it won't find the image but you can check the correct image URL:

```
microk8s.kubectl -n controller-test get pod/controller-0 -o json | jq .spec.containers[0].image
"docker.io:5000/jujusolutions/jujud-operator:2.9.8"
```

Also bootstrap a standard microk8s controller to regression test default behaviour.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1934707

### DIFF
--- a/cloudconfig/podcfg/image.go
+++ b/cloudconfig/podcfg/image.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/docker/distribution/reference"
 	"github.com/juju/charm/v8"
 	"github.com/juju/errors"
 	"github.com/juju/version/v2"
@@ -15,14 +16,13 @@ import (
 )
 
 const (
-	JujudOCINamespace      = "jujusolutions"
-	JujudOCIName           = "jujud-operator"
-	JujuContainerAgentName = "containeragent"
-	JujudbOCIName          = "juju-db"
+	JujudOCINamespace = "jujusolutions"
+	JujudOCIName      = "jujud-operator"
+	JujudbOCIName     = "juju-db"
 )
 
 // GetControllerImagePath returns oci image path of jujud for a controller.
-func (cfg *ControllerPodConfig) GetControllerImagePath() string {
+func (cfg *ControllerPodConfig) GetControllerImagePath() (string, error) {
 	return GetJujuOCIImagePath(cfg.Controller.Config, cfg.JujuVersion, cfg.OfficialBuild)
 }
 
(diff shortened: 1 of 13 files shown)


wallyworld:
[Juju interprets caas-image-repo containing port number incorrectly](https://trello.com/c/iUACsVYi/1660-juju-interprets-caas-image-repo-containing-port-number-incorrectly)

cloudconfig/podcfg/image.go
```diff
@@ -62,21 +61,25 @@ func RebuildOldOperatorImagePath(imagePath string, ver version.Number) string {
 	return tagImagePath(imagePath, ver)
 }
 
-func tagImagePath(path string, ver version.Number) string {
+func tagImagePath(fullPath string, ver version.Number) string {
+	fullPathParts := strings.Split(fullPath, "/")
```
hpidcock, line 65: Lets just remove all this string bashing and instead use github.com/distribution/distribution/reference to Parse image references.

hpidcock (changes requested):

cloudconfig/podcfg/image.go
wallyworld: done

cloudconfig/podcfg/image.go
```diff
@@ -42,44 +42,40 @@ func IsJujuOCIImage(imagePath string) bool {
...
-	splittedPath := strings.Split(path, ":")
-	path = splittedPath[0]
-	if len(splittedPath) > 1 {
-		verString = splittedPath[1]
+func tagImagePath(fullPath string, ver version.Number) (string, error) {
+	ref, err := reference.Parse(fullPath)
+	if err != nil {
+		return "", errors.Trace(err)
 	}
+	imageNamed := ref.(reference.Named)
```
hpidcock, line 70: this could panic, but probably ok if it did. (e.g. digest only path isn't a reference.Named)

hpidcock (approved):

cloudconfig/podcfg/image.go
wallyworld: i thought about returning an error here but checked the call sites and we are good. maybe i should return an error anyway just as a safety check

wallyworld:
$$merge$$

PR #13137, opened 2021-07-06
Cache: Coherence

SimonRichardson:
The cache would fail to notify if you changed from a value and back to
the original value (flip flopping). This is because the underlying hash value 
was never written. 

This has two consequences:

 1. The cache could never emit a change of the original value
 2. Because we never cached the new value, if the new value was the same
 as the old one, we would still emit the event.

The fix is easy, just save the hash after the notify and we can have
better cache coherency.

## QA steps

If you change the model-config logging-config to a new value and back to the 
old value, it should now correctly go back to the old value. Flip-flopping of values 
in the logging-config exposes this relatively easily.

```sh
$ juju bootstrap lxd test
$ juju model-config logging-config="<root>=WARN"
$ juju model-config logging-config="<root>=DEBUG"
```

View `$ juju debug-log -m controller` to ensure that the flip-flop happens.

### DIFF
--- a/core/cache/watcher.go
+++ b/core/cache/watcher.go
@@ -156,6 +156,7 @@ func (w *ConfigWatcher) configChanged(topic string, value interface{}) {
 		return
 	}
 	w.notify()
+	w.hash = hash
 }
 
 // StringsWatcher will return what has changed.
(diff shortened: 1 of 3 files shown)


core/cache/watcher.go
```diff
@@ -156,6 +156,7 @@ func (w *ConfigWatcher) configChanged(topic string, value interface{}) {
 		return
 	}
 	w.notify()
+	w.hash = hash
```
SimonRichardson, line 159: This is the actual fix

jameinel (approved):

core/cache/watcher.go
jameinel: Did you play around with whether this should happen before or after the notify? (I think after is correct, but I wanted to make sure we thought about it.)

jameinel (approved):

SimonRichardson:
$$merge$$

core/cache/watcher.go
SimonRichardson: Yeah I did, but the conclusion I came to was the same.

PR #13146, opened 2021-07-08
address Merge comments

hmlanigan:
Address 2 comments made in #13131. Fix output and add functions to handle config required by providers pre and post bootstrap.

## QA steps

Same as #13121.  There should be no change to current behavior.

### DIFF
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -219,10 +219,7 @@ juju_bootstrap() {
 
 	pre_bootstrap
 
-	# When double quotes are added to ${series}, the juju bootstrap
-	# command looks correct, and works outside of the harness, but
-	# does not run, goes directly to cleanup.
-	command="juju bootstrap ${series} --build-agent=${BUILD_AGENT} ${cloud} ${name} -d ${model} ${BOOTSTRAP_ADD_ON}"
+	command="juju bootstrap ${series} --build-agent=${BUILD_AGENT} ${cloud} ${name} -d ${model} ${BOOTSTRAP_ADDITIONAL_ARGS}"
 	# keep $@ here, otherwise hit SC2124
 	${command} "$@" 2>&1 | OUTPUT "${output}"
 	echo "${name}" >>"${TEST_DIR}/jujus"
@@ -233,13 +230,15 @@ juju_bootstrap() {
 # pre_bootstrap contains setup required before bootstrap specific to providers
 # and shouldn't be used by any of the tests directly.
 pre_bootstrap() {
+	# ensure BOOTSTRAP_ADDITIONAL_ARGS is defined, even if not necessary.
+	export BOOTSTRAP_ADDITIONAL_ARGS=""
 	case "${BOOTSTRAP_PROVIDER:-}" in
 	"vsphere")
 		echo "====> Creating image simplestream metadata for juju ($(green "${version}:${cloud}"))"
 
 		image_streams_dir=${TEST_DIR}/image-streams
 		setup_vsphere_simplestreams "${image_streams_dir}" "${BOOTSTRAP_SERIES}"
-		export BOOTSTRAP_ADD_ON="--metadata-source ${image_streams_dir}"
+		export BOOTSTRAP_ADDITIONAL_ARGS="--metadata-source ${image_streams_dir}"
 		;;
 	esac
 }
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -71,7 +71,7 @@ bootstrap() {
 		cloud="${manual_name}"
 		;;
 	*)
-		echo "Unexpected bootstrap cloud (${BOOTSTRAP_PROVIDER})."
+		echo "Unexpected bootstrap provider (${BOOTSTRAP_PROVIDER})."
 		exit 1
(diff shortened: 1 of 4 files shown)


tests/includes/juju.sh
```diff
@@ -218,30 +217,53 @@ juju_bootstrap() {
 		;;
 	esac
 
-	image_metadata=""
+	pre_bootstrap
```
hmlanigan, line 220: Here is the primary change for review, and the rest of this file.

tests/includes/juju.sh
```diff
@@ -218,30 +217,53 @@ juju_bootstrap() {
...
+
+	# When double quotes are added to ${series}, the juju bootstrap
+	# command looks correct, and works outside of the harness, but
+	# does not run, goes directly to cleanup.
+	command="juju bootstrap ${series} --build-agent=${BUILD_AGENT} ${cloud} ${name} -d ${model} ${BOOTSTRAP_ADD_ON}"
```
jameinel, line 225: As this is bash, I think what you would want is:

command="juju bootstrap \"${series}\" ..."
otherwise the " ends up exiting the string, and you can't do:
command='juju bootstrap "${series}"...'
because then none of the variables are expanded.
I'm a little surprised we're naming the controller just based on the series and not something like ${series}-$RANDOM

tests/includes/juju.sh
```diff
@@ -218,30 +217,53 @@ juju_bootstrap() {
...
-	fi
-
-	# When double quotes are added to ${series}, the juju bootstrap
-	# command looks correct, and works outside of the harness, but
-	# does not run, goes directly to cleanup.
-	#shellcheck disable=SC2086
-	juju bootstrap ${series} \
-		--build-agent=${BUILD_AGENT} \
-		"${cloud}" "${name}" -d "${model}" ${image_metadata} "$@" 2>&1 | OUTPUT "${output}"
+		export BOOTSTRAP_ADD_ON="--metadata-source ${image_streams_dir}"
```
jameinel, line 242: BOOTSTRAP_ADDITIONAL_ARGS ?
I'm not sure that it is better.

tests/includes/juju.sh
```diff
@@ -218,30 +217,53 @@ juju_bootstrap() {
...
+		;;
+	esac
+}
 
-	echo "${name}" >>"${TEST_DIR}/jujus"
+# post_bootstrap contains actions required after bootstrap specific to providers
+# and shouldn't be used by any of the tests directly.  Calls post_add_model
+# models are added during bootstrap.
+post_bootstrap() {
```
jameinel, line 250: We might want to put post_bootstrap into a bash 'trap' so that it always runs whether or not bootstrap succeeds. Then again, maybe that happens too late in the happy case?

tests/includes/juju.sh
```diff
@@ -218,30 +217,53 @@ juju_bootstrap() {
...
+	case "${BOOTSTRAP_PROVIDER:-}" in
+	"vsphere")
+		rm -r "${TEST_DIR}"/image-streams
+		;;
+	esac
+	post_add_model
```
jameinel, line 256: It seems a little odd that post_add_model is part of post_bootstrap, but i guess you do need to do that step here. I'm just wondering if it should be explicitly called above as:
post_bootstrap
post_add_model

or whether it is cleaner to embed it as you have done.

tests/includes/juju.sh
jameinel: Also, do we run these scripts with bash's 'set -u', which means we need to make sure BOOTSTRAP_ADD_ON has some value (even just the empty string), or it will fail on non-vsphere.

tests/suites/caasadmission/task.sh
```diff
@@ -6,20 +6,20 @@ test_caasadmission() {
 
 	set_verbosity
 
-  case "${BOOTSTRAP_PROVIDER:-}" in
-    "k8s")
-      echo "==> Checking for dependencies"
-      check_dependencies petname
```
jameinel, line 12: I assume the rest of these are just spaces vs tabs? As long as we end up consistent

jameinel (approved):

tests/suites/caasadmission/task.sh
hmlanigan: yes, static analysis caught them, ran shfmt on the files.

tests/includes/juju.sh
hmlanigan: a bit better than the original.

tests/includes/juju.sh
hmlanigan: At least in this case, it's easier to debug bootstrap failing later if post_bootstrap does not run.  Something to consider if in the future if options are added.

tests/includes/juju.sh
hmlanigan: Name is the name of the controller, which ends up as ctrl-xxxxxxxx.

Ensured that BOOTSTRAP_ADDITIONAL_ARGS is always set.

Removed the comment @ 222-224 as it no longer applies.

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #13148, opened 2021-07-09
Select a random AZ instead of the first one when starting a machine

wallyworld:
Juju selects the AZ to used when starting a machine based on the AZ with the lowest machine count. If more than one AZ is suitable, Juju would always select the first one (based on alphabetical order). This has the effect of loading up the first AZ if there are multiple models/controllers in use. This PR instead selects a random AZ out of those which have the smallest machine count.

The selection of AZ is based on the premise that the AZ distribution is loaded in when the provisioner worker starts. However, the logic was flawed because the set up of the AZ distribution required that the running instances be loaded first. And this was only done later when a watcher event is received. So the initial AZ map was always empty after an agent restart, messing up the distribution. So that is fixed as well.

Some unit tests have also been tweaked to use the new worker behaviour to run the test - previously they were triggering logic based on a retry signal, but they should be using the machine watcher notification.

## QA steps

The issue is easily seen - bootstrap to AWS and add a machine; the machine would always start in AZ "a". And the next in "b" etc. And after an agent restart, "a" would be used for the next machine.
Now, `juju add-machine` always selects a random AZ out of those that are eligible (as seen in juju status).

## Bug reference

https://bugs.launchpad.net/bugs/1933690

### DIFF
--- a/worker/provisioner/provisioner_task.go
+++ b/worker/provisioner/provisioner_task.go
@@ -6,6 +6,7 @@ package provisioner
 import (
 	stdcontext "context"
 	"fmt"
+	"math/rand"
 	"sort"
 	"strings"
 	"sync"
@@ -163,13 +164,6 @@ func (task *provisionerTask) Wait() error {
 
 func (task *provisionerTask) loop() error {
 
-	// Get existing machine distributions.
-	err := task.populateAvailabilityZoneMachines()
-	// Not all providers implement ZonedEnviron
-	if err != nil && !errors.IsNotImplemented(err) {
-		return errors.Trace(err)
-	}
-
 	// Don't allow the harvesting mode to change until we have read at
 	// least one set of changes, which will populate the task.machines
 	// map. Otherwise we will potentially see all legitimate instances
@@ -262,6 +256,13 @@ func (task *provisionerTask) processMachines(ids []string) error {
 	ctx := task.cloudCallCtxFunc(stdcontext.Background())
 	// Populate the tasks maps of current instances and machines.
 	if err := task.populateMachineMaps(ctx, ids); err != nil {
+		return errors.Trace(err)
+	}
+
+	// Get existing machine distributions.
+	err := task.populateAvailabilityZoneMachines()
+	// Not all providers implement ZonedEnviron
+	if err != nil && !errors.IsNotImplemented(err) {
 		return errors.Trace(err)
 	}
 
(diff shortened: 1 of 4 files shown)


worker/provisioner/provisioner_task.go
```diff
@@ -853,31 +854,58 @@ func (task *provisionerTask) machineAvailabilityZoneDistribution(
...
-		}
-		if zoneMachines.ExcludedMachineIds.Contains(machineId) {
-			task.logger.Debugf("machine %s does not match az %s: excluded machine id",
-				machineId, zoneMachines.ZoneName)
-			continue
+	// Make a map of zone machines keyed on count.
+	zoneMap := make(map[int][]*AvailabilityZoneMachine)
+	for _, zm := range zoneMachines {
+		machineCount := zm.MachineIds.Size()
+		zmList := append(zoneMap[machineCount], zm)
```
hpidcock, line 867: Could this not just be?
`zoneMap[machineCount] = append(zoneMap[machineCount], zm)`

wallyworld:
[AZ distribution regression](https://trello.com/c/Q9XqZ9lS/1662-az-distribution-regression)

worker/provisioner/provisioner_task.go
```diff
@@ -853,31 +854,58 @@ func (task *provisionerTask) machineAvailabilityZoneDistribution(
...
+	var zoneCounts []int
+	for k := range zoneMap {
+		zoneCounts = append(zoneCounts, k)
+	}
+	sort.Ints(zoneCounts)
+
+	var machineZone string
+done:
+	// Starting with the lowest count first, find a suitable AZ.
+	for _, count := range zoneCounts {
```
hpidcock, line 881: Let's try to avoid goto.
```
for _, count := range zoneCounts {
	zmList := zoneMap[count]
	for len(zmList) > 0 {
		// Pick a random AZ to try.
		index := rand.Intn(len(zmList))
		zoneMachines := zmList[index]
		if !zoneMachines.MatchesConstraints(cons) {
			task.logger.Debugf("machine %s does not match az %s: constraints do not match",
				machineId, zoneMachines.ZoneName)
		} else if zoneMachines.FailedMachineIds.Contains(machineId) {
			task.logger.Debugf("machine %s does not match az %s: excluded in failed machine ids",
				machineId, zoneMachines.ZoneName)
		} else if zoneMachines.ExcludedMachineIds.Contains(machineId) {
			task.logger.Debugf("machine %s does not match az %s: excluded machine id",
				machineId, zoneMachines.ZoneName)
		} else {
			// Success, we're out of here.
			machineZone = zoneMachines.ZoneName
			break
		}
		// Zone not suitable so remove it from the list and try the next one.
		zmList = append(zmList[:index], zmList[index+1:]...)
	}
	if machineZone != "" {
		break
	}
}
```

hpidcock (approved):
LGTM just a few things

worker/provisioner/provisioner_task.go
```diff
@@ -853,31 +854,58 @@ func (task *provisionerTask) machineAvailabilityZoneDistribution(
...
+				task.logger.Debugf("machine %s does not match az %s: excluded machine id",
+					machineId, zoneMachines.ZoneName)
+				goto next
+			}
+			// Success, we're out of here.
+			machineZone = zoneMachines.ZoneName
+			break done
+		next:
+			// Zone not suitable so remove it from the list and try the next one.
+			zmList = append(zmList[:index], zmList[index+1:]...)
```
hpidcock, line 907: Does this work for the last item?

hpidcock: it does

hpidcock: :hamburger:

wallyworld: yup

wallyworld:
$$merge$$

worker/provisioner/provisioner_task_test.go
```diff
@@ -206,11 +209,88 @@ func (s *ProvisionerTaskSuite) TestProvisionerRetries(c *gc.C) {
 	s.instanceBroker.CheckCallNames(c, "StartInstance", "StartInstance")
 }
 
+func (s *ProvisionerTaskSuite) TestEvenZonePlacement(c *gc.C) {
```
manadart, line 212: Multiple tests including this one now fail the race jobs.

manadart: Sorted by #13155.

PR #13149, opened 2021-07-09
Loggo wire up labels.

SimonRichardson:
The following wires up the labels for the loggo library. Ensuring that we also
handle labels correctly in the debug-log command.

The change is simple yet powerful for helping tag various logs together as a
singular cohort, rather than individual models (tagging loggers). By tagging
a logger to a given tag you can get all the logs for a given tag. Although simple
in the premise, it is useful for debugging things like HTTP requests to
charmhub as an example.

The move away from global loggers (loggers created at the package level, see
below) and to dependency injected loggers has one downside; discoverability.

```go
package apiserver

var logger = loggo.GetLogger("juju.apiserver")
```

The discoverability of loggers now becomes almost impossible without deep diving 
into the code and tracing all the forked branches. With tagging, this should become
a lot simpler, we just need to ensure that logs are correctly tagged.

## QA steps

It's best to use a multi-terminal session for this (tmux)

```sh
$ juju bootstrap lxd test
$ juju find windows
```

To see that the trace doesn't come through when using `exclude-label`

```sh
$ juju debug-log -m controller --exclude-label=http
```

### DIFF
--- a/apiserver/params/internal.go
+++ b/apiserver/params/internal.go
@@ -950,6 +950,7 @@ type LogMessage struct {
 	Module    string    `json:"mod"`
 	Location  string    `json:"loc"`
 	Message   string    `json:"msg"`
+	Labels    []string  `json:"lab"`
 }
 
 // ResourceUploadResult is used to return some details about an
--- a/state/logs.go
+++ b/state/logs.go
@@ -321,6 +321,7 @@ type logDoc struct {
 	Location string        `bson:"l"` // "filename:lineno"
 	Level    int           `bson:"v"`
 	Message  string        `bson:"x"`
+	Labels   []string      `bson:"c,omitempty"` // e.g. http
 }
 
 type DbLogger struct {
@@ -368,6 +369,7 @@ func (logger *DbLogger) Log(records []LogRecord) error {
 			Location: r.Location,
 			Level:    int(r.Level),
 			Message:  r.Message,
+			Labels:   r.Labels,
 		})
 	}
 	_, err := bulk.Run()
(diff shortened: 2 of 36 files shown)


apiserver/params/internal.go
```diff
@@ -950,6 +950,7 @@ type LogMessage struct {
 	Module    string    `json:"mod"`
 	Location  string    `json:"loc"`
 	Message   string    `json:"msg"`
+	Labels    []string  `json:"lab"`
```
achilleasa, line 953: Really minor nit; maybe it would read better like `msg`; feel free to ignore
 
```suggestion
	Labels    []string  `json:"lbl"`
```

state/logs.go
```diff
@@ -760,6 +765,14 @@ func (t *logTailer) paramsToSelector(params LogTailerParams, prefix string) bson
 		sel = append(sel,
 			bson.DocElem{"m", bson.M{"$not": bson.RegEx{Pattern: makeModulePattern(params.ExcludeModule)}}})
 	}
+	if len(params.IncludeLabel) > 0 {
+		sel = append(sel,
+			bson.DocElem{"c", bson.M{"$in": params.IncludeLabel}})
```
achilleasa, line 770: Do we need some kind of index for this query (and the $nin one)?

achilleasa (approved):
LGTM

state/logs.go
SimonRichardson: I 100% bet we don't have any indexes on this stuff. I'll check, might be a follow-up PR.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

state/logs.go
SimonRichardson: https://github.com/juju/juju/pull/13160

PR #13157, opened 2021-07-12
Integration: Run integration sub test

SimonRichardson:
This allows running integration sub-tests whilst inside the integration
tests. What this provides is a useful way to do the following:

    ./main.sh deploy test_deploy_charms

This only runs the deploy charms tests and skips any of the other tests.

We want to employ this from our CI setup, where we have so many tests now that
they're getting timed out because they're taking too long. If we can shard the tests
out based on the suites and individual tests, we can better positive results from our
CI.

## QA steps

Ensure that it only runs the bundles tests.

```sh
(cd tests && ./main.sh deploy test_deploy_bundles)
```

### DIFF
--- a/tests/suites/examples/task.sh
+++ b/tests/suites/examples/task.sh
@@ -1,5 +1,5 @@
 test_examples() {
-	if [ "$(skip 'test_example')" ]; then
+	if [ "$(skip 'test_examples')" ]; then
 		echo "==> TEST SKIPPED: example tests"
 		return
 	fi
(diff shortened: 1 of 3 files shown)


tests/suites/examples/task.sh
```diff
@@ -1,5 +1,5 @@
 test_examples() {
-	if [ "$(skip 'test_example')" ]; then
+	if [ "$(skip 'test_examples')" ]; then
```
SimonRichardson, line 2: This was worded incorrectly and should be named after the function name.

manadart (approved):
Nice.

SimonRichardson:
$$merge$$

PR #13158, opened 2021-07-12
fix network health bash test

hmlanigan:
ubuntu-bionic key is 3 in current config, not 4.

Remove references to focal units which are no longer part of this test.
## QA steps

```sh
(cd tests ; ./main.sh -v -p aws network)
```


achilleasa (approved):
QA works as expected

hmlanigan:
$$merge$$

PR #13159, opened 2021-07-12
Integration: Always archive integration logs

SimonRichardson:
Sometimes cleaning up juju or cleaning up functions within the
integration tests will cause the scripts to bomb out. We can't do
anything when this happens, so instead we should archive the logs first
before attempting and then archive again just before completing. That
way we can ensure that we have at least some logs.

## QA steps

Run the examples integration tests and archive the output to a tar.gz file.

```sh
(cd tests && ./main.sh -l test -v -a /tmp/output.tar.gz -x output.txt examples)
```

### DIFF
--- a/tests/main.sh
+++ b/tests/main.sh
@@ -252,6 +252,8 @@ cleanup() {
 
 	echo "==> Cleaning up"
 
+	archive_logs "partial"
+
 	cleanup_jujus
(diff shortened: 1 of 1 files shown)


manadart (approved):

SimonRichardson:
$$merge$$

PR #13160, opened 2021-07-13
State logs: index label fields

SimonRichardson:
This follows on from https://github.com/juju/juju/pull/13149.

When using debug-logs it helps if we index the log fields correctly. The
following change ensures that we index the labels field (in this
instance it's the "c" field). As the field is optional, we now need to
create a sparse index for mongo.

See: https://docs.mongodb.com/manual/core/index-sparse/

## QA steps

It's best to use a multi-terminal session for this (tmux)

```sh
$ juju bootstrap lxd test
$ juju find windows
```

To see that the trace doesn't come through when using `exclude-label`

```sh
$ juju debug-log -m controller --exclude-label=http
```

### DIFF
--- a/state/logs.go
+++ b/state/logs.go
@@ -63,13 +63,22 @@ type ModelSessioner interface {
 	ModelUUID() string
 }
 
+// LogIndex represents a way to configure a series of log indexes.
+type LogIndex struct {
+	Key    []string
+	Sparse bool
+}
+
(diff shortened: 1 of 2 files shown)


achilleasa (approved):
LGTM

SimonRichardson:
$$merge$$

PR #13163, opened 2021-07-13
CLI Deploy: Fix panic

SimonRichardson:
The following fixes a panic when deploying a charm that isn't found.
This was caused by a missing test that tested for an error in the API
result. The fix is simple, we just need to ensure that we correctly
handle the error.

## QA steps

The following shouldn't panic.

```sh
$ juju bootstrap lxd test
$ juju deploy fail
```


hmlanigan (approved):
No panics hit, worked as expected.

$ juju deploy fail
ERROR resolving with preferred channel: resolving error: The Charm with the given name was not found in the Store.

nit: Should 'Store' be charmhub?

SimonRichardson:
> $ juju deploy fail
> ERROR resolving with preferred channel: resolving error: The Charm with the given name was not found in the Store.
> 
> nit: Should 'Store' be charmhub?

I'll come back to this, I wanted to get rid of the panic first.

hmlanigan:
Just reported as https://bugs.launchpad.net/juju/+bug/1936003.  I've updated the bug.

PR #13164, opened 2021-07-13
Fixes unit logging appearing on machine log

tlm:
This change makes it so Unit logging comes from the unit logger when
viewing from the Juju controller.

## Checklist

## QA steps

See bug for repo steps and https://github.com/canonical/operator/issues/557

## Bug reference

https://bugs.launchpad.net/juju/+bug/1933548


worker/uniter/runner/jujuc/context.go
```diff
@@ -30,6 +31,10 @@ type Context interface {
 	actionHookContext
 	unitCharmStateContext
 	workloadHookContext
+
+	// GetLogger returns a juju loggo Logger for the supplied module that is
+	// correctly wired up for the given context
+	GetLogger(module string) loggo.Logger
```
wallyworld, line 37: There's already a GetLogger() on HookContext. I think that does the right thing. So we should just use that and not have to override it here. Or is the embedded HookContext one not suitable?

wallyworld (approved):

tlm:
$$merge$$

PR #13165, opened 2021-07-13
deploy by revision: update charmhub client to use new related api functionality

hmlanigan:
This is setup for implementation of deploy by revision feature work.

Allow for install by revision to not require a base, and find the base data in the response.  Also allow for resources to be specified by revision when installing by revision.

## QA steps

```console
# run live unit tests for charmhub client

# there should be no change to current behavior
$ juju deploy juju-qa-test --channel 2.0/edge
Located charm "juju-qa-test" in charm-hub, revision 9
Deploying "juju-qa-test" from charm-hub charm "juju-qa-test", revision 9 in channel 2.0/edge
$ juju refresh juju-qa-test --channel latest/edge
Added charm-hub charm "juju-qa-test", revision 15 in channel edge, to the model
Leaving endpoint in "alpha": info
```


hmlanigan:
!!build!!

SimonRichardson (commented):
There is something strange going on with the API. I think installing by a revision should never send a base. As they ignore bases with revisions it feels like we're violating reasonable assumptions where the API doesn't give us back arm64 if we ask for it.

The following should error.

```
curl -XPOST -s https://api.staging.charmhub.io/v2/charms/refresh -H 'Content-type: application/json' -d '{"context": [], "actions": [{"name": "ubuntu", "revision": 1, "base": {"architecture": "arm64", "name": "NA", "channel": "NA"},  "action": "install", "instance-key": "a-test"}], "fields": ["bases","revision","id","download","version","resources","publisher","summary","type"]}' | jq
```

hmlanigan:
Two bugs filed on the api
https://bugs.launchpad.net/snapstore-server/+bug/1936418 - Charmhub: Stricter API validation for base in refresh request 
https://bugs.launchpad.net/snapstore-server/+bug/1936434 - Charms: Refresh API always returns revision info even for revisions that won't be downloadable

SimonRichardson (approved):

hmlanigan:
$$merge$$

PR #13166, opened 2021-07-14
Add support for bespoke service principals in Azure credentials

wallyworld:
Juju on Azure needs to support credentials built on custom service principals. Such credentials will have a different application id and application object id to the hard coded defaults. We add a new optional "application-object-id" attribute to Juju credentials. It is optional because older credentials don't have it. When Juju needs the application object id, and the default service principal is in use, it simply uses the official juju app hard coded default. If the default service principal is not used, the application object id is read from the credential attributes.

When juju add-credential is run, we now grab the application object id and store it in the credential. This will usually be just the same as the hard coded default. Older credentials won't have it so there is code to insert the correct value later.

As a small refactor, the initialisation of the environ was split up into SetCloudSpec() and SetConfig() to conform with current best practise.

NB older 2.8 clients will no longer be able to talk to Azure controllers since they do not know about the new credential attribute.

## QA steps

The steps to verify that the new way of looking up a non-default application object id are quite involved - see the bug.
For QA, we can test the usual behaviour, by bootstrapping with two different credentials:

```
    cred1:
      auth-type: service-principal-secret
      application-id: 60a04dc9-1857-425f-8076-5ba81ca53d66
      application-password: xxxxx
      application-object-id: 8b744cea-179d-4a73-9dff-20d52126030a
      subscription-id: 2eebf55a-1e02-45d8-a299-02aed8aea00b
```
```
    cred2:
      auth-type: service-principal-secret
      application-id: 60a04dc9-1857-425f-8076-5ba81ca53d66
      application-password: xxxxx
      subscription-id: 2eebf55a-1e02-45d8-a299-02aed8aea00b
```

One has the application-object-id, the other doesn't. In both cases, we can set up a machine with disk encryption.

```
juju create-storage-pool des azure encrypted=true vault-name-prefix=juju
juju add-machine --constraints="root-disk-source=des"
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1929436

### DIFF
--- a/provider/azure/config_test.go
+++ b/provider/azure/config_test.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	fakeApplicationId     = "00000000-0000-0000-0000-000000000000"
+	fakeApplicationId     = "60a04dc9-1857-425f-8076-5ba81ca53d66"
 	fakeSubscriptionId    = "22222222-2222-2222-2222-222222222222"
 	fakeStorageAccountKey = "quay"
 )
--- a/provider/azure/credentials.go
+++ b/provider/azure/credentials.go
@@ -19,9 +19,10 @@ import (
 )
 
 const (
-	credAttrAppId          = "application-id"
-	credAttrSubscriptionId = "subscription-id"
-	credAttrAppPassword    = "application-password"
+	credAttrAppId               = "application-id"
+	credAttrApplicationObjectId = "application-object-id"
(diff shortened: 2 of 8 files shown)


wallyworld:
[azure custom service principal](https://trello.com/c/ePs9NG4E/1668-azure-custom-service-principal)

ycliuhw (approved):
LGTM, thank you!

wallyworld:
$$merge$$

PR #13172, opened 2021-07-16
Fix azure instance queries to account for old completed deployments

wallyworld:
Using a custom Azure resource group, you can bootstrap and then destroy the controller and re-use the group for the next bootstrap. However, Juju had a buggy method of counting machines in a resource group - it would query the completed Deployment records. Because these records are retained on the group, the next bootstrap would think there are too many machines and exit.

This PR changes how machines are queried - it still has to look for deployments because there's a window where the deployment is created but the machine has not shown up yet. So Juju will now query any _incomplete_  deployments, and augment that list with any machines as well. Any machines found take precedence over deployments. The method to gather instances uses the same logic as for EC2 - a short retry loop to allow the requested instances time to get created.

Also fix some lint issues and the resources API version needed to be bumped.

## QA steps

bootstrap to a previously created resource group, add a machine to get 2 azure deployments recorded, destroy the controller, manually delete the resource group contents, and bootstrap again
```plain
juju bootstrap --config resource-group-name=test --no-default-model azure/eastus aztest
juju add-machine
juju destroy controller aztest
```
use the azure portal to see there's 2 Deployments recorded; manually delete resource group content
bootstrap again
`juju bootstrap --config resource-group-name=test --no-default-model azure/eastus`

Previously this would fail as per the bug.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1935979

### DIFF
--- a/provider/azure/environ.go
+++ b/provider/azure/environ.go
@@ -18,7 +18,7 @@ import (
 	keyvaultservices "github.com/Azure/azure-sdk-for-go/services/keyvault/2016-10-01/keyvault"
 	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2018-02-14/keyvault"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-08-01/network"
-	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2020-06-01/resources"
 	legacystorage "github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2017-10-01/storage" // Pin this legacy storage API to 2017-10-01 since it's only used for unmanaged storage in models was created in 2.2 or earlier.
 	azurestorage "github.com/Azure/azure-sdk-for-go/storage"
 	"github.com/Azure/go-autorest/autorest"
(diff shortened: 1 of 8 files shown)


wallyworld:
[azure bootstrap mis-count of images](https://trello.com/c/7UtUF4nJ/1671-azure-bootstrap-mis-count-of-images)

provider/azure/environ.go
```diff
@@ -1464,49 +1462,9 @@ func (env *azureEnviron) deleteVirtualMachine(
 	return nil
 }
 
-// Instances is specified in the Environ interface.
```
wallyworld, line 1467: This block is moved below to be near the other instance query apis

provider/azure/environ.go
```diff
@@ -1700,6 +1805,71 @@ func isControllerDeployment(deployment resources.DeploymentExtended) bool {
 	return false
 }
 
+// allProvisionedInstances returns all of the instances
+// in the given resource group.
+func (env *azureEnviron) allProvisionedInstances(
+	ctx context.ProviderCallContext,
+	resourceGroup string,
+	controllerUUID string,
```
ycliuhw, line 1813: if "controllerUUID" != "", we only look for controller instances rather than instances managed by the current controller?

provider/azure/environ.go
```diff
@@ -1604,25 +1562,170 @@ func (env *azureEnviron) updateResourceControllerTag(
...
 func (env *azureEnviron) AllRunningInstances(ctx context.ProviderCallContext) ([]instances.Instance, error) {
-	return env.AllInstances(ctx)
+	return env.allInstances(ctx, env.resourceGroup, true, "", runningInstStates...)
+}
+
+// gatherInstances tries to get information on each instance id
+// whose corresponding insts slot is nil.
+// This function returns environs.ErrPartialInstances if the
+// insts slice has not been completely filled.
+func (env *azureEnviron) gatherInstances(
```
ycliuhw, line 1629: Azure doesn't have an API for querying instances by a slice of IDs?

provider/azure/environ.go
wallyworld: yes, we look for controller instances with that controller uuid

ycliuhw (approved):
Just two questions, otherwise, LGTM thanks!

provider/azure/environ.go
wallyworld: sadly no - you can ask for instances in a given resource group, but the api has no filter parameter

wallyworld:
$$merge$$

PR #13179, opened 2021-07-20
Update charm error messages

SimonRichardson:
The following PR attempts to standardise the error messages when
deploying a charm that contains information from a metadata format v1
with information from a metadata format v2.

Having non-nil manifests in the database has made this much harder, as
we can no longer inspect if it's nil (lack of a manifest). Instead, we have
to check if bases exist and or containers exist in the metadata to determine
the correct format. To improve this situation we now have an upgrade step
that removes it, if it's empty.

Additionally, I've removed the format logic in core/charm and moved it
to juju/charm/v8. That way we've got consistent data in the same location.
Ideally, I'd rather not leak format information out and just use feature
detection; if there are containers do X. Instead we have if format v2 do
X.

## QA steps

Notice the error message tells you why it failed.

```sh
$ juju bootstrap microk8s m8
$ juju add-model test
$ git clone --depth 1 https://github.com/canonical/alertmanager-operator.git
$ juju deploy ./alertmanager-operator
ERROR  attempting to deploy "./alertmanager-operator": containers without a manifest.yaml not valid
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1936281


manadart (approved):
Nice. Works.

SimonRichardson:
$$merge$$

PR #13185, opened 2021-07-21
Logging: Reset optional fields

SimonRichardson:
When the logsink reads the log records on the connection, it reuses one
log record for all reads. The problem occurs when there is a partial log
record that can have missing entities or labels. As entity and labels
are optional; if they happen to be set, but the next one doesn't have
them, then the ReadJSON (unmarshaller) doesn't clear ALL the fields, it
just updates what is required. This results in labels from the previous
read being in all future reads until a new record happens to have a new
entity or label.

Assuming a stream of log records from A->B->C->D was sent and only B had
an entity, all subsequent records (C->D) would have the same entity as
B.

The code is simple, just create a new instance of LogRecord.

## QA steps

Ideally, you need two terminals running (tmux is wise).

#### First terminal

```sh
```

#### Second terminal

```sh
$ juju bootstrap lxd test
$ juju info bad
$ juju model-config -m controller "logging-config='<root>=DEBUG"
```

#### Expectations

There should be no labels on any logs that aren't `juju.apiserver.charmhub.client` or `juju.apiserver.charmhub.client.transport`.

### DIFF
--- a/apiserver/logsink/logsink.go
+++ b/apiserver/logsink/logsink.go
@@ -304,6 +304,9 @@ func (h *logSinkHandler) receiveLogs(socket *websocket.Conn,
 		defer close(logCh)
 		var m params.LogRecord
 		for {
+			// ResetOptionalFields will remove any stale data for a log record
+			// that might have stale data when reading the JSON.
+			m.ResetOptionalFields()
 			// Receive() blocks until data arrives but will also be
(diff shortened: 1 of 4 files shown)


achilleasa (approved):
LGTM

SimonRichardson:
$$merge$$

PR #13186, opened 2021-07-22
Charm: Allow series and manifest.yaml

SimonRichardson:
The following adds the ability to have a series and a manifest, only to
treat this style of charm as a V1.

This is a regression introduced when moving the format detection from
v1 to v2.

## QA steps

```sh
$ juju bootstrap microk8s m8s
$ juju deploy cs:dex-auth
```


manadart (approved):

SimonRichardson:
$$merge$$

PR #13188, opened 2021-07-22
[2.9] Fix filter by tag bug in equinix provider

achilleasa:
The original filtering logic had a bug where the filtering code would
retain instances when the intersection of the device tags and query tags
was not empty. As a result, devices would survive the filtering as long
as they had at least one of the queried tags present.

When destroying controllers, the provider searches for instances with
the following tags:
- juju-is-controller=true
- juju-controller-uuid=$UUID

As a result, the prior filtering logic would match *all* controllers
(since all of them have the `juju-is-controller=true` tag set) and cause them
to be wrongly expunged.

## QA steps
Bootstrap 2 controllers in any region (e.g. `fr` or `am`). Then delete one of the controllers and ensure that the second remains intact.


achilleasa:
!!build!! (windows tests aborted)

SimonRichardson (approved):

achilleasa:
$$merge$$

PR #13189, opened 2021-07-23
Download and deploy the controller charm from charmhub at bootstrap

wallyworld:
There's a new "juju-controller" charm published to charmhub as beta.

When bootstrapping, we now deploy that charm. Previously there was only the option to deploy a local controller charm.

## QA steps

juju bootstrap aws --no-default-model
juju status

You will see a controller application and unit for the bootstrap machine.

Or you can do

juju bootstrap aws --no-default-model --controller-charm-risk=edge

or

juju bootstrap aws --no-default-model --controller-charm-path=/path/to/local/charm

### DIFF
--- a/cmd/juju/commands/bootstrap.go
+++ b/cmd/juju/commands/bootstrap.go
@@ -249,6 +249,7 @@ type bootstrapCommand struct {
 	noHostedModel   bool
 
 	ControllerCharmPath string
+	ControllerCharmRisk string
 
 	// Force is used to allow a bootstrap to be run on unsupported series.
 	Force bool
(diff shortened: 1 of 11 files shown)


wallyworld:
[juju controller charm](https://trello.com/c/v58jssJ4/1684-juju-controller-charm)

cmd/juju/commands/bootstrap.go
```diff
@@ -329,9 +331,12 @@ func (c *bootstrapCommand) SetFlags(f *gnuflag.FlagSet) {
 	f.BoolVar(&c.noSwitch, "no-switch", false, "Do not switch to the newly created controller")
 	f.BoolVar(&c.Force, "force", false, "Allow the bypassing of checks such as supported series")
 	f.BoolVar(&c.noHostedModel, "no-default-model", false, "Do not create a default model")
-	f.StringVar(&c.ControllerCharmPath, "controller-charm", "", "Path to a locally built controller charm")
+	f.StringVar(&c.ControllerCharmPath, "controller-charm-path", "", "Path to a locally built controller charm")
+	f.StringVar(&c.ControllerCharmRisk, "controller-charm-risk", "beta", "The controller charm risk if not using a local charm")
 }
 
+var controllerCharmRisks = set.NewStrings("edge", "beta", "candidate", "stable")
```
tlm, line 338: Is there a need to put these constants in a separate package that represents the business for the controller charm?

cmd/jujud/agent/controllercharm.go
```diff
@@ -0,0 +1,265 @@
+// Copyright 2021 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package agent
```
tlm, line 4: Should this be in it's own package as theoretically it doesn't need to be tied to the bootstrap process.

cmd/juju/commands/bootstrap.go
wallyworld: I've changed to use the Risk values from the charms repo

cmd/jujud/agent/controllercharm.go
wallyworld: As with the step to download and unpack the dashboard (done just before the controller charm set up), this is part of bootstrap in that it needs to be done as part of setting up the controller. It imports apiserver, state, environ and so needs to be done in a place where we don't run into import loops or break layering etc.

tlm (approved):
lgtm

wallyworld:
$$merge$$

PR #13190, opened 2021-07-23
Move to Pubsub v2

SimonRichardson:
The following moves to pubsub v2, which doesn't require everyone to wait
for every subscriber to be done. Instead, if you want the old
functionality you can call the high-level function in the package.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju enable-ha
$ juju deploy cs:~jameinel/ubuntu-lite-9 ubuntu
$ juju deploy cs:~jameinel/ubuntu-lite-9 ubuntux
$ juju deploy cs:~jameinel/ubuntu-lite-9 ubuntuz
$ juju remove-machine 1 -m controller
$ juju enable-ha
```

Check the raft logs to see if there are any errors.

```sh
juju model-config -m controller logging-config="<root>=INFO;juju.worker.lease.raft=TRACE;juju.core.raftlease=TRACE"
```
## Bug reference

TBA

### DIFF
--- a/apiserver/observer/request_notifier.go
+++ b/apiserver/observer/request_notifier.go
@@ -18,18 +18,17 @@ import (
 // Hub defines the only method of the apiserver centralhub that
 // the observer uses.
 type Hub interface {
-	Publish(topic string, data interface{}) (<-chan struct{}, error)
+	Publish(topic string, data interface{}) (func(), error)
 }
 
 // RequestObserver serves as a sink for API server requests and
 // responses.
 type RequestObserver struct {
-	clock              clock.Clock
-	hub                Hub
-	logger             loggo.Logger
-	connLogger         loggo.Logger
-	pingLogger         loggo.Logger
-	apiConnectionCount func() int64
+	clock      clock.Clock
+	hub        Hub
+	logger     loggo.Logger
+	connLogger loggo.Logger
+	pingLogger loggo.Logger
 
 	// state represents information that's built up as methods on this
 	// type are called. We segregate this to ensure it's clear what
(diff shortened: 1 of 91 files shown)


apiserver/observer/request_notifier.go
```diff
@@ -95,7 +94,7 @@ func (n *RequestObserver) Login(entity names.Tag, model names.ModelTag, fromCont
 		if !n.state.fromController {
 			n.connLogger.Infof("agent login: %s for %s", n.state.tag, n.state.model)
 		}
-		n.hub.Publish(apiserver.ConnectTopic, apiserver.APIConnection{
+		_, _ = n.hub.Publish(apiserver.ConnectTopic, apiserver.APIConnection{
```
jameinel, line 97: This seems dangerous that we're dropping the error return on the floor unilaterally.
Thank you for catching it, but it should be something we follow up on quickly. Even if the answer is just logging that we got an error.

worker/modelcache/manifold.go
```diff
@@ -5,7 +5,7 @@ package modelcache
 
 import (
 	"github.com/juju/errors"
-	"github.com/juju/pubsub"
+	"github.com/juju/pubsub/v2"
```
jameinel, line 8: Does dependency engine context.Get() support passing into an interface? This is the sort of thing that feels like instead of:
```
var hub *pubsub.StructuredHub
if err := context.Get(..., &hub)
```
Into a different way of grabbing the instance of an interface that we need.

That isn't really *this* patch as much as it is a factor for something we should be considering for patterns that we use.

jameinel (approved):
It would be nice to have more local interfaces and have them work, but this does seem to be just a rename, and catching a couple places where we were ignoring errors.

SimonRichardson:
$$merge$$

PR #13191, opened 2021-07-23
[dev] Refactor charm downloader

achilleasa:
This PR is a precursor to the work for server-side bundle deployments. It:
- refactors what was previously known as download strategies in `core/charm` into a store- and storage-agnostic charm downloader (in `core/charm/downloader`).
- extracts the repository implementations from the `charms` facade, refactors them (primarily to be able to pass macaroons through) and moves them to the `core/charm/repository` package.
- provides re-usable services (currently living in `facades/client/charms/services`) which wire the state/storage/repository factory implementations into the core/charm/downloader. This allows the charm download functionality to be re-used between facades (and by the apiserver charm upload code).

The `facades/client/charms/services` provides a repository factory which when provided with a `corecharm.Source` yields a suitable `core/charm/repository` instance. In addition, the factory memoizes the returned repositories allowing them to be re-used (this addresses a TODO in the charms facade).

Finally, all charm-related methods from the `applications` facade which where deprecated in Juju 2.9 now return a NotSupported error advising the caller to use the methods from the charms facade instead.

Please make sure to read the extended commit logs for additional information!

## QA steps

Pretty much everything in the knock spreadsheet: 
- deploy local, cs, ch charms
- juju download x
- juju refresh

PR #13195, opened 2021-07-26
add logger label of charmhub

hmlanigan:
Add CHARMHUB logger label.  Ensure its used by adding to the logger in the client creation, rather than the callers.

## QA steps

```
$ juju model-config  "logging-config='#charmhub=TRACE'" -m controller

# while watching the output of `juju debug-log -m controller`, run the following and verify output.
$ juju deploy juju-qa-test
$ juju info juju-qa-test
$ juju find juju-qa-test
```


achilleasa (approved):
LGTM. You should probably forward port this to develop when you get the chance so any conflicts due to the changes introduced by #13191 can be resolved early on.

hmlanigan:
$$merge$$

PR #13196, opened 2021-07-27
Add another whitelisted network error to bootstrap check

wallyworld:
When bootstrapping, we connect to the controller api to check that things have come up correctly. There's an allowable set of errors that cause a retry (eg i/o timeout). Anything else is fatal. A new case has emerged of an error that needs to be allowed for: "network is unreachable".

## QA steps

bootstrap to aws

## Bug reference

https://bugs.launchpad.net/juju/+bug/1938019


hpidcock (approved):
LGTM

wallyworld:
$$merge$$

PR #13200, opened 2021-07-28
Deploy by revision resources prep

hmlanigan:
A group of smaller commits in preparation for deploy by resource revision.
- Update charmhub.DownloadOneFromRevision to not require a base.  Both the download and install actions react the same way with regards to bases and fields.  For two facades, this involves a new method of series verification, as it can no longer be done by the api call.
- A little clean up.

There should be no difference in current behavior unless deploying a charm with a specific resource revision, then the error is changed from:
ERROR Instance key '4d27f01e-5c45-4553-8960-3c99cb983936' invalid 'base' in 'download' for charm_id=Hw30RWzpUBnJLGtO71SX8VDWvd3WrjaJ name=juju-qa-test
to
ERROR charm resource "foo-file" at revision 3 not found

This will be fixed in the near future with deploy charm by revision and requiring resource revisions.

## QA Steps

$ juju deploy juju-qa-test --resource foo-file=3
# no error for below
$ juju deploy juju-qa-test


achilleasa (approved):
LGTM

hmlanigan:
$$merge$$

PR #13201, opened 2021-07-29
Use separate matching functions for block and filesystem volume devices

wallyworld:
There's a hard to reproduce issue deploying a charm with block storage on Azure - the incorrect device path is being given to the charm to use. It appears there might be an issue if, external to the charm, a partition is created on that block device; this could confuse the matching algorithm. There's 2 cases where we need to determine what block device represents a juju volume:
- some block storage is needed by the charm
- the charm requests filesystem storage and this is done by creating a volume and adding a filesystem on top

This PR separates the matching in the 2 cases, to avoid a potential incorrect match if there's a partition added to what Juju expects to be block storage. We also prefer to use the devicelink for the mount point instead of disk-by-uuid.

## QA steps

The issue in the bug could not be reproduced, but I deploy a charm with both block and filesystem storage to
- lxd
- azure
- google
- aws

And checked juju storage --format yaml and also the /etc/fstab on the unit's machine to ensure everything as set up as expected.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1936752

### DIFF
--- a/storage/path.go
+++ b/storage/path.go
@@ -27,12 +27,12 @@ func BlockDevicePath(device BlockDevice) (string, error) {
 	if device.HardwareId != "" {
 		return path.Join(diskByID, device.HardwareId), nil
 	}
-	if device.UUID != "" {
-		return path.Join(diskByUUID, device.UUID), nil
-	}
 	if len(device.DeviceLinks) > 0 {
 		// return the first device link in the list
 		return device.DeviceLinks[0], nil
+	}
+	if device.UUID != "" {
+		return path.Join(diskByUUID, device.UUID), nil
 	}
 	if device.DeviceName != "" {
 		return path.Join(diskByDeviceName, device.DeviceName), nil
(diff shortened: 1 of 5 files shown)


storage/path.go
```diff
@@ -27,13 +27,13 @@ func BlockDevicePath(device BlockDevice) (string, error) {
...
 		return path.Join(diskByID, device.HardwareId), nil
 	}
-	if device.UUID != "" {
-		return path.Join(diskByUUID, device.UUID), nil
-	}
 	if len(device.DeviceLinks) > 0 {
 		// return the first device link in the list
 		return device.DeviceLinks[0], nil
 	}
+	if device.UUID != "" {
```
ycliuhw, line 34: could we add a comment why `DeviceLinks` takes higher priority to return here?

ycliuhw (approved):
lgtm thanks

wallyworld:
$$merge$$

storage/path.go
wallyworld: It's somewhat arbitrary - there's no real reason to choose one over the other. The bug seemed to show that a UUID path was involved but because it's been impossible to reproduce, it's hard to definitively say what order is preferred. Once this gets tested on site, we can revisit if needed.

wallyworld:
[azure storage bug](https://trello.com/c/NZi5Oudx/1682-azure-storage-bug)

PR #13204, opened 2021-07-30
Remove the now-unused ApplicationWatcherFacade

benhoyt:
Well, strictly speaking it's used but only in FilterNone mode (see https://github.com/juju/juju/pull/13090), so we might as well call `WatchApplications` directly once again.

This is a partial revert of https://github.com/juju/juju/pull/11979, when `ApplicationWatcherFacade` was added, plus a bit of rejiggering to bring things up to date.

This change *should* be a no-op functionality-wise. For QA, I'm running the same manual tests as described in https://github.com/juju/juju/pull/13090.


wallyworld (approved):

benhoyt:
$$merge$$

PR #13210, opened 2021-08-03
Add note and todo.

hmlanigan:
Avoid breaking a snowflake OpenStack cloud in the future.

Comments only have changed.


SimonRichardson (approved):
It would be nice to query the network for the UUID to make sure it exists, but for now, the NOTE is fine.

hmlanigan:
$$merge$$

PR #13211, opened 2021-08-03
Fix data race (accessing charmInfo) in TestV2CharmExitsApplicationWorker

benhoyt:
This fixes a data race accessing the charmInfo returned by the mocked `ApplicationCharmInfo` function. This was happening because we synchronize by waiting for the `ApplicationCharmInfo` call itself, but then write to `charmInfo` straight away, and the `charmFormat()` function may not be finished reading it yet, causing a data race.

The fix removes the first call to `ApplicationCharmInfo` in the application worker (there's no need to call it twice), and then waits for the main worker's `Life()` call to synchronize and know when the worker has definitely finished the `charmFormat()` call before changing `charmInfo` fields.

See `go test -race ./worker/caasunitprovisioner/` output [here](https://pastebin.canonical.com/p/yXbfhF8d7H/).


hpidcock (approved):
LGTM thank-you

benhoyt:
$$merge$$

PR #13217, opened 2021-08-05
Only 1 origin in deployCommand stuct, remove the duplication.

hmlanigan:
deployCommand should have only 1 origin, not 2.  Remove the old one.

## QA steps

There should be no change to current charm deployment behavior.

```console 
$ juju deploy cs:ubuntu
$ juju deploy juju-qa-test --channel 2.0/stable
$ juju deploy ./testcharms/charm-repo/bionic/ubuntu-plus
```

### DIFF
--- a/cmd/juju/application/deployer/charm.go
+++ b/cmd/juju/application/deployer/charm.go
@@ -47,7 +47,6 @@ type deployCharm struct {
 	flagSet          *gnuflag.FlagSet
 	model            ModelCommand
 	numUnits         int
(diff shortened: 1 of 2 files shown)


SimonRichardson (approved):
I know what happened here, we originally didn't have an origin as part of `application.CharmID` that came later when we added bundles. As bundles was a later addition we didn't go back and clean this up. 

Thanks for doing this, these are good changes!

hmlanigan:
$$merge$$

PR #13218, opened 2021-08-06
Show application data set by the remote app in show-unit output

wallyworld:
`juju show-unit foo` displays relation data as set by the app/unit foo is related to.
In the case of application relation data, we were showing the data that is set by the nominated unit, not the data that was set by the other side.

## QA steps

I used local copies of the controller and dashboard charms
juju deploy controller
juju deploy dashboard
juju relate controller dashboard
juju show-unit dashboard/0

App data would have been empty but is now
```
    application-data:
      controller-url: wss://10.160.126.204/api
      is-juju: "True"
      model-url-template: wss://10.160.126.204/model/${modelUUID}/api
```


wallyworld:
[show-unit fix](https://trello.com/c/loxdFqfl/1700-show-unit-fix)

ycliuhw (approved):
LGTM thanks

wallyworld:
$$merge$$

PR #13221, opened 2021-08-06
Status watch flag

juanmanuel-tirado:
A very common user behaviour is to watch the status of deployments for a while. Actually, in the documents this is done using the `watch` command in conjunction with `juju status`. I have defined an additional flag for the `status` command called `--watch` followed by a duration value that runs 'status' commands every given interval of time endlessly.

## QA steps

To test simply deploy any application and watch the iterative status calls.

```sh
juju deploy hello-juju
# print status and exit
juju status
# print the status every two seconds
juju status --watch=2s
```
The operation continues until the program is killed.

## Documentation changes

The flag is documented in the command docs.


SimonRichardson (commented):
It would be nice to use the `AllWatcher` to update the status rather than calling `runStatus` every time (it's an expensive operation on large models), but that's rather complicated to implement.

cmd/juju/status/status.go
```diff
@@ -377,6 +383,28 @@ func (c *statusCommand) Run(ctx *cmd.Context) error {
...
+	err := c.runStatus(ctx)
+	if err != nil || c.watch == 0 {
+		return err
+	}
+
+	// repeat the call using a ticker
+	ticker := time.NewTicker(c.watch)
+
+	for range ticker.C {
+		err := c.runStatus(ctx)
```
jameinel, line 402: The one usability problem with this vs `watch --color juju status --color` is that it just scrolls more and more on the screen, rather than what `watch` does (by putting a fresh screen on each refresh, etc).

So I'm tempted to keep the command line arg, but it actually forks out to `watch -n X --color` (if juju status itself is already in color).

juanmanuel-tirado: That's true. If we want, we can replicate the watch behaviour using ANSI scapes like `fmt.Printf("\033[2J")` to clear the screen.

cmd/juju/status/status.go
```diff
@@ -384,13 +387,17 @@ func (c *statusCommand) runStatus(ctx *cmd.Context) error {
...
-	return nil
+	return numLines, nil
 }
 
 func (c *statusCommand) Run(ctx *cmd.Context) error {
 	defer c.close()
 
-	err := c.runStatus(ctx)
+	if c.watch != 0 {
+		fmt.Printf("\u001Bc")
```
SimonRichardson, line 397: Does this work for windows?

juanmanuel-tirado: This is is valid for ANSI compatible terminals but it has to be tested on Windows.

juanmanuel-tirado: I have added additional helpers just for the windows case based on https://github.com/inancgumus/screen/blob/master/clear_windows.go. Anyway, this has to be tested.

cmd/juju/status/status.go
```diff
@@ -134,6 +138,10 @@ Examples:
     # Provide output as valid JSON
     juju status --format=json
 
+    # Watch the status of the mysql application every five seconds
+	# Only available for unix-based systems.
+    juju status --watch 5s mysql
+
 Further reading:
 
     https://juju.is/docs/command/status
```
jnsgruk, line 147: A nit, but this URL doesn't exist

juanmanuel-tirado: Good point. I guess if there are more of these urls in the help for the other commands.

jnsgruk: Yeh, quite likely!

cmd/juju/status/screen_helper_windows.go
```diff
@@ -0,0 +1,86 @@
+// +build windows
```
SimonRichardson, line 1: We can remove this file if we don't support windows?

SimonRichardson (commented):
Once we remove the windows file, we should land this.

cmd/juju/status/screen_helper_unix.go
```diff
@@ -0,0 +1,14 @@
...
+
+// +build !windows
+
+package status
+
+import "fmt"
+
+// ClearScreen removes any character from the terminal
+// using ANSI scape characters.
+func ClearScreen() {
```
SimonRichardson, line 12: This can just be put back in the other file now.

juanmanuel-tirado:
!!merge!!

SimonRichardson (approved):

SimonRichardson:
$$merge$$

PR #13224, opened 2021-08-06
close the edge case producing a nil origin

hmlanigan:
If juju deploy of a charmhub charm with resources, failed after AddCharm was called, then subsequent deploy attempts of the same charm would fail in deployResource due to a nil origin.  The charm had already been downloaded, but no origin was available, and no error viewed when attempting to find the origin.

In the above case, down load the charm a 2nd time so we can have an application and origin.

## QA steps

```console

# cause a charmhub charm with resources to fail after AddCharm during deploy.
$ juju deploy juju-qa-test --storage store=10G --to lxd
Located charm "juju-qa-test" in charm-hub, revision 13
ERROR adding storage to lxd container not supported

# deploy the charm again, without options which cause a failure.
$ juju deploy juju-qa-test
Located charm "juju-qa-test" in charm-hub, revision 13
Deploying "juju-qa-test" from charm-hub charm "juju-qa-test", revision 13 in channel stable
```

### DIFF
--- a/apiserver/facades/client/charms/services/storage.go
+++ b/apiserver/facades/client/charms/services/storage.go
@@ -56,13 +56,21 @@ func (s *CharmStorage) PrepareToStoreCharm(charmURL *charm.URL) error {
 	}
 
 	if ch.IsUploaded() {
-		// Since the charm is already uploaded this call should not fail.
-		// We can safely ignore errors here.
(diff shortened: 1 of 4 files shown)


hmlanigan:
Closing - the change which caused the nil origin is being removed in https://github.com/juju/juju/pull/13277

PR #13237, opened 2021-08-12
Add secrets manager api client and wire up hook commands

wallyworld:
Add the secrets manager client API and wireup to the hook commands. The backend is still todo.

The client api provides 2 methods - create a secret and get a secret value. The create api takes a config to tell it how to create the secret. Hook commands don't yet support password secrets.

To make the code simpler, the secret value data has been changed from a `[]byte` to a `string`.

## QA steps

None yet, just unit tests.


core/secrets/secret.go
```diff
@@ -0,0 +1,88 @@
...
+		Scope: ScopeApplication,
+		Path:  strings.Join(nameParts, "."),
+		Params: map[string]interface{}{
+			PasswordLength:       length,
+			PasswordSpecialChars: specialChars,
+		},
+	}
+}
+
+var pathRegexp = regexp.MustCompile("^[a-zA-Z0-9]+(\\.[a-zA-Z0-9]+)*$")
```
benhoyt, line 74: Might be better specified with a backtick-string to avoid the double `\\` (regex literals normally are).

api/secretsmanager/client.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+// Create creates a new secret.
+func (c *Client) Create(cfg *secrets.SecretConfig, value secrets.SecretValue) (string, error) {
+	if err := cfg.Validate(); err != nil {
+		return "", errors.Trace(err)
+	}
+
+	dataOrNil := func(value secrets.SecretValue) map[string]string {
```
benhoyt, line 32: Clearer to just write this use-once function inline, like so?

```go
var data map[string]string
if value != nil {
    data = value.EncodedValues()
}
```

apiserver/params/secrets.go
```diff
@@ -0,0 +1,40 @@
...
+type CreateSecretArgs struct {
+	Args []CreateSecretArg `json:"args"`
+}
+
+// CreateSecretArg holds the args for creating a secret.
+type CreateSecretArg struct {
+	Type   string                 `json:"type"`
+	Path   string                 `json:"path"`
+	Scope  string                 `json:"scope"`
+	Params map[string]interface{} `json:"params"`
```
benhoyt, line 16: Is there somewhere the valid keys/values for `Params` will be documented? Maybe a doc comment on each of these fields, with an example?

core/secrets/secret.go
```diff
@@ -0,0 +1,88 @@
...
+		Type:  TypeBlob,
+		Scope: ScopeApplication,
+		Path:  strings.Join(nameParts, "."),
+	}
+}
+
+// TODO(wallyworld) - use a schema to describe the config
+const (
+	PasswordLength       = "password-length"
+	PasswordSpecialChars = "password-special-chars"
```
benhoyt, line 58: Ah, I see -- here's kinda where they're documented.

benhoyt (commented):
Looking good so far!

benhoyt (approved):

apiserver/params/secrets.go
wallyworld: params still evolving, i'll add a todo and some doc

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
[secrets client api](https://trello.com/c/rwrL2ylj/1711-secrets-client-api)

PR #13241, opened 2021-08-16
Force nested lxd containers to be bionic by default in bash tests

wallyworld:
Nested focal lxd containers cannot run snap due to an apparmor namespace issue.
CI tests are failing because we try and start a focal workload machine when running on LXD.
For now, until we port the test infrastructure to use something like ephemeral nodes, change the default workload series to bionic.

## QA steps

```sh
cd tests
./main smoke
```

### DIFF
--- a/tests/includes/juju.sh
+++ b/tests/includes/juju.sh
@@ -219,9 +219,19 @@ juju_bootstrap() {
 		esac
 	fi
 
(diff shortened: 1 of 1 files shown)


wallyworld:
[use bionic workload machines by default in bash tests](https://trello.com/c/tdI7Z1P3/1716-use-bionic-workload-machines-by-default-in-bash-tests)

SimonRichardson (approved):

PR #13247, opened 2021-08-17
Create a channel in a more permissive manner

SimonRichardson:
The following allows migrations and exporting of bundles in a more
permissive manner. In that, they allow you to create a channel without a
risk. This risk is then set to stable once we normalise it at a later
stage, which ensures we always have the right value.

## QA steps

```sh
$ juju bootstrap lxd test
$ juju  deploy cs:ubuntu  ubuntua
$ juju deploy cs:ubuntu  ubuntub
$ juju deploy cs:ubuntu  ubuntuc
$ juju mongo
# Replace $MODEL_ID with your model id.
$ db.applications.updateOne({_id:"$MODEL_ID:ubuntuc"}, {$set:{"charm-origin.channel.risk":""}})
$ exit
$ juju export-bundle
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1939601

### DIFF
--- a/state/migration_export.go
+++ b/state/migration_export.go
@@ -1947,11 +1947,7 @@ func (e *exporter) getCharmOrigin(doc applicationDoc, defaultArch string) (descr
 
 	var channel charm.Channel
 	if origin.Channel != nil {
-		var err error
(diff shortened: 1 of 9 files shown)


achilleasa (approved):

SimonRichardson:
$$merge$$

PR #13256, opened 2021-08-18
Ensure unsubscribe on error

SimonRichardson:
If we subscribe to a given topic, we need to ensure that we unsubscribe
when we get an error. If we don't do this, then we can end up having a
instance hanging around forever. This could end up being a leak of
resources.

Unfortunately, pubsub is never tied to the lifecycle of an instance. It 
would be nice to either pass in a tomb or a context to do auto
unsubscribing once a lifecycle is complete.

## QA steps

Tests pass.


manadart (approved):
Good catch.

SimonRichardson:
$$merge$$

PR #13257, opened 2021-08-18
Only update homebrew release on juju/juju

SimonRichardson:
It's pointless running homebrew on other repos, it just doesn't make
sense to do that. So instead we should check if we can run it.

See:

 - https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions
 - https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context


jameinel (approved):
I think this handles my issues around having homebrew trying to publish from my local copy

SimonRichardson:
$$merge$$

PR #13259, opened 2021-08-18
fix bundle changes sort infinite loop

hmlanigan:
The current sort algorithm for bundle changes causes some unit tests to timeout.  The panic traces show an infinite loop in the sort.

Replacing the bundle changes sort algorithm for a topological sort.  This brought to light some bugs with bundle changes not having all necessary requirements listed for the sort to be accurate for deployment. Such as ensuring the machines are created in the order of specification so that machine IDs match.  And ensuring the upgrade charm is run after the new charm is added to the controller.

This change does change the order of output for deploying a bundle, however it should not change the results.

## QA steps

Run many types of bundle deployments including:
- Using existing machines
- Bundles that will fail, such as series mismatch with charm or defined machine.
- Bundles which upgrade charms
- Bundles specifying machines, using --map-existing for some already existing machines in the model.

### DIFF
--- a/core/bundle/changes/changes.go
+++ b/core/bundle/changes/changes.go
@@ -1228,7 +1228,7 @@ func (cs *changeset) dependents() map[string][]string {
 func (cs *changeset) sorted() ([]Change, error) {
 	// Exit it early if there is nothing to sort.
 	if len(cs.changes) == 0 {
-		return []Change{}, nil
+		return nil, nil
 	}
 
 	// Create a map to sort the data.
@@ -1242,7 +1242,7 @@ func (cs *changeset) sorted() ([]Change, error) {
 		data[c.Id()] = c.Requires()
 		dataOrder[i] = c.Id()
 	}
-	sortedChangeIDs, err := toposort_flatten(dataOrder, data)
+	sortedChangeIDs, err := toposortFlatten(dataOrder, data)
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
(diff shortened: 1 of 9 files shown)


core/bundle/changes/changes.go
```diff
@@ -1225,29 +1225,110 @@ func (cs *changeset) dependents() map[string][]string {
...
+}
+
+// toposort_flatten performs a topographical flattened sort on the provided
+// data.  dataOrder is a slice of the originally ordered changes. data is a
+// map of change to requirements.  To be idempotent, requirements must be in
+// the same order each time.  Use along with data to ensure that this method
+// is deterministic.
+//
+// Adapted from https://tylercipriani.com/blog/2017/09/13/topographical-sorting-in-golang/
+func toposort_flatten(dataOrder []string, data map[string][]string) ([]string, error) {
```
hmlanigan, line 1271: This is the main change for the PR

hmlanigan:
!!build!!

core/bundle/changes/changes.go
```diff
@@ -1225,29 +1225,110 @@ func (cs *changeset) dependents() map[string][]string {
...
-			if !done.Contains(r) {
-
-				// This change requires a change which is not yet listed.
-				// Push this change at the end of the list and retry later.
-				changes = append(changes, change)
-				continue mainloop
+func (cs *changeset) sorted() ([]Change, error) {
+	// Exit it early if there is nothing to sort.
+	if len(cs.changes) == 0 {
+		return []Change{}, nil
```
hpidcock, line 1231: nil, nil

core/bundle/changes/changes.go
hpidcock, line 1271: the function name should probably be topoSortFlatten

hpidcock: also would be nice to add a test for this function directly.

hpidcock: Also would be good to attribute with some more information: CC BY-SA 4.0 "If supplied, you must provide the name of the creator and attribution parties, a copyright notice, a license notice, a disclaimer notice, and a link to the material. CC licenses prior to Version 4.0 also require you to provide the title of the material if supplied, and may have other slight differences."

hpidcock (approved):

core/bundle/changes/changes.go
hmlanigan: tests are in changessort_test.go

hmlanigan:
Addressed comments and rebased to develop

hmlanigan:
$$merge$$

PR #13206, opened 2021-07-30
Fix deploy with resource revision for Charmhub charms

hmlanigan:
Part of work to deploy a charm by revision.  A pieces of the work to ensure that when you deploy a Charmhub charm by revision, that you also specify the resource revisions to be used.  Enforcing this will be the next step.

When resolving resources, if it has a revision and is not a resource which has already been uploaded, look for it specifically with download action of Charmhub refresh including a resource-revision.  Add these results to the ones being resolved.  When using the download action of the Charmhub refresh command with a charm revision, resource revisions must be specified, do they will not be included in the results.

Drive by to remove FindDownloadURL mocks from tests.  It's been refactored and renamed to GetDownloadURL.

## QA steps

```sh
$  juju deploy juju-qa-test 
Located charm "juju-qa-test" in charm-hub, revision 13
Deploying "juju-qa-test" from charm-hub charm "juju-qa-test", revision 13 in channel stable
# note that foo-file revision 2 is listed below.
$ juju resources juju-qa-test
Resource  Supplied by  Revision
foo-file  charmstore   2

$ juju deploy juju-qa-test --resource foo-file=3 test-resource-revision
Located charm "juju-qa-test" in charm-hub, revision 13
Deploying "test-resource-revision" from charm-hub charm "juju-qa-test", revision 13 in channel stable
# note that foo-file revision  is listed below.
$ juju resources test-resource-revision
Resource  Supplied by  Revision
foo-file  charmstore   3
```

PR #13260, opened 2021-08-18
[dev] Implement async charm downloads

achilleasa:
This PR is part of the work for server-side bundle expansion. In particular, it implements the plumbing for enabling asynchronous charm downloads. The changes are hidden behind a **controller** feature flag so they can be tested without breaking the existing, synchronous, deploy flow.

The solution for async charm downloads consists of the following components:
- A watcher for detecting applications that reference a charm still pending to be downloaded
- A controller-level facade for exposing the watcher to clients and for triggering an asynchronous charm download
- The `charmdownloader` per-model worker which starts the watcher and makes charm download requests.

## Enabling support for asynchronous charm downloads

To avoid breaking the current synchronous deployment behavior, the PR changes are behind a **controller** feature flag which must be enabled as follows:

```
 juju controller-config 'features=[async-charm-downloads]'
```

### How AddCharm worked before the changes

Prior to these changes, `AddCharm` (client/charms facade) used a two-phase approach for adding a charm to state/blob store. That logic has been extracted in a previous PR into a dedicated [service](https://github.com/juju/juju/blob/develop/apiserver/facades/client/charms/services/storage.go#L52-L110).

Using just the charm URL, it would first attempt to create an entry in the `charms` collection (and bail out early if the doc already existed) with the `PendingUpload` flag set (see next paragraph for an explanation). Then, it would fetch the charm archive, upload it to the blob store and update the charm document with the charm metadata **and** the blob location/SHA256 while also setting the `PendingUpload` flag to `false`. 

### Changes to the AddCharm logic for async charm downloads

The `AddCharm` (client/charms facade) will execute a different code path when the controller flag is set. First, we try to lookup the charm document by the provided charm URL. Note that the existing charm lookup code does **not** return charm documents that are either placeholders or have the pending upload flag set to `true`. To support the async charm download use-case (where we have the metadata in the charm doc but the pending upload flag is still set), the lookup code will first examine the controller feature flags and _also_ allow lookup of pending charms. 

If the charm doc is present then `AddCharm` is essentially a no-op. However, the call is expected to return back the resolved origin for the charm. However, as charmhub charms may use the same revision in multiple channels, we still need to resolve the origin based on the preferred origin that the client passed through and return the result back.

Next, we fetch the charm archive. For this use-case we are only actually interested in the charm metadata and the LXD profiles. To avoid implementing this at this stage (according to the server-side bundle expansion spec we will provide a dedicated facade for the new deploy logic that will perform batch metadata-only lookups but that functionality is not yet available by the repository implementations) we emulate the metadata lookup by downloading the archive and collecting the metadata while dropping the rest of the blob contents.

Then, a new (to avoid messing with `AddCharm` and friends) state method called `AddCharmMetadata` is invoked which like `AddCharm` also accepts a `state.CharmInfo` parameter which has the `StoragePath` and `SHA256` intentionally left blank. The two blank fields instruct the `insertCharmOps` (state/charm.go) helper to set the `PendingUpload` flag to true.

So, now we end up with a charm document that:
- has all metadata populated.
- has the client-provided macaroons populated.
- has **no** storage path or SHA256 values set.
- has the `PendingUpload` flag set to true.

## Watching and reacting to pending charms.

To trigger asynchronous charm downloads, the PR provides a new watcher which monitors application documents for changes (also when they get **added** to the collection; see extended commit logs for more details). For each changed application document, it looks up the referenced charm and checks whether it is still pending to be downloaded. If that's the case, the **application ID** is emitted by the watcher.

A new worker (`charmdownloader`) and matching facades are provided which expose the watcher and provide the means for the worker to request from the controller to download pending charms for a list of applications. 

The download flow is pretty straightforward and uses the `core/charm/downloader` package. Prior to commencing a download we update the application status message to "downloading charm" and we update it again to "downloaded charm" when the download is complete.

## Changes to the blob retrieval endpoints

When the manifest deployer (in uniter) tries to fetch the blob data via the HTTP charm-related endpoints, we want to be able to distinguish between an unknown blob path (which yields a 404) and a blob that is pending to be downloaded. To this end, the PR introduces a new error type (NotYetAvailable) which is mapped to a 409 (Conflict) to let the client know that the blob data is not yet available but will be in the future and they need to retry the request.

## Changes to the manifest deployer (uniter)

The manifest deployer was written with the assumption that the blob would always be available (as blobs where synchronously downloaded when `AddCharm` was invoked during deployment). While this will generally still be the case even in the async charm case (charms are typically small and we will be able to download them before the unit machine gets provisioned), there is still a chance that the blob might not yet be available when the deployer tries to retrieve it (e.g. we deploy to a manual machine).

To address such cases, the deployer read bundle logic has been wrapped in a retry loop with a (jittered) exponential backoff delay generator. In addition, the `IsFatalError` handler of the retrier treats NotYetAvailable errors as non-fatal.

## QA steps

```console
$ juju bootstrap lxd --bootstrap-series bionic --no-dashboard

# Check that regular deployments work
$ juju deploy wordpress

# Enable async charm deployments
$ juju controller-config 'features=[async-charm-downloads]'

# In another tab watch juju status
$ watch -c juju status --color

# Deploy a charm
$ juju deploy percona-cluster -n 3

# The percona application should first display a "downloading charm" message (maintenance mode)
# followed by a "downloaded charm" message (unknown mode) and the units should show up green
```


api/charmdownloader/client_test.go
```diff
@@ -0,0 +1,16 @@
...
+	gc "gopkg.in/check.v1"
+)
+
+type charmDownloaderSuite struct {
+}
+
+var _ = gc.Suite(&charmDownloaderSuite{})
+
+func (s *charmDownloaderSuite) TestWatchApplicationsWithPendingCharms(c *gc.C) {
+}
```
SimonRichardson, line 16: Guessing this needs filling out?

apiserver/charms.go
```diff
@@ -532,7 +538,17 @@ func (h *charmsHandler) processGet(r *http.Request, st *state.State) (
...
 // object.
 func sendJSONError(w http.ResponseWriter, req *http.Request, err error) error {
-	logger.Errorf("returning error from %s %s: %s", req.Method, req.URL, errors.Details(err))
+	if errors.IsNotYetAvailable(err) {
+		// This error is typically raised when trying to fetch the blob
+		// contents for a charm which is still pending to be downloaded.
+		//
+		// We should log this at debug level to avoid unnecessary noise
+		// in the logs.
+		logger.Debugf("returning error from %s %s: %s", req.Method, req.URL, errors.Details(err))
```
SimonRichardson, line 547: Can we have a better message here...

`charm currently not available, returning error from ...`

apiserver/facades/client/charms/client.go
```diff
@@ -380,6 +412,77 @@ func (a *API) addCharmWithAuthorization(args params.AddCharmWithAuth) (params.Ch
...
+	// is to be downloaded.
+	if _, err := a.backendState.Charm(charmURL); err == nil {
+		// Try to lookup the origin associated with the application.
+		// Due to the way that charm deploys are implemented by the
+		// client, if the actual deploy call fails, the resolved origin
+		// returned by this method will not be recorded against the
+		// application (as the application doc will not be created).
+		//
+		// If we are unable to figure out the origin, we still need
+		// to execute the remainder of this method
```
SimonRichardson, line 433: You're in a weird state if the backend state has the charm, but doesn't have the uploaded charm origin. Are we not opening a race here?

apiserver/facades/client/charms/client.go
```diff
@@ -380,6 +412,77 @@ func (a *API) addCharmWithAuthorization(args params.AddCharmWithAuth) (params.Ch
...
+	repo, err := a.getCharmRepository(requestedOrigin.Source)
+	if err != nil {
+		return corecharm.Origin{}, errors.Trace(err)
+	}
+
+	// TODO(achilleasa):
+	// At this stage we are only interested in the charm metadata and lxd
+	// profile. However, the repo does not (yet) support metadata-only
+	// lookups so we will use the download API and extract the metadata
+	// we need.
```
SimonRichardson, line 453: You can...

```
curl -XPOST -s https://api.charmhub.io/v2/charms/refresh -H 'Content-type: application/json' -d '{"context": [], "actions": [{"name": "juju-controller", "base": {"name": "ubuntu", "channel": "16.04", "architecture": "amd64"}, "channel":"beta", "action": "download", "instance-key": "a-test"}], "fields":["metadata-yaml"]}' | jq
```

apiserver/charms.go
achilleasa: I intentionally kept it the same as the error above it for consistency; you will also get this information in the logged error details.

apiserver/facades/client/charms/client.go
achilleasa: The way it works ATM is:
- AddCharm X -> returns ResolvedOrigin
- Deploy X (app) + ResolvedOrigin -> records origin to app document

Because this is not transactional, if the Deploy call fails (e.g. intermittent network issues) you have the charm document but no origin. So if you retry the `juju deploy X` command, we still have to fetch the metadata and re-calculate the resolved origin (which might in an edge case change between running the two `juju deploy` commands) so the `Deploy` can work as expected.

Note that in the future, all these things will happen in a single transaction (and in a dedicated facade) so I don't anticipate any issues. For now, this approach allows us to test the feature without accidentally breaking the existing deploy flow.

apiserver/facades/client/charms/client.go
achilleasa: Indeed, both store APIs support this functionality (although only charmhub supports batch lookups IIRC). The next step forward is to implement a batch metadata lookup method in the repository interface which the new deploy code will use (group requirements by store and make batch lookup requests).

Since this facade will eventually be deprecated and superseded by the new deploy code, I didn't want to spend time (in this PR) to get that working so I opted for this hacky shortcut which has the same effect (with the obvious caveat that we fetch and drop the blob but this is all behind a feature flag so it's fine for now).

api/charmdownloader/client_test.go
achilleasa: I should actually get rid of that file... It doesn't really do anything worth testing

hmlanigan (changes requested):
Found an issue during qa

apiserver/facades/client/charms/client.go
```diff
@@ -380,6 +412,77 @@ func (a *API) addCharmWithAuthorization(args params.AddCharmWithAuth) (params.Ch
...
+	charmURL, err := charm.ParseURL(args.URL)
+	if err != nil {
+		return corecharm.Origin{}, err
+	}
+
+	// Check if a charm doc already exists for this charm URL. If so, the
+	// charm has already been queued for download so this is a no-op. We
+	// still need to return back the recorded origin from where the charm
+	// is to be downloaded.
+	if _, err := a.backendState.Charm(charmURL); err == nil {
```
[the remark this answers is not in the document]
achilleasa: Hmm... Looks like I don't fully understand how the process works :-( 

AFAICT from the original version of the code, charm URLs (the ID in charm documents) are supposed to be unique; the code explicitly checks for the presence of a charm doc and treats AddCharm as a no-op. However, in such a case, the original version of the code would also resolve the origin **again** and return it to the client.

I believe this is wrong. Resolving the origin again might yield a different origin (because the developer released a new version in the meantime) which means that Juju deploys blob B (fetched after resolving the origin to **R0**) but we may record a _potentially different_ origin R' against the application doc (via the subsequent deploy steps).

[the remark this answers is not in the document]
SimonRichardson: So this is known and probably what epochs are intended to solve...?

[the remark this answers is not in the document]
achilleasa: So, is the right ™️ solution here to preserve the original behavior and always resolve the origin?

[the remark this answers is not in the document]
achilleasa: Ok, after chatting with Simon, I know have a better understanding about how charmhub revisions can point to the same blob but in different channels. The above bit was probably correct until https://github.com/juju/juju/pull/13206 landed (the actual introduction of `UploadedCharmOrigin` landed before this one as part of the refactoring work in https://github.com/juju/juju/pull/13191. 

So I will rework the PR to drop `UploadedCharmOrigin` call and ensure we always resolve the origin when the charm is already downloaded. On the bright side, once we transition to server-side bundle expansion, the controller will resolve the origin just once and this re-resolve step will be redundant.

worker/charmdownloader/interfaces.go
```diff
@@ -0,0 +1,25 @@
+// Copyright 2021 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package charmdownloader
+
+import (
+	"github.com/juju/juju/core/watcher"
+	"github.com/juju/names/v4"
```
hmlanigan, line 8: need stanzas

worker/charmdownloader/manifold.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/juju/juju/api/base"
+	client "github.com/juju/juju/api/charmdownloader"
+)
+
+// ManifoldConfig describes the resources used by the charmdownloader worker.
+type ManifoldConfig struct {
+	APICallerName string
+	Logger        Logger
+
+	NewCharmDownloaderAPI func(base.APICaller) CharmDownloaderAPI
```
hmlanigan, line 21: Usually everything in the config is validated.  A comment on why we're not validating NewCharmDownloaderAPI would be helpful.

hmlanigan (approved):
A few small items.  QA went well, including some extra tests.  LGTM

worker/charmdownloader/manifold.go
achilleasa: I changed the validation code to inject the default client if one is not provided.

achilleasa:
!!build!!

worker/charmdownloader/manifold.go
achilleasa: Actually I had to roll this back as the manifold config is not a pointer so I can't patch it while validating. I did add a comment however explaining that it is an optional field and that the default client will be used if no c-tor is specified.

achilleasa:
$$merge$$

(NOTE: check-multi-juju is lying: https://jenkins.juju.canonical.com/job/github-make-check-juju/12594/consoleText)

juanmanuel-tirado:
Is not this the error?

`Generating facade schema...
go install -mod=mod -tags ""  -ldflags "-s -w -extldflags -static -X github.com/juju/juju/version.GitCommit=66e483affb98a0f50b08b843d5028a86920cdc64 -X github.com/juju/juju/version.GitTreeState=clean -X github.com/juju/juju/version.build=" -v $MAIN_PACKAGES
Reading package lists...
Building dependency tree...
Reading state information...
python3 is already the newest version (3.8.2-0ubuntu2).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Using  for docs check.
Traceback (most recent call last):
  File "./scripts/generate-docs.py", line 79, in <module>
    main(sys.argv)
  File "./scripts/generate-docs.py", line 57, in main
    doc_generator = GENERATORS[args[1]]()
  File "/home/jenkins/go/src/github.com/juju/juju/scripts/jujuman.py", line 14, in __init__
    self.version = self._version()
  File "/home/jenkins/go/src/github.com/juju/juju/scripts/jujuman.py", line 25, in _version
    juju_version = self.run_juju('version')
  File "/home/jenkins/go/src/github.com/juju/juju/scripts/jujuman.py", line 22, in run_juju
    return subprocess.check_output(cmd).strip().decode('utf-8')
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 493, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1704, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'juju'`

achilleasa:
@juanmanuel-tirado well, the run that I've linked above seems to be green.... 😕

PR #13261, opened 2021-08-19
Move snap proxy setting to cloud init runcmd rather than bootcmd

wallyworld:
It seems on some machines the snapd seeding process take take a while such that the cloud init boot cmds to set up snap proxies fail.
This PR defers the snap proxy setup to run cmds instead and also add a small script to wait for snapd to be seeded.

## QA steps

bootstrap
juju model-config snap-http-proxy="http://10.0.0.1:8000"
juju add-machine

ensure machine comes up
grep cloudinit-output-log on machine to see
```
+ printf Attempt 1 to wait for snapd to be seeded...\n
Attempt 1 to wait for snapd to be seeded...
+ snap wait core seed.loaded
+ break
+ snap set system proxy.http=http://10.0.0.1:8000
```

## Bug reference

https://bugs.launchpad.net/juju/+bug/1940445

### DIFF
--- a/cloudconfig/cloudinit/cloudinit_ubuntu.go
+++ b/cloudconfig/cloudinit/cloudinit_ubuntu.go
@@ -279,6 +279,23 @@ func (cfg *ubuntuCloudConfig) addRequiredPackages() {
 	}
 }
 
+var waitSnapSeeded = `
+n=1
+while true; do
+
+printf "Attempt $n to wait for snapd to be seeded...\n"
+snap wait core seed.loaded && break
+if [ $n -eq 5 ]; then
+  echo "snapd not initialised"
+  break
+fi
+
+echo "Wait for snapd failed, retrying in 5s"
+sleep 5
+n=$((n+1))
+done
+`[1:]
+
 // Updates proxy settings used when rendering the conf as a script
 func (cfg *ubuntuCloudConfig) updateProxySettings(proxyCfg PackageManagerProxyConfig) error {
 	// Write out the apt proxy settings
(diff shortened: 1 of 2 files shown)


wallyworld:
[snapd init issue](https://trello.com/c/DDFrIicb/1724-snapd-init-issue)

nicolasbock (approved):
I can conform that with this change I cannot reproduce the issue anymore.

cloudconfig/cloudinit/cloudinit_ubuntu.go
```diff
@@ -291,11 +309,13 @@ func (cfg *ubuntuCloudConfig) updateProxySettings(proxyCfg PackageManagerProxyCo
 			filename))
 	}
 
+	once := sync.Once{}
```
hpidcock, line 312: wouldn't a bool be better suited to this?

```
once := false
addWaitSnapSeeded := func() {
    if once {
         return
    }
    cfg.AddRunCmd(waitSnapSeeded)
    once = true
}
```

cloudconfig/cloudinit/cloudinit_ubuntu.go
```diff
@@ -279,6 +280,23 @@ func (cfg *ubuntuCloudConfig) addRequiredPackages() {
 	}
 }
 
+var waitSnapSeeded = `
+n=1
```
hpidcock, line 284: I'm guessing AddRunCmd inserts this into a script/runs these commands in a shell? Just taking note of no #!

hpidcock (approved):
Thank-you, looks good.

cloudconfig/cloudinit/cloudinit_ubuntu.go
hpidcock: just that sync.Once is a bit heavy for something not requiring synchronisation?

cloudconfig/cloudinit/cloudinit_ubuntu.go
wallyworld: yeah, that's how it works, we use a similar construct elsewhere too

cloudconfig/cloudinit/cloudinit_ubuntu.go
wallyworld: i considered doing it this way also but not sure why I went with sync.Once. i'll change to the func approach

wallyworld:
$$merge$$

PR #13277, opened 2021-08-25
[dev] Always resolve origin for already downloaded charms

achilleasa:
This PR reverts a change introduced by https://github.com/juju/juju/pull/13191 where `AddCharm` (via a call to core/charm/downloader), if the charm was already downloaded, would attempt to locate the application document referencing the charm URL and return back the recorded origin.

However, as the same revision can be shared between channels, we should always attempt to resolve the charm origin based on the preferred origin passed by the client to `AddCharm` even if the charm is already fetched.


hmlanigan (commented):
QA with charm revisions is good now - charm revisions with resources is broken as expected.  I'll work on fixing that later today.

PR #13286, opened 2021-08-27
Fix reportHookError blocking when relation hook fails due to missing relation

hpidcock:
Fix reportHookError blocking when relation hook fails due to missing relation.

If the unit has a queued relation hook, restarted while the same relation is forced removed it enters a worker restart loop from the resolver.

## QA steps

- Wait for a unit to idle
- juju_stop_unit
- poke mongo to update unit's unitstates document queueing a relation-* hook with an invalid relation-id.
- juju_start_unit
- unit should fail gracefully and not enter a worker restart loop

## Documentation changes

N/A

## Bug reference

Somewhat related to https://bugs.launchpad.net/charm-ceph-mon/+bug/1940983

### DIFF
--- a/worker/uniter/uniter.go
+++ b/worker/uniter/uniter.go
@@ -941,6 +941,7 @@ func (u *Uniter) reportHookError(hookInfo hook.Info) error {
 	// hook is interrupted (e.g. unit agent crashes), rather than immediately
 	// after attempting a runHookOp.
 	hookName := string(hookInfo.Kind)
+	hookMessage := string(hookInfo.Kind)
 	statusData := map[string]interface{}{}
 	if hookInfo.Kind.IsRelation() {
 		statusData["relation-id"] = hookInfo.RelationId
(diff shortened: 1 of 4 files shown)


hpidcock:
[Fix reportHookError in uniter](https://trello.com/c/rBHN4vQW/1735-fix-reporthookerror-in-uniter)

hpidcock:
!!build!!

worker/uniter/uniter_test.go
```diff
@@ -12,10 +12,12 @@ import (
 	"syscall"
 
 	corecharm "github.com/juju/charm/v8"
+	"github.com/juju/charm/v8/hooks"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	jc "github.com/juju/testing/checkers"
 	gc "gopkg.in/check.v1"
+	"gopkg.in/yaml.v3"
```
wallyworld, line 20: We use yaml.v2 almost everywhere else I think. There appears to be a few places where v3 has crept in. Would be nice to be consistent.

worker/uniter/uniter.go
```diff
@@ -941,20 +941,25 @@ func (u *Uniter) reportHookError(hookInfo hook.Info) error {
...
 		}
 		relationName, err := u.relationStateTracker.Name(hookInfo.RelationId)
-		if err != nil {
+		if errors.IsNotFound(err) {
+			hookMessage = fmt.Sprintf("%s: %v", hookInfo.Kind, err)
+		} else if err != nil {
 			return errors.Trace(err)
+		} else {
+			hookName = fmt.Sprintf("%s-%s", relationName, hookInfo.Kind)
+			hookMessage = hookName
```
wallyworld, line 958: We could even consider making a hook error message for any error and reporting that via setAgentStatus() below. We are inside a method where something by definition has already gone wrong. Should we bounce the agent if there's an error dealing with an error?

wallyworld (approved):

worker/uniter/uniter_test.go
hmlanigan: Weren't there some reasons to not move to yaml v3?  But we needed it in a very specific case.

hpidcock: yeah thanks for that, this was just an autoimport issue

hpidcock:
$$merge$$

hpidcock:
$$merge$$

PR #13289, opened 2021-08-27
[2.9] Ignore fan subnets when querying equinix API

achilleasa:
When the provider prepares a new instance for which space constraints have been specified, it queries the equinix API for each subnet that Juju specifies in the StartInstance argument.

Before creating the instance (device in equinix terminology) the provider queries the equinix API to resolve each requested Juju subnet ID into a subnet CIDR which is then passed to the equinix create device API. However, the current provider implementation does not filter FAN subnets (FAN is automatically enabled on equinix metal) so the resolution for the FAN subnets fails (a 404 is returned back from the API) and the instance fails to be provisioned.

## QA steps

Deploy a charm on equinix metal using space constraints

## Bug reference
https://bugs.launchpad.net/juju/+bug/1941912

### DIFF
--- a/provider/equinix/environ.go
+++ b/provider/equinix/environ.go
@@ -19,6 +19,7 @@ import (
 	"github.com/juju/juju/cloudconfig/providerinit"
 	"github.com/juju/juju/core/constraints"
 	"github.com/juju/juju/core/instance"
+	"github.com/juju/juju/core/network"
 	"github.com/juju/juju/core/series"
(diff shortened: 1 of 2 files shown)


juanmanuel-tirado:
!!build!!

juanmanuel-tirado (approved):
LG2M

juanmanuel-tirado:
$$merge$$

PR #13290, opened 2021-08-30
Remove app status history when destorying an app; add support for apps in show-status-log

wallyworld:
Investigating a CMR issue, the following problem was observed.
An offer is consumed and the SAAS status entry correctly reflects the app status in the offering model, eg Active.
Remove the SAAS using `juju remove-saas` and consume again. Status is now showing "unknown". But it will be correctly updated next time the offer application changes after that.

The issue is that deleting the SAAS leaves the status history behind. And the next time the offer is consumed, the new status is the same as the last historical value and no update is done, so it is left as unknown.

The issue of not removing app status history also affect local applications. So this PR fixes this for both local and SAAS apps. The implementation is as per what is done for erasing unit history (do it in the Done() method it the destroy operation).

The PR also tweaks the show-status-log CLI tool to support both `application` and `saas` types for printing status history.

As a driveby, fix a few lint issues as well.

## QA steps

deploy 2 models and offer mariadb in one of them
in the other model, juju consume mariadb
then
```
juju show-status-log --type saas mariadb

Time                        Type  Status   Message
30 Aug 2021 14:29:44+10:00  saas  active   ready
```
remove the saas
```
juju remove-saas mariadb
juju show-status-log --type saas mariadb
ERROR while processing the request: fetching status history for "application-mariadb": remote application "mariadb" not found
```
Consume the offer again
```
juju consume othermodel.mariadb
juju show-status-log --type saas mariadb

Time                        Type  Status   Message
30 Aug 2021 14:31:28+10:00  saas  active   ready
```

Also run juju status to see the SAAS status is "Active".

Previously juju status would have shown "unknown".

### DIFF
--- a/api/watcher/watcher.go
+++ b/api/watcher/watcher.go
@@ -603,8 +603,8 @@ func (w *relationStatusWatcher) loop(initialChanges []params.RelationLifeSuspend
 				// at this point, so just return.
 				return nil
 			}
-			new := copyChanges(data.(*params.RelationLifeSuspendedStatusWatchResult).Changes)
-			changes = w.mergeChanges(changes, new)
+			newChanges := copyChanges(data.(*params.RelationLifeSuspendedStatusWatchResult).Changes)
+			changes = w.mergeChanges(changes, newChanges)
 			out = w.out
 		case out <- changes:
 			out = nil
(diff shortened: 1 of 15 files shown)


hpidcock (approved):
Thanks, good to see this fixed.

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
[incorrect saas status](https://trello.com/c/689UdmQc/1736-incorrect-saas-status)

PR #13293, opened 2021-09-01
Record secret rotation config and add watcher infrastructure

wallyworld:
A new watcher is added to notify of changes to secrets rotation info. The watcher triggers when the rotation interval changes; the payload is the rotation interval plus last time rotated. A subsequent PR will add a worker to use this information to fire secret rotation hook events to notify charms that a secret needs to be rotated.

The rotation details are stored in a new collection which the watcher sits on top of.
Secrets get a new "owner" attribute - this is akin to the entity for which the secret is created and which ties to the life cycle of the secret. It may be an application or a unit or a model or a controller (but is not fully wired up yet). Only the owner of a secret may watch for changes. The idea here is that a unit agent will watch for all secrets owned by its charm so that it may run the rotate-secret hook when needed. A future PR will also make only owners (or those granted permission) able to update the secret etc. 

The key components are:
- state watcher
- new bespoke watcher facade and api params
- new method on the secrets manager facade to expose the watcher
- associated api glue and mocks

## QA steps

To test, I added a little bespoke code to the unit agent to watch for secret rotation changes using the new secrets manager API in this PR. The changes were logged.

From a hook context
secret-create -> receive event
secret-update --rotate <interval> -> receive event with interval, check "secretRotate" collection has a record
secret-update --rotate 0 -> receive event with interval set to 0, check "secretRotate" collection has no record


wallyworld:
[secret rotation watcher](https://trello.com/c/H76hSUAl/1731-secret-rotation-watcher)

state/secrets.go
```diff
@@ -313,3 +335,222 @@ func (s *secretsStore) ListSecrets(filter SecretsFilter) ([]*secrets.SecretMetad
...
+		}
+		details = append(details, corewatcher.RotateSecretChange{
+			ID:             ID,
+			URL:            knownDetails.URL,
+			RotateInterval: 0,
+		})
+		return details, nil
+	}
+	if doc.TxnRevno > knownDetails.txnRevNo {
+		URL, err := secrets.ParseURL(doc.URL)
```
hpidcock, line 511: `url`

hpidcock (approved):
Some changes but LGTM

wallyworld:
$$merge$$

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #13294, opened 2021-09-01
Fix DownloadAndStore where charm is already downloaded.

hmlanigan:
If the charm has already been downloaded, the origin returned from DownloadAndStore must include an origin with an ID and Hash.  Otherwise charm upgrades and deploying with resources will fail.

## QA steps

```sh
$ juju deploy juju-qa-test
# deploy the same charm with a new name.
$ juju deploy juju-qa-test juju-qa-test-again

# verify ID/hash in db.
juju:PRIMARY> db.applications.find({},{"charm-origin":1}).pretty()
```

### DIFF
--- a/core/charm/downloader/downloader.go
+++ b/core/charm/downloader/downloader.go
@@ -6,6 +6,7 @@ package downloader
 import (
 	"io"
 	"io/ioutil"
+	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


hmlanigan:
This PR has been consumed by #13295

PR #13298, opened 2021-09-02
[dev] Implement batch charm lookup

achilleasa:
This PR builds on top of #13260 and is part of the work for implementing server-side bundle expansion.

It expands the charm repository interface with a new method (`GetEssentialMetadata`). The new method provides a mechanism for querying the essential (minimum set of metadata for deploying a charm) metadata for a list of charms. 

The key idea here is to have the `AddCharm` facade make a lightweight API call (also leveraging charmhub's support for batch requests) to the store and only fetch the metadata (as opposed to fetching the full charm blob) needed to start a deployment (metadata.yaml, manifest.yaml, config.yaml and lxd-profile.yaml), persist it to state and backfill the remaining bits of metadata (e.g. actions, metrics etc.) when the gets asynchronously downloaded (and before the uniter can retrieve the charm blob). The end goal is to reduce the _perceived_  execution time of `juju deploy` (and improve it even further when server-side bundle expansion gets implemented) for the end-users.

The _charmstore_ implementation is pretty straight forward and requires two API calls per charm. The first call gets all the information we need _with the exception of the lxd profile_. For the latter we need to perform a second call to grab the profile (if present) from the charm archive.

The _charmhub_ implementation is, for the time being, _emulating_ the required behavior. It groups together lookup requests based on the provided macaroon values (note: charmhub does not support macaroons so all requests get lumped up in a single group), then downloading the complete charm for each request and extracts the metadata we need. 

The charmhub API currently _only_ allows us to get the contents of the `metadata.yaml` file (via the /refresh endpoint). We have put together a [spec](https://docs.google.com/document/d/1CPwwYipBV03cBq3cVO2BRW614XZEJ0Phv0tziT3xbVw/edit#) for extending the list of fields we can get out of the refresh endpoint so we can pull the additional metadata we need. Once the required changes to the charmhub API become available to us we will amend the implementation to use it.

Note that the above methods are not currently used but will be wired in in a separate PR.

## QA steps

Just check that juju compiles; the changes are not currently used.

### DIFF
--- a/core/charm/repository/charmstore.go
+++ b/core/charm/repository/charmstore.go
@@ -4,6 +4,7 @@
 package repository
 
 import (
+	"io"
 	"net/url"
 
 	"github.com/go-macaroon-bakery/macaroon-bakery/v3/httpbakery"
@@ -11,6 +12,7 @@ import (
 	charmresource "github.com/juju/charm/v9/resource"
 	"github.com/juju/charmrepo/v7"
 	"github.com/juju/charmrepo/v7/csclient"
+	"github.com/juju/charmrepo/v7/csclient/params"
 	csparams "github.com/juju/charmrepo/v7/csclient/params"
 	"github.com/juju/errors"
 	"gopkg.in/macaroon.v2"
@@ -23,7 +25,9 @@ import (
 // CharmStoreClient describes the API exposed by the charmstore client.
 type CharmStoreClient interface {
 	Get(charmURL *charm.URL, archivePath string) (*charm.CharmArchive, error)
-	ResolveWithPreferredChannel(*charm.URL, csparams.Channel) (*charm.URL, csparams.Channel, []string, error)
+	ResolveWithPreferredChannel(charmURL *charm.URL, channel csparams.Channel) (*charm.URL, csparams.Channel, []string, error)
+	Meta(*charm.URL, interface{}) (*charm.URL, error)
+	GetFileFromArchive(charmURL *charm.URL, filename string) (io.ReadCloser, error)
 }
 
 // CharmStoreRepository provides an API for charm-related operations using charmstore.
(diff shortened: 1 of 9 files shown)


core/charm/repository/charmstore.go
```diff
@@ -135,6 +139,47 @@ func (c *CharmStoreRepository) ListResources(charmURL *charm.URL, _ corecharm.Or
...
+func (c *CharmStoreRepository) GetEssentialMetadata(reqs ...corecharm.MetadataRequest) ([]corecharm.EssentialMetadata, error) {
+	var res = make([]corecharm.EssentialMetadata, len(reqs))
+
+	for reqIdx, req := range reqs {
+		// NOTE(achilleas): due to the way that the charmstore client
+		// was originally implemented we unfortunately need to create a
+		// new client per request.
+		channel := csparams.Channel(req.Origin.Channel.Risk)
+		client, err := c.clientFactory(c.charmstoreURL, channel, req.Macaroons)
+
```
juanmanuel-tirado, line 154: Shouldn't `err` being checked here?

core/charm/repository/charmstore.go
```diff
@@ -135,6 +139,47 @@ func (c *CharmStoreRepository) ListResources(charmURL *charm.URL, _ corecharm.Or
...
+			// It's fine if the charm does not provide an lxd profile
+			if errors.Cause(fetchErr) == params.ErrNotFound {
+				continue
+			}
+
+			return nil, errors.Annotatef(err, "retrieving metadata for %q", req.CharmURL)
+		}
+
+		res[reqIdx].LXDProfile, err = charm.ReadLXDProfile(profileReader)
+		_ = profileReader.Close()
```
juanmanuel-tirado, line 174: The return from the method should be used at least to inform the user that something went wrong there.

juanmanuel-tirado (commented):
Check my minor comments

core/charm/repository/charmstore.go
achilleasa: `charm.ReadLXDProfile` is expected to consume the entire profile payload and either return back an error or return a profile instance. In this scenario, failing to close the reader (which in reality is an http.Response.Body) shouldn't really matter as the actual goal of the method (grab the metadata) has already been met at this point. 

We could however catch the error and log out a warning if you think it's important. Thoughts?

juanmanuel-tirado (approved):
LG2M

achilleasa:
!!build!!

achilleasa:
$$merge$$

PR #13299, opened 2021-09-02
Fix /etc/profile.d/juju-introspection.sh not being POSIX-shell compatible

kot0dama:
/etc/profile.d/juju-introspection.sh contains a few bash arrays, breaking POSIX shell compatibility and at least some packages post-installation as it is loaded as a profile script.

For example, on a model deployed with Juju 2.9.11, installing spamassassin yields a _Syntax error_ as spamassassin post-installation script attempts to run sa-update as debian-spamd user, having /bin/sh as shell.

The change includes variables name changes (and a comment) to avoid future updates breaking POSIX compatibility.
Feel free to update the comment to whatever you see fit.

## QA steps

* Check it builds properly
* Deploy a model with spamassassin installation (eg. charm cs:mailman3-core) and check it doesn't fail during SA install.

## Bug reference

Fixes https://bugs.launchpad.net/juju/+bug/1942430
Probable iteration of https://bugs.launchpad.net/juju/+bug/1770437

### DIFF
--- a/worker/introspection/script.go
+++ b/worker/introspection/script.go
@@ -13,11 +13,11 @@ import (
 
 var (
 	// ProfileDir is the directory where the profile script is written.
-	ProfileDir        = "/etc/profile.d"
-	bashFuncsFilename = "juju-introspection.sh"
+	ProfileDir         = "/etc/profile.d"
+	shellFuncsFilename = "juju-introspection.sh"
 )
 
-// WriteProfileFunctions writes the bashFuncs below to a file in the
+// WriteProfileFunctions writes the shellFuncs below to a file in the
 // /etc/profile.d directory so all bash terminals can easily access the
 // introspection worker.
 func WriteProfileFunctions(profileDir string) error {
@@ -26,17 +26,19 @@ func WriteProfileFunctions(profileDir string) error {
 		return nil
 	}
 	filename := profileFilename(profileDir)
-	if err := ioutil.WriteFile(filename, []byte(bashFuncs), 0644); err != nil {
+	if err := ioutil.WriteFile(filename, []byte(shellFuncs), 0644); err != nil {
 		return errors.Annotate(err, "writing introspection bash funcs")
 	}
 	return nil
 }
 
 func profileFilename(profileDir string) string {
-	return path.Join(profileDir, bashFuncsFilename)
+	return path.Join(profileDir, shellFuncsFilename)
 }
 
-const bashFuncs = `
+// WARNING: This code MUST be compatible with all POSIX shells including
+// /bin/sh and MUST NOT include bash-isms.
+const shellFuncs = `
 juju_agent_call () {
   local agent=$1
   shift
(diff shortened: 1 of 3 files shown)


jameinel:
Sorry about the bot spam.
However, the code that you've defined for the POSIX mode is not actually compatible. At least in my testing I see:
```console
$ t () {
> arr=("$@")
> local -a args
> for i in "${arr[@]}"; do
>   args+=("unit=$i")
> done
> python3 -c 'import sys; print(sys.argv)' "${args[@]}"
> }
$ tt () {
> args=""
> for i in "$@"; do
>   args="$args unit=$i"
> done
> python3 -c 'import sys; print(sys.argv)' "$args"
> }
$ t a b
['-c', 'unit=a', 'unit=b']
$ tt a b
['-c', ' unit=a unit=b']
```

In our previous bash script you end up with 2 arguments with 'unit=X', while in the new form you end up with both an opening space, and a single argument that happens to have a space (and opens with an extra space as well)

I wonder if a better solution is not to try and make it Posix compliant, but just make it clear that it is `#!/bin/bash` rather than `#!/bin/sh`

worker/introspection/script.go
```diff
@@ -127,12 +129,11 @@ juju_stop_unit () {
...
-  arr=("$@")
-  local -a args
-  for i in "${arr[@]}"; do
-    args+=("unit=$i")
+  args=""
+  for i in "$@"; do
+    args="$args unit=$i"
   done
-  juju_agent --post units action=stop "${args[@]}"
+  juju_agent --post units action=stop "$args"
```
jameinel, line 136: This doesn't actually have the same semantic.
However, maybe if we just avoid the "" around $args. It does look like:
```
$ tt () { args=""; for i in "$@"; do   args="$args unit=$i"; done; python3 -c 'import sys; print(sys.argv)' $args; }
$ tt a b
['-c', 'unit=a', 'unit=b']
```
Looks more like what we want to happen.

worker/introspection/script.go
```diff
@@ -141,12 +142,11 @@ juju_start_unit () {
...
-  arr=("$@")
-  local -a args
-  for i in "${arr[@]}"; do
-    args+=("unit=$i")
+  args=""
+  for i in "$@"; do
+    args="$args unit=$i"
   done
-  juju_agent --post units action=start "${args[@]}"
+  juju_agent --post units action=start "$args"
```
jameinel, line 149: same here

jameinel (changes requested):

kot0dama:
I see, indeed you're right removing the quotes around $args fixes it.
 
> I wonder if a better solution is not to try and make it Posix compliant, but just make it clear that it is `#!/bin/bash` rather than `#!/bin/sh`

About this, that won't cut it as it is not executed but source-d by the shell of the user, so every file in /etc/profile.d/ must be POSIX compatible. Another solution would be to add this:
```
# Not bash or zsh?
[ -n "${BASH_VERSION:-}" -o -n "${ZSH_VERSION:-}" ] || return 0
```

Anyways, I'll fix the quote issue and force push again.

kot0dama:
Here it is, let me know if that's better.
Thank you

jameinel (approved):

jameinel:
$$merge$$

jameinel:
I accidentally submitted it for merge before seeing that it was targeting Develop rather than the 2.9 branch. If it doesn't go through, I would appreciate if you would rebase targeting 2.9 and update the target of your PR, but otherwise we can backport this ourselves. Thank you for the POSIX fixes.

PR #13303, opened 2021-09-03
Add extra tracing to relation related workflow

wallyworld:
To aid in debugging relation hook issues, extra (mostly) trace logging has been added to various workers which deal with relations. The bulk of the changes are in the uniter relation handling. Also, in the state package, a child "relationunits" logger is now used to that tracing can be turned on for just that bit and not the rest of state.

As an additional debugging aid, we log a warning if the relation joined processing is wedged waiting for a unit watcher event to tirgger.

For a uniter, a few other changes were made. When a relation is removed, we were not deleting ids from all required maps. And the relation resolver is made to be like the relation created resolver in that it doesn't do any processing if there's another operation pending. Also, the method to update the state in the controller when a hook is committed was unnecessarily returning an error for the relation-broken hook when it is never called for that hook anyway.

Also some drive by lint fixes.

## QA steps

deploy and relate 2 applications
show-status-log to check for relation created, joined, changed hooks
remove the relation and check for departed, broken hooks
add back the relation, rinse and repeat

### DIFF
--- a/api/watcher/watcher.go
+++ b/api/watcher/watcher.go
@@ -9,6 +9,7 @@ import (
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 	"github.com/juju/names/v4"
+	"github.com/kr/pretty"
 	"gopkg.in/tomb.v2"
 
 	"github.com/juju/juju/api/base"
@@ -247,6 +248,7 @@ func (w *stringsWatcher) Changes() watcher.StringsChannel {
 type relationUnitsWatcher struct {
 	commonWatcher
 	caller                 base.APICaller
+	logger                 loggo.Logger
 	relationUnitsWatcherId string
 	out                    chan watcher.RelationUnitsChange
 }
@@ -254,6 +256,7 @@ type relationUnitsWatcher struct {
 func NewRelationUnitsWatcher(caller base.APICaller, result params.RelationUnitsWatchResult) watcher.RelationUnitsWatcher {
 	w := &relationUnitsWatcher{
 		caller:                 caller,
+		logger:                 logger.Child("relationunits"),
 		relationUnitsWatcherId: result.RelationUnitsWatcherId,
 		out:                    make(chan watcher.RelationUnitsChange),
 	}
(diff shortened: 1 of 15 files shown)


wallyworld:
[extra relation debugging](https://trello.com/c/RXUJNmdB/1740-extra-relation-debugging)

api/watcher/watcher.go
```diff
@@ -365,6 +373,9 @@ func (w *remoteRelationWatcher) loop(initialChange params.RemoteRelationChangeEv
 		select {
 		// Send out the initial event or subsequent change.
 		case w.out <- change:
+			if w.logger.IsTraceEnabled() {
+				w.logger.Tracef("sent remote relation change %# v", pretty.Formatter(change))
```
hpidcock, line 377: I'm guessing the IsTraceEnabled for the pretty formatter? Because the only cost of creating the formatter is a reflec.ValueOf which appears fairly cheap.

worker/uniter/relation/statetracker.go
```diff
@@ -123,6 +126,9 @@ func (r *relationStateTracker) loadInitialState() error {
 		r.relationCreated[rel.Id()] = true
 	}
 
+	if r.logger.IsTraceEnabled() {
+		r.logger.Tracef("initialising relation state tracker: %# v", pretty.Formatter(r.stateMgr.(*stateManager).relationState))
```
hpidcock, line 130: could this be something other than *stateManager or be untyped nil?

worker/uniter/relation/statetracker.go
```diff
@@ -181,6 +195,8 @@ func (r *relationStateTracker) joinRelation(rel Relation) (err error) {
 			// Should this be a different error? e.g. resolver.ErrAborted, that
 			// Loop translates into ErrLoopAborted?
 			return resolver.ErrLoopAborted
+		case <-time.After(30 * time.Second):
```
hpidcock, line 198: I don't imagine this loop will spin much but this could blow out to a lot of leaked timers+channels.

hpidcock (approved):

api/watcher/watcher.go
wallyworld: I'm following the conventions used elsewhere which wrap the pretty trace output. I guess the thinking is that every bit adds up in a loaded system.

worker/uniter/relation/statetracker.go
wallyworld: No, stateMgr is always set to a concrete value

worker/uniter/relation/statetracker.go
wallyworld: The expectation is that the watcher *always* triggers with the initial event immediately.
There's no expectation that this will ever be timed out unless there's a terrible bug in the watcher. Nonetheless, I've tweaked to return an error if the timeout ever happens, which will bounce the worker and things can re-sync.

wallyworld:
$$merge$$

PR #13306, opened 2021-09-03
Fix introspection posix shell

jameinel:
This is a rebase of https://github.com/juju/juju/pull/13299 targeting the 2.9 branch.
It changes the `/etc/profile.d/juju-introspection.sh` which had some bash-isms that break if your user's shell is not bash. (dash, or zsh, etc). We can't use bashisms because all shells read /etc/profile.d.

## QA steps

See the original PR.

## Documentation changes

None.

## Bug reference

https://bugs.launchpad.net/juju/+bug/1770437

### DIFF
--- a/worker/introspection/script.go
+++ b/worker/introspection/script.go
@@ -13,11 +13,11 @@ import (
 
 var (
 	// ProfileDir is the directory where the profile script is written.
-	ProfileDir        = "/etc/profile.d"
-	bashFuncsFilename = "juju-introspection.sh"
+	ProfileDir         = "/etc/profile.d"
+	shellFuncsFilename = "juju-introspection.sh"
 )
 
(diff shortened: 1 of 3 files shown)


worker/introspection/script.go
```diff
@@ -141,12 +142,11 @@ juju_start_unit () {
     echo "usage: juju_start_unit <unit-name> [<unit-name>...]"
     return 1
   fi
-  arr=("$@")
-  local -a args
-  for i in "${arr[@]}"; do
-    args+=("unit=$i")
+  args=""
```
achilleasa, line 145: We should probably extract this in a func

achilleasa (approved):

SimonRichardson:
!!build!!

juanmanuel-tirado:
!!build!!

second try

PR #13310, opened 2021-09-06
Github action for windows

SimonRichardson:
The following adds github windows action for client only tests. As we're
only testing clients only and not workloads, we don't _really_ need the
correct version of mongo here. All full end to end tests should be
officially identified and either moved to a better location or
re-written to prevent full stack testing. The CLI shouldn't be
exercising the full stack, that if anything should be apiserver and
down.

## QA steps

Github actions pass.

### DIFF
--- a/.github/workflows/client-tests.yml
+++ b/.github/workflows/client-tests.yml
@@ -26,13 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # macOS-latest is disabled because there is an issue of the tests timing
-        # out. No effort has been done to work out why they currently timeout
-        # on macOS, but we should investigate that to improve our coverage on
-        # other clients.
-        # To turn on macOS, just update the os to include it.
-        # os: [ubuntu-latest, macOS-latest, windows-latest]
-        os: [ubuntu-latest, macOS-latest]
+        os: [ubuntu-latest, macOS-latest, windows-latest]
 
     steps:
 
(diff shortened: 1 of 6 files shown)


.github/workflows/client-tests.yml
```diff
@@ -42,6 +36,31 @@ jobs:
         go-version: 1.14
```
juanmanuel-tirado, line 36: Could we run 1.17 here?

SimonRichardson: This is go version per branch. So 2.9 is 1.14, develop is 1.17, once we move this to develop we can then swap it over :)

juanmanuel-tirado: Perfect.

.github/workflows/client-tests.yml
```diff
@@ -42,6 +36,34 @@ jobs:
...
+        func main() {
+          s, err := series.HostSeries()
+          if err != nil {
+            log.Fatal(err)
+          }
+
+          fmt.Printf("OS: %q\nSeries: %q\n", series.MustOSFromSeries(s).String(), s)
+        }
+        EOF
+        go mod init main
```
hpidcock, line 63: 😏

.github/workflows/client-tests.yml
```diff
@@ -85,19 +95,19 @@ jobs:
...
         go test -v ./cmd/juju/... -check.v -coverprofile=coverage.txt -covermode=atomic
         go test -v ./cmd/plugins/... -check.v
       shell: bash
+
+    - name: "Test client: macOS && windows"
+      if: (matrix.os == 'macOS-latest' || matrix.os == 'windows-latest')
+      run: |
+        # There is a concurrency issue with macos setup for the "./cmd/juju/..." packages.
+        # So we have to limit amount of used CPUs and therefore parallelization
+        GOARCH=amd64 CGO_ENABLED=0 go test -v -p 1 ./cmd/juju/... -check.v
```
hpidcock, line 111: yeah this seems nice

juanmanuel-tirado:
!!build!!

SimonRichardson:
@hpidcock any thoughts on why windows is different for GH Actions. I can't get mongo to play nicely at all. It would be nice if we could remove the jenkins windows box with a GH one.

PR #13311, opened 2021-09-07
Be more lenient if the charm dir is deleted after a relation is joined

wallyworld:
There's an intermittent issue that has been reported where a k8s unit is removed / replaced and the charm directory is removed and the unit agent tries to run a hook and fails. The hook is typically a relation tear down hook like relation-broken. Given the nature of the issue, this PR makes the unit agent more robust rather than provide a definitive root case fix (which is yet to be fully identified).

The only reason the unit agent reads the charm dir while running a relation hook is to check that the charm supports the relation endpoint. This is only relevant when a relation is first created. So the logic is tweaked to not do the check every hook, but just once as the relation is being created. And if the directory is missing, a warning is logged at that point rather than error out and bounce the agent - downstream processing in the agent will error if needed. For operator charms, since the issue is very intermittent, a tweak is also done to remove the directory last, after the dead unit is removed from state - this shouldn't matter but it doesn't hurt.

Also a few lint fixes and a bit of extra logging.

## QA steps

deploy and relate 2 k8s charms
remove and add a relation
check show-status-log to ensure the created, joined, departed, broken hooks have run

## Bug reference

https://bugs.launchpad.net/juju/+bug/1882600

### DIFF
--- a/worker/uniter/relation/statetracker.go
+++ b/worker/uniter/relation/statetracker.go
@@ -4,6 +4,7 @@
 package relation
 
 import (
+	"os"
 	"time"
 
 	"github.com/juju/charm/v8"
@@ -105,7 +106,7 @@ func (r *relationStateTracker) loadInitialState() error {
 
 	// Keep the relations ordered for reliable testing.
 	var orderedIds []int
-	activeRelations := make(map[int]Relation)
+	isScopeRelations := make(map[int]Relation)
 	relationSuspended := make(map[int]bool)
 	for _, rs := range relationStatus {
 		if !rs.InScope {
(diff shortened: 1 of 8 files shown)


wallyworld:
[missing charm directory](https://trello.com/c/9FWtt6cp/1741-missing-charm-directory)

worker/uniter/relation/statetracker.go
```diff
@@ -265,25 +267,30 @@ func (r *relationStateTracker) SynchronizeScopes(remote remotestate.Snapshot) er
...
 		}
+		// Keep track of the remote application
+		r.remoteAppName[id] = rel.OtherApplication()
+
+		// Make sure we ignore relations not implemented by the unit's charm.
+		if !r.RelationCreated(id) {
+			if charmSpec == nil {
+				charmSpec, err = charm.ReadCharmDir(r.charmDir)
+				if err != nil && !os.IsNotExist(errors.Cause(err)) {
+					return errors.Trace(err)
```
hpidcock, line 286: Might be good to add a log message here saying charm deleted skipping x hook

hpidcock (approved):

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #13312, opened 2021-09-07
Remove unnecessary txn watcher started channel/topic

wallyworld:
As seen in the linked bug, there can be a panic if the txn watcher worker bounces and closes a "watcherStarted" channel twice. However, that channel and associated pubsub topic is unnecessary in the first place, so this PR removes them both.

## QA steps

bootstrap and deploy a couple of charms and relate

## Bug reference

https://bugs.launchpad.net/juju/+bug/1942421

### DIFF
--- a/state/pool.go
+++ b/state/pool.go
@@ -113,15 +113,12 @@ type StatePool struct {
 	// hub is used to pass the transaction changes from the TxnWatcher
 	// to the various HubWatchers that are used in each state object created
 	// by the state pool.
-	hub        *pubsub.SimpleHub
-	hubUnsubFn func()
+	hub *pubsub.SimpleHub
 
 	// watcherRunner makes sure the TxnWatcher stays running.
 	watcherRunner *worker.Runner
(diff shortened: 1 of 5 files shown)


wallyworld:
[migration critical bug](https://trello.com/c/CpONdDEI/1743-migration-critical-bug)

state/pool.go
```diff
@@ -181,9 +177,6 @@ func OpenStatePool(args OpenParams) (_ *StatePool, err error) {
 		RestartDelay: time.Second,
 		Clock:        args.Clock,
 	})
-	pool.hubUnsubFn = pool.hub.Subscribe(watcher.TxnWatcherStarting, func(string, interface{}) {
```
SimonRichardson, line 184: I suspect this was caused by the fact we weren't catching every location that the unsubscribe could have been called. Either way, I'm happy that we've removed this, as it's not used. Seems like originally it was created because they could.

SimonRichardson (approved):

wallyworld:
$$merge$$

PR #13313, opened 2021-09-07
lxd: Update IsLXDNotFound to deal with Go HTTP spelling

stgraber:
LXD is now using Go's native HTTP error which comes with different
capitalization, so update to support both.

In the future, it'd be preferable for Juju to gain support for LXD's
StatusError type which allows for direct comparison with HTTP errors as
well as retrieving the HTTP error code directly.

In general, string matching on errors is a very bad idea, and the LXD team
would like to get notified when that's necessary so we can come up with
an alternative.

## QA steps

Unit tests pass.

### DIFF
--- a/container/lxd/server.go
+++ b/container/lxd/server.go
@@ -319,5 +319,5 @@ func (s *Server) SupportedArches() []string {
 // IsLXDNotFound checks if an error from the LXD API indicates that a requested
 // entity was not found.
 func IsLXDNotFound(err error) bool {
-	return err != nil && err.Error() == "not found"
+	return err != nil && (err.Error() == "not found" || err.Error() == "Not Found")
 }


SimonRichardson:
!!build!!

stgraber:
Changed to my @canonical.com to avoid a bug in the CLA check...

As a Canonical employee (member of https://github.com/canonical), the CLA check should consider the contribution valid even if sent from my @ubuntu.com. It's a simple org check similar to that done for the CLA signers.

container/lxd/server.go
```diff
@@ -319,5 +319,5 @@ func (s *Server) SupportedArches() []string {
 // IsLXDNotFound checks if an error from the LXD API indicates that a requested
 // entity was not found.
 func IsLXDNotFound(err error) bool {
-	return err != nil && err.Error() == "not found"
+	return err != nil && (err.Error() == "not found" || err.Error() == "Not Found")
```
jameinel, line 322: I'm ok with this as a stop gap, but it sounds like Stephane already has typed errors that we could be using. Is there some sort of:
lxderr, ok := err.(lxd.Error)
that we should be using instead of string comparison here?
Or even
`strings.Lower(err.Error())` seems better than 2 different capitalizations.

jameinel (approved):

stgraber:
You should be able to use `api.StatusErrorMatch(err, 404)` with the current LXD client. Or can use the function without the `404` to jut get back the HTTP status code and whether you're indeed dealing with a StatusError.

https://github.com/lxc/lxd/blob/master/shared/api/error.go#L37

wallyworld:
Thanks for the fix. I've done a new PR #13315 which targets 2.9 branch and which uses the `api.StatusErrorMatch()` helper

PR #13318, opened 2021-09-08
Add test to ec2 + ensure ec2 SecurityGroup mutates use only group id

hpidcock:
Follow up to #13316 to correct testing of security groups and also handle revoke/delete of ec2 security groups.

## QA steps

- Bootstrap to aws with non-default VPC id
- Deploy charm
- Expose app
- Check sg has opened the ports
- Delete app check sg is deleted

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1942948


wallyworld (approved):
We should probably enhance the relevant CI test as well.

hpidcock:
$$merge$$

PR #13336, opened 2021-09-16
Fix pod-spec uniter exits on pending action op when remote caas container died

hpidcock:
Fixes issue where the uniter gets into a worker retry loop due to a pending run-action op but the remote caas container has gone away/upgraded. 

## QA steps

Get the workload pod of a pod-spec application in a CrashLoopBackOff
Run an action and watch the unit worker/caasoperator for:
```
application-coredns: 11:28:27 ERROR juju.worker.caasoperator.runner exited "coredns/2": unknown operation kind run-action
```

## Documentation changes

N/A

## Bug reference

https://bugs.launchpad.net/juju/+bug/1943776

### DIFF
--- a/worker/uniter/actions/resolver_test.go
+++ b/worker/uniter/actions/resolver_test.go
@@ -10,6 +10,7 @@ import (
 	gc "gopkg.in/check.v1"
 
 	"github.com/juju/juju/worker/uniter/actions"
+	"github.com/juju/juju/worker/uniter/hook"
 	"github.com/juju/juju/worker/uniter/operation"
 	"github.com/juju/juju/worker/uniter/remotestate"
 	"github.com/juju/juju/worker/uniter/resolver"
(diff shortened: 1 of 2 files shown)


worker/uniter/actions/resolver_test.go
```diff
@@ -190,3 +240,12 @@ type mockFailOp struct {
 func (op *mockFailOp) String() string {
 	return op.name
 }
+
+type mockSkipHookOp struct {
```
wallyworld, line 244: I'm curious why we need to create a new skip hook mock when we already have code which creates a skip hook op. I wonder how that was tested.
```
	case operation.RunAction:
		if localState.Hook != nil {
			r.logger.Infof("found incomplete action %v; ignoring", localState.ActionId)
			r.logger.Infof("recommitting prior %q hook", localState.Hook.Kind)
			return opFactory.NewSkipHook(*localState.Hook)
		}
```

wallyworld (approved):

worker/uniter/actions/resolver_test.go
hpidcock: I guess it wasn't tested.

hpidcock: Added a test for the existing behaviour

hpidcock:
$$merge$$

PR #13337, opened 2021-09-16
Filter on offer uuid when processing offer relations on removal of an offer

wallyworld:
When removing an offer, we build up a list of txn ops to apply. If there's active relations to the offer, we return `ErrTransientFailure` to allow cleanup of those relations to complete. Except that we weren't matching on offer uuid and so if the offer were one of many to a given application, the txn logic would fail with a "state changing too quickly" error.

This PR simply adds the missing check, and also does some error processing in the operation `Done` method to give some extra context if there is a contention error.

## QA steps

deploy an application like mysql
create 2 offers, eg
juju offer mysql:db offer1
juju offer mysql:db-admin offer2

deploy mediawiki in another model and relate to the mysql offer1

remove the offer2
(this would have previously given an error)

## Bug reference

https://bugs.launchpad.net/juju/+bug/1873472

### DIFF
--- a/state/applicationoffers.go
+++ b/state/applicationoffers.go
@@ -158,7 +158,7 @@ func (s *applicationOffers) AllApplicationOffers() (offers []*crossmodel.Applica
 // RemoveOfferOperation returns a model operation that will allow relation to leave scope.
 func (s *applicationOffers) RemoveOfferOperation(offerName string, force bool) *RemoveOfferOperation {
 	return &RemoveOfferOperation{
-		offers:          &applicationOffers{s.st},
+		offerStore:      &applicationOffers{s.st},
 		offerName:       offerName,
 		ForcedOperation: ForcedOperation{Force: force},
 	}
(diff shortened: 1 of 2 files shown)


wallyworld:
[remove one offer of many on a given app](https://trello.com/c/TgPBr9A8/1766-remove-one-offer-of-many-on-a-given-app)

state/applicationoffers.go
```diff
@@ -208,6 +208,15 @@ func (op *RemoveOfferOperation) Build(attempt int) ([]txn.Op, error) {
 func (op *RemoveOfferOperation) Done(err error) error {
 	if err != nil {
 		if !op.Force {
+			if errors.Cause(err) == jujutxn.ErrExcessiveContention {
+				relCount, err := op.countOfferRelations(op.offer)
+				if err != nil {
```
ycliuhw, line 213: yeah, 👍 more accurate errors

ycliuhw (approved):
LGTM thanks

wallyworld:
$$merge$$

